PR #20, opened 2015-06-01
Improved clarity of save button

quillford:
Changed from "1.36m material" to "1.36m of material" for slightly improved clarity.


awhiemstra:
Can you move this to master branch? 15.06 is currently string frozen as there are translators working on translating it and I do not want to break all translations.

quillford:
Moved to master in #22.

PR #22, opened 2015-06-02
Improved clarity of save button

quillford:
Same as #20 but  moved to master branch.


awhiemstra:
Thank you, but just changing the string in the extracted translation files will break things the next time we extract messages. Can you perhaps change the source string in resources/qml/SaveButton.qml line 152?

mennoberg:
cura v8
buttons

PR #97, opened 2015-07-03
Prusa i3 Profile

quillford:
Not sure what to do for line 3

```
"icon": "icon_ultimaker2.png",
```
Where are the icons stored and what do they do?


nallath:
We don't use the icons anymore. We should probably remove that.

Issue #93, opened 2015-07-03
Layer slider has poor visibility


fieldOfView:
The slider used to select the layer to view uses the same color as the printer bed checkerboard and has poor visibility when the top layer is shown. The Layer view opens with the top layer visible by default. Depending on the view of the printerbed, it is easily possible to obscure the slider almost entirely.

Switching the backgroundcolor of the slider to the color the buttons have would increase visibility (though those buttons could do with a bit more contrast from the checkerboard grey as well).

nallath:
Thanks for reporting. If you want a quick fix, you can edit the theme.json file (in the resoureces folder). If I remember correctly you need to change the value of "slider_groove". The color is defined by RGBA.

fieldOfView:
Actually, it is slider_groove_border and slider_groove_fill; the light part of the slider is set with slider_groove. Would a pull-request be appreciated?

nallath:
Yup. Thats why we open source things :)

fieldOfView:
@nallath: against which branch? Master or 15.06?

nallath:
FYI: Master is the branch where we continue to commit / create new features. 15.06 is the release branch for bugfixes only.

PR #98, opened 2015-07-03
Improve visibility of layer slider 

fieldOfView:
The layer slider (in Layer viewmode) has poor visibility against the background if the slider overlaps the checkerboard in the 3d scene. This patch changes the color of the slider to the same color that is used for buttons, setting it apart from the checkerboard.

Fixes https://github.com/Ultimaker/Cura/issues/93

### DIFF
--- a/resources/themes/cura/theme.json
+++ b/resources/themes/cura/theme.json
@@ -96,8 +96,8 @@
         "progressbar_control": [12, 169, 227, 255],
 
         "slider_groove": [245, 245, 245, 255],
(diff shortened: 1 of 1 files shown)


fieldOfView:
Let me know if you would like me to resubmit this PR against the master branch. (See https://github.com/Ultimaker/Cura/issues/93#issuecomment-118793655)

nallath:
It's a bug fix, so i'm merging it with 15.06. Thanks for your contribution!

PR #112, opened 2015-07-06
Instructions for making profiles for non UM printers

quillford:


nickthetait:
Nice :smile:

nallath:
Thanks! :)

PR #113, opened 2015-07-06
modified README

quillford:
Removed the description that Cura is unreleased. Added a note as to where to put the JSON profile for custom machines.


BagelOrb:
Thanks! :)

PR #120, opened 2015-07-07
Hephestos for Cura 15.06

CoretechR:
Platform offset is [110,128,81.5]


quillford:
Is this profile necessary? Once editing machine properties from the GUI is added, the standard i3 profile should be close enough. The platform is the same as well.

CoretechR:
OK, Sorry. Didn't know about the editing menu. From what I have read on the forums I thought this feature would not return.

nallath:
What made you think that the feature would not return?

CoretechR:
e.g. https://ultimaker.com/en/community/view/16409-thoughts-about-the-new-cura?page=8
you said you would rely on the community to upload profiles, so I assumed that would be a permanent solution.
But if the settings are coming back that's awesome.
Do you have an ETA on when this will be implemented?

nallath:
We've always relied on the community to upload profiles. All profiles in the old cura (machine wise) were contributions. 
I don't know when this will be implemented but like i said, its not high on our list.

nicanor-romero:
At the moment in BQ we are adjusting and testing the profiles for our printers to get the best results. We will soon be making a pull request to add the profiles for the BQ Witbox and BQ Prusa i3 Hephestos.

nallath:
Allright. Just make a pull request when they are done.

Issue #88, opened 2015-07-02
could not create shader / coredump on 15.06.1 + Ubuntu 64 15.04


Patola:
Just downloaded Cura 64-bits 15.06.1 from http://ultimaker.com/en/products/cura-software
Installed ok on Ubuntu 64-bits 15.04 (had to uninstall libprotobuf-dev and a bunch of other -dev packages though).
When I try to run it, though, it spits out:

[2:46] [4290] [patola@ovo applications]% cura_app.py
2015-07-02 02:46:40,397 - INFO - Loaded plugin ConsoleLogger
2015-07-02 02:46:40,478 - INFO - Loaded plugin LocalFileStorage
2015-07-02 02:46:40,480 - INFO - Loaded plugin MeshView
2015-07-02 02:46:40,482 - INFO - Loaded plugin WireframeView
2015-07-02 02:46:40,484 - INFO - Loaded plugin LayerView
2015-07-02 02:46:40,485 - INFO - Loaded plugin OBJReader
2015-07-02 02:46:40,486 - INFO - Loaded plugin STLReader
2015-07-02 02:46:40,487 - INFO - Loaded plugin OBJWriter
2015-07-02 02:46:40,488 - INFO - Loaded plugin STLWriter
2015-07-02 02:46:40,490 - INFO - Loaded plugin GCodeWriter
2015-07-02 02:46:40,491 - INFO - Loaded plugin CameraTool
2015-07-02 02:46:40,493 - INFO - Loaded plugin TranslateTool
2015-07-02 02:46:40,496 - INFO - Loaded plugin MirrorTool
2015-07-02 02:46:40,498 - INFO - Loaded plugin SelectionTool
2015-07-02 02:46:40,501 - INFO - Loaded plugin ScaleTool
2015-07-02 02:46:40,516 - INFO - Loaded plugin RotateTool
2015-07-02 02:46:40,518 - INFO - Checking for new version of cura
2015-07-02 02:46:40,519 - INFO - Loaded plugin UpdateChecker
2015-07-02 02:46:40,525 - INFO - Loaded plugin USBPrinting
2015-07-02 02:46:40,531 - INFO - Loaded plugin CuraEngineBackend
2015-07-02 02:46:40,536 - WARNING - Could not find any translations matching default for file uranium_qt, falling back to english
2015-07-02 02:46:40,536 - WARNING - Could not find English translations for file uranium_qt. Switching to developer english.
[...]

PR #131, opened 2015-07-10
Fixes shader creation with Ubuntu+nvidia

Patola:
This fixes issue #88 with Ubuntu + nvidia

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
+from OpenGL import GL
 from UM.Qt.QtApplication import QtApplication
 from UM.Scene.SceneNode import SceneNode
 from UM.Scene.Camera import Camera


daid:
I do think this needs a tiny platform dependeny check, as on linux adding this is most likely fine (as the package is easy to install) but on Mac and Windows adding python-opengl as a dependency isn't that great. (And it will possibly break "angle" support on windows)

awhiemstra:
Hmm, the problem I have with this is that it is an Ubuntu-specific bug that we are now solving by adding a global dependency. A simple platform.system() == "Linux" check would at least reduce it to "just" Linux but for Fedora or Arch users that would still introduce an additional dependency. Ideally this would just be fixed upstream, but until it is I would be fine with introducing this workaround as long as it is limited to Ubuntu/Debian based systems.

Patola:
I will remake the patch. Thanks

thopiekar:
Think something like
```
if platform.system() == "Linux":
    if platform.linux_distribution()[0] in ("Ubuntu", ): # Just in case it also happens on Debian, so it can be added
        from OpenGL import GL
```

BurtonQin:
This problem occurred to me after I installed NVDIA driver. Thanks to the answers above, I found it easy to deal with.
1.**ldconf -p | grep libGL.so**
    I found that the first entry is still
    libGL.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libGL.so
    instead of 
    libGL.so (libc6,x86-64) => /usr/lib/nvidia-340-updates/libGL.so
2. **cd /etc/ld.so.conf.d**
    I found many conf files, including   
    i386-linux-gnu_GL.conf  and x86_64-linux-gnu_GL.conf
    **cat x86_64-linux-gnu_GL.conf**
    /usr/lib/nvidia-340-updates
    /usr/lib32/nvidia-340-updates
    but i386-linux-gnu_GL.conf is empty.
    So I just copy the contents in x86-linux-gnu_GL.conf to i386-linux-gnu_GL.conf
3. **sudo ldconfig**
    Everything is OK now☺

thopiekar:
I'll test it today.
Can you report this problem at launchpad? Would be great to have it fixed
in Ubuntu itself.
Another important question: to you still get this segmentation fault, when
closing Cura?

However, I would still keep this fix alive until Ubuntu fixes this problem.
What do you think?

PR #152, opened 2015-07-17
Added machine settings of German RepRap NEO machine

simonCor:
This is the machine: https://www.germanreprap.com/produkte/3d-drucker/neo-3d-drucker/

The configuration works fine fine for mine.


nallath:
Could you also add another platform model? Now it uses the UM2 platform, which might be a bit confusing.

simonCor:
I understand - I will see if I can a custom one for the Neo.

simonCor:
Ok. It now has a custom build platform

nallath:
Allright, thanks for putting in the work :+1:

TZanke:
Hello,

thank you for your work.

I found a problem with the orientation of the build platform. I have to rotate the objects 90° clockwise to get the expected result in prints.

Is it a problem of the custom build platform STL or X/Y zero positions in the JSON file?

Any idea how to fix this?

PR #161, opened 2015-07-21
BQ profiles update

nicanor-romero:
BQ printer profiles updated for Cura version 15.06.02


nallath:
15.06.02 is only for bug-fixes, not for new features. These are already in the main branch, so they will be in the next version.

PR #210, opened 2015-08-02
Fix poor visibility of MessageStack against grid

fieldOfView:
The background of the MessageStack is the same as the (hardcoded) dark color of the buildplate grid. When the view is rotated so the MessageStack overlays the buildplate grid, the visibility of the messagestack is poor. 

This PR changes the backgroundcolor of the MessageStack to a darker shade, so it stands out against the grid. It uses the same color as the buttons, which may or may not be ideal. I'm a bit conflicted about changing the message text color to white for better contrast. But perhaps a different shade should be picked for the MessageStack background?

### DIFF
--- a/resources/themes/cura/theme.json
+++ b/resources/themes/cura/theme.json
@@ -120,7 +120,7 @@
         "save_button_printtime_text": [12, 169, 227, 255],
         "save_button_background": [249, 249, 249, 255],
 
-        "message": [205, 202, 201, 255],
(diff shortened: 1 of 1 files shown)


TMHogenhout:
Hey Aldo, 
great contribution, thank you. I merged it with 15.06, so if there is a next release, it will actually have a visible messagestack ;) I will also pick it up with our UX-designer to prevent the same problem for the 15.10 release. Thanks again!

PR #214, opened 2015-08-03
Adding traceback.print_exc()

thopiekar:
Very useful when CrashHandler couldn't be found or isn't working at all.
The result is that a traceback message will be printed when a exception appears (in python style).

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -3,12 +3,15 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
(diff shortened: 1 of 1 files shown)


awhiemstra:
I have now updated the crash handler to print the exception if there is no QCoreApplication. This should take care of displaying the error when dealing with import or related errors. Please check that it works for you.

The reason I do not want to print an exception by default is that on Windows this will produce an unhelpful "The error log was saved as C:/Program Files/Cura/Cura.log" on any console output, due to a py2exe bug.

Issue #232, opened 2015-08-09
Question: How do I get the option for USB printing to appear?


KJHamblin:
Application Version: 15.06.02/03
Platform: Windows 7/8.1
Qt: Unsure
PyQt: Unsure
Display Driver: Unsure

Steps to Reproduce:
1.  Open Cura 15.06.XX
2.  Load .STL file
3.  Power on Arduino Mega 2560 based printer 
4.  Connect USB to computer

OR

1.  Power on Arduino Mega 2560 based printer 
2.  Connect USB to computer
3.  Open Cura 15.06.XX
4.  Load .STL file

OR

1.  Open Cura 15.06.XX
2.  Power on Arduino Mega 2560 based printer 
3.  Connect USB to computer
4.  Load .STL file

Actual Results:
"Save to Disk" is the only visible option.

Expected results:
"Print with USB" should over take "Save to Disk" and become blue.

Additional Information:
I have a RigidBotBig based on the Arduino Mega 2560 and now that I have my .json profiles complete I cannot get Cura 15.06.03 to give me the option to "Print with USB".  The only option that appears is "Save to Disk."  This is despite the fact that the printer shows up in Device Manager and connects to Pronterface and Cura 15.04 fine.  This is true on both Windows 7 and Windows 8.1.

Is it supported yet or am I doing something incorrect?  

Below is my .json file.  While I do not think this has any impact, I have included it for completeness:

**************************** RIGIDBOT .JSON PROFILE ****************************
```JSON
{

    "id": "rigidbotbig",
    "version": 1, 
    "name": "RigidBotBig",
    "manufacturer": "Invent-A-Part",
    "author": "RBC",
    "platform": "rigidbotbig_platform.stl",
    "platform_texture": "RigidBotBigbackplate.png",

    "inherits": "fdmprinter.json",

    "machine_settings": {

        "machine_width": { "default": 400 },
        "machine_depth": { "default": 300 },
        "machine_height": { "default": 254 },
        "machine_heated_bed": { "default": true },

        "machine_nozzle_size": { "default": 0.4},
        "machine_head_shape_min_x": { "default": 0 },
        "machine_head_shape_min_y": { "default": 0 },
        "machine_head_shape_max_x": { "default": 0 },
        "machine_head_shape_max_y": { "default": 0 },
        "machine_nozzle_gantry_distance": { "default": 0 },
        "machine_gcode_flavor": { "default": "RepRap (Marlin/Sprinter)" },
		
[...]

PR #240, opened 2015-08-10
Auto Baud Rate Detection Speed Up

KJHamblin:
In _connect:
    
1. The 1.5 second sleep command was moved outside the baud_rate for loop.  The bootloader is only accessible one time (at power) up so this added an extra 1.5 seconds to each loop cycle.

2. In place of the 1.5 second sleep command a 0.1 second sleep command was used to prevent an excessive amount of USB commands being set for each baud rate test.

3.  The time_out was reduced from 15 seconds to 1 second.  This should be sufficient and greatly reduces connection time, especially with multiple COM ports present.

### DIFF
--- a/plugins/USBPrinting/PrinterConnection.py
+++ b/plugins/USBPrinting/PrinterConnection.py
@@ -187,6 +187,7 @@ class PrinterConnection(SignalEmitter):
             Logger.log("i", "Could not establish connection on %s, unknown reasons.", self._serial_port)
             return
         
+        time.sleep(1.5) # Ensure that we are not talking to the bootloader. 1.5 sec seems to be the magic number
         # If the programmer connected, we know its an atmega based version. Not all that usefull, but it does give some debugging information.
         for baud_rate in self._getBaudrateList(): # Cycle all baud rates (auto detect)
             
@@ -199,9 +200,9 @@ class PrinterConnection(SignalEmitter):
             else:   
                 if not self.setBaudRate(baud_rate):
                     continue # Could not set the baud rate, go to the next
-            time.sleep(1.5) # Ensure that we are not talking to the bootloader. 1.5 sec seems to be the magic number
+            time.sleep(0.1) # Ensure that we are sending too many commands over USB
             sucesfull_responses = 0
-            timeout_time = time.time() + 15
+            timeout_time = time.time() + 1 #Set a 1sec time_out. 1 sec/ 0.1 sec = 10 commands sent per baud rate tested
             self._serial.write(b"\n")
             self._sendCommand("M105")  # Request temperature, as this should (if baudrate is correct) result in a command with "T:" in it
 


daid:
Did you test this with all possible firmware & hardware combinations?

There are a few odd and strange things in bootloaders and firmwares out there. Which need the longer startup and delays.

KJHamblin:
I have not.  Currently I only have access to an Atmega2560 based design (RigidBot) and it worked fine in that configuration.  I can try different firmware with this hardware and I can try to source more testing on different systems from different people.

Do you have any idea what firmware/hardware combinations are the outliers that would require a timeout of 15 seconds?  If you do I can attempt to target people with those configurations to source more testing.

KJHamblin:
Also as an FYI if you were unaware, this spawned from the issue https://github.com/Ultimaker/Cura/issues/232#issuecomment-129433097 that I created after updating to Cura 15.06

nallath:
I do believe that this is better than what we did before; KJHamblin reported checking times of 45(!) seconds because he has a large number of devices. That feels like a lot. 

Most of the USB printing code i simply 'stole' from the old Cura, without really checking why and what it was doing. Im going to defer to @daid 's judgement on this as he wrote it in the first place.

KJHamblin:
The 45 second delay was my best case scenario!  If the test started on the COM port of my printer (baud rate of 115200) it would need to cycle through 250000 (15 seconds to timeout) and 230400 (15 seconds to timeout).  So ok maybe it was only a 30ish second wait time.

I don't completely understand the code of ```PrinterConnection.py```, but I suspect it makes a list of all COM ports on the system and then cycles through a list of baud rates for every COM port.  At 15 seconds per baud rate test and maybe 4-6 baud rates tested (250000, 230400, 115200, 9600, etc?)  the delay will add up very quickly, especially with multiple COM ports.

@nallath That is fair about @daid's judgement.  I've always considered myself more of a hardware guy than software guy so I respect caution when reviewing and accepting my code.  Just let me know if there is anything I can do to help.

daid:
Note, the old code would case the found baudrate and port for the next run of Cura, so you would only have the long auto-detect times once, after that it tries to previous values first. @nallath is that still in your rewrite?

daid:
(And there are 2 major bootloader problems. One is a board which has a 10 second bootloader timeout. The other is a bootloader which "hangs" forever when it receives data during it's boot timeout time)

KJHamblin:
For those bootloader issues couldn't we keep my change of moving the 1.5 second delay out of the FOR loop, but increasing the delay to 11sec while keeping the time_out at 1 second?  That is still conservative since the device would need to be powered and connected to the computer prior to opening Cura.

awhiemstra:
One of the reasons the timeout is 15 seconds is because we had some serious problems on Mac OSX. That said, the USB printer code in master still needs to be ported to the OutputDevice API and thus currently is not being loaded. When doing that, we can probably also save the detected baud rate properly as it does not do that at the moment.

KJHamblin:
@awhiemstra forgive my ignorance on the inner workings of Cura as I haven't really dug into all of the code yet, but would porting the USB printer code to the OutputDevice AP remove it as a plugin to Cura?  I kind of like it as a plugin since the code is very easy to find, edit and test.  I think this is good since @nallath mentioned it is not an officially supported function.

Also do you have any idea whether the OSX issues had anything to do with Mac hardware or were they just related to OSX itself.  I'm asking because I do not have access to a Mac but I might be able to cobble together a hackintosh for testing.

On another, semi-related note, would someone be able to save me some time and point me to where ```_connect``` (or however the USB connection is initiated) is called in the main Cura application?  I want to see just how far down this USB rabbit hole I can go...

Ghostkeeper:
OSX issues are probably due to the OSX operating system rather than Macintosh hardware, since almost all our software engineers are working on Macintosh hardware that is provided by Ultimaker with some Linux version on it :P

nallath:
I'm currently porting it. It would still be a plugin, but a 'OutputDevicePlugin' (which we recently added). The porting will have quite a bit of impact on the code, so I'm going to close this pull request. I will have a close look at the suggested changes (speeding up the detection / saving previous found baud rates / etc).

KJHamblin:
Sounds good, thanks!

PR #241, opened 2015-08-10
Add RigidBot & RigidBotBig Profiles

KJHamblin:
This adds the RigidBot & RigidBotBig .json and .stl files necessary for
support in Cura 15.06


nallath:
Nice! Thanks for adding it.

PR #258, opened 2015-08-16
Correcting icon path

thopiekar:
Just fixes a simple typo I found.


nallath:
Nice find! Thanks for fixing it.

PR #333, opened 2015-09-04
Change UMO standard upgrades in Add Machine wizard

fieldOfView:
from https://ultimaker.com/en/community/view/17296-cura-1510-open-beta?page=1&sort=#reply-117393:
first opening-select your printer --> Ultimaker Original: by default only the dual extrusion upgrade box is ticked. Should it be the opposite by default only the extruder driver upgrade?

This PR restores the default upgrade to what it was with legacy Cura.


fieldOfView:
Sorry, borked PR.

PR #388, opened 2015-09-12
Fix a few typos in comments and tooltip

markwal:
nog to not

Perhaps to fix the tooltip it has to be in the English translation since it uses the English string here as the lookup for other languages?  If so, feel free to close without merging, obviously.

### DIFF
--- a/resources/qml/ViewPage.qml
+++ b/resources/qml/ViewPage.qml
@@ -35,14 +35,14 @@ UM.PreferencesPage
         }
         Button
         {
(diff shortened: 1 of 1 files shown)


nallath:
It uses the english string for the lookup, but the english needs to be correct as well (obviously). The master branch is our development branch, so your fix should have effect for the release after 15.10 (this would be 16.01).

markwal:
Ah, I thought maybe (without checking) that you might include a translation from English to English

nallath:
Ah, no. Originally we planned to do this (convert 'developer english' into 'english') but we dropped that for the moment as it would cost quite a bit of extra resources / time to do so.

Issue #406, opened 2015-09-19
[master] No open with entry for stl files


thopiekar:
So far I just opened stl files with Cura. Checked my filemanager and found this:

![cura-open-with](https://cloud.githubusercontent.com/assets/1847437/9976832/6759e870-5ef0-11e5-8575-632d1a0e6cd5.png)

I at least expected to find Cura there ("Öffnen mit"=="Open with")

---------------
Version: 15.09.82
Platform: Linux-4.2.0-10-generic-x86_64-with-Ubuntu-15.10-wily
Qt: 5.4.2
PyQt: 5.4.2
Installation using https://launchpad.net/~thopiekar/+archive/ubuntu/cura

thopiekar:
PR #407 gives now the following result:

![cura-open-with2](https://cloud.githubusercontent.com/assets/1847437/9977863/d698cafa-5f16-11e5-846c-c646316ea87a.png)

nickthetait:
:+1:

thopiekar:
Already merged, so closing...

PR #407, opened 2015-09-19
Adding stereolithography mime-type

thopiekar:
While finding the problem I reported at #406 I decided to fix it on my own.
Currently there is no icon for stl files on Linux, that's why I set it to "unknown". Otherwise it doesn't show no icon (probably I'm doing something wrong...).
However, *.stl files are now detected as stereolithography files now and after clicking on these files Cura opens and loads the file as expected.

_Problem here_, which needs to be fixed later: Opening a files creates a "cura.log" in the same directory as the stl file. Don't think this was planned, but if you are sure this is the right behaviour you can feel free merging this PR.

_Another problem_, which occurs is that neither the slicing is triggered nor generating the layers, when opening the layer view.


nallath:
We do want cura log to be in one location. It's already hard enough to get people to give us the logs without spreading them ;)

Ghostkeeper:
Seems to me that Uranium is saving the log in the "current directory" rather than the appdata, from this description, right?

thopiekar:
I would suggest to put logs always at $HOME/cura/cura.log as suggested for OSX. This will make people less confusing. Additionally OSX and POSIX are almost the same, so users would beleave the behaviour is the same on both OS's.

Don't know how the directories are handled in Cura. I would fix that for you if I know where find the part of the code :)

nallath:
@awhiemstra Should we merge this, knowing that it could create the logs in a wrong location?

thopiekar:
In my point of view adding this would give the problem that logs are created in $(pwd) more possibility to mess up all folders where stl files were opened.

I would wait for fixing all problems before.

Don't worry I won't remove this PR in the mean time :smiley:

awhiemstra:
What I am more wondering about in regards to this specific pull request is where you get the "application/sla" mime type from. http://www.iana.org/assignments/media-types/media-types.xhtml is the official list of registered mime types and does not contain "application/sla". As far as I know, mime types not in the x. namespace are meant to be registered and any unregistered mime type should be considered an illegal mime type and ignored.

thopiekar:
Yes, I just followed and guide how to add this file type. They suggested to
use a  template and the URL was part of it. Hope it is save to remove this
URL.

Haven't heard about that. Other applications, like a cd burning application
K3B, register their own mime types (even by inheriting other mime types.
I don't know why STL files are not have not been registered, because I read
many web servers even support this type.

thopiekar:
Took another look at /usr/share/mime/packages/. I have 45 files in there including one for VirtualBox, FreeMind, KDE, KDE5, Apport, APT and other. I would wonder if all of these projects are ignoring standards and curiously all of these files begin with (including the URL)
```
<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
```
https://github.com/Ultimaker/Cura/pull/407/files#diff-4a00f62c36f480614c6b54cadc146459R2

thopiekar:
Ok, by reading http://manpages.ubuntu.com/manpages/vivid/de/man1/dh_installmime.1.html it seems that sla.xml should not be installed by CMake, but by dh_installmime  (http://manpages.ubuntu.com/manpages/vivid/de/man1/dh_installmime.1.html ), which also creates needed postinst/postrm scripts for the packages. I'll open another PR with the only needed addition in /cura.desktop.

PR #410, opened 2015-09-20
display travel moves for current layer

fxtentacle:
This will re-add the move/retraction view that Cura 15.04.2 had by adding a "jump" mesh to the layer view and by adding two new polygon types: MoveCombingType and MoveRetractionType


nallath:
Thanks for the pull request. I'm not able to get the travel moves visible. Could you tell me what steps to take to see the difference.

Could you also add some documentation in the form of comments? This will help in the future for maintenance & such.

fxtentacle:
This is only the GUI update. To actually see the travel moves, you'll also need to recompile CuraEngine with https://github.com/Ultimaker/CuraEngine/pull/242

If you look into the LayerView.py, you'll see that I only added two enum values with colors and that I added _current_layer_jumps, which behaves exactly like the already-existing _current_layer_mesh, except that it contains the jumps and not the mesh.. so I feel that adding comments would be more or less superfluous, because I didn't introduce anything new. I merely extended the current display to handle 2 additional types of polygons.

nallath:
I don't see any travel moves; even with the recompiled engine.

fxtentacle:
I'm using the cura-build scripts and I had to rm -r dist before recompiling to make sure that the build scripts copy over the updated python into the GUI. Maybe you need to do that too?

fxtentacle:
And just to make sure we're clear about what to expect.. this is how it looks for me:
screen shot 2015-09-21 at 11 21 16
The thick light blue lines are moves while the nozzle is retracted. The thin blue lines are moves without extrusion and without retraction.

nallath:
I'm just running things directly from your branch, without any installer. Your screenshot is exactly the behaviour that we want :) 

I'll try to get it to work a bit more.

nallath:
Got it to work. Nice work :)

PR #420, opened 2015-09-22
Adding supported stl mime-type

thopiekar:
Related to the discussion at #407 just made the only needed modification to cura.desktop, that specifies which mime-types are supported by Cura.
This PR just adds the application/sla (as known as *.stl files) entry.

### DIFF
--- a/cura.desktop
+++ b/cura.desktop
@@ -8,5 +8,6 @@ TryExec=/usr/bin/cura_app.py
 Icon=/usr/share/cura/resources/images/cura-icon.png
 Terminal=false
 Type=Application
+MimeType=application/sla
 Categories=Graphics;
 Keywords=3D;Printing;


awhiemstra:
I was not really talking about the XML namespace but the the actual mime type. Maybe the term namespace is wrong but there are a few things like "application/vnd.oasis.opendocument.text" that use a "vnd." prefix to indicate a vendor specific "namespace". The way I read the IETF specs is that any "unofficial" mime type should use an "x." prefix, so in this case it would be "application/x.sla" as there is no official STL mime type.

Other than that, I was mostly wondering where you found that STL should use "application/sla" as mime type instead of "application/stl" or something similar. I could not really find an authoritative source about it.

awhiemstra:
Also, if you want to install the mime type file correctly, you might want to use https://projects.kde.org/projects/kdesupport/extra-cmake-modules/repository/revisions/master/entry/find-modules/FindSharedMimeInfo.cmake

thopiekar:
Found it on Wikipedia (in german): https://de.wikipedia.org/wiki/STL-Schnittstelle
And also at:
* https://my.bluehost.com/cgi/help/306
* http://www.hansenb.pdx.edu/DMKB/dict/tutorials/mime_typ.php
Which are no official sites, I know :disappointed:

awhiemstra:
Ah, the english version does not list a mime type so that got me confused. It is as good a reason as any to use that, so I'll merge this.

thopiekar:
Thanks

PR #450, opened 2015-10-01
Enable heated bed for i3 printer.

hmflash:
When selecting the i3 printer, the heated bed option is not displayed.  This fixes that.


nallath:
Thanks!

PR #456, opened 2015-10-03
prevent "GCode of death" ;)

fxtentacle:
I know that playing with master isn't always safe, but the cura master built managed to somehow produce a GCode, which will start with a HUGE extrusion without movements. The result is that my Ultimaker2 freezes and you can hear the extruder motor scream in pain. And afterwards, everything is clogged up and the nozzle is sticking to the heatbed and I needed to disassemble the feeder to get out all of the filament scraps. So that kind of GCode is REALLY annoying.

Here's the start of the buggy gcode file:

```
;FLAVOR:UltiGCode
;TIME:55356
;MATERIAL:62176
;MATERIAL2:0
;FLAVOR:UltiGCode
;TIME:44321
;MATERIAL:68468
;MATERIAL2:0
;LAYER:4
G0 X94.979 Y100.853 Z1.000
;TYPE:WALL-INNER
G1 F3000.000 X94.569 Y101.055 E4719.78767
G1 X93.072 Y98.014 E4720.05883
```

To me, that looks like Cura somehow started the slicing multiple times and then interleaved the results. But in any case, that E4719.78767 value is obviously a problem. As a first step towards finding the base issue that is causing bad GCode to be generated, I added a safety check to GCodeWriter.py which will log an error and abort if there is an extrusion difference > 10.0 between any two lines in the GCode.

So after applying this patch, Cura will cancel the GCode writing with an error message if there has been any screwup during slicing that resulted in overly large extrusion value jumps.

### DIFF
--- a/plugins/GCodeWriter/GCodeWriter.py
+++ b/plugins/GCodeWriter/GCodeWriter.py
@@ -5,6 +5,7 @@ from UM.Mesh.MeshWriter import MeshWriter
 from UM.Logger import Logger
 from UM.Application import Application
 import io
+import re
 
 
 class GCodeWriter(MeshWriter):
@@ -19,6 +20,23 @@ class GCodeWriter(MeshWriter):
         scene = Application.getInstance().getController().getScene()
         gcode_list = getattr(scene, "gcode_list")
         if gcode_list:
+            
+            # start extrusion safety check
+            check_lines = []
+            for layer in gcode_list:
+                check_lines.extend(layer.split("\n"))
+            cur_extrude = 0.0;
+            extrude_safety = 10.0;
+            for i, line in enumerate(check_lines):
+                match = re.search(r'E([0-9.]+)', line)
+                if match:
+                    extrude = float( match.group(1) )
+                    if extrude > cur_extrude + extrude_safety:
+                        Logger.log("e", "GCode is not safe! Extrusion gap from "+str(cur_extrude)+" to "+str(extrude))
+                        return False
+                    cur_extrude = extrude
+            # end extrusion safety check                        
+
             for gcode in gcode_list:
                 stream.write(gcode)
             return True


Ghostkeeper:
I think this warrants some public discussion on whether sanity checks on G-code should be the task of Cura's front-end. There are 3 options here, all with valid reasoning:

a) Cura checks the G-code it gets from CuraEngine, because the G-code it gets from CuraEngine is input and all input should be checked.
b) G-code checking is an example of post-processing, which should be a separate plug-in, possibly to be delivered by default along with Cura.
c) Cura should transparently show the G-code to the user since it is the task of the slicer to slice properly, i.e. the slicer is trusted.

I'm leaning most towards b) but that is slightly more work. This pull request could be merged for the time being until a post-processing plug-in is made. What are your thoughts?

daid:
Actually a check like this is in Marlin already.

fxtentacle:
I think it should be added, if only temporary, for two reasons:

1. To aid in spotting the underlying bug. I am fine with blindly trusting the slicing engine, but this probably wasn't caused by a bug in the slicer. There obviously is a race condition bug somewhere, which caused two CuraEngine results to get interleaved into one output file. As is, it is very hard to figure out "how did this happen" because the developer only notices after the fact, when the file has been fully written. So maybe this should even be moved a bit up in the chain inside Cura, so that we can throw an exception as soon as Cura receives the first wrong line from CuraEngine, or as soon as Cura mixes up the first two slicing results into one GCode stream. Since I believe that the bad GCode is caused by Cura's communication with the slicing engine, I would argue that this check should be inside Cura until the underlying communication bug (inside Cura) has been found and fixed. 

2. My Ultimaker2 still has the 15.04.2 firmware, that it arrived with. That firmware apparently does not have any protection against this kind of GCode, or the firmware-internal protection didn't work correctly. At least the GCode had a really bad impact on my machine and it froze the firmware in a state where I couldn't abort with the menu, but I could hear the motors screaming. That's obviously a super bad customer experience... I do think that Cura (since it's from Ultimaker too) should have a built-in protections to avoid causing damage to Ultimaker hardware. I agree that for the long term, that should probably be a post-processing plugin which is enabled by default. But I'm not sure if a post-processing plugin could throw an exception inside Cura early enough to aid in debugging (#1) so I think the plugin will have to come later.

awhiemstra:
Hmm, coming back to this, what I am mostly interested in is how this affects save performance. Adding a few seconds to a long GCode file does not really matter, as long as it does not cause the rest of the application to hang.

daid:
@awhiemstra are you sure you want parsing like this in the "frontend" code?
The current implementation of "max" 10mm extrusion will also break dual extrusion retractions which are more then 10mm.

awhiemstra:
Not really, no, as I see it, any engine bug causing this should just be fixed.

On the other hand, I can see some point to detecting errors in the gcode, since it makes things easier to debug. But we should probably have a proper backend testing suite for that instead.

awhiemstra:
@fxtentacle I am planning on adding a post-processing signal to the GCode writer that would allow you to write this as a plugin we can include in development builds instead of needing to integrate it in the GCode writer. Would that be an acceptable solution for you?

daid:
We can also add a check for "the GCode of death" in the CuraEngine/tests/runTests.py script, I've been setting up and running this script as a stability checker for the engine. Right now it throws away all the engine output, but it could verify the engine output.

This way we can catch this before it ever makes it in a release.

awhiemstra:
Yeah, that would be the best place to have it, in the automated tests we are setting up.

nickthetait:
I :heart: this news about automated tests

ChrisTerBeke:
This PR is so old it would be possible to hide a dead body here. Closing for now.

PR #479, opened 2015-10-13
Typo's in PrinterConnection.py

mpbomil:
Some of the "threading.Thread.daemon"s have typo's.


awhiemstra:
Indeed they do. Thank you for spotting them.

PR #480, opened 2015-10-14
Added profiles for 2nd Gen printers from BQ

nicanor-romero:


nallath:
Are we still accepting new machines for 15.10 @awhiemstra. I know we talked about this, but i cant quite recall what we decided back then.

awhiemstra:
We are. We decided that profiles are simply data files and as such have little impact on what the code does.

awhiemstra:
And merged. Thank you :)

PR #490, opened 2015-10-18
Add back placeholder expansion for start/end gcode

markwal:
See issue Ultimaker/CuraEngine#240

The intent of this change it to allow token replacement in start/end
gcode similar to the old Cura. Tokens are of the form
{material_bed_temperature} where the name is any setting name that is
passed to the engine.

In addition, four new tokens are added that expand to full gcode lines
for setting temperatures: {set_bed_temperature},
{set_bed_temperature_and_wait}, {set_extruder_temperatures}, and
{set_extruder_temperatures_and_wait}. These make it convenient to
control placement in the machine profile start/end gcode, but maintain the same
same generated temperature gcode as previously provided by
FffGcodeWriter.cpp

This change will generate the same temperature gcode as FffGcodeWriter
at the head, if none of the temperature tokens are present in the
machine profile. Otherwise, they're suppressed.

This change also depends on a corresponding change in CuraEngine that
removes the temperature gcode generation from FffGcodeWriter.cpp and
Wireframe2gcode.cpp as per suggestion from @BagelOrb in
Ultimaker/CuraEngine#240

I'm not sure I got the per object stuff correct because I'm not sure I tracked it properly to the backend.  The intention was to retain the logic that picks up the material temperatures from the per object setting that this replaces in FffGcodeWriter.cpp, but in my testing it always picked up the temperature from the global settings.


markwal:
Ah, I figured out some more about the per object settings.  I'll test more carefully with different objects using different extruders tomorrow and post my findings and/or updates tomorrow.  In the meantime, any comments or feedback welcome.

awhiemstra:
While you are doing processing before sending things to the engine, I would rather do this processing on save to GCode. Doing it on save would make sure there are no additional changes to the GCode and would allow you to do processing on every line of GCode instead of just the start/end. In addition, it would allow the engine to insert its own placeholders.

Next to that, I would prefer to have a generic "replace this placeholder with the value for this setting" rather than just limiting it to temperature values.

markwal:
Ok. Where do you suggest for post processing? The old Cura did it before.

It does any setting already.

BagelOrb:
@awhiemstra I wouldn't want the engine to insert placeholders. The engine should provide valid gcode. The post-processing plugins are generally used for quick hacks, which could better have been implemented in the engine itself.

I definitely don't see this as a reason not to merge this pull request (yet).

I do agree with @awhiemstra that the token replacements should be possible on all settings.


I think the token replacement procedure should consider only standard setting replacement; no adding new tags like ' set_extruder_temperatures_and_wait'.
Instead the code you deleted from the engine should be placed back and a couple of settings should be introduced: 'material_print_temp_wait' and 'material_bed_temp_wait' and  'material_print_temp_prepend' and 'material_bed_temp_prepend' . These should be boolean settings signifying whether to wait for the temperature to be reached and whether CuraEngine should prepend the temperature commands (or whether they are already included in the starting gcode). These last two boolean values should be computed by the start gcode parser. 

I hope you see why and how I would like to see the code now. Please ask if not.

awhiemstra:
@markwal I'd put it in plugins/GCodeWriter/GCodeWriter.py's write method. That way the replacement is only done once instead of every time it is sent to the engine. Note that you will need to construct the list of objects and groups manually in that case.

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -313,12 +328,83 @@ def _onChanged(self):
...
+            "set_bed_temperature_and_wait": "",
+            "set_extruder_temperatures": "",
+            "set_extruder_temperatures_and_wait": ""
+        }
+        if profile.getSettingValue("machine_gcode_flavor") == "UltiGCode":
+            return temp_gcode
+
+        if profile.getSettingValue("machine_heated_bed"):
+            temp = profile.getSettingValue("material_bed_temperature")
+            temp = int(temp) if temp else 0
```
awhiemstra: There should be no need to explicitly cast to int, that should already be taken care of by the Profile class.

awhiemstra:
> It does any setting already.

You are right, I overlooked that. Might be worth a small comment in the code.

markwal:
Seems like there isn't agreement on whether to do it in the gcode writer? I'm Ok implementing it either way.

If I add the new settings rather than new tokens and just do the conditional in the backend, this change gets dramatically simpler. In fact, if its ok with you, I'd postpone the wait conditional to a future PR and just do the prepend ones since that's just cleanly additional functionality.

markwal:
OK, I pushed a new version that puts the temperature gcode generation back to the backend but controls with the new settings (including the conditional wait) suggested by @BagelOrb.  Removed the per object loop from the frontend since now it serves no purpose.

Added the comment suggested by @awhiemstra and I think it is more clear now that I'm using settings directly and avoiding the "tokens" local. 

I haven't moved it to the gcode writer since I was waiting for consensus on that one.  I'd lean toward leaving it up front so that the engine generates complete and working gcode.

BagelOrb:
I think there is some confusion due to the fact that the frontend has a GCodeWriter.py and the CuraEngine has a (totally unrelated) file FffGcodeWriter.h /.cpp

It looks good from my end of the spectrum (CuraEngine).

Are you good to merge, @awhiemstra ?

Are `material_print_temp_wait` and `material_bed_temp_wait` included settings and are `material_print_temp_prepend` and `material_bed_temp_prepend` settings which are automatically set by the start gcode parser?

markwal:
Yes. The prepends are computed based on start gcode and the waits are settable in the profile.

markwal:
@awhiemstra Need anything from me?

awhiemstra:
Hmm, I still am not completely happy with doing this every time things are being sliced. Any chance you can at least cache the formatted start/end code so if no changes happen to temperature or start/end code we do not do all the formatting etc. again?

markwal:
OK, I can rig that up, but are you sure?  You'd take the risk of missing a cache invalidation and the extra code and logic for it over a string replacement with a hash lookup in a string that is lets say 1K at most against a slice that is going to generate stuff that is 3 orders of magnitude bigger?  Is that really where you want to spend extra logic?  I generally only want to add caches where I'm certain the memory is going to pay me pretty well for the saved code execution.

awhiemstra:
The problem here is that sending the slice data is done on the main thread and as such blocks the UI. If this were done on a thread - and python would do threads properly - I would have no issues with doing this every slice. However, we are already running into issues where the main UI blocks for half a second. As long as we do not have a proper solution for that, anything that can increase that delay needs a lot of thought. Since both the GCode and the temperature do not change often, caching the result seems a reasonably safe thing to do to gain speed at the cost of a little memory.

markwal:
Ah. Got it. Ok.

markwal:
Hang on. We have it so that it will do any setting, so that means cache invalidation has to happen when any setting change happens.  If when I format it, I keep track of which settings are referenced and then check if any of those settings have changed before doing the format, I think I'm going to be at a similar number of cycles.

Or I could keep track of any referenced setting and then distribute code wherever settings are set and do a lookup to see if the cache needs to be invalidated.  Is there such a choke point?  And where?

Or another possibility would be to have another profile setting that says whether token replacement is desired, then we can skip this for any profile that gets no benefit.

What do you think?

awhiemstra:
You can connect to the settingValueChanged signal of the active profile to track value changes of settings. This gets a string as parameter containing the key of the setting that changed. If you have a list of settings used in the token replacement, you can then see if the key is in that list and clear the cache.

Or alternatively, you could invalidate the cache on every setting change and see if that gets you acceptable speed. I am mostly worried about platform interactions here, where moving a model on the build plate can get quite slow. Platform interaction does not change settings so should not invalidate the cache.

awhiemstra:
@markwal I did some performance updates recently that include moving the entire slice start code into its own job, since this was a major bottle neck. With that, I do not really have any other objections to doing the start/end code processing before slicing without caching.

markwal:
Oh nice. I had intended to use your last suggestion and test, just hadn't got around to it yet.

markwal:
Rebased to master and moved the expansion to StartSliceJob.py

awhiemstra:
Thank you! And merged. I hope it was not too arduous. :)

markwal:
Thanks.  No problem.  Thanks for putting it in the job, keeps it simple.

plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -90,12 +105,28 @@ def run(self):
...
+            Logger.log("w", "Unabled to do token replacement on start/end gcode %s", traceback.format_exc())
+            return str(value).encode("utf-8")
+
     def _sendSettings(self, profile):
         msg = Cura_pb2.SettingList()
-        for key, value in profile.getAllSettingValues(include_machine = True).items():
+        settings = profile.getAllSettingValues(include_machine = True)
+        start_gcode = settings["machine_start_gcode"]
+        settings["material_bed_temp_prepend"] = not "{material_bed_temperature}" in start_gcode
+        settings["material_print_temp_prepend"] = not "{material_print_temperature" in start_gcode
```
BagelOrb: You missed a '}' in this line.
Mistake has been corrected.

resources/machines/fdmprinter.json
```diff
@@ -17,6 +17,12 @@
             "description": "Gcode commands to be executed at the very end - separated by \\n.",
             "default": "M104 S0\nM140 S0\n;Retract the filament\nG92 E1\nG1 E-1 F300\nG28 X0 Y0\nM84"
         },
+        "material_bed_temp_wait": {
+            "default": true
```
BagelOrb: You left out the descriptions of these new machine settings, but perhaps they were added for all machine settings in between you first wrote this code and when the PR was merged.
Has been fixed.

markwal: Oh yeah. Back in Oct 1, no descriptions. Funny i didn't notice on merging though. Now they stick out like a sore thumb.

PR #513, opened 2015-10-28
Fix unset text colors on bright backgrounds

ad1217:
Adds a text_default color for use where the background color is set to
something bright. This avoids falling back on the user's theme, in which
the default text color might be also be bright.


awhiemstra:
Nice work, thank you!

PR #533, opened 2015-11-10
Adjust initial view to be slightly from the side

fieldOfView:
The default view on the build volume shows the volume head-on to match how a user sees his/her printer. When an object is added to the center of the buildplate, this default view causes the y and z axes to align, making it harder to manipulate the object (translate, rotate, scale):
view1

Attached patch makes a small change to the default view to be slightly from the side, so the two axes don't align (and also places the camera slightly closer to the build volume so everything is less tiny):
view1_patched

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -169,7 +169,7 @@ class CuraApplication(QtApplication):
         self._physics = PlatformPhysics.PlatformPhysics(controller, self._volume)
 
         camera = Camera("3d", root)
-        camera.setPosition(Vector(0, 250, 900))
+        camera.setPosition(Vector(-80, 250, 700))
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I agree, but I understood that having the camera start head-on was an explicit request from @mennoberg.

mennoberg:
@Ghostkeeper This explicit request came out from that we want to match the orientation of the physical buildplate and virtual buildplate. Physical buildplate is only fully visible when standing in front, but I agree with @fieldOfView, functionality over orientation.

fieldOfView:
See the screenshot of the proposed new view; you are still not looking "through" a side pane, so the view remains a "natural" look on the full buildplate.

PR #560, opened 2015-11-30
Create wanhao_duplicator_i3.json

nielsswinkels:
I just got started with this printer. They say in the printers manual to use the prusa i3 settings, but that one has a different buildvolume. So I propose this file as it's own profile.

### DIFF
--- a/resources/machines/wanhao_duplicator_i3.json
+++ b/resources/machines/wanhao_duplicator_i3.json
@@ -1,0 +1,35 @@
+{
+    "id": "wanhao_duplicator_i3",
+    "version": 1, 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Congratulations on your new printer :wink:

Looks good. It uses the same buildplate mesh, which could be confusing?

BagelOrb:
Closing this because the json file is following the old standard.
It's a bit of a shame this PR slipped through the cracks.

@nielsswinkels If you make a new PR for a wanhao_duplicatori3**.def**.json according to the new format, we will gladly merge it.

PR #564, opened 2015-12-08
Render postprocessing

awhiemstra:
The Cura-related render postprocessing work.

### DIFF
--- a/plugins/SolidView/SolidView.py
+++ b/plugins/SolidView/SolidView.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2015 Ultimaker B.V.
-# Uranium is released under the terms of the AGPLv3 or higher.
+# Cura is released under the terms of the AGPLv3 or higher.
 
 from UM.View.View import View
 from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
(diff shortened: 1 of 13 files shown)


plugins/SolidView/SolidView.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2015 Ultimaker B.V.
+# Uranium is released under the terms of the AGPLv3 or higher.
```
nallath: Cura :)

plugins/SolidView/SolidView.py
```diff
@@ -0,0 +1,89 @@
...
+from UM.Application import Application
+from UM.Math.Color import Color
+from UM.Preferences import Preferences
+from UM.View.Renderer import Renderer
+
+from UM.View.GL.OpenGL import OpenGL
+
+import math
+
+## Standard view for mesh models.
```
nallath: How is this different from the simple view? I find solid view as a name a bit confusing (as the models displayed are not solid by definition). Solid models would be the stuff that open-scad generates or some voxel implementation.

plugins/SolidView/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2015 Ultimaker B.V.
+# Uranium is released under the terms of the AGPLv3 or higher.
```
nallath: Cura

plugins/SolidView/SolidView.py
awhiemstra: It is different in that it also does the checking if things are inside the build volume and displays overhang. The name "SolidView" comes from the label used it in the view switching dropdown, "Solid". Whether or not that makes sense is up for discussion.

awhiemstra:
This has now been merged.

daid:
Small notes, after this merge, I get an extra view mode called "Simple".
And I did this small change to improve error messages on shader errors:
https://github.com/Ultimaker/Uranium/commit/edefc178e6bed7d2b8a961e7a4b8bffe93c69478
(Was running into a shader compile error due to local edits, but the message did not really help)

PR #568, opened 2015-12-16
Write and Read Profile in g-code

Ghostkeeper:
This feature will write the profile to g-code when saving the g-code, and allows importing settings from g-code with the Import Profile button in the profile manager.

It also creates two new types of plug-ins: Profile Reader and Profile Writer, to fulfil said function as a plug-in (which is handy for later).

A similar pull request is made for Uranium. They should be tested together.

### DIFF
--- a/plugins/GCodeProfileReader/GCodeProfileReader.py
+++ b/plugins/GCodeProfileReader/GCodeProfileReader.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
-from UM.Application import Application #To get the current profile that should be updated with the settings from the g-code.
+from UM.Application import Application #To get the machine manager to create the new profile in.
 from UM.Settings.Profile import Profile
 from UM.Settings.ProfileReader import ProfileReader
 import re #Regular expressions for parsing escape characters in the settings.
--- a/plugins/GCodeProfileReader/GCodeProfileReader.py
+++ b/plugins/GCodeProfileReader/GCodeProfileReader.py
@@ -24,9 +24,9 @@ class GCodeProfileReader(ProfileReader):
     #   Note that the keys of this dictionary are regex strings. The values are
     #   not.
     escape_characters = {
-        "\\\\": "\\", #The escape character.
-        "\\n": "\n",  #Newlines. They break off the comment.
-        "\\r": "\r"   #Carriage return. Windows users may need this for visualisation in their editors.
+        re.escape("\\\\"): "\\", #The escape character.
+        re.escape("\\n"): "\n",  #Newlines. They break off the comment.
+        re.escape("\\r"): "\r"   #Carriage return. Windows users may need this for visualisation in their editors.
     }
 
     ##  Initialises the g-code reader as a profile reader.
@@ -40,7 +40,7 @@ class GCodeProfileReader(ProfileReader):
     #   specified file was no g-code or contained no parsable profile, \code
     #   None \endcode is returned.
     def read(self, file_name):
-        prefix = ";SETTING_" + str(version) + " "
+        prefix = ";SETTING_" + str(GCodeProfileReader.version) + " "
         prefix_length = len(prefix)
 
         #Loading all settings from the file. They are all at the end, but Python has no reverse seek any more since Python3. TODO: Consider moving settings to the start?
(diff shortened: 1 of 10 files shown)


plugins/GCodeProfileReader/GCodeProfileReader.py
```diff
@@ -0,0 +1,56 @@
...
+        super().__init__()
+
+    ##  Reads a g-code file, loading the profile from it.
+    #
+    #   \param file_name The name of the file to read the profile from.
+    #   \return The profile that was in the specified file, if any. If the
+    #   specified file was no g-code or contained no parsable profile, \code
+    #   None \endcode is returned.
+    def read(self, file_name):
+        version = 1 #IF YOU CHANGE THIS FUNCTION IN A WAY THAT BREAKS REVERSE COMPATIBILITY, INCREMENT THIS VERSION NUMBER!
```
awhiemstra: Please put this version number at the start of the class as a constant, as is done in Profile and other places. That makes it clearer.

plugins/GCodeProfileReader/GCodeProfileReader.py
```diff
@@ -0,0 +1,56 @@
...
+        prefix = ";SETTING_" + str(version) + " "
+        
+        #Loading all settings from the file. They are all at the end, but Python has no reverse seek any more since Python3. TODO: Consider moving settings to the start?
+        serialised = "" #Will be filled with the serialised profile.
+        try:
+            with open(file_name) as f:
+                for line in f:
+                    if line.startswith(prefix):
+                        serialised += line[len(prefix):-1] #Remove the prefix and the newline from the line, and add it to the rest.
```
awhiemstra: Save len(prefix) to a variable before doing the for loop, that saves a bunch of calls to len()

plugins/GCodeProfileReader/GCodeProfileReader.py
```diff
@@ -0,0 +1,56 @@
...
+        except IOError as e:
+            Logger.log("e", "Unable to open file %s for reading: %s", file_name, str(e))
+            return None
+
+        #Unescape the serialised profile.
+        escape_characters = { #Which special characters (keys) are replaced by what escape character (values).
```
awhiemstra: I generally prefer to have static data declarations like these as class attributes if possible, so that there is only one instance of them, instead of recreating the dict every time the function runs.

plugins/GCodeProfileReader/GCodeProfileReader.py
```diff
@@ -0,0 +1,56 @@
...
+                              #Note: The keys are regex strings. Values are not.
+            "\\\\": "\\", #The escape character.
+            "\\n": "\n",  #Newlines. They break off the comment.
+            "\\r": "\r"   #Carriage return. Windows users may need this for visualisation in their editors.
+        }
+        escape_characters = dict((re.escape(key), value) for key, value in escape_characters.items())
```
awhiemstra: Why not insert items in the dictionary directly using re.escape()? That would avoid a messy-looking inline for loop.

plugins/GCodeWriter/GCodeWriter.py
```diff
@@ -21,6 +22,40 @@ def write(self, stream, node, mode = MeshWriter.OutputMode.TextMode):
...
+
+    ##  Serialises the profile to prepare it for saving in the g-code.
+    #
+    #   The profile are serialised, and special characters (including newline)
+    #   are escaped.
+    #
+    #   \param profile The profile to serialise.
+    #   \return A serialised string of the profile.
+    def _serialiseProfile(self, profile):
+        version = 1 #IF YOU CHANGE THIS FUNCTION IN A WAY THAT BREAKS REVERSE COMPATIBILITY, INCREMENT THIS VERSION NUMBER!
```
awhiemstra: See above. But maybe you can use Profile.ProfileVersion instead of using a different version? That would avoid having to keep these two in sync.

plugins/GCodeWriter/GCodeWriter.py
```diff
@@ -21,6 +22,40 @@ def write(self, stream, node, mode = MeshWriter.OutputMode.TextMode):
...
+        #Escape characters that have a special meaning in g-code comments.
+        escape_characters = { #Which special characters (keys) are replaced by what escape character (values).
+                              #Note: The keys are regex strings. Values are not.
+            "\\": "\\\\", #The escape character.
+            "\n": "\\n",  #Newlines. They break off the comment.
+            "\r": "\\r"   #Carriage return. Windows users may need this for visualisation in their editors.
+        }
+        escape_characters = dict((re.escape(key), value) for key, value in escape_characters.items())
+        pattern = re.compile("|".join(escape_characters.keys()))
+        serialised = pattern.sub(lambda m: escape_characters[re.escape(m.group(0))], serialised) #Perform the replacement with a regular expression.
```
awhiemstra: It seems silly that you need to do all this regular expression magic for some simple escaping. But I suppose there is not really an easier solution...

plugins/GCodeWriter/GCodeWriter.py
Ghostkeeper: I'm moving this version number up to the start of the class.

But I do think it's generally a good idea to keep this separate from `GCodeProfileReader`, so that the developer must explicitly update the version of the reader in order to update the profile file format. If the version of the reader is not updated, it should then None instead of interpreting the file format wrongly.

plugins/GCodeWriter/GCodeWriter.py
Ghostkeeper: Right now it does all the replacements with one big regular expression. I looked for a better solution but the only real alternative is to iterate over the elements of the dictionary and explicitly substitute one by one, but that has the disadvantage that the result of the first substitution can be interpreted as an escaped character in the second: `\\n` gets unescaped to `\n` which gets unescaped to a newline character.

Python has built-in methods for escaping and unescaping for python code, URLs and a few others, but not for g-code. There is no built-in method for replacing multiple strings at the same time.

PR #570, opened 2015-12-18
Import Legacy Settings

Ghostkeeper:
This feature allows importing profiles from old Cura versions (version 15.04.03 and earlier). In the profile manager of Cura, simply click import and select current_profile.ini from the legacy Cura configuration folder, and the settings should be translated to the new Cura version.


plugins/LegacyProfileReader/DictionaryOfDoom.json
```diff
@@ -0,0 +1,75 @@
+{
+    "source_version": "15.04",
+    "target_version": 1,
```
nallath: Shouldn't that be a 2?

Ghostkeeper: No, it's actually the profile version, which is still at 1, not the API version. The `source_version` is not used at the moment by the way.

PR #574, opened 2015-12-21
Fixing error message, which appears when loading Cura.

thopiekar:
It happens because the variable version has just one entry.
In case of a unversioned aka master codebase the only content is "master", so there is no second entry at all.
Now a len(version) checks whether there is a second entry or not.

I also wanted to fix the same in master, but it is already fixed there - https://github.com/Ultimaker/Cura/commit/e7fc881440a4b92598fb4bf64648c96b47e17bb5

### DIFF
--- a/cura/CuraSplashScreen.py
+++ b/cura/CuraSplashScreen.py
@@ -21,8 +21,12 @@ class CuraSplashScreen(QSplashScreen):
 
         painter.setFont(QFont("Roboto", 20))
         painter.drawText(0, 0, 203, 230, Qt.AlignRight | Qt.AlignBottom, version[0])
-        painter.setFont(QFont("Roboto", 12))
-        painter.drawText(0, 0, 203, 255, Qt.AlignRight | Qt.AlignBottom, version[1])
+
+        # "master" release does not have an second entry in variable version.
+        # In this case skip to load version[1]...
+        if len(version) > 1:
+            painter.setFont(QFont("Roboto", 12))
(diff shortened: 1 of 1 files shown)


nallath:
Thanks :)

thopiekar:
No problem. Just noticed it when I saw my problem with the libGLESv2 error
;)

fieldOfView:
Note that this was already fixed by @daid in the master branch, and by me in the 2.1 branch.
The 15.10 branch is pretty stale.

thopiekar:
Good that you mention that. Why is 15.10 that stale? Shouldn't it be the
next release?
And what is 2.1 for? Is it an updated 15.04 branch?

I'm a bit confused :/

Ghostkeeper:
You're answering your own question! But it is confusing indeed.

Version 15.10 was decided in the end to never be released. It was only a closed beta. 2.1 is progressed from 15.10 and 15.06 (so it has the new structure with Uranium and all), and it is scheduled to replace 15.04 for good.

thopiekar:
Ahh, ok. And the next release in 2016 will still have the format 16.XY?

fieldOfView:
No, from 2.1 on we have semantic versioning! MajorVersion.MinorVersion.BugfixVersion

thopiekar:
Thank you for the info. Now I feel almost up-to-date :)
According to this I corrected my PPA at launchpad.net, so 2.1 is used for
stable builds.

Regards

PR #575, opened 2015-12-22
Terminate engine when a ToolOperation starts

fieldOfView:
Currently if the user starts using a Tool (eg the Move tool by moving an object, or the LayFlat tool) while a slice operation is ongoing, the slice operation continues during the operation. This unnecessarily loads the CPU, and makes the UI less responsive than it could be, because after the operation has finished we can be certain that a reslice is necessary.

This PR terminates the engine process when a ToolOperation starts, thus freeing up CPU cycles to perform the ToolOperation. It also makes "continuous slicing" slightly less obnoxious because it tries to get out of the way more.


nickthetait:
:+1:

Issue #229, opened 2015-08-07
[master] WARNING - Could not find any translations matching default for file uranium_qt, falling back to english


thopiekar:
Would be interesting where Cura expects it's i18n files. However, this is the message I get here:
```
2015-08-07 13:11:30,584 - WARNING - Could not find any translations matching default for file uranium_qt, falling back to english
2015-08-07 13:11:30,585 - WARNING - Could not find English translations for file uranium_qt. Switching to developer english.
2015-08-07 13:11:30,587 - WARNING - Could not find any translations matching default for file cura_qt, falling back to english
2015-08-07 13:11:30,587 - WARNING - Could not find English translations for file cura_qt. Switching to developer english.
```
And the location of the files, while installing the files via CMake:
```
thopiekar@home:~$ ls /usr/share/cura/resources/i18n/ -la                                                                                                               
insgesamt 84                                                                                                                                                           
drwxr-xr-x 7 root root  4096 Aug  6 22:32 .                                                                                                                            
drwxr-xr-x 8 root root  4096 Jul 31 23:09 ..                                                                                                                           
-rw-r--r-- 1 root root  2818 Aug  6 18:51 cura.pot                                                                                                                     
-rw-r--r-- 1 root root  8515 Aug  6 18:51 cura_qt.pot                                                                                                                  
drwxr-xr-x 2 root root  4096 Aug  6 22:32 de
drwxr-xr-x 2 root root  4096 Aug  6 22:32 es
-rw-r--r-- 1 root root 40688 Aug  6 18:51 fdmprinter.json.pot
drwxr-xr-x 2 root root  4096 Aug  6 22:32 fi
drwxr-xr-x 2 root root  4096 Aug  6 22:32 fr
drwxr-xr-x 2 root root  4096 Aug  6 22:32 ru
```
Same for uranium:
```
thopiekar@home:~$ ls /usr/share/uranium/resources/i18n -la
insgesamt 48
drwxr-xr-x 7 root root 4096 Aug  5 18:36 .
drwxr-xr-x 5 root root 4096 Aug  3 18:17 ..
drwxr-xr-x 2 root root 4096 Aug  5 18:36 de
drwxr-xr-x 2 root root 4096 Aug  5 18:36 es
drwxr-xr-x 2 root root 4096 Aug  5 18:36 fi
drwxr-xr-x 2 root root 4096 Aug  5 18:36 fr
drwxr-xr-x 2 root root 4096 Aug  5 18:36 ru
[...]

PR #576, opened 2015-12-23
Fixing i18n build and install

thopiekar:
* removes custom target copy-translations
* updating TODO to remove the language list with a nice algorithmus
* renaming file to po_file in foreach-loop
* etc.

Fixes issue #229

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,8 @@ if(NOT ${URANIUM_SCRIPTS_DIR} STREQUAL "")
         # build directory to the source resources directory. This is mostly a convenience
         # during development, normally you want to simply use the install target to install
         # the files along side the rest of the application.
-        add_custom_target(copy-translations)
 
-        #TODO: Properly install the built files. This should be done after we move the applications out of the Uranium repo.
+        #TODO: Find a way to get a rid of this list of languages.
         set(languages
             en
             x-test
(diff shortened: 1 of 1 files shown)


thopiekar:
Any thoughts? Feedback?

Ghostkeeper:
Looks okay to me, you've made the distinction between mo-files and po-files more clear at least. But I don't dare to touch this while @awhiemstra is on vacation. He'll be back next year (in a week, that is).

TMHogenhout:
I've tested it by creating both .pot and .po files with the fix and without the fix. The only difference is that it doesn't break on non-standard symbols, basically everything goes except for the double quote. 
weirdcharactersetfromhell

thopiekar:
Hmm, I don't see why my fix should break pot to po generation. The only
part I touched is the po to mo generation. Or did I get you wrong?

Could you tell how you usually generate the (po) files? Would be great if I
could reproduce it here.

Thank you for your feedback :)

TMHogenhout:
I have no idea how your fix, fixes this seemingly unrelated issue, but I'm very happy that it did. We had the problem that we couldn't use certain symbols in the json translatable strings, even though .json -> .pot is UTF-8 -> UTF-8. So under the hood something was going wrong and somehow your fix, fixed that. 

I have tested it several times. Before the .pot file would simply be cut off right after the line where there would be for example a quote (very annoying). So yeah this really saves me a lot of work. Thanks!

How we generate the po files? Old school msgmerge

thopiekar:
Maybe the magic was in renaming the variable "file" into "po_file". After
reading in the CMake docs that "file" is already some kind of function I
worried about leaving this.

No problem. Can you recommend a po-editor I can use on Linux? Additionally,
will it be needed to use msgmerge to update the po file?

TMHogenhout:
Unfortunately once I merged it; the magic was gone. It still worked but the side-effect of the script not breaking on non-standard symbols disappeared. So then it must be something happening (or not happening) in your branch. Well hopefully I can get a second look after the 2.1 release.. 

I would love to, but I don't really use a po-editor. I think it is needed to use something like msgmerge, because if your don't merge the old .po with the new .pot; you have to ask for completely new translations every time. Now we only need new translations for the strings that have been added. You can see the result here: https://github.com/Ultimaker/Cura/blob/2.1/resources/i18n/de/cura.po

On line 165 there is an empty translation; that's a new still to be translated string. And in the bottom of the file you can see strings that are no longer used (but we don't throw them away because maybe they'll be used again in the future). 

Thanks again!

thopiekar:
No problem :+1: and thank you for your great explanation.

Ah, ok, so I would need to run msgmerge on my de language now, too. I think it would be helpful to have a command for updating po with a newer pot.
Well, po-editor is only for editing po files not to update them from pot files, as far as I know.

I played with that application and I wonder that po-editor is converting multi-line entries into single-line entries. See https://github.com/thopiekar/Cura/commit/739baacf86e0af5f0d6af4d547e0f6d2e7ea9946

TMHogenhout:
or you just wait till the new translations come online in a few weeks from now.. Then you don't have to translate it either.. Obviously you can only do this if you have roughly the same translated strings. 

O yes it would be very helpful to have a command to update the po files with the newer pot files. That's basically the only step in the workflow that's not automated. And I actually found it to be a lot of work to do this with the command line for all the different files. Also msgmerge does a perfect job, I only have to adapt things like the id afterwards.

That's one hell of a good point there. I tested whether that would be a problem by reproducing it manually on the x-test thing for both the fdmprinter language file and cura language file. As you can see no problem there (pfew).
testcura1
testcura2
testfdm1
testfdm2

PR #582, opened 2016-01-04
To add a support for Malyan M180 to Cura 2.1

khudko83:
Kindly ask you to add a support for Malyan M180 to Cura 2.1
I believe it will be helpful for a lot of users!
Please find attached json file for M180, which was developed by Malyan and worked with Cura 15.06.

[m180.txt](https://github.com/Ultimaker/Cura/files/77364/m180.txt)

PR #583, opened 2016-01-04
Malyan180 support

Ghostkeeper:
This is a translation of the machine that Malyan made for version 15.06. The settings should be the same as what Malyan provided back then.

See also https://github.com/Ultimaker/Cura/pull/582, which was not a pull request but a request for work. I just did the translation.

### DIFF
--- a/resources/machines/m180.json
+++ b/resources/machines/m180.json
@@ -1,0 +1,38 @@
+{
+    "id": "m180",
+    "version": 1,
+    "name": "Malyan M180",
+    "manufacturer": "Malyan",
+    "icon": "icon_ultimaker.png",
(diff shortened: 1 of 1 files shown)


BagelOrb:
The machine settings should be overrides if you want to merge this file to master...

nallath:
My bad. I'll fix that.

fieldOfView:
What's the point of having this profile if the Malyan M180 needs a .x3g file (instead of .gcode), which we don't create.

BagelOrb:
There is a tool which converts .gcode to .x3g. People are free to use that tool however they like. 
https://github.com/whpthomas/GPX

PR #588, opened 2016-01-12
Don't include origin in boundingbox

fieldOfView:
Fixes CURA-679

AxisAlignedBox() initializes as a 0 width, 0 height box around the origin.


soldanr:
I would love to see the origin on the bounding box :+1:

fieldOfView:
@soldanr please make a separate issue for that, because this PR is (1) already closed and (2) about something else than your suggestion.

PR #592, opened 2016-01-14
Scanning i18n directory for languages

thopiekar:
Instead of having a list of languages hardcoded, we scan the i18n directory for available languages.
It was a TODO, which was left after fixing po-file installation. After finding the macro I decided to fix it here.
The reference to the added macro is there as an URL.


thopiekar:
@TMHogenhout I'm sure you will like to test this :smile:

Ghostkeeper:
Looks good to me. I'll test it and if it works I'll merge it.

thopiekar:
Ok, so would you like to have the same in Uranium?

TMHogenhout:
Yes please :smiley:

PR #603, opened 2016-01-19
Adding new supported image formats

thopiekar:
Don't know how long we support opening images, but this commit adds them to the *.desktop file for Linux based desktops.
So "open with"-actions will now show Cura for these media-types.


Ghostkeeper:
Good idea. We should also check if 3MF and OBJ need to be added.

thopiekar:
Have them in my own packaging files, which are overwriting these provided here. I just added the mime types for images here, because I'm sure they are called like this.

I can add another PR with the other mime types if you like.

PR #612, opened 2016-01-23
Alternative fix for NVidia graphic cards

thopiekar:
As I upgraded my computer to a developer version of Ubuntu, I noticed that the OpenGL module, which is needed for our workaround, is currently broken.
So I browsed the web and found the reason why it is needed at all and found a alternative.
The reason for the problem is that Qt5 is dynamicly loading libGL.so instead of libGL.so.1, as the OpenGL module loads libGL.so.1.
So if you install the closed-source nvidia drivers it only creates a link from libGL.so.1 to it's binaries and the result is that PyQt5/Qt5 tries to load Mesa binaries together with NVidia binaries.
By importing the OpenGL module you preload the libGL.so.1, but this can also be done directly by using ctypes.

* Replaced the OpenGL fix with the ctypes fix
* Added a TODO

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -57,8 +57,10 @@ numpy.seterr(all="ignore")
 
 if platform.system() == "Linux": # Needed for platform.linux_distribution, which is not available on Windows and OSX
     # For Ubuntu: https://bugs.launchpad.net/ubuntu/+source/python-qt4/+bug/941826
-    if platform.linux_distribution()[0] in ("Ubuntu", ): # Just in case it also happens on Debian, so it can be added
-        from OpenGL import GL
+    if platform.linux_distribution()[0] in ("Ubuntu", ): # TODO: Needs a "if X11_GFX == 'nvidia'" here. The workaround is only needed on Ubuntu+NVidia drivers. Other drivers are not affected, but fine with this fix.
+        import ctypes
+        from ctypes.util import find_library
+        ctypes.CDLL(find_library('GL'), ctypes.RTLD_GLOBAL)
 
 try:
     from cura.CuraVersion import CuraVersion


nallath:
The link in your comment mentiones a bug in qt4. (For Ubuntu: https://bugs.launchpad.net/ubuntu/+source/python-qt4/+bug/941826). Is this also an issue for qt5 (if so, i'll merge it)

thopiekar:
Yes, the bug is still present in Qt5. I even think it is more driver related, as we get these segmentation faults, when using the closed-source nvidia driver.

Looks like there is something terribly wrong with the driver that both problems appear.

awhiemstra:
I am tempted to ask you to put this into a file called "FixUbuntusStupidShit.py", since it is completely ridiculous that these packaging issues cause such major breakage. Especially if they now managed to break python-opengl as well. Though I recall @daid mentioning they did that before. 

On a more serious note, the fix looks OK. Though this: "# TODO: Needs a "if X11_GFX == 'nvidia'" here." might prove to be quite a lot more tricky than you'd think. :)

thopiekar:
The problem with Python-OpenGL in Ubuntu Xenial (development release) is a problem related to Python 2.5, which is the default Python 3 interpreter there.
Importing OpenGL, as we did before, does exactly the same as my commit here, but ctypes is a function provided by Python itself, so we don't need Python-OpenGL anymore.

I just wonder about other projects. How do they handle these problems?! Most interesting is how do they catch the segmentation fault?!

daid:
In 3.5 years, they broke numpy 3 times, and python-opengl twice. That I know of.

thopiekar:
Seems to be the best combination we get here: Ubuntu + Python :smile: 

However, what do you think about this PR? It would drop our dependency on Python-OpenGL...
_And:_ If I find some time for it I'll try to define better if-clauses for the next release (e.g. 2.1) and move it to Uranium, where we actually use Qt5's OpenGL class/module :wink:

thopiekar:
A user just reported an alternative fix by modifying the system files.
It would make this PR and the old fix not needed anymore, but it still does not fix the segmentation faults.
https://github.com/Ultimaker/Cura/pull/131#issuecomment-176088664

thopiekar:
Now got an answer by the maintainer of the nvidia-driver and he is sure that there are so issues.
So I wrote him about our errors and also sent him the output of an PyQt5 example:
```
thopiekar@home:~$ /usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 142, in initializeGL
    self.gl = self.context().versionFunctions()
ImportError: No module named 'PyQt5._QOpenGLFunctions_4_5_Compatibility'
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 166, in resizeGL
    self.gl.glViewport((width - side) // 2, (height - side) // 2, side,
AttributeError: 'GLWidget' object has no attribute 'gl'
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 152, in paintGL
    self.gl.glClear(
AttributeError: 'GLWidget' object has no attribute 'gl'
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 152, in paintGL
    self.gl.glClear(
AttributeError: 'GLWidget' object has no attribute 'gl'
QOpenGLShaderProgram: could not create shader program
QOpenGLShader: could not create shader
bool QOpenGLTextureBlitter::create() Could not link shader program:
 ""
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 152, in paintGL
    self.gl.glClear(
AttributeError: 'GLWidget' object has no attribute 'gl'
  File "/usr/share/doc/pyqt5-examples/examples/opengl/hellogl.py", line 152, in paintGL
    self.gl.glClear(
AttributeError: 'GLWidget' object has no attribute 'gl'
```

If we are lucky we can even remove the workaround at all. Is there some questions left about this PR?

nallath:
Nope! Merging it :)

thopiekar:
Thanks. I asked the maintainer of the nvidia package and he said that this
isn't actually a problem with the driver package but a problem in PyQt5. It
seems that PyQt5 isn't handling the driver correctly. He is sure it needs
to be fixed in PyQt5 like it has been already done in PyQt4.

nallath:
Well, we could try to write a patch for pyqt 5 and try and get that upstream. Not sure how hard that would be though.

thopiekar:
Just googled for commits and patches. Best of all is that PyQt people
beleave again it is related to the nvidia packaging.

I will contact the maintainer again, what he thinks about the feedback, eg.
in this bug report:
https://bugs.launchpad.net/ubuntu/+source/nvidia-graphics-drivers-319-updates/+bug/1241988

nallath:
Cool. Thanks for all the work!

PR #613, opened 2016-01-23
Adding support for additional object files

thopiekar:
* 3MF files
* OBJ files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,9 @@ if(NOT APPLE AND NOT WIN32)
     install(DIRECTORY cura DESTINATION lib/python${PYTHON_VERSION_MAJOR}/dist-packages FILES_MATCHING PATTERN *.py)
     install(FILES ${CMAKE_BINARY_DIR}/CuraVersion.py DESTINATION lib/python${PYTHON_VERSION_MAJOR}/dist-packages/cura)
     install(FILES cura.desktop DESTINATION ${CMAKE_INSTALL_DATADIR}/applications)
+    install(FILES cura.sharedmimeinfo
+            DESTINATION ${CMAKE_INSTALL_DATADIR}/mime/packages/
+            RENAME cura.xml )
 else()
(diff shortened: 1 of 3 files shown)


thopiekar:
I wasn't sure about the name for obj files, so I took "application/prs.wavefront-obj" for traditional *.obj files and I also found "application/vnd.ms-3mfdocument" for 3MF files, as found here:
```
application/vnd.ms-3mfdocument			3mf
```
https://git.fedorahosted.org/cgit/mailcap.git/plain/mime.types

But this page is telling something else: http://fileformats.archiveteam.org/wiki/3MF

What do you think?

Ghostkeeper:
I'm guessing you found http://stackoverflow.com/questions/18641754/mime-type-of-wavefronts-obj-and-mtl too.

thopiekar:
Correct, but I didn't take '''text/plain''', because every file format
(.txt and all other based on '''text/plain''') would take Cura as
supporting application. So I created a XML, which explains that all
.obj files are called prs.[...] and is based on text/plain, so you can
still open .txt with text-editors and .obj with text-editors and Cura.

I'll add this XML during this day.

awhiemstra:
http://www.iana.org/assignments/media-types/application/vnd.ms-3mfdocument
So IANA agrees with the mime type you found. And since they are the final registration point of mime types, that seems the correct one to use. You might want to file a bugreport for shared-mime-info that points them to that type, since shared-mime-info should have all IANA defined types by default iirc.

As for OBJ, unfortunately, like STL there is no proper defined mime type for it. Ideally we'd have some additional lines to distinguish them from text files by more than just the extension, but that might prove tricky.

thopiekar:
Ah, ok. I was just google'ing for these mime types, but didn't know about a official resource.
Good to know this mime-type is correct.

Correct, these files just don't simply start with a specific header.
@awhiemstra: What do you think? Could we keep it like this or should we wait, until there is a better solution?

Ghostkeeper:
This is fine for now. I don't think a better solution will present itself in the near future.

PR #618, opened 2016-01-29
Profiles rework

fieldOfView:
This is a reimagining of how profiles work in Cura. Instead of automatically creating new profiles when the user tries to change a setting on a "readonly" profile and having multiple changed profiles, there is only one "working profile" per machine type. This is more akin to how Legacy Cura worked.

This PR also implements different profiles for different printer types, variants (nozzle sizes in the UI) and materials. Finally it contains numerous little UX improvements related to profiles.

Also see Uranium PR: https://github.com/Ultimaker/Uranium/pull/98

Fixes or contributes to:
CURA-547
CURA-430 (hopefully)
CURA-805
CURA-199
CURA-402
CURA-583
CURA-626
CURA-24
CURA-751
CURA-569
CURA-568
CURA-432


resources/machines/ultimaker2_extended_plus_025.json
```diff
@@ -7,7 +7,8 @@
     "platform": "ultimaker2_platform.obj",
     "platform_texture": "ultimaker2plus_backplate.png",
     "inherits": "ultimaker2_extended_plus.json",
-    "variant": "0.25mm Nozzle",
+    "variant": "0.25 mm",
```
Ghostkeeper: Why did you remove the `Nozzle` text from these variant names?

resources/qml/LoadProfileDialog.qml
```diff
@@ -1,64 +0,0 @@
...
-        }
-        Button
-        {
-            text: catalog.i18nc("@action:button","Cancel");
-
-            onClicked: base.visible = false;
-        }
-    }
-}
-
```
Ghostkeeper: Why is LoadProfileDialog removed? I feel like I'm missing something here.

resources/machines/ultimaker2_extended_plus_025.json
fieldOfView: In the UI, the field is labeled "Nozzle" now. So the full UI would read:
"Nozzle: 0.25mm Nozzle"

resources/qml/LoadProfileDialog.qml
fieldOfView: LoadProfileDialog was never used in Cura. I added it two weeks ago, as a last ditch attempt to get some strings in before the string freeze. Unfortunately the UX changed during development of the feature (plus I needed the strings in Uranium, not in Cura).

resources/machines/ultimaker2_extended_plus_025.json
Ghostkeeper: Ah, right. Nevermind then :)

PR #620, opened 2016-01-29
Fast Simple<->Advanced Switch

awhiemstra:
Rather than reloading the panel every time the mode changes, use a StackView and create static instances of SimpleMode and AdvancedMode.

Requires https://github.com/Ultimaker/Uranium/pull/102


nallath:
Looks good. Can be merged

PR #625, opened 2016-02-01
i18n: de: Set of (better) translations for german

thopiekar:
I just reviewed the translations and some some parts, which I (as native speaker) would translate another way.
The commits in this PR should be good documented.
Would love to see feedback ( maybe from @TMHogenhout ) to understand why you exactly translated things your way.
For example "X-Ray" is untranslated, but I don't see a reason, why you don't use "Röntgen". I don't think it sounds so bad.

### DIFF
--- a/resources/i18n/de/cura.po
+++ b/resources/i18n/de/cura.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: Cura 2.1\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2016-01-18 11:54+0100\n"
-"PO-Revision-Date: 2016-02-01 17:26+0100\n"
+"PO-Revision-Date: 2016-02-01 17:33+0100\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -613,10 +613,9 @@ msgid "0.0 m"
 msgstr "0,0 Meter"
 
 #: /home/tamara/2.1/Cura/resources/qml/JobSpecs.qml:218
-#, fuzzy
 msgctxt "@label"
 msgid "%1 m"
-msgstr "%1 Meter"
+msgstr "%1 m"
 
 #: /home/tamara/2.1/Cura/resources/qml/SidebarHeader.qml:29
 #, fuzzy
--- a/resources/i18n/de/cura.po
+++ b/resources/i18n/de/cura.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: Cura 2.1\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2016-01-18 11:54+0100\n"
-"PO-Revision-Date: 2016-02-01 18:47+0100\n"
+"PO-Revision-Date: 2016-02-01 18:48+0100\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
@TMHogenhout is on vacation right now, so you probably won't hear from her in the next two weeks (and in the past week). I've taken responsibility for the translations for now, which basically means that I have to check and test the Dutch, Castilian and Italian translations that are still to come.

I agree with a lot of these, especially Konfiguration and Aktualisierung. But the translation bureau is made by professional translators and the stance of our supervisors in the past has been that the professional translation has more authority than the translation from a developer, even if it is the native language of that developer. This kind of trust has led to hilarious situations in the past, such as when Ultimaker's website showed in huge letters on the front page: "MEET YOUR ULTIMAKER". But nevertheless, I'm going to have to pass this to higher up.

That comma will have to be looked at though. I don't think there would be any string parsing on that, is there? I can't imagine that we hadn't thought about that the radix could change for other languages. If it is dependent on radix, we should change that.

Ghostkeeper:
The comma wasn't parsed of course but probably gets replaced by a period as soon as the material amount becomes more than 0 metres. We should take a good look at the way we print international numbers in general, also in setting fields and such. For this particular label, something like this would be a good solution I think:

Change

    text: base.printMaterialAmount <= 0 ? catalog.i18nc("@label", "0.0 m") : catalog.i18nc("@label", "%1 m").arg(base.printMaterialAmount)

into

    text: catalog.i18nc("@label", "%1 m").arg(base.printMaterialAmount <= 0 ? 0.0 : base.printMaterialAmount)

And maybe format it a bit such that it always shows 1 decimal. This is of course not Thopiekar's to fix but maybe someone would like to take a look into this. I'll see what I can do at my boss to determine whether this pull request should be merged.

thopiekar:
@Ghostkeeper Thank you for still taking care about this PR. I hope it never sound to you like ~~"You are not making a great job - my translations are better"~~. I'm really sorry, if it ever sounded like this!

I just wanted to share how your current translations (which are not bad at all!) sound to me and, as I have the tools to modify the po-files, I just decided to make this PR to share my experiences with you.

Well, in my point of view, I would implement something like:
```
>>> i=1.2
>>> [int(i), round(i-int(i),2)]
[1, 0.2]
```
So, the line you mentioned should look something like:
```
visual_printMaterialAmount = max(0.0, base.printMaterialAmount)
catalog.i18nc("@label", "{0}.{1} m").format(int(visual_printMaterialAmount), round(visual_printMaterialAmount-int(visual_printMaterialAmount),2))
```
Don't know whether it was from a qml or py file, but I think about something like that. Seems like base.printMaterialAmount can also be > 0.0 .

With something like "{0}.{1} m" it should be very easy to translate it into different languages.

fieldOfView:
I guess it is a bit of a "dutchism" to keep some terms in english even if we have dutch words for them. For example we would never say "spruitstuk" for "nozzle" in a 3d printing context. So we would prefer "X-Ray" over "Röntgen", but I can see how the latter would be more apt in german. 

It is hard to imagine that noone has tackled the radix in floats translation before in PyQt/QML. But a cursory search did not show me any better options...

Ghostkeeper:
> @Ghostkeeper Thank you for still taking care about this PR. I hope it never sound to you like "You are not making a great job - my translations are better". I'm really sorry, if it ever sounded like this!

> I just wanted to share how your current translations (which are not bad at all!) sound to me and, as I have the tools to modify the po-files, I just decided to make this PR to share my experiences with you.

Don't worry, man. I didn't take it this way, and if I did I would eagerly accept critique. I'm thankful for your contributions. Know that you're becoming somewhat of a legend around the office due to all your contributions.

Javascript does have internationalisation options for numbers, such as [`toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString). We could take a look at those. In Python there's the [locale](https://docs.python.org/3/library/locale.html) package.

Ghostkeeper:
I have the green light from my supervisor to incorporate these improvements to the German translations. I'll merge this pull request now, and if we ever get the chance to improve the locale-specific float-to-string conversion we'd have to check back at the internationalisation files anyway.

PR #629, opened 2016-02-04
i18n: de: Another set of translation improvements

thopiekar:
First I found this:
cura-import-image2
and fixed it (untested!) via https://github.com/Ultimaker/Cura/commit/14f09989048ad8025a5a2ff89d1312026a2ba7fd
Please note that I don't know which other strings are affected of line: https://github.com/Ultimaker/Cura/commit/14f09989048ad8025a5a2ff89d1312026a2ba7fd#diff-3fddbff0922570e8c821286a70dc0bdaR1219

I can also check other strings if you can tell me where "Ready to " also appears.

While doing this I also found:
https://github.com/Ultimaker/Cura/commit/6d475ce08c50f6cf98e274c5125e19ce03f420aa


thopiekar:
Please note the comment of commit https://github.com/thopiekar/Cura/commit/dd95ea71cf9b3ebbc4048517e2cfb015393abe6c

It might be possible that this also appears in other languages.
{0} is not the location of the generated file, but it is the name of the removable drive..

bildschirmfoto4

Ghostkeeper:
The commit of 14f09989048ad8025a5a2ff89d1312026a2ba7fd might happen in other languages too. It'd be better to have the translation key be something like `Ready to {0}` and then fill in `{0}` with `Save to Removable Drive`. I expect the changes you made to dd95ea71cf9b3ebbc4048517e2cfb015393abe6c to be done properly in other languages. If it isn't, that's the fault of the translators.

thopiekar:
I wouldn't even call it {0}. It might be less confusing when you call it directly {removable_drive_name}.
Same for ```Ready to {0}``` -> ```Ready to {task_name}```.

thopiekar:
https://github.com/Ultimaker/Uranium/pull/108 is still open.

PR #642, opened 2016-02-24
add prusa I3 XL  machine profile and STL

guigashm:
add prusa I3 XL  machine profile and STL


nallath:
Thanks for adding it :)

guigashm:
;DD glad to help.

thopiekar:
@nallath: there is a "icon" property pointing at a ultimaker icon. Is this correct?!

nallath:
If i recall correctly, the icon bit in the machine is depricated. We should really remove that, but we never got to it.

PR #658, opened 2016-03-09
[2.1] i18n: de: Correcting tooltip

thopiekar:
Just found another translation error.
This commit corrects "Bereit zum Über USB drucken" into "Bereit zum Drucken über USB".
It is almost related to another set of corrections I sent before, but didn't found this one as I didn't had my printer connected.

cura


nallath:
Thanks :)

PR #664, opened 2016-03-14
USBPrinting: Cleanups to USBPrinterManager

thopiekar:
While looking for the code in Cura, which tell about the firmware names I found this list of if checks. So before adding my own printer here I just began cleaning up this code by using dictionaries.

* Please verify this PR before merging. It is untested...


Ghostkeeper:
Good idea.

The code looks good, except that it uses some strange indenting in the if-else statements. In this case it wouldn't go wrong but if we modify that code and don't see the indenting it could lead to strange results. We always use 4 spaces. If that messes up the vertical alignment that's the lesser of two evils.

I'll test this momentarily and then merge once the indenting is fixed.

plugins/USBPrinting/USBPrinterManager.py
```diff
@@ -135,36 +135,44 @@ def getInstance(cls, engine = None, script_engine = None):
...
+                                      "ultimaker2_extended_plus" : "MarlinUltimaker2extended-plus.hex",
+                                      }
+        machine_with_heated_bed    = {"ultimaker_original"       : "MarlinUltimaker-HBK-{baudrate}.hex",
+                                      }
 
         ##TODO: Add check for multiple extruders
 
+        hex_file = None
+        if not machine_instance.getMachineSettingValue("machine_heated_bed"):
+           if  machine_type in machine_without_heated_bed.keys():
```
thopiekar: @Ghostkeeper Yes, you are right! For a reason there are only 3 spaces between 158 and 159!
I should really go and setup Eclipse+PyDev for coding on Cura except using a simple editor (Kate).

thopiekar:
Should be corrected now :dancer:

Ghostkeeper:
It can't seem to find any firmware for UM2-series printers. Looking more closely at the code, it won't work with any machine that has heated bed other than UMO.

Because it checks for `machine_heated_bed` to determine from which dictionary to read, and for machines with a heated bed that setting is always `True`. Only for UMO will it set `machine_heated_bed` to `False` or `True` depending on what the user does in the add machine wizard. This can be fixed by moving all UM2 printers except UM2go and the UMO+ printer to the `machine_with_heated_bed` dictionary. Only BQ Witbox, UMO and UM2go are then in the `machine_without_heated_bed` dictionary.

Further, it seems that at line 164 it checks whether the key is present in `machine_with_heated_bed` but then at line 165 looks up the key in `machine_without_heated_bed`, which could result in a key error.

thopiekar:
Hope no additional commits are needed now... (fingers crossed)

thopiekar:
Just wait for a moment. I'm rethinking my if-clauses again...

thopiekar:
I guess it makes more sense now :smile:

Ghostkeeper:
Works!

thopiekar:
Yeah!

PR #666, opened 2016-03-15
Fixing small issues in the code (#2)

thopiekar:


setup.py
```diff
@@ -3,11 +3,10 @@
 
 from distutils.core import setup
 import py2exe
-import UM
-import UM.Qt
-import cura
+import UM    #@UnusedImport
```
Ghostkeeper: This import is used down at line 27: `UM.__file__`

Ghostkeeper:
I don't necessarily agree with the changes to WindowsRemovableDrivePlugin.py. Generally it's advisable to use `from ctypes import windll`-type importing rather than `import ctypes`. That's a bit of a subjective thing though so I'd merge this if it weren't for the one comment that I added to the change of setup.py. If you'll remove the comment you placed there then it's good to go as far as I'm concerned.

thopiekar:
Ok, the idea behind using ctypes.windll instead of windll was to save reimporting it (would be available as ctypes.windll and windll).
I'll remove the comment soon. For a reason PyDev said it is unused, but I didn't check whether it is true.

thopiekar:
Done :wink:

PR #668, opened 2016-03-17
[master] Generate cura.desktop

thopiekar:
Should make the copy of cura.desktop in cura-build unnecessary.
Additionally it makes it possible to install cura into a non-default path, while cura.desktop is still using the correct path.


awhiemstra:
Actually, this does not obsolete the copy in cura-build. cura-build installs a script into /usr/bin/ that will run the Cura in /opt with all the right paths set, like LD_LIBRARY_PATH. This will still be needed.

thopiekar:
Wouldn't it be better then to rename /usr/bin/cura into /usr/bin/cura.real and call the script you mention /usr/bin/cura? I know that different projects, like MythTV, do it like that.

thopiekar:
Are there any questions left? Something that I missed?

awhiemstra:
Hmm, no, it seems to be OK for the rest, it at least fixes the "different install folder than /usr" thing.

thopiekar:
Ok, so I need to make another PR in ```cura-build``` to make the fix complete?

thopiekar:
:eyes:

awhiemstra:
You might want to look at the pull request I made for cura-build. Basically, it changes everything to in the end produce AppImage files instead of debian packages. That one kind of obsoletes the need for a change in cura-build.

thopiekar:
Yes, in the past it was meant to be used with some changes/pull requests I made for cura-build.
It was left here for a while so I decided to update the PR here and merge it, because it shouldn't break anything.

Just to be sure: How are the internal builds working today? Is the menu entry still working fine?

PR #673, opened 2016-03-18
WIP: [master] RemovableDriveOutputDevice: Adding basic UDisks2 support

thopiekar:
The code doesn't work at the moment. For a reason Cura doesn't show the devices, which were found by the new module.

The "legacy" aka old module creates dictionaries of:
{mount_point_1: label_1, etc.}
Whereas my modules creates:
{udisk2_device_name_1: label_1, etc.}

Converting everything to mointpoints and then rescan all devices which fit to the mount points (eg. for ejecting) is no option in my point of view.
It would be much better so have something like ```drive.getMountPoint()``` where the mount point will get requested from UDisks2.


awhiemstra:
I have had plans for UDisks support for this for a long time, nice to see someone working on it.

While there is nothing wrong with the code, I do think it is not radical enough. :) If we are using UDisks anyway, there is no need for a thread that runs every 5 seconds, because UDisks should be sending DBus signals when a new device is plugged in or removed. We can react to those signals rather than scanning.

Though this means not inheriting the basic removable device class in the plugin. On the other hand, that class is purely a convenience and should not block us from doing better implementations where needed.

> Converting everything to mointpoints and then rescan all devices which fit to the 
> mount points (eg. for ejecting) is no option in my point of view.

Mount points are an implementation detail and should not be exposed. Since everything is (should) be under control of the plugin, adding a property or method to expose the mount point should not be much of a problem.

thopiekar:
Yes, I saw the comment in the code ^^

Ok, but I'm not sure whether I can use ```dbus``` to bind the signal or need to take QDBus (which I remember from the past, when I was playing with PyQt4).

But as I'm not dealing here with mount points at all, how does Cura know here to store the files? I browsed the code and couldn't find any hint. That's why I believed it using the ID to create the paths, eg. for```os.path.join(id, gcode_to_be_stored)``` or ```os.path.isdir(id)```.

awhiemstra:
> Ok, but I'm not sure whether I can use dbus to bind the signal or need to take QDBus 
> (which I remember from the past, when I was playing with PyQt4).

If it cannot do that it is a rather useless dbus module. :p But at least according to https://dbus.freedesktop.org/doc/dbus-python/api/ it is possible.

> But as I'm not dealing here with mount points at all, how does Cura know here to store 
> the files? I browsed the code and couldn't find any hint. That's why I believed it using the 
> ID to create the paths, eg. foros.path.join(id, gcode_to_be_stored) or os.path.isdir(id).

Yes, that is what it currently does. You will need to extend the API a little to make it get the path from the plugin instead of assuming "id == path".

thopiekar:
@awhiemstra Well, this is maybe possible since PyQt5, but I remember when I was coding with PyQt4 that I had to use ```QThread``` except of ```thread``` or ```threading```. My application was freezing with them. So I just asked ^^

Ok, thanks!

Issue #465, opened 2015-10-07
Printing with USB problem with Ultimaker 2 Go


mesut-pehlivan:
After pressing "print", the nozzle immediately starts printing without heating up, and the bed also doesn't move to its position. It seems like the start gcode never runs. Printing with SD-card works fine.
I've experienced this with any 15.09.*\* version so far.

peteruithoven:
(This is an issue for Ultimaker2's in general.)
Those machine's gcode flavor are set to Ultigcode, which only properly works from an SD card. 
To support USB printing the gcode flavor of the machine has to be changed to reprap and there needs to be start- and end gcode.
Would it be possible to change the flavor "temporarily" from a plugin? Would it be possible to change the start- and end gcode depending on the gcode flavor? Even easier would be that the settings could be automatically changed based on a usb printing flag (this would hide more implementation details and give more freedom). 

Enabling USB printing is required to support wireless printing using products like the Doodle3D WiFi-Box, Octoprint and Astroprint.

Ghostkeeper:
Post processing plug-ins are able to change such things, as a temporary solution.

peteruithoven:
But this would require a seperate plugin? And even a setting that users need to change manually?
It also doesn't solve the lack of default start and end gcode for the Ultimaker2, we'd have to add the Ultimaker2 reprap flavor start and end code in our plugin... Shouldn't this be included in the default machine settings of Cura? 
Being able to usb print Ultigcode might be an even better longterm solution, but it will take a while for people to update their firmware.

awhiemstra:
The proper solution for this is that the printer should somehow do its start-up sequence when trying to print over USB. However, I do not know if and when this will happen.

companje:
I think we need to find out 'if and when' this will happen. Because without this people won't be able to use the USBPrinting-plugin or the Doodle3D-plugin on Ultimaker printers without manually changing their settings files.

Ghostkeeper:
I don't see that proper solution happening anytime soon. The firmware developers have their hands full at the moment. Apparently it's fairly involved.

peteruithoven:
So would it be possible to have something like a `connection type` setting in Cura? This could filled with things like 'usb' or 'card'. 
This then should change the flavor from ultigcode to reprap and use the necessary start and end gcode. 
This might be implemented by adding connection type specfic machine settings? 

The more hacky solution would be to always enforce the reprap flavor and ignoring the start and end gcode when the ultigcode flavor is used.

[...]

PR #689, opened 2016-04-06
Add start and end GCode to enable printing through WiFi

mesut-pehlivan:
The UltiGCode flavor does not use the "machine_start_gcode" and "machine_end_gcode". Adding this start and end gcode will enable us to print through USB and/or WiFi if we switch the printer flavor to RepRap before printing (through the use of plugins).

### DIFF
--- a/resources/machines/ultimaker2.json
+++ b/resources/machines/ultimaker2.json
@@ -34,8 +34,8 @@
         }
     ],
     "machine_settings": {
-        "machine_start_gcode" : { "default": "" },
-        "machine_end_gcode" : { "default": "" },
+        "machine_start_gcode" : { "default": "G21\nG90\nM107\nG28\nG1 Z15 F9000\nG92 E0\nG1 F9000" },
+        "machine_end_gcode" : { "default": "M107\nG91\nG1 E-1 F300\nG1 Z+5.5 E-5 X-20 Y-20 F9000\nG28\nM84\nG90" },
         "machine_width": { "default": 223 },
         "machine_depth": { "default": 223 },
         "machine_height": { "default": 205 },


nallath:
Then this is the wrong way of doing it. If you change the flavour to reprap, you should at that point change the start / end g-code.

mesut-pehlivan:
The tricky thing is where do we then store the start / end gcode for printers that use UltiGCode? 
To us it doesn't make sense to add Ultimaker2/UltiGCode specific configuration to our wireless printing plugin. We'd like to add the start / end g-code configuration to Cura so the bigger community can help maintain it and all plugin creators (Octoprint, Astroprint, Printr, 3DPrinterOS) can access it. 

In the long run it might make sense to add connection type (wifi, usb, ...) or flavor specific start/end gcode to printer configurations, this would allow Cura to automatically use the appropriate start/end gcode. Or have the Ultimaker2 "simply" support UltiGCode over usb of course. 

Also see the discussion in: https://github.com/Ultimaker/Cura/issues/465

awhiemstra:
In my opinion, having the start/end g-code in the machine definition files is the correct place, since it is required for USB printing as well. What I am more wondering about is if it is not better to leave the flavour at UltiGCode and then send the start/end gcode manually, instead of having the engine append it.

mesut-pehlivan:
The problem with generating g-code in UltiGCode is that the extrusion is different than when you would generate it in RepRap. If I recall correctly UltiGCode uses mm³ as extrusion, reprap uses mm of filament.

awhiemstra:
Yes, UltiGCode uses mm3. I would expect that that does not need to change between USB printing and SD card printing. If it does need to change then changing the favour makes sense.

daid:
Actually, when printing trough USB on an UM2, then it acts like an UMO. So printing in mm makes sense then.

Only firmware bug is that UltiGCode is not seen trough USB printing for the UM2, and thus is always falls back to UMO compatibility.

nallath:
For some reason I thought that print with USB did the same as printer over SD. 

I do agree with the proposed change to add a prefered connection type of some sorts to the machine (or even supported types). We already did something like this with prefered file type.

daid:
https://github.com/Ultimaker/Ultimaker2Marlin/blob/master/Marlin/UltiLCD2_menu_print.cpp#L403
Relevant hack in the UM2 code that is the source of this mess.

peteruithoven:
So I think we agree on that this is or requires a hack anyway, but what do you guys think about the pull request?

fieldOfView:
As an alternative approach, you could switch gcode-flavor from UltiGCode to Reprap. Here's a definition that does that:
https://gist.github.com/fieldOfView/ae85f755fc70d1c65e6968a870dfef0c
Tested with OctoPrint, which has the same usb-based problem.

fieldOfView:
I am currently testing adding the start/end gcode like this:
```
    "overrides": {
        "machine_start_gcode" : {
            "default_value": "",
            "value": "\"\"  if machine_gcode_flavor == \"UltiGCode\" else \"G21 ;metric values\\nG90 ;absolute positioning\\nM82 ;set extruder to absolute mode\\nM107 ;start with the fan off\\nG1 X10 Y0 F4000;move X/Y to min endstops\\nG28 Z0 ;move Z to bottom endstops\\nG1 Z15.0 F9000 ;move the platform to 15mm\\nG92 E0 ;zero the extruded length\\nG1 F200 E10 ;extrude 10 mm of feed stock\\nG92 E0 ;zero the extruded length again\\nG1 F9000\\n;Put printing message on LCD screen\\nM117 Printing...\""
        },
        "machine_end_gcode" : {
            "default_value": "",
            "value": "\"\"  if machine_gcode_flavor == \"UltiGCode\" else \"M104 S0 ;extruder heater off\\nM140 S0 ;heated bed heater off (if you have it)\\nG91 ;relative positioning\\nG1 E-1 F300  ;retract the filament a bit before lifting the nozzle, to release some of the pressure\\nG1 Z+0.5 E-5 X-20 Y-20 F9000 ;move Z up a bit and retract filament even more\\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\\nM84 ;steppers off\\nG90 ;absolute positioning\""
        },
```
That way, the start/end gcode evaluate to an empty string when the gcode flavor is set to ultigcode (as is the case currently), but if the user switches to reprap flavor the start/end gcode is used.

fieldOfView:
What would be good start/end gcode sequences for UM2 though?

peteruithoven:
@fieldOfView How are those override values connected to the gcode flavor? 

At Doodle3D we tested with the gcodes @mesut-pehlivan proposed. Those are based on the gcode we use on the Doodle3D WiFi-Box.

fieldOfView:
In 2.2 (not sure about 2.1), the value can be a function. In this case the function uses the setting of the machine_gcode_flavor to switch between two possible values (empty string or my gcode). 

I am working on a plugin that lets the user easily switch between gcode flavors: https://github.com/fieldOfView/MachineSettingsActionPlugin

peteruithoven:
Nice! 

And that sounds like a very handy plugin.

fieldOfView:
This is what it currently looks like:
machinesettings

Issue #662, opened 2016-03-11
Download firmware as needed (into TMP)


thopiekar:
Related to issue #627, I think about a feature to be able to add a download URL for firmware to the JSON files.

The URL could be even point to a "virtual file", eg.:

```
http://www.printer.vendor.org/download/awesome-printer_latest.hex
```

and Cura will follow according to the http response to

```
http://www.printer.vendor.org/download/awesome-printer_999.hex
```

.. so vendors will be able to provide updates, without modifying the Cura package or waiting for a new release.

Here is a suggestion how the firmware update process could work.

```
1. Check for a local copy of a firmware
2.1 Check for the URL tag in the printer JSON (If there is no tag skip to 4.)
2.2 Check for a available internet connection (If there is no connection skip to 4.)
2.3 Ask the server for the md5sum of the latest firmware, which is online.
2.4 Create a md5sum of the file for comparisons 
3. Download the firmware as needed as tmpfile (which will be removed automaticly when the file gets .close()'d again - so no mess with old firmware)
4. Flash the latest firmware.
```

Finally I would also love to see a function adding the .hex manually. This could be done using two seperate tabs, one for "automatic" and one for "manual" firmware update.

ianpaschal:
Closing for ~2 years of inactivity.

Ghostkeeper:
The "manual" upload is implemented now. Maybe that can be considered sufficient.

PR #699, opened 2016-04-14
Output device & printer connection rework

nallath:
USB printing now uses a more generalised way to talk to printers. This should greatly simplify the creation of printer connections (doodle 3D, octoprint, printr,  etc)

CURA-1339


thopiekar:
Looks good, but what about firmware upgrades? It looks like there is no list of machine_id to firmware_filename anymore. Will this PR remove the firmware upgrade tool at all?

nallath:
We're going to rework how firmware upgrades are done. It shouldn't remove the firmware upgrading though. Might be that I've been a bit over zealous with removing old / unused code.

thopiekar:
Ok, it was part of ```plugins/USBPrinting/USBPrinterManager.py``` where I contributed approx a month ago. I just added my printer to this list and I wondered where it will be located after this PR.
Hope you'll take a look on this: https://github.com/Ultimaker/Cura/issues/662 :wink:

Ghostkeeper:
Firmware upgrade should no longer be put into USBPrinterManager since it is not specific to USB connections.

nallath:
I slightly disagree with Ghostkeeper here. Firmware upgrading can be in USBConnection. The point is that firmware upgrading is not specific to -just- USB connections (it's just one of multiple ways to do it)

Ghostkeeper:
I suppose you're right in that it probably needs a different implementation for each type of connection.

thopiekar:
Yes, USBConnection should include firmware flashing and what we know as USB
printing should go to something like "serial printing". To be honest I
don't see a reason why we couldn't print over Bluetooth serial connections,
for example.

nallath:
It is possible, but i think it would be better to just make a BluetoothPrinterOutputDevice (instead of over complicating it even more).
I very much doubt that anyone will make the bluetooth interface in the first place.

thopiekar:
That's true, but the data exchange does not 'really' happen over USB. The
data is only bridged over USB while the real data is truly exchanged
between the USB-serial converter on the board. So you are not forced to use
USB. Think about the WiFi-serial chips (ESP**** don't remember the correct
name) or (old-school) connections via RS232 cable.
They are all possible and the Firmware on the printer does not care (that
much) how data was transferred.

However, that's only my point of view.

awhiemstra:
If a bluetooth serial connection acts like a serial connection (by creating a proper node in /dev) then it would be picked up regardless of whatever we call the plugin. In my opinion, naming the plugin should also communicate the intended scope of the plugin. Since we do not want to litter the USB plugin with a bunch of code to set up bluetooth connections, calling it USB printing plugin is fine.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+        self._progress = 0
+        self._head_x = 0
+        self._head_y = 0
+        self._head_z = 0
+        self._connection_state = ConnectionState.CLOSED
+
+    def requestWrite(self, node, file_name = None, filter_by_machine = False):
+        raise NotImplementedError("requestWrite needs to be implemented")
+
+    # Signals:
```
awhiemstra: Please also document signals. Though these are fairly obvious, adding information about parameters passed is quite important since those can not be inferred from a simple signal declaration.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+#   The assumption is made the printer is a FDM printer.
+#
+#   Note that a number of settings are marked as "final". This is because decorators
+#   are not inherited by children. To fix this we use the private counter part of those
+#   functions to actually have the implementation.
+#
+#   For all other uses it should be used in the same way as a "regular" OutputDevice.
+class PrinterOutputDevice(OutputDevice, QObject):
+    def __init__(self, device_id, parent = None):
+        QObject.__init__(self, parent)
```
awhiemstra: Just using super().__init__(parent, device_id) should also work correctly.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+    hotendTemperaturesChanged = pyqtSignal()
+    targetHotendTemperaturesChanged = pyqtSignal()
+
+    headPositionChanged = pyqtSignal()
+
+    connectionStateChanged = pyqtSignal(str)
+
+    ##  Get the bed temperature of the bed (if any)
+    #   This function is "final" (do not re-implement)
+    #   /sa _getBedTemperature implementation function
```
awhiemstra: You should use \ for doxygen keywords, not /.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+    @pyqtProperty(float, notify = bedTemperatureChanged)
```
awhiemstra: I actually realised the other day that this is not necessary if we separate the property declaration from the getter function, since then you can safely override the getter function without breaking the property. For example:

`
def getBedTemperature(self):
    raise NotImplementedError()

bedTemperature = pyqtProperty(float, fget = getBedTemperature, notify = bedTemperatureChanged)
`

Unfortunately, since slots are only available as decorators, we cannot use the same trick.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+    ##  Protected setter for the current bed temperature.
+    #   This simply sets the bed temperature, but ensures that a signal is emitted.
+    #   /param temperature temperature of the bed.
+    def _setBedTemperature(self, temperature):
+        self._bed_temperature = temperature
+        self.bedTemperatureChanged.emit()
+
+    ##  Get the bed temperature if connected printer (if any)
+    @pyqtProperty(int, notify = bedTemperatureChanged)
+    def bedTemperature(self):
```
awhiemstra: This is a duplicate of the same property already declared above.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+
+    ##  Implementation function of setTargetHotendTemperature.
+    #   /param index Index of the hotend to set the temperature of
+    #   /param temperature Temperature to set the hotend to (in deg C)
+    #   /sa setTargetHotendTemperature
+    def _setTargetHotendTemperature(self, index, temperature):
+        Logger.log("w", "_setTargetHotendTemperature is not implemented by this output device")
+
+    @pyqtProperty("QVariantList", notify = targetHotendTemperaturesChanged)
+    def targetHotendTemperatures(self):
```
awhiemstra: Since you have a getter that returns a list, why not add a setter that accepts a list? (In addition to the index based setter).

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+    #   This simply sets the hotend temperature, but ensures that a signal is emitted.
+    #   /param index Index of the hotend
+    #   /param temperature temperature of the hotend (in deg C)
+    def _setHotendTemperature(self, index, temperature):
+        self._hotend_temperatures[index] = temperature
+        self.hotendTemperaturesChanged.emit()
+
+    ##  Attempt to establish connection
+    def connect(self):
+        pass
```
awhiemstra: Shouldn't this raise NotImplementedError? Seems both this and the other should always be reimplemented.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+
+    ##  Set the X position of the head.
+    #   This function is "final" (do not re-implement)
+    #   /param x x position head needs to move to.
+    #   /param speed Speed by which it needs to move (in mm/minute)
+    #   /sa _setHeadx implementation function
+    @pyqtSlot("long")
+    @pyqtSlot("long", "long")
+    def setHeadX(self, x, speed = 3000):
+        self._setHeadX(x, speed)
```
awhiemstra: Why not call `_setHeadPosition(x, self._head_y, self._head_z, speed)`? That would save reimplementing four separate functions that would effectively do the same.

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+
+    ##  Set the progress of any currently active process
+    #   /param progress Progress of the process.
+    def setProgress(self, progress):
+        self._progress = progress
+        self.progressChanged.emit()
+
+##  The current processing state of the backend.
+class ConnectionState(IntEnum):
+    CLOSED = 0
```
awhiemstra: Can we just use upper camel case for enums? All caps is ugly (and I do not like shouting in code).

cura/PrinterOutputDevice.py
```diff
@@ -0,0 +1,303 @@
...
+    CONNECTING = 1
+    CONNECTED = 2
```
awhiemstra: It seems to me like this lacks several important states, like "printing" for example.

plugins/USBPrinting/ControlWindow.qml
```diff
@@ -25,7 +25,8 @@ UM.Dialog
             Label
             {
                 //: USB Printing dialog label, %1 is head temperature
-                text: catalog.i18nc("@label","Extruder Temperature %1").arg(manager.extruderTemperature)
+                Component.onCompleted: console.log(manager.hotendTemperatures)
```
awhiemstra: Just like print statements, please remove debug things. :)

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -119,51 +89,52 @@ def __init__(self, serial_port, parent = None):
...
+
     firmwareUpdateComplete = pyqtSignal()
 
     endstopStateChanged = pyqtSignal(str ,bool, arguments = ["key","state"])
 
-    @pyqtProperty(float, notify = progressChanged)
-    def progress(self):
-        return self._progress
+    def _setTargetBedTemperature(self, temperature):
+        Logger.log("d", "Setting bed temperature to %s", temperature)
```
awhiemstra: It might be useful to also log for which device this is happening, in case you have multiple devices.

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -456,9 +363,9 @@ def showControlInterface(self):
     ##  Send a command to printer. 
     #   \param cmd string with g-code
     def sendCommand(self, cmd):
-        if self.isPrinting():
+        if self._progress:
```
awhiemstra: This should check for a certain state, not some arbitrary value.

plugins/USBPrinting/__init__.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
-from . import USBPrinterManager
+from . import USBPrinterOutputDeviceManager
```
awhiemstra: How does this work? Is this a missing file? Because it is not in this review.

thopiekar:
@awhiemstra I agree with that. However, if the plugin will have more features in the future, we should rethink that naming :wink:

cura/PrinterOutputDevice.py
nallath: I copied this from how USBPrinting did it. I usually like to make specific as to what parents use what. (Now it's clear that parent is used by QObject and device_id by output device)

cura/PrinterOutputDevice.py
nallath: oh. Derp. Will fix that.

cura/PrinterOutputDevice.py
nallath: I was a bit in dubio here. Not sure why I chose to not add a notImplemented here. But don't feel very strong about it either way, so I'll change it.

cura/PrinterOutputDevice.py
nallath: I thought we decided on using all caps for enums? Not very sure though.

cura/PrinterOutputDevice.py
nallath: This is intentional. Printing has nothing to do with connection state.

plugins/USBPrinting/__init__.py
nallath: Thats a missing file. Renamed it but messed up adding it I guess.

cura/PrinterOutputDevice.py
nallath: Removed it.

cura/PrinterOutputDevice.py
nallath: So. Change it or leave it as is? I'm slightly in favour of leaving it so we have all this "documentation" in a single class (instead of having two systems to handle this). Although there is something to be said for it being easier to read this way.

cura/PrinterOutputDevice.py
nallath: Done.

cura/PrinterOutputDevice.py
nallath: I see self._head_y as a stored value (posibiliy to be updated by the listen thread or some other means). Setting the headPosition should use new values (instead of stored ones).

plugins/USBPrinting/USBPrinterOutputDevice.py
nallath: I could change this, but it would only be reimplementing data that's implied by the progress. (-1 && 0 == not printing, anything else is doing something).

cura/PrinterOutputDevice.py
nallath: Changed it. The issue was caused when not using named arguments (which it now does)

cura/PrinterOutputDevice.py
awhiemstra: I am fine with leaving it as is for now. I will investigate if it is also possible to do this with pyqtSlot, in which case I would prefer that approach later on as it makes inheritance less complicated.

cura/PrinterOutputDevice.py
awhiemstra: We never formally agreed on it iirc and I have always been against it. As I said, all other enums/constant values in Uranium and Cura use upper camel case as far as I know, so I would at least prefer to keep that consistent.

cura/PrinterOutputDevice.py
awhiemstra: But a "busy" state would make sense. As would an "error" state in case of connection errors.

plugins/USBPrinting/USBPrinterOutputDevice.py
awhiemstra: The point is to make the API explicit. If you compare against a state, you know it is in that state. If you just compare against "magic values" you need to always figure out what those magic values are.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -20,21 +21,22 @@
...
 from PyQt5.QtCore import QUrl, QObject, pyqtSlot, pyqtProperty, pyqtSignal, Qt
 from UM.i18n import i18nCatalog
 i18n_catalog = i18nCatalog("cura")
 
-class USBPrinterManager(QObject, SignalEmitter, OutputDevicePlugin, Extension):
+
+##  Manager class that ensures that a usbPrinteroutput device is created for every connected USB printer.
+class USBPrinterOutputDeviceManager(QObject, SignalEmitter, OutputDevicePlugin, Extension):
     def __init__(self, parent = None):
         QObject.__init__(self, parent)
```
awhiemstra: Same as the other, please use super().__init__().

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -170,48 +172,53 @@ def _getDefaultFirmwareName(self):
...
-        if self._printer_connections[serial_port].isConnected():
-            self.getOutputDeviceManager().addOutputDevice(self._printer_connections[serial_port])
-        else:
-            self.getOutputDeviceManager().removeOutputDevice(serial_port)
-        self.printerConnectionStateChanged.emit()
+    def addOutputDevice(self, serial_port):
+        device = USBPrinterOutputDevice.USBPrinterOutputDevice(serial_port)
+        device.connect()
+        device.connectionStateChanged.connect(self._onConnectionStateChanged)
+        device.progressChanged.connect(self.progressChanged)
```
awhiemstra: Ideally, you should connect to signals before calling a method that may cause those signals to be emitted, unless you intentionally want to ignore those signals. In this case, if device.connect() is fast enough (or uses blocking calls) you end up with not being informed about connection state changes.

cura/PrinterOutputDevice.py
nallath: Fair enough; consistency > standards.

plugins/USBPrinting/__init__.py
```diff
@@ -19,5 +19,5 @@ def getMetaData():
     }
 
 def register(app):
-    qmlRegisterSingletonType(USBPrinterManager.USBPrinterManager, "UM", 1, 0, "USBPrinterManager", USBPrinterManager.USBPrinterManager.getInstance)
-    return {"extension":USBPrinterManager.USBPrinterManager.getInstance(),"output_device": USBPrinterManager.USBPrinterManager.getInstance() }
+    qmlRegisterSingletonType(USBPrinterOutputDeviceManager.USBPrinterOutputDeviceManager, "UM", 1, 0, "USBPrinterManager", USBPrinterOutputDeviceManager.USBPrinterOutputDeviceManager.getInstance)
```
awhiemstra: Considering this code is not in Uranium, this should not be registered to the "UM" module. Instead, I'd use "Cura" or event "Cura.USBPrinting" so you do not end up with collisions.

plugins/USBPrinting/USBPrinterOutputDevice.py
nallath: Except that in this case this is not done by the API, but by the implementation of the API (not being able to do certain things when printing is specific to USB printing).

plugins/USBPrinting/__init__.py
nallath: Fair enough. This got through before we had formal reviews. I'm guessing this "mistake" is made more often for other plugins.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
nallath: This was old copied code, but I will change it anyway.

plugins/USBPrinting/__init__.py
```diff
@@ -19,5 +19,5 @@ def getMetaData():
     }
 
 def register(app):
-    qmlRegisterSingletonType(USBPrinterManager.USBPrinterManager, "UM", 1, 0, "USBPrinterManager", USBPrinterManager.USBPrinterManager.getInstance)
-    return {"extension":USBPrinterManager.USBPrinterManager.getInstance(),"output_device": USBPrinterManager.USBPrinterManager.getInstance() }
+    qmlRegisterSingletonType(USBPrinterOutputDeviceManager.USBPrinterOutputDeviceManager, "Cura", 1, 0, "USBPrinterManager", USBPrinterOutputDeviceManager.USBPrinterOutputDeviceManager.getInstance)
```
awhiemstra: Come to think of it, the Qt documentation for qmlRegisterSingletonType states:

> NOTE: A QObject singleton type instance returned from a singleton type provider is owned by the QML engine. 
> For this reason, the singleton type provider function should not be implemented as a singleton factory.

This is another one of the reasons for using proxy objects in Uranium. I expect it never was a problem because of how Python and PyQt handle objects, but it is something to be aware of. I do not currently know an easy solution, but please do add a comment here that we are basically violating QML api. :)

nallath: And done ;)

PR #701, opened 2016-04-15
UM2+ profile improvements

THeijmans:
A few improvements to um2+ machine definitions & profiles.

### DIFF
--- a/resources/machines/ultimaker2plus.json
+++ b/resources/machines/ultimaker2plus.json
@@ -11,6 +11,12 @@
     "inherits": "ultimaker2.json",
 
     "overrides": {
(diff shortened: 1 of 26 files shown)


nallath:
Quickly glancing it I see the following issues with it: 

- Setting the print speed to anything below 10 causes outer wall speed to become negative (which is an impossible (error) value, which prevents Cura from slicing). This is especially an issue if the outer wall speed is hidden. 
- For all variants of the UM2+, the same inheritance function is used for "speed_wall_x". It would be better to move it up one level (into UM2+ instead of in the nozzle variants)

But in most cases it works a lot better than the hard-coded values of the profiles that we had before.

Issue #704, opened 2016-04-18
RuntimeError


danilius:
Got this error which prevents Cura from starting:

Traceback (most recent call last):
  File "/opt/cura/bin/cura", line 19, in <module>
    import Arcus
RuntimeError: the sip module implements API v11.0 to v11.1 but the Arcus module requires API v11.2

nallath:
Oh this one again. Did you do a clean install of cura?

danilius:
I installed the deb. I did download the source code and managed to compile it once or twice and then some bugs crept in and I gave up the ghost. So, how do I clean up the mess? That will probably cure this.

nallath:
Someone else did report this isse on the cura-build (https://github.com/Ultimaker/cura-build/issues/40). Maybe @awhiemstra @thopiekar or @Ghostkeeper know a bit more about this (I seem to recall that we fixed it, but can't find anything about this :()

thopiekar:
I guess here is something messed up. Can you give exact info about what you've done? E.g:
1. Was an older version of cura installed as deb before?
2. Have you tried to install one of the beta packages _after_ that?
3. What is the name of your own built package (filename)?

Additionally, please give detailed info about your installation. I can just guess you are using Linux and a APT-based distribution.

Thanks!

danilius:
1) Yes
2) Yes
3) Cura-2.1.0-Linux.deb which I downloaded from here: http://software.ultimaker.com/Cura_open_beta/

I reinstalled the 15.06 version since I need that for work.

thopiekar:
And the old version is working fine?

danilius:
Yup.

awhiemstra:
It is working because that did not use SIP generated bindings. We had this issue before, for some reason either Python or Sip is ignoring all the paths we set up to run Cura and uses a system provided version. @danilius Do you have python3-sip or a similar package installed?

danilius:
Nope, not to my knowledge. But I can install it if you think it will make a difference.

awhiemstra:
Well no, I expect it to be the cause of this issue. What seems to happen is that python or sip decides it should use the system version and that conflicts with the sip used to compile Arcus. Unfortunately I do not know of a workaround at this moment.

danilius:
Oh well.

thopiekar:
Ok, please do the following (I'm sorry that I can't test it here locally).

1. Please reinstall Cura 2.1.0
[...]

PR #708, opened 2016-04-18
Force PYTHONPATH to be in the top of the sys.path list.

kakaroto:
This fixes https://github.com/Ultimaker/Cura/issues/704

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -5,6 +5,15 @@
 
 import sys
 
+# It looks like setuptools creates a .pth file in
+# the default /usr/lib which causes the default site-packages
+# to be inserted into sys.path before PYTHONPATH.
+# This can cause issues such as having libsip loaded from
+# the system instead of the one provided with Cura, which causes
+# incompatibility issues with libArcus
+sys.path.insert(1, os.environ.get('PYTHONPATH', ''))
+
+
 def exceptHook(hook_type, value, traceback):
     import cura.CrashHandler
     cura.CrashHandler.show(hook_type, value, traceback)


thopiekar:
This is completely related to ```cura-build```, because this won't happen, if you install cura ```normally```.
I would prefer to find a solution in ```cura-build```.

However, as far as I read about that problem, ```PYTHONPATH``` is appended to ```sys.path```. So it is only not preferred. What you should do is to check for ```PYTHONPATH``` existence and move it to the front in ```sys.path``` to give it a higher priority.

awhiemstra:
The code would become quite a bit more ugly if you have to do manual merging of paths. I suppose you could add a check to only add it if it is not at the front, but more than that quickly becomes overkill, considering this is a workaround for what is effectively an upstream bug. Since pythonpath should always be at the front I am fine with a small change to ensure that it is in the front.

nallath:
So merge this or not?

thopiekar:
In my point of view: no.
It it simply too 'cura-build' specific.

awhiemstra:
This could probably be fixed in a different way if we start using cx_freeze to create self-contained binaries on Linux. Unfortunately that would require a bit more work in the build system.

Personally, I consider this workaround the same as the one for the Ubuntu driver thing: stupid upstream issues that we really should not have to work around, but still have to to get stuff working. As the cura-build packages are the "official" binaries at the moment, I want to accept this PR provided the check is added to see if both sys.path and PYTHONPATH start with the same thing.

kakaroto:
I agree with @thopiekar in that it's a hack that should be in cura-build more than in cura. I guess I should have added a `# HACK ALERT` at the start of the comment.
As for checking for PYTHONPATH position before adding it, that's not necessary, since it can be there 100 times, it won't change anything, the order is what matters, and having duplicated entries will not affect any behavior, so I don't see a point in moving the PYTHONPATH or adding it only if it's not there, etc...
Also note that sys.path[0] will always be the current working directory from which the app was launched, then sys.path[1] is /usr/bin then it should be PYTHONPATH (that's why I insert it in position 1, not position 0).

Anyways, I think a nicer fix would be to apply this patch in cura-build before creating the bundle, so it stays out of the Cura repo, but is still there in the final deb.

Also note, there was something similar that we had to do a few months ago for windows as well : https://github.com/alephobjects/Cura/commit/f430de51c28074ba47734d4b5a65e3032d0759b1 &
https://code.alephobjects.com/T224 
That's also a hack, but without it, it wouldn't work for some users.

thopiekar:
@kakaroto Thank you! I never took such a detailed look at sys.path. While looking at your explanations I just made your patch more concrete. This was actually what I missed in this PR.
Hope you'll like it.

Btw. I took your commit from your ```master``` branch, so your effort is visible in my PR as well :wink:

kakaroto:
@thopiekar thanks!
One issue with your commit 839bc1fec8b78188494ae9174218a20d10e4dbd2 is that PYTHONPATH won't be in `sys.path[-1]`, so it will never be inserted. Just run ipython with a custom PYTHONPATH then print sys.path to see its content.

sys.path will contain [`pwd`, /usr/bin, $PYTHONPATH, system libs, $PATH] so there are a few directories that come after it, and it won't be the last item.

Trying to see if PYTHONPATH is in the first 2 or 3 items in the list would be more complicated, that's why you should just add it without checking. If it's already there, then you duplicate it, and it has no effect, if it's not, then you changed the order and when python looks for a library, and hits it again in its search list, it will again have no effect since PYTHONPATH will have already been checked.

thopiekar:
Oh, ok, I'll make another concept.

However, this is what I get on Ubuntu Xenial:
```
thopiekar@hp:~$ PYTHONPATH=BLA python
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.path
['', '/home/thopiekar/BLA', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/usr/local/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages/PILcompat', '/usr/lib/python2.7/dist-packages/gtk-2.0', '/usr/lib/python2.7/dist-packages/wx-3.0-gtk2']
```
Here it works.

kakaroto:
@thopiekar yeah, it works for me too (although, I have /usr/bin in position 1 before PYTHONPATH), but it didn't work for a colleague and doesn't work for people in #704.
Also try with python3, it might give a different result. For me, with python 2.7, I do get a few directories before PYTHONPATH, but not with python 3.4

PR #710, opened 2016-04-19
[master] PYTHONPATH fix reworked

thopiekar:
* Added markers for the issues at GitHub. Don't know anything about the JIRA reports.
* Used @kakaroto's work at #708 
* More details in the commits.


thopiekar:
Oh, that needs testing of course. I don't use DEB's made by cura-build at all...

kakaroto:
As explained in https://github.com/Ultimaker/Cura/pull/708#issuecomment-212027838 the check for sys.path[-1] will always be false and will cause the fix to fail.

thopiekar:
Now done the check the other way round. @kakaroto, can you test it?

kakaroto:
@thopiekar It looks good to me that way now.
I've asked my colleague to test it since he's the one with the problem.
Thanks!

thopiekar:
No problem. Hope it will work :)

nickthetait:
I can confirm that this fixed it for me on Debian Jessie.
Test procedure:
install Cura 2.1.0
Try running it, get the error message "RuntimeError: the sip module implements API v11.0 to v11.1 but the Arcus module requires API v11.2"
Add the 4 lines into /opt/cura/bin/cura
Try running it, and it works :relaxed:

thopiekar:
Great @nickthetait.

I just noticed that I missed something, but now the PR should be complete. Let's see what @awhiemstra says about that (tomorrow) :wink:

cura_app.py
```diff
@@ -3,8 +3,23 @@
...
 
+#WORKAROUND: GITHUB-704 GITHUB-708
+# It looks like setuptools creates a .pth file in
+# the default /usr/lib which causes the default site-packages
+# to be inserted into sys.path before PYTHONPATH.
+# This can cause issues such as having libsip loaded from
+# the system instead of the one provided with Cura, which causes
+# incompatibility issues with libArcus
+if "PYTHONPATH" in os.environ.keys():                # If PYTHONPATH is used
+    PYTHONPATH_real = os.path.realpath(os.environ["PYTHONPATH"])
```
awhiemstra: Does realpath() work on PYTHONPATH if PYTHONPATH is something like "/home/ahiemstra/lib/python3:/home/ahiemstra/lib/python3.4:/home/ahiemstra/Projects/some_project"? Because that is a valid PYTHONPATH according to https://docs.python.org/3.4/using/cmdline.html#envvar-PYTHONPATH

thopiekar: No, I guess that won't work, so I say for now this is not implemented. Will
add this as soon as possible.

thopiekar: That looks interesting.. It works only on the first element, but it also makes sense - it just cares about the beginning of the string.

```
>>> os.path.realpath("BLA:/foo/bar:foobar")
'/home/thopiekar/BLA:/foo/bar:foobar'
```

That brings me to a good idea...

thopiekar:
@nickthetait Could you test it again, please?
I see now enough PYTHONPATHs in the code now now :sweat_smile:

nickthetait:
Retested and verified working :sunglasses:

thopiekar:
Thanks :+1:

kakaroto:
Great catch @awhiemstra.
The latest patch from @thopiekar isn't proper though. the sys.path variable is a list, so you can't insert "foo:bar" into it, it needs to be ["foo", "bar"]. Basically each element of the list needs to be a path, not a comma-separated list of paths. 
The way I'd do it is to make PYTHONPATH_real into a list, then check only the first element, then append the two lists:

```
    PYTHONPATH = os.environ["PYTHONPATH"]
    PYTHONPATH = PYTHONPATH.split(os.pathsep)
    PYTHONPATH_real = []
    for path in PYTHONPATH:
         PYTHONPATH_real.append(os.path.realpath(path))
     if sys.path[1] != PYTHONPATH_real[0]:      # .. check whether PYTHONPATH is placed incorrectly.
         for path in PYTHONPATH_real:
             sys.path.remove(path)    # If so, remove all PYTHONPATH elements..
         sys.path = sys.path[:1] + PYTHONPATH_real + sys.path[1:] # and add it at the correct place again.
```

thopiekar:
Oh, yes, I need to take a look on it again :disappointed:

thopiekar:
@kakaroto What do you think?

kakaroto:
@thopiekar nice, logic looks good to me, and I like the way you did it.
Only thing is a 'typo' in the comment "# Insert it at 1 before os.curdir, which is 0." should be 'after'. Looks good to me otherwise.
@nickthetait time to test/confirm it works ? @awhiemstra time to review it again ?

Ghostkeeper:
It's so awesome to see you guys working together on this.

nickthetait:
Retested and .... failure 
`Traceback (most recent call last):
  File "/opt/cura/bin/cura", line 19, in <module>
    for PATH in PYTHONPATH:                                 # Now beginning with the last PATH
TypeError: 'NoneType' object is not iterable`

line 18 needs to be `PYTHONPATH.reverse()                                    # and reverse it, because we always insert at 1` because apparently python list methods work in place and return `None`. With this line changed everything works.

Thanks for all the great hacking @thopiekar. It is hard to develop anything without a test environment!

kakaroto:
@thopiekar it doesn't work. @nickthetait told me it fails with :

``` File "/opt/cura/bin/cura", line 19, in <module>
    for PATH in PYTHONPATH:                                 # Now beginning with the last PATH
TypeError: 'NoneType' object is not iterable```

I found that the list.reverse() function reverses the list in place, but it returns nothing, so the :
`PYTHONPATH = PYTHONPATH.reverse()`
actually assigns `None` to `PYTHONPATH`.
Instead it should be :
`PYTHONPATH.reverse()`

thopiekar:
@nickthetait @kakaroto There was never a PR before where I ever facepalmed myself that much like here.
The fix is easy ```.reverse()``` only modifies the current variable and returns not the result just ```None```
Please remove the ```PYTHONPATH = ``` in line 18.

kakaroto:
@thopiekar lol, same here, I kept facepalming at my 'looks good' reviews when they weren't. And it's such a trivial thing to fix too and yet it looks complicated when looking at all the comments here :)
I also just realized that os.pathsep is NOT the same thing as os.path.sep .. Good job python devs!

Thanks @thopiekar for the fix and good support :)

thopiekar:
No problem. At the beginning I also thought it will be very easy to fix
this 😉

By the way, Github is really missing a 'facepalmed smiley' 😂

awhiemstra:
Technically speaking the variables PYTHONPATH and PATH are not entirely according to our code standards but considering that the environment variables are that way I am fine with it, so merging this.

nickthetait:
I just tried out 2.1.1 which still exhibits this problem. I think this PR needs to be added to 2.1 branch too

PR #725, opened 2016-04-27
Uniqbot by Unimatech machine profile

zerr0:


nallath:
Note that machine_head_shape_min_x is no longer used. 
If you want to have a correct head size, use  machine_head_polygon && machine_head_with_fans_polygon

PR #727, opened 2016-04-29
Add Printrbot Play Definitions

cruisetheblues:
added two printrbot play definition files (.json). One for a standard
printrbot play, the other for a printrbot play with the optional larger
heated bed.

### DIFF
--- a/resources/machines/printrbot-play-htbed.json
+++ b/resources/machines/printrbot-play-htbed.json
@@ -1,0 +1,63 @@
+{
+    "id": "printrbot-play-htbed",
+    "name": "Printrbot Play With Heated Bed",
+    "manufacturer": "Printrbot",
+    "version": 1,
+    "inherits": "fdmprinter.json",
+    "machine_settings": {
+        "machine_width": {
+            "default": 100
+        },
+        "machine_height": {
+            "default": 130
+        },
+        "machine_depth": {
+            "default": 203
+        },
+        "machine_center_is_zero": {
+            "default": false
+        },
+        "machine_nozzle_size": {
+            "default": 0.4
+        },
+        "machine_head_shape_min_x": {
+            "default": 75
+        },
+        "machine_head_shape_min_y": {
+            "default": 18
+        },
+        "machine_head_shape_max_x": {
+            "default": 18
+        },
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Looks good, but for one thing. Ultimaker's marketing team wants to have all printers other than Ultimaker's in the "Other" category in the add-machine wizard. This means that the manufacturer has to be "Other". I'm sorry :( Could you change that for us?

fieldOfView:
@Ghostkeeper Er... https://github.com/Ultimaker/Cura/pull/725

Ghostkeeper:
Ah, we missed it there apparently.

Ghostkeeper:
Fixed in d8c53f7a1f0d59ed3663329c6891d3c1462b7c21.

nickthetait:
I don't understand. Is only Ultimaker allowed to set the manufacturer field correctly?

awhiemstra:
The manufacturer field is currently incorrectly named and is in reality used as category. This will be fixed for 2.2.

cruisetheblues:
That's kind of a bummer, sorting by manufacturer would be very helpful but I will change it. Thanks.

nallath:
@nickthetait; We were asked for the release that we "ship" to have two categories:
"Ultimaker" & "Other". We actually misuse the manufacturer field for this (subdividing stuff into "made by us" and !"made by us"). This is actually a nasty hack and we shouldn't do it that way.
If you look at the settings_rework branch, you can see that we're already making changes to do this right.

In my opinion we should set the data to be right and ensure that certain wizards can filter based on the data. 

So no worries, this by no means a way of Ultimaker to control stuff.

nickthetait:
I see what you mean now, thx

thopiekar:
I see a lot of duplicate values compared with ```printrbot-play```. Maybe you can try to setup ```printrbot-play-htbed``` to inherit ```printrbot-play```, which is also done on Ultimaker printers with upgrades.

For example: https://github.com/Ultimaker/Cura/blob/master/resources/machines/ultimaker2_extended_plus_025.json#L10

BagelOrb:
This has been fixed by another PR, which got merged.

Issue #697, opened 2016-04-12
Drop down window for object selection.


Gareee:
Loading an object that scaled properly makes selecting them fairly easy,l, but if you load something scaled way too small, if becomes a real pain to select them. You have to scroll in as far as you can, and fiddle with the preview rotation to see if you can get the object into view to select it. If there was a drop down object selector, you could just select the obj by its loaded name. This would also come in handy for objects you scale up, which then push other objects way off the print bed space, that you need to relocate. Once selected in the dropdown, objects should not only have focus, but the preview display should also auto center on them as well for faster manipulation back to the print bed.

fieldOfView:
An object list / "scene graph" could also help for selecting objects once they are grouped, so their Per Object Settings can be adjusted.

Ghostkeeper:
Hehe, please warn me before you tell Jaime that we're implementing that (if we are). I'd like to be outside of the blast zone before then.

nallath:
Ugh. Meshlist panel. My years old Nemesis. Are you now trying to infect Cura as well? ;)

fieldOfView:
If it improves UX, yes. 
Sounds like I will have to hear you out on the faults (and merits) of meshlists over a couple of beers (or cookies)

nallath:
I love it when it works. It's just very difficult to get all the behaviours right.

Gareee:
A panel was just my first thought.. there could easily be a menu option listing the loaded objects for selection. Or maybe even a icon based drop down list at the top of the translation overlay controls to select the loaded objects that way before using rotate/scale, ect.

Also a question that just sprung to mind.. in the older cura, you can make a selectable number of duplicates.. in the new version, you can only make one at a time. Will the multiple duplicates option be returning in a new build?

awhiemstra:
The multiple duplicate option will most likely not be returning as such. Instead we will be making sure duplicate works on multiple objects at the same time and we want to introduce copy-paste functionality.

Gareee:
Copy and paste might do the trick then, since you could repeatedly paste the same object down. The reason I asked, was I created some game tokens (Like 1 1/4" discs) and in the old cura I was printing 18 of them at a time. I loaded on, replicated it 17 times, and was pretty much done. Thats obviously a lot longer process in the new version.

I also recorded a video to show the issue I'm having in some cases selecting a loaded object, and why I'd like to either see them automatically focused upon loading, or to be able to select them from a list of loaded objects somewhere. 

[...]

PR #733, opened 2016-05-04
Add a preference for scaling up tiny objects

fieldOfView:
When objects have been modeled with meters as units, they become so tiny that the cannot be selected in Cura to be scaled up.

See https://github.com/Ultimaker/Cura/issues/697#issuecomment-209518061

Also see https://github.com/Ultimaker/Uranium/pull/132


Ghostkeeper:
I thought we already did something like this automatically?

fieldOfView:
Only the other way around, scaling down large meshes.

nallath:
No reason not to add this. It's simple enough and fixes the problem in almost all use cases.

PR #737, opened 2016-05-10
Listing of all paths for every OS

thopiekar:
Only cosmetics

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,10 @@ Use [this](https://github.com/Ultimaker/Uranium/wiki/Bug-Reporting-Template) tem
 For crashes and similar issues, please attach the following information:
 
 * (On Windows) The log as produced by dxdiag (start -> run -> dxdiag -> save output)
-* The Cura GUI log file, located at (Windows) $User/AppData/Local/cura/cura.log, (OSX) $User/.cura/cura.log, (Ubuntu) $USER/.local/share/cura
+* The Cura GUI log file, located at
+  * $User/AppData/Local/cura/cura.log (Windows)
+  * $User/.cura/cura.log (OSX)
+  * $USER/.local/share/cura (Ubuntu/Linux)
 * The Cura Engine log, using Help -> Show Engine Log
 
 Dependencies


README.md
```diff
@@ -13,7 +13,10 @@ Use [this](https://github.com/Ultimaker/Uranium/wiki/Bug-Reporting-Template) tem
 For crashes and similar issues, please attach the following information:
 
 * (On Windows) The log as produced by dxdiag (start -> run -> dxdiag -> save output)
-* The Cura GUI log file, located at (Windows) $User/AppData/Local/cura/cura.log, (OSX) $User/.cura/cura.log, (Ubuntu) $USER/.local/share/cura
+* The Cura GUI log file, located at
+  * $User/AppData/Local/cura/cura.log (Windows)
```
fieldOfView: More correct:
%USERPROFILE%\AppData\Local\cura\cura.log
$User does not exist on Windows, and Windows uses backslashes

README.md
```diff
@@ -13,7 +13,10 @@ Use [this](https://github.com/Ultimaker/Uranium/wiki/Bug-Reporting-Template) tem
...
+  * $User/.cura/cura.log (OSX)
+  * $USER/.local/share/cura (Ubuntu/Linux)
```
fieldOfView: Since $USER evaluates to the user name, not the user home, wouldn't ~ be better?

thopiekar: I would even use $HOME for Linux here. I guess OSX will have the same here or something similar.
For Windows we could use %APPDATA% I think. But this is less self-explaining.

Moreover, if this "guide" is meant to be used by the end-users I wouldn't expect they all know what the variables mean. So instead of using the variables or relative paths I would mention the "real" paths, like ```/home/<username>/.local/cura/cura.log```. Same for the other OS's.
The only problem I see with Windows is, that ```%APPDATA%``` has been moved for many times, when looking into the past.

awhiemstra: Also, in master $HOME/.cura is invalid for OSX, it uses $HOME/Library/Applciation Support/cura

PR #741, opened 2016-05-14
BQ Hephestos2: Removing override of support_enable

thopiekar:
"support_enable" is already set as "enabled": false in fdmprinter.json.
So this line is unneeded here.


nallath:
As usual; Thanks for the contribution :)

thopiekar:
No problem :wink: (I guess bq should be more thankful than you for caring about their product here.)

PR #742, opened 2016-05-14
BQ Hephestos2: Adding missing machine_preferences

thopiekar:
Without that Cura will by default display "ABS" as material (first material in the list).
This will add the missing material setup to the JSON.


fieldOfView:
I'm not sure machine_preferences will survive the work currently going on for 2.2.
If that construct does survive the rework, I think it may be worth adding the preference to fdmprinter.json instead of putting it on many different definitions.

thopiekar:
Makes sense, because we can expect that almost every fdmprinter.json can
print PLA in "normal" quality.

PR #771, opened 2016-06-06
Splashscreen: Another (and last) correction

thopiekar:
Now all the needed changes should be done.

### DIFF
--- a/cura/CuraSplashScreen.py
+++ b/cura/CuraSplashScreen.py
@@ -23,7 +23,7 @@ class CuraSplashScreen(QSplashScreen):
         version = Application.getInstance().getVersion().split("-")
         buildtype = Application.getInstance().getBuildType()
         if buildtype:
-            version += " (%s)" %(buildtype)
+            version[0] += " (%s)" %(buildtype)
 
         painter.setFont(QFont("Proxima Nova Rg", 20 ))
         painter.drawText(0, 0, 330 * self._scale, 230 * self._scale, Qt.AlignHCenter | Qt.AlignBottom, version[0])


Ghostkeeper:
The splash screen still displays the same for me (from source), and the code changes seem like an improvement, so down the merge funnel it goes!

thopiekar:
The first part of this was merged by @nallath in saturday. You'll have to set CURA_BUILDTYPE in cmake to see any changes.

For me it is working now and showing the branding ”(PPA)" appended to the version number.

BTW. We are splitting there by "-", but what is usually being splitted there? Is it some kind of version comment?
I'm not that familiar with versioning...

Ghostkeeper:
Yes, I saw that merge from Nallath, so I was aware of what you were working on :smile:

We'd usually be splitting on `.` since we number versions with periods, as in `2.1.2`. But I don't know by heart what our build server puts in CURA_VERSION right now.

thopiekar:
Hmm, the set CURA_VERSION in CMake goes as string to the splash screen and
to the error reporting plugin. In the splash screen code there it is
splitted by "-". I guess you wanted to use "2.1.2-Beta" here, so in the
first line you'll have the version number and in the second line the
comment that this is a beta release.
When looking at the latest beta for Windows, this doesn't seem to be used.
Only "2.1.1" appears..

awhiemstra:
That's because that build is simply called 2.1.1, no extra version so nothing after the -.

thopiekar:
Ok, what about calling the second row directly CURA_VERSION_BUILDDATE (CMake variable)?

So you think designing the dialog wouldn't look like this?:

$CURA_VERSION ($CURA_BUILD_TYPE)
$CURA_EXTRA_VERSION

But I agree, using BUILD_TYPE is there a bit misleading.. The idea was to add a suffix which represents the way CURA was distributed 🤔

PR #779, opened 2016-06-08
[master] CURA-1597: Windows unicode fix

thopiekar:
This should remove the wrong character for Cura master on Windows.
If you check the diff you see actually no difference, but on Windows there is.

#CURA-159


thopiekar:
Ok, this PR is crap. I guess this PR is the right solution for it: https://github.com/Ultimaker/Uranium/pull/136

PR #781, opened 2016-06-09
Don't ignore the directory but the link, too.

thopiekar:


fieldOfView:
I'm going to need a bit more context. What/who does this do/help?

thopiekar:
Just helps me at the moment I guess. It isn't related to any CURA-* task. Just added a this ignore before as I had the debian/ folder, but now I replaced it with a symlink and now it works with both.
When building debian packages you'll need the debian folder in the root of the source, but at the moment this is not part of the Cura code here..

Ghostkeeper:
It helps Debian developers with having a cleaner repository when they type `git status`, more easily add files with tab completion, and not accidentally commit compiled binaries if they'd be so stupid to do `git add -A`.

Can't hurt for the rest unless we're making files in the root directory that start with `debian`.

PR #782, opened 2016-06-09
[master] Adding missing colons

thopiekar:
Only cosmetics.. Just found these missing colons when looking for a reason for the non-working-shortcuts, so just added them.

(CURA-1603)

### DIFF
--- a/resources/qml/Actions.qml
+++ b/resources/qml/Actions.qml
@@ -54,7 +54,7 @@ Item
 
     Action
     {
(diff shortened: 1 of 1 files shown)


fieldOfView:
We're actually leaning towards _not_ having the semicolons in newer code.
From https://wiki.qt.io/Qt_Quick_Best_Practices:
> Multiple properties on same line are separated by a semicolon 
Other than that, there should be few semicolons (only where readability is improved by having them)

PR #787, opened 2016-06-10
Minor Exception handling Fixes

awhiemstra:
This fixes an issue where an exception in MachineManager::setActiveMachine would completely prevent the application from starting. In addition, I have improved the uncaught exception handling in Cura so it is a bit more useful and properly aborts on certain kinds of exceptions.

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -5,6 +5,8 @@ import webbrowser
 
 from PyQt5.QtCore import QT_VERSION_STR, PYQT_VERSION_STR, QCoreApplication
 from PyQt5.QtWidgets import QDialog, QDialogButtonBox, QVBoxLayout, QLabel, QTextEdit
+
+from UM.Logger import Logger
 from UM.i18n import i18nCatalog
 catalog = i18nCatalog("cura")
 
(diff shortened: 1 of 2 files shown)


cura/CrashHandler.py
```diff
@@ -34,7 +45,7 @@ def show(exception_type, value, tb):
 
     label = QLabel(dialog)
     layout.addWidget(label)
-    label.setText(catalog.i18nc("@label", "<p>An uncaught exception has occurred!</p><p>Please use the information below to post a bug report at <a href=\"http://github.com/Ultimaker/Cura/issues\">http://github.com/Ultimaker/Cura/issues</a></p>"))
+    label.setText(catalog.i18nc("@label", "<p>A fatal exception has occurred that we could not recover from!</p><p>Please use the information below to post a bug report at <a href=\"http://github.com/Ultimaker/Cura/issues\">http://github.com/Ultimaker/Cura/issues</a></p>"))
```
sedwards2009: Why is the application referring to itself as 'we'? ;-)

nallath: https://en.wikipedia.org/wiki/Royal_we

Ghostkeeper: Maybe the collection of bits is referring to themselves as "we"?

PR #791, opened 2016-06-11
[master] Importing and using Logger correctly

thopiekar:
Usually we import Logger a different way across our code. Fixing this
kind of typo just keeps this convention.

### DIFF
--- a/cura/SettingOverrideDecorator.py
+++ b/cura/SettingOverrideDecorator.py
@@ -8,7 +8,7 @@ from UM.Signal import Signal, signalemitter
 from UM.Settings.ContainerStack import ContainerStack
 from UM.Settings.InstanceContainer import InstanceContainer
 from UM.Settings.ContainerRegistry import ContainerRegistry
-import UM.Logger
+from UM.Logger import Logger
 
 from UM.Application import Application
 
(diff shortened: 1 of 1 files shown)


thopiekar:
Oh, btw. I guess it also should be called UM.Logger.Logger.log(*args) before, as the UM.Logger module includes a Logger class, which finally has .log(*args).

fieldOfView:
@awhiemstra has recently introduced a new way of importing UM modules. Importing UM.Logger and using UM.Logger.log() should work. Also see ExtruderManager.py

thopiekar:
I left ExtruderManager.py unchanged, because I thought you are still working on it.
I can review this file, too, if you like.

So are we going to change the way of importing the modules in general?

fieldOfView:
> So are we going to change the way of importing the modules in general?

In general, I believe new code is supposed to use the new-style imports. But because we have been working towards a release, most of us did not change our habbits yet. This will likely change the coming week(s).

nallath:
The "new" style imports are usefull for circular dependencies and having less issues with namespace polution.

thopiekar:
Ok, if you like you can merge this. At least this would keep one coding style being used at the moment.

thopiekar:
Closing this?

fieldOfView:
Yup

PR #792, opened 2016-06-11
[master] code cleanup

thopiekar:
Just take a look at the two commits below.

### DIFF
--- a/cura/SettingOverrideDecorator.py
+++ b/cura/SettingOverrideDecorator.py
@@ -51,8 +51,8 @@ class SettingOverrideDecorator(SceneNodeDecorator):
     def getActiveExtruder(self):
         return self._extruder_stack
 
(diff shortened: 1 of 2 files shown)


cura/SettingOverrideDecorator.py
```diff
@@ -51,8 +51,8 @@ def __deepcopy__(self, memo):
     def getActiveExtruder(self):
         return self._extruder_stack
 
-    def _onSettingChanged(self, instance, property):
-        if property == "value":  # Only reslice if the value has changed.
+    def _onSettingChanged(self, instance, attribute): # Reminder: 'property' is a built-in function
+        if attribute == "value":  # Only reslice if the value has changed.
```
fieldOfView: I would prefer ```property_name``` over ```attribute```. The word "property" is fairly embedded in the settings architecture (a setting has several properties such as type, value, min_value, etc), so using a synonym does not help in the clarity of what is being passed here. Good catch though.

PR #795, opened 2016-06-12
[master] CuraEngineBackend: Removing duplicate close()

thopiekar:
close() is already defined in this class. This removes the useless code.


Ghostkeeper:
Whoa, that might have resulted in some nasty bugs where sockets are kept open.

PR #797, opened 2016-06-12
Adding MM2 as well as a solid model for bed

XistorLogic:
Wanted to add the MendelMax 2 as a printer option.

Believe I have everything in the right locations, will gladly move/fix if needed.


thopiekar:
Thank you for your PR :+1:

There are 3 things I worry about:
1. You are using upper case letters in 'id', but only lower case letters should be used.
2. Additionaly you are using spaces in your 'id'. Please replace them with '_'
3. The filename of your STL does  not fit to: "<id>_platform.stl"

Ghostkeeper:
There's a few more. As you've seen, we changed the file format a bit since version 2.1. Here's what needs to be fixed for it to be working with 2.2:

- [ ] `version` should be `2`.
- [ ] `manufacturer`, `author`, `icon`, `platform` and `file_formats` should be in a subnode called `metadata`.
- [ ] Instead of defining `default` for each of the settings, define `default_value`.
- [ ] The `inherits` field no longer takes the extension, so it should just be `fdmprinter`.
- [ ] It's probably wise to make the `id` be equal to the file name without extension for now.
- [ ] We are using the extension `.def.json` for definition files. Otherwise it won't get picked up by our mime type detection as being a machine definition.
- [ ] Instead of setting `manufacturer` to other, we now use a new metadata field called `category` (also in `metadata`). This should be set to `Other` for now. Instead, you can set `manufacturer` to be the actual manufacturer of the printer again!

I'm sorry if it's too much work. Would you need help?

awhiemstra:
> It's probably wise to make the id be equal to the file name without extension for now.

Actually, you can just leave out the ID, it is no longer used anywhere, the internal ID we use is generated from the file name instead.

thopiekar:
Thank you for the info! 👍

XistorLogic:
Thanks for the feedback! Will change asap

BagelOrb:
@XistorLogic don't forget!

nickthetait:
... you still around @XistorLogic ?

BagelOrb:
This definition is outdated it cannot be merged.

PR #800, opened 2016-06-14
Formulate layerview logic using numpy to speed it up. 

Johan3DV:
Also changed layer data packets from engine to make it possible.

Needs the corresponding pull requests to Uranium and CuraEngine to work.


Johan3DV:
The goal of this pull request is to speed-up the current layer view of Cura without restricting further development with more features. The following comment can also be found in an accompanying PR for CuraEngine..

The way I see it the problem with the current layer view implementation is that the data is sent from the engine to the front-end in polygon packets that are really small. This leads to for-loops with many iterations in the front-end code which are really slow.

The current implementation of sending polygons to the front-end sends polygons that contain a single line segment each. This is not only inefficient in terms of memory usage but also make the processing in Cura slow. To start out with I added a class in the back-end called PolygonCompiler which buffered polygons of the same type and sent them to the front-end packed together into larger polygons to mitigate the problem.

It turned out to only help so much because of the structure of the actual data. The way a polygon is currently defined, it's a collection of lines of the same type (infill, skin, etc) with a common line width. This is not really a good way of representing the data CuraEngine produces. Line widths vary and only the insets contain continuous segments of the same type. It seems more appropriate to pack the layer data in another way. The polygon message sent from the engine to the front-end in this PR has been redefined so that for each line segment the polygon defines the line width and the type. This also allows the computations in the front-end to be expressed much more efficiently using numpy.

**The big problem with the implementation in this PR is that it's currently a protocol breaking change**. CuraEngine and Cura from before and after this PR won't be compatible for viewing polygons.

I think the alternative to this protocol breaking change is to make another type of layer message in the protocol which allows one to define another type of polygon according to the one in this PR. A front-end supporting these later protocol version would then need to notify the back-end which type of message to use and if no such notification is received the back-end falls back to the currently used type of layer/polygon message. These cases needs to be handled both in the front-end and in the back-end.

Resolving the protocol breaking changes would add more complexity to the code. I don't see any reason for using the current implementation of the polygon message in the future. Even if the code responsible for handling the data in the front-end was rewritten using C++ the current polygon is still not flexible enough. Another more subtle protocol breaking change would be sending the point coordinates of the polygon not as int64 but as floats, which are actually used for displaying them, which would reduce message size. The question arises whether the precision offered by int64 will be beneficial to Cura in the future or just be a memory overhead. There might be other ways of specifying the line data that would be better from a memory point of view.

nallath:
I don't see that much issue with breaking protocol. Especially not since this is a pretty big problem, we never got the time for to fix properly.

BagelOrb:
See the accompanying CuraEngine PR: https://github.com/Ultimaker/CuraEngine/pull/354

awhiemstra:
I agree with @nallath that protocol changes are not an issue. The protocol used is an internal implementation detail and _should_ be simple to change. We already changed it for 2.2 anyway because of other features, so having an addiitonal change is no issue anyway.

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
         return self.createMeshOrJumps(False)
 
     def createMeshOrJumps(self, make_mesh):
-        builder = MeshBuilder()
-
+        builder = MeshBuilder() # This is never really used, only the mesh_data inside
+        mesh_data = builder._mesh_data
```
awhiemstra: This is changing with the simple_convex_hull branch, but this is actually a pretty bad practice. You are accessing a protected instance variable from outside. Note that python has no real concept of protected or private, but by convention anything starting with an underscore should at the very least be treated as protected. Note that there is a getter for MeshBuilder's mesh data (getMeshData) so there is no need for this. In addition, you actually would not need MeshBuilder and could just have used MeshData's constructor directly, but this is changing for the simple_convex_hull branch.

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
+        index_pattern = numpy.array([[0,3,2,0,1,3]],dtype = numpy.int32 )
```
awhiemstra: Codestyle: space after ,

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
+        if make_mesh:
+            for polygon in self._polygons:
+                line_count += polygon._mesh_line_count
+        else:
+            for polygon in self._polygons:
+                line_count += polygon._jump_count
+            
+        
+        # Reserve the neccesary space for the data upfront
+        mesh_data.reserveFaceAndVerticeCount( 2*line_count, 4*line_count )
```
awhiemstra: Codestyle: space around operators, not after and before ( ).

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
-                normal = normals[i - 1]
-
-                point1 = Vector(data = start - normal)
-                point2 = Vector(data = start + normal)
-                point3 = Vector(data = end + normal)
-                point4 = Vector(data = end - normal)
-
-                builder.addQuad(point1, point2, point3, point4, color = poly_color)
-
+            #if make_mesh and (polygon.type == LayerPolygon.MoveCombingType or polygon.type == LayerPolygon.MoveRetractionType):
```
awhiemstra: Commented code can just be removed, no need to keep it around, git does that for us.

cura/LayerPolygon.py
```diff
@@ -14,52 +14,108 @@ class LayerPolygon:
...
+        self._color_map = self.__color_map
+        
+        # type == LayerPolygon.InfillType or type == LayerPolygon.SkinType or type == LayerPolygon.SupportInfillType
+        # Should be generated in better way, not hardcoded.
+        self._orInfillSkin = numpy.array([0,0,0,1,0,0,1,1,0,0],dtype=numpy.bool)
+        
+        self._build_cache_line_mesh_mask = None
+        self._build_cache_needed_points = None
+        
+    def build_cache(self):
```
awhiemstra: Codestyle: method names should use lowerCamelCase, not underscores

cura/LayerPolygon.py
```diff
@@ -14,52 +14,108 @@ class LayerPolygon:
...
+        self._vertex_count = self._mesh_line_count + numpy.sum( self._types[1:] == self._types[:-1])
+
+        # Buffering the colors shouldn't be necessary as it is not 
+        # re-used and can save alot of memory usage.
+        self._colors = self.__color_map[self._types]
+        self._color_map = self.__color_map
+        
+        # type == LayerPolygon.InfillType or type == LayerPolygon.SkinType or type == LayerPolygon.SupportInfillType
+        # Should be generated in better way, not hardcoded.
+        self._orInfillSkin = numpy.array([0,0,0,1,0,0,1,1,0,0],dtype=numpy.bool)
```
awhiemstra: Not really happy with the naming here. I had to really look what this variable means.

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
+            # Create an array with normals and tile 2 copies to match size of points variable
+            normals = numpy.tile( polygon.getNormals()[index_mask.ravel()], (1,2))
+
+            # Scale all normals by the line width of the current line so we can easily offset.
+            normals *= (polygon.lineWidths[index_mask.ravel()] / 2)
+
+            # Create 4 points to draw each line segment, points +- normals results in 2 points each. Reshape to one point per line
+            f_points = numpy.concatenate((points-normals,points+normals),1).reshape((-1,3))
+            # index_pattern defines which points to use to draw the two faces for each lines egment, the following linesegment is offset by 4 
+            f_indices = ( index_pattern + numpy.arange(0,4*len(normals),4,dtype=numpy.int32).reshape((-1,1)) ).reshape((-1,3))
```
awhiemstra: I'd rather see this written out a bit more. Numpy is hard enough to follow without doing multiple operations in a single line.

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
+            #    continue
+            #if not make_mesh and not (polygon.type == LayerPolygon.MoveCombingType or polygon.type == LayerPolygon.MoveRetractionType):
+            #    continue
+            
+            index_mask = numpy.logical_not(polygon._jump_mask) if make_mesh else polygon._jump_mask
+
+            # Create an array with rows [p p+1] and only save those we whant to draw based on make_mesh
+            points = numpy.concatenate((polygon.data[:-1],polygon.data[1:]),1)[index_mask.ravel()]
+            # Line types of the points we want to draw
+            line_types = polygon._types[index_mask]
```
awhiemstra: Protected member access outside a subclass.

cura/Layer.py
```diff
@@ -61,39 +68,59 @@ def createJumps(self):
...
-                point4 = Vector(data = end - normal)
-
-                builder.addQuad(point1, point2, point3, point4, color = poly_color)
-
+            #if make_mesh and (polygon.type == LayerPolygon.MoveCombingType or polygon.type == LayerPolygon.MoveRetractionType):
+            #    continue
+            #if not make_mesh and not (polygon.type == LayerPolygon.MoveCombingType or polygon.type == LayerPolygon.MoveRetractionType):
+            #    continue
+            
+            index_mask = numpy.logical_not(polygon._jump_mask) if make_mesh else polygon._jump_mask
```
awhiemstra: Protected member access outside a subclass.

cura/LayerPolygon.py
```diff
@@ -97,3 +154,16 @@ def getNormals(self):
...
+        [1.0, 1.0, 1.0, 1.0],
+        [1.0, 0.0, 0.0, 1.0],
+        [0.0, 1.0, 0.0, 1.0],
+        [1.0, 1.0, 0.0, 1.0],
+        [0.0, 1.0, 1.0, 1.0],
+        [0.0, 1.0, 1.0, 1.0],
+        [1.0, 0.74, 0.0, 1.0],
+        [0.0, 1.0, 1.0, 1.0],
+        [0.0, 0.0, 1.0, 1.0],
+        [0.5, 0.5, 1.0, 1.0]])
```
awhiemstra: Please put the closing brace on a newline so it becomes clearer where things end.

cura/LayerPolygon.py
```diff
@@ -14,52 +14,108 @@ class LayerPolygon:
...
+        self._vertex_end += vertex_offset
+        
+        vertices[self._vertex_begin:self._vertex_end, :] = self._data[index_list, :]
+        colors[self._vertex_begin:self._vertex_end, :] = numpy.tile(self._colors,(1,2)).reshape((-1,4))[needed_points_list.ravel()] 
+        colors[self._vertex_begin:self._vertex_end, :] *= numpy.array([[0.5, 0.5, 0.5, 1.0]], numpy.float32)
+
+        self._index_begin += index_offset
+        self._index_end += index_offset
+        
+        indices[self._index_begin:self._index_end,:] = numpy.arange(self._index_end-self._index_begin, dtype=numpy.int32).reshape((-1,1))
```
awhiemstra: Like above, please write this out a bit more or add some additional documentation. The reason we did not use numpy before is because the Cura 15.04 code was completely undocumented and incredibly hard to follow. I would like to avoid the same issue with this code.

cura/Layer.py
Johan3DV: This has been adapted to use MeshBuilder after simple_convex_hull was merged.

cura/LayerPolygon.py
Johan3DV: I've added some more documenting comments to the code. The problem with writing out this code is that each individual statement don't have much merit on it's own and I think it's therefore hard to name.

sedwards2009:
`LayerPolygon.build()` could probably be made a little bit faster by caching values from `self` using a local variable instead of using `self.foo_bar` the whole time. Also there are a couple numpy arrays which have constant contents. They are a candidate for being reused and not reallocated everytime `build()` runs.

Johan3DV:
As far as I know LayerPolygon.build() is only called once.

Johan3DV:
The real reason for having the buildCache() and build() separately in LayerPolygon was build() call happens outside the progress calculated part in ProcessSlicedLayersJob. Perhaps the better option is to merge buildCache() and build() and re write the progress calculation.

cura/Layer.py
```diff
@@ -62,38 +69,48 @@ def createJumps(self):
 
     def createMeshOrJumps(self, make_mesh):
         builder = MeshBuilder()
-
+        index_pattern = numpy.array([[0, 3, 2, 0, 1, 3]], dtype = numpy.int32 )
```
awhiemstra: You could probably make this a class member and save recreating the index_pattern every time this method is called.

awhiemstra:
Hi Johan,

There are currently merge conflicts. Can you merge master into your branch and update the PR?

Johan3DV:
I will have a look at this later tonight.

I'm not sure if something should be done about the LayerPolygon naming. As discussed in the PR to CuraEngine, LayerPolygon is not representing a Polygon but more of a general path segment. Should LayerPolygon be renamed or should a new class LayerPathSegment be created?

fieldOfView:
I just saw this in action for the first time. Wow. Just wow.
Thanks @Johan3DV !

PR #803, opened 2016-06-15
[CURA-1615] Updating firmware manager for API 3

thopiekar:
This updates the firmware updater to use the new MachineManagerModel

Additionally renamed "machine_type" to "machine_id", which makes it more
selfexplaining, when reading the code.

My printers firmware is not available here and has no heated bed, so
please test this commit!

### DIFF
--- a/cura/MachineManagerModel.py
+++ b/cura/MachineManagerModel.py
@@ -486,13 +486,6 @@ class MachineManagerModel(QObject):
         return False
 
     @pyqtProperty(bool, notify = globalContainerChanged)
-    def hasHeatedBed(self):
-        if self._active_container_stack:
-            return bool(self._global_container_stack.getMetaDataEntry("machine_heated_bed", False))
-
-        return False
-
-    @pyqtProperty(bool, notify = globalContainerChanged)
     def hasVariants(self):
         if self._active_container_stack:
             return bool(self._global_container_stack.getMetaDataEntry("has_variants", False))
--- a/cura/MachineManagerModel.py
+++ b/cura/MachineManagerModel.py
@@ -486,6 +486,13 @@ class MachineManagerModel(QObject):
         return False
 
     @pyqtProperty(bool, notify = globalContainerChanged)
+    def hasHeatedBed(self):
+        if self._active_container_stack:
+            return bool(self._global_container_stack.getMetaDataEntry("machine_heated_bed", False))
+
+        return False
+
+    @pyqtProperty(bool, notify = globalContainerChanged)
     def hasVariants(self):
         if self._active_container_stack:
             return bool(self._global_container_stack.getMetaDataEntry("has_variants", False))
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -128,11 +128,18 @@ class USBPrinterOutputDeviceManager(QObject, SignalEmitter, OutputDevicePlugin, 
         return USBPrinterOutputDeviceManager._instance
 
     def _getDefaultFirmwareName(self):
+        # Check whether getGlobalContainerStack() returns None or not...
+        global_container_stack = Application.getInstance().getGlobalContainerStack()
+        if not global_container_stack:
+            Logger.log("c", "getGlobalContainerStack() returned None")
+            Logger.log("i", "Closing firmware-updater UI as a consequence")
+            self._firmware_view.close()
+        
         # Detecting id of the current machine
-        machine_id = Application.getInstance().getGlobalContainerStack().getBottom().id
+        machine_id = global_container_stack.getBottom().id
         
         # Detecting whether it has a heated bed
-        machine_has_heated_bed = Application.getInstance().getGlobalContainerStack().getProperty("machine_heated_bed", "value")
+        machine_has_heated_bed = global_container_stack.getProperty("machine_heated_bed", "value")
         
         if platform.system() == "Linux":
             baudrate = 115200
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -13,7 +13,6 @@ from UM.Qt.ListModel import ListModel
 from UM.Message import Message
 
 from cura.CuraApplication import CuraApplication
-from cura.MachineManagerModel import MachineManagerModel
 
 import threading
 import platform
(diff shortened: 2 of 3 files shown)


cura/MachineManagerModel.py
```diff
@@ -486,6 +486,13 @@ def hasMaterials(self):
         return False
 
     @pyqtProperty(bool, notify = globalContainerChanged)
+    def hasHeatedBed(self):
+        if self._active_container_stack:
+            return bool(self._global_container_stack.getMetaDataEntry("machine_heated_bed", False))
```
fieldOfView: machine_heated_bed is not a metadata entry

thopiekar: Is there any reference where I can find these values?

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +129,9 @@ def getInstance(cls, engine = None, script_engine = None):
         return USBPrinterOutputDeviceManager._instance
 
     def _getDefaultFirmwareName(self):
-        machine_instance = Application.getInstance().getMachineManager().getActiveMachineInstance()
-        machine_type = machine_instance.getMachineDefinition().getId()
+        machine_manager_model = MachineManagerModel()
```
fieldOfView: Initialising a MachineManager instance to get at a single setting value seems a bit wasteful (MachineManager comes at quite some overhead). Come to think of it I think you could also use ```Application.getInstance().getGlobalContainerStack().getProperty("machine_heated_bed", "value")``` to check the setting.

That way you don't have to introduce a single-use method to MachineManagerModel.

fieldOfView: ...and ```Application.getInstance().getGlobalContainerStack().getBottom().id``` to get at the machine id

cura/MachineManagerModel.py
thopiekar: OK.. I guess I found something..

fieldOfView: fdmprinter.def.json

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,12 @@ def getInstance(cls, engine = None, script_engine = None):
         return USBPrinterOutputDeviceManager._instance
 
     def _getDefaultFirmwareName(self):
-        machine_instance = Application.getInstance().getMachineManager().getActiveMachineInstance()
-        machine_type = machine_instance.getMachineDefinition().getId()
+        # Detecting id of the current machine
+        machine_id = Application.getInstance().getGlobalContainerStack().getBottom().id
```
awhiemstra: Note that getGlobalContainerStack() can return None. I am not sure if this method can be called in any situation where that is the case but if so, you might want to check for it.

thopiekar: What do you expect as reaction on this? Just close the dialog again and throw an critical message to the logs?

awhiemstra: Probably the easiest yes. We usually simply return a default value in these situations and let the caller code deal with it.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,19 @@ def getInstance(cls, engine = None, script_engine = None):
         return USBPrinterOutputDeviceManager._instance
 
     def _getDefaultFirmwareName(self):
-        machine_instance = Application.getInstance().getMachineManager().getActiveMachineInstance()
-        machine_type = machine_instance.getMachineDefinition().getId()
+        # Check whether getGlobalContainerStack() returns None or not...
+        global_container_stack = Application.getInstance().getGlobalContainerStack()
+        if not global_container_stack:
+            Logger.log("c", "getGlobalContainerStack() returned None")
+            Logger.log("i", "Closing firmware-updater UI as a consequence")
```
fieldOfView: I don't think this warrants a _critical_ error. A critical error means the application cannot (should not) continue. This hypothetical case means the function cannot complete. Also I don't think it is necessary to create 2 log entries. I suggest ```Logger.log("e", "There is no global container stack. Can not update firmware")```

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,19 @@ def getInstance(cls, engine = None, script_engine = None):
...
+            self._firmware_view.close()
+        
+        # Detecting id of the current machine
+        machine_id = global_container_stack.getBottom().id
```
fieldOfView: This comment seems a bit superfluous (we can read). If you want to add a comment here, say something like "the bottom of the containerstack is the machine definition")

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,19 @@ def getInstance(cls, engine = None, script_engine = None):
...
+        
+        # Detecting whether it has a heated bed
```
fieldOfView: Superfluous comment.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,19 @@ def getInstance(cls, engine = None, script_engine = None):
         return USBPrinterOutputDeviceManager._instance
 
     def _getDefaultFirmwareName(self):
-        machine_instance = Application.getInstance().getMachineManager().getActiveMachineInstance()
-        machine_type = machine_instance.getMachineDefinition().getId()
+        # Check whether getGlobalContainerStack() returns None or not...
```
fieldOfView: Superfluous comment.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -128,8 +128,19 @@ def getInstance(cls, engine = None, script_engine = None):
...
+        global_container_stack = Application.getInstance().getGlobalContainerStack()
+        if not global_container_stack:
+            Logger.log("c", "getGlobalContainerStack() returned None")
+            Logger.log("i", "Closing firmware-updater UI as a consequence")
+            self._firmware_view.close()
```
awhiemstra: You should also return after this, otherwise it will still try to update firmware.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
thopiekar: True :confused:

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
fieldOfView: Doh! missed that one

thopiekar: Hmm, I would ```return None``` here, but on the other hand ```return ""``` would be save here, isn't it?

awhiemstra: Since it is expected to return a string, I'd return an empty string.

cura/MachineManagerModel.py
Ghostkeeper: Also note that the global container stack may be None if there is no printer yet (i.e. the user is still in initial set-up).

thopiekar: Ok, then it is absolutely correct to close the dialog again.

PR #805, opened 2016-06-15
Differencing between Platform from UM.* and UM.Scene.*

thopiekar:
Just fixes the redefinition of Platform...


Ghostkeeper:
Thanks. One reason to use fully qualified names eh :wink:

thopiekar:
No need to thank me. I just made the break here 🙊

PR #816, opened 2016-06-20
Printrbot Z-height

Jerooney:
the Z-height was incorrect


nallath:
Test pull request

Ghostkeeper:
That's not the Z-height. That's the Y-coordinate of one of the sides of the print head. Since the PrintrBot has a print "arm" extending beyond one of the sides of the printer, the print head extends more or less to infinity in that direction (though wider than the build plate makes no sense of course; that is just to allow for modified printers).

nallath:
Yeah, this was a test pull request made by jeroen to get a feel for what pull requests are. Hence my comment. Just ignore it ;)

PR #810, opened 2016-06-16
[CURA-1615] Plugins: Updating GCodeProfileReader and SliceInfo

thopiekar:
I got the plugin most of all working. At least the "successfully"
imported profile XY" dialog appears. But sadly the profile does not
appear in the list of profiles after that. I can only guess something is
blocking here.
Additionally it should be noted that G-Code exported from Cura 2.1.x
does not work here anymore on Cura 2.2.x.

PR #818, opened 2016-06-22
[CURA-1445] SliceInfo: Providing data from all containers

thopiekar:
Depends on and includes #810 

Contributes to CURA-1445


fieldOfView:
It's a bit tricky to distill which commits of this PR are new, compared to PR 810.

thopiekar:
https://github.com/thopiekar/Cura/compare/master-CURA-1615...thopiekar:master-CURA-1445?expand=1

PR #819, opened 2016-06-22
Machine actions

nallath:
CURA-1385


cura/MachineActionManager.py
```diff
@@ -0,0 +1,142 @@
...
+from UM.Logger import Logger
+from UM.PluginRegistry import PluginRegistry  # So MachineAction can be added as plugin type
+
+from UM.Settings.ContainerRegistry import ContainerRegistry
+from UM.Settings.DefinitionContainer import DefinitionContainer
+
+from PyQt5.QtCore import QObject, pyqtSlot
+
+##  Raised when trying to add an unknown machine action as a required action
+class UnknownMachineAction(Exception):
```
fieldOfView: Exception-class names end in *Error, so UnknownMachineActionError

cura/MachineActionManager.py
```diff
@@ -0,0 +1,142 @@
...
+    pass
+
+
+##  Raised when trying to add a machine action that does not have an unique key.
+class NotUniqueMachineAction(Exception):
```
fieldOfView: Exception-class names end in *Error, I suggest MachineActionKeyExistsError

cura/MachineActionManager.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class MachineActionManager(QObject):
+    def __init__(self, parent = None):
+        super().__init__(parent)
+
+        self._machine_actions = {}  # Dict of all known machine actions
+        self._required_actions = {}  # Dict of all required actions by machine reference.
+        self._supported_actions = {}  # Dict of all supported actions by machine reference
+        self._first_start_actions = {}  # Dict of all actions that need to be done when first added by machine reference
```
fieldOfView: Suggestion: ```_add_printer_actions```, first start suggest first starting Cura (which is only half true)

cura/MachineActionManager.py
```diff
@@ -0,0 +1,142 @@
...
+        if action.getKey() not in self._machine_actions:
+            self._machine_actions[action.getKey()] = action
+        else:
+            raise NotUniqueMachineAction("MachineAction with key %s was already added. Actions must have unique keys.", action.getKey())
+
+    ##  Get all actions supported by given machine
+    #   \param machine The machine you want the supported actions of
+    #   \returns set of supported actions.
+    @pyqtSlot(str, result = "QVariantList")
+    def getSupportedActions(self, machine_id):
```
fieldOfView: To prevent confusion between stacks and definitions, I think it would be better to name the argument ```definition_id``` instead of ```machine_id```

plugins/UltimakerMachineActions/UMOCheckupMachineAction.qml
```diff
@@ -0,0 +1,271 @@
...
+                anchors.top: parent.width < checkupMachineAction.width ? parent.top : startCheckButton.bottom
+                anchors.topMargin: parent.width < checkupMachineAction.width ? 0 : UM.Theme.getSize("default_margin").height/2
+                anchors.left: parent.width < checkupMachineAction.width ? startCheckButton.right : parent.left
+                anchors.leftMargin: parent.width < checkupMachineAction.width ? UM.Theme.getSize("default_margin").width : 0
+                text: catalog.i18nc("@action:button", "Skip Printer Check");
+                onClicked: manager.setFinished()
+            }
+        }
+
+        Item
```
fieldOfView: This would benefit from being in a Grid instead of an Item; you would not have to do all the anchoring

resources/definitions/ultimaker_original.def.json
```diff
@@ -14,12 +14,7 @@
         "has_materials": true,
         "preferred_material": "*pla*",
         "preferred_quality": "*normal*",
-        "pages": [
-            "SelectUpgradedParts",
-            "UpgradeFirmware",
-            "UltimakerCheckup",
-            "BedLeveling"
-        ]
+        "supported_actions":[ "UMOCheckup", "UpgradeFirmware"]
```
fieldOfView: Missing bedlevel?
Nitpick to the max: extraneous leading space: ["UMOCheckup"

resources/qml/Cura.qml
```diff
@@ -771,6 +771,38 @@ UM.MainWindow
...
+        onMachineAdded:
+        {
+            firstRunWizard.start(id)
+        }
+    }
+
+    // Dialog to handle first run machine actions
+    UM.Wizard
+    {
+        id: firstRunWizard;
```
fieldOfView: It's not a first run wizard, but a machineActionsWizard

resources/definitions/ultimaker_original_plus.def.json
```diff
@@ -11,11 +11,7 @@
         "icon": "icon_ultimaker.png",
         "platform": "ultimaker2_platform.obj",
         "platform_texture": "UltimakerPlusbackplate.png",
-        "pages": [
-            "UpgradeFirmware",
-            "UltimakerCheckup",
-            "BedLeveling"
-        ]
+        "supported_actions":[ "UMOCheckup", "UpgradeFirmware", "BedLevel"]
```
fieldOfView: Same nitpick: extraneous leading space: ["UMOCheckup"

plugins/UltimakerMachineActions/UMOCheckupMachineAction.qml
nallath: I know. I just copied this code from the old wizards.

resources/definitions/ultimaker_original.def.json
nallath: Oh. Here i was thinking we only wanted bed level for the heated bed. Oh well, it's easy enough to add.

cura/MachineActionManager.py
nallath: Fixed

cura/MachineActionManager.py
nallath: Fixed

resources/qml/Cura.qml
nallath: Fixed

resources/qml/MachineAction.qml
```diff
@@ -0,0 +1,15 @@
...
+    // Connect the finished property change to completed signal.
+    property var finished: manager.finished
+    onFinishedChanged: if(manager.finished) {completed()}
+    signal completed()
+
+    function reset()
+    {
+        manager.reset()
+    }
+}
```
fieldOfView: It [c|w]ould be better to include the pageTitle (and pageDescription) that is in each MachineAction here, much like how PreferencePage.qml does it, to enforce some uniformity.

PR #820, opened 2016-06-23
Add OctoPrint upload as an output device plugin

markwal:
This isn't ready to be merged as I still have some testing to do.  In particular, I need to figure out whether the hang at quit is my fault or also happens in master without my changes.  But I'd appreciate it if you'd take a look and give feedback.

### DIFF
--- a/plugins/OctoPrintOutputDevice/OctoPrintOutputDevice.py
+++ b/plugins/OctoPrintOutputDevice/OctoPrintOutputDevice.py
@@ -1,0 +1,232 @@
+# OctoPrintOutputDevice
+
+import os.path
+
+from PyQt5 import QtNetwork
+from PyQt5.QtCore import QTemporaryFile, QUrl
+
+from UM.Application import Application
+from UM.Logger import Logger
+from UM.Message import Message
+from UM.Job import Job
+from UM.Mesh.WriteMeshJob import WriteMeshJob
+from UM.Mesh.MeshWriter import MeshWriter
+from UM.Scene.Iterator.BreadthFirstIterator import BreadthFirstIterator
+from UM.OutputDevice.OutputDevice import OutputDevice
+from UM.OutputDevice import OutputDeviceError
+
+from UM.i18n import i18nCatalog
+catalog = i18nCatalog("cura")
+
+
+from enum import Enum
+class OutputStage(Enum):
+    ready = 0
+    writing = 1
+    uploading = 2
+
+
+class OctoPrintOutputDevice(OutputDevice):
+    def __init__(self, name="OctoPrint", host="http://octopi.local", apiKey=""):
+        super().__init__(name)
+
+        self.setName(name)
+        description = catalog.i18nc("@action:button", "Save to {0} ({1})").format(name, host)
+        self.setShortDescription(description)
+        self.setDescription(description)
+
+        self._stage = OutputStage.ready
+        self._host = host
+        self._apiKey = apiKey
+
+        self._qnam = QtNetwork.QNetworkAccessManager()
+        self._qnam.authenticationRequired.connect(self._onAuthRequired)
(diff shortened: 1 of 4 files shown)


markwal:
It does seem as though the hang occurs on quit on a rebuild without my changes, so it appears unrelated.

nallath:
You might want to have a look at the PrinterOutputDevice. That way your plugin can also benefit from https://github.com/Ultimaker/Cura/tree/feature_print_monitoring

fieldOfView:
Awesome. We're currently swamped with work because of an upcoming deadline, but I'll try to have a look at this tomorrow or over the weekend.

markwal:
@nallath hey, the print monitoring is pretty nifty.  I'll looking at adding print monitoring capability to a future iteration.  For now though, upload only is the main functionality I was after.

markwal:
@fieldOfView I hope the deadline went ok.  I've done some more testing and cleaned up a few loose ends, so I'm finished with this set of work until/unless there is something you all see that I should address.

fieldOfView:
@markwal the deadline is still ongoing. In my spare time I have been looking to get bonjour discovery to work, and to eventually make your implementation work with @nallath's print monitor (which was just merged into master today). Rest assured that we've not forgotten about this pull, though it will take a little while longer before it (or its descendant) is pulled.

markwal:
@fieldOfView Ah, I misunderstood the timeline.  I rebased and tested on the new master with the print monitor in it (not using it, just making sure there are no new incompatibilities).  I also added an open in browser button to link to the OctoPrint built-in print monitor.

fieldOfView:
I have put some of my private time into an alternative approach, using the PrinterOutputDevice as @nallath suggested:
https://github.com/fieldofview/OctoPrintPlugin
This ties in nicely with the print monitor. It also uses bonjour/zeroconf to help in discovery of the OctoPrint instance on the network, though you still have to enter the API key by hand.

markwal:
@fieldOfView, @nallath Seems like we should combine both approaches and allow the user to upload for later printing, monitor using OctoPrint web UI (or other OctoPrint phone app, etc.) directly *or* use Cura to monitor.

markwal:
I used UM.Preferences, should I instead use UM.Settings to store the OctoPrint instance url's an api_keys?  I store one for each instance so that the user can have a farm of OctoPrint's with unique api_key's if they wish.

fieldOfView:
Yes, using UM.Settings to store values per instance makes sense.

I'm interested in getting the best of our approaches in master. Functionally I would like to extend my PrinterOutputDevice-based approach with the following:
* Ability to manually enter an IP or hostname for cases where bonjour discovery does not work
* A way to easily open OctoPrint in the browser
* A preference to not start printing immediately
Can you think of any more things you would need/want?

fieldOfView:
Unfortunately our Product Management has decided not to include an OctoPrint plugin in Cura.

@markwal: I think the best thing to do for now is to host your plugin as a separate plugin, much like I do with mine. Thanks for your contribution. On a personal note, I hope PM changes their mind (as they do from time to time).

nallath:
I'm kindof tempted to just make a unofficial build where it is included.

fieldOfView:
Message to our PM:
17qbfv

fieldOfView:
@markwal: Just so you know I have created a repository from your plugin, containing just the files and commits for the plugin:
https://github.com/fieldOfView/OctoPrintUpload

markwal:
To the imaginary PM who doesn't have a github account:

Seems like you should go for this.  The downside is miminal: It's only appearance in a fresh install is in the Extensions menu so unlikely to confuse the user nor distract someone from buying an OctoPrint.  And the upside from the OctoPrint community who will be the only ones to find it on that menu will be considerable goodwill.

PR #821, opened 2016-06-23
[CURA-1615] GCodeProfileReader/-Export: Rework

thopiekar:
The exported data is now the same as you can find in .curaprofile files, but when deserializing the data back into a ContainerInstance and returning the new container like done in ```CuraProfileReader``` the imported profile still does not appear in the "settings window -> profiles".

I even get this error when "reimporting" the same container again:
```[35m2016-06-23 19:38:18,136[0m - [33mWARNING[0m - [37m.../UM/Settings/ContainerRegistry.py (addContainer [238]): Container of type <class 'UM.Settings.InstanceContainer.InstanceContainer'> and id BH2_SimpleCube_10mm already added[0m```

Would be great if someone could give me hints on that.
Thanks!

### DIFF
--- a/plugins/GCodeWriter/GCodeWriter.py
+++ b/plugins/GCodeWriter/GCodeWriter.py
@@ -23,7 +23,7 @@ class GCodeWriter(MeshWriter):
     #   It can only read settings with the same version as the version it was
     #   written with. If the file format is changed in a way that breaks reverse
     #   compatibility, increment this version number!
-    version = 1
+    version = 2
 
     ##  Dictionary that defines how characters are escaped when embedded in
     #   g-code.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Ghostkeeper** [comment]
When exporting, you are saving the same ID as the container (which is a good choice, I think). When importing it again, you are giving it the same ID as in the file again. Container IDs must be unique for each type.

There is a function in ContainerRegistry that thinks up a unique name for you: https://github.com/Ultimaker/Uranium/blob/master/UM/Settings/ContainerRegistry.py#L324
It adds the "#2" after the name if the name already exists. I think that would be a good solution.

**Comment by thopiekar** [comment]
Ready for testing :ok_hand:

File: plugins/GCodeWriter/GCodeWriter.py [hunk 1]
```diff
@@ -68,23 +68,21 @@ def _serialiseSettings(self, settings):
...
         prefix_length = len(prefix)
 
-        all_settings = InstanceContainer("G-code-imported-profile") #Create a new 'profile' with ALL settings so that the slice can be precisely reproduced.
-        all_settings.setDefinition(settings.getBottom())
-        for key in settings.getAllKeys():
-            all_settings.setProperty(key, "value", settings.getProperty(key, "value")) #Just copy everything over to the setting instance.
-        serialised = all_settings.serialize()
+        global_stack = Application.getInstance().getGlobalContainerStack()
+        container_with_profile = global_stack.findContainer({"type": "quality"})
+        serialized = container_with_profile.serialize()
```
fieldOfView: This only includes the "quality" profile settings and omits the "user" profile settings. So any settings the user has made that are not stored in the quality profile will not be stored with the gcode profile. We should have the settings from both "quality" and "user" combined into one.

fieldOfView:
I merged this so it stands a better chance at being tested. See the above comment about writing the "user" settings.

plugins/GCodeWriter/GCodeWriter.py
thopiekar: I know. Just followed Arjen's suggestion to save ’saved’ profiles first.
We can add an additional ticket for this on JIRA, so they (Arjen, Tim and
Jasper)  can decided when we want this fully supported.

fieldOfView: In that case, consider this part reviewed and tested.
Has the CURA-1615 ticket been updated with a link to the new issue?

thopiekar: No, the new issue needs to be added. Didn't opened it so far.
Am 05.07.2016 22:20 schrieb "Aldo Hoeben" <notifications@github.com>:

PR #822, opened 2016-06-24
Print monitor

fieldOfView:
CURA-1036 and CURA-338


cura/CuraApplication.py
```diff
@@ -235,6 +239,9 @@ def __init__(self):
 
             self._recent_files.append(QUrl.fromLocalFile(f))
 
+    def _onEngineCreated(self):
```
awhiemstra: Uh, why not do this in registerObjects? That gets a valid engine passed to it that you can add the image provider to.

cura/PrinterOutputDevice.py
```diff
@@ -57,6 +61,39 @@ def requestWrite(self, node, file_name = None, filter_by_machine = False):
     # it also sends it's own device_id (for convenience sake)
     connectionStateChanged = pyqtSignal(str)
 
+    timeElapsedChanged = pyqtSignal()
+
+    timeTotalChanged = pyqtSignal()
+
+    jobStateChanged = pyqtSignal()
+
+    jobNameChanged = pyqtSignal()
```
awhiemstra: I personally prefer to put the signal next to the property it is a change notification of, since that requires less searching. But we have no hard requirement for it.

cura/PrinterOutputDevice.py
```diff
@@ -74,6 +111,30 @@ def setTargetBedTemperature(self, temperature):
...
+        return self._time_elapsed
+
+    ## Total time of the print
+    #  Note that timeTotal - timeElapsed should give time remaining.
+    @pyqtProperty(float, notify=timeTotalChanged)
+    def timeTotal(self):
+        return self._time_total
+
+    @pyqtSlot(float)
+    def setTimeTotal(self, new_total):
```
awhiemstra: Out of curiosity, why are these exposed to QML? Is there ever a reason to set this from QML?

resources/qml/MonitorButton.qml
```diff
@@ -0,0 +1,183 @@
...
+        text: catalog.i18nc("@label:", "Abort Print")
+        onClicked: { Cura.MachineManager.printerOutputDevices[0].setJobState("abort") }
+
+
+        style: ButtonStyle
+        {
+            background: Rectangle
+            {
+                border.width: UM.Theme.getSize("default_lining").width
+                border.color: !control.enabled ? UM.Theme.getColor("action_button_disabled_border") :
```
awhiemstra: I have been changing these in other places as well, please do not use so many nested ternary operators and instead just write them out as a proper if-then-else construction. This gets really hard to follow otherwise.

resources/qml/Sidebar.qml
```diff
@@ -33,6 +35,22 @@ Rectangle
         tooltip.hide();
     }
 
+    function strPadLeft(string, pad, length) {
```
awhiemstra: Quick! Let's remove this method and break all of nodejs!

(https://developers.slashdot.org/story/16/03/23/0652204/how-one-dev-broke-node-and-thousands-of-projects-in-11-lines-of-javascript for those that do not get it)

resources/qml/Sidebar.qml
```diff
@@ -33,6 +35,22 @@ Rectangle
...
+
+    function getPrettyTime(time)
+    {
+        var hours = Math.floor(time / 3600)
+        time -= hours * 3600
+        var minutes = Math.floor(time / 60);
+        time -= minutes * 60
+        var seconds = Math.floor(time);
+
+        var finalTime = strPadLeft(hours, "0", 2) + ':' + strPadLeft(minutes,'0',2)+ ':' + strPadLeft(seconds,'0',2);
```
awhiemstra: Inconsistent use of quotes here.

resources/qml/Sidebar.qml
```diff
@@ -191,23 +277,152 @@ Rectangle
         }
     }
 
-    Rectangle {
+    // Item that shows the print monitor properties
+    Column
```
awhiemstra: Can we please move this to its own file? Just dumping it in here is rather ugly. Additionally, we would be able to use loader to avoid constantly having this in memory.

cura/PrinterOutputDevice.py
nallath: eh. The setTimeTotal doesn't really need to be exposed. The property for getting time elapsed and time total are needed (as they are displayed)

resources/qml/Sidebar.qml
nallath: Because i didn't know about that one :$

PR #823, opened 2016-06-24
CuraApplication: Fixing message about unknown color

thopiekar:
Doesn't contribute to any CURA-* issue.

Just needs @fieldOfView 's review here. ```UM.Theme.colors.message_text``` was not existing, so I took ```UM.Theme.colors.text_hover```. I'm sure we want something else here.


fieldOfView:
WTF does that code do anyway? 
It apparently colors the percentage number in a progress message stack.
https://github.com/Ultimaker/Uranium/commit/6bf0763e61a6ec6a192bf706b5a9e7ff9d4fc434 and then https://github.com/ultimaker/Cura/commit/720577ff8eb43df2dded5d49f3cb50658f82fcfe

Seems awfully contrived to me.
Good catch, but rather than taking another color, I would much rather this be cleaned up.

thopiekar:
UM.Qt.Bindings.Theme.Theme needs a QmlEngine to work. After that it takes the color "text_hover" from colors as QColor, where I ask for .name(), which actually does a conversation from ARGB to hex. color.

thopiekar:
Yes, it is pretty ugly, but does it's job :sweat_smile:

fieldOfView:
Thanks for the nod, @thopiekar!

thopiekar:
No problem. Just closed that after I saw your great set of commits :wink:

PR #824, opened 2016-06-24
[CURA-1771 + CURA-1445] SliceInfo: Make exception non-fatal again

thopiekar:
Related to CURA-1445 and CURA-1771

Needs: https://github.com/Ultimaker/Uranium/pull/147


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -108,4 +108,4 @@ def _onWriteStarted(self, output_device):
             Logger.log("i", "Sent anonymous slice info to %s", self.info_url)
             f.close()
         except Exception as e:
-            Logger.logException("e", e)
+            Logger.log("e", "Exception occured: %s" %(e))
```
fieldOfView: Suggestion: ```Exception occured while sending slice data: %s```

thopiekar: The output in the console will say it comes from SliceInfo.py, so I think it is enough.

awhiemstra:
logException does not make an exception fatal, instead it just makes sure the entire backtrace is printed, so is generally preferred when doing exception handling like this. If the problem is too much debug output, you might want to consider introducing some more specific exception handling for things like timeouts.

thopiekar:
Ok, I should better take a closer look at the original log.

thopiekar:
Found the reason in Uranium.
It tries to format a line which was taken from the sources
```
[35m2016-06-27 11:25:43,058[0m - [31m[47mCRITICAL[0m - [37m.../cura/CrashHandler.py (show [24]):   File "/usr/lib/python3.5/urllib/request.py", line 1313, in unknown_open[0m
  File "/usr/lib/python3.5/urllib/request.py", line 1313, in unknown_open ()
    raise URLError('unknown url type: %s' % type) ()
Error in sys.excepthook:
```

thopiekar:
Depends on: https://github.com/Ultimaker/Uranium/pull/147

thopiekar:
Now also depends on https://github.com/Ultimaker/Uranium/pull/152

PR #825, opened 2016-06-26
support prints up to 15cm width with GRR NEO Printer

TZanke:
NEO Firmware moves head 1cm to the right after Y endstop is pressed. This shrinks build volume to 14cm width. With this extra commands, the head moves back to the endstop. Now 15cm width builds are possible.

Additional Commands:
G91
G1 Y-10.0000 F9000
G92 Y0
G90


Ghostkeeper:
I believe you. Thanks for the contribution!

PR #826, opened 2016-06-26
Fixed the default nozel size for German RepRap Neo

simonCor:
Changed the default nozel size of GRR neo to 1.75mm. Was undefined before and defaulted to the wrong one.


Ghostkeeper:
Perhaps you could remove the `support_enable` defaulting to true. In our experience not printing support the first time clicks better with novice users, and I can't imagine that being something specific to GRR Neo.

I think you meant to fix the filament diameter and the title of this pull request is just a mistake, but if you did mean to change the nozzle size, take a look at the `machine_nozzle_size` setting!

ChrisTerBeke:
@simonCor Any plans to make the suggested changes?

ChrisTerBeke:
Closing because requested changes were never implemented and probably not important anymore.

PR #828, opened 2016-06-27
Better call it "Print via USB"

thopiekar:
I think this sounds better, because I understand "USB" as an technology. So you wouldn't say "Print with USB", but would fit if you say "Print with USB cable". So "Print via USB" should be more correct here.


Ghostkeeper:
I agree. We have an English copywriter though. I'll ask her.

thopiekar:
✌

PR #829, opened 2016-06-27
[CURA-1284]: Adding experimental icon

thopiekar:
Only a experimental icon is missing in resources/themes/cura/icons.

_NOTE_: The icons only appear after creating a new printer. At the moment I don't have an explanation for this.

Depends on: https://github.com/Ultimaker/Uranium/pull/148


fieldOfView:
Aww, you made this **way** more complicated than it needed to be. All that was needed for this issue is a new icon for the "experimental" _category_, to set that category apart from the "special modes" category. No need to add an icon to each setting...

thopiekar:
And I understood the task like "adding an icon to every entry to make, so we can move all entries to regular categories in the future". I didn't thought about replacing the "unknown" icon for the "experimental" category :disappointed:

thopiekar:
And yes, changing only the icon for the category is a very easy task.

PR #838, opened 2016-07-01
BQ Hephestos2 - Preheating temperature fix

thopiekar:
When starting a print with the "custom" GCode by BQ, the printer resets
the nozzle temperature to 210°C when printing via SD card (no matter
what you set on Cura) and only preheats the nozzle to 180°C when
printing via USB.
So currently the printer begins to print via USB at 180°C and reaches
the correct temperature eg. while printing the brim.

### DIFF
--- a/resources/definitions/bq_hephestos_2.def.json
+++ b/resources/definitions/bq_hephestos_2.def.json
@@ -14,7 +14,7 @@
     },
 
     "overrides": {
-        "machine_start_gcode": { "default_value": "; -- START GCODE --\nM800        ; Custom GCODE to fire start print procedure\n; -- end of START GCODE --" },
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I should've asked this before merging, but are you sure that print_temperature gets interpreted by BQ?

thopiekar:
And I should test my commit better. My change on the JSON was ignored by ```~/.cache/cura/definitions``` (are we probably caching to much here?) and yes, this is not correct:
```; -- START GCODE --
M800        ; Custom GCODE to fire start print procedure
M109 S{print_temperature} ;Makes sure the temperature is correct before printing
```; -- end of START GCODE --
Found similar lines in other definitions so I guessed it gets replaced. Is there any way to get the temperature here?

Thanks!

Ghostkeeper:
I see it too, in the RigidBot definition. I think if that still works since the settings rework, it would use the key `material_print_temperature`. Maybe that is why it wasn't filling it in for you.

thopiekar:
Thanks! Will give it a try :+1:

PR #846, opened 2016-07-04
Material Editing Support

awhiemstra:
Adds support for duplicating materials and editing them.

Depends on https://github.com/Ultimaker/Uranium/pull/150


awhiemstra:
Hang on, just realised I forgot to update the printer/profiles pages

cura/Settings/ContainerManager.py
```diff
@@ -0,0 +1,63 @@
...
+# Cura is released under the terms of the AGPLv3 or higher.
+
+from PyQt5.QtCore import QObject, pyqtSlot, pyqtProperty, pyqtSignal
+
+import UM.Settings
+
+from UM.i18n import i18nCatalog
+catalog = i18nCatalog("cura")
+
+class ContainerManager(QObject):
```
Ghostkeeper: You need to document this class.

cura/Settings/ContainerManager.py
```diff
@@ -0,0 +1,63 @@
...
+    @pyqtSlot(str, str)
+    def renameContainer(self, container_id, new_name):
+        pass
+
+    @pyqtSlot(str)
+    def removeContainer(self, container_id):
+        pass
+
+    @pyqtSlot(str, str, str)
+    def setContainerMetaDataEntry(self, container_id, entry_name, entry_value):
```
Ghostkeeper: In other places we've held the convention (originating from @awhiemstra too) that the Data in metadata is not capitalised; i.e. it was not considered a new word.

cura/Settings/MaterialSettingsVisibilityHandler.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__(parent = parent, *args, **kwargs)
+
+        material_settings = set([
+            "material_print_temperature",
+            "material_bed_temperature",
+            "material_standby_temperature",
+            "cool_fan_speed",
+            "retraction_amount",
+            "retraction_speed",
+        ])
```
Ghostkeeper: This is sort of double data. We have a different set of visible settings in CuraApplications, and as I now understand it this will make these six additional settings visible at every launch (I may be wrong on this, maybe it's every time you start the material manager?).

If you really want a separate set of visible settings in the material manager, then I suggest that you at least make a setting for it and define that setting here. The setting doesn't necessarily need to be modifiable from the interface.

awhiemstra: This is a visibility handler. It specifies which settings should be visible through SettingDefinitionsModel. In this case, the list of settings supported by the XML profiles is a fixed, limited list so there is no reason to use a preference. If anything, the dictionary in XmlMaterialProfile should be re-used for this, but that would create a tricky dependency on a plugin.

The fact that CuraApplication speficies a certain default set of visible settings is rather unrelated to this. There is no duplication of data with the list of visible settings there since they are used for different purposes.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -3,32 +3,196 @@
...
+        registry = UM.Settings.ContainerRegistry.getInstance()
+
+        all_containers = registry.findInstanceContainers(GUID = self.getMetaDataEntry("GUID"))
+        most_generic = all_containers[0]
+        for container in all_containers:
+            # Find the "most generic" version of this material
+            # This is a bit of a nasty implementation because of the naive assumption that anything with a shorter
+            # id is more generic. It holds for the current implementation though.
+            if len(most_generic.id) > len(container.id):
+                most_generic = container
```
Ghostkeeper: After thinking a bit more about it, even with our current implementation doesn't quite hold. If a material specifies a profile for `ultimaker2_plus_0.4` and one for `ultimaker2_extended_plus`, then the profile for `ultimaker2_plus_0.4` will get chosen over `ultimaker2_extended_plus`.

I'd still rather have you checking with multiple filter steps, like we discussed. It will be more code, but it could prevent some hard-to-debug bugs in the future.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -3,32 +3,196 @@
...
+
+        for container in all_containers:
+            definition_id = container.getDefinition().id
+            if definition_id == "fdmprinter":
+                continue
+
+            if definition_id not in machine_container_map:
+                machine_container_map[definition_id] = container
+            elif len(container.id) < len(machine_container_map[definition_id].id):
+                machine_container_map[definition_id] = container
```
Ghostkeeper: There you go again!

resources/qml/Preferences/MaterialView.qml
```diff
@@ -0,0 +1,248 @@
...
+                    onEditingFinished: Cura.ContainerManager.setContainerMetaDataEntry(base.containerId, "material", text)
+                }
+
+                Label { width: base.firstColumnWidth; height: parent.rowHeight; verticalAlignment: Qt.AlignVCenter; text: catalog.i18nc("@label", "Color") }
+
+                Row
+                {
+                    width: base.secondColumnWidth;
+                    height:  parent.rowHeight;
+                    spacing: UM.Theme.getSize("default_margin").width/2
```
Ghostkeeper: Speaking of spacing, the personal space of that little `/` is being invaded.

cura/Settings/ContainerManager.py
```diff
@@ -29,21 +29,115 @@ def __init__(self, parent = None):
...
+        containers = self._registry.findContainers(None, id = container_id)
         if not containers:
-            return
+            return ""
+
+        container = containers[0]
+
+        new_container = None
+        new_name = self._registry.uniqueName(container.getName())
+        if hasattr(container, "duplicate"):
```
nallath: What is the "duplicate" attribute of a container? I'd like to see a line of documentation explaining why you are doing this.

cura/Settings/ContainerManager.py
```diff
@@ -29,21 +29,115 @@ def __init__(self, parent = None):
...
-        pass
+    ##  Change the name of a specified container to a new name.
+    #
+    #   \param container_id \type{str} The ID of the container to change the name of.
+    #   \param new_id \type{str} The new ID of the container.
+    #   \param new_name \type{str} The new name of the specified container.
+    #
+    #   \return True if successful, False if not.
+    @pyqtSlot(str, str, str, result = bool)
+    def renameContainer(self, container_id, new_id, new_name):
```
nallath: Name and ID are used rather intechangably in other parts of the code right? So there is no real need to request both of them.

cura/Settings/ContainerManager.py
```diff
@@ -29,21 +29,115 @@ def __init__(self, parent = None):
...
+    #
+    #   \param merge_into_id \type{str} The ID of the container to merge into.
+    #   \param merge_id \type{str} The ID of the container to merge.
+    #
+    #   \return True if successfully merged, False if not.
+    @pyqtSlot(str, result = bool)
+    def mergeContainers(self, merge_into_id, merge_id):
+        containers = self._registry.findContainers(None, id = container_id)
+        if not containers:
+            return False
```
nallath: Bit of logging for the failure cases would be nice.

cura/Settings/ContainerManager.py
awhiemstra: It is a function that only exists for InstanceContainer at the moment. I will add a bit of explanation.

cura/Settings/ContainerManager.py
awhiemstra: None except that means that this bit of API will always assume that, which is in my opinion a bad assumption to make.

Ghostkeeper: Name and ID can't be forced to be equal to each other. Otherwise we'll get a lot of name clashes if we try to make a Normal Quality profile for ABS 0.4mm and one for ABS 0.8mm and one for PLA 0.4mm, etc...

ID needs to be unique, and name is user-visible. That's what I held so far.

fieldOfView: For user-generated profiles names and ids are kept "similar", and names are made to be unique.
Read-only profiles can have a name that is different from their id.

This was done so the user can identify their user-generated profiles by the filename.

PR #848, opened 2016-07-05
BQ Hephestos2: Heat up nozzle while leveling

thopiekar:
Saves time and preheats the nozzle to the correct temperature (set by Cura).
Btw. the BQ support explained me the reason why they are assuming 210°C needs to set here. The reason is that the printer is just meant to be used with PLA, so they are only expecting to use exactly 210°C. Just explained them why this is incorrect when think about the possibility that Cura can set the temperature on it's own.


thopiekar:
Merci

PR #857, opened 2016-07-08
[CURA-1445 + CURA-1615] Changes on SliceInfo and GCodeWriter

thopiekar:
I think this is the way YouMagine expects the data to be sent.

### DIFF
--- a/plugins/GCodeWriter/GCodeWriter.py
+++ b/plugins/GCodeWriter/GCodeWriter.py
@@ -5,6 +5,7 @@ from UM.Mesh.MeshWriter import MeshWriter
 from UM.Logger import Logger
 from UM.Application import Application
 from UM.Settings.InstanceContainer import InstanceContainer #To create a complete setting profile to store in the g-code.
+from UM.Settings.ContainerRegistry import ContainerRegistry
 import re #For escaping characters in the settings.
 
 ##  Writes g-code to a file.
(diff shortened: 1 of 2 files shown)


thopiekar:
Depends on https://github.com/Ultimaker/Uranium/pull/152

thopiekar:
The changes on the GCodeWriter are more related to CURA-1615..

plugins/GCodeWriter/GCodeWriter.py
```diff
@@ -69,8 +70,13 @@ def _serialiseSettings(self, settings):
         prefix_length = len(prefix)
 
         global_stack = Application.getInstance().getGlobalContainerStack()
-        container_with_profile = global_stack.findContainer({"type": "quality"})
-        serialized = container_with_profile.serialize()
+        # Collecting all data from the top of the stack
+        temporarySettings= InstanceContainer("")
+        temporarySettings.setDefinition(ContainerRegistry.getInstance().findDefinitionContainers(id="fdmprinter")[0])
+        for key in global_stack.getAllKeys():
```
nallath: This does not get all settings; This only gets settings for single extrusion machines.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -88,7 +88,7 @@ def _onWriteStarted(self, output_device):
                 "settings": global_container_stack.serialize(), # global_container with references on used containers
                 "version": Application.getInstance().getVersion(),
                 "modelhash": "None",
-                "printtime": print_information.currentPrintTime.getDisplayString(),
+                "printtime": print_information.currentPrintTime.getDurationInISOFormat(),
```
nallath: I might be wrong, but i think they just want print time in seconds.

thopiekar: I think we should wait until we get feedback from YouMagine to be sure :wink:

plugins/GCodeWriter/GCodeWriter.py
thopiekar: Can you give a small hint how to get per extruder settings?

Ghostkeeper: `global_stack`.getAllKeys() gets all keys in FDMPrinter. You'd have to call this function on an extruder stack to get the keys in FDMExtruder. I believe those stacks are held in Cura/Settings/ExtruderManager.py.

PR #858, opened 2016-07-08
[CURA-1831] SSL workaround for OSX

thopiekar:
Basically uses a workaround I found here:
http://stackoverflow.com/a/28052583

Tested on a MacBook Air at the HQ.

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -47,82 +47,78 @@ class SliceInfo(Extension):
         Preferences.getInstance().setValue("info/asked_send_slice_info", True)
 
     def _onWriteStarted(self, output_device):
-        try:
-            if not Preferences.getInstance().getValue("info/send_slice_info"):
-                Logger.log("d", "'info/send_slice_info' is turned off.")
-                return # Do nothing, user does not want to send data
+        if not Preferences.getInstance().getValue("info/send_slice_info"):
+            Logger.log("d", "'info/send_slice_info' is turned off.")
+            return # Do nothing, user does not want to send data
 
-            global_container_stack = Application.getInstance().getGlobalContainerStack()
+        global_container_stack = Application.getInstance().getGlobalContainerStack()
 
-            # Get total material used (in mm^3)
-            print_information = Application.getInstance().getPrintInformation()
-            material_radius = 0.5 * global_container_stack.getProperty("material_diameter", "value")
+        # Get total material used (in mm^3)
+        print_information = Application.getInstance().getPrintInformation()
+        material_radius = 0.5 * global_container_stack.getProperty("material_diameter", "value")
 
-            # TODO: Send material per extruder instead of mashing it on a pile
-            material_used = math.pi * material_radius * material_radius * sum(print_information.materialAmounts) #Volume of all materials used
+        # TODO: Send material per extruder instead of mashing it on a pile
+        material_used = math.pi * material_radius * material_radius * sum(print_information.materialAmounts) #Volume of all materials used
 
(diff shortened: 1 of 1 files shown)


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -47,82 +47,78 @@ def messageActionTriggered(self, message_id, action_id):
...
-                    ssl_context = ssl._create_unverified_context()
-                    server_request = urllib.request.urlopen(self.info_url, data = binary_data, timeout = 1, context = ssl_context)
-                else:
-                    server_request = urllib.request.urlopen(self.info_url, data = binary_data, timeout = 1)                Logger.log("i", "Sent anonymous slice info to %s", self.info_url)
-                server_request.close()
-            except Exception as e:
-                Logger.logException("e", "An exception occurred while trying to send slice information")
-        except:
-            # We really can't afford to have a mistake here, as this would break the sending of g-code to a device
-            # (Either saving or directly to a printer). The functionality of the slice data is not *that* important.
```
fieldOfView: Are we no longer concerned about this?
As far as I understand there this PR only fixes one possible issue, but are we sure that is all that can ever go wrong?

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -112,9 +114,12 @@ def _onWriteStarted(self, output_device):
...
             # Submit data
             try:
-                f = urllib.request.urlopen(self.info_url, data = binary_data, timeout = 1)
-                Logger.log("i", "Sent anonymous slice info to %s", self.info_url)
-                f.close()
+                if Platform.isOSX():
+                    ssl_context = ssl._create_unverified_context()
+                    server_request = urllib.request.urlopen(self.info_url, data = binary_data, timeout = 1, context = ssl_context)
+                else:
+                    server_request = urllib.request.urlopen(self.info_url, data = binary_data, timeout = 1)                Logger.log("i", "Sent anonymous slice info to %s", self.info_url)
```
fieldOfView: Euhm, this line                                  is wonky.

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: And as far as I understand the issue on JIRA it always happens on OSX...
I applied it here and successfully sends all data.

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Yes, I should better just append the keyword, when being on OSX.

fieldOfView: That's not the problem; scroll to the right and you'll find a Logger statement that should be on the next line.

plugins/SliceInfoPlugin/SliceInfo.py
Ghostkeeper: What fieldOfView means is that there might be other exceptions occurring in the future and we basically NEVER want to block the slicing when anything fails in the sending of slicing data.

I am of the opinion that if there is an exception we haven't caught yet, that is a coding error on our side. We should explicitly code this piece to be resilient to internet interruptions and such and test this well, rather than hide our coding errors behind a try: catch block.

nallath: I agree with that, but the point is that we can never test everything. The functionality of this bit of code isn't that important but it pretty much breaks everything if we don't do a general try catch and something does happen. 

I'd prefer to leave the big try catch block in.

thopiekar: I hate signals :angry:

thopiekar:
Ok, we can close this again..

PR #867, opened 2016-07-11
CURA-1615: Storing all settings into G-Code

thopiekar:
Replaces: https://github.com/Ultimaker/Cura/pull/857


sedwards2009:
This pull request is really the domain of CURA-1727 "GCodeProfileReader: Broken and needs update".

thopiekar:
Oops, I used 1615 here, because this number goes to all plugin fixes (all
the small fixes). Is renaming the PR enough? 🙄

sedwards2009:
I think this code is going to have to be different anyway. How about we just park this PR here for a while as I sort out CURA-1727. (I still might have to 'borrow' some of the code from here anyway. ;-) )

PR #869, opened 2016-07-11
CURA-1787 Added message when switching to one_at_a_time.

jackha:
CURA-1787

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -165,6 +165,13 @@ class BuildVolume(SceneNode):
     def getBoundingBox(self):
         return self._volume_aabb
 
+    def _buildVolumeMessage(self):
+        Message(catalog.i18nc(
+            "@info:status",
+            "The build volume height has been reduced due to the value of the"
+            " \"Print Sequence\" setting to prevent the gantry from colliding"
+            " with printed objects."), lifetime=10).show()
+
     def _onGlobalContainerStackChanged(self):
         if self._active_container_stack:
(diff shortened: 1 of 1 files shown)


cura/BuildVolume.py
```diff
@@ -189,6 +192,11 @@ def _onSettingPropertyChanged(self, setting_key, property_name):
         if setting_key == "print_sequence":
             if Application.getInstance().getGlobalContainerStack().getProperty("print_sequence", "value") == "one_at_a_time":
                 self._height = self._active_container_stack.getProperty("gantry_height", "value")
+                Message(catalog.i18nc(
+                    "@info:status",
+                    "The build volume height has been reduced due to the value of the"
+                    " \"Print Sequence\" setting to prevent the gantry from colliding"
+                    " with printed objects."), lifetime=10).show()
```
nallath: Why not use the default lifetime?

fieldOfView:
I would show the same message _onGlobalContainerStackChanged (if appropriate)

cura/BuildVolume.py
jackha: The default is 30 secs, this feels better.

nallath: The 30 seconds was introduced because people mentioned that the 10 seconds was too short, because they often missed important messages. 
I don't really mind which of the two we go for, but it should be a concious decision. 
@fieldOfView What's your opinion on this?

jackha:
I was wondering if it made sense, but I think it is indeed nice to have this message on start of Cura as well.

PR #879, opened 2016-07-21
Remove the Global tab

fieldOfView:
This removes the concept of a "Global" tab.

Settings that are not settable per extruder are now shown on the extruder tabs, and their values can be saved into per-extruder profiles. These settings have a "link" icon that informs the user that editing a value here will change it for all extruders. 

The global stack loads the same profiles (variant, material, quality) as extruder 0. A new property "global_inherits_stack" is introduced in fdmprinter.def.json that points to an extruderstack. This property would normally be a simple "function" pointing to the value of eg support_extruder_nr.  Values from this extruder are copied into the global stack user profile ("current settings"). This way global settings can mirror the settings set per extruder. Settings that are not settable per extruder are automatically copied from the last changed extruder stack.

Settings in the advanced setting view have a new context-menu "Copy value to all extruders" that copies the value to all extruder stacks (including the global stack).

For single extrusion printers there is no change.

Technical considerations:
* For the backend, nothing changes. There is still a global stack, but the global stack is not shown to the user, and it is automatically populated.
* No changes to Uranium.

Testing considerations:
For testing it can be beneficial to show the global tab. This can by editing SidebarHeader.qml (see the change in https://github.com/Ultimaker/Cura/commit/dd78561e25cfb33fcbb2e196c96bedd84c8cc62e)
It may be necessary to clear your profile, or at least to add a new Jedi.

CURA-1758


cura/Settings/MachineManager.py
```diff
@@ -220,15 +220,16 @@ def _onGlobalPropertyChanged(self, key, property_name):
                                 # Remove from the value from the other stacks as well, unless the
                                 # top value from the other stacklevels is different than the new value
                                 for container in extruder_stack.getContainers():
-                                    if container == extruder_stack.getTop():
-                                        continue
+                                    # Commented out this line for a bit. Don't know why it's here, but it is causing
+                                    # issues with reset button.
```
fieldOfView: This was just to skip the top container in this check. The value is removed from the top profile if the first encountered non-top container contains the same value.

nallath: Those two lines broke the reset button being handled properly, but removing it also created a recursion issue. I've since then fixed it.

PR #881, opened 2016-07-22
Stripe objects that are outside the build volume

fieldOfView:
This PR adds a two-color stripe-pattern to objects that are not entirely inside the buildvolume. This way they can not be confused with objects that are rendered with a (future) grey materialcolor.

There is no CURA-issue for this, hence the PR.

### DIFF
--- a/plugins/SolidView/SolidView.py
+++ b/plugins/SolidView/SolidView.py
@@ -34,9 +34,10 @@ class SolidView(View):
             self._enabled_shader = OpenGL.getInstance().createShaderProgram(Resources.getPath(Resources.Shaders, "overhang.shader"))
 
         if not self._disabled_shader:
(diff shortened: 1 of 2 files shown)


nallath:
Kinda trippy. Should I merge this into master? Seems ta work allright, but not sure if we already want this for next release.

awhiemstra:
I am not really happy with this change. In my opinion, objects outside the build plate look really out of place now.

fieldOfView:
The model looking "out of place" is the whole idea, because these objects _are_ out of place. We need to show that there is something wrong with the object(-placement), in a way that does not conflict with any (future) material-color.

The stripe pattern is in screen-space on purpose; this way it is a stripe-pattern regardless of the geometry and regardless of the view-direction; if the stripe-pattern were applied to eg a sphere in model- or world-space, depending on the view-direction it would show as stripes, curves or even concentric circles.

Having said that, I am open for suggestions.

PR #889, opened 2016-07-25
[CURA-1923] SliceInfo: Sending data threaded

thopiekar:
The idea is to make the sending process, so it won't block the UI until the data is sent or the timeout has reached.
This commit does the following:
* Increasing the timeout to 5s
-> As we are sending our data threaded now this won't hurt. At least for clients in slow networks.
* Moving section for "sending data to info_url" into a seperate thread.
* Starting the thread instead and collecting all threads in a list
* Removing all threads from the list, if they have already finished their work.
* When the SliceInfo Extension gets ```__del__```'d, it will wait for the threads until they finish.

Contributes to CURA-1923

PS: Also adds a Logger.logException to get debug info, when something goes wrong.


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -24,6 +21,33 @@
 
 catalog = i18nCatalog("cura")
 
+class SliceInfoThread(Thread):
```
awhiemstra: Instead of using Thread directly, why not create a Job subclass for this? Those are specifically designed for this.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -24,6 +21,33 @@
...
+    url = None
+
+    def __init__(self, url, data):
+        Thread.__init__(self)
+        self.url = url
+        self.data = data
+
+    def run(self):
+        if not self.url or not self.data:
+            Logger.log("e", "URL or DATA for sending slice info was not set!")
```
awhiemstra: Shouldn't this return? Otherwise you are trying to send nonexistent data.

thopiekar: Yes, of course. Fixed that :+1:

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Done :ok_hand:

thopiekar:
So in summary, this kind of fixing the blocking tasks is fine?

awhiemstra:
Yes, especially when it involves IO we generally move things to a job so that we do not block the main thread.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -24,6 +22,34 @@
 
 catalog = i18nCatalog("cura")
 
+class SliceInfoThread(Job):
+    data = None
+    url = None
+
+    def __init__(self, url, data):
+        Job.__init__(self)
```
awhiemstra: For python 3, you should use super().__init__()

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -112,19 +152,11 @@ def _onWriteStarted(self, output_device):
...
-                kwoptions["context"] = ssl._create_unverified_context()
-            try:
-                f = urllib.request.urlopen(self.info_url, **kwoptions)
-                Logger.log("i", "Sent anonymous slice info to %s", self.info_url)
-                f.close()
-            except Exception as e:
-                Logger.logException("e", "An exception occurred while trying to send slice information")
+            # Sending slice info non-blocking
+            reportJob = SliceInfoThread(self.info_url, binary_data)
+            self.runningJobs.append(reportJob)
```
awhiemstra: In general, jobs are meant to act in a fire-and-forget manner. JobQueue manages the jobs and their assignment to worker threads, after which they are run and then destroyed. There should be no real need to keep track of them unless you need the result of a job, in which case you should use its finished signal. So I do not think you need to do all this tracking here.

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: :ok_hand:

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Done :ok_hand:

sedwards2009:
Regarding whether it is a good idea to fix blocking issues like this, I would much prefer an improved architecture where one thread is responsible for all communication with the back-end and that it uses a message passing style IPC to communicate with the UI.

Why do I prefer this? Because I this ad hoc system of spawning multiple threads which share the back-end connection and is coordinated using signals is fragile and hard to get right. There are too many moving parts, too many interactions and too many assumptions about when which signal occur when. Putting one thread and component in charge with a clearly defined API would centralise things and make it much easier to coordinate.

awhiemstra:
I think this has little to do with how the backend communication is done, since this code does not really rely on that. I do not think an approach like that works for all the cases where we have potentially blocking IO.

nallath:
How would the slice info rely at all on the backend / engine? It simply waits for an onWriteStarted event and sends some info in the background.

thopiekar:
So basically this can be merged`? At least it fixes the problem for now.

I also find it strange that some code is waiting here until all signals are progressed. It makes sense when you emit signals which return a value, but it doesn't if there is nothing you should care about.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -24,6 +21,34 @@
...
+
+        if Platform.isOSX():
+            kwoptions["context"] = ssl._create_unverified_context()
+
+        try:
+            f = urllib.request.urlopen(self.url, **kwoptions)
+            Logger.log("i", "Sent anonymous slice info to %s", self.url)
+            f.close()
+        except Exception:
+            Logger.logException("e", "An exception occurred while trying to send slice information")
```
fieldOfView: Like with the update-checker, I would prefer to use Logger.log() instead of Logger.logException. There is not much use to dump the whole traceback to the log on every slice. Just a simple warning will do.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -112,19 +139,10 @@ def _onWriteStarted(self, output_device):
...
-            except Exception as e:
-                Logger.logException("e", "An exception occurred while trying to send slice information")
+            # Sending slice info non-blocking
+            reportJob = SliceInfoJob(self.info_url, binary_data)
+            reportJob.start()
         except:
             # We really can't afford to have a mistake here, as this would break the sending of g-code to a device
             # (Either saving or directly to a printer). The functionality of the slice data is not *that* important.
-            pass
+            Logger.logException("e", "Exception raised while sending slice info") # But we should be notified about these problems of course.
```
fieldOfView: Sam as above. Logger.logException is useful for when you need to find out why we reached this point. We know exactly how this point was reached. A logged warning which includes the exception-text would do.

thopiekar: Ok, I agree with that here. But ```pass``` here killed time here, because I thought my code above was absolutely correct.

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Hmm, then I would define better which Exceptions appear if you don't have internet.
For every other situation I would print the log like this.
Logger.logException was meant to be used for printing Exceptions. I guess what it needs here is something like Logger.logException(depth=X) to cut all the uninteresting part away.

thopiekar: @awhiemstra What do you think?

Ghostkeeper: I suggest catching URLError and logging that with logException, but catching HTTPError and logging that with just a log message that the file could not be fetched from the internet.

awhiemstra: I would say, try to be specific about which exceptions to catch, then log those less verbose. Those @Ghostkeeper mentions sound like the right direction. Anything else would be an unexpected exception and should bubble upwards to CrashHandler.

thopiekar: What about unexpected errors (neither URLError nor HTTPError)? I would catch them with logException, too.

awhiemstra: Quoting myself:

> Anything else would be an unexpected exception and should bubble upwards to CrashHandler.

Ghostkeeper: Ah, but should it, @awhiemstra? We've said before that we don't want this slice info to ever break the program regardless of any programming errors...

awhiemstra: In that case we should catch any exception, yes.

plugins/SliceInfoPlugin/SliceInfo.py
fieldOfView: I can hardly imaging ```pass``` is the command that takes time. It is the python equivalent of a "null operation".

thopiekar:
Could someone make a little review again? 
Thanks :wink:

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -24,6 +21,36 @@
...
+            kwoptions["context"] = ssl._create_unverified_context()
+
+        try:
+            f = urllib.request.urlopen(self.url, **kwoptions)
+            Logger.log("i", "Sent anonymous slice info to %s", self.url)
+            f.close()
+        except urllib.error.HTTPError as http_exception:
+            Logger.log("e", "An exception occurred while trying to send slice information: %s" %(repr(http_exception)))
+        except Exception: # Includes urllib.error.HTTPError, was discussed to be handled like any other exception 
+            Logger.logException("e", "An exception occurred while trying to send slice information")
```
fieldOfView: We don't have "rules" to decide when to use logException() vs just a log() that includes the exception text, but I would only use logException in cases where the context/traceback might be unknown (ie from very generic/often used classes such as SettingFunction, Validator, ContainerRegistry, which gets called from many different classes/cases). 
In this case it just add unnecessary log verbosity. But since it was already used here, I'll let it slide.

PR #893, opened 2016-07-26
[CURA-1223] 3MFReader: Doing selftest and fail on broken files

thopiekar:
This commit adds a selftest before the "result" gets returned.
It should break on this/these functions and popup a message about using
the online repair tool.
I compared the content of both files for a long time, googled about the
format, but wasn't able to find a proper fix.
More routines will likely need to be added here, but with those I have
it is working so far very well.

Sadly it is not possible to override the default message, so two
messages will appear. Additionally, the URL of the link is not
clickable/executable from the UX. Just acting like normal text.

Contributes to CURA-1223

### DIFF
--- a/plugins/3MFReader/ThreeMFReader.py
+++ b/plugins/3MFReader/ThreeMFReader.py
@@ -27,7 +27,7 @@ except ImportError:
 ##    Base implementation for reading 3MF files. Has no support for textures. Only loads meshes!
 class ThreeMFReader(MeshReader):
     def __init__(self):
-        super(ThreeMFReader, self).__init__()
+        super().__init__()
         self._supported_extensions = [".3mf"]
 
         self._namespaces = {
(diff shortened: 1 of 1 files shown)


plugins/3MFReader/ThreeMFReader.py
```diff
@@ -116,4 +120,11 @@ def read(self, file_name):
         except Exception as e:
             Logger.log("e", "exception occured in 3mf reader: %s", e)
 
+        try: # Selftest - There might be more functions that should fail 
+            result.getBoundingBox()
```
awhiemstra: Maybe you can expand this check to also see if the bounding box is not valid? Then you'd catch both the situation where we load an invalid file as well as an empty file.

thopiekar: Done. Hope ```.isValid()``` is enough here.

PR #896, opened 2016-07-29
Added option to only display top layers in layerview

nallath:
Got really fed up with not being able to look at a single layer. 

Put this in a pull request as we don't have a corresponding ticket for this (and this was therefore built in my own time)


fieldOfView:
Reviewed and tested. Looks good. Awesomeness all around.

nallath:
Yay :)

PR #898, opened 2016-07-30
Create shades of colors for each extruder

fieldOfView:
If two extruders have the same material it is impossible to see if multiple objects will print with the same extruder of with different extruders. This PR creates a shade of the material-color per extruder.

This PR relies on https://github.com/Ultimaker/Uranium/pull/155 for parsing and changing the rgb color strings used in the configuration files and qml.

### DIFF
--- a/cura/Settings/ExtrudersModel.py
+++ b/cura/Settings/ExtrudersModel.py
@@ -4,6 +4,7 @@
 from PyQt5.QtCore import Qt, pyqtSignal, pyqtProperty
 
 import UM.Qt.ListModel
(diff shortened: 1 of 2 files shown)


fieldOfView:
I just thought of a much better way to do this, which would also work with more than 2 extruders and can be extended to layerview more easily: The colorshade would apply objects printed with the not-active-extruder. Selecting an extruder in the sidebar would "dim" the objects printed with the other extruder(s), this highlighting the objects printed with the currently active extruder.

I'll prepare a new PR for that.

PR #901, opened 2016-08-01
Scanning for CuraEngine and stop if it was not found

thopiekar:
So far I ran 2 or 3 times into the problem that my engine was not up to
date. The bad thing here is that there is always an updated version from
my PPA on the PC. So first this commit gives the possibility to look for
CuraEngine in the $PATH directories. Before I had to copy it always
manually, eg. via "cp $(which CuraEngine) bin/CuraEngine" in my
workbench.

Second, people can get into the situation that CuraEngine is missing at
all. So before making Cura loop and try to use an executable that does
not exist, better raise an Exception here.
An additional info message tells about the location being used.

Does not contribute to any JIRA issue (I think), but makes my life
easier.

### DIFF
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -13,7 +13,6 @@ from UM.Resources import Resources
 from UM.Settings.Validator import ValidatorState #To find if a setting is in an error state. We can't slice then.
 from UM.Platform import Platform
 
-
 import cura.Settings
 
 from cura.OneAtATimeIterator import OneAtATimeIterator
@@ -32,7 +31,6 @@ import Arcus
 
 from UM.i18n import i18nCatalog
 catalog = i18nCatalog("cura")
-
 
 class CuraEngineBackend(Backend):
     ##  Starts the back-end plug-in.
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -37,7 +37,6 @@ class CuraEngineBackend(Backend):
     #
     #   This registers all the signal listeners and prepares for communication
     #   with the back-end in general.
-    
     def __init__(self):
         super().__init__()
 
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -54,7 +54,7 @@ class CuraEngineBackend(Backend):
         if Platform.isWindows():
             default_engine_location += ".exe"
         if Platform.isLinux() and not default_engine_location:
-            if not os.getenv('PATH'):
+            if not os.getenv("PATH"):
                 raise OSError("There is something wrong with your Linux installation.")
             for pathdir in os.getenv('PATH').split(os.pathsep):
                 execpath = os.path.join(pathdir, executable_name)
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -39,16 +39,35 @@ class CuraEngineBackend(Backend):
     #
     #   This registers all the signal listeners and prepares for communication
     #   with the back-end in general.
+    
+    executableName = "CuraEngine"
```
awhiemstra: Is this supposed to be public? Otherwise please mark it as protected/private.

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -39,16 +39,35 @@ class CuraEngineBackend(Backend):
...
-        default_engine_location = os.path.join(Application.getInstallPrefix(), "bin", "CuraEngine")
+        default_engine_location = None
+        if os.path.exists(os.path.join(Application.getInstallPrefix(), "bin", self.executableName)):
+            default_engine_location = os.path.join(Application.getInstallPrefix(), "bin", self.executableName)
         if hasattr(sys, "frozen"):
-            default_engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), "CuraEngine")
+            default_engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), self.executableName)
         if Platform.isWindows():
             default_engine_location += ".exe"
+        if Platform.isLinux() and not default_engine_location:
```
awhiemstra: Part of me thinks it is a shame to limit this to Linux. On the other hand, I do not think this happens often on the other two platforms.

thopiekar: Well, if that problem goes on someones nerves then she/he can improve the code here. I wouldn't care about that here at the moment.

plugins/CuraEngineBackend/CuraEngineBackend.py
thopiekar: Good question. I just wanted to use it here, because "CuraEngine" was mentioned multiple times. So I decided to define it once at one place.

Do you think it is needed to make it private here?

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -39,16 +39,35 @@ class CuraEngineBackend(Backend):
...
+            if not os.getenv('PATH'):
```
Ghostkeeper: We use double quotes for strings.

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -54,7 +54,7 @@ def __init__(self):
         if Platform.isWindows():
             default_engine_location += ".exe"
         if Platform.isLinux() and not default_engine_location:
-            if not os.getenv('PATH'):
+            if not os.getenv("PATH"):
                 raise OSError("There is something wrong with your Linux installation.")
             for pathdir in os.getenv('PATH').split(os.pathsep):
```
Ghostkeeper: Also here, please :P Sorry for the nitpick.

thopiekar: I hate that this ```'``` key is so easy to type on QWERTY keyboards. Always using ```'``` because it is comfortable ^^

Ghostkeeper: We require it to keep in line with the other programming languages we're using. I.e. C++.

thopiekar:
Click'n'merge? :wink:

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -39,16 +39,34 @@ class CuraEngineBackend(Backend):
     #
     #   This registers all the signal listeners and prepares for communication
     #   with the back-end in general.
+
```
awhiemstra: Should remove the stray enter here.

awhiemstra:
Feel free to merge after removing the stray enter.

thopiekar:
Thanks

PR #902, opened 2016-08-01
Show "ghosts" for all objects on the buildplate in layerview

fieldOfView:
The current implementation of layerview only shows a grey "ghost" of the currently selected object. If the user enters layerview without an object selected, she is presented with an empty buildplate until slicing is completed and layers are processed.

This PR shows a ghost for all objects instead of only for the selected object(s), this giving the user a decent view of where objects are located and where to click to select an object.

CURA-1601 is slated for a later sprint, but it seems such a shame to delay this now that we have made such great improvements to Layer View.

### DIFF
--- a/plugins/LayerView/LayerView.py
+++ b/plugins/LayerView/LayerView.py
@@ -33,7 +33,7 @@ class LayerView(View):
     def __init__(self):
         super().__init__()
         self._shader = None
-        self._selection_shader = None
+        self._ghost_shader = None
         self._num_layers = 0
         self._layer_percentage = 0  # what percentage of layers need to be shown (Slider gives value between 0 - 100)
         self._proxy = LayerViewProxy.LayerViewProxy()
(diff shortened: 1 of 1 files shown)


awhiemstra:
Yay awesome

awhiemstra:
One issue though: The overlay flag disables depth testing which results in some pretty interesting rendering artifacts:


screenshot_20160802_102550

Ghostkeeper:
Can I keep them?

fieldOfView:
I will look in to using render passes instead of using overlay.

awhiemstra:
Oops I seem to have accidentally merged this

nickthetait:
I don't follow what you mean by
> disables depth testing

Ghostkeeper:
It disables the [depth buffer](https://en.wikipedia.org/wiki/Z-buffering). The depth buffer is used by shaders in 3D rendering. It's basically a drawing of the scene, but instead of the scene colour it draws a shade based on the distance to the camera. After the depth buffer is drawn, the shader draws every triangle in the scene again but first tests if the depth towards a patch of surface is equal to the depth buffer. If the distance is equal to the distance in the depth buffer, that means that the patch of surface is not obscured by something else (i.e. the surface is in front).

Disabling the depth buffer just draws all surfaces. The result is that whichever surface gets drawn last ends up overwriting the colour of the other surfaces, and so ends up on top.

nickthetait:
gotcha :ok_hand: thanks for explaining

PR #904, opened 2016-08-02
Too small timeout of autoscale notification in 2.1.2

iakapulko:
Huge amount of users (and me too, unfortunately) miss popups of autoscale feature. Time, filament and money lost because of this. So, why not to adjust timeout of this popup or not make this option off by default?


Dim3nsioneer:
Autoscale is a very dangerous feature imho. Default on advanced mode should be 'off'.

Ghostkeeper:
I suspect you didn't mean to make this pull request pulling random commits into the 2.1 branch, so let's ignore those commits and treat this as an issue.

There's a setting in Preferences called "Scale large files" to toggle the auto-scaling. It's somewhat strange to let the default of that setting depend on whether the user is in advanced mode.

I somewhat like the feature myself. It's great for what I do with 3D printers: print cosmetic stuff. However I agree that it's better for 3 out of our 4 target audiences to turn it off by default (prototyping, tinkering and manufacturing). What do you think, @fieldOfView?

fieldOfView:
I agree not to have the autoscale feature default depend on the simple/advanced switch. Perhaps the feature could be implemented a bit smarter and only scale down objects that are way too large (eg twice or more larger than what would fit). That way there will be fewer occasions where the object is printer _just a little_ too small.

Note that the cousin-setting that scales extremely small objects up only scales by factors of 10 (this was meant for the scenario where objects are modeled with meter units instead of millimeter units). Perhaps the auto-scale-_down_  could do something similar? Would that solve anything?

Dim3nsioneer:
I agree with @fieldOfView that scales around 90-100% are the most dangerous as the user doesn't see it immediately from just the size. What about having a different color for an autoscaled object?

fieldOfView:
In 2.2, object colors will represent materials, so changing the color for (auto)scaled objects won't be an option. For that version we have already changed the time the autoscale message is shown from 10 seconds to 30 seconds.

LilBub:
I got hit with this recently and due to the color scheme of this program (very pale colors), it is easy to miss a grey window.  As others noted, since it is automatic as a default setting, it can be a big problem.  

Perhaps a larger, Windows-like pop up would have helped.  

But my preference would be behavior similar to Cura 15.04.x  

Many models will not fit the buildplate by default positions...yet WILL if the model is rotated.  No scaling needed.

fieldOfView:
I am closing this because it is not a valid pull request. We are still discussing what is the best way to handle this for 2.2 and beyond.

Dim3nsioneer:
Shall we open a new issue or do you have it on your internal list?

fieldOfView:
It is on our internal tracker as CURA-2050.

PR #905, opened 2016-08-03
Make camera animation feel snappier

Ghostkeeper:
I've tweaked the camera animation a bit.
The difference is that it no longer eases into the animation, making it react more immediate. This makes it feel snappier. The duration is made a little shorter to make the ease-out feel approximately the same.

Let me know what you think.

### DIFF
--- a/cura/CameraAnimation.py
+++ b/cura/CameraAnimation.py
@@ -11,8 +11,8 @@ class CameraAnimation(QVariantAnimation):
     def __init__(self, parent = None):
         super().__init__(parent)
         self._camera_tool = None
-        self.setDuration(500)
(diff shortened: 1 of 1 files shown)


fieldOfView:
The difference is subtle. I actually don't know what I prefer; most of the time the animation stutters because slicing is (re-)started at the same time as the camera animation.

ChrisTerBeke:
Just manually changed those 2 lines on my machine and I like the changes a little bit better than the previous settings.

Ghostkeeper:
Depends on the performance of Cura on your computer a bit, I guess, if the animation stutters when slicing is restarted. Though since then we've had a few tweaks to the way that the subprocess for slicing is started, so maybe it's not so bad now. I'll create a ticket for it.

PR #906, opened 2016-08-03
Show "ghosts" for all objects on the buildplate in layerview (#2)

fieldOfView:
The current implementation of layerview only shows a grey "ghost" of the currently selected object. If the user enters layerview without an object selected, she is presented with an empty buildplate until slicing is completed and layers are processed.

This PR shows a ghost for all objects instead of only for the selected object(s), this giving the user a decent view of where objects are located and where to click to select an object. It is a followup to https://github.com/Ultimaker/Cura/pull/902, and now uses a layer pass to render the layer data on top of the ghosts.

CURA-1601 is slated for a later sprint, but it seems such a shame to delay this now that we have made such great improvements to Layer View.

### DIFF
--- a/plugins/LayerView/LayerPass.py
+++ b/plugins/LayerView/LayerPass.py
@@ -1,30 +1,112 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
-from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
+from UM.Logger import Logger
+from UM.Application import Application
 from UM.Resources import Resources
+from UM.Preferences import Preferences
+from UM.Signal import Signal
+from UM.Job import Job
+
+from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
 from UM.Scene.SceneNode import SceneNode
 from UM.Scene.ToolHandle import ToolHandle
-from UM.Application import Application
-
+from UM.Mesh.MeshBuilder import MeshBuilder
 from UM.View.RenderPass import RenderPass
 from UM.View.RenderBatch import RenderBatch
 from UM.View.GL.OpenGL import OpenGL
+
+import numpy
 
 ## RenderPass used to display g-code paths.
 class LayerPass(RenderPass):
     def __init__(self, width, height):
         super().__init__("layerview", width, height)
 
+        self._controller = Application.getInstance().getController()
+
         self._shader = None
         self._tool_handle_shader = None
         self._gl = OpenGL.getInstance().getBindingsObject()
-        self._scene = Application.getInstance().getController().getScene()
+        self._scene = self._controller.getScene()
 
-        self._layerview = None
+        self._layer_view = None
+
+        self._controller.getScene().getRoot().childrenChanged.connect(self._onSceneChanged)
+        self._max_layers = 0
+        self._current_layer_num = 0
+        self._current_layer_mesh = None
+        self._current_layer_jumps = None
+        self._top_layers_job = None
+        self._activity = False
+        self._old_max_layers = 0
+
+        Preferences.getInstance().addPreference("view/top_layer_count", 5)
+        Preferences.getInstance().addPreference("view/only_show_top_layers", False)
+        Preferences.getInstance().preferenceChanged.connect(self._onPreferencesChanged)
+
+        self._solid_layers = int(Preferences.getInstance().getValue("view/top_layer_count"))
+        self._only_show_top_layers = bool(Preferences.getInstance().getValue("view/only_show_top_layers"))
 
     def setLayerView(self, layerview):
-        self._layerview = layerview
+        self._layer_view = layerview
+
+    def getActivity(self):
+        return self._activity
+
+    def getCurrentLayer(self):
+        return self._current_layer_num
+
+    def getMaxLayers(self):
+        return self._max_layers
+
+    def resetLayerData(self):
+        self._current_layer_mesh = None
+        self._current_layer_jumps = None
+
+    def setLayer(self, value):
+        if self._current_layer_num != value:
+            self._current_layer_num = value
+            if self._current_layer_num < 0:
+                self._current_layer_num = 0
+            if self._current_layer_num > self._max_layers:
+                self._current_layer_num = self._max_layers
+
+            self._startUpdateTopLayers()
+
+            self.currentLayerNumChanged.emit()
+
+    def calculateMaxLayers(self):
+        scene = self._controller.getScene()
+        self._activity = True
+
+        self._old_max_layers = self._max_layers
+        ## Recalculate num max layers
+        new_max_layers = 0
+        for node in DepthFirstIterator(scene.getRoot()):
+            layer_data = node.callDecoration("getLayerData")
+            if not layer_data:
+                continue
+
+            if new_max_layers < len(layer_data.getLayers()):
+                new_max_layers = len(layer_data.getLayers()) - 1
+
+        if new_max_layers > 0 and new_max_layers != self._old_max_layers:
+            self._max_layers = new_max_layers
+
+            # The qt slider has a bit of weird behavior that if the maxvalue needs to be changed first
+            # if it's the largest value. If we don't do this, we can have a slider block outside of the
+            # slider.
+            if new_max_layers > self._current_layer_num:
+                self.maxLayersChanged.emit()
+                self.setLayer(int(self._max_layers))
+            else:
+                self.setLayer(int(self._max_layers))
+                self.maxLayersChanged.emit()
+        self._startUpdateTopLayers()
+
+    maxLayersChanged = Signal()
+    currentLayerNumChanged = Signal()
 
     def render(self):
         if not self._shader:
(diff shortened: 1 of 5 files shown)


plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2015 Ultimaker B.V.
```
Ghostkeeper: Happy new year! I've updated this for you in 9e4b7e29ccec1cfb76ef0ac87b65bf4c7ca840b1.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,86 @@
...
+
+        self._shader = None
+        self._tool_handle_shader = None
+        self._gl = OpenGL.getInstance().getBindingsObject()
+        self._scene = Application.getInstance().getController().getScene()
+
+        self._layerview = None
+
+    def setLayerView(self, layerview):
+        self._layerview = layerview
```
Ghostkeeper: This seems like the kind of thing you would want to do in the constructor.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+        super().__init__("layerview", width, height)
+
+        self._controller = Application.getInstance().getController()
+
+        self._shader = None
+        self._tool_handle_shader = None
+        self._gl = OpenGL.getInstance().getBindingsObject()
+        self._scene = self._controller.getScene()
+
+        self._controller.getScene().getRoot().childrenChanged.connect(self._onSceneChanged)
```
Ghostkeeper: You could use `self._scene` here.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+            if self._current_layer_num > self._max_layers:
+                self._current_layer_num = self._max_layers
+
+            self._startUpdateTopLayers()
+
+            self.currentLayerNumChanged.emit()
+
+    def calculateMaxLayers(self):
+        scene = self._controller.getScene()
+        self._activity = True
```
Ghostkeeper: This one is never set back to `False` again. Perhaps you can document what this does and how it differs from `self._busy`?

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+            if not layer_data:
+                continue
+
+            if new_max_layers < len(layer_data.getLayers()):
+                new_max_layers = len(layer_data.getLayers()) - 1
+
+        if new_max_layers > 0 and new_max_layers != self._old_max_layers:
+            self._max_layers = new_max_layers
+
+            # The qt slider has a bit of weird behavior that if the maxvalue needs to be changed first
```
Ghostkeeper: Grammar. Fixed this for you in be2b821cb318004d722533bc3c912954a67fdf0a.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+
+        tool_handle_batch = RenderBatch(self._tool_handle_shader, type = RenderBatch.RenderType.Overlay)
+        tool_handle_has_items = False
+
+        for node in DepthFirstIterator(self._scene.getRoot()):
+            if isinstance(node, ToolHandle):
+                tool_handle_batch.addItem(node.getWorldTransformation(), mesh = node.getSolidMesh())
+                tool_handle_has_items = True
+
+            if type(node) is SceneNode and node.getMeshData() and node.isVisible():
```
Ghostkeeper: Probably safer to use `elif node.getMeshData and node.isVisible():` rather than checking specifically for `SceneNode` objects, seeing as how we're inclined to use subclasses of `SceneNode` from time to time.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+    currentLayerNumChanged = Signal()
+
+    def render(self):
+        if not self._shader:
+            self._shader = OpenGL.getInstance().createShaderProgram(Resources.getPath(Resources.Shaders, "default.shader"))
+        if not self._tool_handle_shader:
+            self._tool_handle_shader = OpenGL.getInstance().createShaderProgram(Resources.getPath(Resources.Shaders, "toolhandle.shader"))
+
+        self.bind()
+        self._gl.glDisable(self._gl.GL_DEPTH_TEST)
```
Ghostkeeper: Wouldn't this cause the layers to mix up their depth ordering again?

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def resetLayerData(self):
+        self._current_layer_mesh = None
+        self._current_layer_jumps = None
+
+    def setLayer(self, value):
+        if self._current_layer_num != value:
+            self._current_layer_num = value
+            if self._current_layer_num < 0:
+                self._current_layer_num = 0
```
Ghostkeeper: Does this mean that you can't visualise halfway through the raft any more?

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,273 @@
...
+                break
+
+        if self._cancel or not layer_data:
+            return
+
+        layer_mesh = MeshBuilder()
+        for i in range(self._solid_layers):
+            layer_number = self._layer_number - i
+            if layer_number < 0:
+                continue
```
Ghostkeeper: No rafts?

plugins/LayerView/LayerView.py
```diff
@@ -178,103 +78,29 @@ def event(self, event):
...
-        if not jump_mesh or jump_mesh.getVertices() is None:
-            jump_mesh = None
+            if not self._composite_pass:
+                self._composite_pass = self.getRenderer().getRenderPass("composite")
 
-        self.setResult({"layers": layer_mesh.build(), "jumps": jump_mesh})
+            self._old_layer_bindings = self._composite_pass.getLayerBindings()
+            layer_bindings = self._old_layer_bindings[:] # make a copy
+            layer_bindings.append("layerview")
+            self._composite_pass.setLayerBindings(layer_bindings)
```
Ghostkeeper: Why does it need to go via this construct, rather than via just `self._composite_pass.getLayerBindings().append("layerview")`?

plugins/LayerView/layerview_composite.shader
```diff
@@ -0,0 +1,80 @@
+[shaders]
+vertex =
+    uniform highp mat4 u_modelViewProjectionMatrix;
+    attribute highp vec4 a_vertex;
+    attribute highp vec2 a_uvs;
+
+    varying highp vec2 v_uvs;
```
Ghostkeeper: We don't have textures in Layer View.

plugins/LayerView/layerview_composite.shader
```diff
@@ -0,0 +1,80 @@
...
+    const vec3 y_axis = vec3(0.0, 1.0, 0.0);
+    const vec3 z_axis = vec3(0.0, 0.0, 1.0);
+
+    void main()
+    {
+        kernel[0] = 0.0; kernel[1] = 1.0; kernel[2] = 0.0;
+        kernel[3] = 1.0; kernel[4] = -4.0; kernel[5] = 1.0;
+        kernel[6] = 0.0; kernel[7] = 1.0; kernel[8] = 0.0;
+
+        vec4 result = vec4(0.965, 0.965, 0.965, 1.0);
```
Ghostkeeper: This is a magic number. Please document it.

plugins/LayerView/layerview_composite.shader
```diff
@@ -0,0 +1,80 @@
...
+
+    const vec3 x_axis = vec3(1.0, 0.0, 0.0);
+    const vec3 y_axis = vec3(0.0, 1.0, 0.0);
+    const vec3 z_axis = vec3(0.0, 0.0, 1.0);
+
+    void main()
+    {
+        kernel[0] = 0.0; kernel[1] = 1.0; kernel[2] = 0.0;
+        kernel[3] = 1.0; kernel[4] = -4.0; kernel[5] = 1.0;
+        kernel[6] = 0.0; kernel[7] = 1.0; kernel[8] = 0.0;
```
Ghostkeeper: Might be useful to document this Gaussian kernel.

plugins/LayerView/LayerPass.py
fieldOfView: This is moved here verbatim from LayerView.py:
https://github.com/Ultimaker/Cura/blob/master/plugins/LayerView/LayerView.py#L124
This behavior did not change in this PR

plugins/LayerView/LayerPass.py
fieldOfView: This is moved here verbatim from LayerView.py:
https://github.com/Ultimaker/Cura/blob/master/plugins/LayerView/LayerView.py#L136
This behavior did not change in this PR

plugins/LayerView/LayerPass.py
fieldOfView: This is moved here verbatim from LayerView.py:
https://github.com/Ultimaker/Cura/blob/master/plugins/LayerView/LayerView.py#L239
This behavior did not change in this PR

PR #908, opened 2016-08-04
Make quality profiles global

awhiemstra:
Changes quality profiles to keep track of changes and make the quality a "global" thing. In addition, we now store global settings in the global stack again. This does not yet implement everything that we discussed but it contains the bare minimum.


awhiemstra:
Everything now works except for import/export, though I am not entirely happy with it. The import/export should in my opinion be done in a follow up since this is already becoming way too large.

cura/Settings/ContainerManager.py
```diff
@@ -350,6 +351,194 @@ def importContainer(self, file_url):
...
+            UM.Logger.log("w", "No quality container found in stack %s, cannot create profile", global_stack.getId())
+            return False
+
+        UM.Application.getInstance().getMachineManager().blurSettings.emit()
+
+        unique_name = UM.Settings.ContainerRegistry.getInstance().uniqueName(quality_container.getName())
+        unique_id = unique_name.lower()
+        unique_id.replace(" ", "_")
+
+        stacks = [ s for s in cura.Settings.ExtruderManager.getInstance().getMachineExtruders(global_stack.getId()) ]
```
Ghostkeeper: You could just use `list(...)` to convert a generator into a list, instead of this list comprehension.

cura/Settings/ContainerManager.py
```diff
@@ -394,7 +583,34 @@ def _updateContainerNameFilters(self):
             name_filter = "{0} ({1})".format(mime_type.comment, suffix_list)
             self._container_name_filters[name_filter] = entry
 
-    # Factory function, used by QML
-    @staticmethod
-    def createContainerManager(engine, js_engine):
-        return ContainerManager()
+    def _getQualityContainers(self, quality_name, container_type = "quality_changes"):
```
Ghostkeeper: It seems somewhat illogical to have a function that says it gets quality containers, but secretly may return quality-changes containers. Why did you collate these two types of containers?

cura/Settings/ContainerManager.py
```diff
@@ -350,6 +351,194 @@ def importContainer(self, file_url):
...
+        UM.Application.getInstance().getMachineManager().blurSettings.emit()
+
+        new_name = UM.Settings.ContainerRegistry.getInstance().uniqueName(new_name)
+        new_id = new_name.lower()
+        new_id.replace(" ", "_")
+
+        for container in self._getQualityContainers(quality_name):
+            stack_id = container.getMetaDataEntry("extruder", global_stack.getId())
+
+            UM.Settings.ContainerRegistry.getInstance().renameContainer(container.getId(), new_name, stack_id + "_" + new_id)
```
Ghostkeeper: Why does the stack_id get prepended before the new id? This needs some documentation.

cura/Settings/ContainerManager.py
```diff
@@ -350,6 +351,194 @@ def importContainer(self, file_url):
...
+
+        UM.Application.getInstance().getMachineManager().activeQualityChanged.emit()
+        return True
+
+    @pyqtSlot(str, result = str)
+    def duplicateQuality(self, quality_name):
```
Ghostkeeper: So this function may duplicate quality as well as quality-changes profiles? I really think these two should be more separate.

cura/Settings/ContainerManager.py
```diff
@@ -394,7 +583,34 @@ def _updateContainerNameFilters(self):
...
+            stacks = [ s for s in cura.Settings.ExtruderManager.getInstance().getMachineExtruders(global_stack.getId()) ]
+            stacks.insert(0, global_stack)
+
+            for stack in stacks:
+                material_ids.append(stack.findContainer(type = "material").getId())
+
+        containers = UM.Settings.ContainerRegistry.getInstance().findInstanceContainers(**criteria)
+        for container in containers:
+            if filter_by_material and container.getMetaDataEntry("material") not in material_ids:
+                continue
```
Ghostkeeper: This at least warrants some documentation. Why do you only want profiles with a material that is currently in one of the extruder stacks?

cura/Settings/MachineManager.py
```diff
@@ -26,16 +26,18 @@ def __init__(self, parent = None):
...
+        self.globalContainerChanged.connect(self.activeVariantChanged)
+        self.globalContainerChanged.connect(self.activeQualityChanged)
+
         self._active_stack_valid = None
         self._onGlobalContainerChanged()
 
         ExtruderManager.getInstance().activeExtruderChanged.connect(self._onActiveExtruderStackChanged)
         self._onActiveExtruderStackChanged()
 
         ##  When the global container is changed, active material probably needs to be updated.
```
Ghostkeeper: This documentation needs to be moved up as well. The code below is about when the active extruders change.

cura/Settings/MachineManager.py
```diff
@@ -489,22 +395,70 @@ def activeMaterialId(self):
...
+        stacks.insert(0, self._global_container_stack)
+
+        for stack in stacks:
+            material_container = stack.findContainer(type = "material")
+            if not material_container:
+                continue
+
+            key = ""
+            if stack == self._global_container_stack:
+                key = "global"
```
Ghostkeeper: Why not `self._global_container_stack.getId()`? No need for this extra case, right?

cura/Settings/ContainerManager.py
awhiemstra: No. It creates quality-changes containers. However, it accepts both a quality-changes container name or a quality container name. For quality changes, it can simply duplicate them. For quality, it creates a new set of quality changes based on the supplied quality.

cura/Settings/MachineManager.py
```diff
@@ -489,22 +395,70 @@ def activeMaterialId(self):
...
+
+        stacks = [ s for s in ExtruderManager.getInstance().getMachineExtruders(self._global_container_stack.getId()) ]
+        stacks.insert(0, self._global_container_stack)
+
+        for stack in stacks:
+            material_container = stack.findContainer(type = "material")
+            if not material_container:
+                continue
+
+            key = ""
```
Ghostkeeper: Unused assignment.

cura/Settings/MachineManager.py
```diff
@@ -669,40 +525,95 @@ def setActiveVariant(self, variant_id):
...
+                material = stack.findContainer(type = "material")
+                criteria["material"] = material.getId()
+
+            stack_quality = UM.Settings.ContainerRegistry.getInstance().findInstanceContainers(**criteria)
+            if not stack_quality:
+                criteria.pop("extruder")
+                stack_quality = UM.Settings.ContainerRegistry.getInstance().findInstanceContainers(**criteria)
+                if not stack_quality:
+                    stack_quality = quality_container
+                stack_quality = stack_quality[0]
```
Ghostkeeper: This doesn't work `if not stack_quality`, because `quality_container` is not a list.

cura/Settings/MachineManager.py
```diff
@@ -669,40 +525,95 @@ def setActiveVariant(self, variant_id):
...
+
+        quality_type = quality_container.getMetaDataEntry("quality_type")
+        if not quality_type:
+            quality_type = quality_changes_container.getName()
+
+        stacks = [ s for s in ExtruderManager.getInstance().getMachineExtruders(self._global_container_stack.getId()) ]
+        stacks.insert(0, self._global_container_stack)
+
+        for stack in stacks:
+            extruder_id = stack.getId() if stack != self._global_container_stack else None
```
Ghostkeeper: If `stack == self._global_container_stack` then extruder_id becomes `None` and you'll be filtering on `"extruder": None`. In that case you're popping the "extruder" from the criteria. I think it's better to first check for `stack != self._global_container_stack` and then only add the extruder criterium inside that if statement.

resources/qml/Cura.qml
```diff
@@ -459,7 +459,8 @@ UM.MainWindow
         target: Cura.Actions.addProfile
         onTriggered:
         {
-            Cura.MachineManager.newQualityContainerFromQualityAndUser();
+//             Cura.MachineManager.newQualityContainerFromQualityAndUser();
```
Ghostkeeper: Remove commented code.

resources/qml/Menus/ProfileMenu.qml
```diff
@@ -38,7 +38,7 @@ import Cura 1.0 as Cura
         id: customProfileInstantiator
         model: UM.InstanceContainersModel
         {
-            filter: menu.getFilter({ "read_only": false });
+            filter: { "type": "quality_changes", "extruder": null, "definition": Cura.MachineManager.filterQualityByMachine ? Cura.MachineManager.activeDefinitionId : "fdmprinter" };
```
Ghostkeeper: `"extruder": null`? What's that doing?

resources/qml/Preferences/ProfileTab.qml
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2015 Ultimaker B.V.
```
Ghostkeeper: What year are you living in? :stuck_out_tongue:

resources/qml/Preferences/ProfileTab.qml
```diff
@@ -0,0 +1,38 @@
...
+// Uranium is released under the terms of the AGPLv3 or higher.
```
Ghostkeeper: That may be so, but I'm more interested in the terms of Cura.

resources/qml/Preferences/ProfileTab.qml
```diff
@@ -0,0 +1,38 @@
...
+import QtQuick.Controls 1.1
+
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+Tab
+{
+    id: base
+
+    property string extruderId: "";
```
Ghostkeeper: `extruderId` is not used?

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -106,13 +106,15 @@ UM.ManagementPage
...
             onClicked: importDialog.open();
+            enabled: false
         },
         Button
         {
             text: catalog.i18nc("@action:button", "Export")
             iconName: "document-export"
             onClicked: exportDialog.open()
-            enabled: currentItem != null
+//             enabled: currentItem != null
```
Ghostkeeper: Remove commented code.

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -106,13 +106,15 @@ UM.ManagementPage
             text: catalog.i18nc("@action:button", "Import");
             iconName: "document-import";
             onClicked: importDialog.open();
+            enabled: false
```
Ghostkeeper: Aww :(

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -173,7 +175,7 @@ UM.ManagementPage
 
             Label {
                 id: defaultsMessage
-                visible: currentItem && !currentItem.metadata.has_settings
+                visible: false
```
Ghostkeeper: Why is this never visible any more?

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -187,69 +189,31 @@ UM.ManagementPage
             }
         }
 
-        ScrollView {
-            id: scrollView
-
+        TabView
```
Ghostkeeper: Does this mean that it doesn't scroll any more?

resources/qml/Settings/SettingItem.qml
```diff
@@ -27,7 +27,6 @@ Item {
 
     // Create properties to put property provider stuff in (bindings break in qt 5.5.1 otherwise)
     property var state: propertyProvider.properties.state
-    property var settablePerExtruder: propertyProvider.properties.settable_per_extruder
```
Ghostkeeper: Why was this removed?

resources/qml/Preferences/ProfileTab.qml
nallath: http://i3.kym-cdn.com/photos/images/facebook/000/209/945/D6PfW.jpg

cura/Settings/MachineManager.py
awhiemstra: Except that filtering on "extruder": None is a valid filtering criteria since it means it will only match containers that do not have an extruder set.

resources/qml/Menus/ProfileMenu.qml
awhiemstra: The same as above (as null is converted to None by PyQt): It excludes containers that have extruder set.

resources/qml/Preferences/ProfilesPage.qml
awhiemstra: Because I have no proper solution for it. The check that is there currently is broken since when the quality_changes container is empty that does not mean the underlying quality container is empty. I found the entire thing rather dubious so I hid it until I could discuss it, but then forgot to discuss it.

resources/qml/Preferences/ProfilesPage.qml
awhiemstra: Check ProfileTab.qml. TableView already scrolls.

PR #909, opened 2016-08-05
Bugfix: Can't create shaders using Nvidia drivers on Linux Mint

Nihlus:
Issue #88 is back again, this time on Linux Mint. This pull request adds it to the check in CuraApplication, which fixes the issue. Mint reports itself as "LinuxMint" in python, despite being an Ubuntu derivation.

I'm not all that good with Python, so I couldn't implement the more general fix suggested in the TODO comment.


thopiekar:
No problem. Collecting all the distributions here is good enough 😉

awhiemstra:
Hmm, now that we do not need an extra dependency we could simply remove the distro check.

thopiekar:
But are you sure it isn't a packaging issue? I mean with the closed source nvidia driver?
That's why I added this check here in the past...

thopiekar:
Added "LinuxMint" via https://github.com/Ultimaker/Cura/commit/6b2935379f46429b125ba3a786324a6617296ef0
I just moved the workaround to ```cura_app.py``` to have all workarounds at one place (if possible).

Issue #910, opened 2016-08-05
[master] Empty/blank text in Menu lists


nickthetait:
Application Version:
Master branch of Cura 021a33d6bd5114d3ff7edb8fa0a85c91ba55d08d
Master branch of Uranium 48ec68d98f8fa7af311b0793135ea88d1b7759ac
Platform: Debian Jessie
Qt: 5.6
PyQt: PyQt5_gpl-5.6
Display Driver: Intel Corporation Xeon E3-1200 v2/3rd Gen Core processor Graphics Controller (rev 09) - i915

Steps to Reproduce:
Setup development environment
Run Cura from source code

Actual Results:
Opening a menu entry, many entries are displayed as unselectable or with no text at all. Clicking one of these misbehaving entries does nothing except close the menu. Correctly displayed entries function as expected.

Menus with only broken items: Edit, Preferences, Help
Mixed broken/working menus: File, Settings, Right Click in scene view
Fully working menus: View, Extensions 
![screenshot - 08052016 - 10 25 29 am](https://cloud.githubusercontent.com/assets/11818614/17444246/2f1630b4-5afc-11e6-829e-b798bf9bcabf.png)
![screenshot - 08052016 - 10 25 11 am](https://cloud.githubusercontent.com/assets/11818614/17444245/2f15a54a-5afc-11e6-967a-f045dfbdbf66.png)
![screenshot - 08042016 - 11 10 46 am](https://cloud.githubusercontent.com/assets/11818614/17444247/2f173d60-5afc-11e6-82f9-d52c32ebbf4e.png)

Expected results:
Able to use all menu entries

Additional Information:
First reported here: https://github.com/Ultimaker/CuraEngine/issues/375#issuecomment-237258003
Making a new ticket to better figure out root cause and to help people who run into this problem in the future.

Able to load and slice a model fine.

Each time a menu is opened/closed this error message is thrown:
file:///home/aleph/Software/AlephObjects/Cura2/QT/install/5.6/gcc_64/qml/QtQuick/Controls/Private/ScrollViewHelper.qml:137:9: QML Binding: Property 'raised' does not exist on Item.

[...]

PR #914, opened 2016-08-08
Support for the X3D format

sevaa:
Initial support for the X3D scene file format (see http://www.web3d.org/).

For now, just mesh import. Supports all mesh geometry nodes in Rendering, Geometry3D, Geometry2D,  and CAD geometry modules, specifically:

 - Box 
 - Cone 
 - Cylinder 
 - Sphere
 - ElevationGrid 
 - Extrusion 
 - IndexedFaceSet
 - IndexedTriangleFanSet
 - IndexedTriangleSet
 - IndexedTriangleStripSet
 - TriangleFanSet
 - TriangleSet
 - TriangleStripSet
 - IndexedQuadSet 
 - QuadSet
 - Disk2D
 - Rectangle2D
 - TriangleSet2D

No support for materials, color, or textures. Normals in the file (if any) are ignored. Light sources, viewpoints, sensors, world, shaders, scripts also ignored.

There are some more geometries in the standard: Text, GeoElevationGrid, and 4 NURBS based surfaces. No way I'm hand-tesselating those.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -150,7 +150,7 @@ class StartSliceJob(Job):
                     obj.id = id(object)
                     verts = mesh_data.getVertices()
                     indices = mesh_data.getIndices()
-                    if not indices is None:
+                    if indices is not None:
                         verts = numpy.array([verts[vert_index] for face in indices for vert_index in face])
                     else:
                         verts = numpy.array(verts)
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -148,7 +148,12 @@ class StartSliceJob(Job):
 
                     obj = group_message.addRepeatedMessage("objects")
                     obj.id = id(object)
-                    verts = numpy.array(mesh_data.getVertices())
+                    verts = mesh_data.getVertices()
+                    indices = mesh_data.getIndices()
+                    if not indices is None:
+                        verts = numpy.array([verts[vert_index] for face in indices for vert_index in face])
+                    else:
+                        verts = numpy.array(verts)
 
                     # Convert from Y up axes to Z up axes. Equals a 90 degree rotation.
                     verts[:, [1, 2]] = verts[:, [2, 1]]
--- a/plugins/X3DReader/X3DReader.py
+++ b/plugins/X3DReader/X3DReader.py
@@ -66,6 +66,7 @@ class X3DReader(MeshReader):
                 theScene.addDecorator(group_decorator)
                 for node in self.sceneNodes:
                     theScene.addChild(node)
+                theScene.setSelectable(True)
             elif len(self.sceneNodes) == 1:
                 theScene = self.sceneNodes[0]
             else: # No shapes read :(
@@ -700,7 +701,6 @@ def addQuadFlip(bui, a, b, c, d, ccw):
 # - Find an outer vertex with the smallest angle and no vertices inside its adjacent triangle
 # - Remove the triangle at that vertex
 # - Repeat until done
-# Note that n is the count of vertices in the face, but the `face` array is one element bigger, with nth element same as the 0th one
 # Vertex coordinates are supposed to be already in the mesh builder object
 def addFace(bui, indices, ccw):
     # Resolve indices to coordinates for faster math
--- a/plugins/X3DReader/X3DReader.py
+++ b/plugins/X3DReader/X3DReader.py
@@ -1,0 +1,792 @@
+# Seva Alekseyev with National Institutes of Health, 2016
+
+from UM.Mesh.MeshReader import MeshReader
+from UM.Mesh.MeshBuilder import MeshBuilder
+from UM.Logger import Logger
+from UM.Math.Matrix import Matrix
+from UM.Math.Vector import Vector
+from UM.Scene.SceneNode import SceneNode
+from UM.Scene.GroupDecorator import GroupDecorator
+from UM.Job import Job
+from math import pi, sin, cos, sqrt
+import numpy
+
+EPSILON = 0.000001 # So very crude. :(
+
+try:
+    import xml.etree.cElementTree as ET
+except ImportError:
+    import xml.etree.ElementTree as ET
+    
+    
+DEFAULT_SUBDIV = 16 # Default subdivision factor for spheres, cones, and cylinders    
+
+class X3DReader(MeshReader):
+    def __init__(self):
+        super().__init__()
+        self._supported_extensions = [".x3d"]
+        self._namespaces = {}
+        self.defs = {}
+        
+    def read(self, file_name):
+        try:
+            self.sceneNodes = []
+            self.fileName = file_name
+            self.transform = Matrix()
+            self.transform.setByScaleVector(Vector(1000, 1000, 1000))
+            
+            tree = ET.parse(file_name)
+            self.processNode(tree.getroot())
+            
+            if len(self.sceneNodes) > 1:
+                theScene = SceneNode()
+                group_decorator = GroupDecorator()
+                theScene.addDecorator(group_decorator)
+                for node in self.sceneNodes:
+                    theScene.addChild(node)
+            elif len(self.sceneNodes) == 1:
+                theScene = self.sceneNodes[0]
+            else: # No shapes read :(
+                return None
+            theScene.setName(file_name)
+        except Exception as e:
+            Logger.log("e", "exception occured in x3d reader: %s", e)
+
+        try:
+            boundingBox = theScene.getBoundingBox()
+            boundingBox.isValid()
+        except:
+            return None
+
+        return theScene
+    
+    # ------------------------- XML tree traversal
+  
+    def processNode(self, xmlNode):
+        xmlNode =  self.resolveDefUse(xmlNode)
+        if xmlNode is None:
+            return
+        
+        tag = xmlNode.tag
+        if tag in ("X3D", "Scene", "Group", "StaticGroup", "CADAssembly", "CADFace", "CADLayer", "CADPart", "Collision"):
+            self.processChildNodes(xmlNode)
+        elif tag == "LOD":
+            self.processNode(xmlNode[0])
+        elif tag == "Transform":
+            self.processTransform(xmlNode)
+        elif tag == "Shape":
+            self.processShape(xmlNode)
+            
+            
+    def processShape(self, xmlNode):
+        # Find the geometry and the appearance inside the Shape
+        geometry = appearance = None
+        for subNode in xmlNode:
+            if subNode.tag == "Appearance" and not appearance:
+                appearance = self.resolveDefUse(subNode)
+            elif subNode.tag in self.geometryImporters and not geometry:
+                geometry = self.resolveDefUse(subNode)
+        
+        # TODO: appearance is completely ignored. At least apply the material color...        
+        if not geometry is None:
+            try:
+                bui = MeshBuilder()
+                self.geometryImporters[geometry.tag](self, geometry, bui)
+                
+                bui.calculateNormals()
+                bui.setFileName(self.fileName)
+                
+                sceneNode = SceneNode()
+                if "DEF" in geometry.attrib:
+                    sceneNode.setName(geometry.tag + "#" + geometry.attrib["DEF"])
+                else:
+                    sceneNode.setName(geometry.tag)
+                     
+                sceneNode.setMeshData(bui.build().getTransformed(self.transform))
+                sceneNode.setSelectable(True)
+                self.sceneNodes.append(sceneNode)
+                
+            except Exception as e:
+                Logger.log("e", "exception occured in x3d reader while reading %s: %s", geometry.tag, e)
+        
+    # Returns the referenced node if the node has USE, the same node otherwise.
+    # May return None is USE points at a nonexistent node
+    # In X3DOM, when both DEF and USE are in the same node, DEF is ignored.
+    # Big caveat: XML node objects may evaluate to boolean False!!! 
+    def resolveDefUse(self, node):
+        USE = node.attrib.get("USE")
+        if USE:
+            return self.defs.get(USE, None)
+
+        DEF = node.attrib.get("DEF")            
+        if DEF:
+            self.defs[DEF] = node 
+        return node
+    
+    def processChildNodes(self, node):
+        for c in node:
+            self.processNode(c)
+            Job.yieldThread()
+    
+    # Since this is a grouping node, will recurse down the tree.
+    # According to the spec, the final transform matrix is:
+    # T * C * R * SR * S * -SR * -C
+    # Where SR corresponds to the rotation matrix to scaleOrientation
+    # C and SR are rather exotic. S, slightly less so. 
+    def processTransform(self, node):
+        rot = readRotation(node, "rotation", (0, 0, 1, 0)) # (angle, axisVactor) tuple
+        trans = readVector(node, "translation", (0, 0, 0)) # Vector
+        scale = readVector(node, "scale", (1, 1, 1)) # Vector
+        center = readVector(node, "center", (0, 0, 0)) # Vector
+        scaleOrient = readRotation(node, "scaleOrientation", (0, 0, 1, 0)) # (angle, axisVactor) tuple
+        
+        # Store the previous transform; in Cura, the default matrix multiplication is in place        
+        prev = Matrix(self.transform.getData()) # It's deep copy, I've checked
+        
+        # The rest of transform manipulation will be applied in place
+        gotCenter = (center.x != 0 or center.y != 0 or center.z != 0)
+        
+        T = self.transform
+        if trans.x != 0 or trans.y != 0 or trans.z !=0:
+            T.translate(trans)
+        if gotCenter:
+            T.translate(center)
+        if rot[0] != 0:
+            T.rotateByAxis(*rot)
+        if scale.x != 1 or scale.y != 1 or scale.z != 1:
+            gotScaleOrient = scaleOrient[0] != 0
+            if gotScaleOrient:
+                T.rotateByAxis(*scaleOrient)
+            # No scale by vector in place operation in UM
+            S = Matrix()
+            S.setByScaleVector(scale)
+            T.multiply(S)
+            if gotScaleOrient:
+                T.rotateByAxis(-scaleOrient[0], scaleOrient[1])
+        if gotCenter:
+            T.translate(-center)
+            
+        self.processChildNodes(node)
+        self.transform = prev
+    
+    # ------------------------- Geometry importers
+    # They are supposed to fill the MeshBuilder object with vertices and faces, the caller will do the rest
+    
+    # Primitives
+
+    def geomBox(self, node, bui):
+        size = readFloatArray(node, "size", [2, 2, 2])
+        bui.addCube(size[0], size[1], size[2])
+    
+    # The sphere is subdivided into nr rings and ns segments
+    def geomSphere(self, node, bui):
+        r = readFloat(node, "radius", 0.5)
+        subdiv = readIntArray(node, 'subdivision', None)
+        if subdiv:
+            if len(subdiv) == 1:
+                nr = ns = subdiv[0]
+            else:
+                (nr, ns) = subdiv
+        else:
+            nr = ns = DEFAULT_SUBDIV
+            
+        
+        lau = pi / nr  # Unit angle of latitude (rings) for the given tesselation
+        lou = 2 * pi / ns  # Unit angle of longitude (segments)
+        
+        bui.reserveFaceAndVertexCount(ns*(nr*2 - 2), 2 + (nr + 1)*ns)
+    
+        
+        # +y and -y poles
+        bui.addVertex(0, r, 0)
+        bui.addVertex(0, -r, 0)
+        
+        # The non-polar vertices go from x=0, negative z plane counterclockwise -
+        # to -x, to +z, to +x, back to -z
+        for ring in range(1, nr):
+            for seg in range(ns):
+                bui.addVertex(-r*sin(lou * seg) * sin(lau * ring),
+                          r*cos(lau * ring),
+                          -r*cos(lou * seg) * sin(lau * ring))
+                
+        vb = 2 + (nr - 2) * ns  # First vertex index for the bottom cap
+    
+        # Faces go in order: top cap, sides, bottom cap.
+        # Sides go by ring then by segment.
+    
+        # Caps
+        # Top cap face vertices go in order: down right up
+        # (starting from +y pole)
+        # Bottom cap goes: up left down (starting from -y pole)
+        for seg in range(ns):
+            addTri(bui, 0, seg + 2, (seg + 1) % ns + 2)
+            addTri(bui, 1, vb + (seg + 1) % ns, vb + seg)
+    
+        # Sides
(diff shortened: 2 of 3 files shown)


sevaa:
The file below is a renamed ZIP file that contains X3D scenes for every kind of geometry that we support. As a reference implementation, to see what those scenes are supposed to look like, I recommend the free player called InstantReality ( http://www.instantreality.org/ ).

[BasicGeometryTests.ZIP.txt](https://github.com/Ultimaker/Cura/files/407093/BasicGeometryTests.ZIP.txt)

Ghostkeeper:
Looks good at first glance. Haven't done a real code review yet.

Geometry is the only thing it needs to import, since Cura has no support for (and no need for) colours, materials and textures. The user selects the material, after all. Normals are computed by Uranium automatically. Light sources, viewpoints, world and shaders are determined by Uranium and we don't want sensors and scripts to run anyway.

nallath:
Did we create an internal issue for review / testing this?

Ghostkeeper:
Not yet, as far as I know.

sevaa:
> Cura has no support for (and no need for) colours

Well, there is a _colors_ array in a MeshBuilder, so on some level Cura has colors :) But I'm not populating it.

Anyways, I've tried importing with a fairly sizable model (attached), and Cura froze for 10-15 seconds, with no activity in the console, well after my code has returned. Not  good, but I was not sure if it was my fault, or Cura's.

[1YSA-ribbon-rainbow.zip.txt](https://github.com/Ultimaker/Cura/files/409102/1YSA-ribbon-rainbow.zip.txt)

fieldOfView:
> Well, there is a colors array in a MeshBuilder, so on some level Cura has colors

That is _Uranium_, not Cura ;)

fieldOfView:
From what I saw during a brief peek is that the object seems to be loaded as a LOT of grouped subobjects. I count ~200 shapes in the X3D. I don't think we have ever loaded that many "objects" into Cura, and it seems like Cura does not like it. Additionally, curaengine complains a lot about faces not connecting.

sevaa:
I'm translating every Shape object from the scene into a separate UM SceneNode object. Took that design from the 3MF reader. I can, instead, merge all meshes in the scene into one big SceneNode. Please advice.

As for the "faces not connecting" message, I saw it on the most innocent geometry (e. g. on OneIndTriFan.x3d from the reference geometry archive linked above). Since that one is perfectly connected, I think the message is bogus. Please trace the code, watch the contents of _indices, and let me know what am I missing.

fieldOfView:
Paging @BagelOrb for the "faces not connecting" thing (I don't think it is bogus, but I know little to nothing about CuraEngine).

The usecase for the 3MF reader importing objects into separate SceneNodes is mainly dual extrusion prints, to make it easier to import an object that consists of multiple parts that should be printed with multiple extruders. It was not meant to load hundreds of SceneNodes, but like I've said I don't think we have even tried to do that so it could be there's something fairly trivial that could fix this on our side and make it perform better. I just don't know.

sevaa:
The geometry I've mentioned is an octahedron. The vertices go:

[ 0.  1.  0.]
[ 0.  0. -1.]
[-1.  0.  0.]
[ 0.  0.  1.]
[ 1.  0.  0.]
[ 0. -1.  0.]

The faces go:

[0 1 2]
[0 2 3]
[0 3 4]
[0 4 1]
[5 1 4]
[5 4 3]
[5 3 2]
[5 2 1]

There's the top vertex on the +Y axis, then a square in the XZ plane, then the bottom in the -Y axis. The triangles connect top to the square sides, then the bottom. Please tell me what's unconnected about this.

And yet I get errors:

> 2016-08-09 11:17:50,055 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 0.
> 2016-08-09 11:17:50,056 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 0.
> 2016-08-09 11:17:50,056 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 0.
> 2016-08-09 11:17:50,056 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 1.
> 2016-08-09 11:17:50,057 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 1.
> 2016-08-09 11:17:50,057 - DEBUG - ...\UM\Backend\Backend.py (_backendLog [80]): [Backend] [ERROR] Couldn't find face connected to face 1.

nickthetait:
Went and tried out this PR a little, seems to be working overall. I have not even glanced at the code itself.
screenshot - 08102016 - 03 12 03 pm

Being able to compare side by side identical files in both X3D and STL formats would make verification much easier. And therefore make it an easy choice for Ultimaker to merge. Sevaa, are you able to provide a few test cases like this in both formats? This would help determine what is causing the "unconnected face" problem.

Below are a few items that I ran into. Can't know for sure if they are bugs in the PR, Cura or Uranium mostly because I'm still suffering from https://github.com/Ultimaker/Cura/issues/910 

For some reason I was not able to load the OneExtrusion.x3d file.
2016-08-10 15:06:30,427 - ERROR - .../cura/../plugins/X3DReader/X3DReader.py (read [32]): exception occured in x3d reader: not well-formed (invalid token): line 14, column 2

There is something strange relating to how objects are grouped. 
Individual nodes are interpreted by Cura as distinct objects, but there seems not to be a selectable group for the whole entire object. In this picture you can see that the object center is based on one of the cylinders but not the whole group if that makes sense.
screenshot - 08102016 - 03 13 46 pm

A similar thing occurs when putting part of the model outside the print area. The grey on this castle is indicating that it is out of bounds and cannot be printed. What I think should happen is the entire castle should be marked grey/unprintable in this case.

There is definitely poor performance with the ribbon-rainbow file (the most complex one) but I did not invest any time to figure out why...

I'm quite fond of the castle model :japanese_castle: It is a great test case. Did you design it sevaa?

fieldOfView:
Something weird is going on in those screenshots. Looks like the normals are not being correctly calculated.

sevaa:
My copy of OneExtrusion loads fine. I've replaced it in the archive and reuploaded into the comment above.

There was no setSelectable call for the whole scene in case of a multiobject scene. Added, committed, pushed.

Side-by-side STL comparison seems to be a good idea, but I'll have to investigate it a bit. I've tried converting with x3d2stl from InstantReality, but it produces bogus models that error out in Cura. I'm on it...

sevaa:
Found another case where indexed meshes are mishandled, this time in Uranium: UM.Mesh.MeshData.calculateNormalsFromIndexedVertices(). It allocates the normals array as 3*face_count, then runs through all the faces, assuming indices in the face will run all 3*n possible values. That is clearly not so in case of nontrivially indexed meshes. A good portion of the normals array remains zero.

Can someone please explain to me what is the expected structure of the normals array in a MeshBuilder? It's neither by face nor by vertex. Are they face normals, vertex normals, or face normals applied to vertices (if so, which face)?

sevaa:
Shape merging is DEFINITELY the way to go. The ribbon model, and even a larger one (attached), load in reasonable time.

They still take forever to slice though.

[4OSH-ribbon-rainbow.zip.txt](https://github.com/Ultimaker/Cura/files/416025/4OSH-ribbon-rainbow.zip.txt)

fieldOfView:
Hi @sevaa, we have not forgotten about you or this plugin. Due to holidays we are currently rather understaffed, so it will take a little longer before we have the time to give this PR the attention it deserves.

sevaa:
Understandable. I might still have some performance enhancements for large meshes, but the basic functionality is all in place.

plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -148,7 +148,12 @@ def run(self):
 
                     obj = group_message.addRepeatedMessage("objects")
                     obj.id = id(object)
-                    verts = numpy.array(mesh_data.getVertices())
+                    verts = mesh_data.getVertices()
+                    indices = mesh_data.getIndices()
+                    if not indices is None:
```
nallath: Codestyle. SHould be 'if indicides is not None' (Makes it easier to read what is going on)

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
+# Seva Alekseyev with National Institutes of Health, 2016
```
nallath: Note that the code is also AGPL v3, as it's based on other AGPL v3 code (which infects this code)

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+except ImportError:
+    import xml.etree.ElementTree as ET
+    
+# TODO: preserve the structure of scenes that contain several objects
+# Use CADPart, for example, to distinguish between separate objects    
+    
+DEFAULT_SUBDIV = 16 # Default subdivision factor for spheres, cones, and cylinders
+
+class Shape:
+    def __init__(self, v, f, ib, n):
```
nallath: A bit more documentation  or clearer names would be nice here. It doesn't do much now, but it could be that it will grow in size in the future. 
Personally i think that using self.verts = verts is better readable than self.verts = v

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+            
+            self.transform = Matrix()
+            self.transform.setByScaleFactor(scale)
+            self.index_base = 0
+            
+            # Traverse the scene tree, populate the shapes list
+            self.processChildNodes(xml_scene)
+            
+            if self.shapes:
+                bui = MeshBuilder()
```
nallath: Why call it bui? Seems a bit of a weird / undiscriptive name

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+                scene.setName(file_name)
+            else:
+                return None
+            
+        except Exception as e:
+            Logger.log("e", "exception occured in x3d reader: %s", e)
+
+        try:
+            boundingBox = scene.getBoundingBox()
+            boundingBox.isValid()
```
nallath: This line doesn't do anything, unless bounding box is None (which causes the exception to be raised).
The exception is also far to broad. If you want an exeption that is that broad, document why and let the logger write the exception somewhere.

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+            try:
+                self.verts = self.faces = [] # Safeguard 
+                self.geometry_importers[geometry.tag](self, geometry)
+                m = self.transform.getData()
+                verts = numpy.array([m.dot(vert)[:3] for vert in self.verts])
+                self.shapes.append(Shape(verts, self.faces, self.index_base, geometry.tag))
+                self.index_base += len(verts)
+                
+            except Exception as e:
+                Logger.log("e", "exception occured in x3d reader while reading %s: %s", geometry.tag, e)
```
nallath: Have a look at Logger.logException. That logging already handles the printing of exception so you don't have to.

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+    # According to the spec, the final transform matrix is:
+    # T * C * R * SR * S * -SR * -C
+    # Where SR corresponds to the rotation matrix to scaleOrientation
+    # C and SR are rather exotic. S, slightly less so. 
+    def processTransform(self, node):
+        rot = readRotation(node, "rotation", (0, 0, 1, 0)) # (angle, axisVactor) tuple
+        trans = readVector(node, "translation", (0, 0, 0)) # Vector
+        scale = readVector(node, "scale", (1, 1, 1)) # Vector
+        center = readVector(node, "center", (0, 0, 0)) # Vector
+        scaleOrient = readRotation(node, "scaleOrientation", (0, 0, 1, 0)) # (angle, axisVactor) tuple
```
nallath: bit nitpicking, but we use underscore seperation for local members (so it should be scale_orient).

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+            
+        self.processChildNodes(node)
+        self.transform = prev
+    
+    # ------------------------- Geometry importers
+    # They are supposed to fill the MeshBuilder object with vertices and faces, the caller will do the rest
+    
+    # Primitives
+
+    def geomBox(self, node):
```
nallath: I think naming it something like; createGeomBox is more clear. The naming is a bit confusing now.

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+        self.addQuad(4, 0, 3, 7)   # +x
+        self.addQuad(7, 3, 2, 6)   # -z
+        self.addQuad(6, 2, 1, 5)   # -x
+        self.addQuad(5, 1, 0, 4)   # +z
+        self.addQuad(7, 6, 5, 4)  # -y
+    
+    # The sphere is subdivided into nr rings and ns segments
+    def geomSphere(self, node):
+        r = readFloat(node, "radius", 0.5)
+        subdiv = readIntArray(node, 'subdivision', None)
```
nallath: Use " instead of '

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+def pointInsideTriangle(vx, next, prev, nextXprev):
+    vxXprev = vx.cross(prev)
+    r = ratio(vxXprev, nextXprev)
+    if r < 0:
+        return False;
+    vxXnext = vx.cross(next);
+    s = -ratio(vxXnext, nextXprev)
+    return s > 0 and (s + r) < 1
+    
+def toNumpyRotation(rot):
```
nallath: What rotation to numpy rotation? Euler angles? Quaternion? Axis angle? Bit of documentation would be nice here :)

plugins/X3DReader/X3DReader.py
```diff
@@ -0,0 +1,902 @@
...
+# For stability, taking the ration between the biggest coordinates would be better; none of that, either.   
+def ratio(a, b):
+    if b.x > EPSILON or b.x < -EPSILON:
+        return a.x / b.x
+    elif b.y > EPSILON or b.y < -EPSILON:
+        return a.y / b.y
+    else:
+        return a.z / b.z    
+    
+def pointInsideTriangle(vx, next, prev, nextXprev):
```
nallath: Bit more documentation would be nice. I don't see what all the parameters are right away.

nallath:
I've left some remarks. Nothing really big. Code looks pretty good so far, so as far as I'm concerned we could merge this pretty soon and ship it with the next release (note that I haven't tested it yet, but I don't expect any problems there).

sevaa:
Understood. Will address on Monday, I'm AFK now.

Some of the questionable bits, like the usage of Logger, were copied from the 3MF reader verbatim, and that's in master :) Just sayin'.

If they insist on having a verb in geometry handler function names, I'm use "process", that's consistent with the rest of the code.

nallath:
Yeah the 3mf writer was written before we had code reviews, so it doesn't adhere to our coding standards. 
We're getting pretty close to a release, so we won't put much effort into making sure that old code adheres to it. We will need to tackle it at some point.

sevaa:
Addressed all the code review issues.

Ghostkeeper:
I'm assigned for the component test. I found these issues:
* All coordinates are currently interpreted as metres. Convention in the maker community is to interpret unmarked units as millimetres. Is this a hard requirement of the X3D format? If so, we can leave it this way, as long as the CAD applications would output designs in metres as well. Most of the test models provided in the .zip are scaled down to 2% or so. Also the 3D printed heart I downloaded from web3d.org had to be scaled down to 0.1%.
* Normals are still weird. Most of them seem to have vertex-interpolated normals instead of face-interpolated. This makes it hard to discern the shape of the mesh, as can be seen in both of nickthetait's screenshots.

Performance is nice now, by the way. On par with other mesh readers, at least. I've tried a few X3D models from web3d.org. Everything loaded fine. Only the OneElevation.x3d file from your zip of examples may have not loaded entirely as intentional, but it depends on how you subdivide a quad into triangles. I guess it could be interpreted either way.

I'll take a quick look if I can give you a hint on those normals.

Ghostkeeper:
Welp, the normals problem really seems to be a problem in Uranium. I tried for a bit, but couldn't fix it directly and we haven't the resources to inspect it further right now. I guess it'll have to do with faulty normals for now.

The root of the problem is that the normals list always needs to be exactly as long as the vertices list per the getNormals() function in MeshBuilder. If the mesh is indexed and you still want normals per face instead of per vertex, this will have to change or the indexing needs to be dereferenced before then.

sevaa:
1. See http://www.web3d.org/documents/specifications/19775-1/V3.3/index.html , section 4.3.6:

> ISO/IEC 19775 defines the initial base unit of measure of the world coordinate system to be metres. However, the world coordinate units may be modified by specifying a different length unit using the UNIT statement.

I do respect the UNIT statement. The code that generates those ribbon models of mine doesn't, unfortunately, care for real life scale; as far as it's concerned, a unit of length is one angstrom :) Those models represent protein molecules.

2. This could be yet another artifact of Cura (or Uranium) mishandling an indexed mesh. STL and 3MF meshes are not indexed, they're just arrays of vertex triples. X3D has numerous provisions for nontrivial indexing. I don't calculate normals myself, I feed the array of vertices and indices (e. g. faces) to a `SceneNode` object, and it does the rest.

sevaa:
About normals: let me see how does Uranium do it. So, the normals array should be same length as the vertices array **but** normals should correspond to faces, right? I presume, Cura doesn't expect normal averaging for shared vertices, right? A bit of index-to-face resolution at normal generation time will probably do the trick. However, I can't fold those changes under this pull request, since Uranium is a different repo.

The X3D standard does calls for normal averaging in case of indexed meshes. That's why the three pyramids at http://www.jishop.com/temp/x3d/#TriVsIndTri look differently - the bottom one is indexed, so smooth shading is possible. However, faithful model display is not the point of Cura, and for slicing purposes, normals are irrelevant, are they?

In case I go ahead and start messing with Uranium's normal generation code, please tell me, how can I check if normals are still off?

nallath:
OBJ also uses indexed faces. So you could check and see if OBJ loading works correctly.

Ghostkeeper:
Normals in Cura are only for display, as a preview before printing. In practice, it works best if you use per-face normals, because it allows you to see the edges very clearly. In games and such they use normal averaging to create the illusion of high resolution / smooth surfaces.

awhiemstra:
Normals are handled the way the GPU expects them: one normal per vertex. Smoothed normals are simply an artefact of the direction of those normals. For OBJ, I implemented loading normals from file to make it possible to give the build platform smoothed corners. But OBJ is a pretty simple format that has the same assumptions as STL for most of the geometry from what I recall.

In addition, we always recalculate the normals for printable objects iirc. You might actually want to avoid that if X3D provides proper normal information.

sevaa:
X3D supports a data structure for mesh normals, but doesn't require it. When given, there's a option for normals being specified per-face or per-vertex.

I, for now, ignore them in the file altogether, and let the platform code recalculate them :)

nickthetait:
I don't think that WIP tag is accurate anymore. @sevaa you are not aware of anything else that needed fixing right?

@nallath is this being considered for inclusion in 2.3?

nallath:
Yep. It's in code review now, but there is nothing to review since last time. I'll have a look tomorrow.

thopiekar:
@nickthetait My fault..

nickthetait:
cool :)

sevaa:
There's a problem with normal generation on indexed meshes that should be fixed in Uranium. There's an unresolved issue with OneExtrusion.x3d that doesn't reproduce on my end (I think they were testing with an old version of the scene). Other than than, no.

Performance on huge models could probably use some enhancement, but that's an open ended endeavor :)

nallath:
I don't think we will have time to fix the normal generation for 2.3 :( We're in the stage where we only fix blocking & Critical bugs and i don't think we see that issue as either.

nickthetait:
In that case, since 2.3 has its own branch now it would be safe to merge this into master right? Obviously need to create a task for fixing the normals, but the longer this sits un-merged higher chance conflicts will arise...

Ghostkeeper:
Well, the 2.3 branch is under discussion. We've had the command from the product lead to go ahead and continue on master, to prevent having to merge every time and prevent having to test both versions. A few of us disagree with that decision and would opt to only test the 2.3 branch, like we did with 2.1. But for now the branch is sort of in limbo.

Ghostkeeper:
Just got word from tha boss that he finds it acceptable to have untranslated strings for this feature. So as far as he's concerned, this issue can be merged.

Ghostkeeper:
Works, apart from the two issues I originally found (scaling, which was a won't-fix, and the normal vector issue, which is made into a separate issue on our internal tracker).

nickthetait:
Hooray for being merged \o/

Patola:
Hi, just a question about the practical use of the X3D format. I've seen that "No support for materials, color, or textures"; so, it means that currently I cannot use the format to convey information about double/triple extrusion, as I can with .3mf? Is that right?

BTW, congratulations for the format support. I am eager to help testing this feature (as soon as I get Cura compiled).

sevaa:
The X3D format supports materials, color, and textures. But the import module ignores those while importing, so they won't be reflected in Cura in any way. See the Aug 9 discussion earlier in this thread.

BagelOrb:
Triple extrusion? What kind of machine do you have?

I wonder whether cura is up for that task..

Op 21 sep. 2016 18:22 schreef "Seva Alekseyev" <notifications@github.com>:

The X3D format supports materials, color, and textures. But the import
module ignores those while importing, so they won't be reflected in Cura in
any way. See the Aug 9 discussion earlier in this thread.

—
You are receiving this because you were mentioned.
<https://github.com/Ultimaker/Cura/pull/914#issuecomment-248665027>, or mute
the thread

Patola:
@BagelOrb I have a diamond extruder.
@sevaa Thanks. Any plans to support this in the near future? Will I be able to model something in Blender, export it to x3d, import it in Cura then have the different materials automatically be assigned to different extruders, or at least be selectable to do so?

sevaa:
According to Ghostkeeper@Aug 9, Cura doesn't support colors. Once it does, I'll gladly work on importing them along with the rest of the X3D.

Patola:
@sevaa I understand, but in the meantime wouldn't it be sensible to use one of the indexed features (colors or materials) and assign them differently to each extruder, if the printer has more than one extruder and the file has more than one color/material?

sevaa:
No way to test. Also, let them release the basic patch first, I'm sure I've missed something and there will be feedback.

awhiemstra:
We have no support for setting extruders/materials etc. from files for any format. So adding that would be a lot of work and out of scope of the initial X3D support.

Patola:
@awhiemstra I understand. What I'm asking for is not part of file importing. But I understand that Cura's workflow for dual (or triple, etc.) extrusion is way far from ideal. It would be much simpler if different materials could automatically assigned to different extruders right from the modeling stage. Should/can I open a new bug report asking for that workflow change?

fieldOfView:
@Patola: yes, a new feature request would be better than trying to repurpose this one.

A wise man once said "He who would learn to fly one day must first learn to stand and walk and run and climb and dance; one cannot fly into flying.". Import first, reading your brain to know how what material you want to print what model with later.

Patola:
Just for reference, I created #1014

PR #918, opened 2016-08-10
Update UI texts to be compliant to Ultimaker dictionary

fieldOfView:
A number of issues were found where the UI uses terms that do not comply to the terms set in the "Ultimaker dictionary". 

CURA-2053

### DIFF
--- a/plugins/PerObjectSettingsTool/__init__.py
+++ b/plugins/PerObjectSettingsTool/__init__.py
@@ -11,15 +11,15 @@ i18n_catalog = i18nCatalog("cura")
 def getMetaData():
     return {
         "plugin": {
-            "name": i18n_catalog.i18nc("@label", "Per Object Settings Tool"),
+            "name": i18n_catalog.i18nc("@label", "Per Model Settings Tool"),
             "author": "Ultimaker",
             "version": "1.0",
-            "description": i18n_catalog.i18nc("@info:whatsthis", "Provides the Per Object Settings."),
+            "description": i18n_catalog.i18nc("@info:whatsthis", "Provides the Per Model Settings."),
             "api": 3
         },
         "tool": {
-            "name": i18n_catalog.i18nc("@label", "Per Object Settings"),
-            "description": i18n_catalog.i18nc("@info:tooltip", "Configure Per Object Settings"),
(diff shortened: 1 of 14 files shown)


plugins/PerObjectSettingsTool/__init__.py
```diff
@@ -11,15 +11,15 @@
 def getMetaData():
     return {
         "plugin": {
-            "name": i18n_catalog.i18nc("@label", "Per Object Settings Tool"),
+            "name": i18n_catalog.i18nc("@label", "Per Model Settings Tool"),
```
nallath: Shouldn't we also change the entire plugin file name?

resources/definitions/fdmprinter.def.json
```diff
@@ -2109,8 +2109,8 @@
             {
                 "cool_fan_enabled":
                 {
-                    "label": "Enable Cooling Fans",
-                    "description": "Enables the cooling fans while printing. The fans improve print quality on layers with short layer times and bridging / overhangs.",
+                    "label": "Enable Print Cooling",
```
nallath: Dict sez; The text "Enable Cooling Fans" in Cura should be changed in "Enable Print Cooling Fans".

resources/qml/PrintMonitor.qml
```diff
@@ -44,7 +46,7 @@ Column
         delegate: Loader
         {
             sourceComponent: monitorItem
-            property string label: catalog.i18nc("@label", "Bed Temperature")
+            property string label: catalog.i18nc("@label", "Build plate")
```
nallath: Shouldn't it be build plate temperature?

fieldOfView: See UXD-80; "The text temperature (in Bed Temperature ) can be left out because the header is already called "Temperatures"."

plugins/PerObjectSettingsTool/__init__.py
fieldOfView: In this PR I tried to only change user-facing text.

Ghostkeeper: Please forgive me if I still type Per Object Settings Tool once in a while :wink:

PR #920, opened 2016-08-11
Adding "approximately equal" to the amount of material needed

thopiekar:
I think it makes sense to add this character here, because the amount needed is only a approximated value, and without the end-user might expect a more accurate value.

### DIFF
--- a/resources/qml/JobSpecs.qml
+++ b/resources/qml/JobSpecs.qml
@@ -209,7 +209,7 @@ Rectangle {
                         lengths = ["0.00"];
                         weights = ["0"];
                     }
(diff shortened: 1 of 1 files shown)


fieldOfView:
I would actually prefer "%1 m / ~%2 g"

The extruded length is fairly precise (give or take some slippage), the weight is much more of a guesstimate due to rounding, inaccurate material diameter and inaccurate density info. Also notice the single squiggle "approximately" vs your double squiggle "approximately equal to".

thopiekar:
Ok, makes sense.

I just read this part of the job specs like: <material_icon> ≈ (%1 m, %2 g)
Just as if the symbol was a variable in the user interface..

fieldOfView:
A "≈" only makes sense if you have two items to compare: A ≈ B.

PR #922, opened 2016-08-14
CURA-2079: Finish firmware upload process on errors

thopiekar:
Looking at the old code shows that the download process was never
completely finished. The according function was only return'ed, but no
signal was emitted to the window to give the user the possibility to
close it again.

Optionally, it would be useful here to pass messages to the window
dialog to inform the user about problems.

However, this code is untested because:
1. I don't know where put my printers *.hex-file, when coding here with
my editor,
2. Have no idea how to reproduce the IO errors reported at JIRA, so
fingers crossed here.

Contributes to CURA-2079

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -182,7 +182,8 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
 
         if len(hex_file) == 0:
             Logger.log("e", "Unable to read provided hex file. Could not update firmware")
-            return 
+            self._updateFirmware_completed()
+            return
 
         programmer = stk500v2.Stk500v2()
         programmer.progressCallback = self.setProgress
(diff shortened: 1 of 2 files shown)


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -227,8 +235,8 @@ def updateFirmware(self, file_name):
     def firmwareUpdateFinished(self):
         return self._firmware_update_finished
 
-    def resetFirmwareUpdateFinished(self):
-        self._firmware_update_finished = False
+    def resetFirmwareUpdate(self, update_has_finished = False):
```
fieldOfView: Is this called from anywhere other than from _updateFirmware_completed above? If not, why is it a public method, and what is the use of the argument?

thopiekar: No problem. To be honest I just saw this code while working on _updateFirmware and decided to make it more common.

PR #924, opened 2016-08-15
Machine Settings plugin

fieldOfView:
This PR adds a machine action to allow editing certain settings from the machine_* category (nozzle width, start/end gcode, etc). Changes to these settings are stored in variants.

At this point it does not support printers that use variants (UM2+) or printers that have multiple extruders (UM3).

The PR also includes a "custom" printer that has the Machine Settings plugin as a "first run" action, providing an easy way to set up a printer that has no official profile.


plugins/MachineSettingsAction/MachineSettingsAction.py
```diff
@@ -0,0 +1,87 @@
...
+        new_variant.addMetaDataEntry("type", "variant")
+        new_variant.setDefinition(global_container_stack.getBottom())
+        UM.Settings.ContainerRegistry.getInstance().addContainer(new_variant)
+        global_container_stack.replaceContainer(variant_index, new_variant)
+
+    def _onContainerAdded(self, container):
+        # Add this action as a supported action to all machine definitions
+        if isinstance(container, UM.Settings.DefinitionContainer) and container.getMetaDataEntry("type") == "machine":
+            if container.getProperty("machine_extruder_count", "value") > 1:
+                # Multiextruder printers are not currently supported
```
nallath: A spot of logging to indicate that a user tried this is probably a good idea.

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -0,0 +1,475 @@
...
+                            {
+                                machineEndGcodeProvider.setPropertyValue("value", machineEndGcodeField.text)
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
```
nallath: Isn't it better to put the logic for the head poly in the machine action .py?

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -0,0 +1,475 @@
+// Copyright (c) 2016 Ultimaker B.V.
```
nallath: Copyright here is different from the other bits. Is this copyrighted by UM or by you?

plugins/MachineSettingsAction/MachineSettingsAction.py
fieldOfView: The user doesn't necessarily try anything here.
A spot of debug logging doesn't hurt though.

plugins/MachineSettingsAction/MachineSettingsAction.qml
fieldOfView: I wanted to use the SettingPropertyProvider as much as possible.

cura/BuildVolume.py
```diff
@@ -240,9 +240,11 @@ def _onGlobalContainerStackChanged(self):
             self._active_container_stack.propertyChanged.connect(self._onSettingPropertyChanged)
 
             self._width = self._active_container_stack.getProperty("machine_width", "value")
+            machine_height = self._active_container_stack.getProperty("machine_height", "value")
             if self._active_container_stack.getProperty("print_sequence", "value") == "one_at_a_time":
-                self._height = self._active_container_stack.getProperty("gantry_height", "value")
-                self._buildVolumeMessage()
+                self._height = min(self._active_container_stack.getProperty("gantry_height", "value"), machine_height)
+                if(self._height < machine_height):
```
nallath: the () are not needed.

cura/BuildVolume.py
```diff
@@ -258,9 +260,11 @@ def _onSettingPropertyChanged(self, setting_key, property_name):
 
         rebuild_me = False
         if setting_key == "print_sequence":
+            machine_height = self._active_container_stack.getProperty("machine_height", "value")
             if Application.getInstance().getGlobalContainerStack().getProperty("print_sequence", "value") == "one_at_a_time":
-                self._height = self._active_container_stack.getProperty("gantry_height", "value")
-                self._buildVolumeMessage()
+                self._height = min(self._active_container_stack.getProperty("gantry_height", "value"), machine_height)
+                if(self._height < machine_height):
```
nallath: Same here. () not needed.

cura/BuildVolume.py
fieldOfView: doh, too much javascript!

PR #926, opened 2016-08-16
[CURA-2079] Adding detailed error messages to firmware updater dialog

thopiekar:
Just read https://github.com/Ultimaker/Cura/commit/0435706736b165cbb83b9a58c09dbc5ce2b21f06
and take a look at the code.

Short: Added different error messages, which appear whenever a exception was catched.

### DIFF
--- a/plugins/USBPrinting/FirmwareUpdateWindow.qml
+++ b/plugins/USBPrinting/FirmwareUpdateWindow.qml
@@ -37,6 +37,26 @@ UM.Dialog
                     //: Firmware update status label
                     return catalog.i18nc("@label","Firmware update completed.")
                 }
+                else if (manager.progress == -4)
+                {
+                    //: Firmware update status label
+                    return catalog.i18nc("@label","Firmware update failed due to missing firmware.")
+                }
+                else if (manager.progress == -3)
+                {
+                    //: Firmware update status label
+                    return catalog.i18nc("@label","Firmware update failed due to an input/output error.")
+                }
+                else if (manager.progress == -2)
+                {
+                    //: Firmware update status label
+                    return catalog.i18nc("@label","Firmware update failed due to an communication error.")
+                }
+                else if (manager.progress == -1)
+                {
+                    //: Firmware update status label
+                    return catalog.i18nc("@label","Firmware update failed due to an unknown error.")
+                }
                 else if (manager.progress == 0)
                 {
                     //: Firmware update status label
--- a/plugins/USBPrinting/FirmwareUpdateWindow.qml
+++ b/plugins/USBPrinting/FirmwareUpdateWindow.qml
@@ -32,7 +32,7 @@ UM.Dialog
             }
 
             text: {
-                if (manager.firmwareUpdateCompleteStatus)
+                if (manager.firmwareUpdateCompleteStatus && !manager.progress > 0)
                 {
                     //: Firmware update status label
                     return catalog.i18nc("@label","Firmware update completed.")
--- a/plugins/USBPrinting/FirmwareUpdateWindow.qml
+++ b/plugins/USBPrinting/FirmwareUpdateWindow.qml
@@ -32,40 +32,51 @@ UM.Dialog
             }
 
             text: {
-                if (manager.firmwareUpdateCompleteStatus && !manager.progress > 0)
+                if (manager.errorCode == 0)
                 {
-                    //: Firmware update status label
-                    return catalog.i18nc("@label","Firmware update completed.")
-                }
-                else if (manager.progress == -4)
+                    if (manager.firmwareUpdateCompleteStatus)
+                	{
+	                    //: Firmware update status label
+    	                return catalog.i18nc("@label","Firmware update completed.")
+        	        }
+        	        if (manager.progress == 0)
+                	{
+	                    //: Firmware update status label
+    	                return catalog.i18nc("@label","Starting firmware update, this may take a while.")
+                	}
+                	else
+                	{
+                    	//: Firmware update status label
+                    	return catalog.i18nc("@label","Updating firmware.")
+                    }
+        	    }
+                else 
                 {
-                    //: Firmware update status label
-                    return catalog.i18nc("@label","Firmware update failed due to missing firmware.")
-                }
-                else if (manager.progress == -3)
-                {
-                    //: Firmware update status label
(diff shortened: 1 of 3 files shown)


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -198,26 +198,53 @@ def _updateFirmware(self):
...
+    ##  Private function which makes sure that firmware update process has failed by an unknown error
+    def _updateFirmware_completed_unknown(self):
+        return self._updateFirmware_completed_common(progress = -1)
+    
+    ##  Private function which makes sure that firmware update process has sucessfully completed/ended
+    def _updateFirmware_completed_sucessfully(self):
+        return self._updateFirmware_completed_common(progress = 100)
+    
+    ##  Private common function which makes sure that firmware update process has completed/ended with a set progress state
+    def _updateFirmware_completed_common(self, progress, max_progress = 100):
```
fieldOfView: Instead of masquerading an errorcode in progress it would be better to have an actual error argument here, and a @pyqtProperty named errorCode.

thopiekar:
@fieldOfView, do you see a reason why my errorCode does not get updated?
It is accessible in the Python code, but doing changes to it does not do any changes to the QML dialog :disappointed:

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, parent = None):
         self._update_thread = threading.Thread(target = self._updateThread)
         self._update_thread.setDaemon(True)
 
+        self._errorCode = 0
```
fieldOfView: codestyle: self._error_code

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -173,6 +173,7 @@ def connect(self):
 
     ##  Private function (threaded) that actually uploads the firmware.
     def _updateFirmware(self):
+        self._errorCode = 0
```
fieldOfView: codestyle: self._error_code

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -57,6 +59,10 @@ def progress(self):
             progress += device.progress
         return progress / len(self._usb_output_devices)
 
+    @pyqtProperty(int, notify = progressChanged)
+    def errorCode(self):
+        return self._errorCode
```
fieldOfView: This never gets set to anything but 0.

plugins/USBPrinting/FirmwareUpdateWindow.qml
```diff
@@ -32,20 +32,51 @@ UM.Dialog
...
-                else if (manager.progress == 0)
-                {
-                    //: Firmware update status label
-                    return catalog.i18nc("@label","Starting firmware update, this may take a while.")
-                }
-                else
+                    if (manager.firmwareUpdateCompleteStatus)
+                	{
+	                    //: Firmware update status label
+    	                return catalog.i18nc("@label","Firmware update completed.")
```
fieldOfView: Indentation is weird

plugins/USBPrinting/FirmwareUpdateWindow.qml
```diff
@@ -32,20 +32,51 @@ UM.Dialog
...
+                	else if (manager.errorCode == 4)
+                	{
+	                    //: Firmware update status label
+                    	return catalog.i18nc("@label","Firmware update failed due to missing firmware.")
+                	}
+                	else
+                	{
+                    	//: Firmware update status label
+                    	return catalog.i18nc("@label", "Unknown error code: %1").arg(manager.errorCode)
+                    }
```
fieldOfView: Javascript has switch/case expression to make this more readable
(proper indentation helps as well)

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -182,7 +183,7 @@ def _updateFirmware(self):
 
         if len(hex_file) == 0:
             Logger.log("e", "Unable to read provided hex file. Could not update firmware")
-            self._updateFirmware_completed()
+            self._updateFirmware_completed_missing_firmware()
```
fieldOfView: Sorry for not catching this before. The codestyle for private method names in python is _lowerCamelCase().

fieldOfView:
As far als I can see, the _errorCode (should be _error_code) in USBPrinterOutputDeviceManager.py never gets set, only the one in USBPrinterOutputDevice.py is set.

plugins/USBPrinting/FirmwareUpdateWindow.qml
thopiekar: Used the regular text editor of Eclipse for QML files. Possibly it just enabled tabs again. However, I'll fix that.

plugins/USBPrinting/FirmwareUpdateWindow.qml
thopiekar: Ok, thank you for the hint.

thopiekar:
@fieldOfView: Yes, that's what I think, too. While looking at functions like "setProgress", etc., I thought, that both instances get inherited at the end. So changing self._error_code in USBPrinterOutputDevice should also change the one in USBPrinterOutputDeviceManager.
Do you see a way how to pass the error code correctly to USBPrinterOutputDeviceManager? :confused:

thopiekar:
AAHH!!! Probably got it!

(Will give it a try later. Have to learn again for my exams here.)

plugins/USBPrinting/FirmwareUpdateWindow.qml
thopiekar: Yes, a lot of tabulators. Removed them..

plugins/USBPrinting/FirmwareUpdateWindow.qml
thopiekar: Done

thopiekar:
Yeah :dancer: 

spectacle h11989

fieldOfView:
Works as expected now. Tip for future reference: instead of creating a number of method for error-states, you could also have introduced an enum with descriptive names and a single method.

PR #929, opened 2016-08-17
Update Innovo Inventor JSON and STL files

innovo3d:
Made some changes to the Inventor JSON file:
- made machine center is zero true
- added turn heated bed off to end g code
- changed top/bottom thickness from 0.30mm to 1.2mm (don’t know why I
put it to 0.3 in the first place!!)
- changed material bed temp to 0 by default

As for the STL file, it’s been updated to a much smaller file, 810kb
(down from 13mb).  I thought I updated this before but apparently not.

I hope this update can be merged soon, because I plan on launching a
makerspace with these printers and everyone will be using Cura so they
need these settings

### DIFF
--- a/resources/machines/innovo-inventor.json
+++ b/resources/machines/innovo-inventor.json
@@ -13,7 +13,7 @@
         "machine_height": {"default": 290},
         "machine_depth": {"default": 300},
         "machine_heated_bed": { "default": true},
-        "machine_center_is_zero": {"default": false},
+        "machine_center_is_zero": {"default": true},
         "machine_nozzle_size": {"default": 0.4},
         "machine_head_shape_min_x": {"default": 43.7},
         "machine_head_shape_min_y": {"default": 19.2},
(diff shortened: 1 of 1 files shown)


nallath:
Note; I don't expect that we will release 2.1.4. I'm willing to merge this (the changed look good), but you're better off converting it to 2.2, which will be released soon-ish.

Ghostkeeper:
This has already been converted for 2.2: a8109f6267461a9b985957c275ec63ac2c1bcf99

nallath:
Fair enough. Then this can be merged.

fieldOfView:
So it is merged, but will not be released? Er... ok...

nallath:
I don't expect that we will release another one. Doesn't mean it won't happen. We also never expected 15.04 to be supported as long as it was, but still it happened. 

Besides, the work is already done, it doesn't cost us anything to merge it.

PR #930, opened 2016-08-17
Update def.json and STL file

innovo3d:
Updated JSON file:
- changed machine center is zero to true
- added turn heated bed off to end g code
- changed top/bottom thickness from 0.3 to 1.2mm
- added machine preferences as I did in 2.1 (is this still used?)

Updated STL file.  The last file was 13mb! this one is 810kb!


Ghostkeeper:
This PR went fine. Good job!

PR #943, opened 2016-08-23
[CURA-2007] resolved settings

jackha:
Example: bed temperature.
There is only one bed temperature, but multiple extruders with multiple materials can prefer their own bed temperatures. A resolvement has been added to "merge" those settings. 
Actual bed temperature can now be averaged from extruder settings.


cura/CuraApplication.py
```diff
@@ -925,3 +926,8 @@ def addAdditionalComponent(self, area_id, component):
         self._additional_components[area_id].append(component)
 
         self.additionalComponentsChanged.emit(area_id)
+
+    @pyqtSlot(str)
```
nallath: This is pretty nasty debug code. If you want to use this more often, i'd suggest moving the import out of the function.

resources/qml/Settings/SettingTextField.qml
```diff
@@ -106,7 +106,16 @@ SettingItem
             {
                 target: input
                 property: "text"
-                value: propertyProvider.properties.value
+                value:  {
+                    if ((propertyProvider.properties.resolve != "None") && (stackLevel != 0)) {
```
nallath: So only if the user-setting is not set we return the resolve? So if the user sets the setting and doesn't save it, the resolve is not used. When the user sets the setting and makes a quality change out of it, the resolve is used?

resources/qml/Settings/SettingView.qml
```diff
@@ -124,7 +125,7 @@ ScrollView
                 id: inheritStackProvider
                 containerStackId: Cura.MachineManager.activeMachineId
                 key: model.key
-                watchedProperties: [ "global_inherits_stack" ]
+                watchedProperties: [ "global_inherits_stack", "value" ]
```
nallath: Why add value here? We already have a property provider that is looking at value. There is a pretty good reason (as documented above) to have a dedicated global_inherits_stack provider.

cura/CuraApplication.py
jackha: Ah yes forgot to move the import out of the function. I need this function almost every time I debug qml's :-) (on my computer console.log disappears...)

resources/qml/Settings/SettingView.qml
jackha: Old leftover... removed it

PR #955, opened 2016-08-29
Fixed duplicate material. CURA-1969

jackha:


plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -89,15 +89,15 @@ def setName(self, new_name):
...
-    def setProperty(self, key, property_name, property_value, container = None):
-        if self.isReadOnly():
-            return
-
-        super().setProperty(key, property_name, property_value)
-
-        basefile = self.getMetaDataEntry("base_file", self._id)  #if basefile is none, this is a basefile.
-        for container in UM.Settings.ContainerRegistry.getInstance().findInstanceContainers(base_file = basefile):
-            container._dirty = True
+    # def setProperty(self, key, property_name, property_value, container = None):
```
jackha: What kind of impact does this have? I don't see any negative effects..

nallath: It pretty much means that if we set the property of one of the containers that is created from a single file, not all of them get updated. I believe they are intended to work that way, so i'd not remove it lightly.

PR #958, opened 2016-08-31
update Printrbot Simple definition

weevilgenius:
* Add platform mesh
* Default nozzle size is 0.4mm
* Updated start gcode to correct bed origin after auto-level


nallath:
I've also cherry-picked your changes into our release branch.

PR #967, opened 2016-09-03
[CURA-1445] Adding hashes

thopiekar:
... for every model placed on the build plate.

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -84,14 +84,13 @@ class SliceInfo(Extension):
                 return # Do nothing, user does not want to send data
 
             # Listing all files placed on the buildplate
-            filenames = []
+            modelhashes = []
             for node in DepthFirstIterator(CuraApplication.getInstance().getController().getScene().getRoot()):
                 if type(node) is not SceneNode or not node.getMeshData():
-                    continue    
-                filenames.append(node.getMeshData().getFileName())
+                    continue
+                modelhashes.append(node.getMeshData().getHash())
 
             # Creating md5sums and formatting them as discussed on JIRA
-            modelhashes = [hashlib.md5(open(filename, 'rb').read()).hexdigest() for filename in filenames]            
             modelhash_formatted = ""
             if modelhashes:
                 modelhash_formatted = str(modelhashes[0])
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -107,27 +107,6 @@ class SliceInfo(Extension):
             # TODO: Send material per extruder instead of mashing it on a pile
             material_used = math.pi * material_radius * material_radius * sum(print_information.materialLengths) #Volume of all materials used
 
-            # Get model information (bounding boxes, hashes and transformation matrix)
-            models_info = []
-            for node in DepthFirstIterator(Application.getInstance().getController().getScene().getRoot()):
-                if type(node) is SceneNode and node.getMeshData() and node.getMeshData().getVertices() is not None:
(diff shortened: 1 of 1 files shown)


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -80,6 +83,22 @@ def _onWriteStarted(self, output_device):
...
+                if type(node) is not SceneNode or not node.getMeshData():
+                    continue    
+                filenames.append(node.getMeshData().getFileName())
+
+            # Creating md5sums and formatting them as discussed on JIRA
+            modelhashes = [hashlib.md5(open(filename, 'rb').read()).hexdigest() for filename in filenames]            
+            modelhash_formatted = ""
+            if modelhashes:
+                modelhash_formatted = str(modelhashes[0])
+                if len(modelhashes) > 1:
```
nallath: You don't need to check for length. If there is only one entry, the for loop won't be triggered in the first place.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -80,6 +83,22 @@ def _onWriteStarted(self, output_device):
...
 
+            # Listing all files placed on the buildplate
+            filenames = []
+            for node in DepthFirstIterator(CuraApplication.getInstance().getController().getScene().getRoot()):
+                if type(node) is not SceneNode or not node.getMeshData():
+                    continue    
+                filenames.append(node.getMeshData().getFileName())
+
+            # Creating md5sums and formatting them as discussed on JIRA
+            modelhashes = [hashlib.md5(open(filename, 'rb').read()).hexdigest() for filename in filenames]
```
nallath: It shouldn't make a hash out of the filename, but a hash out of the model data. Otherwise renaming a file will generate a new hash.

fieldOfView: I think the hash will need to be made when importing the model(s), because otherwise even positioning the model on the buildplate will generate a new hash.

thopiekar:
Don't ask me where this commit message is from: https://github.com/Ultimaker/Cura/pull/967/commits/dd02243790b1952c21ec4c95c7ac84d7625ae593
I just wanted to commit my changes, the message appeared "from the dark" and even when I aborted the commit was applied like this to my branch.
If you want I can try to rebase.

Back to the topic: Now the code uses UM's own hashing functions.

plugins/SliceInfoPlugin/SliceInfo.py
Ghostkeeper: It would get triggered, but only once, right?

In any case, it's easier to just do `modelhash_formatted = ",".join(modelhashes)` afterwards.

Ghostkeeper:
Looks like some merge was still in progress there, that the commit message file was still storing that commit message? Or it failed to reset the file afterwards.

thopiekar:
Should I recommit it? Or can I leave it like that?

fieldOfView:
I would not worry about the commit message too much, but have a look at @Ghostkeeper's suggestion here: https://github.com/Ultimaker/Cura/pull/967/commits/6688b99caee09d3ae8e102bd31a8ddcf72d3b0b2#r77596102

thopiekar:
@fieldOfView Many thanks! Forgot about it :eyes:

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Nope, this code will be executed, when saving G-code or clicking on "Print via USB", so not that often.

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: This is great! \o/

PR #968, opened 2016-09-03
CuraEngineBackend: Cleanup engine lookup

thopiekar:
Just noticed that my engine at /usr/bin/ couldn't be found correctly
again.
Made the test at line 62 correct by checking for it's existence and
renamed some variables, which were confusing (for me).

### DIFF
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -44,27 +44,28 @@ class CuraEngineBackend(Backend):
         executable_name = "CuraEngine"
         if Platform.isWindows():
             executable_name += ".exe"
-        default_engine_location = executable_name
+
+        engine_location = executable_name
         if os.path.exists(os.path.join(Application.getInstallPrefix(), "bin", executable_name)):
-            default_engine_location = os.path.join(Application.getInstallPrefix(), "bin", executable_name)
+            engine_location = os.path.join(Application.getInstallPrefix(), "bin", executable_name)
         if hasattr(sys, "frozen"):
-            default_engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), executable_name)
-        if Platform.isLinux() and not default_engine_location:
+            engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), executable_name)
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -44,27 +44,28 @@ def __init__(self):
...
         if hasattr(sys, "frozen"):
-            default_engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), executable_name)
-        if Platform.isLinux() and not default_engine_location:
+            engine_location = os.path.join(os.path.dirname(os.path.abspath(sys.executable)), executable_name)
+        if Platform.isLinux() and not os.path.exists(engine_location):
             if not os.getenv("PATH"):
                 raise OSError("There is something wrong with your Linux installation.")
             for pathdir in os.getenv("PATH").split(os.pathsep):
-                execpath = os.path.join(pathdir, executable_name)
+                execpath = os.path.join(pathdir, engine_location)
```
Ghostkeeper: Did you mean to change this one? It's not the renamed variable. This can have repercussions since the `engine_location` variable changes throughout multiple iterations of the for loop.

thopiekar: Will take another look on this later.
Btw. from what I know about registering keys into our settings storage...
What happens if the location changes?
E.g. someone switched from my PPA to our bundled installation, would the
code take another look and check for the new location?
I don't think so 🤔

Mit freundlichen Grüßen/With kind regards
- Thomas Karl Pietrowski

Ghostkeeper: Well not *during* the run of the program. I think that taking that into account would be somewhat over-engineered.

I guess that after the new run, it would take whatever CuraEngine executable comes first on the PATH.

thopiekar: Ok, then it should be fine. I wouldn't care about changes during runtime,
too. (This is not how Cura should be used at all.)

ChrisTerBeke:
This PR is outdated and there is no agreement on a proper implementation. Closing for now.

thopiekar:
Also not convinced of my own work..

PR #977, opened 2016-09-07
i18n: de: Adding translated comment

thopiekar:
Adding german translation primary as example for CURA-570.


thopiekar:
As nobody seems to care...

Ghostkeeper:
Sorry, slipped through the cracks! :( The addition is fine.

PR #994, opened 2016-09-15
Speed up switching between multiextrusion machines

fieldOfView:
ExtruderManager.addMachineExtruders is supposed to create a cache, but this cache was always rebuilt because of an indentation error.

This PR does a little refactoring on behavior that had grown because of the error; ExtrudersModel should not respond to the cache being rebuilt, but instead it should respond to when the global containerstack changes to a definition with different extruders.

I came across this when trying to figure out why the ExtrudersModel did not reset when selecting a single extrusion printer after selecting a multiextrusion printer. That is because for single extrusion printers the code was correctly using the cache, which resulted in not updating the ExtrudersModel.

This is in a PR because there is a non-negligible regression hazard, and there is no issue for it.

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -13,8 +13,12 @@ import UM.Settings.SettingFunction
 #
 #   This keeps a list of extruder stacks for each machine.
 class ExtruderManager(QObject):
-    ##  Signal to notify other components when the list of extruders changes.
+    ##  Signal to notify other components when the list of extruders for a machine definition changes.
     extrudersChanged = pyqtSignal(QVariant)
+
+    ## Signal to notify other components when the global container stack is switched to a definition
+    #  that has different extruders than the previous global container stack
+    globalContainerStackDefinitionChanged = pyqtSignal()
 
     ##  Notify when the user switches the currently active extruder.
     activeExtruderChanged = pyqtSignal()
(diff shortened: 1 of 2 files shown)


cura/Settings/ExtruderManager.py
```diff
@@ -13,9 +13,13 @@
 #
 #   This keeps a list of extruder stacks for each machine.
 class ExtruderManager(QObject):
-    ##  Signal to notify other components when the list of extruders changes.
+    ##  Signal to notify other components when the list of extruders for a machine definition changes.
     extrudersChanged = pyqtSignal(QVariant)
 
+    ## Signal to notify other components when the global container stack is switched to a definition
+    #  that has different extruders than the previous global container stack
+    globalContainerStackDefinition = pyqtSignal()
```
awhiemstra: I prefer to call signals that trigger when something changes "somethingChanged", so in this case "extruderDefinitionsChanged" or something similar.

cura/Settings/ExtruderManager.py
```diff
@@ -25,6 +29,7 @@ def __init__(self, parent = None):
         self._extruder_trains = { } #Per machine, a dictionary of extruder container stack IDs.
         self._active_extruder_index = 0
         UM.Application.getInstance().globalContainerStackChanged.connect(self.__globalContainerStackChanged)
+        self._global_container_stack_definition = None
```
awhiemstra: Since this is used to store the ID of the definition and not the actual DefinitionContainer using "_global_container_stack_definition_id" is clearer.

cura/Settings/ExtruderManager.py
fieldOfView: I quite agree. I don't know why I named it like this.

cura/Settings/ExtruderManager.py
fieldOfView: Agreed

fieldOfView:
I have fixed the commit, changing the names, and rebased it to the current master. The rebasing seems to have thrown off the review.

awhiemstra:
Looks good. Feel free to merge this once today's craziness is over.

PR #995, opened 2016-09-15
Render Layerview ghosts behind layers

fieldOfView:
This is a follow-up to https://github.com/Ultimaker/Cura/pull/906

I'm trying to rendeer the ghosts behind the layerview. I am currently stumped why the layers RenderPass seems to be affected by the underlying renderpass.


awhiemstra:
The rendering issue was due to a bug in Uranium's RenderPass which did not properly clear the framebuffer. This has now been fixed.

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,270 @@
...
+
+                if self._current_layer_mesh:
+                    batch_has_items = True
+                    batch.addItem(node.getWorldTransformation(), self._current_layer_mesh)
+
+                if self._current_layer_jumps:
+                    batch_has_items = True
+                    batch.addItem(node.getWorldTransformation(), self._current_layer_jumps)
+
+                if batch_has_items:
```
awhiemstra: Rather than using multiple booleans to track this, why not check that batch.items contains anything? Seems simpler to me. :)

plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,270 @@
...
+
+        self._shader = None
+        self._tool_handle_shader = None
+        self._gl = OpenGL.getInstance().getBindingsObject()
+        self._scene = self._controller.getScene()
+
+        self._scene.getRoot().childrenChanged.connect(self._onSceneChanged)
+        self._max_layers = 0
+        self._current_layer_num = 0
+        self._current_layer_mesh = None
```
awhiemstra: I personally think it is nicer to keep the logic for the mesh creation, maximum layer count etc. in LayerView and make the render pass just deal with rendering. Maybe even create a separate class for it, since there is something to be said about it not being the responsibility of LayerView either.

plugins/LayerView/layerview_composite.shader
```diff
@@ -0,0 +1,82 @@
...
+    {
+        gl_Position = u_modelViewProjectionMatrix * a_vertex;
+        v_uvs = a_uvs;
+    }
+
+fragment =
+    uniform sampler2D u_layer0;
+    uniform sampler2D u_layer1;
+    uniform sampler2D u_layer2;
+    uniform sampler2D u_layer3;
```
awhiemstra: This uniform does not seem to be used.

plugins/LayerView/layerview_composite.shader
```diff
@@ -0,0 +1,82 @@
...
+
+    uniform float u_imageWidth;
```
awhiemstra: Neither are these two. I guess they are also in the original composite shader and should be fixed there as well.

fieldOfView:
Unfortunately I have to close this PR (again) because it is against 2.3

PR #1002, opened 2016-09-16
[CURA-1445] SliceInfo: Using HTTPS again

thopiekar:
The problem seems to be related to our HTTP(S) server.
It is expected that the problem will be solved after the weekend.

Contributes to CURA-1445


fieldOfView:
I think you meant to commit a bit more...

thopiekar:
Nope, it is a server issue as far as I analysed with Jordy @ YM. There is no HTTPS request the YM server answers to.
All the SliceInfo (only <= 14.x) doesn't get into the server. So respecting privacy I just made this PR.
On Monday Wilco should care about it..

fieldOfView:
Then what's the use of this PR? All it does is add a newline at the end of a file.

thopiekar:
Oh, looks like Jack only applied his HTTP workaround only on master.
Well, then you don't see any changes here. I changed the URL on master and cherry-picked it on 2.3, because I thought it will be needed here, too..

thopiekar:
So you can git merge or just close it. Feel free..

PR #1003, opened 2016-09-16
Added kossel platform based on jrocholl (GPLv3)

Patola:
Added a stl file for the kossel heatbed. Created from jrocholl's original STL files for kossel plastic parts (all GPLv3) and the public domain geometry from the mk_visual_calc repository.


thopiekar:
You'll also need to set the filename in your definition. That's done eg. for UM2+ at:
https://github.com/Ultimaker/Cura/blob/master/resources/definitions/ultimaker2_plus.def.json#L12

Patola:
I think it is ok now. I also corrected the STL since it was shifted up on the Z axis. I have tested my files for both visualization and printing and they work correctly on my kossel. I still have some issues... the square checkerboard pattern kind of breaks the visualization, on the other hand it correctly lightens up the part where the print is permitted; although it is a circle, I approximated with 8 "disallowed" areas around the square. There is some "border" on the square checkerboard, though, that I did not configure as disallowed but still does not allow objects to be placed. But in general, this kossel table and STL seem to work well. Can someone test and commit it to Cura please?

thopiekar:
@Patola, but I guess https://github.com/Ultimaker/Cura/pull/1003/commits/ab46312ed7405cc2e673e46b7a25dc19fc4e8c54 isn't needed here on Ultimaker/Cura, is it?

Patola:
I am a disaster at github. I made that commit thinking it was going to my own fork of Cura. My apologies... How do I cancel this particular commit? I am not finding the way to do that.

thopiekar:
No problem. You can get a rid of the two commits using the following instructions:
http://stackoverflow.com/questions/927358/how-to-undo-last-commits-in-git
Remember: They talk about 1 commit, but as you reverted the one you need to undo two commit.
At the end you won't see both commits :wink:

Patola:
I think it worked. I did undo 2, push rebase, and now the README.md commit and revert to not appear anymore. Thanks again, Thomas!

thopiekar:
No problem and well done, @Patola :+1:

Ghostkeeper:
The border appears because of brim, skirt or raft. Making the brim, skirt or raft smaller will reduce the border as well.

Ghostkeeper:
Seems to work fine. The z-fighting with the build plate could be solved by moving the mesh 1mm lower or so, but it's fine for now. If you wish to do that, you could make another PR.

AGPL is incompatible with GPLv3, but since it is a separate file, we are allowed to publish the GPLv3 file alongside the code.

PR #1007, opened 2016-09-19
Render Layerview ghosts behind layers

fieldOfView:
This is hopefully the final iteration of a PR to render layerview ghosts for all objects on the buildplate behind the rendered layers.

### DIFF
--- a/plugins/LayerView/LayerPass.py
+++ b/plugins/LayerView/LayerPass.py
@@ -1,118 +1,30 @@
 # Copyright (c) 2016 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
-from UM.Logger import Logger
-from UM.Application import Application
-from UM.Resources import Resources
-from UM.Preferences import Preferences
-from UM.Signal import Signal
-from UM.Job import Job
-
 from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
+from UM.Resources import Resources
 from UM.Scene.SceneNode import SceneNode
 from UM.Scene.ToolHandle import ToolHandle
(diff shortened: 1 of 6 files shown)


plugins/LayerView/LayerPass.py
```diff
@@ -0,0 +1,79 @@
...
+        self.bind()
+
+        tool_handle_batch = RenderBatch(self._tool_handle_shader, type = RenderBatch.RenderType.Overlay)
+
+        for node in DepthFirstIterator(self._scene.getRoot()):
+            if isinstance(node, ToolHandle):
+                tool_handle_batch.addItem(node.getWorldTransformation(), mesh = node.getSolidMesh())
+
+            elif isinstance(node, SceneNode) and node.getMeshData() and node.isVisible():
+                layer_data = node.callDecoration("getLayerData")
```
nallath: Only one node wil ever have layerdata right? So if we actually find the layer data, we can stop checking the other nodes. Won't be that much faster, though.

plugins/LayerView/LayerView.py
```diff
@@ -91,7 +110,7 @@ def beginRendering(self):
 
         if not self._ghost_shader:
             self._ghost_shader = OpenGL.getInstance().createShaderProgram(Resources.getPath(Resources.Shaders, "color.shader"))
-            self._ghost_shader.setUniformValue("u_color", Color(0, 0, 0, 64))
+            self._ghost_shader.setUniformValue("u_color", Color(32, 32, 32, 96))
```
nallath: We might want to consider finding a way to get these kind of colors out of the hardcoded spectrum (buildplate also has them). For now it's not a really big issue.

PR #1013, opened 2016-09-22
SliceInfo: Defining container_serialized as empty, before trying to s…

thopiekar:
…erialize

While discussing with Jordy about the code I just found a problematic case where sending slice info could fail.
(This shouldn't happen, but..) If the first container is not serializable, then the for-loop will crash because container_serialized is not set for the if clause.
Additionally the if-clause is at the moment useless, because it is always true, because the content of container_serialized gets never resetted.

Contributes to CURA-1445

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -119,6 +119,7 @@ class SliceInfo(Extension):
             }
             for container in global_container_stack.getContainers():
                 container_id = container.getId()
+                container_serialized = ""
                 try:
                     container_serialized = container.serialize()
                 except NotImplementedError:


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -119,6 +119,7 @@ def _onWriteStarted(self, output_device):
             }
             for container in global_container_stack.getContainers():
                 container_id = container.getId()
+                container_serialized = ""
```
Ghostkeeper: I don't see why this is necessary? The `if` clause below can only be reached if the `try` did not throw an exception, and if the `try` did not throw an exception, `container_serialized` will always be overwritten by the result of `container.serialize()`. Q.E.D.?

thopiekar: Additionally when serializing fails `container_serialized` is always filled with old data, but the container_id is always new. So at the end you'll set wrong data into a new container ID (keyword of the dict).

Ghostkeeper: The for loop starts (line 120). The first container is not serialisable, so you hit the `Except` in line 125, and consequently the `continue`, and you start over at line 120. In that loop, `container_serialized` is indeed undefined but you never hit the check against it.

For the second container, it tries again, but this time it's serialisable so line 124 successfully executes, `container_serialized` is defined, and at line 129, the `if` statement becomes `True` (as long as the resulting serialisation is not empty).

I think I'm missing something here then? It's not terribly important, of course...

thopiekar:
Yes, but shouldn't throw `if container_serialized` an error which says that `container_serialized` is undefined?
I'm a bit confused now.. or am I overlooking something here?
And for example, what if the 3 container isn't serializable again? It won't update `container_serialized` and fill it to the dict with the id of the new (third) container... 🤔

plugins/SliceInfoPlugin/SliceInfo.py
thopiekar: Yes, but shouldn't the code throw at `if container_serialized` an error which says that `container_serialized` is undefined?
I'm a bit confused now.. or am I overlooking something here?
And for example, what if the 3 container isn't serializable again? It won't update `container_serialized` and fill it to the dict with the id of the new (third) container... 🤔

thopiekar: @Ghostkeeper: You are so right! For a reason I misunderstood the `continue` here.
Will remove that line in the new PR.

Ghostkeeper: Glad it was cleared up. We've put this on through component test in the meanwhile :smiley:

PR #1015, opened 2016-09-22
[CURA-1445] Sending bundle of serialized data

thopiekar:
After an discussion with the developers at YM, we decided to send all data in one keyword.
So now all data is bundled into a JSON dictionary, where "" is the global_container_stack and it's containers go into the same JSON, too, by using their IDs as keywords.
More on JIRA...

Contributes to CURA-1445

Includes/Fixes #1013

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -109,7 +109,6 @@ class SliceInfo(Extension):
             containers = { "": global_container_stack.serialize() }
             for container in global_container_stack.getContainers():
                 container_id = container.getId()
-                container_serialized = ""
                 try:
                     container_serialized = container.serialize()
                 except NotImplementedError:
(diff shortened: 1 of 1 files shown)


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -105,30 +106,32 @@ def _onWriteStarted(self, output_device):
...
+                else:
+                    Logger.log("i", "No data found in %s to be serialized!", container_id)
+
             # Bundle the collected data
             submitted_data = {
                 "processor": platform.processor(),
                 "machine": platform.machine(),
                 "platform": platform.platform(),
-                "settings": global_container_stack.serialize(), # global_container with references on used containers
+                "settings": json.dumps(containers), # global_container with references on used containers
```
Ghostkeeper: Documentation is outdated.

thopiekar: This is done with https://github.com/Ultimaker/Cura/pull/1015/commits/10d2d7902ea35a135afe249791305a8c447b6d1d

thopiekar:
Reverted 1fd169c6697f268588733e7a6d65e80d587b51ab as discussed in another PR.

Issue #530, opened 2015-11-03
Please create an AppData file for Cura


hughsie:
Please consider writing and installing an AppData file with the application description and some screenshots, else Cura looks really bad in the GNOME and KDE Software Centers. We'd love to showcase more applications, but without the extra data file we can't. I use Cura quite a lot, and would love to feature it on the front page. If you need help or an example file, please just ask.

See http://people.freedesktop.org/~hughsient/appdata/ for details; thanks!

Richard

nallath:
Shouldn't take too much effort I guess.

hughsie:
I'm getting reports from people that cura isn't visible in gnome-software and I'm pointing them to this bug. Would you like me to try to create an AppData PR for this functionality?

nallath:
Please do. We're very close to a release, so our minds are focused on other things right now. 
Sorry that this issue fell through the cracks

PR #1031, opened 2016-10-06
Add an example AppData file so that Cura is visible in Linux software…

hughsie:
… centers

Fixes: https://github.com/Ultimaker/Cura/issues/530

### DIFF
--- a/cura.appdata.xml
+++ b/cura.appdata.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2016 Richard Hughes <richard@hughsie.com> -->
+<component type="desktop">
+  <id>cura.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>AGPL-3.0 and CC-BY-SA-4.0</project_license>
+  <name>Cura</name>
+  <summary>The world's most advanced 3d printer software</summary>
+  <description>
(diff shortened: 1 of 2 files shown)


cura.appdata.xml
```diff
@@ -0,0 +1,31 @@
...
+      <li>Experts are able to customize 200 settings to achieve the best results</li>
+      <li>Optimized profiles for Ultimaker materials</li>
+      <li>Supported by a global network of Ultimaker certified service partners</li>
+      <li>Print multiple objects at once with different settings for each object</li>
+      <li>Cura supports STL, 3MF and OBJ file formats</li>
+      <li>Open source and completely free</li>
+    </ul>
+  </description>
+  <screenshots>
+    <screenshot type="default" width="1200" height="675">https://people.freedesktop.org/~hughsient/temp/appdata-cura.png</screenshot>
```
fieldOfView: We should probably host a more up to date screenshot somewhere on software.ultimaker.com

Ghostkeeper: You can take this screenshot: http://software.ultimaker.com/Cura.png

hughsie: Update PR, thanks.

awhiemstra:
Note that this breaks the install target because it tries to install the appdata file from CMAKE_BINARY_DIR whereas it is located in CMAKE_SOURCE_DIR

PR #1032, opened 2016-10-06
Fix um2p profilenames

fieldOfView:
This PR ensures consistency between quality_type and filename/id for the UM2+ quality profiles. The PR can only be merged if we know for sure the debian build will have a clean resources/quality folder.


fieldOfView:
I don't think it is a good idea to do this now, so I retract this PR.

PR #1035, opened 2016-10-10
Fix slow writing of gcode files to some usb sticks

fieldOfView:
The GCodeWriter writes to the USB drive line by line. On some USB drives this causes writing to take multiple minutes to write a file that takes tens of seconds on other USB drives. Writing the whole gcode string at once greatly reduces the total write-time.

CURA-2544

### DIFF
--- a/plugins/GCodeWriter/GCodeWriter.py
+++ b/plugins/GCodeWriter/GCodeWriter.py
@@ -54,8 +54,7 @@ class GCodeWriter(MeshWriter):
         scene = Application.getInstance().getController().getScene()
         gcode_list = getattr(scene, "gcode_list")
         if gcode_list:
-            for gcode in gcode_list:
-                stream.write(gcode)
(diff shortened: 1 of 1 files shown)


sedwards2009:
I believe that writing one big string is going to be faster than many small strings. That said, it is worth having a look at the cause of the problem. That output stream doesn't seem to be getting the buffering it needs. We could explicitly set a buffer size in `RemovableDriveOutputDevice.py`'s `requestWrite()` method. Something in the range or 256KB should make a difference.

fieldOfView:
Good point. I've done some preliminary tests with io.BufferedWriter, and that also fixes the problem. Let's await tomorrow's decision if this issue is indeed important enough to fix in the current sprint.

fieldOfView:
This PR has been superseded by this commit: https://github.com/Ultimaker/Cura/commit/d04e6a664417108141ae9fae319c037588d468a1

PR #1042, opened 2016-10-13
Only reset the setting for the currently selected extruder.

sedwards2009:
When clicking on the little revert arrow for a setting in the UI, it reverts the setting for all extruders not just the one which is currently selected. This PR fixes this and just changes the current extruder.

CURA-2640 Restore to default not per extruder


fieldOfView:
The methodname and documentation need updating

PR #1055, opened 2016-10-18
Remove duplicate translations causing failure to build on Linux

ad1217:
All of these lines exist higher in the file, and cause msgfmt to error out when building.


Ghostkeeper:
Yeah, I had noticed them, but only after I added the (then) secret strings for the UM3. I fixed them then. We'll make those public today, so this PR won't be necessary any more then. Merging this now would only give a merge conflict.

Thanks for paying attention!

PR #1056, opened 2016-10-18
Use Global quality profiles for the Global Stack on Multiextrusion machines

awhiemstra:
This changes the behaviour related to the global stack to make sure that we use a global-container specific quality profile. In addition, we use an empty material and variant for the global stack. This fixes issues where values from the wrong profiles are used in another extruder, due to the primary extruder profile being used as global quality profile and the second extruder falling back to these values. The same would happen with materials and variants in that stack, which is why we are setting them to empty.

Ideally, we would simply use the empty quality container for the global stack, however, this interferes with the quality selection code, which depends on the quality_type being available. Since the empty container can only have one quality_type, using the empty profile would result in only having a single quality available.

### DIFF
--- a/cura/QualityManager.py
+++ b/cura/QualityManager.py
@@ -82,8 +82,9 @@ class QualityManager:
     #   \param material_containers (Optional) \type{List[ContainerInstance]} If nothing is specified then
     #                               the current set of selected materials is used.
     #   \return the matching quality container \type{ContainerInstance}
-    def findQualityByQualityType(self, quality_type, machine_definition=None, material_containers=None):
-        criteria = {"type": "quality"}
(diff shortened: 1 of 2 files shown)


cura/QualityManager.py
```diff
@@ -82,8 +82,9 @@ def __fetchQualityTypeDictForMaterial(self, machine_definition, material):
     #   \param material_containers (Optional) \type{List[ContainerInstance]} If nothing is specified then
     #                               the current set of selected materials is used.
     #   \return the matching quality container \type{ContainerInstance}
-    def findQualityByQualityType(self, quality_type, machine_definition=None, material_containers=None):
-        criteria = {"type": "quality"}
```
nallath: Update documentation to include possible kwargs options.

PR #1057, opened 2016-10-18
Fix install directories for Python files on non-Debian Linux

ad1217:
uses similar code as in libArcus to pick the correct Python directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,21 +26,27 @@ if(NOT ${URANIUM_SCRIPTS_DIR} STREQUAL "")
 endif()
 
 find_package(PythonInterp 3.4.0 REQUIRED)
+if(EXISTS /etc/debian_version)
+    # Running on a debian-based system, which requires special handling for python modules.
+    set(PYTHON_SITE_PACKAGES_DIR lib/python${PYTHON_VERSION_MAJOR}/dist-packages CACHE STRING "Directory to install Python bindings to")
+else()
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Hmm, we have a developer on Arch, who hasn't encountered this problem as far as I know.

awhiemstra:
Probably because I disabled compilation of Cura translations on my machine.

ad1217:
This isn't relating to translations, but to packaging. I maintain `cura-git` in the AUR (as well as the related `-git` packages), and have been patching `Cmakelists.txt` to use the correct directories when installing. This would remove the need for manual patching.

thopiekar:
This PR is not needed at all. All we need to change is to install to /usr/lib/python{MAJOR}.{MINOR} because there are already tools inside debhelper (packaging tools) which handle this situation!

thopiekar:
@ad1217 So if this is the layout Arch uses, you can (in my point of view) feel free to kill the case for Debian.

ad1217:
Ah, that works too. Should it also be changed in libArcus as well then?

thopiekar:
Yes, please. As we are updating now our build system, I'm sure we can take a look whether we can't get rid of this here, too.
@awhiemstra Isn't it? :wink:

nickthetait:
Any developments on this? Is it still relevant or needed?

Ghostkeeper:
Ah, yes there have been developments: https://github.com/Ultimaker/Cura/commit/6484df3e60d194e0d31f7704865b552e50423591

I think this is no longer needed, but could someone test on a non-Debian Linux installation?

ad1217:
Yeah, this got fixed a while ago, sorry I forgot to close this PR.

PR #1058, opened 2016-10-19
Changed MachineManager.isActiveStackValid to areStacksValid. CURA-2730

jackha:
Goal: User cannot create a profile that is invalid.

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -34,7 +34,7 @@ class MachineManager(QObject):
         self.globalContainerChanged.connect(self.activeVariantChanged)
         self.globalContainerChanged.connect(self.activeQualityChanged)
 
-        self._active_stack_valid = None
+        self._stacks_valid = None
         self._onGlobalContainerChanged()
 
         ExtruderManager.getInstance().activeExtruderChanged.connect(self._onActiveExtruderStackChanged)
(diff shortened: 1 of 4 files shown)


cura/Settings/MachineManager.py
```diff
@@ -405,12 +424,12 @@ def clearUserSettingAllCurrentStacks(self, key):
...
-    #   Note that the _active_stack_valid is cached due to performance issues
-    #   Calling _checkStackForErrors on every change is simply too expensive
-    @pyqtProperty(bool, notify = activeStackValidationChanged)
-    def isActiveStackValid(self):
-        return bool(self._active_stack_valid)
+    ##  Check if none of the stacks contain error states
+    #   Note that the _stacks_valid is cached due to performance issues
+    #   Calling _checkStack(s)ForErrors on every change is simply too expensive
+    @pyqtProperty(bool, notify = stacksValidationChanged)
+    def areStacksValid(self):
```
awhiemstra: While this just changes the original property, I think the naming of this (and also the original) is wrong. A stack with an error value is still a valid stack, it just has an error. An invalid stack would be something like a stack with a container that does not exist or other major programmatic error, not a stack with just an error value.

cura/Settings/MachineManager.py
```diff
@@ -337,16 +342,30 @@ def addMachine(self, name, definition_id):
     def _createUniqueName(self, container_type, current_name, new_name, fallback_name):
         return UM.Settings.ContainerRegistry.getInstance().createUniqueName(container_type, current_name, new_name, fallback_name)
 
-    ##  Convenience function to check if a stack has errors.
-    def _checkStackForErrors(self, stack):
+    ##  Convenience function to check if a stack is valid.
+    def _checkStackValid(self, stack):
```
awhiemstra: I think it is a good idea to move this to ContainerStack as a hasErrors() or similar (see also my other comment about the "valid" naming) rather than keeping it here. MachineManager is already way too overcrowded and this just adds more.

PR #1060, opened 2016-10-20
CURA-1511: Adding section to the general settings

thopiekar:
This commit adds a section to the general page to set the log-level for the console output.
It depends of a PR made for Uranium (https://github.com/Ultimaker/Uranium/pull/172).

Contributes to CURA-1511

### DIFF
--- a/resources/qml/Preferences/GeneralPage.qml
+++ b/resources/qml/Preferences/GeneralPage.qml
@@ -347,5 +347,67 @@ UM.PreferencesPage
                 onCheckedChanged: UM.Preferences.setValue("info/send_slice_info", checked)
             }
         }
+
+        Label
+        {
+            font.bold: true
+            visible: checkUpdatesCheckbox.visible || sendDataCheckbox.visible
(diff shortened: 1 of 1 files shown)


nickthetait:
Bumping this PR as it is over a year old now :open_mouth: but never got a response :cry:

Ghostkeeper:
When it has a Jira ticket number attached I tend to assume that it's some ticket in our current sprint and will be handled in our day-to-day workflow, so I won't do anything with it.

This Jira ticket was not in our current sprint and was eventually closed by our project manager for being older than 3 months. I'll duplicate that ticket and put it up so that we can judge whether or not we want this added.

LipuFei:
I think the logging level should be passed through either a parameter (e.g. argv) and/or an environment variable, or something similar. Putting this in preferences is even less convenient because you need to run Cura first in order to set the logging level, then close and start again.

ChrisTerBeke:
We have decided not to implement this. Developer features are best triggered via a command line argument.

PR #1071, opened 2016-10-23
Fix error when starting with no active machine

fieldOfView:
This PR fixes an error when starting Cura without an active machine.

This should be applied to both the 2.3 and master branch; the "critical" unhandled exception error also happens with the released 2.3.0, but it only shows in the logs. Apropos, shouldn't a "critical" error trigger the CrashHandler instead of failing silently?


Ghostkeeper:
I also cherry-picked this to 2.3.
This seems to make the profile drop-down yellow before any printer is selected, which is actually kind of expected behaviour.

PR #1074, opened 2016-10-24
Fix UM2 start gcode

fieldOfView:
This PR adds homing the X/Y axes, which is currently missing. As a result the UM2/UM2+ will not have the correct origin and may start printing outside the build volume if the gcode flavor is set to RepRap (eg for USB printing or for OctoPrint).


fieldOfView:
Wrong branch.

PR #1075, opened 2016-10-24
Fix homing UM2/UM2+ when printing with RepRap flavor

fieldOfView:
This PR adds homing the X/Y axes, which is currently missing. As a result the UM2/UM2+ will not have the correct origin and may start printing outside the build volume if the gcode flavor is set to RepRap (eg for USB printing or for OctoPrint).

It would be nice if this fix could be included in 2.3.1

### DIFF
--- a/resources/definitions/ultimaker2.def.json
+++ b/resources/definitions/ultimaker2.def.json
@@ -21,7 +21,7 @@
         "machine_name": { "default_value": "Ultimaker 2" },
         "machine_start_gcode" : {
             "default_value": "",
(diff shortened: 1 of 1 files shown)


nallath:
I'd love to merge this into 2.3.1, but it's not really a critical issue. It does affect quite a big base of printers out in the wild, so unless we absolutely know for sure that this won't break anything...

fieldOfView:
> it's not really a critical issue
It is though... It is not a pretty sound when the gantry tries to push the printer 20cm through the backplate.

For those who print with UltiGCode, there is 0% chance this change does anything for them (because they get an empty start gcode string, like they already do)

nallath:
Fair enough.

PR #1078, opened 2016-10-25
Render an indication of the origin of the machine

fieldOfView:
This PR adds a small indication at the machine origin. This is especially important for those machines that don't have a machine mesh that shows which end of the buildplate is the front.

The PR renders three thin cubes, similar to how toolhandles draw thick lines. This way the origin axes draw with a bit of body, over the blue-ish buildvolume wireframe cube.

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -33,6 +33,9 @@ PRIME_CLEARANCE = 1.5
 ##  Build volume is a special kind of node that is responsible for rendering the printable area & disallowed areas.
 class BuildVolume(SceneNode):
     VolumeOutlineColor = Color(12, 169, 227, 255)
+    XAxisColor = Color(255, 0, 0, 255)
+    YAxisColor = Color(0, 0, 255, 255)
+    ZAxisColor = Color(0, 255, 0, 255)
 
     raftThicknessChanged = Signal()
 
@@ -44,6 +47,8 @@ class BuildVolume(SceneNode):
         self._depth = 0
 
         self._shader = None
+
+        self._origin_mesh = None
 
         self._grid_mesh = None
         self._grid_shader = None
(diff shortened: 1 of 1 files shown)


thopiekar:
It sounds interesting 😃 
Do you have a screenshot around?

Regards 😉

fieldOfView:
No, but I can make one:
origin

thopiekar:
Very kind, thanks. Looks cool.

I can imagine that there could be the argument that this could confuse beginners or whatever.. In that case it should be very easy to add an option to dis-/enable it 😉

fieldOfView:
I don't understand, what could be confusing about it?

Note that the screenshot is fairly zoomed in on a fairly small custom machine. For most normal printers the effect is more subtle.

thopiekar:
Don't ask me. There are also many things, which don't disturb me, but others.
In my point of view there is nothing wrong with it.

Just had this idea in mind, no criticism, only brainstorming..

Ghostkeeper:
@nallath was just adding this feature, recently. Don't know what became of his implementation though.

What happens when `machine_center_is_zero` is activated?

fieldOfView:
When machine_center_is_zero is activated, the origin-indication moves to the center of the buildplate. In this case it still functions as an indication of the orientation as well, because the axes point towards positive x,y & z (with the same colors that are used for the x,y,z axes that are used elsewhere, i.e. in the tools)

Ghostkeeper:
The indicator should still be part of the build. Last I checked...

Cyberman-tM:
Interesting. To clarify: I'm using the public build, 2.3.1 - I assumed that this would be already in there?
I've now downloaded 2.4 beta, but so far was unable to start it. It just disappears...
I should check the system logs, maybe something else is interfering.

Ghostkeeper:
Possibly it had something to do with the settings folder. If it's getting an error during the process of upgrading your files from the format of 2.3 to the format of 2.4, Cura just tends to stop without warning.

PR #1079, opened 2016-10-25
Velleman Vertex Config

mistertilapia:
Added a config file for the Velleman Vertex K8400

### DIFF
--- a/resources/definitions/velleman_vertex.def.json
+++ b/resources/definitions/velleman_vertex.def.json
@@ -1,0 +1,56 @@
+{
+    "id": "velleman_vertex",
+    "version": 2,
+    "name": "Velleman Vertex",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Cade Cyphers",
+        "manufacturer": "Velleman",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
+    },
+    
+    "overrides": {
+        "machine_name": {
+            "default_value": "Velleman Vertex"
+        },
+        "machine_heated_bed": {
+            "default_value": false
+        },
+        "machine_width": {
+            "default_value": 190
+        },
+        "machine_height": {
+            "default_value": 200
+        },
+        "machine_depth": {
+            "default_value": 180
+        },
+        "machine_center_is_zero": {
+            "default_value": false
+        },
+        "machine_nozzle_size": {
+            "default_value": 0.35
+        },
+        "material_diameter": {
+            "default_value": 1.75
+        },
+        "machine_nozzle_heat_up_speed": {
+            "default_value": 2
+        },
+        "machine_nozzle_cool_down_speed": {
+            "default_value": 2
+        },
+        "machine_gcode_flavor": {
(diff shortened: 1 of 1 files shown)


resources/definitions/velleman_vertex.def.json
```diff
@@ -0,0 +1,56 @@
...
+    "version": 2,
+    "name": "Velleman Vertex",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Cade Cyphers",
+        "manufacturer": "Velleman",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
```
Ghostkeeper: This comma here gives a syntax error.

resources/definitions/velleman_vertex.def.json
```diff
@@ -0,0 +1,56 @@
...
+            "default_value": false
+        },
+        "machine_width": {
+            "default_value": 190
+        },
+        "machine_height": {
+            "default_value": 200
+        },
+        "machine_depth": {
+            "default_value": 180
```
Ghostkeeper: Going by the specs online, I think Height and Depth are switched around here. In Cura (a little unconventially) the width is the X axis, the depth is the Y axis and the height is the Z axis. It's as if you're looking from the front side of the machine, so to say. Could you verify that this is correct?

Ocarthon:
I also created the config for this printer and used the following values:
width = 180
depth = 180
height = 190

i took the values from the Repetierhost version for this printer.

But if this get's added, we should also add the version for the one with two extruders

mistertilapia:
The specs online are strangely incorrect, as the print area is most definitely wider than it is deep. I believe the config is correct, and perhaps Velleman has the measurements ordered strangely.
I can double check when I get home if needed, though.

BagelOrb:
Please rectify the problems of this definition file so that we can incorp[orate it into Cura.

tbackdahl:
I made a different attempt for both single and dual head versions.
The machine uses printable coordinated x=0 to 200, y=20 to 200 regardless of how many extruders you use, thereby the different machine_width for the two models.
New files:
vertex_k8400.def.json
vertex_k8400_dual.def.json
vertex_k8400_dual_1st.def.json
vertex_k8400_dual_2nd.def.json
Vertex_build_panel.stl
Sorry if I did this in the wrong way -- I am new to git.

Ghostkeeper:
You should commit these files to your fork of Cura that you used to make this pull request. They'll update here then and we can merge them if they are correct.

BagelOrb:
Alternatively you could close this pull request and open a new one with
tree new files.

Op 24 feb. 2017 12:06 schreef "Ghostkeeper" <notifications@github.com>:

tbackdahl:
Sorry about the cryptic message. I am not the same person who made this pull request.
Now, I finally managed to make a new pull request #1474.
I had not understood that I had to fork the project before I could commit changes and make pull requests.

Ghostkeeper:
Ah, sorry. I didn't see that you weren't the original poster of this pull request. The new pull request does work and makes this pull request obsolete, so I'm closing this one.

PR #1083, opened 2016-10-26
Fix errors when switching from UM3 to UM2.

fieldOfView:
This PR fixes "critical" errors caused (I think) by "empty_material" not having a definition. The error also happens in the 2.3 branch, so it should probably be cherry-picked.


nallath:
Done :)

PR #1116, opened 2016-11-07
Add dialog to set number of copies when duplicating models

fieldOfView:
This PR adds a dialog to allow the user to set the number of copies when duplicating models.

The copies are still naively set in the center of the buildplate, after which it is up to the buildplate physics to push them apart.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -686,10 +686,9 @@ class CuraApplication(QtApplication):
             while current_node.getParent() and current_node.getParent().callDecoration("isGroup"):
                 current_node = current_node.getParent()
 
-            new_node = copy.deepcopy(current_node)
-
             op = GroupedOperation()
             for _ in range(count):
+                new_node = copy.deepcopy(current_node)
                 op.addOperation(AddSceneNodeOperation(new_node, current_node.getParent()))
             op.push()
 
--- a/resources/qml/Actions.qml
+++ b/resources/qml/Actions.qml
@@ -232,7 +232,7 @@ Item
     Action
     {
         id: multiplyObjectAction;
-        text: catalog.i18nc("@action:inmenu","&Duplicate Model");
+        text: catalog.i18nc("@action:inmenu","&Duplicate Model...");
         iconName: "edit-duplicate"
     }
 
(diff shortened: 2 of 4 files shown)


nickthetait:
We actually implemented this recently too. Sorry for not creating a PR before now, just escaping 3 weeks backlog of non-cura2 stuff... :cry: 
Anyway, these should be the two related commits:
https://code.alephobjects.com/rCTfdf96ababdc12b7fb6f15e7ef437c53879016e58
https://code.alephobjects.com/rCT032edaa7e6f0fb0b401385a9993e7900cd347fb4

If our solution is preferred over fieldOfViews' I'll actually make a PR for it.

nallath:
As for any bugfixed, please also make pull requests for those.

This was one of those features that we had on our tracker for quite some time, but more important stuff kept popping up all the time.

fieldOfView:
I have made some suggestions on the code at alephobjects. Good to see we started the same way; at some point my code looked almost exactly the same as theirs, but I think mine has a couple less issues.

Biggest functional difference is that I left the menu entry at "Duplicate Model..." instead of changing it to "Multiply Models...". I guess Multiply is more correct here (though multiple duplicates can exist of a single model). Which do we prefer?

awhiemstra:
Personally I would prefer not to have a dialog at all and support duplicating with multiple selected objects and copy/paste. But if people insist on a dialog, I would use multiply personally.

fieldOfView:
> support duplicating with multiple selected objects 

That's another issue. This is about creating - say - 50 copies at once instead of repeatedly having to duplicate/select all/group/duplicate/ungroup/select all/group/duplicate...

nickthetait:
Popup/multiply method is just like Legacy Cura. It restores the ability to produce the same object hundreds of times easily. This is not feasible just by using duplicate.

Ghostkeeper:
If you put in 3, it creates 3 new instances of the selected mesh. Which means that there are now 4 of them. I thought that was a bit unintuitive, but it's the same behaviour as legacy.

For that reason, and since there are no further complaints in these comments, I decided to merge this.

fieldOfView:
@Ghostkeeper: that's why I initially stayed away from the "multiply" wording. The label "Number of copies" is correct though; the selected object is the original, so it is not a copy.

PR #1118, opened 2016-11-08
Updates to Machine Settings

fieldOfView:
This PR fixes some things in relation to the Machine Settings plugin:
* Remove hardcoded container index
* Fix compatibility with machines that use variants (UM2+)
* Tweak start/end gcode editing for readability: fixed-width font and no linewrapping
* Get gcode flavor options from SettingPropertyProvider instead of hardcoding

~~Todo:~~
* ~~Fix compatibility with multiextrusion printers (UM3, UMO Dual, custom)~~

~~Requires https://github.com/Ultimaker/Uranium/pull/183~~

I originally wanted to add compatiblity for multiextrusion printers in this same PR. I now think this should be done in a new PR because otherwise this PR would become to big and take too long. Please consider the implemented changes.


fieldOfView:
Which settings from fdmextruder.def.json do we want to be easily settable from Machine Settings for printers that have multiple extruders?

At least machine_nozzle_offset_x and machine_nozzle_offset_y.  How about start- and end position and gcode? Are those important enough to include here?

fieldOfView:
I have changed the MachineSettings plugin to get the gcode flavors from the setting (ie: fdmprinter.def.json) instead of hardcoding a subset. I noticed that the human-readable form of "UltiGcode"" is "Ultimaker 2", whereas the human-readable form of Griffin is "Griffin", not "Ultimaker 3". For consistency (also with legacy Cura) I would call "UltiGcode" by its own name. I can change that in this same PR.

Ghostkeeper:
I think it'd be good to have the start and end g-code of the extruders in the machine settings as well. The g-code to perform an extruder switch will differ vastly between different printers, and it allows for wiping the nozzle on a wipe cushion, moving to the switching bay, etc.

nickthetait:
> Which settings from fdmextruder.def.json do we want to be easily settable from Machine Settings for printers that have multiple extruders?

Being able to pick the number of nozzles. Last I checked, it is not possible to configure a printer for multi-extrusion without directly manipulating the .def.json.

fieldOfView:
> Being able to pick the number of nozzles.

Yeah, that's a bit of a tricky one with the current architecture. If you look at the definition of the UM3, you will see that there are separate definition files for the extruders. Adding an extruder to a printer would mean adding such a file. Not impossible, but quite a bit of work.

nickthetait:
How about not needing to create those files? Would it be sufficient to default each extruder to use whatever is defined by the printer.def.json?

fieldOfView:
The settings for the extruders need to be stored somewhere (the aforementioned extruder offset & switch gcode), so we need files. For printers that don't define their own extruders, the extruders would simply point to fdmextruder.def.json (which is the extruder equivalent of fdmprinter.def.json), and the MachineSettings plugin would create a variant per extruder (like it currently creates a variant for the printer)

fieldOfView:
The PR currently fixes Machine Settings for the UM2+, adds all the G-Code flavors CuraEngine supports, and improves G-Code legibility. I intend to fix support for multiextrusion printers in a new PR, but feel this PR would drag on too long and would become too complex if we wait for it.

Ready for review.

plugins/MachineSettingsAction/MachineSettingsAction.py
```diff
@@ -21,27 +21,56 @@ def __init__(self, parent = None):
...
-        new_variant.setDefinition(global_container_stack.getBottom())
-        UM.Settings.ContainerRegistry.getInstance().addContainer(new_variant)
-        global_container_stack.replaceContainer(self._container_index, new_variant)
+            if variant and variant.getId() == "empty_variant":
+                # There is an empty variant that we can use to store the machine settings
+                container_index = global_container_stack.getContainerIndex(variant)
+                machine_settings_variant = self._createMachineVariant(global_container_stack, container_index)
+            else:
+                # Add a second variant before the current variant to store the machine settings
+                machine_settings_variant = self._createMachineVariant(global_container_stack)
```
Ghostkeeper: The default way of finding a specific profile (e.g. the material) is by asking the container stack for the first profile with a specific type, like we do in line 39 here. I fear that that would all break if there are two containers with the same type in one stack. That makes me hesitatant to merge this.

Ghostkeeper:
Done a review. My only concern was this: https://github.com/Ultimaker/Cura/pull/1118/commits/aa9856e83bc653bb02489a9be496520699bb8ce1#r88426555

Ghostkeeper:
I'm not sure what to do with this so I'd like a second opinion before we merge it. I'll ask @nallath tomorrow in person.

fieldOfView:
I agree with your review comment.

It would be better if the plugin would add a container of type "machine_settings" instead of "variant". Effectively it would work the same, settingwise. Getting this container stored was looking tricky, so I ended up with the current double variant.

fieldOfView:
I've given this some more thought, reconstructing how I came to my current solution. ContainerStack.findContainer() iterates over its _containers, which is a list. Lists have an order, so if you search for a container of type "variant", you always get the first variant container first: https://github.com/Ultimaker/Uranium/blob/master/UM/Settings/ContainerStack.py#L367. This should work. Having said that, it would need testing beyond the testing that I already did.

I think it is actually very powerful if a machine can have multiple levels of variants.

The "machine_settings" solution I mentioned above would have more implications outside the Machine Settings plugin; UMOUpgradeOptions would have to use the same "machine_settings" container (or the heated bed kit option can not be overridden with Machine Settings), a resource type would have to be registered and an additional folder would be created. Finally there would have to be an upgrade step to convert the date from 2.3 variants to machine_settings for machines which have had options set using Machine Settings. It can be done, but the current implementation is simpler.

Ghostkeeper:
Having discussed this, we've reached consensus that it is indeed risky to create multiple containers of the same type on a stack. Not only with the `findContainer`, but also with how this works with multiple extruders and with the variant that gets referred to in the material profiles.

Our preferred solution is to use a different type, a "machine changes" so to say, that is specific to the machine. It might need to be put on top of the variant and/or be specific to the variant though. Having an extra folder and resource type is no problem. The upgrade for version 2.4 can be arranged (it is one of the open issues for different reasons anyway).

fieldOfView:
Ok. Having just finished upgrading my UM2 to a UM2+, I'll see what I can do.

fieldOfView:
Added 1 new commit (and rebased the branch to be up to date with master).

The changes in MachineSettings and UMOUpgradeOptions are now stored in a "definition_changes" container which sits between the definition and the variant. This way there can be no conflicts between the definition changes/settings and the variant.

The definition_changes are now stored in the "machine_instances" folder. When I tackle multi-extrusion MachineSettings, I can make sure those machine_changes containers are stored in the "extruder_trains" folder.

Ghostkeeper:
Thanks. The new commit looks good. Nothing to comment.
A component test (and quick regression test) will give the final verdict.

fieldOfView:
Note that since there's no 2.3 to 2.4 upgrade yet, previously created variants in existing machines will override definition_changes settings

Ghostkeeper:
I didn't test upgrading from version 2.3. I'll add a note to the version upgrade issue we still have open.

PR #1125, opened 2016-11-09
Modify reading of Coordinate/@point attribute to allow comma separators

vincentmarchetti:
This change allows reading X3D IndexedTriangleSet nodes in which the list of coordinates contains commas as separators. Commas  are permitted by the specification of the XML encoding of the X3D standard: see ISO/IEC 19776-1:2015	(X3D Encoding:XML V3.3) Section 5.1.2;
http://www.web3d.org/documents/specifications/19776-1/V3.3/Part01/EncodingOfFields.html#5.1.2

### DIFF
--- a/plugins/X3DReader/X3DReader.py
+++ b/plugins/X3DReader/X3DReader.py
@@ -700,7 +700,11 @@ class X3DReader(MeshReader):
                 if not c is None:
                     pt = c.attrib.get("point")
                     if pt:
-                        co = [float(x) for x in pt.split()]
+                        # allow the list of float values in 'point' attribute to 
+                        # be separated by commas or whitespace as per spec of 
+                        # XML encoding of X3D 
+                        # Ref  ISO/IEC 19776-1:2015 : Section 5.1.2
+                        co = [float(x) for vec in pt.split(',') for x in vec.split()]
                         num_verts = len(co) // 3
                         self.verts = numpy.empty((4, num_verts), dtype=numpy.float32)
                         self.verts[3,:] = numpy.ones((num_verts), dtype=numpy.float32)


Ghostkeeper:
An improvement, but it's not complete. The ISO specifies this as a valid coordinate string: `1 1 1, 0 0 0, `
That would break here though, because it splits on spaces as well as commas and therefore there would be two empty strings in the resulting sequence (since there is nothing between the comma and the space).

vincentmarchetti:
For the example you posed, with a terminal comma, the first split on commas does have an empty string as the last element of a list; when this empty string is split on whitespace that
[pyramid_0.x3d.zip](https://github.com/Ultimaker/Cura/files/583317/pyramid_0.x3d.zip)

 returns an empty list, so does not contribute to the returned list of converted floats.

I have tested  on a  test file illustrating this case (attached) and it imports into a Cura app built with the revised code.

Ghostkeeper:
The example also has a comma and a space halfway through the string. It should first split it into two:

    ["1 1 1", " 0 0 0"]

Then it should split the second string with an empty string at the beginning:

    [["1", "1", "1"], ["", "0", "0", "0"]]

But if it works on that pyramid file, I'll believe you, since it has `0 14.66e-3 0, 12.7e-3 -7.34e-3 0, -12.7e-3 -7.34e-3 0, 0 0 20.73e-3,` as one of the coordinates.

brutzman:
Be careful if inserting commas... the XML encoding does not allow them in the middle of coordinate points, just between them.  This significantly helps author sleuthing if a pair of values is broken by losing intervening whitespace, or other issues emerge in a reeeally loooong set of 3-tuple points.  The array structure is more visible.

Legal .x3d examples:
<Coordinate point='0 0 0 1 0 0 0 1 0 0 0 1'/>
<Coordinate point='0 0 0, 1 0 0, 0 1 0, 0 0 1'/>

Illegal .x3d example:
<Coordinate point='0, 0, 0, 1 00, 0 1, 0, 0, 0 1'/>

Further background, incidentally... this is checked via XML schema and regular expression (regex) validation, as part of X3D Quality Assurance (QA).  All intervening comments are removed in X3D Canonical Form (supporting XML digital signature for authentication, or XML encryption for privacy).

Ghostkeeper:
The XML encoding has nothing to say about commas in coordinate points. The XML specification has no concept of "coordinate points", and neither has the encoding of XML files.

In any case, this pull request seems to make the provided tetrahedron file work, and is according to spec. All other X3D test cases I have also still work (though the normals are still broken, but that is a different story).

PR #1142, opened 2016-11-13
Initial support for JellyBOX printer

filipgoc:
Adds the machine definition and mesh for JellyBOX (http://imade3d.com) for rudimentary support.


resources/definitions/jellybox.def.json
```diff
@@ -0,0 +1,35 @@
...
+    "version": 2,
+    "name": "JellyBOX",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "IMADE3D",
+        "manufacturer": "IMADE3D",
+        "category": "Other",
+        "platform": "jellybox_platform.stl",
+        "platform_offset": [ 0, 0, 0],
```
Ghostkeeper: Currently the platform seems to be _just_ a little bit too high, as it goes into the build volume and makes the build plate appear foggy. It would help if you made the second coordinate slightly negative, moving the build plate down. `[0, -0.3, 0]` seems to work well.

resources/definitions/jellybox.def.json
```diff
@@ -0,0 +1,35 @@
...
...write Preparing\nM140 S{material_bed_temperature}              ;set bed temperature and move on\nM104 S{material_print_temperature}             ;set extruder temperature and move on\nM206 X10.0 Y0.0       ;set x homing offset for default bed leveling\nG21                   ;metric values\nG90                   ;absolute positioning\nM107                  ;start with the fan off\nM82                   ;set extruder to absolute mode\nG28                   ;home all axes\nM203 Z5               ;slow Z speed down for greater accuracy when probing\nG29                   ;auto bed leveling procedure\nM203 Z7               ;pick up z speed again for printing\nM190 S{material_bed_temperature}              ;wait for the bed to reach desired temperature\nM109 S{material_print_temperature}             ;wait for the extruder to reach desired temperature\nG92 E0                ;reset the extruder position\nG1 F200 E5            ;extrude 5mm of feed stock\nG92 E0                ;reset the extruder position again\nM117 Print starting   ;write Print starting\n---------------------------------------------\n; ; ; Jellybox Printer Start Script End ; ; ;\n_____________________________________________"
```
Ghostkeeper: Are you sure that these `--------------------` and `____________` work? Or was it just scaffolding?
Marlin doesn't usually accept that sort of thing as g-code.

filipgoc: tested and works. 
however, it's a good point - i'll comment it out to make sure. 

also, look how awesome it looks: 
image

filipgoc: (maybe in the future i'll modify the summary so that it contains the proper gcode overwrites, too, for now I'm good)

resources/definitions/jellybox.def.json
Ghostkeeper: Thanks. It can be merged once this is done, as far as I'm concerned.

filipgoc:
I made the requested changes 12 days ago.
Anything else I should do to get this approved?

Ghostkeeper:
Ah, sorry. I get notifications for messages and commits to pull requests, but the notifications for commits I usually just delete thinking that people are still working on the pull request. So when there is no message I don't usually look at the pull requests any more. I'll test it now.

filipgoc:
No problem. I did actually say 'ready to go,' but inline in the code as a response to the original comments - and it somehow initiated a 'review' instead of appending the message here. Unexpected messaging behavior...

Thanks! Sweet. Now off to make the material switching magic working.

PR #1147, opened 2016-11-15
Find/filter settings in sidebar

fieldOfView:
This PR adds a filter-as-you-type textfield to filter the settings in the sidebar. This makes it easier to find a setting if the user has made a lot of settings visible. The implementation also shows all settings when filtering, regardless of their visibility setting. "Invisible" settings have a contextmenu item to make them visible. 

With this PR a detour through the Setting Visibility preference page is no longer needed for less-often used features.

CURA-918

This is an extension of work done by @victor9999 for Aleph Objects (Lulzbot):
https://code.alephobjects.com/T562
Note that the initial commit seems huge, but 90% of it is just wrapping the existing code in an Item.

**Todo:**
* ~~If Cura is closed when the sidebar is being filtered, all categories will be expanded when Cura is restarted~~ Fixed 
* ~~It could be nice if there were a "lookingglass" toggle to show/hide the filter textfield so it does not take up sidebar realestate when not in use.~~ Added
* ~~An "x" to clear the filter textfield?~~ Added


fieldOfView:
> If Cura is closed when the sidebar is being filtered, all categories will be expanded when Cura is restarted

I see two ways of preventing this:
* Catch the application closing, restore the expanded categories before the preferences are written
* Add a "don't write to preferences" flag to SettingPreferenceVisibilityHandler
I think the latter is a better approach, because the former only works if the application is closed cleanly

awhiemstra:
Neither solution is actually the correct one. We watch the "expendedCategories" property for changes, then when it changes, we update the preferences property related to this. So the solution is to not update the preferences when you're busy filtering stuff. Note that preventing the visibility handler from saving to preferences will actually break one of your listed features, as it will no longer save which settings have been made visible. Expanded/collapsed is separate fro m this.

fieldOfView:
Thanks, I was making this too complicated in my head.

Ghostkeeper:
Or to save the filter as a preference too and restore the entire state on start-up?

nallath:
This feature is amazing!

fieldOfView:
Yeah, CURA-918 is something I proposed in the 2.1 era, but it got postponed until st. Juttemis.
I'm sort of struggeling with a nice spot to put a toggle for the filter though.

fieldOfView:
> save the filter as a preference

Nah, filters like this are not often persisting between app relaunches. Or do you have a certain usecase in mind?

nickthetait:
I can't think of a reason why the filter should be persistent.

awhiemstra:
I also don't think the filter should be persistent, that will only lead to confusion.

thopiekar:
I also don't see why it should be useful.
But to be fair the only compromise I see here is to show the last keyword in grey.
So when the user clicks into the text field the greyed text is still visible and pressing enter will use that one.
On the other hand if the user is looking for something different he can begin to type in her/his new keyword as usual.
What do you think?

fieldOfView:
I think we should not reinvent search-as-you-type fields to the point where they behave differently in Cura than in any other application unless we have a very good usecase.

14bitVoid:
I also don't think what search filter you use should be made persistent. It's not what you'd expect but just a quick search

fieldOfView:
The filter textfield is now toggleable.
Ready for review.

resources/qml/Settings/SettingView.qml
```diff
@@ -9,229 +9,406 @@ import QtQuick.Layouts 1.1
...
-                        return Cura.MachineManager.activeMachineId;
-                    }
-                    if(inheritStackProvider.properties.limit_to_extruder != null && inheritStackProvider.properties.limit_to_extruder >= 0)
+                    if(findingSettings)
                     {
-                        //We have limit_to_extruder, so pick that stack.
-                        return ExtruderManager.extruderIds[String(inheritStackProvider.properties.limit_to_extruder)];
+                        expandedCategories = definitionsModel.expanded.slice();
+                        definitionsModel.expanded = ["*"];
+                        definitionsModel.showAncestors = true;
```
Ghostkeeper: Won't this also show normal parents, such as "Line Width" is a parent of "Wall Line Width"? Is that intended?

resources/qml/Settings/SettingView.qml
```diff
@@ -9,229 +9,406 @@ import QtQuick.Layouts 1.1
...
+            rightMargin: UM.Theme.getSize("default_margin").width
+        }
+        height: visible ? UM.Theme.getSize("setting_control").height : 0
+        Behavior on height { NumberAnimation { duration: 100 } }
 
-            source:
+        TextField
+        {
+            id: filter;
+            visible: !monitoringPrint
```
Ghostkeeper: Unnecessary, right? Its parent is `filterContainer`, which should also be invisible then if I'm not mistaken.

resources/qml/Settings/SettingView.qml
```diff
@@ -9,229 +9,406 @@ import QtQuick.Layouts 1.1
...
+            property var expandedCategories
+            property bool lastFindingSettings: false
+
+            onTextChanged:
             {
-                target: provider
-                property: "containerStackId"
-                when: model.settable_per_extruder || (inheritStackProvider.properties.limit_to_extruder != null && inheritStackProvider.properties.limit_to_extruder >= 0);
-                value:
+                definitionsModel.filter = {"label": "*" + text};
```
Ghostkeeper: Shouldn't it be filtering on `"*" + text + "*"`?

Ghostkeeper: I think this would also continue to search in English even if you set the language differently. That is a change that would need to be put through to Uranium though, in the way that this filter is applied.

resources/qml/Settings/SettingView.qml
```diff
@@ -9,229 +9,406 @@ import QtQuick.Layouts 1.1
...
-                key: model.key
-                watchedProperties: [ "limit_to_extruder" ]
+                filter.text = "";
             }
+        }
 
-            UM.SettingPropertyProvider
+        MouseArea
+        {
+            id: mouseArea
```
Ghostkeeper: It would do us well to give this a better name. I have no idea what this mouse area is used for.

resources/qml/Sidebar.qml
```diff
@@ -256,6 +256,44 @@ Rectangle
...
+    {
+        id: toggleFilterButton
+
+        anchors.right: parent.right
+        anchors.rightMargin: UM.Theme.getSize("default_margin").width
+        anchors.top: headerSeparator.bottom
+        anchors.topMargin: UM.Theme.getSize("default_margin").height
+
+        height: settingsModeSelection.height
+        width: visible ? height : 0
```
Ghostkeeper: Are these to show an animation or something? If it's not visible this shouldn't have any effect, I'd think.

resources/qml/Sidebar.qml
```diff
@@ -214,7 +214,7 @@ Rectangle
                 anchors.left: parent.left
                 anchors.leftMargin: model.index * (settingsModeSelection.width / 2)
                 anchors.verticalCenter: parent.verticalCenter
-                width: parent.width / 2
+                width: 0.5 * parent.width - (modesListModel.get(index).showFilterButton ? toggleFilterButton.width : 0)
```
awhiemstra: Since you're already changing the width of the toggleFilterButton there is no need to add the extra check here, especially since it make the expression a lot more complicated. 

Also, this will remove the button width twice due to operator precedence.

resources/qml/Sidebar.qml
awhiemstra: A visible item still has a width and a height. Anything that layouts based on an item's width or height will thus still use the original width and height of the item. Strictly speaking, I do not think that is the case here but changing width here will make the expression above simpler.

resources/qml/Settings/SettingView.qml
fieldOfView: The filtering here is lifted directly from the filtering as it is applied on the Setting Visibility page. That also filters on ```"*" + text```. The underlying findContainer just checks ```if "*" in filter``` and does not care where the ```"*"``` is.

resources/qml/Settings/SettingView.qml
fieldOfView: That is intended. The same logic is used on the setting visibility page; it is important for learn-ability of where a setting lives.

nallath:
What is everyones opinion about this? 

I'd love to get this in before the beta release, so we have ~10 days to do that.

resources/qml/Sidebar.qml
fieldOfView: The ternary switch here is not for the width of the toggleFilterButton, but for whether the toggleFilterButton is shown for this mode, ie: 0 for "simple", toggleFilterButton.width for "advanced". The toggleFilterButton hides itself when the "simple" tab is showing, to not clutter up the simple mode.

resources/qml/Sidebar.qml
fieldOfView: This is used so the "advanced" button expands to its original width when the user is in "simple" mode. In "simple" mode the toggle that is only relevant in the "advanced" mode is simply hidden altogether to avoid confusion and to maintain simplicity. When the user switches to "advanced" mode the "advanced" button is resized slightly to make room for the toggle button.

resources/qml/Settings/SettingView.qml
fieldOfView: You are indeed not mistaken.

resources/qml/Settings/SettingView.qml
fieldOfView: Point taken. Will ```hoverMouseArea``` be enough of an indication that this mouseArea is used to catch hover events?

Ghostkeeper: Yeah that's fine. If I'd read that I would think that it is meant to denote the area where it highlights something if you move the cursor over it.

resources/qml/Sidebar.qml
awhiemstra: I get that; my point is that you are already changing the button's width based on whether it should be visible or not. So the end result is the same: if you show the button, its width will be > 0 and the other two buttons resize based on the button's width. If the button is hidden, its width will be 0 and the buttons will not subtract any size.

Ghostkeeper:
As far as I've heard, the current opinion is that the risk has been reduced enough that we can consider merging it as long as we do a bit of regression testing.

resources/qml/Sidebar.qml
fieldOfView: Ok, so what do you suggest?
```width: 0.5 * parent.width - toggleFilterButton.width``` ?
That will decrease the width of the "simple" button too. The idea is that the "simple" button is not changed, and the "advanced" button is slightly less wide to make room for the toggle button, but only when it is selected.

fieldOfView: Changed to ```model.showFilterButton```, which is indeed clearer. Thanks (why didn't I think of that?)

nallath:
I'm not that much a fan of the looking glass icon, but it's definately not enough to prevent this pretty awesome addition from being merged!

fieldOfView:
>I'm not that much a fan of the looking glass icon

Not a fan of the toggle, not a fan of using a looking glass as an icon for it, or not a fan of this specific looking glass? Any suggestions for change?

fieldOfView:
Thanks @victor9999 and @nickthetait for your work getting this started.

nallath:
I've got various levels of dislike for parts of it. 
The most pressing one is that it feels a bit wrong that there is a second button on the custom tab itself.

PR #1151, opened 2016-11-16
CURA-2611: Adding workaround to avoid pyc files

thopiekar:
This will avoid pyc files when using Cura on Windows or any other OS,
where it was frozen.

Contributes to CURA-2611
Contributes to CURA-2387
Worked on this while working on CURA-1767 (Had to wait for finished builds...)


fieldOfView:
Wouldn't this also be fixed if the installer didn't run Cura with elevated privileges (admin rights)? Normally Cura cannot write to its own folder since it is in Program Files.

I guess this is still nice to have for those people who troubleshoot everything with "run as administrator"

thopiekar:
Yes, but there is an additional problem around this. You probably remember that there was a problem, that you can't drag and drop when starting from the installer.. That's CURA-1767.

thopiekar:
Oh, and for those who are using "Run as administrator" I would love to write an easter egg..
Maybe adding a dialog and ask whether they really want to continue with admin rights.. and ask that eg. 15 times and place the dialog at a random place on the screen 😝

thopiekar:
As a reference the fix on cura-build: https://github.com/Ultimaker/cura-build/pull/93

Ghostkeeper:
Why don't we want .pyc files in plug-in folders?

thopiekar:
Well, if you can tell python to avoid pyc in Cura but except the plugin
dir, then go for it. I think skipping pyc everywhere is better then doing
mess and not cleaning it up with the uninstaller correctly. I really hate,
when I have to remove things on my own when there is a uninstaller. It is
its job to remove (if I want) everything.
Also 'normal' users don't want to dig into some directories to get a clean
install.
I think we have to expect always the stupidest user as this point and make
it really user friendly.

However, his is only my opinion and the PR here is lying around here for
months. I doubt there is any interest in tidiness on our customers
computers.

Ghostkeeper:
I guess it would be ideal to have the uninstaller also remove the .pyc files...

thopiekar:
In my opinion avoiding making mess is better than (not) cleaning it up properly.
So if we don't create these files, we also don't need to care about removing them ☝️

And getting there is only about clicking on the merge button. Can't be more simple..

fieldOfView:
Something has gone very wrong with this PR

PR #1157, opened 2016-11-18
Correcting German translation

thopiekar:
Contributes to CURA-2862


Ghostkeeper:
CURA-2862 is called "Merge UM3NetworkPrinting into Cura". I think you mentioned the wrong Jira issue?

thopiekar:
Ooops, noted the ticket name here on a leaflet, but mixed it up with another ticket I was working on yesterday.
It is CURA-2812. Shall I make a new commit with a correct "Contributes to" tag?

Ghostkeeper:
Nah, if we just make a comment in CURA-2812 it should be fine. It happens all the time, and a comment has always worked.

PR #1164, opened 2016-11-20
PrinterOutputDevice: Add isConnectionState

thopiekar:
Adding isConnectionState here to make direct comparisons, whether the
current state is the correct one.
Looks better than `self._connection_state == <my_connection_state>`  in
my point of view.

### DIFF
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -366,6 +366,11 @@ class PrinterOutputDevice(QObject, OutputDevice):
             self._connection_state = connection_state
             self.connectionStateChanged.emit(self._id)
 
+    ##  Check the connection state of this output device.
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I disagree. It abstracts something that is already very simple, which makes it more of a black box.

Also, the new function is not used anywhere.

thopiekar:
Yes, of course. What I probably dislike here is the underline here.
Well, not at the moment, but I'm rewriting the USB printing plugin incl. Gcode postprocessor ( at the moment you can enable a dryrun, so you can test the communication without extrusion) to be modular.

However, the code is not published at the moment because I want to complete my whole work on it 😉

nickthetait:
That sounds very useful thopiekar! :heart_eyes:

PR #1179, opened 2016-11-26
Unity: Adding workaround Linux-only

thopiekar:
Previously removed in https://github.com/Ultimaker/Uranium/pull/188 and moved here.
Before the workaround was applied on every OS, but now we are adding the environment variable only on Linux.


Ghostkeeper:
See the discussion in https://github.com/Ultimaker/Uranium/pull/188

PR #1183, opened 2016-11-26
Fix issues introduced in the feature_find_settings merge

fieldOfView:
This PR fixes two minor issues introduced in the feature_find_settings merge:
* When starting Cura in simple mode, the filter toggle is incorrectly showing up on the "advanced" tab
* A qml error about anchoring

### DIFF
--- a/resources/qml/Sidebar.qml
+++ b/resources/qml/Sidebar.qml
@@ -166,15 +166,6 @@ Rectangle
         anchors.topMargin: UM.Theme.getSize("default_margin").height
     }
 
-    currentModeIndex:
-    {
-        var index = parseInt(UM.Preferences.getValue("cura/active_mode"))
-        if(index)
-        {
-            return index;
-        }
-        return 0;
-    }
     onCurrentModeIndexChanged:
     {
         UM.Preferences.setValue("cura/active_mode", currentModeIndex);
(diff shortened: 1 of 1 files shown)


resources/qml/Sidebar.qml
```diff
@@ -166,15 +166,6 @@ Rectangle
...
 
-    currentModeIndex:
-    {
-        var index = parseInt(UM.Preferences.getValue("cura/active_mode"))
-        if(index)
-        {
-            return index;
-        }
-        return 0;
-    }
```
Ghostkeeper: Wait, but this is still being used at (new) line 262, where you added additional checks?

fieldOfView: Yes. The important bit of this code is moved to Component.onCompleted.
The problem I'm fixing here is that currentModeIndex was being set before the modesListModel was populated.

Ghostkeeper:
Still getting these QML errors that might be caused by a filter such as this:

    qml: TableViewSelection: index out of range
    qml: TableViewSelection: index out of range

But I can't attribute them to this feature. One of my largest gripes with QML is that the error reporting is so bad. There are no more problems with anchoring though, and it doesn't display the magnifying glass on start-up in recommended mode. Still seems to work fine for the rest. Thanks for the fix!

fieldOfView:
I've been trying to track down these TableViewSelection notices since before the 2.3 beta

awhiemstra:
Those messages come from the Preferences dialog and started showing up iirc when changing from Qt 5.4 to Qt 5.5.

PR #1194, opened 2016-11-29
Adding Upgraded 250x150 PB heated Bed

samirabaza:
cura1
Adding Upgraded 250x150 PB heated Bed
STL + definition file


Ghostkeeper:
There's a couple of problems:
* On all supported platforms except Windows, file names are case sensitive. The `.STL` extension in the file names doesn't match up with the `.stl` in the definition.
* The Pheatedbed.STL file is not used at the moment. Please remove it. The other STL file is much heavier, which is undesirable, but still within acceptable bounds.

If you commit to the master branch of your Cura fork, it should get updated here too.

samirabaza:
Ghostkeeper
please review the changes and give me your feedback

Ghostkeeper:
The platform mesh loads properly now. The machine works. One little thing though: The platform mesh seems to be a bit too high, causing the platform to intersect in the build volume. As a result, the entire build plate is grey.

Adding the following line to the metadata of the machine definition seems to work (inserted at line 12):

            "platform_offset": [0, -0.3, 0],

samirabaza:
Noted that Ghostkeeper.
the file is already updated

PR #1195, opened 2016-11-29
Initial support for elliptic build plate shape

fieldOfView:
This PR provides an option to change the build plate shape to elliptic for polar printers.
polar
An elliptic build volume is drawn, and disallowed area polygons are added so that only an elliptic area remains.

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -10,6 +10,7 @@ from UM.Application import Application
 from UM.Resources import Resources
 from UM.Mesh.MeshBuilder import MeshBuilder
 from UM.Math.Vector import Vector
+from UM.Math.Matrix import Matrix
 from UM.Math.Color import Color
 from UM.Math.AxisAlignedBox import AxisAlignedBox
 from UM.Math.Polygon import Polygon
@@ -218,10 +219,16 @@ class BuildVolume(SceneNode):
 
         else:
             # Bottom and top 'ellipse' of the build volume
+            aspect = 1.0
+            scale_matrix = Matrix()
+            if self._width != 0:
+                # Scale circular meshes by aspect ratio if width != height
+                aspect = self._height / self._width
+                scale_matrix.compose(Vector(1, 1, aspect))
             mb = MeshBuilder()
             mb.addArc(max_w, Vector.Unit_Y, center = (0, min_h - 0.2, 0), color = self.VolumeOutlineColor)
             mb.addArc(max_w, Vector.Unit_Y, center = (0, max_h, 0),  color = self.VolumeOutlineColor)
-            self.setMeshData(mb.build())
+            self.setMeshData(mb.build().getTransformed(scale_matrix))
 
             # Build plate grid mesh
             mb = MeshBuilder()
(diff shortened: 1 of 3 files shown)


nallath:
You are making a lot of people happy with this! I'll have a look right away.

resources/definitions/fdmprinter.def.json
```diff
@@ -137,6 +137,21 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_shape":
+                {
+                    "label": "Build plate shape",
+                    "description": "The shape of the build plate without taking unprintable areas into account.",
+                    "default_value": "Rectangular",
```
nallath: This should be "rectangular" (as per key you used in the options).

nallath:
ellipicalbuildplatefail

resources/definitions/fdmprinter.def.json
fieldOfView: Good catch. Fixed

cura/BuildVolume.py
```diff
@@ -177,34 +183,75 @@ def rebuild(self):
...
+            self._grid_mesh = mb.build()
+
+        else:
+            # Bottom and top 'ellipse' of the build volume
+            aspect = 1.0
+            scale_matrix = Matrix()
+            if self._width != 0:
+                # Scale circular meshes by aspect ratio if width != height
+                aspect = self._height / self._width
+                scale_matrix.compose(Vector(1, 1, aspect))
```
fieldOfView: That's what I meant in the first place. Fixed.

fieldOfView:
Eh? Wt..

If the other half is visible from below, that suggests an error in the winding order of the faces more than an issue with UVs. But the winding order is just fine for me. Will see if I can repro this on linux, or on another machine.

fieldOfView:
Euhm, in your screenshot (probably not related to this issue)... How come Dual Extrusion Overlap is showing in the Build Plate Adhesion category, and why is it showing at all for a single extrusion printer?

nallath:
Euh. Good one. I have no idea why that went wrong.

Ghostkeeper:
Good catch.
Turns out that the dual extrusion category header gets hidden for single extrusion, but we don't disable the two settings Dual Extrusion Overlap and Ooze Shield Enabled for single extrusion. Because there is no header, it appears as if the setting is under the Build Plate Adhesion category. Collapsing the Build Plate Adhesion category doesn't hide the Dual Extrusion Overlap setting.

fieldOfView:
Back to the rendering issue with the elliptic buildplate: it looks like on Linux, the indices I set are being ignored, and instead the faces are always drawn as a triangle strip.

fieldOfView:
I was not adding indices correctly. Took a while to figure this out because Windows was rendering it correctly regardless.

nickthetait:
So people can now use Cura 2 with almost any delta printer right? Are there other features that are still needed?

fieldOfView:
* Handling of unreachable areas due to extruder offsets is not as effective as for rectangular printers
* Auto scaling / scale to max does not work very well on elliptic platforms

There may be lingering issues with center_is_zero in the engine (I know there were issues before, and the feature gets next to 0 testing at UM because their printers don't use that feature)

nallath:
Scale to max doesn't work that well on any printer. So not really a loss there.

PR #1198, opened 2016-11-29
Remove incompatible plugins

thopiekar:
All of the plugins where not updated for about half a year. Haven't tested them on my own, but I guess they are all incompatible now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,7 @@ Please checkout [cura-build](https://github.com/Ultimaker/cura-build)
 
 Third party plugins
 -------------
-* [Print Cost Calculator](https://github.com/nallath/PrintCostCalculator): Calculates weight and monetary cost of your print.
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I have tested X3G writer more recently and as far as I know it still works.
Turns out that I locally incremented the API version without having committed it. Did that now.

The Barbarian Plug-in still works as far as I know. Haven't tested it since 2.3.0.

Don't know about the print cost calculator.

nallath:
The barbarian plugin isn't that impressive, so it should work without a hitch.

thopiekar:
What about the Print Cost Calculator? It is now more or less integrated into Cura since we have materials support, isn't it?

fieldOfView:
It would be great if Print Cost Calculator could be deprecated, but the materials support still is not complete; I cannot enter spool weight and cost, and print cost is not calculated.

PR #1204, opened 2016-12-02
QML: Remove name of application

thopiekar:
For me having "Cura" under the logo actually doesn't make sense, as it is quite obvious by the logo that we are talking about _Cura_ here. So in my opinion we can prefix the application version here with "version: " instead.

Untested but it should work (fingers crossed)

### DIFF
--- a/resources/qml/AboutDialog.qml
+++ b/resources/qml/AboutDialog.qml
@@ -40,7 +40,7 @@ UM.Dialog
     {
         id: version
 
-        text: "Cura %1".arg(UM.Application.version)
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
@Appesteijn Anything we should change here with respect to the 'rename' to Ultimaker Cura?

Appesteijn:
Yeah Thomas is right, it is a little bit too much Cura :)

thopiekar:
Haha, well, who knows.. maybe one day we will dance to the name 😅

ChrisTerBeke:
Merged. Not gonna be in 3.0 yet because string freeze.

PR #1231, opened 2016-12-09
Changed urllib import to be more restrictive

Nihlus:
On certain Linux platforms, urllib in py3 no longer accepts a top-level module import. Restricting the imported modules to just "parse" fixes this issue.

This has been tested and is verified working on the following configuration:
Linux Mint 18 x86 Cinnamon


Ghostkeeper:
Good change. Thanks!

Still works on Ubuntu.

PR #1234, opened 2016-12-10
Add definition and bed mesh for Kossel Pro

ex-nerd:
Kossel Pro is a larger version of the Kossel Mini, which is already represented here.

Source for generating these files can be found here:

https://github.com/dotscad/openbeam/blob/master/kossel_pro_build_platform/


Ghostkeeper:
Looks good! Thanks!

PR #1236, opened 2016-12-12
Add open source projects to the About dialog

fieldOfView:
This PR adds a list of projects, their description and licenses to the About dialog, similar to how legacy had it (except this PR adds an actual hyperlink).

The list of projects may have to be expanded, it could very well be that I missed some project(s).

### DIFF
--- a/resources/qml/AboutDialog.qml
+++ b/resources/qml/AboutDialog.qml
@@ -15,7 +15,7 @@ UM.Dialog
     title: catalog.i18nc("@title:window","About Cura")
 
     minimumWidth: 450 * Screen.devicePixelRatio
-    minimumHeight: 500 * Screen.devicePixelRatio
+    minimumHeight: 550 * Screen.devicePixelRatio
     width: minimumWidth
     height: minimumHeight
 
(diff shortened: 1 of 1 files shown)


thopiekar:
Thanks!

I personally find it correct and fair to add all the used projects into the about dialog, but I personally would only add projects, which are needed by the core.
For example Cura could be used without any usage of USB (PySerial) or network connection (Python-ZeroConf). If it was possible I would tend to give our the plugins the possibility to append entries to the dialog. So USB printing will add PySerial to this list unless it is activated, for example.

However, this is just an idea/improvement I see here. Apart from this I'm fine and think it is better then not to mention these projects.

Ghostkeeper:
Should we also include SIP?

Ghostkeeper:
When I try to pull this, it says there are "no common ancestors" and proceeds to download 50MB of commits, and then proceeds to refuse unrelated histories.

fieldOfView:
This seems to have been merged. Shall I close the PR?

Ghostkeeper:
Uh, yeah. I'd expect Github to show it as being "merged"...

PR #1240, opened 2016-12-13
Add machine_shape to delta printer definitions

fieldOfView:
This PR adds the elliptical machine shape to delta printer definitions. If you got it, flaunt it.

While fixing this I also saw and fixed some z-fighting between the build plate mesh and the machine mesh for the Kossel printers.


Ghostkeeper:
> While fixing this I also saw and fixed some z-fighting between the build plate mesh and the machine mesh for the Kossel printers.

Noticed that too. Didn't want to bother the author with such a minor thing. Thanks for fixing that up as well.

PR #1241, opened 2016-12-13
Fixes to Add Machine dialog

fieldOfView:
This PR fixes some minor issues with the Add Machine dialog:

* the last printer in the "other" list is cut off by the machine name field
* the right-end side of the sections can not be clicked
* when switching to another section, an invisible printer stays selected (and QML complains)
* the purpose of the textfield is unclear


fieldOfView:
This has been merged by @Ghostkeeper

Ghostkeeper:
Merged this. I think it was not displaying as such because I had to rebase on master.

PR #1243, opened 2016-12-13
Implement material cost

fieldOfView:
This PR implements the forgotten material cost feature.

In this PR, filament weight per spool and cost per spool are stored per material guid in the preferences (instead of in the material profile). It can be argues that the price (and to a lesser extend the spool size) is not a material property but something that is set eg by the reseller. Furthermore this enables editing these values for readonly materials without affecting the xml profile.

Filament length on the spool as well as cost per meter are calculated on the material view tab.

TODO:
* ~~use cost per meter to calculate print cost on the print stats area~~
* ~~make currency a settable preference~~ fixed

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -217,6 +217,10 @@ class CuraApplication(QtApplication):
         Preferences.getInstance().addPreference("mesh/scale_tiny_meshes", True)
         Preferences.getInstance().addPreference("cura/dialog_on_project_save", True)
         Preferences.getInstance().addPreference("cura/asked_dialog_on_project_save", False)
+
+        Preferences.getInstance().addPreference("cura/currency", "€")
+        Preferences.getInstance().addPreference("cura/material_settings", "{}")
+
         for key in [
             "dialog_load_path",  # dialog_save_path is in LocalFileOutputDevicePlugin
             "dialog_profile_path",
--- a/cura/PrintInformation.py
+++ b/cura/PrintInformation.py
@@ -6,6 +6,7 @@ from PyQt5.QtCore import QObject, pyqtSignal, pyqtProperty, pyqtSlot
 from UM.Application import Application
 from UM.Qt.Duration import Duration
 from UM.Preferences import Preferences
+from UM.Settings import ContainerRegistry
 
 import cura.Settings.ExtruderManager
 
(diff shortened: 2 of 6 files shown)


nallath:
I'd love to add this, but i will need to discuss this first. We're technically in feature freeze right now...

fieldOfView:
Ah damn, I seem to have misnamed the branch; I meant to name it "fix_missing_material_cost" ;-)

fieldOfView:
Since this PR is not going to be merged for 2.4, I might as well mention a couple of deficiencies with the current implementation:
* changing the density or the filament diameter does not immediately update the cost per meter on the Materials page
* changing the cost per meter on the Materials page does not immediately update the print cost in the print specs area on the viewport

Ghostkeeper:
A few things surface in my mind when testing this.

* It's really quite confusing that the cost per meter and the filament length don't update when you update the weight and cost of a spool. The cost of the current print in the main window is less pressing, since it doesn't serve to me as a validation of whether what I just entered is correct-ish.
* I think the tildes `~` at the computed filament length and cost per meter should be removed. At least in the latter case since it already says "(Approx.)" before it. Otherwise the "(Approx.)" should be removed.
* These two fields are editable regardless of the "Edit" button, which is inconsistent with the rest. Why did we have that "Edit" button anyway? If we have the "Edit" button as a sort-of "Apply Changes" mechanism, perhaps you can use that to signal when to update the calculations of filament length and cost per meter.

fieldOfView:
> It's really quite confusing that the cost per meter and the filament length don't update when you update the weight and cost of a spool. 

Yes, as noted above. This needs to be fixed.

> These two fields are editable regardless of the "Edit" button.

This is by design; the user must be able to enter the amount of money spent on the spool; we cannot fill that in for them because the price may be different per reseller.

fieldOfView:
I have fixed the updating of the filament length and cost per meter in the materials view, as well as the in the job specs area. 

Ready for review again.

Ghostkeeper:
Will test tomorrow.

Ghostkeeper:
I sort of found a bug, I think... Sorta. Not sure.

I had a copied material which I had completely modified: BLA (bollylactic acid), copied from PLA. It had a changed material type and a changed name. When I modify the material cost of BLA, it will also modify the cost of PLA accordingly. This is because these preferences-per-material are stored per GUID. That's logical, because the GUID is intended to point to a specific type of filament made by a specific manufacturer, and therefore this manufacturer has a specific cost for his material and all profiles for that material should have the same cost.

However if a user wants to create a wholly new type of material in Cura, such as BLA, then it's a bit annoying. This will be fixed by letting the user modify the GUID, so I don't think we have to tackle it now. That's why I'm still merging it. For the rest it works great. Thanks for the contribution!

fieldOfView:
This is a logical consequence of https://github.com/Ultimaker/Cura/issues/1328, and a lack of a way to create a new material (as opposed to duplicating an existing material) and/or creating a new GUID for a material. For me this is a separate issue.

PR #1246, opened 2016-12-14
Fix version upgrade with user variants

fieldOfView:
This PR fixes three related issues with the 2.3 to 2.4 version upgrade when there are Machine Settings.

Steps to reproduce the issues:
* Launch Cura 2.3.1 with an empty config
* Add a UM2
* Open machinesettings
* Close Cura 2.3.1 and launch Cura 2.4

Result: the UM2 cannot be loaded, because the versionupgrade screws up.


sedwards2009:
I think these changes are only going to make things better, so I see no reason why not to take them in now.

sedwards2009:
..and thanks.

PR #1265, opened 2016-12-23
Fix scale of Add Machine and Select (Per Object) Settings dialogs...

fieldOfView:
... on fairly high, but non-"retina" screens.

I have a new laptop with a 13" 1080p display, which I use at its native resolution. This works fine for most dialogs, except the mentioned two which are shown comically large.


Ghostkeeper:
It looks the same on both my screens, but we have no high density screens to test the actual issue with.

Regardless, I'm going to merge it on the basis that it should improve the display in theory and that it doesn't break what isn't broken yet, so it can only improve things.

PR #1260, opened 2016-12-20
Tweak the hovered state of the machine selection

fieldOfView:
This PR tweaks the theming of the machine selection button to match the other controls in the header bar. It irks me that the machine selection widget behaves differently from the other two buttons. It is a purely cosmetic change.

PR #1266, opened 2016-12-23
Add a margin to the highlight rect, making it more symmetrical

fieldOfView:
This is a purely cosmetic tweak which makes the hover effect slightly more distinct, and balance more evenly; the right-most mode-button also has a margin to the right as opposed to hugging the side of the window.

Sorry I did not catch this in one go with this PR: https://github.com/Ultimaker/Cura/pull/1260

### DIFF
--- a/resources/qml/Sidebar.qml
+++ b/resources/qml/Sidebar.qml
@@ -131,6 +131,7 @@ Rectangle
                             anchors.left: parent.left
                             anchors.right: parent.right
                             anchors.bottom: parent.bottom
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is how it looks after the change:
adjustedhighlightbar

I'm not sure I agree with this change because it only adjusts the highlight bar. This makes the machine name align exactly with the left side of the bar without margin, which makes it less symmetrical in some way. I'm curious to hear what other people think.

Issue #1155, opened 2016-11-17
Request for code review of new feature: GCode Reader


nickthetait:
For ~2 months Aleph Objects has been working on the ability to import and render GCode files in Cura. At this point the feature has matured to a usable level so we conducted an internal code review. I would like to hear Ultimaker's input before we start refactoring and cleanup.

Here is the code review page: https://code.alephobjects.com/D5.
Our tracking task which covers the history of design and implementation: https://code.alephobjects.com/T466
If you would like to give it a spin, checkout the GCODEReader branches of https://code.alephobjects.com/diffusion/CT/ and https://code.alephobjects.com/diffusion/U/

**Screen shots**
Reference Benchy model
![Reference benchy model](https://cloud.githubusercontent.com/assets/11818614/20401940/f02fed32-acb7-11e6-910b-c86599dbeaa1.png)

Imported GCode
![Imported GCode](https://cloud.githubusercontent.com/assets/11818614/20401948/fa7450d0-acb7-11e6-9faf-db5e51b61bc5.png)

Inspecting some layers
![Inspecting some layers](https://cloud.githubusercontent.com/assets/11818614/20401953/012c6f84-acb8-11e6-9962-c279bbed73f2.png)

Inspecting a single layer
![Inspecting a single layer](https://cloud.githubusercontent.com/assets/11818614/20401957/0882d89a-acb8-11e6-8b08-d1fc23cd7470.png)

**Features**
* Robust: It is more generic than Legacy Cura's implementation, allowing it to work with the output from Cura (Legacy and 2.x) and even Slic3r. It may cope with files produced by proprietary slicers, but this has not been tested.

* Performance: my non-scientific test shows Cura 2 (in a development environment) imports/renders ~2x faster than Legacy Cura (pre-built and installed) using a regular sized Benchy

* Drag & drop works

* Only one GCode file can be imported at a time

* Pause the slicer + explanation to user why GCode files can't be re-sliced

* Display filename in the usual place

* Probably others that I forgot :sleeping:

**Known deficiency**
This feature is based on a slightly out-dated (pre 2.3) codebase. Needs to be ported over to master branch.

**Potential TODOs**
Additional features could make it more useful. Many of these require some information which is not (yet) provided in the GCode file such as nozzle size, filament density, or E Steps.
* Beautification and integration with UI theme system :nail_care:
[...]

Issue #577, opened 2015-12-23
[2.1] Open file dialog: *.gcode not in list of supported files


thopiekar:
While trying to open sample gcode provided by bq on their homepage for the Hephestos 2, I noticed that the "open file" has .gcode not in the list of supported files.
Should be shown when the needed plugin is enabled.

Ghostkeeper:
Directly loading g-code from file was dropped with the rewrite between version 15.04 and 15.06. It might get re-inserted later but it's currently a regression.

The primary reason it's not in yet is because due to performance issues the layer view currently shows the layers per-layer as flat planes. When showing the g-code such an optimisation is impossible since the g-code of a layer doesn't need to be flat.

fieldOfView:
It is not just that the filedialog does not list .gcode as a input format; Cura cannot parse gcode at this point. Layerview is not implemented as a gcode viewer, but it gets data specially prepared/formatted for layerview. This is not a trivial thing to fix.

nallath:
I've put a bit of effort into it at some point and it's not that hard. The main problem would be making it fast. You would need to make a GCodeReader plugin that creates polygon data for the layerview to display.

nallath:
This is fixed.

PR #1267, opened 2016-12-23
Add a Gcode reader plugin

kakaroto:
This is part of the GCodeReader work (https://code.alephobjects.com/T466 and https://github.com/Ultimaker/Cura/issues/577) which requires the ability to open and parse GCode files.
Review was done here : https://code.alephobjects.com/D7 and has already been approved as can be seen here : https://github.com/Ultimaker/Cura/issues/1155
This work has been rebased on the latest ultimaker/master and modified to fit the latest changes. It has been tested after the rebased and proved to be functional.
Please note that this depends on this Uranium PR being merged first : https://github.com/Ultimaker/Uranium/pull/200


Ghostkeeper:
I found a few bugs in this implementation.

A: I'm getting this error when I try to load a g-code file:

    2016-12-27 11:11:26,442 - CRITICAL - cura.CrashHandler.show [32]: An uncaught exception has occurred!
    2016-12-27 11:11:26,444 - CRITICAL - cura.CrashHandler.show [35]: Traceback (most recent call last):
    2016-12-27 11:11:26,444 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Cura/cura/CuraApplication.py", line 499, in event
    2016-12-27 11:11:26,444 - CRITICAL - cura.CrashHandler.show [35]:     return super().event(event)
    2016-12-27 11:11:26,445 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Qt/QtApplication.py", line 192, in event
    2016-12-27 11:11:26,445 - CRITICAL - cura.CrashHandler.show [35]:     event._function_event.call()
    2016-12-27 11:11:26,445 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Event.py", line 157, in call
    2016-12-27 11:11:26,446 - CRITICAL - cura.CrashHandler.show [35]:     self._function(*self._args, **self._kwargs)
    2016-12-27 11:11:26,446 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Signal.py", line 150, in emit
    2016-12-27 11:11:26,446 - CRITICAL - cura.CrashHandler.show [35]:     func(dest, *args, **kwargs)
    2016-12-27 11:11:26,447 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Cura/cura/CuraApplication.py", line 1091, in _readMeshFinished
    2016-12-27 11:11:26,447 - CRITICAL - cura.CrashHandler.show [35]:     op.push()
    2016-12-27 11:11:26,447 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Operations/Operation.py", line 50, in push
    2016-12-27 11:11:26,448 - CRITICAL - cura.CrashHandler.show [35]:     Application.getInstance().getOperationStack().push(self)
    2016-12-27 11:11:26,448 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Operations/OperationStack.py", line 52, in push
    2016-12-27 11:11:26,448 - CRITICAL - cura.CrashHandler.show [35]:     operation.redo()
    2016-12-27 11:11:26,448 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Operations/AddSceneNodeOperation.py", line 35, in redo
    2016-12-27 11:11:26,449 - CRITICAL - cura.CrashHandler.show [35]:     self._node.setParent(self._parent)
    2016-12-27 11:11:26,449 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Scene/SceneNode.py", line 254, in setParent
    2016-12-27 11:11:26,449 - CRITICAL - cura.CrashHandler.show [35]:     scene_node.addChild(self)
    2016-12-27 11:11:26,450 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Scene/SceneNode.py", line 308, in addChild
    2016-12-27 11:11:26,451 - CRITICAL - cura.CrashHandler.show [35]:     scene_node._transformChanged()
    2016-12-27 11:11:26,451 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Scene/SceneNode.py", line 605, in _transformChanged
    2016-12-27 11:11:26,452 - CRITICAL - cura.CrashHandler.show [35]:     self.transformationChanged.emit(self)
    2016-12-27 11:11:26,452 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Signal.py", line 154, in emit
    2016-12-27 11:11:26,452 - CRITICAL - cura.CrashHandler.show [35]:     signal.emit(*args, **kwargs)
    2016-12-27 11:11:26,453 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Signal.py", line 154, in emit
    2016-12-27 11:11:26,453 - CRITICAL - cura.CrashHandler.show [35]:     signal.emit(*args, **kwargs)
    2016-12-27 11:11:26,453 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Signal.py", line 150, in emit
    2016-12-27 11:11:26,454 - CRITICAL - cura.CrashHandler.show [35]:     func(dest, *args, **kwargs)
    2016-12-27 11:11:26,454 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Cura/cura/../plugins/CuraEngineBackend/CuraEngineBackend.py", line 327, in _onSceneChanged
    2016-12-27 11:11:26,454 - CRITICAL - cura.CrashHandler.show [35]:     self.pauseSlicing()
    2016-12-27 11:11:26,455 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Cura/cura/../plugins/CuraEngineBackend/CuraEngineBackend.py", line 196, in pauseSlicing
    2016-12-27 11:11:26,455 - CRITICAL - cura.CrashHandler.show [35]:     self.backendStateChange.emit(BackendState.Disabled)
    2016-12-27 11:11:26,455 - CRITICAL - cura.CrashHandler.show [35]:   File "/usr/lib/python3.5/enum.py", line 274, in __getattr__
    2016-12-27 11:11:26,455 - CRITICAL - cura.CrashHandler.show [35]:     raise AttributeError(name) from None
    2016-12-27 11:11:26,456 - CRITICAL - cura.CrashHandler.show [35]: AttributeError: Disabled

It does load the g-code file though. What I expect is the result is that the slicing widget now says it's preparing to slice, indefinitely. This also prevents me from saving the g-code to a USB stick, sending it via a USB cable or printing via wifi.

B: Layer numbers are misaligned. Layer 1 is empty. Layer 2 has the g-code with the brim and such that is normally the first layer. And so on.

C: Line thickness seems to be too thick, at least for the Ultimaker 3. I see Z-fighting between the inner walls and the outer walls, which indicates that they overlap. Also, travel moves are way thicker than what they normally are.

D: Loading a g-code file shows two progress bars, one indefinite saying that it's loading the .gcode file and one actual progress bar that says it's parsing the g-code. I realise though that this is probably a limitation of Uranium's structure so I won't consider this a real problem if you can't fix it easily.

E: Multiply models doesn't work any more. It gives this error:

    2016-12-27 11:23:34,212 - CRITICAL - cura.CrashHandler.show [32]: An uncaught exception has occurred!
    2016-12-27 11:23:34,214 - CRITICAL - cura.CrashHandler.show [35]: Traceback (most recent call last):
    2016-12-27 11:23:34,214 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Cura/cura/CuraApplication.py", line 677, in multiplyObject
    2016-12-27 11:23:34,215 - CRITICAL - cura.CrashHandler.show [35]:     new_node = copy.deepcopy(current_node)
    2016-12-27 11:23:34,215 - CRITICAL - cura.CrashHandler.show [35]:   File "/usr/lib/python3.5/copy.py", line 166, in deepcopy
    2016-12-27 11:23:34,216 - CRITICAL - cura.CrashHandler.show [35]:     y = copier(memo)
    2016-12-27 11:23:34,216 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Scene/SceneNode.py", line 89, in __deepcopy__
    2016-12-27 11:23:34,216 - CRITICAL - cura.CrashHandler.show [35]:     copy.addDecorator(deepcopy(decorator, memo))
    2016-12-27 11:23:34,216 - CRITICAL - cura.CrashHandler.show [35]:   File "/usr/lib/python3.5/copy.py", line 166, in deepcopy
    2016-12-27 11:23:34,216 - CRITICAL - cura.CrashHandler.show [35]:     y = copier(memo)
    2016-12-27 11:23:34,217 - CRITICAL - cura.CrashHandler.show [35]:   File "/home/ruben/Projects/Uranium/UM/Scene/SceneNodeDecorator.py", line 24, in __deepcopy__
    2016-12-27 11:23:34,217 - CRITICAL - cura.CrashHandler.show [35]:     raise NotImplementedError("Subclass {0} of SceneNodeDecorator should implement their own __deepcopy__() method.".format(str(self)))
    2016-12-27 11:23:34,217 - CRITICAL - cura.CrashHandler.show [35]: NotImplementedError: Subclass <cura.SliceableObjectDecorator.SliceableObjectDecorator object at 0x7f3bead28748> of SceneNodeDecorator should implement their own __deepcopy__() method.

This happens when loading a _normal_ mesh, not g-code, but it happens only on your branch. Master is fine.

F: Reloading all models doesn't reload the g-code. Not really an issue if it doesn't, but it might be unexpected for the user.

G: Dual extrusion offset is not taken into account. I understand though that this was a limitation of the information that was knowable and can't be fixed yet until we change the g-code header format?

H: The print job name becomes "Pre-sliced file test". I'd expect it to get the name of the file, just like other models. Perhaps it shouldn't be prefixed with the printer name though.

Ghostkeeper:
Did some more testing with a few things I realised later might be a problem.

I: Retracted travel moves are shown in the normal colour rather than the light blue colour we normally use.

J: Z-hop causes each piece to be recognised as a separate layer. This will be a difficult one to fix if you're recognising layers by the Z height. Prime towers are also a big problem then.

K: The search icon and part of the Advanced/Custom Mode button is still visible on the settings panel.

nickthetait:
Kakaroto is out on vacation this week, so he'll respond after he gets back...

kakaroto:
Hi @Ghostkeeper,
A quick drop from vacation after seeing these responses, thanks for testing!

A: The bug is probably because you hadn't merged the change in Uranium too as noted in the PR description.
That Uranium PR was merged already so just update Uranium and that bug should disappear.

B: This is probably because the layer 1 contains the move after homing, since each moves in a specific Z height is considered as a separate layer.

C: This was discussed here : https://code.alephobjects.com/D5#inline-47 
The line_width is hardcoded to 0.5. If you think a different/better value is needed, we can change that. I agree that moves should have a thinner line. I'm not sure though about actually calculating the line width based on E values, as it doesn't sound trivial.

D: Yes, I found it annoying as well, but I don't think we have a choice. I believe the second indefinite progress bar is created by Uranium for the "Loading files" and the GCodeReader plugin only creates that second one. If it's possible to have an API in Uranium to control that first progress bar, then it might be better to have the plugin simply set the progress bar value on it rather than create its own progress bar, but that would only work for the single-file-load of GCode files, it wouldn't be useful for the STL file loading plugin which can load multiple files at the same time. The Gcode progressbar is also used for cancelling a load operation, I don't know how much change would be needed to handle cancellation from the Uranium-based progress bar.

E: I have no idea what's with this deepcopy issue. maybe @victor9999 can comment on it ? It definitely does seem like something is missing and needs to be fixed before the merge.

F: I don't think reloading is handled by the plugin, I think it comes from Cura/Uranium and I'm guessing that reloading is just a fancy name for "reset all transformations", but since gcode models can't be moved/scaled/etc.. that won't do anything.

G: Yeah, this was discussed before, and I guess this needs a new issue to be created for it. Unless you have a suggestion to improve things right now in the current implement. Your last comment here https://github.com/Ultimaker/Cura/issues/1155#issuecomment-261630756 implies that we should leave the offset to Marlin (which means it wouldn't be backward-compatible), as for the header, it's UM specific, so wouldn't work for all machines either.

H: I think the file you used was called "test.gcode" :) It does use the name of the file, it just prefixes it with "Pre-sliced file". 

I: retraction moves are using the type : `LayerPolygon.MoveRetractionType ` and I don't see any other retraction types in cura/LayerPolygon.py that could be used for the retraction moves. the color of the lines are determined by that type value.

J: Yes, that might be annoying. The problem is that we had to make each Z change into its own layer, this is because of sequential printing. It was discussed here: https://code.alephobjects.com/T466#7323
I suppose two solutions are available to us here:
* If we find the ';LAYER' comment in the file (CuraEngine-generated gcode), then we could ignore Z changes and parse the entire file basing our layers on the value of the ';LAYER' comments.
* We do not consider ourselves on a new layer unless there is an actual extrusion on that Z offset. That should probably eliminate issues with the Z-hop, and would also probably fix the 'start.gcode move counting as layer 1' issue you mentioned in 'B'.

K: Ah that's a good one! Was hard for me to understand what you meant until I played around with the interface and reproduced it. 
@victor9999 If before loading the gcode, you select 'Custom' in the Print setup, then there's a search button that appears next to 'Custom', if you then load the gcode file, that search icon will remain even though the print setup has been disabled.

nallath:
Reloading is completely borked right now. It should do more than just reset the transformation, it should load the entire file again. The problem is what should be done if there are more nodes in a single file (3mf for instance).

kakaroto:
That makes sense, you export stl from openscad for example, then want to reload, so it needs to re-read the file. 
I'm still assuming that Cura would be the one handling the reload, not the plugin itself. As for the multi-node stuff, I guess nodes should have a single parent node or at least the reload function would build the list of files to reload as a set and multiple nodes would have the same 'file' origin. Shouldn't be much of a problem.

kakaroto:
@Ghostkeeper 
I've just pushed a couple of commits from @victor9999 that address most of the issues discussed here.
B: and J: Now if the parser finds ';LAYER' comments in the gcode file, it will only create a new layer if it finds that comment during parsing. Current behavior is kept if the file has no such comment (slic3r generated files for example). This should fix the z-hop issue you mentioned, and will also eliminate the start.gcode post-homing move that counted as the first layer, so the first layer now does contain the actual first layer of the print.

C: line width has now been set to 0.4 and the line width during moves has been set to 0.2 so they'd be thinner than extrusions.

E: @victor9999 fixed the deepcopy issue.

K: the search button is now hidden.

I believe these are all the issues that were remaining to be fixed.

Please review these latest commits to make sure it's all done properly, since those commits have not been reviewed before by anyone on your team.

Thanks!

plugins/GCodeReader/GCodeReader.py
```diff
@@ -0,0 +1,303 @@
...
+            return None
+
+    def _getFloat(self, line, code):
+        value = self._getValue(line, code)
+        try:
+            return float(value)
+        except:
+            return None
+
+    def _onHideMessage(self, message):
```
awhiemstra: User-interaction wise, it is not really nice to abort the loading when closing the message, since it is not very explicit. We have support for explicit actions on Messages, so it is better to add an explicit "Cancel" action to the message and not show the close button.

awhiemstra:
Aaand its merged. Thanks a lot for your work!

kakaroto:
Thanks for the smooth review process and the merge!

PR #1274, opened 2016-12-28
Add infill_angles setting to control direction of lines and zig zag infill.

smartavionics:
This setting provides 4 options:

Default - the original directions, 45 and 135.

Uniform - one each of 45, 90, 135 and 180.

Stronger X - 45, 90, 135 and 90 again - provides more strength in X direction.

Stronger Y - 45, 0, 135 and 0 again - provides more strength in Y direction.

I really wanted to be able to enter the comma separated lists of angles into a text box in the gui but the limits on length and validation mess that up so, instead, I am using these fixed options which cover my requirements at least.


Ghostkeeper:
This belongs to https://github.com/Ultimaker/CuraEngine/pull/427

Changes here look good.

BagelOrb:
@Ghostkeeper isn't it possible to have a string setting, so that the user isn't limited to the combinations defined here?

@smartavionics I think your predefined settings are a bit weird.
The order of uniform is better like this: `45, 135, 90, 0`
Stronger X and Stronger Y are actually less strong than the orthogonal option; while the X/Y is in the same direction every odd layer, the even layers alternate, meaning that the infill would hang 4 layers above the previous infill line along that direction, which makes it weaker than the orthogonal option. In contrast the orthogonal option always has infill lines which can rest on two layers below.

Also the engine uses an empty list as the default fallback to normal behaviour.

All in all I think you should have:
```
"": "Standard",
"45,135": "Diagonal",
"0,90": "Axis Aligned",
"45,90,135,180": "Demi-Cardinal",
"90": "Along X",
"0": "Along Y"
```

Note that some changes I requested on the engine side may have repercussions on the `enabled` value in the Cura frontend.

smartavionics:
I wanted to use a string setting but the UI doesn't let you enter a comma separated list of numbers.

A couple of thoughts about your new settings list above.

1 - standard and diagonal give the same pattern (maybe that's what you want)

2 - remember that you are not going to be using the stronger in X/Y options with 10% infill, the infill amount is more likely to be 80%+ (I am currently using 90%) so all of the infill is very well supported. I can assure you that the parts are stronger using the angles I chose.

BagelOrb:
Yeah the standard option should be removed. I thought the behaviour was
different when you are using e.g. triangles infill, but I'm not sure.

Ah yes, in the case of > 50% infill it will be stronger.

These names are not really transparent, though. I really think we should
allow for string settings rather than doing it this way.


Op 6 jan. 2017 16:00 schreef "Mark Burton" <notifications@github.com>:

I wanted to use a string setting but the UI doesn't let you enter a comma
separated list of numbers.

A couple of thoughts about your new settings list above.

1 - standard and diagonal give the same pattern (maybe that's what you want)

2 - remember that you are not going to be using the stronger in X/Y options
with 10% infill, the infill amount is more likely to be 80%+ (I am
currently using 90%) so all of the infill is very well supported. I can
assure you that the parts are stronger using the angles I chose.

—
You are receiving this because you commented.
<https://github.com/Ultimaker/Cura/pull/1274#issuecomment-270920639>, or mute
the thread

smartavionics:
Yes, a string setting would be best as then people can use whatever angles they think are best.

Sorry, I am not familiar with the UI framework used by Cura and I don't really do much python programming so I am not in a good position to investigate adding the capability of having a string setting option in the UI. Can you provide some help with this? (or at least provide a hint as to how it should be done?)

BagelOrb:
The same holds for me. I'm a C++ programmer and not too familliar with the frontend.

Perhaps @sedwards2009 , @Ghostkeeper , @nallath or @jackha can help?
They should be able to point you to the right places for adding a string type setting without any error handling.

smartavionics:
Understood. I would have thought that adding an unchecked string entry widget to the UI could be useful for other reasons in the future so, hopefully, there will be some enthusiasm for making that happen. In the meantime, I will keep using this PR and not worry about it being integrated. Cheers.

BagelOrb:
Let's just keep this open as a reminder to not forget this feature.

smartavionics:
Hello again. I have worked out what was required to allow the Infill Line Directions setting to be an arbitrary string. It needed the maximum length of the setting string to be increased and the regexp validator changed when the setting type is "str" (as opposed to "int", "float", etc.)

BagelOrb:
Awesome!

I couldn't imagine it would be difficult, but I don't know much about frontend code.

I think we should merge this PR and the corresponding CuraEngine PR (https://github.com/Ultimaker/CuraEngine/pull/427)

We just need to review the code and test it. (CURA-3275 on our internal issue tracker)

Also we're keeping a list of community contributions. How much work would you say you have put in? I would estimate it around 16 hours; what do you think?

smartavionics:
Great, thanks.
I haven't kept track of the time I have spent on this but that's a reasonable estimate, probably not more than that.

fieldOfView:
Note that as is, there is no validation of the input whatsoever. No warnings, no errors. Can the engine handle just any input on this setting?

With infinite power comes infinite responsibility ;-)

BagelOrb:
Hmm good point. I think the engine will crash with arbitrary input like `^%@$`

BagelOrb:
Actually I don't think so when looking at the code.
We'll just have to try.

BagelOrb:
It just defaults to normal behaviour then.

smartavionics:
Hi, that's deliberate because the text setting type is a rather generic "str" so I didn't want to have a regexp that checked for a comma separated list of numbers. We could introduce a new type, i.e. "number-list" or similar and then the UI could check the setting for validity.

As it stands, the engine parses the list and it expects a comma separated list of numbers (whitespace is ignored). If it comes across a bad character, it gives up and just uses what it has found so far. If the list is empty, it uses default values of 45, 135 (just like old times).

fieldOfView:
> We could introduce a new type, i.e. "number-list" or similar

Or we could introduce a property "validator_pattern" in SettingDefinition, which could be used to set a custom validator for a setting.

smartavionics:
> Or we could introduce a property "validator_pattern" in SettingDefinition, which could be used to set a custom validator for a setting.

Yes, that sounds the way to go because it could be used for other settings in the future.

BagelOrb:
Interesting stuff, but I don't think we should put all that weight on Marks
shoulders.

Let's approve this PR as it is and create new issues when we feel it's time
to improve on this kind of stuff.

Op 17 jan. 2017 12:07 schreef "Mark Burton" <notifications@github.com>:

smartavionics:
I agree with Tim, not because I am lazy(!) but because introducing that new validator_pattern needs some consideration which is kinda orthogonal to the current PR. Sure, in the future, the infill directions setting could benefit from using such a feature in which case it can be updated to use it.

fieldOfView:
Agreed. This can be fixed in a followup PR

Ghostkeeper:
Having a sort of "list" as setting data type would allow us to create an interface element for it though.

smartavionics:
So to validate it would be zero or more strings (containing any characters except commas) separated by commas?

Ghostkeeper:
I'd leave the regex validator as it is right now. That is more applicable to strings in general. If we have a list or list-of-ints type one day we can correct the regex for it then.

smartavionics:
Now this setting can be used with any of the infill patterns. The default value has been changed to be the
empty string which the engine will understand to mean use the traditional default angles.

smartavionics:
I've added the [int] setting type and also a new setting skin_angles (aka "Top/Bottom Line Directions").

smartavionics:
BTW, it's pretty much impossible to discern the direction of the skin lines because the Cura display draws the skin with no gaps between the lines. It would be better if there was a tiny gap between the lines.

Ghostkeeper:
> BTW, it's pretty much impossible to discern the direction of the skin lines because the Cura display draws the skin with no gaps between the lines.

We're fixing that in Cura v2.5. The result can currently be seen in the `layer_view3_cleanup` branch.

resources/definitions/fdmprinter.def.json
```diff
@@ -1089,6 +1097,14 @@
                     "value": "'lines' if infill_sparse_density > 25 else 'grid'",
                     "settable_per_mesh": true
                 },
+                "infill_angles":
+                {
+                    "label": "Infill Line Directions",
+                    "description": "A list of integer line directions to use. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the traditional default angles (45 and 135 degrees for the lines and zig zag patterns and 45 degrees for all other patterns).",
+                    "type": "[int]",
+                    "default_value": "[ ]",
```
BagelOrb: This setting should not be enabled for the Cubic Subdivision infill type. I don't think you made the changes to Cubic Subdivision infill for it to apply the infill angles (and I don't think you should).

resources/definitions/fdmprinter.def.json
```diff
@@ -908,6 +908,14 @@
                     "value": "top_bottom_pattern",
                     "settable_per_mesh": true
                 },
+                "skin_angles":
+                {
+                    "label": "Top/Bottom Line Directions",
+                    "description": "A list of integer line directions to use when the top/bottom layers use the lines or zig zag pattern. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the traditional default angles (45 and 135 degrees).",
+                    "type": "[int]",
+                    "default_value": "[ ]",
```
BagelOrb: These settings should both only apply for linear patterns. You need something like `"enabled": "skin_pattern != 'concentric'"`

For the infill patterns that the angles don't apply for Concentric infill, Concentric 3D infill and cubicsubdiv.

smartavionics:
OK, now disables those settings as per your requests.

BagelOrb:
Ready to merge. Looking forward to testing!

resources/qml/Settings/SettingTextField.qml
```diff
@@ -98,9 +98,9 @@ SettingItem
 
             selectByMouse: true;
 
-            maximumLength: 10;
+            maximumLength: (definition.type == "[int]") ? 20 : 10;
 
-            validator: RegExpValidator { regExp: (definition.type == "int") ? /^-?[0-9]{0,10}$/ : /^-?[0-9]{0,9}[.,]?[0-9]{0,10}$/ } // definition.type property from parent loader used to disallow fractional number entry
+            validator: RegExpValidator { regExp: (definition.type == "[int]") ? /^\[(\s*-?[0-9]+\s*,)*(\s*-?[0-9]+)\s*\]$/ : (definition.type == "int") ? /^-?[0-9]{0,10}$/ : /^-?[0-9]{0,9}[.,]?[0-9]{0,10}$/ } // definition.type property from parent loader used to disallow fractional number entry
```
BagelOrb: I can't fill in a comma or a semi-colon, so I can't really test this :/
@nallath could you help with getting this to work?

nallath: Uh, using a , in a int is not possible. Why would you even want to?

BagelOrb: It's not an `int` it's a new type called `[int]`: a comma separated list of integers.

BagelOrb: We made a new validator, but how do we change what characters the field accepts?

Ghostkeeper: The regex for integers is `-?[0-9]{0,10}`, as stated in the second option of the double ternary operator. The 0 in `{0,10}` is to allow removing the number altogether. It should actually be 1. The 10 should actually be infinite but we disallow numbers with more than 10 digits. If it were `{1,infinite}` that would be equivalent to `+`. So that becomes `-?[0-9]+`.

From this follows that the regex for comma separated lists of integers would be `(-?[0-9]+,)*(-?[0-9]+)`.

Allow for whitespace everywhere, and it becomes `(\s*-?[0-9]+\s*,)*(\s*-?[0-9]+)\s`.

Now enclose this in square brackets and it becomes `^\[(\s*-?[0-9]+\s*,)*(\s*-?[0-9]+)\s\]$`. This is exactly the regex that smartavionics got. The only difference I'd make is to limit the integer number length again.

Ghostkeeper: It seems that the problem is that this only accepts proper final values, not intermediary values while the user is typing. The only reason why the brackets are not a problem is because the default value is an empty lists with pre-filled brackets. If the user would start off without brackets, he could never type anything.

I think the solution is to allow zero-length integers in this field, then interpret that as the number 0 (or erase it from the list). Actually replacing all the `+` in the regex with `{0,10}` would fix it.

smartavionics:
Well, for me, using the code in my PR, I can type integer values separated by commas with optional whitespace and enclosed in square brackets OK. Personally, I would ditch the square brackets but that's just a matter of taste. What's the problem people are seeing?

Ghostkeeper:
The problem we were seeing is that you can't type `[10, 20, 30]` because the intermediary value while you're typing is `[10,]` and this is not an accepted state by that regex. So it blocks the comma when you go from `[10]` to `[10,]`.

smartavionics:
Hi, well that's weird, because I've just tested doing exactly that and have no problems. The regexp is
/^\[?(\s*-?[0-9]{0,9}\s*,)*(\s*-?[0-9]{0,9})\s*\]?$/

smartavionics:
Ah, but there is another problem because the regexp allows the leading [ to be missing but the backend code that parses the list requires the leading [ to be present. So, we either need to make the [ optional in the back end or enforce it in the qml regexp.

resources/qml/Settings/SettingTextField.qml
BagelOrb: I changed the regex to do exactly that. I also allow for removing the brackets.

BagelOrb:
Oh I've tested the backend by omitting the brackets and it seemed to work fine.
Let's just leave it like this and see whether our testers or users can actually break it - I couldn't.

PR #1288, opened 2017-01-02
Layer View 3D, with detection of OpenGL context versions.

jackha:
Summary:
- Data for Layer view has extra info: line_dimensions, extruders, line_types
- Removed misusing color alpha channel for extruder number. Extruders is now a new separate attribute.
- Compatibility mode for layer view if OpenGL context does not meet requirements (or force_layer_view_compatibility_mode). Compatibility mode is based on old layer view, but with a few more features.
- Minimum layer + maximum layer.
- Checkboxes to toggle layer types (travel moves, support, adhesion, skin, infill), and per extruder.
- Needs updates from Uranium https://github.com/Ultimaker/Uranium/pull/204

OpenGL 4.1 core and 2.0:
- Added 4.1 core shaders.
- Layer view compatibility mode uses OpenGLContext.isLegacyOpenGL and preference force_layer_view_compatibility_mode


fieldOfView:
A minor suggestion: please try to use meaningful commit messages, even in private and experimental branches. At some point this code is going to be merged into master, and all the "wip", "cleanup" etc commit messages will become meaningless.

jackha:
No problem, it was a minor addition.

fieldOfView:
Finally took some time to checkout the branch. Nice work!

The dual layer slider was a surprise; its a bit of an engineer's design ;-). A functionality that I miss with the dual sliders is being able to select - say - a 5 layer thick "stack" and move it up/down like before.

Here's a quick sketch of an alternative, where both upper and lower limit share a single track:
slider
You could make the "inbetween" grey area draggable as well, moving both of the limits together. If you make it so that the upper track is draw a bit higher than the lower track, you can make it so that there is still a bit of grey left when you have a single layer selected. Taking it one step further would be to have the layer number box move to the limit "nub" you touched last (upper, middle, lower), so it is also possible to enter exact numbers for those.

fieldOfView:
For the curious... Here's the state of things:
state-of-things

Ghostkeeper:
Love the stacked layer view slider idea!

jackha:
@fieldOfView Thanks for the comments. The dual slider is indeed my engineers design. A new proposal updates the slider as well as some other elements. Your proposal with the draggable 5 layer stack can also be incorporated :-)

cura/LayerDataBuilder.py
```diff
@@ -48,28 +48,63 @@ def setLayerThickness(self, layer, thickness):
 
         self._layers[layer].setThickness(thickness)
 
-    def build(self):
+    #   material color map: [r, g, b, a] for each extruder row.
+    #   line_type_brightness: compatibility layer view uses line type brightness of 0.5
+    def build(self, material_color_map, line_type_brightness = 1.0):
```
nallath: Could you use doxygen style markup here? Makes it a bit more consistent (use \param & \type{dict} etc)

cura/LayerDataBuilder.py
```diff
@@ -48,28 +48,63 @@ def setLayerThickness(self, layer, thickness):
...
 
         self.addVertices(vertices)
+        colors[:, 0:3] *= line_type_brightness
         self.addColors(colors)
         self.addIndices(indices.flatten())
 
+        material_colors = numpy.zeros((line_dimensions.shape[0], 4), dtype=numpy.float32)
+        for extruder_nr in range(material_color_map.shape[0]):
+            material_colors[extruders == extruder_nr] = material_color_map[extruder_nr]
+        material_colors[line_types == LayerPolygon.MoveCombingType] = colors[line_types == LayerPolygon.MoveCombingType]
```
nallath: As just discussed; Numpy arrays work differently. Please add a comment here to indicate that.

cura/LayerPolygon.py
```diff
@@ -19,13 +19,16 @@ class LayerPolygon:
     
     __jump_map = numpy.logical_or(numpy.logical_or(numpy.arange(11) == NoneType, numpy.arange(11) == MoveCombingType), numpy.arange(11) == MoveRetractionType)
     
-    def __init__(self, mesh, extruder, line_types, data, line_widths):
+    ##  LayerPolygon
+    #   line_thicknesses: array with type as index and thickness as value
```
nallath: Document what the parameters mean.

cura/LayerPolygon.py
```diff
@@ -60,12 +63,11 @@ def buildCache(self):
...
         numpy.logical_and(self._build_cache_needed_points, self._build_cache_line_mesh_mask, self._build_cache_needed_points )
-        
+
         self._vertex_begin = 0
         self._vertex_end = numpy.sum( self._build_cache_needed_points )
         
-
-    def build(self, vertex_offset, index_offset, vertices, colors, indices):
+    def build(self, vertex_offset, index_offset, vertices, colors, line_dimensions, extruders, line_types, indices):
         if (self._build_cache_line_mesh_mask is None) or (self._build_cache_needed_points is None ):
```
nallath: You didn't write this, but the () are unneeded. Might as well remove them.

cura/LayerPolygon.py
```diff
@@ -60,12 +63,11 @@ def buildCache(self):
...
         # Mark points as unneeded if they are of types we don't want in the line mesh according to the calculated mask
         numpy.logical_and(self._build_cache_needed_points, self._build_cache_line_mesh_mask, self._build_cache_needed_points )
-        
+
         self._vertex_begin = 0
         self._vertex_end = numpy.sum( self._build_cache_needed_points )
         
-
-    def build(self, vertex_offset, index_offset, vertices, colors, indices):
+    def build(self, vertex_offset, index_offset, vertices, colors, line_dimensions, extruders, line_types, indices):
```
nallath: Document the function (when to use, what does it do)

plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
```diff
@@ -24,6 +27,14 @@
 catalog = i18nCatalog("cura")
 
 
+def colorCodeToRGBA(color_code):
```
nallath: Document what format of input it expects and what kind of output it returns.

plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
```diff
@@ -144,7 +159,33 @@ def run(self):
                 self._progress.setProgress(progress)
 
         # We are done processing all the layers we got from the engine, now create a mesh out of the data
-        layer_mesh = layer_data.build()
+
+        # Find out colors per extruder
+        global_container_stack = Application.getInstance().getGlobalContainerStack()
+        manager = ExtruderManager.getInstance()
+        extruders = list(manager.getMachineExtruders(global_container_stack.getId()))
```
nallath: Why convert it to a list here?

plugins/LayerView/LayerView.py
```diff
@@ -60,17 +61,30 @@ def __init__(self):
...
+
         Preferences.getInstance().preferenceChanged.connect(self._onPreferencesChanged)
 
         self._solid_layers = int(Preferences.getInstance().getValue("view/top_layer_count"))
         self._only_show_top_layers = bool(Preferences.getInstance().getValue("view/only_show_top_layers"))
+        self._compatibility_mode = True  # for safety
 
         self._wireprint_warning_message = Message(catalog.i18nc("@info:status", "Cura does not accurately display layers when Wire Printing is enabled"))
 
+    def _resetSettings(self):
```
nallath: just _reset might be a better name for this. I'm also fine leaving this as is.

plugins/LayerView/LayerView.py
```diff
@@ -136,6 +154,71 @@ def setLayer(self, value):
...
+        if self._minimum_layer_num != value:
+            self._minimum_layer_num = value
+            if self._minimum_layer_num < 0:
+                self._minimum_layer_num = 0
+
+            self._startUpdateTopLayers()
+
+            self.currentLayerNumChanged.emit()
+
+    def setLayerViewType(self, layer_view_type):
```
nallath: Documentation please :) What input is expected for layer_view_type?

plugins/LayerView/LayerView.py
```diff
@@ -136,6 +154,71 @@ def setLayer(self, value):
...
+        self._layer_view_type = layer_view_type
+        self.currentLayerNumChanged.emit()
+
+    def getLayerViewType(self):
+        return self._layer_view_type
+
+    def setExtruderOpacity(self, extruder_nr, opacity):
```
nallath: Documentation + bit more defensive coding. We don't handle the case of trying to set the opacity of extruder_nr "OMGWTFBBQ" (which never exists)

plugins/LayerView/LayerView.qml
```diff
@@ -15,11 +15,36 @@ Item
 
     Slider
     {
+        id: slider2
```
nallath: I think you can come up with a better descriptive name ;)

plugins/LayerView/LayerView.qml
```diff
@@ -114,4 +145,129 @@ Item
             }
         }
     }
+
+    Rectangle {
+        anchors.left: parent.left
+        anchors.verticalCenter: parent.verticalCenter
+        anchors.top: slider_background.bottom
+        anchors.topMargin: UM.Theme.getSize("default_margin").height
+        //anchors.leftMargin: UM.Theme.getSize("default_margin").width
```
nallath: Unused code. Remove this

plugins/LayerView/LayerView.qml
```diff
@@ -114,4 +145,129 @@ Item
...
+            anchors.left: parent.left
+            text: catalog.i18nc("@label","Compatibility mode")
+            visible: UM.LayerView.compatibilityMode
+        }
+
+        ColumnLayout {
+            id: view_settings
+            anchors.top: UM.LayerView.compatibilityMode ? compatibilityModeLabel.bottom : layerTypeCombobox.bottom
+            anchors.topMargin: UM.Theme.getSize("default_margin").height
+            x: UM.Theme.getSize("default_margin").width
```
nallath: Why not use anchors.left & leftMargin?

plugins/LayerView/LayerView.qml
```diff
@@ -114,4 +145,129 @@ Item
...
+
+            CheckBox {
```
nallath: This would be better with a repeater & a listModel. This is a whole lot of code duplication.

plugins/LayerView/LayerViewProxy.py
```diff
@@ -36,6 +37,12 @@ def currentLayer(self):
         if type(active_view) == LayerView.LayerView.LayerView:
             return active_view.getCurrentLayer()
 
+    @pyqtProperty(int, notify = currentLayerChanged)
+    def minimumLayer(self):
+        active_view = self._controller.getActiveView()
+        if type(active_view) == LayerView.LayerView.LayerView:
```
nallath: Why not ductype it? If a view has a getMinimumLayer() function, it is good enough. It shouldn't matter if it is a layerview or not.

plugins/LayerView/LayerViewProxy.py
```diff
@@ -44,13 +51,84 @@ def busy(self):
             return active_view.isBusy()
 
         return False
-    
+
+    @pyqtProperty(bool)
+    def compatibilityMode(self):
+        active_view = self._controller.getActiveView()
+        if type(active_view) == LayerView.LayerView.LayerView:
```
nallath: See previous comment

plugins/LayerView/layers.shader
```diff
@@ -3,33 +3,163 @@ vertex =
...
     attribute lowp vec4 a_color;
+    attribute lowp vec4 a_material_color;
+
     varying lowp vec4 v_color;
+    varying float v_line_type;
+
     void main()
     {
         gl_Position = u_modelViewProjectionMatrix * a_vertex;
-        // shade the color depending on the extruder index stored in the alpha component of the color
```
nallath: Why remove the documentation?

plugins/LayerView/layers.shader
```diff
@@ -3,33 +3,163 @@ vertex =
...
-        v_color = (a_color.a == u_active_extruder) ? a_color : a_color * u_shade_factor;
-        v_color.a = 1.0;
+        v_color = a_color;
+        if ((a_line_type != 8.0) && (a_line_type != 9.0)) {
```
nallath: a_line_type seems to be an enum. Could you also use it as such. I have no idea what 8.0 and 9.0 indicate right now.

plugins/LayerView/layers.shader
```diff
@@ -3,33 +3,163 @@ vertex =
...
+        }
+
+        v_line_type = a_line_type;
     }
 
 fragment =
     varying lowp vec4 v_color;
+    varying float v_line_type;
+
+    uniform int u_show_travel_moves;
```
nallath: Why is this an int? As far as I can see it's used as a bool.

plugins/LayerView/layers.shader
```diff
@@ -3,33 +3,163 @@ vertex =
...
+    in lowp vec4 a_material_color;
+
+    out lowp vec4 v_color;
+    out float v_line_type;
+
+    void main()
+    {
+        gl_Position = u_modelViewProjectionMatrix * a_vertex;
+        v_color = a_color;
+        if ((a_line_type != 8) && (a_line_type != 9)) {
```
nallath: See comment about enum. Also you now compare it to the integer 8 (instead of 8.0 that you used previously)

plugins/LayerView/layers3d.shader
```diff
@@ -0,0 +1,346 @@
...
+        vec4 g_vertex_offset_horz_head;
+
+        float size_x;
+        float size_y;
+
+        if ((v_extruder_opacity[0][v_extruder[0]] == 0.0) && (v_line_type[0] != 8) && (v_line_type[0] != 9)) {
+            return;
+        }
+        // See LayerPolygon; 8 is MoveCombingType, 9 is RetractionType
+        if ((u_show_travel_moves == 0) && ((v_line_type[0] == 8) || (v_line_type[0] == 9))) {
```
nallath: See comment about enum

plugins/LayerView/layers3d.shader
```diff
@@ -0,0 +1,346 @@
...
+        if ((u_show_infill == 0) && (v_line_type[0] == 6)) {
+            return;
+        }
+
+        if ((v_line_type[0] == 8) || (v_line_type[0] == 9)) {
+            // fixed size for movements
+            size_x = 0.1;
+            size_y = 0.1;
+        } else {
+            size_x = v_line_dim[0].x / 2 + 0.01;  // radius, and make it nicely overlapping
```
nallath: Might want to use a overlap variable for this (in case this is used in other places).

plugins/LayerView/layers3d.shader
```diff
@@ -0,0 +1,346 @@
...
+        g_vertex_delta = gl_in[1].gl_Position - gl_in[0].gl_Position;
+        g_vertex_normal_horz_head = normalize(vec3(-g_vertex_delta.x, -g_vertex_delta.y, -g_vertex_delta.z));
+        g_vertex_offset_horz_head = vec4(g_vertex_normal_horz_head * size_x, 0.0);
+
+        g_vertex_normal_horz = normalize(vec3(g_vertex_delta.z, g_vertex_delta.y, -g_vertex_delta.x));
+
+        g_vertex_offset_horz = vec4(g_vertex_normal_horz * size_x, 0.0); //size * g_vertex_normal_horz;
+        g_vertex_normal_vert = vec3(0.0, 1.0, 0.0);
+        g_vertex_offset_vert = vec4(g_vertex_normal_vert * size_y, 0.0);
+
```
nallath: Wouldn't it be better to make a function that handles the creation of verts? This seems like an awfull lot of code duplication.

plugins/LayerView/layers3d.shader
```diff
@@ -0,0 +1,346 @@
...
+    void main()
+    {
+        mediump vec4 finalColor = vec4(0.0);
+        float alpha = f_color.a;
+
+        finalColor.rgb += f_color.rgb * 0.3;
+
+        highp vec3 normal = normalize(f_normal);
+        highp vec3 lightDir = normalize(u_lightPosition - f_vertex);
```
nallath: Codestyle; light_dir

plugins/LayerView/layers3d.shader
```diff
@@ -0,0 +1,346 @@
...
+
+fragment41core =
+    #version 410
+    in lowp vec4 f_color;
+    in lowp vec3 f_normal;
+    in lowp vec3 f_vertex;
+
+    out vec4 frag_color;
+
+    uniform mediump vec4 u_ambientColor;
```
nallath: Codestyle: u_ambient_color

plugins/LayerView/layerview_composite.shader
```diff
@@ -63,6 +63,74 @@ fragment =
...
+
+    const vec3 x_axis = vec3(1.0, 0.0, 0.0);
+    const vec3 y_axis = vec3(0.0, 1.0, 0.0);
+    const vec3 z_axis = vec3(0.0, 0.0, 1.0);
+
+    out vec4 frag_color;
+
+    void main()
+    {
+        kernel[0] = 0.0; kernel[1] = 1.0; kernel[2] = 0.0;
```
nallath: Add a bit of documentation here that this is a blur kernel (sorta)

plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
jackha: Easier debugging and not sure if an iterator will evaluate to False if it doesn't return any objects.

plugins/LayerView/LayerView.qml
jackha: Haha, yes and its going to be removed in the near future

plugins/LayerView/LayerView.qml
jackha: No reason and I fixed it.

plugins/LayerView/LayerView.qml
jackha: Well, the only thing the items have in common is "CheckBox". All elements have unique properties.

plugins/LayerView/LayerViewProxy.py
jackha: I copied the method how current properties work. Maybe if we add another layer view we can change it. For now it doesn't matter.

plugins/LayerView/layers.shader
jackha: It was not correct any more, but now I've added a new correct comment.

plugins/LayerView/layers.shader
jackha: Bools are available only from OpenGL 4.1.

plugins/LayerView/layers3d.shader
jackha: Uhm yes, why didn't I do that? It's so convenient ^^. Doing it now.

plugins/LayerView/layerview_composite.shader
jackha: I didn't know that ^^

plugins/LayerView/layers3d.shader
jackha: Are you sure? There are a lot of bindings and attributes that have this camelCase. I started renaming and realized that I practically had to rename all... partial work is in branch rename_shader_variables

nallath: I think all of our codestyle was like that. Could be that someone ignored it in the past.

awhiemstra: We never really had (and still don't have) a coding style for either GLSL or public properties. I would say it is fine as is, if we really want to change it we should do that separately.

plugins/LayerView/layers.shader
jackha: After a small discussion we can leave it as is for now. I've added a comment.

PR #1290, opened 2017-01-02
Add a setting to select skin pattern for first layer

14bitVoid:
This pull req creates a setting to change the skin pattern for just the first layer. I like concentric skin for bed adhesion but not for anything else so this allows to set just the layer that does the bed adhesion to concentric.
I've used this for a while but now I decided to make a PR for it, it was tested on the Ultimaker 2 printer.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -890,6 +890,20 @@
                     "default_value": "lines",
                     "settable_per_mesh": true
                 },
+                "top_bottom_pattern_0":
+                {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Ah, I thought of something just after I merged this. It would've been nice to make this setting inherit from the old Top/Bottom Pattern setting, so that changing the Top/Bottom Pattern doesn't leave the first layer unchanged for reasons that are normally hidden to the user.

I'm going to add that inheritance to the setting. I hope you don't mind.

PR #1298, opened 2017-01-04
Correct y axis print head size for printrbot simple

Zazcallabah:
Print head size settings for printrbot simple had y min and y max mixed up,
making "print one at a time" always print in the wrong order.


Ghostkeeper:
We can't test this on an actual Printrbot, but I believe you. Thanks for the fix!

PR #1317, opened 2017-01-07
Fix connections to devices not based on arduino

TheRikke:
stk500v2 failed to release the serial port on timeout. Fixes Cura unable to connect to not arduino based printers on Windows. Bug did not occur on Linux due to different serial port behavior.


nallath:
Thanks for the fix :)

PR #1324, opened 2017-01-09
Add definition file for 101Hero

TheRikke:
This is a definition file for the 101Hero. A Kickstarter/Indiegogo campain product. It is based on Marlin 1.0.0 firmware.


awhiemstra:
Looks good to me. It would be nice to have a build platform mesh for it though, any chance you can provide that?

TheRikke:
Sorry, but I can't provide a platform mesh. Not that good at modeling. Just a developer.

Issue #1232, opened 2016-12-09
Tower position when center=0,0 (bug?) Cura 2.3.1 and 2.4beta


karabas:
I set center as 0,0
1 Shadow of tower does not know about it at all  and displayed as if center at  0,0 = false
2 Cura accepts only positive values as tower position. So I can place the tower only right far corner of the bed.

Ghostkeeper:
Probably the prime locations as well, then.

karabas:
The same in 2.4 beta. Tower placement  for delta is very limited due to this bug.
![2017-01-11_18-01-49](https://cloud.githubusercontent.com/assets/67457/21853613/e9290c16-d828-11e6-911c-e2c0ce234377.png)

fieldOfView:
As it says in the 2.4 release notes, delta support is largely untested, because none of the Cura regulars have any deltas around. So we rely on reports like this to get things fixed. Thanks @karabas!

PR #1336, opened 2017-01-11
Fix prime disallowed areas when center is zero 

fieldOfView:
This PR fixes how Cura determines the disallowed areas when machine_center_is_zero is true. It fixes https://github.com/Ultimaker/Cura/issues/1232

The PR contains a minor hack to ignore the extruder prime position when it is not set in the printer definition. Without this hack, Cura plonks a fairly large disallowed area in the center of the buildplate, because the extruder_prime_position is not settable inside Cura and it defaults to (0,0).

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -538,8 +538,11 @@ class BuildVolume(SceneNode):
             prime_tower_size = self._global_container_stack.getProperty("prime_tower_size", "value")
             machine_width = self._global_container_stack.getProperty("machine_width", "value")
             machine_depth = self._global_container_stack.getProperty("machine_depth", "value")
-            prime_tower_x = self._global_container_stack.getProperty("prime_tower_position_x", "value") - machine_width / 2 #Offset by half machine_width and _depth to put the origin in the front-left.
-            prime_tower_y = - self._global_container_stack.getProperty("prime_tower_position_y", "value") + machine_depth / 2
+            prime_tower_x = self._global_container_stack.getProperty("prime_tower_position_x", "value")
+            prime_tower_y = - self._global_container_stack.getProperty("prime_tower_position_y", "value")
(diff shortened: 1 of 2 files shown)


cura/BuildVolume.py
```diff
@@ -570,8 +573,17 @@ def _computeDisallowedAreasPrime(self, border_size, used_extruders):
         machine_width = self._global_container_stack.getProperty("machine_width", "value")
         machine_depth = self._global_container_stack.getProperty("machine_depth", "value")
         for extruder in used_extruders:
-            prime_x = extruder.getProperty("extruder_prime_pos_x", "value") - machine_width / 2 #Offset by half machine_width and _depth to put the origin in the front-left.
-            prime_y = machine_depth / 2 - extruder.getProperty("extruder_prime_pos_y", "value")
+            prime_x = extruder.getProperty("extruder_prime_pos_x", "value")
+            prime_y = - extruder.getProperty("extruder_prime_pos_y", "value")
+
+            #Ignore extruder prime position if it is not set
+            if prime_x == 0 and prime_y == 0:
```
awhiemstra: I wonder if this assumption is safe, since I can imagine 0,0 being a valid prime position for a non-machine_center_is_zero machine. However, since it only affects disallowed areas it is not really important and we can fix it if we ever come across it.

awhiemstra:
One minor comment that is not blocking. For the rest it looks fine, so merged.

cura/BuildVolume.py
fieldOfView: It probably isn't, but I could not find a good way to check if the setting was actually set in the machine definition. 

Perhaps a better way to handle this would be to make the default value for extruder_prime_pos_x and _y something like -1000000 (can we do -INF in default_values?).

awhiemstra: Yeah, I figured it was that. If the setting is a float, we can use any valid float as default value, including infinity or NaN.

PR #1337, opened 2017-01-11
Hide profiles selection when hiding settings

fieldOfView:
This PR ensures the selection of profiles (variant/material/quality) is not shown when hiding the settings (eg when showing a gcode file). Showing the dropdown boxes for profile selection enables the user to change settings, and this should not be possible.


awhiemstra:
I noticed it too, was wondering if I needed to mention it or not. This also fixes it. :)

Also, congratulations on getting PR #1337 ;)

PR #1339, opened 2017-01-12
Add Cartesio config files

maukcc:


resources/definitions/cartesio.def.json
```diff
@@ -0,0 +1,52 @@
...
+    "version": 2,
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Scheepers",
+        "manufacturer": "Cartesio bv",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
+        "has_machine_materials": true,
```
awhiemstra: Note that this only works if there are actual per-machine materials, otherwise this will result in no materials displayed.

resources/variants/cartesio_0.25.inst.cfg
```diff
@@ -0,0 +1,15 @@
...
+definition = cartesio
+
+[metadata]
+author = Cartesio
+type = variant
+
+[values]
+machine_nozzle_size = 0.25
+machine_nozzle_tip_outer_diameter = 1.05
+speed_wall = =round(speed_print / 1.25, 1)
```
awhiemstra: Since you are using this formula in all 4 nozzle configurations, it should be easy to set these for the entire machine using something along the lines of:

```
"overrides": {
  "speed_wall": {
    "value": "round(speed_print / 1.25, 1)"
  }
}
```

This also applies to the other two speed settings.

If you did not intend to override the default formula for speed calculation, I would recommend just removing them.

resources/definitions/cartesio.def.json
maukcc: We are not really comfortable with the current standard materials, and we do not want customers to pick them. We are making material files with machine/nozzle specific settings that they can download

resources/variants/cartesio_0.25.inst.cfg
maukcc: I haven't really crystallized these settings yet, but wanted to be ready for the future :)

resources/definitions/cartesio.def.json
awhiemstra: In that case, you might want to simply disable "has_materials" for now. You could then update the definition once the materials are actually available. I think that would be better than having an empty material selector.

resources/variants/cartesio_0.25.inst.cfg
awhiemstra: Ah right, so you are planning on actually making them different? In that case, it is fine to leave them as is.

resources/definitions/cartesio.def.json
maukcc: Where are the material settings actually located in this repository?

awhiemstra: The material files are in https://github.com/ultimaker/fdm_materials . The original reason for that is to be able to share them between Cura and the UM3.

maukcc: OK. So We will leave this as is and change the material files over there?

maukcc: I guess this is ready to merge?
There will be updates to the "extruder start code" as soon as the program matures.

Ghostkeeper:
This part works, so I'm merging it. The changes you made to FDM Materials don't quite work yet. I've made a pull request on Mauk's fork to fix those. So for the moment, you won't see any new materials or changed material settings for the Cartesio printer. Once the FDM Materials are merged, those should appear as well.

PR #1340, opened 2017-01-12
PowerManagement: Initial commit

thopiekar:
The provider for Linux shall work, but further tests are needed for Windows and mostly for the OSX provider.

### DIFF
--- a/plugins/PowerManagement/PowerManagementLinux.py
+++ b/plugins/PowerManagement/PowerManagementLinux.py
@@ -1,0 +1,70 @@
+'''
+Created on 10.01.2017
+
+@author: thopiekar
+'''
+
+import dbus
+
+from UM.Extension import Extension
+
+from UM.Logger import Logger
+
+class PowerManagement(Extension):
+    def __init__(self):
+        super().__init__()
+                
+        self.suspendInhibition = None
+        
+        __init_api_list__ = [self.__init_api_solid__,
+                             self.__init_api_freedesktop__,
+                             ]
+        
+        while __init_api_list__:
+            try:
+                __init_api_list__[0]()
+                return
+            except:
+                del __init_api_list__[0]
+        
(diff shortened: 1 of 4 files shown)


plugins/PowerManagement/PowerManagementLinux.py
```diff
@@ -0,0 +1,70 @@
+'''
+Created on 10.01.2017
+
+@author: thopiekar
+'''
+
+import dbus
```
awhiemstra: I would personally prefer to use Qt's DBus module instead, so we only need one DBus implementation.

thopiekar: Is that really necessary?

plugins/PowerManagement/PowerManagementLinux.py
```diff
@@ -0,0 +1,70 @@
...
+                return
+            except:
+                del __init_api_list__[0]
+        
+        Logger.log("e", "We were not able to provide any power management provider!")
+        return
+    
+    def __init_api_solid__(self):
+            self.solid_session = None
+            devobj = dbus.SessionBus().get_object("org.kde.kded5",
```
thopiekar: But the freedesktop interface doesn't allow to pass the application icon, afaik.
I couldn't find a way to pass the icon at the moment, but I'm planning to find it out later.

thopiekar:
Well, I might be useful when sending very fat prints to a UM3 and the user just closes the lid.

In the future I'm planning to bring up a new plugin, but I'm not allowed to share any details in public at the moment. I can come over later and explain.

plugins/PowerManagement/PowerManagementLinux.py
Ghostkeeper: It removes a dependency with relative ease. Would be very nice. Our build system is already quite complicated, as you know.

thopiekar:
Created an own repo for it.

Issue #1316, opened 2017-01-07
Convert legacy .ini machine profiles to .def.json machine definitions


probonopd:
How does one convert legacy `.ini` machine profiles to `.def.json` machine definitions?
I would like to convert [this `.ini`](https://gist.githubusercontent.com/probonopd/0dc2afa55fc31caa5ce1f5eea0f3255b/raw/11f440410d935760df20374bcbcbdc47515f03ca/RF100.ini) which came with the machine to the  `.def.json` format that newer versoins of Cura require that I could put in `~/.local/share/cura/definitions`, __without losing any information__.

I found [LegacyCuraMachineConverter](https://github.com/BagelOrb/LegacyCuraMachineConverter) but it seems unmaintained.

Especially:
* [This `.ini`](https://gist.githubusercontent.com/probonopd/0dc2afa55fc31caa5ce1f5eea0f3255b/raw/11f440410d935760df20374bcbcbdc47515f03ca/RF100.ini) seems to be "machine" related, yet newer Cura versions seem to suggest that there are "Printer" settings, and separate from that, "Profiles". The `.ini` that came with the machine seems to be a mixture of both. How do I split this into "Printer" and "Profile"?
* Are there instructions on how to do this, or a converter to do it automatically? (I see that there is `Cura/plugins/LegacyProfileReader/` but no documentation that I could find)
* Are the keys named the same in the `.ini` and `.def.json` formats? (I see that there is `Cura/plugins/LegacyProfileReader/DictionaryOfDoom.json` but no documentation that I could find)
* Are all keys from `.ini` supported in `.def.json` and vice versa?

Using http://quillford.github.io/CuraProfileMaker/ only gives a very rudimentary profile, compared to profiles like [this one](https://github.com/Ultimaker/Cura/blob/master/resources/definitions/fdmprinter.def.json).

nallath:
There is indeed a difference between a "profile", which describes things like "print this on normal / high quality" or material "pla"/ "abs". The profiles can also describe something like "Abs high quality needs to be printed like this". To even further complicate things, there can be a "Generic ABS high quality" and a "Ultimaker 3 high quality" profile.

There are also machine definitions; These define standards of a machine. Think about physical dimensions, but also settings that due to it's setup should be considered default (eg; Bowden style printers tend to support higher acceleration)

The keys are different. Have a look at the "dictionary of doom". It contains a mapping of all settings from legacy to 2.x. 

[...]

PR #1350, opened 2017-01-15
Add Renkforce RF100

probonopd:
Add [this printer](https://www.conrad.de/de/renkforce-rf100-3d-drucker-starter-kit-inkl-filament-single-extruder-1507428.html). This file is based on the `.ini` file that came with it, as per https://github.com/Ultimaker/Cura/issues/1316

### DIFF
--- a/resources/definitions/renkforce_rf100.def.json
+++ b/resources/definitions/renkforce_rf100.def.json
@@ -1,0 +1,87 @@
+{
+    "id": "RF100",
+    "version": 2,
+    "name": "Renkforce RF100",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "author": "Simon Peter (based on RF100.ini by Conrad Electronic SE)",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "manufacturer": "Renkforce",
+        "platform_offset": [0, 0, 0],
+        "visible": true
+    },
+
+    "overrides": {
+        "adhesion_type": {
+            "default_value": "skirt"
+        },
+        "bottom_thickness": {
+            "default_value": 0
+        },
+        "infill_sparse_density": {
+            "default_value": 15
+        },
+        "machine_depth": {
+            "default_value": 100
+        },
+        "machine_end_gcode": {
+            "default_value": ";End GCode\nM104 S0 ;extruder heater off\nM140 S0 ;heated bed heater off (if you have it)\nG91 ;relative positioning\nG1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure\nG1 Z+0.5 E-5 X-20 Y-20 F{speed_travel} ;move Z up a bit and retract filament even more\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\nM84 ;steppers off\nG90 ;absolute positioning"
+        },
+        "machine_gcode_flavor": {
+            "default_value": "RepRap (Marlin/Sprinter)"
+        },
+        "machine_height": {
+            "default_value": 100
+        },
+        "machine_name": {
+            "default_value": "Renkforce RF100"
+        },
+        "machine_start_gcode": {
+            "default_value": ";Sliced at: {day} {date} {time}\nG21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 ;move X/Y to min endstops\nG28 Z0 ;move Z to min endstops\nG1 Z15.0 F{speed_travel} ;move the platform down 15mm\nG92 E0 ;zero the extruded length\nG1 F200 E3 ;extrude 3mm of feed stock\nG92 E0 ;zero the extruded length again\nG1 F{speed_travel}\nM117 Printing..."
+        },
+        "machine_width": {
+            "default_value": 100
+        },
+        "material_diameter": {
+            "default_value": 1.75
+        },
+        "retraction_amount": {
+            "default_value": 2
+        },
+        "speed_layer_0": {
+            "default_value": 30
+        },
+        "speed_topbottom": {
+            "default_value": 30
+        },
+        "speed_travel": {
+            "default_value": 50
+        },
+        "speed_wall_0": {
+            "default_value": 25
+        },
+        "speed_wall_x": {
+            "default_value": 35
+        },
+        "support_xy_distance": {
+            "default_value": 0
+        },
+        "support_z_distance": {
+            "default_value": 0.1
+        },
+        "top_thickness": {
+            "default_value": 0.5
+        }
+    },
+
+    "categories": {
+        "material": {
+            "settings": {
+                "material_bed_temperature": {
+                    "visible": false
+                }
+            }
+        }
+    }
+}
--- a/resources/definitions/renkforce_rf100.def.json
+++ b/resources/definitions/renkforce_rf100.def.json
@@ -1,14 +1,13 @@
 {
     "id": "RF100",
     "version": 2,
-    "name": "renkforce RF100",
+    "name": "Renkforce RF100",
     "inherits": "fdmprinter",
     "metadata": {
         "author": "Simon Peter (based on RF100.ini by Conrad Electronic SE)",
         "category": "Other",
         "file_formats": "text/x-gcode",
-        "manufacturer": "renkforce",
-        "platform_offset": [0, 0, 0],
+        "manufacturer": "Renkforce",
         "visible": true
     },
 
@@ -17,7 +16,7 @@
             "default_value": "skirt"
         },
         "bottom_thickness": {
-            "default_value": 0.5
+            "value": 0.5
         },
         "brim_line_count": {
             "default_value": 20.0
@@ -44,25 +43,16 @@
             "default_value": 10.0
         },
         "infill_before_walls": {
-            "default_value": 1.0
+            "default_value": true
         },
         "infill_overlap": {
-            "default_value": 15.0
-        },
-        "infill_sparse_density": {
             "default_value": 15.0
         },
         "layer_0_z_overlap": {
             "default_value": 0.22
         },
-        "layer_height": {
-            "default_value": 0.1
-        },
         "layer_height_0": {
             "default_value": 0.3
-        },
-        "line_width": {
-            "default_value": 0.4
         },
         "machine_depth": {
             "default_value": 100
@@ -77,10 +67,7 @@
             "default_value": 100
         },
         "machine_name": {
-            "default_value": "renkforce RF100"
-        },
-        "machine_nozzle_size": {
-            "default_value": 0.4
+            "default_value": "Renkforce RF100"
         },
         "machine_start_gcode": {
             "default_value": ";Sliced at: {day} {date} {time}\nG21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 ;move X/Y to min endstops\nG28 Z0 ;move Z to min endstops\nG1 Z15.0 F{speed_travel} ;move the platform down 15mm\nG92 E0 ;zero the extruded length\nG1 F200 E3 ;extrude 3mm of feed stock\nG92 E0 ;zero the extruded length again\nG1 F{speed_travel}\nM117 Printing..."
@@ -89,13 +76,10 @@
             "default_value": 100
         },
         "material_bed_temperature": {
-            "default_value": 70.0
+            "visible": false
         },
         "material_diameter": {
             "default_value": 1.75
-        },
-        "material_flow": {
-            "default_value": 100.0
         },
         "material_print_temperature": {
             "default_value": 210.0
@@ -216,16 +200,6 @@
         },
         "wall_thickness": {
             "default_value": 0.8
-        }
-    },
-
-    "categories": {
-        "material": {
-            "settings": {
-                "material_bed_temperature": {
-                    "visible": false
-                }
-            }
         }
     }
 }
--- a/resources/definitions/renkforce_rf100.def.json
+++ b/resources/definitions/renkforce_rf100.def.json
@@ -115,12 +115,6 @@
         "ooze_shield_enabled": {
             "default_value": true
         },
-        "prime_tower_enable": {
-            "default_value": true
-        },
-        "prime_tower_size": {
-            "default_value": 12.24744871391589
-        },
         "raft_airgap": {
             "default_value": 0.22
         },
--- a/resources/definitions/renkforce_rf100.def.json
+++ b/resources/definitions/renkforce_rf100.def.json
@@ -58,7 +58,7 @@
             "value": "100"
         },
         "machine_end_gcode": {
-            "value": ";End GCode\nM104 S0 ;extruder heater off\nM140 S0 ;heated bed heater off (if you have it)\nG91 ;relative positioning\nG1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure\nG1 Z+0.5 E-5 X-20 Y-20 F{speed_travel} ;move Z up a bit and retract filament even more\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\nM84 ;steppers off\nG90 ;absolute positioning"
+            "value": "';End GCode\nM104 S0 ;extruder heater off\nM140 S0 ;heated bed heater off (if you have it)\nG91 ;relative positioning\nG1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure\nG1 Z+0.5 E-5 X-20 Y-20 F{speed_travel} ;move Z up a bit and retract filament even more\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\nM84 ;steppers off\nG90 ;absolute positioning'"
         },
         "machine_gcode_flavor": {
             "value": "RepRap (Marlin/Sprinter)"
@@ -70,7 +70,7 @@
             "value": "Renkforce RF100"
         },
         "machine_start_gcode": {
-            "value": ";Sliced at: {day} {date} {time}\nG21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 ;move X/Y to min endstops\nG28 Z0 ;move Z to min endstops\nG1 Z15.0 F{speed_travel} ;move the platform down 15mm\nG92 E0 ;zero the extruded length\nG1 F200 E3 ;extrude 3mm of feed stock\nG92 E0 ;zero the extruded length again\nG1 F{speed_travel}\nM117 Printing..."
+            "value": "';Sliced at: {day} {date} {time}\nG21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 ;move X/Y to min endstops\nG28 Z0 ;move Z to min endstops\nG1 Z15.0 F{speed_travel} ;move the platform down 15mm\nG92 E0 ;zero the extruded length\nG1 F200 E3 ;extrude 3mm of feed stock\nG92 E0 ;zero the extruded length again\nG1 F{speed_travel}\nM117 Printing...'"
         },
         "machine_width": {
             "value": "100"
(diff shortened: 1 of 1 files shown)


probonopd:
Don't merge yet: Using this profile results in improper slicing. Only the outer walls are there, the entire inside of the model is ignored. Why?

probonopd:
Still getting
> Unable to slice with the current settings. The following settings have errors: Prime Tower X Position, Prime Tower Y Position.

probonopd:
It slices http://www.thingiverse.com/thing:692523 entirely wrong - why?

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
+{
+    "id": "RF100",
+    "version": 2,
+    "name": "renkforce RF100",
```
Ghostkeeper: We'd better capitalise the R. I base that on this: https://www.conrad.de/de/renkforce-rf100-3d-drucker-starter-kit-inkl-filament-single-extruder-1507428.html

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+    "inherits": "fdmprinter",
+    "metadata": {
+        "author": "Simon Peter (based on RF100.ini by Conrad Electronic SE)",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "manufacturer": "renkforce",
```
Ghostkeeper: Similarly, this R should also be capitalised.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+        "platform_offset": [0, 0, 0],
```
Ghostkeeper: There is no platform mesh for this printer. This line doesn't do anything at the moment. Perhaps you'd like to add a platform mesh? Or otherwise we should remove this.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+        "visible": true
+    },
+
+    "overrides": {
+        "adhesion_type": {
+            "default_value": "skirt"
+        },
+        "bottom_thickness": {
+            "default_value": 0.5
```
Ghostkeeper: Like I said [here](https://github.com/Ultimaker/Cura/issues/1316#issuecomment-274355866), it's wise to overwrite `"value"` instead of `"default_value"`. This setting change won't have any effect otherwise.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+        },
+        "brim_line_count": {
+            "default_value": 20.0
```
Ghostkeeper: Not a blocking problem, but it's better to overwrite `brim_width` rather than `brim_line_count`, otherwise users can't modify `brim_width` any more (because `brim_line_count` overrides `brim_width`).

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": 0.22
+        },
+        "layer_height": {
+            "default_value": 0.1
+        },
+        "layer_height_0": {
+            "default_value": 0.3
+        },
+        "line_width": {
+            "default_value": 0.4
```
Ghostkeeper: This is already the default in Cura 2, so it can be left out.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": 10.0
+        },
+        "infill_before_walls": {
+            "default_value": 1.0
+        },
+        "infill_overlap": {
+            "default_value": 15.0
+        },
+        "infill_sparse_density": {
+            "default_value": 15.0
```
Ghostkeeper: Are you sure that the infill density of this printer should be different from the default infill density of other printers?

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": true
+        },
+        "cool_min_layer_time": {
+            "default_value": 5.0
+        },
+        "cool_min_speed": {
+            "default_value": 10.0
+        },
+        "infill_before_walls": {
+            "default_value": 1.0
```
Ghostkeeper: This should be a boolean setting, I believe. I don't know how your script ended up with 1.0.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": 15.0
+        },
+        "infill_sparse_density": {
+            "default_value": 15.0
+        },
+        "layer_0_z_overlap": {
+            "default_value": 0.22
+        },
+        "layer_height": {
+            "default_value": 0.1
```
Ghostkeeper: This is already the default in Cura 2.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": "RepRap (Marlin/Sprinter)"
+        },
+        "machine_height": {
+            "default_value": 100
+        },
+        "machine_name": {
+            "default_value": "renkforce RF100"
+        },
+        "machine_nozzle_size": {
+            "default_value": 0.4
```
Ghostkeeper: This is already the default in Cura 2.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": 100
+        },
+        "material_bed_temperature": {
+            "default_value": 70.0
+        },
+        "material_diameter": {
+            "default_value": 1.75
+        },
+        "material_flow": {
+            "default_value": 100.0
```
Ghostkeeper: This is already the default in Cura 2. There's a lot of these below, too.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,231 @@
...
+        "wall_thickness": {
+            "default_value": 0.8
+        }
+    },
+
+    "categories": {
+        "material": {
+            "settings": {
+                "material_bed_temperature": {
+                    "visible": false
```
Ghostkeeper: You should just put this `"material_bed_temperature"` in the overrides along with the rest. It doesn't matter that this one edits the `visible` property rather than `default_value`. Perhaps you would put a white space between them.

Ghostkeeper:
Also, thank you for creating the AppImage system 😄 We are making thankful use of it.

probonopd:
Thanks for your thorough review @Ghostkeeper, great learning for me, i'll implement your changes now.

resources/definitions/renkforce_rf100.def.json
probonopd: This was an artifact of the automatic conversion. Removing it for now.

resources/definitions/renkforce_rf100.def.json
probonopd: As commented [here](https://github.com/Ultimaker/Cura/issues/1316#issuecomment-274361823) I am not sure about this - should I replace __all__ occurences of `default_value` with `value` or just the ones you specifically commented on here?

resources/definitions/renkforce_rf100.def.json
probonopd: The original ini file had `brim_line_count = 20` and I would like to convert this as "truthfully" as possible for starters, to be improved from there. I am not sure how to do the calculation from `brim_line_count` to `brim_width`.

resources/definitions/renkforce_rf100.def.json
probonopd: The original ini file has `perimeter_before_infill = False`. So I assume it is correct to set this to `true` here. (How would I find out what is better for this printer, are there test prints that let you deterministically experiment with settings like these? A more general question well beyond just this setting...)

resources/definitions/renkforce_rf100.def.json
probonopd: No. This is what came out of the automatic conversion. Looking at the ini file, I don't see why this should be here, so deleting for now.

resources/definitions/renkforce_rf100.def.json
probonopd: Deleting.

resources/definitions/renkforce_rf100.def.json
probonopd: Deleting.

resources/definitions/renkforce_rf100.def.json
probonopd: Deleting.

resources/definitions/renkforce_rf100.def.json
probonopd: Deleting.

resources/definitions/renkforce_rf100.def.json
probonopd: Changed. The value 70 for `material_bed_temperature` doesn't make much sense for a printer without a heated bed either.

resources/definitions/renkforce_rf100.def.json
Ghostkeeper: I replied to your question here: https://github.com/Ultimaker/Cura/issues/1316#issuecomment-274371670

resources/definitions/renkforce_rf100.def.json
Ghostkeeper: It's simply done by dividing by the brim line width. But a brim line count is fine too.

resources/definitions/renkforce_rf100.def.json
Ghostkeeper: There are no test prints that deterministically experiment with such settings. There are so many settings, and they have so wildly differing functionality that it is virtually impossible to design a global test for all or even many of them.

Simply `true` should do well, but that is also the default so you can leave this setting out. If we ever decide to change the default then this printer would change along, which is a good thing.

Ghostkeeper:
Ah, wait. There was still the issue of `default_value` vs. `value`. I shouldn't have approved that review. Don't merge this yet.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,205 @@
...
+            "value": "0.22"
+        },
+        "layer_height_0": {
+            "value": "0.3"
+        },
+        "machine_depth": {
+            "value": "100"
+        },
+        "machine_end_gcode": {
+            "value": ";End GCode\nM104 S0 ;extruder heater off\nM140 S0 ;heated bed heater off (if you have it)\nG91 ;relative positioning\nG1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure\nG1 Z+0.5 E-5 X-20 Y-20 F{speed_travel} ;move Z up a bit and retract filament even more\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\nM84 ;steppers off\nG90 ;absolute positioning"
```
Ghostkeeper: These are directly interpreted as Python code, so it breaks at the first word `End` which is not a known variable. This can be fixed either by adding additional quotes around the whole thing (like `"value": "';End GCode'"`) or by making this one a `default_value` again instead of `value`, since for these particular settings the `value` is never defined yet.

The same goes for the GCode Flavour, Machine Name and Start GCode below.

fieldOfView: Isn't this by itself enough reason to prefer the use of "default_value" over "value" in definition files? That's how most of the Ultimaker profiles work too.

Ghostkeeper: Yeah, true. I guess it works either way. For settings that start with `machine_` it's generally safe to use "default_value", since there is no "value" that overrides it, but I thought this was a bit complex to communicate to our new contributor.

probonopd: It seems a bit "esoteric" to me that there are differences in the semantics based on which key is used - are you sure this is how you want it to be?

Ghostkeeper: There are no differences in semantics based on key. It's just how the data plays out.

There are two options: Either use `"value": "';End GCode'"` (with the extra quotes) or use `"default_value": ";End GCode"` (without the extra quotes) and in the latter case, make sure there is no `"value"` in a parent definition that overrides the `default_value`. Either are fine with me.

In my opinion, `default_value` could be replaced entirely with `value` anyway. But that is a major change that will not happen for some time, and perhaps not at all because I seem to be in the minority with that opinion.

probonopd: Changed.

resources/definitions/renkforce_rf100.def.json
```diff
@@ -0,0 +1,205 @@
...
+            "value": "0.22"
+        },
+        "layer_height_0": {
+            "value": "0.3"
+        },
+        "machine_depth": {
+            "value": "100"
+        },
+        "machine_end_gcode": {
+            "value": "';End GCode\nM104 S0 ;extruder heater off\nM140 S0 ;heated bed heater off (if you have it)\nG91 ;relative positioning\nG1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure\nG1 Z+0.5 E-5 X-20 Y-20 F{speed_travel} ;move Z up a bit and retract filament even more\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\nM84 ;steppers off\nG90 ;absolute positioning'"
```
Ghostkeeper: This line still fails to parse, since `\n` already gets unescaped when the string is read from the JSON, and then it evaluates a multi-line string which has a syntax error:

    2017-02-07 11:57:43,968 - ERROR - UM.Settings.SettingFunction.__init__ [36]: Parse error in function (';Sliced at: {day} {date} {time}
    G21 ;metric values
    G90 ;absolute positioning
    M82 ;set extruder to absolute mode
    M107 ;start with the fan off
    G28 X0 Y0 ;move X/Y to min endstops
    G28 Z0 ;move Z to min endstops
    G1 Z15.0 F{speed_travel} ;move the platform down 15mm
    G92 E0 ;zero the extruded length
    G1 F200 E3 ;extrude 3mm of feed stock
    G92 E0 ;zero the extruded length again
    G1 F{speed_travel}
    M117 Printing...') for setting: EOL while scanning string literal (eval, line 1)

It fails because it tries this:

    eval("';End GCode
    M104 S0 ;extruder heater off
    <...>
    G90 ;absolute positioning'")

That is a syntax error. The same happens with the `machine_start_gcode` setting.

probonopd: Sorry for me being a bit slow on this, but it is not clear to me what I am supposed to do here.

Ghostkeeper: No worries, I think we all agree that it's very confusing. There are two ways to fix this, and either one is fine:

* Escape the backslashes in the string (double escaping for the newlines). I.e. replace `\n` with `\\n`. This directly fixes the syntax error at hand.
* Still go for the other route and make these text settings use `default_value` (as they happen to not get overwritten by any `value` property by chance). To do this, change `"value"` into `"default_value"` and remove the extra quotes again.

probonopd:
Hope I got it right this time :-)

Ghostkeeper:
Apparently not.

The same holds for all other strings in the JSON. If it's value, it gets interpreted as Python and this triggers "undefined variable" errors for your strings.
This time, I've identified them all: `adhesion_type`, `machine_gcode_flavor`, `machine_name`, `retraction_combing`, `support_pattern` and `support_type` still give syntax errors (apparently they give only one error per definition at a time). Changing these to `"default_value"` makes the printer load properly and still uses the correct settings. Since I've already made these changes, I'll make a pull request on your pull request.

Ghostkeeper:
Oh, well apparently pushing is allowed. Well, this works too.

probonopd:
These settings in `.local/share/cura/user/Renkforce+RF100_current_settings.inst.cfg` significantly increase first layer adhesion for me - should we make them the default?

- material_flow = 110
- skirt_line_count = 3
- speed_layer_0 = 25.0

Ghostkeeper:
Sounds good. I'll make these changes in the Renkforce definition file.

Ghostkeeper:
Material flow is a bit weird to change in a printer definition file. It increases flow by 10% in your case, for the entire print. It'll create a bit of overextrusion on purpose, but if you're having underextrusion problems that might be a problem with your printer hardware.

I'm making the change for now. If you find a better way to solve the problem on your end, we can make further changes.

probonopd:
Sorry to bother you again, but possibly this is helpful for others who want to write profiles, too.

- I think that `F{travel_speed}` is wrong, since the printer moves way too slow in the start and end g-code commands that use it. Can you confirm?
- Also, is there a way not to cool down the head immediately at the end of the print, for multiple prints in quick succession?
- For easier access when removing the printed object, I would like to move the platform down at the end of the print. Is it a good idea to do move down in relative position, will Cura consider the maximum print area when doing this? Or do I need to have an absolute position (essentially moving the platform down all the way because I do not know how high the object is?)

What do you think about the following start and end g-code?

```
; Start GCpde
G21 ;metric values
G90 ;absolute positioning
M82 ;set extruder to absolute mode
M107 ;start with the fan off
G28 X0 Y0 ;move X/Y to min endstops
G28 Z0 ;move Z to min endstops
G1 Z15.0 ;move the platform down 15mm
G92 E0 ;zero the extruded length
G1 F200 E3 ;extrude 3mm of feed stock
G92 E0 ;zero the extruded length again
;Put printing message on LCD screen
M117 Printing...

;End GCode
M117 Finishing...
G91 ;relative positioning
G1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure
G1 Z+0.5 E-5 X-20 Y-20 F6000 ;move Z up a bit and retract filament even more
G28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way
G1 Z30.0 F6000 ;move the platform down 30mm
M104 S0 ;extruder heater off
M140 S0 ;heated bed heater off (if you have it)
M84 ;steppers off
G90 ;absolute positioning
M117 Done
```

probonopd:
I also noticed that in fast mode, infill speed is twice as high as print or travel speed, and this does not seem to work well, at least with my filament. This must be a bug?

Ghostkeeper:
> I think that F{travel_speed} is wrong, since the printer moves way too slow in the start and end g-code commands that use it. Can you confirm?

Ah, yes. It should get multiplied by 60 because that `F` is interpreted as mm/min rather than mm/s.

> Also, is there a way not to cool down the head immediately at the end of the print, for multiple prints in quick succession?

The engine always writes a temperature at the end. There is no way to prevent that currently. Only a post-processing script could take care of that.

> Is it a good idea to do move down in relative position, will Cura consider the maximum print area when doing this?

No. If you put that in the end g-code it gets copied over into the output blindly. Cura won't analyse that g-code to find whether there are any moves that would exceed the build volume. Since you know that the nozzle is always at the highest layer at the end of a print, you could do a relative move of 1cm and reduce the build volume height by 1cm or so. You could also do an absolute move to `G0 Z{machine_height}` to bring the build plate all the way down. I find that this long Z move helps me to hear when the print is finished anyway :smiley: We've also had some success with devising a move for the print head that actually pushes the object automatically off the build plate and then continues with printing the next object, but that is really specific to the model you're printing.

> What do you think about the following start and end g-code?

There's one `G1` command in the end g-code that could send your nozzle crashing into the print:

    G1 Z+0.5 E-5 X-20 Y-20 F6000 ;move Z up a bit and retract filament even more

This moves the bed 0.5mm up. If the nozzle is directly on the print there it will crash into your model. If your nozzle isn't directly on the print there but your print is less than 0.5mm high it will crash into the build plate. For the rest the g-code looks fine.

> I also noticed that in fast mode, infill speed is twice as high as print or travel speed, and this does not seem to work well, at least with my filament. This must be a bug?

Hmm, the Renkforce RF100 sets the infill speed specifically to 50.0. A lot of profiles set the infill speed to be equal to the print speed. There's no profile in all of Cura that sets it to be equal to twice the print speed. Only the Mankati Fullscale XT+ sets it to 100. I don't know where that value comes from.

probonopd:
Something is wrong with the infill speed when using __Low Quality__ or __High Quality__ -- what do I need to fix in the machine definition?

screenshot from 2017-04-30 20-18-24
screenshot from 2017-04-30 20-18-51
screenshot from 2017-04-30 20-19-26

Ghostkeeper:
The generic Low Quality profile and High Quality profile override the default with a different infill speed. There's more of these settings, too. See https://github.com/Ultimaker/Cura/blob/master/resources/quality/low.inst.cfg and https://github.com/Ultimaker/Cura/blob/master/resources/quality/high.inst.cfg

The way to prevent that would be to write your own quality profiles such that it doesn't revert to the generic ones.

PR #1355, opened 2017-01-18
Fix typos in kossel_pro gcode variable usage

ex-nerd:
Replace double-braces with single-braces so that `material_bed_temperature` and `material_print_temperature` are interpolated properly.


awhiemstra:
Seems like a simple enough fix. Merged.

PR #1359, opened 2017-01-19
Added min_infill_area setting.

smartavionics:
This allows the user to specify in mm^2 the minimum area of infill regions.
Areas smaller than this will be merged into the surrounding skin rather
than being filled with infill.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1230,6 +1230,16 @@
                     "default_value": true,
                     "enabled": "infill_sparse_density > 0",
                     "settable_per_mesh": true
+                },
+                "min_infill_area":
+                {
+                    "label": "Min Infill Area",
+                    "description": "Don't generate areas of infill smaller than this (use skin instead).",
+                    "unit": "mm²",
+                    "type": "float",
+                    "minimum_value": "0",
+                    "default_value": 0,
+                    "settable_per_mesh": true
                 }
             }
         },
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1233,7 +1233,7 @@
                 },
                 "min_infill_area":
                 {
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1230,6 +1230,16 @@
                     "default_value": true,
                     "enabled": "infill_sparse_density > 0",
                     "settable_per_mesh": true
+                },
+                "min_infill_area":
+                {
+                    "label": "Min Infill Area",
```
Ghostkeeper: We'll want this label to say "Minimum Infill Area" in full, rather than abbreviating it.

smartavionics:
Done.

Ghostkeeper:
So quick! I'm in the process of testing. The simple cases seem to work at least. I'm testing with the Voronoi Yoda at the moment.

smartavionics:
I am trying the Yoda now. It seems to me that for complex infill regions (which this model has) this setting is not useful because when the infill areas are subdivided you end up with little sub-areas which become filled with skin rather than being filled with infill.
My rational behind this feature is to improve surface skin quality by removing small segments of infill that appear underneath small raised features on the surface (e.g. text or lines). I wasn't really thinking of the affect it would have on complex infill shapes within the model.

Ghostkeeper:
I was a bit confused, because it doesn't seem to work when the top/bottom thickness is 0. After discussing this, we decided that it's probably good that the feature doesn't work if skin is disabled, although it is a bit unintuitive. Requires some thought.

In any case, the feature does seem to work as intended for the rest. So I'm going to merge this.

fieldOfView:
@Smartavionics: you're trying the what now? ;)

smartavionics:
The Voronoi Yoda that you mentioned above.

fieldOfView:
Nevermind, I was thinking of something else alltogether

PR #1368, opened 2017-01-23
Corrected confusion between 3MF and X3D

Patola:
It seems the msgid's regarding X3D were wrongly translated as for 3MF in spanish. Corrected three messages.

### DIFF
--- a/resources/i18n/es/cura.po
+++ b/resources/i18n/es/cura.po
@@ -51,19 +51,19 @@ msgstr "Rayos X"
 #, fuzzy
 msgctxt "@label"
 msgid "X3D Reader"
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Sorry. It's nice of you to fix that, but this file is currently off to the translation bureau we're hiring for the Spanish translation (as well as Dutch, Italian, German, Finnish, French and Turkish). Once they're done I'm going to use their new .po file. That would constitute a "merge conflict" as it were, though it's not a merge done by Git but manually by me.

When I get their new translation back, I'll take special care to make sure these are correct. We've had some trouble with texts being marked as "fuzzy" by msgmerge but the translators not taking care to review them.

Issue #1371, opened 2017-01-23
Colors not defined + Colors with inconvenient connection in theme.js?


rudowinger:
Hey,

While working on a "dark mode" adobeésque theme for Cura, I stumbled on a few hiccups that I want to report:

For a large part, the theming works very good and Cura can be enhanced very well with using the theme subsystem!!!

Here you can see a screenshot of my dark theme in progress:
![cols2](https://cloud.githubusercontent.com/assets/14955635/22224611/85cffab2-e1bf-11e6-845d-c03ae0fc2294.jpg)

... but as you can see, some colors have inconvenient connections or cannot be customized at all.

I made a complete-red theme to check for un-themeable parts:
![cols](https://cloud.githubusercontent.com/assets/14955635/22224650/a947f170-e1bf-11e6-91c0-15773ce579fd.jpg)

The issues I found: (maybe some things come from a plugin and use system wide values such as primary color? Help in that case is very much appreciated!)

Unthemeable parts:
- Screen "Print Setup - Recommended": Background field on "infill", Background field on "Enable Support, Build Plate adhesion", Background of "Need help improving..."
- Background of Slicer Status Window
- Tooltip background of "View - Layers" slider

Would be great to have independent theme options:
- "3D view - Layers" tooltip background and caption text
- "Slicer Status" text caption field, i.e. "Slicing...", Button "Save To file"
- "3D view" - Title, Metrics, Time & Filament Stats

Hope, my feedback helps, thank you for reading :)
If I missed an option that /is/ in fact an independent setting, I'm thankful if you point it out

fieldOfView:
This PR removes the white blocks:
https://github.com/Ultimaker/Cura/issues/1372

fieldOfView:
Forgot to say: that dark theme looks good!

rudowinger:
Thanks (also for your quick Pull requests!) You can have it, when it's done

fieldOfView:
@rudowinger have you been able to check the 2.5 beta? It contains the fixes to the theme system you need to fix your dark theme.

rudowinger:
@fieldOfView So it is in Beta? Was too dumb to compile - thanks for letting me know - will look into it!

fieldOfView:
Yes, the theme fixes are in the 2.5 beta.
https://ultimaker.com/en/blog/39116-introducing-cura-25-beta

[...]

PR #1372, opened 2017-01-23
Fix Rectangles that always show as white regardless of the theme

fieldOfView:
This PR removes a number of unthemable white blocks from the sidebar.
See the white areas here: 
borrowed screenshot

Fixes https://github.com/Ultimaker/Cura/issues/1371


awhiemstra:
Note that any rectangle with colour "transparent" could also be replaced with a simple Item instead.

fieldOfView:
I agree. I'll amend my PR

fieldOfView:
I have replaced transparent rectangles with items throughout Cura.

Ghostkeeper:
I went into theme.json and replaced all colours (using the regex `\[\d{1,3},\s*\d{1,3},\s*\d{1,3},\s*\d{1,3}\]`) with pure red `[255, 0, 0, 255]`. The result is now this:

pureredtheme

As it looks now, it seems that one colour is not yet themeable: The colour of the icons for print preview on the top right (white on the screenshot). Clicking things in this interface is very hard; it requires some guesswork. But I wasn't able to make anything else appear other than pure red. The other things you see in the screenshot are the build plate, which is intended because it has a texture, and the colours of the materials (the little yellow blocks), which is also intended.

Merging this right now is fine with me as it is a strict improvement. But would you like to make the last little detail themeable as well?

fieldOfView:
"mooi rood is niet lelijk" ;-)

The two icons in the sidebar are on my list of things to fix. The print monitor icon has colored overlays. These icons are the only SVG files among the icons where the color is taken from the SVG file. The icon should be split into two icons; the printer icon, and a status overlay. The status overlay can then be colored using the theme instead of the SVG file.

fieldOfView:
The two header tab buttons are now also themeable.

Issue #1369, opened 2017-01-23
wrong %userprofile% folder with roaming profile (W7)


ljoets:
Hello,
I have to deploy latest stable Cura ( version 2.3.1) on W7x64 computers in a school.
Users have itinerant windows profiles on a sambaedu3 server.
In theory, %userprofile% is on a network drive. And I expected to find .cura folder on it, but .cura folder stays on C:\Users
Is there a method to specify %userprofile% at launch ?
Thanks.
I had this profil problem with precedent version, and found a workaround #(https://github.com/Ultimaker/Cura/issues/505) but cfg files goes to C:\users and I can't find profile.py in C:\Programs\Cura 2.3

ljoets:
I found here where the windows userprofile is defined : 
https://github.com/Ultimaker/Cura/blob/d391639b48e3e3adc3af6c935e7595490ed4025a/cura_app.py
But I can't find this file int the path of an installed version of Cura.
How can I change that file after install ?
Thanks

fieldOfView:
Unfortunately this cannot be easily changed with the official releases, as the relevant code is distributed in a compiled form. 

This needs to be changed in the source, in two places actually. I'll prepare a pull request so it gets fixed in Cura 2.5. At this point it is unlikely the fix is going to make it in the 2.4 release.

Technical analysis: Cura does try to do the right thing by asking Python to expand the "~" directory to the user profile folder. This should work, but apparently there are situations where it doesn't. Instead of ```os.path.expanduser("~/AppData/Local/cura/")``` we should probably use ```os.path.expandvars("%LOCALAPPDATA%\cura")```. That way we ask what the OS says, instead of Python.

awhiemstra:
This is what the Python documentation says about expanduser:

> On Windows, HOME and USERPROFILE will be used if set, otherwise a combination of HOMEPATH and HOMEDRIVE will be used. An initial ~user is handled by stripping the last directory component from the created user path derived above.

So if that does not actually work, then there is a bug in Python. What does the %USERPROFILE% variable contain for you?

fieldOfView:
Regardless of what USERPROFILE returns, it is better to use the LOCALAPPDATA variable, because a user can change the path of only the local appdata folder.

ljoets:
Ok, thanks.
On computers I manage in our school (windows 7x64 in a domain),
echo %userprofile% -> C:\Users\<my_login>
echo %appadata% -> K:\profil (it is what I need)
[...]

Issue #216, opened 2015-08-04
15.06.03 Raft Not Displaying in Layers View 


Fuzionist1:
The raft is not displaying in Layers view. Instead, the first layer shown is as if there was no raft at all. I have tried complex and completely basic models. 
![capture](https://cloud.githubusercontent.com/assets/13238025/9066902/3254a258-3aa8-11e5-9141-60003a918759.PNG)

Xisle:
It looks like the raft is just not being rendered or included in the layer count. The model gets raised up when Raft is present though.

BagelOrb:
Fixed in next release

PR #1375, opened 2017-01-24
Fix the location of the configuration folder on Windows

fieldOfView:
~/AppData/Local may not be correctly expanded to the AppData folder on systems with itinerant profiles (eg corporate/educational networks where the user profile is stored on the network)

See https://github.com/Ultimaker/Cura/issues/1369 for discussion
Also see the Uranium PR: https://github.com/Ultimaker/Uranium/pull/216

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -50,7 +50,7 @@ import cura.CuraApplication
 import cura.Settings.CuraContainerRegistry
 
 if Platform.isWindows() and hasattr(sys, "frozen"):
-    dirpath = os.path.expanduser("~/AppData/Local/cura/")
+    dirpath = os.path.expandvars("%LOCALAPPDATA%\cura")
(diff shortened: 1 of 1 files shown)


fieldOfView:
Though I feel the PR is more correct than the current situation, it does not fix the problem at hand (#216) and the usecase for what it fixes is just too small. The review burden is high enough as it is, so I am closing this PR.

Ghostkeeper:
I'd say, keep this PR open until the review burden is down a bit.

fieldOfView:
If you insist...

Ghostkeeper:
As far as I'm concerned this is a good change, but it requires someone on Windows to test. @sedwards2009 or @jackha ?

LipuFei:
We also need a Windows whose profile is stored remotely...

PR #1378, opened 2017-01-25
Layerview Legend

fieldOfView:
Now that the code is there to have the colors in the layerview based on the theme, it is fairly trivial to add a legend of the layer "polygon" types to the viewport. This PR does just that.

The PR builds on the themeable viewport branches:
https://github.com/Ultimaker/Cura/pull/1370 (included in this PR)
https://github.com/Ultimaker/Uranium/pull/215 (required by this PR)

legend

CURA-1037 (?)

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -34,15 +34,17 @@ PRIME_CLEARANCE = 6.5
 
 ##  Build volume is a special kind of node that is responsible for rendering the printable area & disallowed areas.
 class BuildVolume(SceneNode):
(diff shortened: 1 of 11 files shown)


fieldOfView:
The list of commits also includes that of the feature_theme_viewport branch. Only the "Add a legend to Layer View" commit is relevant to this PR.

Also note that the feature_theme_viewport branch uses the colors that are currently in master. This legend makes it visible that we use the same color for multiple polygon types.

jackha:
Thanks!

PR #676, opened 2016-03-21
portuguese

CoderSquirrel:
Started to make translation for portuguese

PR #1379, opened 2017-01-25
Added translation files for Brazilian Portuguese

Patola:
As per directions given on PR #676, I created the translation files for brazilian portuguese. All strings are translated and have been subjected to the latest version of template file using msgmerge from gettext.

### DIFF
--- a/resources/i18n/ptbr/fdmprinter.def.json.po
+++ b/resources/i18n/ptbr/fdmprinter.def.json.po
@@ -1,0 +1,5178 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Uranium json setting files\n"
+"Report-Msgid-Bugs-To: http://github.com/ultimaker/uranium\n"
+"POT-Creation-Date: 2016-12-28 10:51+0000\n"
+"PO-Revision-Date: 2017-01-24 01:00-0300\n"
+"Last-Translator: Cláudio Sampaio <patola@makerlinux.com.br>\n"
+"Language-Team: LANGUAGE\n"
+"Language: ptbr\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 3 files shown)


resources/i18n/ptbr/fdmprinter.def.json.po
```diff
@@ -0,0 +1,5178 @@
...
+msgctxt "speed_travel label"
+msgid "Travel Speed"
+msgstr "Velocidade de Viagem"
+
+#: fdmprinter.def.json
+msgctxt "speed_travel description"
+msgid "The speed at which travel moves are made."
+msgstr ""
+"Velocidade em que ocorrem os movimentos de viagem (movimentação do extrusor "
+"s
```
Ghostkeeper: Hmm, I think Github isn't happy with these long files. This comment was meant to be placed at the translation of "Anti Overhang Mesh".

Patola:
Thans for the comments! Yes, I think stripping all quotes from 'Jerk' is better. It is just that it is already used as a jargon term, so translating to a portuguese word would confuse people instead of clarifying. For 'anti support mesh', the reason was that the translation of 'overhang' to portuguese is an awkward expression - "seções pendentes" - and using "anti seções pendentes" would make it even more awkward. But awkward does not equal obscure, it would be easily understood this way also. Should I edit my PR to make these changes?

Ghostkeeper:
We found that the plural forms were wrong, by the way. I corrected them on master.

Ghostkeeper:
Anti support mesh is fine as it is then.

If you feel like editing the "Jerk" thing, feel free. I've already merged this pull request, so please make another one if you feel like making the modification.

Patola:
I came with a more appropriate translation for anti-overhang so I got rid of 'anti-suporte'. Also, I made the changes to jerk with quotes. PR#1386

PR #1386, opened 2017-01-27
Fixed little things according to advice from Ghostkeeper

Patola:
Fixed  two typos, removed quotes from jerk, changed 'anti-suporte' to 'anti-pendências' so that it does not conflict with future changes.


Ghostkeeper:
"Pendente" sounds like a good solution.

PR #1387, opened 2017-01-28
Add preference to reduce the widths of lines shown in the layer view.

smartavionics:
All lines are reduced in width by 10%. Doing this makes it possible to
discern the edges of the skin lines which, otherwise, merge together.

Without this preference enabled it's not at all easy when looking at the layer view to discern the directions of the skin lines because they butt up to each other with no gaps. This preference simply scales the widths of all lines in the layer view to 90% of their original width and so the individual skin lines become visible.

One thing I noticed is that adding this preference requires me to resize the preferences dialog otherwise the lower down preferences start to disappear. Can the dialog be made to auto-size?

### DIFF
--- a/plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
+++ b/plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
@@ -4,6 +4,7 @@
 import gc
 
 from UM.Job import Job
+from UM.Preferences import Preferences
 from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
 from UM.Scene.SceneNode import SceneNode
 from UM.Application import Application
@@ -86,6 +87,8 @@ class ProcessSlicedLayersJob(Job):
 
         current_layer = 0
 
+        reduce_line_widths = bool(Preferences.getInstance().getValue("view/reduce_line_widths"))
+
         for layer in self._layers:
             abs_layer_number = layer.id + abs(min_layer_number)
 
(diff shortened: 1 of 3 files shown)


smartavionics:
Preference not enabled...
screenshot_2017-01-28_08-32-15
Preference enabled...
screenshot_2017-01-28_08-32-49

smartavionics:
I've now changed the preferences dialog to use two columns for the viewport behaviour preferences which reduces the height required.

fieldOfView:
@jackha has been doing a lot of nice work to improve the layer view, making this mostly redundant (though it could be useful to add this in the fallback for no geometry shader):
https://github.com/Ultimaker/Cura/pull/1288

About the preferences dialog and options not fitting... Could you post a screenshot of what the preference window looks like for you? For me (us?) there is ample room for options on the General page. What OS are you on, and what sort of display?

smartavionics:
I look forward to seeing the new shader.

I am using Ubuntu 16.04.

When the viewport behaviour preferences are in a single column, the lower half of the "Send (anonymous) print information" line is missing until I resize the dialog (make it taller).

My latest commit on this branch formats the viewport behaviour options as two columns which gets round the problem.

To be honest, I would be quite happy to have this preference always enabled (and hence it doesn't need to be an preference at all) so that the edges of the skin lines are always visible. OK, that would mean that all the other lines would always be thinner by 10%, but so what?

Ghostkeeper:
Honestly, once the new layer view shader is implemented, I'd rather not have this option in. It renders the lines as square tubes (with interpolated normals to make it look round-ish). That works really well to see the direction and size of the lines.

Perhaps if this would be transformed into a floating point setting that adjusts the line width with a scale, it would still be useful. Repetier has this, and it's sometimes useful to compensate for weird printers like the Ultimaker 2 who measure their flow rate differently. Cura does this automatically so it's unnecessary, but for custom FDM printers (or to compensate for bugs) it might still be useful.

Before anyone tries to implement that though, I'd like to hear more opinions from my colleagues.

smartavionics:
I agree if a new and better layer view shader is on the way, it makes little sense to add this PR. I will just use it myself until the new shader arrives in master. Feel free to close this PR if you don't think it has a future.

PR #1395, opened 2017-01-31
Added anchor_skin_in_infill setting.

smartavionics:
When enabled, skin areas are increased in size so that they project into
the infill by at least the distance between infill lines.

Add checkbox to enable skin anchoring feature.


smartavionics:
Thinking about the min_skin_angle_for_expansion setting some more. I reckon the warning is the wrong way round. It should warn when the angle is shallow (because that increases the shrink distance). Also, the default value should be closer to 90 because that will give a small amount of shrink which is a more sensible default than the yards of shrink it defaults to at the moment.

smartavionics:
Well, maybe not closer to 90 as that makes an extremely small shrink distance so maybe 45 is better. I'm still having problems thinking about this angle. I guess it's because the original feature used a shrink width and I'm still more familiar with that.

BagelOrb:
A shallow angle means a large shrink width, which means that we don't apply the expansion often, but only in a few cases where the angle is really shallow.
If the shrink would be really small then almost all skin would be expanded, also the skin you see on https://ultimaker.com/en/community/39214-feature-add-angle-for-top-bottom-internal-layer-generation

I've aded a minimum angle warning of 2 degrees. Lower than 2 degrees would be absurd and would result in undesirable results for small models.

PR #1398, opened 2017-02-02
Added combing_retract_before_outer_wall setting.

smartavionics:
This boolean setting controls whether travel moves to the first point in an
outer wall will always involve a retraction. IMHO, forcing a retraction has
two benefits:

1 - avoids taking the ooze that would occur during the travel to the outer
surface.

2 - the slight pause when un-retracting could help reduce any ripples
introduced by the rapid movement hot-end movement.


Ghostkeeper:
Something went quite wrong with this pull request!

BagelOrb:
I didn't refer these yet: https://github.com/Ultimaker/CuraEngine/pull/453

PR #1407, opened 2017-02-06
French translation - Little correction

DelphinPETER:


nallath:
Whoops! Thanks for fixing it.

PR #1408, opened 2017-02-07
3mf speedup (CURA-3215)

nallath:
This should significantly speed up working with 3mf files. Loading should be ~10 times as fast and saving should be ~3 times as fast.

### DIFF
--- a/plugins/3MFReader/ThreeMFReader.py
+++ b/plugins/3MFReader/ThreeMFReader.py
@@ -40,97 +40,6 @@ class ThreeMFReader(MeshReader):
         self._base_name = ""
         self._unit = None
 
-    def _createNodeFromObject(self, object, name = ""):
-        node = SceneNode()
-        node.setName(name)
-        mesh_builder = MeshBuilder()
-        vertex_list = []
-
-        components = object.find(".//3mf:components", self._namespaces)
-        if components:
-            for component in components:
-                id = component.get("objectid")
-                new_object = self._root.find("./3mf:resources/3mf:object[@id='{0}']".format(id), self._namespaces)
-                new_node = self._createNodeFromObject(new_object, self._base_name + "_" + str(id))
-                node.addChild(new_node)
-                transform = component.get("transform")
-                if transform is not None:
-                    new_node.setTransformation(self._createMatrixFromTransformationString(transform))
-
-        # for vertex in entry.mesh.vertices.vertex:
-        for vertex in object.findall(".//3mf:vertex", self._namespaces):
-            vertex_list.append([vertex.get("x"), vertex.get("y"), vertex.get("z")])
-            Job.yieldThread()
-
-        xml_settings = list(object.findall(".//cura:setting", self._namespaces))
-
-        # Add the setting override decorator, so we can add settings to this node.
-        if xml_settings:
-            node.addDecorator(SettingOverrideDecorator())
-
-            global_container_stack = Application.getInstance().getGlobalContainerStack()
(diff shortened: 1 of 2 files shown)


fieldOfView:
Would it be possible to keep a fallback to the old code in case libsavitar is not available?

nallath:
Err, hadn't really thought about it, but yes that should be possible. I don't expect much problems though, as the savitar lib (now still private) is pretty darn easy to compile, as it only has SIP as a dependancy.

sedwards2009:
The problem with keeping a fallback is that we would have two 3MF parser implementations to maintain instead of just one.

jackha:
I don't think we want a fallback, after merging and updating build systems the old 3mf loading/saving is history.

fieldOfView:
My reason to ask for a fallback is that I think relying on a compiled libsavitar makes setting up a dev-system more complex and may increase the barrier for contribution. Note that there is currently a python fallback for stl reading if numpy-stl is not available.

jackha:
That's a good one.

thopiekar:
+1, for the fallback code. Also remembered @nallath about fallback code in our STLReader in the past and it was quite useful as we had some problems with NumpySTL (Isn't it? Or am I remembering something wrong?..)

thopiekar:
@fieldOfView Correct, I would be forced to invest time in packaging Numpy for my PPA 😄
(However, users already ask for a NumPy-STL package 🙄 )

nallath:
The STL reader is significantly easier and isn't still in "flux" (Not that 3mf changes much, but it can change).

sedwards2009:
I'll copy my response to this question on JIRA here up on github, then @fieldOfView can see it too.

"""
I'm against the idea of keeping two code paths. It just means twice the places where bugs can appear, twice the amount of work when ever we change the 3MF code, and to be honest we are only going to QA the C++ version. The pure Python version will quickly bit-rot and its quality will drop.

If you can get libArcus built, then the 3MF C++ part is easy.

A much better way of lowering the barrier to contribution would be to generate and publish Python wheels for libArcus and 3MF, and a nightly builds of CuraEngine. That would bring Cura set up down to grabbing cura and uranium git repos, installing the wheels (maybe via pip) and copying an engine binary to the right place.
"""

thopiekar:
So what about moving the old code into a separate file (3MFReaderLegacy.py) and let the community care about bugs in there? So whenever a problem with 3MF occurs we just focus as an company on our solution.

Btw. is libSavitar opensource and accessible?

thopiekar:
Ok, stupid question.. now it is..

fieldOfView:
> If you can get libArcus built, then the 3MF C++ part is easy.

The thing is, I've never had to build libarcus on windows. libarcus has seen so little changes that I could just pick a compiled version from a binary release of Cura and work with that. libSavitar sounds like it will see more changes - if only because of 3mf changes (but I have a feeling libSavitar is going to be more than just 3mf, or you would have named it lib3mf).

awhiemstra:
I sure hope Savitar is not going to see a lot of changes, since it is going to be part of the environment on the new build system (because Python) and not things we compile every time.

nallath:
I do like the wheels idea though. That would make it a whole lot easier to get started. 

As for changes, right now libSavitar is 3mf only, but it might not stay that way. I don't think there is a big chance that it will start supporting other file formats soon, but for the off chance that we do, i gave it a generic name (of a bad guy! whoo!)

thopiekar:
My PPA is ready for the merge :muscle:

Ghostkeeper:
It better be!

PR #1409, opened 2017-02-07
Improve performance of switching variant/material/quality

awhiemstra:
These are the changes needed in the Cura code related to https://github.com/Ultimaker/Uranium/pull/221 . The most imporant change here is to postpone the containersChanged signal of the active container stacks.


nallath:
Whoops. Missed this one. My bad.

PR #1412, opened 2017-02-08
Pre-heat Bed and Redesign Print Monitor

Ghostkeeper:
This adds a feature to pre-heat the print bed in order to have it be warm by the time you're done slicing and such. It also basically redesigns the side bar of the print monitor.

Implements issue CURA-3161.

### DIFF
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -170,6 +170,12 @@ class PrinterOutputDevice(QObject, OutputDevice):
     def address(self):
         Logger.log("w", "address is not implemented by this output device.")
 
+    ##  A human-readable name for the device.
+    @pyqtProperty(str, constant = True)
+    def name(self):
+        Logger.log("w", "name is not implemented by this output device.")
+        return ""
+
     @pyqtProperty(str, notify = errorTextChanged)
     def errorText(self):
         return self._error_text
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2017 Ultimaker B.V.
+# Cura is released under the terms of the AGPLv3 or higher.
+
 from UM.i18n import i18nCatalog
 from UM.OutputDevice.OutputDevice import OutputDevice
 from PyQt5.QtCore import pyqtProperty, pyqtSignal, pyqtSlot, QObject
@@ -357,14 +360,14 @@ class PrinterOutputDevice(QObject, OutputDevice):
         result = []
         for material_id in self._material_ids:
             if material_id is None:
-                result.append("#800000FF") #No material.
+                result.append("#00000000") #No material.
                 continue
 
             containers = self._container_registry.findInstanceContainers(type = "material", GUID = material_id)
             if containers:
                 result.append(containers[0].getMetaDataEntry("color_code"))
             else:
-                result.append("#800000FF") #Unknown material.
+                result.append("#00000000") #Unknown material.
         return result
 
     ##  Protected setter for the current material id.
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -260,6 +260,23 @@ class PrinterOutputDevice(QObject, OutputDevice):
     def _setTargetBedTemperature(self, temperature):
         Logger.log("w", "_setTargetBedTemperature is not implemented by this output device")
 
+    ##  Pre-heats the heated bed of the printer.
+    #
+    #   \param temperature The temperature to heat the bed to, in degrees
+    #   Celsius.
+    #   \param duration How long the bed should stay warm, in seconds. Defaults
+    #   to a quarter hour.
+    @pyqtSlot(float, float)
+    def preheatBed(self, temperature, duration):
+        Logger.log("w", "preheatBed is not implemented by this output device.")
+
+    ##  Cancels pre-heating the heated bed of the printer.
+    #
+    #   If the bed is not pre-heated, nothing happens.
+    @pyqtSlot()
+    def cancelPreheatBed(self):
+        Logger.log("w", "cancelPreheatBed is not implemented by this output device.")
+
     ##  Protected setter for the current bed temperature.
     #   This simply sets the bed temperature, but ensures that a signal is emitted.
     #   /param temperature temperature of the bed.
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -264,8 +264,7 @@ class PrinterOutputDevice(QObject, OutputDevice):
     #
     #   \param temperature The temperature to heat the bed to, in degrees
     #   Celsius.
-    #   \param duration How long the bed should stay warm, in seconds. Defaults
-    #   to a quarter hour.
+    #   \param duration How long the bed should stay warm, in seconds.
     @pyqtSlot(float, float)
     def preheatBed(self, temperature, duration):
         Logger.log("w", "preheatBed is not implemented by this output device.")
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -165,6 +165,11 @@ class PrinterOutputDevice(QObject, OutputDevice):
             self._job_name = name
             self.jobNameChanged.emit()
 
+    ##  Gives a human-readable address where the device can be found.
+    @pyqtProperty(str, constant = True)
+    def address(self):
+        Logger.log("w", "address is not implemented by this output device.")
+
     @pyqtProperty(str, notify = errorTextChanged)
     def errorText(self):
         return self._error_text
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -203,7 +203,9 @@ class PrinterOutputDevice(QObject, OutputDevice):
             self._target_bed_temperature = temperature
             self.targetBedTemperatureChanged.emit()
 
-    ##
+    ##  The duration of the time-out to pre-heat the bed, in seconds.
+    #
+    #   \return The duration of the time-out to pre-heat the bed, in seconds.
     @pyqtProperty(int)
     def preheatBedTimeout(self):
         return self._preheat_bed_timeout
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -45,6 +45,7 @@ class PrinterOutputDevice(QObject, OutputDevice):
         self._job_name = ""
         self._error_text = ""
         self._accepts_commands = True
+        self._preheat_bed_timeout = 900 #Default time-out for pre-heating the bed, in seconds.
 
         self._printer_state = ""
         self._printer_type = "unknown"
@@ -198,6 +199,11 @@ class PrinterOutputDevice(QObject, OutputDevice):
         if self._target_bed_temperature != temperature:
             self._target_bed_temperature = temperature
             self.targetBedTemperatureChanged.emit()
+
+    ##
+    @pyqtProperty(int)
+    def preheatBedTimeout(self):
+        return self._preheat_bed_timeout
 
     ## Time the print has been printing.
     #  Note that timeTotal - timeElapsed should give time remaining.
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
@@ -318,14 +318,14 @@ class NetworkPrinterOutputDevice(PrinterOutputDevice):
         if auth_state == AuthState.AuthenticationRequested:
             Logger.log("d", "Authentication state changed to authentication requested.")
             self.setAcceptsCommands(False)
-            self.setConnectionText(i18n_catalog.i18nc("@info:status", "Connected over the network to {0}. Please approve the access request on the printer.").format(self.name))
+            self.setConnectionText(i18n_catalog.i18nc("@info:status", "Connected over the network. Please approve the access request on the printer."))
             self._authentication_requested_message.show()
             self._authentication_request_active = True
             self._authentication_timer.start()  # Start timer so auth will fail after a while.
         elif auth_state == AuthState.Authenticated:
             Logger.log("d", "Authentication state changed to authenticated")
             self.setAcceptsCommands(True)
-            self.setConnectionText(i18n_catalog.i18nc("@info:status", "Connected over the network to {0}.").format(self.name))
+            self.setConnectionText(i18n_catalog.i18nc("@info:status", "Connected over the network."))
             self._authentication_requested_message.hide()
             if self._authentication_request_active:
                 self._authentication_succeeded_message.show()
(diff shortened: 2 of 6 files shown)


resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+                        border.width: UM.Theme.getSize("default_lining").width
+                        border.color: UM.Theme.getColor("lining")
+                        visible: printerConnected
+                        anchors.left: parent.left
+                        anchors.leftMargin: UM.Theme.getSize("default_margin").width
+                        anchors.verticalCenter: materialName.verticalCenter
+                    }
+                    Text //Material name.
+                    {
+                        id: materialName
```
awhiemstra: Shouldn't these also be hidden if there is no printer connected?

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+                model: machineExtruderCount.properties.value
+                delegate: Rectangle
+                {
+                    id: extruderRectangle
+                    color: UM.Theme.getColor("sidebar")
+                    width: extrudersGrid.width / 2 - UM.Theme.getSize("sidebar_lining_thin").width / 2
+                    height: UM.Theme.getSize("sidebar_extruder_box").height
+                    Layout.fillWidth: index == machineExtruderCount.properties.value - 1 && index % 2 == 0
+
+                    Text //Extruder name.
```
awhiemstra: You're a bit inconsistent with using Label or Text. Since we use custom fonts etc. it is not much of a problem, but it is nice to be consistent. I would suggest using Label as it properly falls back to using system fonts etc.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+            id: bedCurrentTemperature
+            text: printerConnected ? connectedPrinter.bedTemperature + "°C" : ""
+            font: UM.Theme.getFont("large")
+            color: UM.Theme.getColor("text")
+            anchors.right: bedTargetTemperature.left
+            anchors.rightMargin: UM.Theme.getSize("setting_unit_margin").width
+            anchors.top: parent.top
+            anchors.topMargin: UM.Theme.getSize("default_margin").height
+        }
+        Rectangle //Input field for pre-heat temperature.
```
awhiemstra: It should in theory be possible to access SettingTextField from here and include it, rather than copying everything over. However, you might need to link a bunch of properties. Just something to consider, not blocking.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+            sourceSize.height: height
+            color: UM.Theme.getColor("text")
+            visible: preheatCountdown.visible
+            source: UM.Theme.getIcon("print_time")
+            anchors.right: preheatCountdown.left
+            anchors.rightMargin: UM.Theme.getSize("default_margin").width / 2
+            anchors.verticalCenter: preheatCountdown.verticalCenter
+        }
+
+        Timer
```
awhiemstra: I think it is better to put this timer into the actual print device. That way the pre-heat timer becomes a device specific thing and it would become possible to have multiple devices pre-heating. In addition, it makes things a lot simpler when (if?) the machine exposes the pre-heat timer.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+            property var endTime: new Date() //Set initial endTime to be the current date, so that the endTime has initially already passed and the timer text becomes invisible if you were to update.
+            function update()
+            {
+                var now = new Date();
+                if (now.getTime() < endTime.getTime())
+                {
+                    var remaining = endTime - now; //This is in milliseconds.
+                    var minutes = Math.floor(remaining / 60 / 1000);
+                    var seconds = Math.floor((remaining / 1000) % 60);
+                    preheatCountdown.text = minutes + ":" + (seconds < 10 ? "0" + seconds : seconds);
```
awhiemstra: Why not use Qt.formatTime ?

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+            anchors.verticalCenter: preheatButton.verticalCenter
+        }
+
+        Button //The pre-heat button.
         {
-            sourceComponent: monitorItem
-            property string label: catalog.i18nc("@label", "Build plate")
-            property string value: printerConnected ? Math.round(connectedPrinter.bedTemperature) + "°C" : ""
+            id: preheatButton
+            text: preheatCountdownTimer.running ? catalog.i18nc("@button Cancel pre-heating", "Cancel") : catalog.i18nc("@button", "Pre-heat")
```
awhiemstra: "@action:button" for the context

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+                        text: control.text;
+                    }
+                }
+            }
+
+            onClicked:
+            {
+                if (!preheatCountdownTimer.running)
+                {
+                    connectedPrinter.preheatBed(preheatTemperatureInput.text, connectedPrinter.preheatBedTimeout);
```
awhiemstra: There's a lot of logic in this QML. As mentioned above, it is probably a good idea to move this into the device. At the very least, most of the logic should be moved to Python.

cura/PrinterOutputDevice.py
```diff
@@ -323,6 +348,28 @@ def materialNames(self):
...
+    def materialColors(self):
+        result = []
+        for material_id in self._material_ids:
+            if material_id is None:
+                result.append("#00000000") #No material.
+                continue
+
+            containers = self._container_registry.findInstanceContainers(type = "material", GUID = material_id)
+            if containers:
+                result.append(containers[0].getMetaDataEntry("color_code"))
```
Ghostkeeper: Qt expects it to be in ARGB, so I'm having to convert this.

Ghostkeeper: Our colour codes are in RRGGBB order, not RRGGBBAA or AARRGGBB.

fieldOfView:
FYI, there are some strange things going on when switching between printers while in the print monitor:
switch

To reproduce:
* I have a (not-connected) UM3 selected in Cura, and go to the print monitor (left)
* I switch to a (OctoPrint connected) UM2+ while in the monitor (middle)
* I switch back to the UM3 (right)

Also note the preheat button seems to have its text rendered twice.

I am curious to see how this will work out with machines with (say) 5 extruders.
This UI seems to have mostly (only?) considered the UM3. For a lot of printers, it is always going to say "unknown material". 

Could the tooltip of the preheat-button use the sidebar tooltip style instead?

fieldOfView:
Could the sidebar check if the device implements preHeatBed? As it is now, I just get a 15 minute countdown and nothing else happens. Where does the 15 minutes come from?

I find the 3 temperatures divided into two areas quite confusing.

Why is hotend in a different font/size than the rest?
hotend

Ghostkeeper:
> there are some strange things going on when switching between printers while in the print monitor

Reproduced this as well. That will need some work. It seems to resolve itself upon switching to settings view and back to print monitor.

> Also note the preheat button seems to have its text rendered twice.

Yeah I don't know what causes that. It's only once listed in the QML. It seems it was rendered once in bold and once in normal face...

> I am curious to see how this will work out with machines with (say) 5 extruders.

It should render the last extruder with double width if the last extruder is on the left side. This condition should also work with 5 extruders or 4.

> This UI seems to have mostly (only?) considered the UM3. For a lot of printers, it is always going to say "unknown material".

I've considered the UM3 and UM2/UMO, while keeping in mind that other printers may connect via USB as well and that Ultimaker might manufacture wholly different printers in the future. The "unknown material" is unfortunate though. I don't really know what we can do about it. Shall we leave it invisible if the material is unknown? I guess that would be better.

> Could the tooltip of the preheat-button use the sidebar tooltip style instead?

I tried that but didn't get it to work immediately and had to scrap it for time constraints. I'll try again now that we have a little bit more time.

> lots of these QML warnings when switching back to the UM3

Yeah, probably due to that update bug.

> Could the sidebar check if the device implements preHeatBed?

Well, as it stands now, if the printer doesn't support actual pre-heating with a cooldown, the idea is to send a normal g-code for heating the bed and then sending another g-code that sets the bed temperature back to 0 once the timer has ran out. It is possible that the user closes Cura, thereby interrupting the timer and preventing Cura from ever setting the bed temperature back to 0. Unless the printer implements its own timer, there is no real way to fix this, since the connection is just broken so Cura can't exert any influence. If the device doesn't implement `setTargetBedTemperature` either, then you get a warning about that in the log; it doesn't disable the button (something I'd like to have changed but it's a bit involved to implement that). Also, USB doesn't implement getting the target bed temperature so that grey number stays at 0 degrees. I'd have liked to remove it but it's also a bit involved unless I always remove it when it's at 0 degrees (also for UM3).

> I find the 3 temperatures divided into two areas quite confusing.

Agreed. It's a bit better if it shows the complete data on the UM3, but still bad. I discussed the option of adding more labels with our UX designer but he didn't want any other text there to "keep it clean".

> Why is hotend in a different font/size than the rest?

Looks like I forgot to add the "font" property to that item. Funny, on my computer it looks the same.

Ghostkeeper:
Thanks for your tests, anyway :smile: It's really the first major UI feature I've implemented.

fieldOfView:
For the record: sorry if my review seemed somewhat brusk. I did not have too much time but wanted to give some feedback on the new functionality.

Ghostkeeper:
> sorry if my review seemed somewhat brusk.

Your feedback is greatly appreciated. I'm doing with it what I deem best. I don't get insulted very easily, you know me.

plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
```diff
@@ -220,6 +222,11 @@ def getProperty(self, key):
     def getKey(self):
         return self._key
 
+    ##  The IP address of the printer.
+    @pyqtProperty(str, constant = True)
+    def address(self):
```
Ghostkeeper: Agreed. Done in e37d8b949e5427d7564da6e3c1af4cc9cbf04bcc.

fieldOfView:
Note that I totally misunderstood the 15 minute timer. I thought it indicated and estimation of the time until the bed temperature was reached (and WTF'ed about the long estimation).

fieldOfView:
>> Also note the preheat button seems to have its text rendered twice.

> I don't know what causes that. I haven't seen it here. It's only once listed in the QML. It seems it was rendered once in bold and once in normal face...

It is because you specify the text in the Button, and then use style to draw another label. The ButtonStyle Label apparently does not replace the Button caption, but is drawn underneath it. Just move the text property to the style.

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -124,6 +124,16 @@ def _homeHead(self):
     def _homeBed(self):
         self._sendCommand("G28 Z")
 
+    ##  A name for the device.
+    @pyqtProperty(str, constant = True)
+    def name(self):
```
Ghostkeeper: Agreed. Done in 0c9b9a3033cfe7e110103db15ce9acd7ec5a5b8a.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
-        anchors.leftMargin: UM.Theme.getSize("default_margin").width
+        id: connectedPrinterHeader
+        width: parent.width
+        height: childrenRect.height + UM.Theme.getSize("default_margin").height * 2
+        color: UM.Theme.getColor("setting_category")
 
         Label
         {
+            id: connectedPrinterNameLabel
+            text: printerConnected ? connectedPrinter.name : catalog.i18nc("@info:status", "No printer connected")
```
Ghostkeeper: It seems to be defined in several places:
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/resources/qml/Sidebar.qml#L32
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/plugins/UM3NetworkPrinting/UM3InfoComponents.qml#L14
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/resources/qml/Preferences/MachinesPage.qml#L140
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/resources/qml/Menus/NozzleMenu.qml#L16
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/resources/qml/MonitorButton.qml#L18
* https://github.com/Ultimaker/Cura/blob/feature_preheat_bed/resources/qml/Menus/MaterialMenu.qml#L16

The one in Sidebar seems to be most closely related. I think it's that one, but I'm not sure how that property ends up here then. I'll refactor this to make it not use this property any more.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
+            font: UM.Theme.getFont("large")
+            color: UM.Theme.getColor("text")
+            anchors.left: parent.left
```
Ghostkeeper: Done in 60812139b78c927b52b91c8264adaa8ff4bb7989.

Note that this will cause the width of the default margin to be used for vertical margin. However it seems we already do this in several places, so I'll follow suit.

resources/qml/PrintMonitor.qml
```diff
@@ -20,49 +20,462 @@ Column
...
-            property string value: printerConnected ? Math.round(connectedPrinter.hotendTemperatures[index]) + "°C" : ""
+            id: extrudersGrid
+            columns: 2
+            columnSpacing: UM.Theme.getSize("sidebar_lining_thin").width
+            rowSpacing: UM.Theme.getSize("sidebar_lining_thin").height
+            width: parent.width
+
+            Repeater
+            {
+                model: machineExtruderCount.properties.value
```
Ghostkeeper: That doesn't work for single-extrusion machines since `extrudersModel` is empty then.

resources/qml/PrintMonitor.qml
Ghostkeeper: I was being consistent in what I thought was right: Label for labels, text for data.

Since you say Label works better if fonts can't be found, I'll change everything to use Label.

resources/qml/PrintMonitor.qml
Ghostkeeper: The text is an empty string then. It needs to have a fallback for the text anyway or we're accessing `connectedPrinter.materialNames` when `connectedPrinter` is `null`.

resources/qml/PrintMonitor.qml
Ghostkeeper: I experimented with this, but `SettingTextField` has functionality that we don't want here, such as applying the value of the setting to the SettingPropertyProvider, and colouring for invalid values. In the end I chose against it.

resources/qml/PrintMonitor.qml
Ghostkeeper: `Qt.formatTime` requires a `QTime` object or `QDateTime` object. I just have an integer in milliseconds. There is no "time interval" object in QT it seems (though [this stupid Wikipedia page](https://en.wikipedia.org/wiki/Qt_interval) ruins a lot of the searching I did on the internet). `QTime` wraps around at midnight. `QDateTime` can be used but constructing it requires constructing first a `QDateTime` object representing midnight of January 1st of the year 0, then adding a number of milliseconds to it, and that just seems like overkill.

This can all be moved into Python, luckily.

PR #1421, opened 2017-02-11
FEATURE: Support wsh usb driver (for mac)

Torsten85:
This pr adds support for the wch driver. This driver is e.g. necessary to connect an anet a8 to a mac.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -267,7 +267,7 @@ class USBPrinterOutputDeviceManager(QObject, OutputDevicePlugin, Extension):
                 pass
         else:
             if only_list_usb:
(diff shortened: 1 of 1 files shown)


fieldOfView:
Shouldn't the non-usb-only path also include the wchusb* variants?

Ghostkeeper:
I'd add it there too, but I don't know how to go about testing my code.

Should we just try without testing? Can't really hurt in this case, perhaps.

thopiekar:
I don't think it will hurt. We are just expanding the list..
Additionally, as far as I know, we don't care about USB printing that much anymore, so I felt free to merge it directly.
Was it a bad idea?

awhiemstra:
Nah, the worst that can happen is that we scan a few more devices to see if they are printers.

Ghostkeeper:
Done in a3af887d3f6a9bd6ef16c6a3f675b6ec08b93756.

PR #1426, opened 2017-02-13
Added basic profile for folgertech

nallath:
As per info provided by Paul Bussiere


resources/definitions/folgertech_FT-5.def.json
```diff
@@ -0,0 +1,29 @@
+{
+    "id": "FolgerTech_FT5",
```
Ghostkeeper: Not that we use this ID any more, but it is not consistent with the file name.

nallath: Forgot to remove that from the file.

Ghostkeeper:
Thanks for your contribution, Paul (and Nallath)!

fieldOfView:
Does that mesh really need to be 2 Mb? It seems like it must have lots of unnecessarily detailed geometry, because the actual model looks fairly simple.

nallath:
Probably not, but I just added the file as I got it. My modeling skills are quite bad, so it would have taken me quite a while to fix it.

PR #1429, opened 2017-02-14
Add missing label color

fieldOfView:
One of the labels in the newly redesigned printmonitor is lacking a color, thus making the hotend temperature unstylable. This PR fixes that.

TL;DR: "Oops, you missed a spot"


awhiemstra:
Seems simple enough, merged. :)

PR #1432, opened 2017-02-15
Feature manual slice / disable auto slice

jackha:
Adds a preference "Slice automatically":
- If turned on, Cura behaves as before, automatically slicing when needed. CuraEngineBackend has been changed a lot to make it more efficient. For example, when moving an object, the engine is now restarted only once instead of twice.
- If turned off, the "Save to File" (or a different device text) changes to Prepare whenever you change something that requires a reslice. During slice you can press Cancel. After slicing, you see "Save to File" again.


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -30,11 +31,12 @@
...
 catalog = i18nCatalog("cura")
 
-class CuraEngineBackend(Backend):
+class CuraEngineBackend(QObject, Backend):
     ##  Starts the back-end plug-in.
     #
     #   This registers all the signal listeners and prepares for communication
     #   with the back-end in general.
+    #   CuraEngineBackend is exposed to qml as well.
     def __init__(self):
```
nallath: I was under the impression that for QObjects you always need to set the parent. This is what we do for every other case where we use QObject. @awhiemstra Could you chime in on this?

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -148,31 +154,46 @@ def getEngineCommand(self):
...
-        if self._pause_slicing or self._block_slicing:
-            return
         self._slice_start_time = time()
-        if not self._enabled or not self._global_container_stack:  # We shouldn't be slicing.
-            # try again in a short time
-            self._change_timer.start()
+        if not self._need_slicing:
+            self.processingProgress.emit(1.0)
+            self.backendStateChange.emit(BackendState.Done)
+            Logger.log("w", "Do not need to slice.")
```
nallath: I find this a vague error message. Did not need to slice or "Nothing to slice" is better imho. (although I don't feel very strong about this)

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -301,41 +307,58 @@ def _onStartSliceCompleted(self, job):
...
+        for node in DepthFirstIterator(self._scene.getRoot()):
+            if node.callDecoration("isBlockSlicing"):
+                enable_timer = False
+                self.backendStateChange.emit(BackendState.Disabled)
+            gcode_list = node.callDecoration("getGCodeList")
+            if gcode_list is not None:
+                self._scene.gcode_list = gcode_list
+
+        if self._use_timer == enable_timer:
+            return
```
nallath: Return either true or false. None correct in most cases, but not if someone explicitly decides to check for determineAutoSlicing() == False
If this is intentional, please update the function documentation accordingly.

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -301,41 +307,58 @@ def _onStartSliceCompleted(self, job):
...
-            if gcode_list is not None:
-                self._scene.gcode_list = gcode_list
-
-        self._block_slicing = block_slicing
-
-        if should_pause or self._block_slicing:
-            self.pauseSlicing()
-        else:
-            self.continueSlicing()
+        self.determineAutoSlicing()
```
nallath: Shouldn't we do something with the result of the determineAutoSlicing function here?

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -350,16 +373,34 @@ def _onSocketError(self, error):
...
 
+    ##  Remove old layer data (if any)
+    def _clearLayerData(self):
+        for node in DepthFirstIterator(self._scene.getRoot()):
+            if node.callDecoration("getLayerData"):
+                node.getParent().removeChild(node)
+                break
+
+    ##  Convenient function: set need_slicing, emit state and clear layer data
+    def needSlicing(self):
```
nallath: Nitpick; I think it should be needsSlicing (as in; This needs Slicing)

plugins/CuraEngineBackend/CuraEngineBackend.py
awhiemstra: The QObject constructor is `QObject(QObject* parent = nullptr)`. This means that if you do not pass a parent instance, it will be unparented and not automatically destroyed when its parent gets destroyed. Since we are dealing with Python here, it is not strictly necessary to give a QObject a parent.

However, if there is a chance that a QObject can survive past the lifetime of the Python wrapper, it *should* have a parent as that prevents Python/PyQt from destroying the Qt/C++ side of the object when the Python side is destroyed. This is quite relevant for things that are created from QML, less so for things created from Python. In general, however, it is best to be safe and implement __init__ of a QObject subclass as `__init__(self, parent = None)`.

plugins/CuraEngineBackend/CuraEngineBackend.py
jackha: Agree, I'm going to change it. It was: "Do not need to slice, optimization possible or programming error" ;-) The intended use is to see if there are unnecessary attempts to slice.

plugins/CuraEngineBackend/CuraEngineBackend.py
jackha: Oops

plugins/CuraEngineBackend/CuraEngineBackend.py
jackha: Ahh it does implicitly. _onChanged looks at _use_timer, which is set in determineAutoSlicing.

PR #1436, opened 2017-02-17
Refactor properties that start with 'get' to avoid confusion

fieldOfView:
…between slots & properties

Case in point: LayerViewProxy.getLayerViewType was decorated as a property but was used/intended as a slot. This PR refactors property names to not start with the verb "get".


nallath:
Thanks for the fix (as always!)

PR #1443, opened 2017-02-18
Printer helloBEEprusa added.

ruimartinsptl:
more info: https://beeverycreative.com/forum/viewtopic.php?f=8&t=389


nallath:
Thanks for adding it :)

fieldOfView:
Does that mesh really need to be 2.6 Mb? It seems like it must have lots of unnecessary details, because the actual model looks fairly simple.

PR #1446, opened 2017-02-19
 Optimise layerview extruder visibility options

fieldOfView:
~~This PR removes a typo~~

### DIFF
--- a/plugins/LayerView/LayerView.py
+++ b/plugins/LayerView/LayerView.py
@@ -73,10 +73,7 @@ class LayerView(View):
         Preferences.getInstance().addPreference("view/force_layer_view_compatibility_mode", False)
 
         Preferences.getInstance().addPreference("layerview/layer_view_type", 0)
-        Preferences.getInstance().addPreference("layerview/extruder0_opacity", 1.0)
-        Preferences.getInstance().addPreference("layerview/extruder1_opacity", 1.0)
-        Preferences.getInstance().addPreference("layerview/extruder2_opacity", 1.0)
-        Preferences.getInstance().addPreference("layerview/extruder3_opacity", 1.0)
+        Preferences.getInstance().addPreference("layerview/extruder_opacities", "")
 
         Preferences.getInstance().addPreference("layerview/show_travel_moves", False)
         Preferences.getInstance().addPreference("layerview/show_support", True)
@@ -392,10 +389,13 @@ class LayerView(View):
 
         self.setLayerViewType(int(float(Preferences.getInstance().getValue("layerview/layer_view_type"))));
 
-        self.setExtruderOpacity(0, float(Preferences.getInstance().getValue("layerview/extruder0_opacity")))
-        self.setExtruderOpacity(1, float(Preferences.getInstance().getValue("layerview/extruder1_opacity")))
-        self.setExtruderOpacity(2, float(Preferences.getInstance().getValue("layerview/extruder2_opacity")))
-        self.setExtruderOpacity(3, float(Preferences.getInstance().getValue("layerview/extruder3_opacity")))
+        extruder_nr = 0
+        for extruder_opacity in Preferences.getInstance().getValue("layerview/extruder_opacities").split(","):
+            try:
+                opacity = float(extruder_opacity)
+            except ValueError:
+                opacity = 1.0
+            self.setExtruderOpacity(extruder_nr, opacity)
 
         self.setShowTravelMoves(bool(Preferences.getInstance().getValue("layerview/show_travel_moves")))
         self.setShowSupport(bool(Preferences.getInstance().getValue("layerview/show_support")))
--- a/plugins/LayerView/LayerView.py
+++ b/plugins/LayerView/LayerView.py
@@ -396,6 +396,7 @@ class LayerView(View):
             except ValueError:
                 opacity = 1.0
             self.setExtruderOpacity(extruder_nr, opacity)
+            extruder_nr += 1
 
         self.setShowTravelMoves(bool(Preferences.getInstance().getValue("layerview/show_travel_moves")))
         self.setShowSupport(bool(Preferences.getInstance().getValue("layerview/show_support")))
--- a/plugins/LayerView/LayerView.qml
+++ b/plugins/LayerView/LayerView.qml
@@ -165,11 +165,11 @@ Item
         Component.onCompleted:
         {
             layerViewTypes.append({
-                text: catalog.i18nc("@title:layerview mode", "Material Color"),
+                text: catalog.i18nc("@label:listbox", "Material Color"),
                 type_id: 0
             })
             layerViewTypes.append({
-                text: catalog.i18nc("@title:layerview mode", "Line Type"),
+                text: catalog.i18nc("@label:listbox", "Line Type"),
                 type_id: 1  // these ids match the switching in the shader
             })
         }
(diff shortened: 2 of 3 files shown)


nallath:
Hmm, merge conflicts. Could you have a look at them? I think Jack got to the code first...

fieldOfView:
Retracted the removal of duplicate code because of conflicts with recent changes by @jackha

fieldOfView:
This PR refactors the duplicated maximum of 4 checkboxes to turn on/off individual extruders in layerview. It replaces the code that currently has a typo.

This PR also moves to lingering compatibility mode preferences from the general page into the layerview options, and it enables translation of the layerview modes.

plugins/LayerView/LayerView.qml
```diff
@@ -248,42 +242,19 @@ Item
...
-                text: "Extruder 2"
-                visible: !UM.LayerView.compatibilityMode && (UM.LayerView.extruderCount >= 2)
-            }
-            CheckBox {
-                checked: view_settings.extruder2_checked
-                onClicked: {
-                    UM.Preferences.setValue("layerview/extruder2_opacity", checked ? 1.0 : 0.0);
-                }
-                text: "Extruder 3"
-                visible: !UM.LayerView.compatibilityMode && (UM.LayerView.etruderCount >= 3)
```
fieldOfView: Here's that typo. Byebye

plugins/LayerView/LayerView.py
```diff
@@ -392,10 +389,13 @@ def _updateWithPreferences(self):
...
-        self.setExtruderOpacity(1, float(Preferences.getInstance().getValue("layerview/extruder1_opacity")))
-        self.setExtruderOpacity(2, float(Preferences.getInstance().getValue("layerview/extruder2_opacity")))
-        self.setExtruderOpacity(3, float(Preferences.getInstance().getValue("layerview/extruder3_opacity")))
+        extruder_nr = 0
+        for extruder_opacity in Preferences.getInstance().getValue("layerview/extruder_opacities").split(","):
+            try:
+                opacity = float(extruder_opacity)
+            except ValueError:
+                opacity = 1.0
+            self.setExtruderOpacity(extruder_nr, opacity)
```
nallath: So you always set the extruderOpacity of the first extruder? I'm guessing you missed a extruder_nr += 1 somehwere in the loop.

plugins/LayerView/LayerView.qml
```diff
@@ -214,72 +218,47 @@ Item
...
-            }
-            CheckBox {
-                checked: view_settings.extruder3_checked
-                onClicked: {
-                    UM.Preferences.setValue("layerview/extruder3_opacity", checked ? 1.0 : 0.0);
+            Repeater {
+                model: UM.LayerView.extruderCount
+                CheckBox {
+                    checked: [undefined, ""].indexOf(view_settings.extruder_opacities[index]) >= 0 || view_settings.extruder_opacities[index] > 0.5
```
nallath: I know what you're doing here, but i don't realy like it. Seems a bit of a weird way of doing this.

awhiemstra: It is very Pythonic, which is pretty wrong in Javascript code :p

plugins/LayerView/LayerView.qml
```diff
@@ -214,72 +218,47 @@ Item
...
-                visible: !UM.LayerView.compatibilityMode && (UM.LayerView.etruderCount >= 3)
-            }
-            CheckBox {
-                checked: view_settings.extruder3_checked
-                onClicked: {
-                    UM.Preferences.setValue("layerview/extruder3_opacity", checked ? 1.0 : 0.0);
+            Repeater {
+                model: UM.LayerView.extruderCount
+                CheckBox {
+                    checked: view_settings.extruder_opacities[index] > 0.5 || view_settings.extruder_opacities[index] == undefined !! view_settings.extruder_opacities[index] == ""
```
nallath: I think this should be || ;) 

I can't test the code unfortunately, as I'm forced to use the compatibility mode.

fieldOfView: Fixed that.

Forced to use the compatibility mode? Then you will be thrilled that I moved the two preferences from the General page to the layerview options.

fieldOfView:
Would be nice if this could get another look. The typo is still there, and this PR replaces the code that the typo is in.

plugins/LayerView/LayerView.py
```diff
@@ -392,10 +389,14 @@ def _updateWithPreferences(self):
 
         self.setLayerViewType(int(float(Preferences.getInstance().getValue("layerview/layer_view_type"))));
 
-        self.setExtruderOpacity(0, float(Preferences.getInstance().getValue("layerview/extruder0_opacity")))
-        self.setExtruderOpacity(1, float(Preferences.getInstance().getValue("layerview/extruder1_opacity")))
-        self.setExtruderOpacity(2, float(Preferences.getInstance().getValue("layerview/extruder2_opacity")))
-        self.setExtruderOpacity(3, float(Preferences.getInstance().getValue("layerview/extruder3_opacity")))
+        extruder_nr = 0
+        for extruder_opacity in Preferences.getInstance().getValue("layerview/extruder_opacities").split(","):
```
awhiemstra: Could we use a different separator? We have run into issues using , as separator before and it is awfully close to localized formatting of floats.

plugins/LayerView/LayerView.qml
```diff
@@ -160,14 +160,18 @@ Item
...
         {
             id: layerViewTypes
-            ListElement {
-                text: "Material Color"
+        }
+
+        Component.onCompleted:
+        {
+            layerViewTypes.append({
+                text: catalog.i18nc("@title:layerview mode", "Material Color"),
```
awhiemstra: `@label:listbox` for context.

plugins/LayerView/LayerView.qml
```diff
@@ -315,6 +294,22 @@ Item
...
             }
+            CheckBox {
+                checked: view_settings.only_show_top_layers
+                onClicked: {
+                    UM.Preferences.setValue("view/only_show_top_layers", checked ? 1.0 : 0.0);
+                }
+                text: catalog.i18nc("@label", "Only Show Top Layers")
+                visible: UM.LayerView.compatibilityMode
+            }
+            CheckBox {
```
awhiemstra: Do we even still need this option? It was mostly a workaround for slow top layer generation.

fieldOfView: For this iteration I did not want to take away functionality.
I have some changes in mind for the draggable rangeslider branch.

plugins/LayerView/LayerView.qml
fieldOfView: Fixed

plugins/LayerView/LayerView.py
fieldOfView: Changed to use | as a separator

plugins/LayerView/LayerView.qml
Ghostkeeper: "Pythonic" means a combination of "elegant" and "using correct language syntax for the job". As far as I'm concerned these two attributes can be extended to other languages as well. If it's good for Python it should be good for Javascript, as long as this is also the most elegant syntax for the job.

awhiemstra: Well, no, to me it means "the way it is done in Python" which is not necessarily the best solution for the job (not even in Python).

plugins/LayerView/LayerView.py
jackha: Except you forgot this one ^^. No problem, already merged and fixed locally on my machine

jackha:
Merged into #1479, which will be merged into master in the near future.

PR #1474, opened 2017-02-24
Velleman Vertex K8400 single and dual head

tbackdahl:
Definition files for single and dual head versions of Velleman Vertex K8400 printer.
The machine uses printable coordinates x=0 to 200, y=20 to 200 and parking position 200, 200 regardless of how many extruders you use.
Thereby the different machine_width for the two models.
The g-code can probably be improved.
USB connection not successful this far.


Ghostkeeper:
These seem to be working fine. Thanks!

PR #1479, opened 2017-02-28
Layer view layout

jackha:
- A lot of hussling around in LayerView.qml: now it's Item -> Rectangle -> ColumnLayout + sliders
- Menu is now next to slider
- Menu incorporates legend as well. Legend consists of small squares which are turned on in color mode line_type.
- Double slider is yet to be replaced
- Compatibility mode also switches travels on top layer(s)


fieldOfView:
Perhaps you could incorporate the changes proposed in this PR: https://github.com/Ultimaker/Cura/pull/1446

jackha:
Yup looking at it. Unfortunately I've made some big layout changes in layerview.qml, which is now a pain in the ## to merge. btw. In the shader I currently cannot set opacity of extruders > 4

fieldOfView:
Stating the obvious: It would have been a pain in the ## if it were merged before....

Maybe the shader should be fixed too then?

Note that it is also a valid response not to merge my code, but then please close the PR, otherwise you lay the burden of updating my changes on me.

jackha:
I use a vec4 in the shader for the opacities, so it cannot be "fixed" easily. I'll leave that for now (I disabled all extruders > 4, they are always on).
I just merged the branch into my local branch and all seems to work. Yes, it should have been merged earlier, last activity was 8 days ago which is far too long. But, better late than never ^^

fieldOfView:
Awesome, thanks. I'll have a look next week to see if I can come up with a proper fix for the shader.

Note that you could limit the repeater I introduced to a maximum of 4 (which you may have already done).

jackha:
No problem. I show extruder 5 on RoVa3D, but the option is disabled. That way the user is informed that extruder 5 is present.

plugins/LayerView/LayerView.py
```diff
@@ -196,8 +194,9 @@ def getLayerViewType(self):
     #   \param extruder_nr 0..3
     #   \param opacity 0.0 .. 1.0
     def setExtruderOpacity(self, extruder_nr, opacity):
-        self._extruder_opacity[extruder_nr] = opacity
-        self.currentLayerNumChanged.emit()
+        if extruder_nr <= 3:
```
LipuFei: Perhaps `if 0 <= extruder_nr <= 3:`

plugins/LayerView/LayerView.py
```diff
@@ -383,12 +378,15 @@ def _updateWithPreferences(self):
...
-        self.setExtruderOpacity(2, float(Preferences.getInstance().getValue("layerview/extruder2_opacity")))
-        self.setExtruderOpacity(3, float(Preferences.getInstance().getValue("layerview/extruder3_opacity")))
+        for extruder_nr, extruder_opacity in enumerate(Preferences.getInstance().getValue("layerview/extruder_opacities").split("|")):
+            try:
+                opacity = float(extruder_opacity)
+            except ValueError:
+                opacity = 1.0
+            self.setExtruderOpacity(extruder_nr, opacity)
 
-        self.setShowTravelMoves(bool(Preferences.getInstance().getValue("layerview/show_travel_moves")))
```
LipuFei: `self.setShowTravelMoves(something)` does `self._show_travel_moves = something` already, so we can use the old statement here.

LipuFei:
- need to resolve conflicts
- some files have unnecessary permission changes: 0644 -> 0755 (a+x)
     - plugins/LayerView/LayerView.py
     - plugins/LayerView/LayerView.qml
     - plugins/LayerView/LayerViewProxy.py
     - resources/themes/cura/theme.json

PR #1481, opened 2017-03-01
Feature override profile

nallath:
This replaces the switch profile dialog with a more explict dialog.

CURA-3221

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -53,7 +53,7 @@ from . import MachineActionManager
 
 from cura.Settings.MachineManager import MachineManager
 from cura.Settings.ExtruderManager import ExtruderManager
-from cura.Settings.CuraContainerRegistry import CuraContainerRegistry
+from cura.Settings.UserChangesModel import UserChangesModel
 from cura.Settings.ExtrudersModel import ExtrudersModel
 from cura.Settings.ContainerSettingsModel import ContainerSettingsModel
 from cura.Settings.MaterialSettingsVisibilityHandler import MaterialSettingsVisibilityHandler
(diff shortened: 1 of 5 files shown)


cura/Settings/UserChangesModel.py
```diff
@@ -0,0 +1,113 @@
...
+                    category_label = self._i18n_catalog.i18nc(category.key + " label", category.label)
+                else:
+                    category_label = category.label
+
+                if self._i18n_catalog:
+                    label = self._i18n_catalog.i18nc(setting_key + " label", stack.getProperty(setting_key, "label"))
+                else:
+                    label = stack.getProperty(setting_key, "label")
+
+                for container in containers:
```
jackha: To me this loop and its contents look like it belongs to ContainerStack, except that UM does not know about resolve. It would not surprise me if this functionality is already duplicated somewhere else. What would be a nice solution for that?

cura/Settings/UserChangesModel.py
```diff
@@ -0,0 +1,113 @@
...
+        for file_name in definition.getInheritedFiles():
+            catalog = i18nCatalog(os.path.basename(file_name))
+            if catalog.hasTranslationLoaded():
+                self._i18n_catalog = catalog
+
+        for stack in stacks:
+            # Make a list of all containers in the stack.
+            containers = []
+            latest_stack = stack
+            while latest_stack:
```
jackha: So the next stack (if any) is usually the global stack?

cura/Settings/UserChangesModel.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, parent = None):
+        super().__init__(parent = parent)
+        self.addRoleName(self.KeyRole, "key")
+        self.addRoleName(self.LabelRole, "label")
+        self.addRoleName(self.ExtruderRole, "extruder")
+        self.addRoleName(self.OriginalValueRole, "original_value")
+        self.addRoleName(self.UserValueRole, "user_value")
+        self.addRoleName(self.CategoryRole, "category")
+
+        Application.getInstance().globalContainerStackChanged.connect(self._update)
```
jackha: Is this useful? 
- If the window's open, I cannot change machine
- If the window's closed, I'm not interested in the contents of the window

cura/CuraApplication.py
```diff
@@ -323,11 +323,26 @@ def _onEngineCreated(self):
...
         self._message_box_callback_arguments = callback_arguments
         self.showMessageBox.emit(title, text, informativeText, detailedText, buttons, icon)
 
+    showDiscardOrKeepProfileChanges = pyqtSignal()
+
+    def discardOrKeepProfileChanges(self):
+        self.showDiscardOrKeepProfileChanges.emit()
+
+    @pyqtSlot(str)
+    def discardOrKeepProfileChangesClosed(self, option):
```
jackha: Why creating this function and call it with "keep" ? I would remove the condition and remove the call with "keep". Then the function should of course be named appropriately

resources/qml/DiscardOrKeepProfileChangesDialog.qml
```diff
@@ -0,0 +1,157 @@
...
+        {
+            anchors.right: parent.right
+            anchors.rightMargin: UM.Theme.getSize("default_margin").width
+            spacing: UM.Theme.getSize("default_margin").width
+            Button
+            {
+                text: catalog.i18nc("@action:button", "Keep");
+                onClicked:
+                {
+                    Printer.discardOrKeepProfileChangesClosed("keep")
```
jackha: You can comment here that "keep" actually does nothing.

cura/CuraApplication.py
nallath: Let's discuss this in the architecture meeting.

cura/Settings/UserChangesModel.py
nallath: Let's discuss this in the architecture meeting.

PR #1484, opened 2017-03-03
Feature: CURA-3397 profile override dialog options

LipuFei:


cura/CuraApplication.py
```diff
@@ -334,7 +335,15 @@ def messageBox(self, title, text, informativeText = "", detailedText = "", butto
     showDiscardOrKeepProfileChanges = pyqtSignal()
 
     def discardOrKeepProfileChanges(self):
-        self.showDiscardOrKeepProfileChanges.emit()
+        choice = Preferences.getInstance().getValue("cura/choice_on_profile_override")
```
nallath: Could you make the "Always ask" a bit more explicit?

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -377,6 +379,40 @@ UM.PreferencesPage
...
+                height: childrenRect.height;
+
+                text: catalog.i18nc("@info:tooltip", ".")
+
+                ComboBox
+                {
+                    id: choiceOnProfileOverrideDropDownButton
+
+                    model: [
+                        "Always ask me this",
```
nallath: These strings are now never translated. You should use the catalog so that they can be internationalised.

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -377,6 +379,40 @@ UM.PreferencesPage
...
+                font.bold: true
+                text: catalog.i18nc("@label", "Override Profile")
+            }
+
+            UM.TooltipArea
+            {
+                width: childrenRect.width;
+                height: childrenRect.height;
+
+                text: catalog.i18nc("@info:tooltip", ".")
```
nallath: Shouldn't the tooltip actually give some helpfull information? If it only displays a dot, we might as well leave it out all together.

LipuFei:
@nallath Did a rebase to make the history clear.
- Added a comment in python code to make the if-else branches clear
- Added tooltip message with translation wrapper
- Added translation wrappers for drop down box options

PR #1472, opened 2017-02-23
[WIP] Use a single draggable range slider in Layerview

fieldOfView:
This PR combines the two sliders in the LayerView with a single rangeslider. The rangeslider has three draggable handles: lower, upper and range. By dragging the range it is possible to always display eg 5 layers.

See https://github.com/Ultimaker/Cura/issues/1444

TODO:
* fix keyboard control
* use PointingRectangle for the layernumber box
* cleanup

Issue #1444, opened 2017-02-18
Range Slider in LayerView


fieldOfView:
I looked into unifying the two sliders in the new layerview. First I checked out the QML RangeSlider, but that does not *quite* do what I want, and it required PyQt5.8 due to an unfortunate packaging error. Instead I created one myself:
https://gist.github.com/fieldOfView/b4565ab77dbff625408bd926760ee3bf

![draggablerangeslider](https://cloud.githubusercontent.com/assets/143551/23093871/81934c0e-f5ee-11e6-9f22-dc999a0122e1.gif)

@jackha, are you working on this or shall I continue and prepare a PR to use this in LayerView?

gabfv:
That's a great feature!

nallath:
I like it!

awhiemstra:
Nice work. As far as I know, @jackha is not currently working on any UI changes for the layer view, so feel free to make a PR.

fieldOfView:
Work in progress here:
https://github.com/Ultimaker/Cura/compare/master...fieldOfView:feature_draggable_range_slider

jackha:
Looks good! Currently I'm not working on any UI changes. Still fine-tuning the layer view itself and working on other issues.

Ghostkeeper:
Processing this and coordinating this with our UX guys and fieldOfView was just planned to be done in the sprint for the following two weeks.

fieldOfView:
I have an implementation that is mostly working, except for keyboard input. It still needs work, but that should be doable in two weeks.

fieldOfView:
Closing this issue in favor of the PR

PR #1495, opened 2017-03-06
Use a single draggable range slider in Layerview 

fieldOfView:
This PR combines the two sliders in the LayerView with a single rangeslider. The rangeslider has three draggable handles: lower, upper and range. By dragging the range it is possible to always display eg 5 layers.

See #1444

TODO:
* ~fix keyboard control~
* ~use PointingRectangle for the layernumber box~
* ~cleanup~

This is a followup to https://github.com/Ultimaker/Cura/pull/1472, which was against master

### DIFF
--- a/plugins/LayerView/LayerView.py
+++ b/plugins/LayerView/LayerView.py
@@ -285,13 +285,15 @@ class LayerView(View):
 
     def event(self, event):
         modifiers = QApplication.keyboardModifiers()
-        ctrl_is_active = modifiers == Qt.ControlModifier
+        ctrl_is_active = modifiers & Qt.ControlModifier
+        shift_is_active = modifiers & Qt.ShiftModifier
         if event.type == Event.KeyPressEvent and ctrl_is_active:
+            amount = 10 if shift_is_active else 1
             if event.key == KeyEvent.UpKey:
-                self.setLayer(self._current_layer_num + 1)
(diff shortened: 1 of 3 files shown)


fieldOfView:
Might I suggest a couple of changes to the legend?

* Remove the words "Show"; the checkboxes already suggest this (and you're not adding "Show" to the extruders) 
* Rename "Travels" to "Travel Moves"
* Change the order so the interesting things are on top (Shell, Infill, Support, Travel Moves on the bottom).

I could include these in the PR.

fieldOfView:
In order for the large-step layer navigation (ctrl+shift+ up/down) to work, this PR in Uranium needs to be merged:
https://github.com/Ultimaker/Uranium/pull/226

nallath:
Is the pull request ready to be reviewed / tested by us?

fieldOfView:
Yes

fieldOfView:
Aww, now layer_view3_layout is merged. Shall I resubmit the PR (again) against master?

jackha:
I'll look at the stuff and make sure it is in the master soon.

maukcc:
Also, is it possible to add the actual layerheight of the layer that is being viewed?
some thing like: 243@12.2mm
This is very helpfull if you want to add some code at a particular layer

PR #1504, opened 2017-03-07
CURA-3470 Return list instead of set for QVariantList

LipuFei:
Fix this crash:
```
Version: 2.4.99-master.20170307132702
Platform: Windows-10-10.0.14393
Qt: 5.7.1
PyQt: 5.7.1

Exception:
TypeError: unable to convert a Python 'set' object to a C++ 'QVariantList' instance
```

### DIFF
--- a/cura/MachineActionManager.py
+++ b/cura/MachineActionManager.py
@@ -105,7 +105,7 @@ class MachineActionManager(QObject):
         if definition_id in self._supported_actions:
             return list(self._supported_actions[definition_id])
         else:
-            return set()
+            return list()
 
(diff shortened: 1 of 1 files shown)


cura/MachineActionManager.py
```diff
@@ -105,7 +105,7 @@ def getSupportedActions(self, definition_id):
         if definition_id in self._supported_actions:
             return list(self._supported_actions[definition_id])
         else:
-            return set()
+            return list()
```
jackha: You can also add a return type to the function definition. That way, mypy can also find this error.

PR #1515, opened 2017-03-10
Simplified Chinese Translate

alphaarea:
Add Chinese Translate
But only translated partial word


Ghostkeeper:
We can't really merge this until there is a complete translation. A lot of these translate to empty, so the text would disappear.

LipuFei:
Hi @alphaarea, a few remarks:
- I suggest to use code `zh-CN` instead of `cn` (reference: [ISO Language Code Table](http://www.lingoes.net/en/translator/langcode.htm)).
- Please don't leave untranslated or blank texts.
- Could you also put the language into the Dropdown list in `resources/qml/Preferences/GeneralPage.qml`, so you can actually select it in the GUI?

ccjjmartin:
@LipuFei Should this PR be closed?  It looks like #2035 is going to be the PR that gets merged.

Ghostkeeper:
The other PR has now been merged, which makes this one obsolete. Thank you, but sorry!

Issue #1296, opened 2017-01-03
machine_start_gcode and dual exrtuder temperature config problem 


flyingKenny:
This problem occurs only with more than one extruder configuration.

I write a custom start code for my dual extrusion machine and want to set the 
_material_print_temperature_ or _material_standby_temperature_ to the first hotend
like:
"machine_start_gcode": { "default_value": ";
-   M140 S{material_bed_temperature}
-   M104 S{material_standby_temperature}; preheat
-   M190 S{material_bed_temperature}; 
-   G29; Probe 
-   M109 S{material_print_temperature}; Final heat

in the gcode: the hotend temperatures are from the definition file or default, not from the materials or gui config.

Is it possible to differ whitch extruders are used in the job and preheat only this.
eg. with a machine_extruder_init_code
in cura 14.12 it was differd by start.gcode and start2.gcode

Ghostkeeper:
There is `machine_extruder_start_code`, which is the g-code to use when switching to a certain extruder. But that would get executed at every extruder switch, not just at the start of the print. It's good enough to do things like setting the bed temperature but not priming since that only needs to be done once.

This sort of thing is really not supported yet at the moment by the engine. We haven't run into this problem ourselves since for the Ultimaker 3 (our only multi-extrusion printer) the default procedure is enough.

Ghostkeeper:
> The flaw I see here:
> All extruders are heated to print temp and then are drawing power for no reason to the unused extruders.

Cura should not attempt to heat up extruders that it doesn't use. We've made specific checks for that. Because it's hella annoying to test a print when it needs to heat up both extruders every time ;)

maukcc:
strange, when UM3 is selected it only heats up the extruder it is going to use
with my Cartesio it does:
;FLAVOR:RepRap
;TIME:285
;Filament used: 0.0913444m
;Layer height: 0.2
;Generated with Cura_SteamEngine master
M190 S50
M104 S205
M104 T1 S235
M104 T2 S245
M104 T3 S250
M109 S205
M109 T1 S235
M109 T2 S245
M109 T3 S250

;LAYER_COUNT:20
;LAYER:0

So it heats up all extruders to " printing temperature initial layer" 
Maybe I need to set variants?

Drayson80:
@Ghostkeeper, au contraire mon amis...
You mention that the UM3 is the only dual printer, but this is not correct. There are a few UMO with official dual upgrade in the feld, so sorry to correct this :-)
[...]

Issue #1279, opened 2016-12-29
extruder offset changeble in GUI


maukcc:
Every time you physically change your extruder you will have to set the offset for the 2nd one.
At present you will have to fiddle with the .def.json file.
Can this be done in the GUI?

fieldOfView:
It's on my personal todo list for 2.5

maukcc:
Also: this needs to be set per printer.
So if I have 5 exactly the same printers, the offsets will probably differ per printer

maukcc:
@fieldOfView Any progress on this? source files to test?
it is driving me insane to manually change this every time :)

fieldOfView:
I have no progress on this to report.

fieldOfView:
Progress is here: #1521

maukcc:
this can be closed

PR #1521, opened 2017-03-11
Multiextruder Machine Settings / Custom Multiextruder printer

fieldOfView:
Machine Settings currently does not support multi-extruder printers. It is also not possible to set the number of extruders.

The main difficulty of setting the number of extruders is that each extruder has its own definition, and these definitions need to have a back-reference to the main definition.

~~This PR is a proof of concept where it is possible to change the number of extruders limited to the number of extruders that have their own definition. So basically it lets the user turn off extruders (eg make a single extruder machine out of a UM3). While this may seem very limiting, the PR also defines 8 extruders for the Custom FDM Printer definition, so it is possible to create a custom printer with up to 8 extruders (though it still defaults to a single extruder machine).~~

~~Adding settings for the extruders (most notably extruder offset and nozzle switch gcode) is next up for this PR.~~

**Update:**
What this PR adds:
* A dropdown field for the number of extruders for the printer, limited to the number of extruders defined in the machine definition
* Up to 8 extruders for the Custom FDM Printer
* Tabs per extruder with settings for nozzle size, nozzle offsets and extruder start & end gcode for any printer with > 1 extruder

What this PR does not add:
* The ability to add extruders to definitions that do not add extruders (eg it is not possible with this PR to change the Ultimaker 2 definition into a dual extrusion machine)
* Nozzle size fields if the definition has variants (because the variant would override the definition_changes anyway)


fieldOfView:
Any hints on why #042f45f cannot be built?

awhiemstra:
Seems it cannot be built because it does not exist. O_o

I do not really know what's going on here. I guess Jenkins initial tried building it but now thinks it no longer exists?

maukcc:
Will this only affect the custom printer, or can we implement this in other "definition".def.json files?
I would love it, if extruders in the side bar are greyed out to show that they are not being used.
And if they are not being used, they are not being heated up, like we are discussing in #1296 (regardless of what material was selected for it)

Ghostkeeper:
The menu to adjust machine settings can be added to any printer definition file. It'll show the menu after the user chooses to add the printer and it'll show a button in the machine manager to restore the window later.

Just add this code to the definition metadata: https://github.com/Ultimaker/Cura/blob/master/resources/definitions/custom.def.json#L13

This pull request would make it possible to adjust extruder settings in multiple extruders with that window as well.

fieldOfView:
> Just add this code to the definition metadata: https://github.com/Ultimaker/Cura/blob/master/resources/definitions/custom.def.json#L13

Only do that if you want the machine settings to pop up directly when adding the printer. For most predefined printers this will be unnecessary.

The plugin adds itself as supported action to all printers, except for printers with multiextrusion because they are not currently handled well. After this PR (which is a work in progress but should be ready before 2.6 goes into feature freeze), the Machine Settings action will be supported by any printer.

fieldOfView:
> I would love it, if extruders in the side bar are greyed out to show that they are not being used.

Please log a separate issue for that, because that is well outside the scope of this PR.

Drayson80:
Hmm.. how could I test/try this? Simply replace the files in my Cura plugin folder or do I have to do something else? (Pls. remind, I´m no software guy, just a mechanical one :-) )

fieldOfView:
> (Pls. remind, I´m no software guy, just a mechanical one :-) )

Then I would say: don't bother yet. It is still very much a work in progress and not functional yet.

Drayson80:
ok, thanks for the info. 
just so annoyed that I always need to switch printers when jumping between single and dual setup...

fieldOfView:
Even with this PR, it is still going to be quicker to switch to another printer using the dropdown menu than it will be to go to the Manage Printers screen, click Machine Settings and then change the number of extruders.

maukcc:
I see extruder offset is also being addressed in this?
If so, then #1279 can be closed.

fieldOfView:
Yes, the x & y offset will be settable per extruder

fieldOfView:
I have tentatively removed the [WIP] from the title of the PR. This can do with an initial round of review & testing

fieldOfView:
Is Jenkins misbehaving again, or have I screwed up?

awhiemstra:
Actually, it seems the network is misbehaving because Jenkins fails during a git checkout... :S

Ghostkeeper:
See CURA-3299.

nallath:
I'm kind of surprised about how little python code it requires for this to work.

fieldOfView:
Testament of a well-thought-out settings system?

I did try a couple more Python-heavy approaches before ending up with this one.

nallath:
So far I've managed to find a minor issue; Updating the offset does not trigger an update of the disallowed areas.

nallath:
Another thing is that the project reader doesn't like definition_changes, but that is an issue of the project reader.

fieldOfView:
I have changed the location where definition_changes containers are stored, because storing them in the folder with the stacks was ill-advised; the upgrade-system was trying to upgrade the definition-changes instance containers to container stacks.

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -671,7 +671,12 @@ Cura.MachineAction
                         propertyProvider.setPropertyValue("value", text);
                         if(forceUpdateOnChange)
                         {
+                            var extruderIndex = ExtruderManager.activeExtruderIndex;
                             manager.forceUpdate();
+                            if(ExtruderManager.activeExtruderIndex != extruderIndex)
+                            {
+                                ExtruderManager.setActiveExtruderIndex(extruderIndex)
+                            }
```
Ghostkeeper: Hmm, the `forceUpdate` emits a `globalContainerStackChanged` signal. Does that change the active extruder index?

jackha:
@fieldOfView: Wow this looks really amazing! I'm currently testing the PR and I wonder if you have the problem as well: it doesn't slice with this machine. Not even with 1 extruder. The progress bar is stuck at 0. Maybe some settings are missing or misplaced?

Error in CuraEngine (with debug mode): /Projects/CuraEngine/src/MeshGroup.cpp:77: cura::ExtruderTrain* cura::MeshGroup::getExtruderTrain(unsigned int): Assertion `extruders[extruder_nr]' failed.
2017-04-18 14:18:28,403 - DEBUG - CuraEngineBackend.CuraEngineBackend._onBackendQuit [545]: Backend quit with return code -6. Resetting process and socket.

cura/CuraApplication.py
```diff
@@ -145,13 +146,15 @@ def __init__(self):
         Resources.addStorageType(self.ResourceTypes.UserInstanceContainer, "user")
         Resources.addStorageType(self.ResourceTypes.ExtruderStack, "extruders")
         Resources.addStorageType(self.ResourceTypes.MachineStack, "machine_instances")
+        Resources.addStorageType(self.ResourceTypes.DefinitionChangesContainer, "machine_settings")
```
awhiemstra: Just a minor comment, but I would prefer we use "definition_changes" for this. We should match the storage name to our internal naming as much as possible in my opinion.

plugins/MachineSettingsAction/MachineSettingsAction.qml
fieldOfView: Yes, globalContainerStackChanged seems to always set the active extruder index to 0, hence this fix.

cura/CuraApplication.py
fieldOfView: Ok, point taken. I wanted to keep the folder close to "machine_instance".

fieldOfView:
@jackha, hmm, I must admit that I had not tried slicing, mostly because slicing-wise I had already tested the base technique before. I will see what is going on.

jackha:
@fieldOfView Thanks!

BagelOrb:
It shouldn't matter on which it relies. The machine_extruder_count should
be exactly the same as the amount (count) of extruders. Isn't that obvious?

Op 18 apr. 2017 17:13 schreef "Aldo Hoeben" <notifications@github.com>:

It seems that somewhere CuraEngine is relying on the number of defined
extruders (metadata.machine_extruder_trains.length) instead of
machine_extruder_count. I'm still trying to find out where by staring at
the code. @BagelOrb <https://github.com/BagelOrb>, you can probably find
this faster than I can.

—
You are receiving this because you were mentioned.
<https://github.com/Ultimaker/Cura/pull/1521#issuecomment-294876777>, or mute
the thread

maukcc:
I am seeing 2 different things:
- the max amount of extruders possible on the machine. (in the definition .def.json file made by the vendor)
- the amount of extruders actually present on the machine (selected in the machine settings by the user)

Ghostkeeper:
The problem with duplicate data is that the data might become inconsistent and this could lead to unexpected results. This is a textbook example! The fix is to depend on only one of the duplicates. In this case, I'd say depend on the extruder list.

There's quite a lot of things in the front-end that also depend on machine_extruder_count.

BagelOrb:
You misunderstood. The machine setting should not be settable by the user.
The format doesn't allow you to make a distinction between possible
extruders and actual extruders.

You'll just manually have to make sure you only use the extruders currently
present on your machine.

Note that what is an extruder in cura actually governs a whole extruder
train: the extruder, the filament, the nozzle etc.

Could you send some pictures of your machine?! I have no idea how there are
multiple extruders defined but not present.

Op 19 apr. 2017 09:43 schreef "MaukCC" <notifications@github.com>:

I am seeing 2 different things:

   - the max amount of extruders possible on the machine. (in the
   definition .def.json file made by the vendor)
   - the amount of extruders actually present on the machine (selected in
   the machine settings by the user)

—
You are receiving this because you were mentioned.
<https://github.com/Ultimaker/Cura/pull/1521#issuecomment-295143582>, or mute
the thread

BagelOrb:
Is hard to find out the extruder count by looking at the number of
extruders defined because of inheritance.

A defe.json file can override a single extruder and leave the rest intact.

Op 19 apr. 2017 09:52 schreef "Tim Kuipers" <t.kuipers@ultimaker.com>:

fieldOfView:
This PR tries to make it possible to define a custom printer with a settable number of extruders. Because of the back- and forward-references between machine definitions and extruder definitions, this is only possible by referencing a maximum number of extruders in the custom printer definition, and then setting the number of extruders to use in the machine instance.

I tried this method by creating a single extruder version of the UM3 using inheritance (which seems to work): https://gist.github.com/fieldOfView/bde7189087721451784865a28f3fc2c1
Unfortunately doing the same trick with an instance_container turns out not to work (yet).

maukcc:
> this is only possible by referencing a maximum number of extruders in the custom printer definition, and then setting the number of extruders to use in the machine instance.

exactly what I mean.
This way I only have to make 1 cartesio.def.json file and then, on 1st start up, the user says how many tools are mounted to it. 
Other wise I would have to make 12 cartesio.def.json files (and its accompanying extruders, variants, materials, quality and so on), which will also probably slow down the startup time of Cura.

So at present when you select Cartesio printer at 1st start up then you get 4 extruders. Which works(ish) for the moment, even if there are only 2 extruders present.

fieldOfView:
Is there an easy way to get a stacktrace out of CuraEngine?

nallath:
The only way i know is with gdb.

fieldOfView:
@jackha: this CuraEngine patch fixes the crash for me: https://github.com/Ultimaker/CuraEngine/pull/501
Would be great if you could test this too.

fieldOfView:
I think there is still some work to do when the user sets the number of extruders to eg 4, then sets the support extruder to the 4th extruder, and then sets the number of extruders back to 2. Same with the other settings where you set an extruder_nr.

Ghostkeeper:
I don't know if it's related to this change, but when I unfocus one of the machine head polygon text fields, I'm getting the log message:

    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:644: ReferenceError: printheadXMinField is not defined

It seems that printheadXMinField is out of scope for the `setHeadPolygon()` function, even though that very text field triggers the call to `setHeadPolygon()`. This error then prevents the head polygon from being updated and makes these text fields ineffective.

Ghostkeeper:
When I switch to one of the extruder tabs I'm getting these error messages:

    file:///usr/local/lib/python3.5/dist-packages/PyQt5/Qt/qml/QtQuick/Controls/ScrollView.qml:348:13: QML Item: Possible anchor loop detected on fill.
    file:///usr/local/lib/python3.5/dist-packages/PyQt5/Qt/qml/QtQuick/Controls/ScrollView.qml:348:13: QML Item: Possible anchor loop detected on fill.
    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:665:23: Unable to assign [undefined] to QString
    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:665:23: Unable to assign [undefined] to QString
    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:665:23: Unable to assign [undefined] to QString
    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:585:43: Unable to assign [undefined] to QString
    file:///home/ruben/Projects/Cura/plugins/MachineSettingsAction/MachineSettingsAction.qml:609:43: Unable to assign [undefined] to QString

Ghostkeeper:
And indeed with the other patch, it slices fine. At least, I did a test with an F model that had support on the second extruder and that worked.

When I set the support pattern to be triangles though, nothing happens. It stays on zigzag. This indicates to me that there is some internal inconsistency still.

When that has been fixed and the error when switching extruder tabs in the machine action window is fixed too, I deem this pull request ready to merge (along with the accompanying pull request in the engine).

fieldOfView:
Thanks for the review. Will look in to the issues you found.

fieldOfView:
> When I set the support pattern to be triangles though, nothing happens. It stays on zigzag. 

I think this is caused by a setting value on the relevant extruder that is being overridden by a value that is still set on the global stack. I will see if I can fix this by moving non-global-only settings to the first extruder stack when switching to multiple extruders (and moving the same settings back to the global stack when switching back to single extruders).

fieldOfView:
Ready for another round of review/tests. I fixed the issues found in the previous round, and fixed a number of issues when switching the number of extruders. The definition_changes containers are now also properly disposed of.

fieldOfView:
> All checks have failed

Not sure if me or Jenkins

nallath:
[WS-CLEANUP] Deleting project workspace...Cannot delete workspace :java.io.EOFException 

Jenkins...

jackha:
@fieldOfView: The support pattern in multi extruder machine is still a problem. Still stays on zig-zag. We think that it's important to fix this, as it may affect other settings as well.

btw. The CuraEngine PR is merged.

fieldOfView:
Could you add reproduction steps? I could not reproduce it. Preferably starting from a clean printer.

jackha:
I used the feature_multiextruder_machinesettings with current master merged into it + master CuraEngine + master Uranium. Then, when I start clean, add fdm printer with 2 extruders it doesn't use the Support Pattern.

fieldOfView:
Ah, I see now. The support pattern only works if the number of extruders is 1 or 8, but not anywhere inbetween. Weird. I think I'll have to dive into CuraEngine again.

fieldOfView:
My CuraEngine fix was the culprit. It could ignore all extruder settings. 
See https://github.com/Ultimaker/CuraEngine/pull/509 for the fix.

jackha:
Ok merged it into master after fixing a small problem. 

Strange that github says it has conflicts... oh well it's merged anyway.

jackha:
btw Did I say Thanks for the big effort! ?!

maukcc:
I am loving it. 
I am having a little mishap: 
I have 4 extruder trains in the definition file, but I choose to use only 2 in my machine settings.
Now Cura lets me see the 2 extruders, but takes the other 2 extruders(which are not there) into account.
For instance I have quality profiles where it is not allowed to make layerheights heigher then 0.3mm if one of the nozzles is smaller then 0.5mm.
Now if one of the non visible extruder trains is set to 0.4, it will not show these quality profiles.

fieldOfView:
> For instance I have quality profiles where it is not allowed to make layerheights higher then 0.3mm if one of the nozzles is smaller then 0.5mm.

Silly question: how do you do that?
In other words: how can I reproduce this?

maukcc:
You can select the Cartesio as machine and try it.

In the definition file we set:
"layer_height": { "maximum_value": "(0.8 * min(extruderValues('machine_nozzle_size')))" },
and then it will not show the quality profiles with a layerheight above that min value.

And that is also the culprit as that line checks all extruderValues/trains and not just the ones actually in use.

maukcc:
No that was not it.
It is the ""has_machine_quality": true," line in the definition file that does the trick.

fieldOfView:
I am working on a patch. Are you running Cura "from source"? If so you could help test the patch.

maukcc:
No sorry, I am running thopiekar`s master version.

fieldOfView:
@maukcc, see #1809

maukcc:
With today’s version of thopiekar`s master version I can:

change extruder amount, but it will not update the amount
So in settings I can change the amount of extruders, ig from 4 to 2, but it will keep showing the 4 extruder tabs. Until I change the machine and then change it back again, then the value has been activated and it only shows the 2 tabs.(or visa versa)

values that are set for unselected extruders are impacting
So if I set extruder_3 to 0.4 nozzle, and I say that the machine has only 2 extruders, It will still read the value of extruder_3, and I can not use larger layer heights.

fieldOfView:
Please start a new issue. This was working, and now it is broken.

PR #1536, opened 2017-03-14
Cartesio changes and 3 more Quality definitions

maukcc:


Ghostkeeper:
We've been looking for a while for a better name for our new 0.8mm nozzle profiles for the Ultimaker 3. It was called Verydraft and Superdraft to force our UX designers to come up with a better name, but they haven't yet. I rather like these new names of "Coarse" and "Extra Coarse".

Ghostkeeper:
All right, I've discussed this with some colleagues and here's what we're going to do. I'm planning to discuss these names (Coarse and Extra Coarse) with UX and see if we can "commandeer" them for our Ultimaker 3 profile names. If they agree then we'll name the 0.3mm profiles "Coarse" and the 0.4mm profiles "Extra Coarse", and we'll come up with a new name for your 0.6mm layers. Probably something in the spirit of "Super Coarse".

You can change the name of that 0.6mm profile later, but we'd like all profiles with the same layer height to have the same name, because in multi-extrusion they can be used together only if they have the same quality type (and it's logical to name the profile after the quality type).

Would that be all right with you?

Also, this pull request has some conflicts. I'd like you to resolve those if possible. Otherwise I'm going to assume that this pull request is leading and should override all other changes made to the Cartesio profiles since your pull request was branched.

maukcc:
We have been printing with larger nozzle sizes for a long while and never used a 0.3mm layerheight. But if you guys feel there is a need for it, then lets put it in. The same for the 0.15 and 0.06 They are just making it harder for the user to choose from the long list, and results stay pretty much the same as 0.1 or 0.2 . If a user wants to go in between of the 0.x layers then we can assume they are proficient enough to alter the layerheight themselves. 
We feel it is important that new users have a simple selection of high, normal, coarse and extra course (0.1, 0.2, 0.4, or 0.6) to choose from and not clutter up the selection with all layer heights in between.

Was there a reason to have 0.06 and not 0.05 ?

In our latest definition file we also related the nozzle size to the max layerheight. Maybe there is a more elegant way of solving this?

I do not really know what the conflicts are, so let the latest version rule.

Ghostkeeper:
I've discussed this with our UX designers and they said that we don't want to use Coarse and Extra Coarse to name the new 0.8mm profiles. I see no reason then to make sure the naming is consistent, so we'll just get this merged as-is. No commandeering.

> We feel it is important that new users have a simple selection of high, normal, coarse and extra course (0.1, 0.2, 0.4, or 0.6) to choose from and not clutter up the selection with all layer heights in between.

The fact that we provide 0.15mm and 0.3mm profiles shouldn't make your list of profiles longer. There's no need to make every printer have every layer height in its list of profiles. It shouldn't clutter up Cartesio users' list of profiles :)

> Was there a reason to have 0.06 and not 0.05 ?

I guess that 0.06 was chosen over 0.05 for quality reasons (0.05mm is just not feasible with the hardware). And the 0.06mm layer height aligns better with the steps that the z-stage of the UM2 printer can make. The z-stage of the Ultimaker 2 can step with 0.02mm accuracy, so I guess it makes sense to use a multiple of that. That's all just guesswork from me though. It was determined while I was just an intern here working on a different project.

> In our latest definition file we also related the nozzle size to the max layerheight. Maybe there is a more elegant way of solving this?

In our own profiles we give a warning for the layer height when it's larger than 75% of the nozzle size. I think a warning would be better than actually setting the maximum. With a warning the setting is coloured yellow. We intend that to mean "Well you could try printing with this but it's probably not going to work." So the user is allowed to try printing with a layer height of 30mm, but we think it's not going to work so well. Instead of setting the maximum value (which absolutely forbids that the user sets this value) we retain power to the user to try weird stuff if he so chooses.

maukcc:
>  It shouldn't clutter up Cartesio users' list of profiles 

How should we go about this?

>  Instead of setting the maximum value (which absolutely forbids that the user sets this value) we retain power to the user to try weird stuff if he so chooses

The thing is that we want a new users to use the "recommended" print setup, and not scare them away with the 101 options in "custom" setup. That is for us to worry about :)  
When using recommended setup, the user will not see any yellow warnings and the program will just slice, giving horrible results. So if the program would give a popup stating there are warnings and the user can proceed at own risk, then I agree.

Ghostkeeper:
> How should we go about this?

As long as you don't provide 0.3mm profiles, Cura shouldn't display any.

> When using recommended setup, the user will not see any yellow warnings and the program will just slice, giving horrible results. So if the program would give a popup stating there are warnings and the user can proceed at own risk, then I agree.

The profile would give a correct layer height (that wouldn't be a warning setting and wouldn't give horrible results). In the recommended set-up the user is not able to set any different layer heights than the ones in your profiles. The user can only adjust the layer height in the "custom" set-up. When he changes it there to a layer height that is too high, it gives a warning colour.

maukcc:
I am missing something :)

The way I go about it is 
1 I select the printer
2 I select a nozzle size
3 I select a material
4 I select the quality profile
5 print

At point 2 you can choose a 0.25mm nozzle and at point 4 a 0.6mm layerheight. 
That will give no warnings.
Is there a way, that when you choose a 0.25 nozzle, all quality profiles over 0.2mm disappear?

Ghostkeeper:
Ah, wait, I've been making a few wrong assumptions. Cartesio doesn't have machine-specific quality profiles. When there are no quality profiles specific to the machine, it'll fall back to the default quality profiles, which are the `normal.inst.cfg`, `draft.inst.cfg`, etc. profiles you were seeing. It'll use all available default profiles then, which explains why you couldn't see how to remove a profile.

There's a metadata property, `has_machine_quality` that when set will make Cura look for quality profiles for the machine. You can then make quality profiles specific to 0.25mm nozzles and PLA materials, and it'll only display those.

    "has_machine_quality": true,

However, this means that you'll need a quality profile for every nozzle-material-layerheight combination you want to support. And that could be a lot of profiles! Ultimaker 2+ and 3 have this set-up right now. You can't select High Quality for TPU on an UM3 because we don't support that, and you can't select High, Normal or Fast for the 0.8mm nozzles, because the layer height is too small for those nozzles.

maukcc:
I will have a look into what we want to do with this. For now we will leave it as is.

The thing is that we like to be as generic as possible. and adding these machine specific quality files do not add anything, except more files. (And I hope these are not all being read at program start-up). Most of the settings are already covered in variants and materials, and adding them in quality just adds to the confusion :)

So for now let it be as-is, and then see how we will deal with it in the future.
Is there a release date for 2.5?

awhiemstra:
> Is there a release date for 2.5?

Beta should go live this afternoon.

maukcc:
AH, just in time then :)

maukcc:
Can this be merged as is?
Or do you need us to change something?

PR #1538, opened 2017-03-14
Mould

Ghostkeeper:
Adds a setting that turns a model into a mould for the model.


BagelOrb:
Derp yes.
Now max val is 90

nickthetait:
Guys I have to admit this is pretty neato :+1:

PR #1539, opened 2017-03-14
Update Renkforce RF100 profile

probonopd:
Update as per discussion in https://github.com/Ultimaker/Cura/pull/1350


Ghostkeeper:
Don't mind the failed checks. We've just introduced this CI system and we're still working out the kinks.

PR #1545, opened 2017-03-15
CURA-3500 Make sure that material XMLs have utf-8 encoding

LipuFei:


fieldOfView:
CURA-3500 is not a release issue?

nallath:
Nope :(

PR #1554, opened 2017-03-20
CURA-3495 Open project/models menu

LipuFei:
Add a dialog asking the user when he/she is opening a Cura project file. He/she can choose to:
1. open as a Cura project, or
2. import models from this project file.
It is possible to set a default option for this. This is configurable in the settings dialog.

Some screenshots:
image
image

PR #1556, opened 2017-03-21
Add more spacing between the label and combobox for override profile option

LipuFei:
In the same way as https://github.com/Ultimaker/Cura/pull/1554


LipuFei:
@jackha reduced the width of the combox too. This one is ready.

PR #1564, opened 2017-03-23
Usb fixes

nallath:
Certain machines don't like sending empty lines. This resolves that.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -560,10 +560,9 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
             line = line[:line.find(";")]
         line = line.strip()
 
-        # Don't send empty lines
+        # Don't send empty lines. But we do have to send something, so send m105 instead.
         if line == "":
-            self._gcode_position += 1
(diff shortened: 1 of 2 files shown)


nallath:
Jira issue -> CURA-3583

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -560,10 +560,9 @@ def _sendNextGcodeLine(self):
             line = line[:line.find(";")]
         line = line.strip()
 
-        # Don't send empty lines
+        # Don't send empty lines. But we do have to send something, so send m105 instead.
         if line == "":
-            self._gcode_position += 1
-            return
+            line = "M105"
```
fieldOfView: Instead of sending an M105, which also causes a response to be sent, couldn't you send an empty comment?
line = ";"

nallath: Smoothieboard don't like the empty lines.

fieldOfView: Ah, I would have run into this issue soon then, because I have just received a smoothieboard clone for a new "project".

PR #1578, opened 2017-03-25
Fail gracefully when libSavitar is not found

fieldOfView:
By catching the ImportError, this prevents the logs being clogged with exceptions in start and when the plugins preference page is touched.

This PR is useful to keep the barrier of entry for contributing to the Cura project low. I have still not managed to compile lib savitar, and the logs do quickly become clogged up with the exceptions caused by not having a working libsavitar.

Note: this must be tested, because I can only test the case where there is no libSavitar, not the case when there is a libSavitar.

### DIFF
--- a/plugins/3MFReader/__init__.py
+++ b/plugins/3MFReader/__init__.py
@@ -31,7 +31,7 @@ def getMetaData() -> Dict:
             "api": 3
         }
     }
-    if "ThreeMFReader" in sys.modules:
+    if "3MFReader.ThreeMFReader" in sys.modules:
         metaData["mesh_reader"] = [
             {
                 "extension": "3mf",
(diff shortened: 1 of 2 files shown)


nallath:
CURA-3595

nallath:
Hmm. This is really odd. Before this pull request the plugins work. After this it complains that it can't find Savitar and won't add the plugins.

fieldOfView:
Ow :/. Like I noted, I have not tested the "happy case". Will see if I can.

fieldOfView:
That should fix it... I was checking for the wrong module (ThreeMFReader vs 3MFReader.ThreeMFReader).

I tested this by patching the binary release now, and both scenarios work for me (with/without Savitar). Without Savitar project loading/saving fails ofcourse.

LipuFei:
I tested this on my Windows with/without Savitar and it works fine.

LipuFei:
@fieldOfView @nallath Is there anything else needs to be done for this?

fieldOfView:
It needs to be merged ;-)

Issue #1582, opened 2017-03-25
can't change filiment dia


thomllama:
got fatal error as soon as I tried to change filament dia to 1.75

Version: 2.5.0-BETA
Platform: Darwin-16.4.0-x86_64-i386-64bit
Qt: 5.7.0
PyQt: 5.7.1

Exception:
Traceback (most recent call last):
  File "/Applications/Cura 2.5.app/Contents/MacOS/plugins/plugins/CuraEngineBackend/CuraEngineBackend.py", line 190, in slice
    self.printDurationMessage.emit(0, [0])
  File "/Users/ultimaker/build/2.5/build/inst/lib/python3.5/site-packages/UM/Signal.py", line 201, in emit
  File "/Users/ultimaker/build/2.5/build/inst/lib/python3.5/site-packages/UM/Signal.py", line 305, in __performEmit
  File "/Users/ultimaker/build/2.5/build/inst/lib/python3.5/site-packages/cura/PrintInformation.py", line 122, in _onPrintDurationMessage
  File "/Users/ultimaker/build/2.5/build/inst/lib/python3.5/site-packages/cura/PrintInformation.py", line 162, in _calculateInformation
ZeroDivisionError: float division by zero

smartavionics:
Deleting all the digits probably sets the value to 0 so as a workaround, pause slicing before changing the dia and un-pause after you have set 1.75.

thomllama:
wouldn't let me even start it up.  crashed 2.4 also. had to delete and use a cleaner to remove all files then reinstall. Installed 2.5 all alone and it seems to be working fine now. 

Do wish they would start having seperate prefs/settings folders especially for beta versions so issues wouldn't cross bleed to the stable versions.

fieldOfView:
> Do wish they would start having seperate prefs/settings folders especially for beta versions so issues wouldn't cross bleed to the stable versions.

That is in the works for Cura 2.6

thomllama:
SWEEETTTT!!!  :P

thomllama:
as a note, since deleting everything and installing 2.5 alone, seems to be working very well. 
Thanks for all the hard work folks!!

fieldOfView:
I have also made a fix for your issue, so it should be fixed before 2.5 goes out of beta. Thanks for the report.
https://github.com/Ultimaker/Cura/issues/1586

thomllama:
Cool. :)

PR #1586, opened 2017-03-26
Fix crash when editing material diameter

fieldOfView:
While editing the diameter value in the materials pane, it can happen that the radius evaluates to 0. This led to a division by zero.

Fixes https://github.com/Ultimaker/Cura/issues/1582

This needs to be cherrypicked to the 2.5 branch

### DIFF
--- a/cura/PrintInformation.py
+++ b/cura/PrintInformation.py
@@ -126,6 +126,7 @@ class PrintInformation(QObject):
             return
 
         # Material amount is sent as an amount of mm^3, so calculate length from that
+        radius = Application.getInstance().getGlobalContainerStack().getProperty("material_diameter", "value") / 2
         self._material_lengths = []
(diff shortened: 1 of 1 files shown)


cura/PrintInformation.py
```diff
@@ -161,8 +160,13 @@ def _calculateInformation(self):
                     else:
                         cost = 0
 
+            radius = Application.getInstance().getGlobalContainerStack().getProperty("material_diameter", "value") / 2
```
LipuFei: Is it possible to get the radius just once outside the `for`-loop like the old code? Or it will cause problems?

fieldOfView: ah, THAT's why I put it all the way up there before. Will fix, thanks

jackha:
Thanks!

PR #1604, opened 2017-03-29
retraction_prime_speed  as a material specific value

maukcc:
added retraction_prime_speed  as a material specific value.
This way, if a material needs a specific setting, the value set in the material xml file will be used. 
Probably a good idea to have this value added in the front end as well, but for now this will do.


Ghostkeeper:
We can't do this. The XML material profiles are not just under control of Cura. They are used by other applications as well, such as the Ultimaker 3 firmware. We can't just put unexpected fields in those XML files.

There's plans to make it extensible with a Cura namespace so that we can put other settings in there.

maukcc:
The thing is that these are really material specific settings. Think of flexible materials, then you also need to set:
- print speed
- prime speed
- prime extra amount

So please put some pressure on those plans :)

Ghostkeeper:
I'll try, but the consensus at the moment is that we're doing one refactor at a time, and the refactor going on at the moment is to deal better with resolve functions.

PR #1615, opened 2017-03-30
Adding sign to pattern match for negative Z value

tamarintech:
Cura will throw an error when it encounters a negative Z value (cannot convert string to float).
Negative Z values are used in many start up routines, nozzle cleaning, etc, etc, etc.

This simply adds a sign to the pattern match.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -566,7 +566,7 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
             if line == "M0" or line == "M1":
                 line = "M105"  # Don't send the M0 or M1 to the machine, as M0 and M1 are handled as an LCD menu pause.
             if ("G0" in line or "G1" in line) and "Z" in line:
-                z = float(re.search("Z([0-9\.]*)", line).group(1))
+                z = float(re.search("Z([-0-9\.]*)", line).group(1))
                 if self._current_z != z:
                     self._current_z = z
         except Exception as e:


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -566,7 +566,7 @@ def _sendNextGcodeLine(self):
             if line == "M0" or line == "M1":
                 line = "M105"  # Don't send the M0 or M1 to the machine, as M0 and M1 are handled as an LCD menu pause.
             if ("G0" in line or "G1" in line) and "Z" in line:
-                z = float(re.search("Z([0-9\.]*)", line).group(1))
+                z = float(re.search("Z([-0-9\.]*)", line).group(1))
```
fieldOfView: ```"Z(-?[0-9\.]*)"``` would be slightly cleaner. Then again perhaps we should be more anal and go for ```Z(-?\d*\.\d+|\d+)```

fieldOfView:
Nice catch!

nallath:
For our reference; CURA-3624

plugins/USBPrinting/USBPrinterOutputDevice.py
Ghostkeeper: If the engine produces invalid g-code such as `G1 Z-0.6-9.2`or something, the proposed solution would give a ValueError that it can't convert this gibberish to a float. The more strict regex that you propose would give a bounds error because group 1 doesn't exist. I think the former is more appropriate, though it might not be as elegant.

tamarintech: For Cura, the validation of the gcode is done by the CuraEngine already, right?
I'm a fan of _as simple as possible regex_ for extracting values from a known upstream and relying on the float to throw ValueErrors for any nonsensical matches.

Maybe we could make the logging for the exception more descriptive so that it is specifically known this is a G1 Z<value> conversion error?

The M0, etc, stuff seems out of place in that block. I can do a separate PR refactoring that exception block and logging if that sounds like something that is wanted...?

nallath: Better error handling is always something that is wanted :)

tamarintech: See https://github.com/Ultimaker/Cura/pull/1670
Could probably make that much better... but that's what I was thinking. :)

ChrisTerBeke:
This was already merged and fixed (according to Jira)?

tamarintech:
Indeed @ChrisTerBeke 
Not sure why it's dangling. Closing it now. :+1:

Issue #1634, opened 2017-04-01
Confirmation dialog for overrides cut off on high-dpi displays on Windows


mqudsi:
As you can see from this screenshot, the confirmation dialog for profile overrides is cut off on high-dpi displays on Windows. This is a regression from previous builds.

![image](https://cloud.githubusercontent.com/assets/606923/24582501/c3bd2630-16f6-11e7-9546-9b67e0471c79.png)

fieldOfView:
Since this dialog is new for Cura 2.5, it is hardly a regression. But it does need to be fixed, thanks for the report.

mqudsi:
Sorry, I meant regression as in "the previous dialog didn't exhibit this problem."

PR #1636, opened 2017-04-02
Fix issues with Workspace and Profile Changes dialogs

fieldOfView:
Some recently added dialogs don't work nicely with Hi DPI screens, and don't use the system theme properly. This PR fixes some of those issues.

Fixes https://github.com/Ultimaker/Cura/issues/1634

### DIFF
--- a/plugins/3MFReader/WorkspaceDialog.qml
+++ b/plugins/3MFReader/WorkspaceDialog.qml
@@ -74,7 +74,7 @@ UM.Dialog
             {
                 id: titleLabel
                 text: catalog.i18nc("@action:title", "Summary - Cura Project")
-                font.pixelSize: 22
+                font.pointSize: 18
             }
             Rectangle
             {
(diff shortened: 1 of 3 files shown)


fieldOfView:
I think Jenkins is having issues. I have only changed qml files, and AFAIK there are no tests for the qml files.

awhiemstra:
It failed on Windows because it could not find the Uranium files. The Windows build slave had some issues so I triggered a rebuild. Now it's fine. :)

Appesteijn:
I have tested this PR and there is a typo which causes a crash: CRITICAL - 3MFReader.WorkspaceDialog._createViewFromQML [234]: QQmlComponent error string file:///Applications/Cura.app/Contents/MacOS/plugins/plugins/3MFReader/WorkspaceDialog.qml:36 Cannot assign to non-existent property "fille"

Besides that this is how my Open Project dialog looks like on my MacBook Pro (2016) 
screen shot 2017-04-05 at 10 55 35

fieldOfView:
@Appesteijn, thanks for reporting. If you change that ```fille``` to ```fill```, does that also fix the layout issue?
See #1651 which fixes the typo

PR #1640, opened 2017-04-03
CURA-3421 Set SettingsComboBox value if we get undefined from resolve

LipuFei:


jackha:
Test failure was a temporary thing that has nothing to do with this issue. PR looks fine, merging and deleting branch

PR #1642, opened 2017-04-03
Feature arrange CURA-3239

jackha:
This PR adds some arrange functions which are used to find "good" locations for objects.

There are 3 parts that benefit from this functionality:
- When loading a model, the new object(s) are placed on "good" locations
- The right mouse button has an "Arrange All Models"
- When multiplying models, the new models are placed on a "good" location as well

### DIFF
--- a/cura/Arrange.py
+++ b/cura/Arrange.py
@@ -1,6 +1,9 @@
-import numpy as np
+import numpy
+from UM.Math.Polygon import Polygon
 
-##  Some polygon converted to an array
+
+##  Polygon representation as an array
+#
 class ShapeArray:
     def __init__(self, arr, offset_x, offset_y, scale = 1):
         self.arr = arr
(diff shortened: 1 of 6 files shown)


cura/Arrange.py
```diff
@@ -0,0 +1,172 @@
...
+        start_prio = 0
+        for i in range(count):
+            new_node = copy.deepcopy(node)
+
+            best_spot = self.bestSpot(
+                offset_shape_arr, start_prio = start_prio, step = step)
+            x, y = best_spot.x, best_spot.y
+            start_prio = best_spot.priority
+            transformation = new_node._transformation
+            if x is not None:  # We could find a place
```
LipuFei: Should we also check `y is not None` here?

cura/Arrange.py
```diff
@@ -0,0 +1,172 @@
...
+
+    ##  Helper to create an Arranger instance
+    #
+    #   Either fill in scene_root and create will find all sliceable nodes by itself,
+    #   or use fixed_nodes to provide the nodes yourself.
+    #   \param scene_root   Root for finding all scene nodes
+    #   \param fixed_nodes  Scene nodes to be placed
+    @classmethod
+    def create(cls, scene_root = None, fixed_nodes = None, scale = 0.5):
+        arranger = Arrange(220, 220, 110, 110, scale = scale)
```
LipuFei: Perhaps get actual print bed dimensions instead of hard-coded numbers?

jackha: I thought about that, but the exact dimensions do not matter here, because we try to center everything. Perhaps in the future if everybody wants to place objects near the rear wall...

cura/Arrange.py
jackha: From the implementation this is not necessary, but I doesn't hurt.

PR #1647, opened 2017-04-04
CURA-3648 Don't save authentication keys

LipuFei:
In combination with https://github.com/Ultimaker/Uranium/pull/240


LipuFei:
We just don't need to save the keys to a project file. Closing this one.

PR #1650, opened 2017-04-04
Mb make magic spiralize setting global

smartavionics:
Tim said to make the magic_spiralize setting no longer settable per mesh (or extruder) so that's what this PR does. When https://github.com/Ultimaker/CuraEngine/pull/488 is merged, this PR has to be merged also otherwise spiralization will break. Thanks.


BagelOrb:
There seems to be some weird commits as part of this PR.

smartavionics:
Yes, I don't know what's happened here. I'll trash this PR and  create a new one based on 2.5.

PR #1656, opened 2017-04-06
Add support for Peopoly Moai (SLA)

fieldOfView:
This PR adds a definition and a set of quality profiles for the upcoming Peopoly Moai:
https://www.kickstarter.com/projects/1554809440/moai-affordable-high-resolution-laser-sla-3d-print

Note that this printer is an SLA printer. Because of this the definition disables/hides a number of features (materials, temperatures (nozzle, buildplate), fan cooling, retraction, supports). Supports are intended to be added using a different tool before slicing.  

In cooperation with Shu (Mark) Peng from Peopoly

### DIFF
--- a/resources/definitions/peopoly_moai.def.json
+++ b/resources/definitions/peopoly_moai.def.json
@@ -1,0 +1,49 @@
+{
+    "id": "peopoly_moai",
+    "version": 2,
+    "name": "Moai",
+    "inherits": "ultimaker",
+    "metadata": {
+        "visible": true,
+        "author": "Aldo Hoeben",
+        "manufacturer": "Peopoly",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "has_machine_quality": true,
+        "has_materials": false
+    },
+
+    "overrides": {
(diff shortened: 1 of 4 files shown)


nallath:
Why does it inherit ultimaker machines? That doesn't make a whole lot of sense.

fieldOfView:
Hmm, weird. I'll fix that.

fieldOfView:
Fixed

jackha:
The definition didn't work without an inheritance. I've added inherits fdmprinter (although it isn't an fdmprinter) and merged it. Maybe we can add an slaprinter.def.json ?

fieldOfView:
'doh! I'm an idiot.

fieldOfView:
> Maybe we can add an slaprinter.def.json ?

That slaprinter.def.json would still need to inherit from fdmprinter. But it could be nice if more SLA printer manufacturers find Cura without materials, temperatures & support usefull

jackha:
No you're not an idiot :stuck_out_tongue_winking_eye:
Yes I hope more SLA printer manufacturers find Cura useful, would be awesome!

PR #1660, opened 2017-04-06
The magic_spiralize setting is no longer settable per mesh (or extruder).

smartavionics:
New PR created as the original was corrupted somehow (based on wrong parent, perhaps?)


BagelOrb:
Whoops. Accidentally merged to 2.5

PR #1670, opened 2017-04-11
First refactor for Z parse section to increase verbosity

tamarintech:
Moved translated M105s (blank, M0, or M1) above and outside of exception block.
Clarified the error message and added the line to it.

Related to https://github.com/Ultimaker/Cura/pull/1615 conversation.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -563,19 +563,19 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
             line = line[:line.find(";")]
         line = line.strip()
 
-        # Don't send empty lines. But we do have to send something, so send m105 instead.
-        if line == "":
+        # Don't send empty lines. But we do have to send something, so send
+        # m105 instead.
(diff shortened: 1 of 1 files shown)


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -563,19 +563,19 @@ def _sendNextGcodeLine(self):
             line = line[:line.find(";")]
         line = line.strip()
 
-        # Don't send empty lines. But we do have to send something, so send m105 instead.
-        if line == "":
+        # Don't send empty lines. But we do have to send something, so send
+        # m105 instead.
+        # Don't send the M0 or M1 to the machine, as M0 and M1 are handled as
+        # an LCD menu pause.
+        if line == "" or line == "M1" or line == "M1":
```
nallath: There is a little mistake here; It should be line == "M0" or line == "M1".

I will fix that in a sperate commit.

tamarintech: .... maybe I just really wanted to make sure it wasn't M1... :D

Sorry. Thanks! :)

PR #1681, opened 2017-04-12
CURA-3540 update profiles when deserializing them

LipuFei:
In combination with https://github.com/Ultimaker/Uranium/pull/243

### DIFF
--- a/plugins/XmlMaterialProfile/XmlMaterialProfile.py
+++ b/plugins/XmlMaterialProfile/XmlMaterialProfile.py
@@ -3,6 +3,7 @@
 
 import copy
 import io
+from typing import Optional
 import xml.etree.ElementTree as ET
 
 from UM.Resources import Resources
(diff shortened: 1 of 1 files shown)


plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -369,8 +370,30 @@ def clearData(self):
         self._dirty = False
         self._path = ""
 
+    def getConfigurationTypeFromSerialized(self, serialized: str) -> Optional[str]:
```
nallath: This isn't an optional right? It always returns a str.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -369,8 +370,30 @@ def clearData(self):
...
+                    raise InvalidInstanceError("Invalid version string '%s': %s" % (entry.text, e))
+                break
+        if version is None:
+            raise InvalidInstanceError("Missing version in metadata")
+        return version
+
     ##  Overridden from InstanceContainer
     def deserialize(self, serialized):
+        # update the serialized data first
+        from UM.Settings.Interfaces import ContainerInterface
```
nallath: I think its smarter to just use import (without from) on the top and use the full name here. Not a 100% sure if doing it like this is more expensive.

awhiemstra: In general, import statements in non-module context are more expensive. This is one of the reasons tools like pylint and style checkers recommend against putting import statements after code.

Issue #1441, opened 2017-02-17
How to override generic material <diameter> ? (and more)


filipgoc:
![image](https://cloud.githubusercontent.com/assets/16858474/23083613/cfd0a230-f55e-11e6-898e-5db5f39b1120.png)

Thanks to `<diameter>2.85</diameter>` close to the root of generic xml materials, the generic materials list 2.85 diameter even for machines that override it to 1.75 - furthermore it cannot be edited by user. 

![image](https://cloud.githubusercontent.com/assets/16858474/23083630/e33238ac-f55e-11e6-9050-863a1a9deaa5.png)

Maybe, `<diameter>___</diameter>` should not be a general property at all… still, how to override this? Or is it a bug to be squashed or a feature? 

The diameter is correctly overridden under print settings, just not under management. 
![image](https://cloud.githubusercontent.com/assets/16858474/23083632/e77d1ada-f55e-11e6-9904-912b856dbe41.png)

( I do wish to use general presets, I think it makes sense and corresponds to Cura’s current architecture.)

The default ```Setting (<setting key="print temperature">200</setting>)```
can be overridden when another one is nested under `<machine>`, which works great, but not the `<diameter>`… 

####  Related question
I would also like to override the `<color_code>` for different machines. Not color name, just the #- - -
I have a feeling it’s the same question.

ntoff:
I just create my own material profiles using this as a template

http://pastebin.com/MLSgU2Zr

I edit that file manually using notepad++ because when I add them through Cura, they don't seem to show up after restarting cura, editing them manually works fine though.

filipgoc:
Thanks @ntoff. Nice template!

However, my question is about overriding the <diameter> in the GLOBAL GENERIC profiles. 

(I have no issues making my own profiles, linking them with a specific machine, and with specific quality presets.)

These are the ones that have "generic_...." as their filename. These are a bit special. 
(example, generic_abs.xml.fdm_material)

They are the default fallback for fdmprinter, and most easily port to load generid quality profiles even when using custom color material profiles for one...

nallath:
I guess it should not really come as a suprise that we never actually tested the diameter property being used correctly (as all of our machines are 2.85). 

I'm also not that convinced that the diameter should be in the material in the first place. It's the machine that defines what the diameter is.

fieldOfView:
[...]

Issue #1308, opened 2017-01-05
Material diameter not loaded from material profile


dWoolridge:
Cura 2.4.0 beta
Windows 7 64-bit

The material diameter shown in the Print Setup panel does not change when loading a new material profile.

To Duplicate:

1. Create a custom material with a diameter of 1.75mm
2. Make sure that the Material->Diameter is checked in "Setting Visibility"
3. Load the default Material "Generic->ABS"
4. Load the default Quality Profile "Normal Quality - 0.1mm
5. Note that the Material->Diameter is shown as 2.85mm
6. Load the custom material with the 1.75mm diameter
7. Note that the Material->Diameter stays 2.85mm 

I have to adjust the flow slightly for each roll of filament that I have to optimize the printing.  I had planned to manage this by creating a Material profile for each roll of filament that I have.

It would also be nice to have a % flow adjustment for each material, rather than that also being saved in the Quality Profile.

Thanks,
--Doug

gregs007:
This also happens in 2.3.1 for Mac

morphlaugh:
Confirmed-- 2.3.1 for Mac also fails to load the 1.75mm diameter from the material profile when selected.

Ghostkeeper:
It seems we have two diameter settings:

* https://github.com/Ultimaker/Cura/blob/master/resources/definitions/fdmprinter.def.json#L1385
* The diameter in the XML profiles. This is obtained here: https://github.com/Ultimaker/Cura/blob/master/plugins/XmlMaterialProfile/XmlMaterialProfile.py#L426 but that variable is never used after that.

fieldOfView:
Seems ```"diameter": "material_diameter"``` needs to be added to ```__material_property_setting_map``` here:
https://github.com/Ultimaker/Cura/blob/master/plugins/XmlMaterialProfile/XmlMaterialProfile.py#L581

fbertram:
I might be missing something here, but as most printers will only accept filament of a certain diameter, it might make sense to move the material diameter to the printer profile. The advantage is, that the generic profiles for materials will be independent from the diameter. In case a printer can really load materials with multiple diameters, separate printer profiles would be required, which is still better than duplicating many material profiles.

Ghostkeeper:
The material profiles are dependent on the diameter though. A 2.85mm filament will require a bit more heat than a 1.75mm filament.

fieldOfView:
[...]

PR #1685, opened 2017-04-12
Filter materials by diameter

fieldOfView:
In Cura, material diameters set in (xml) material profiles override the material diameter set in the machine or variant definition. This causes 1.75mm printers to switch to 2.85mm when selecting eg "Generic PLA". Additionally, 2.85mm machines may show a lot of materials that don't apply to these printers since the merge of a set of materials into the fdm_materials repository.

This PR applies filters to the listed materials so they only show those materials that have relevant diameters. To be robust to 1.75/1.74 and 2.85/3.0 mm materials, the PR takes an approximated material_diameter defined in the printer definition and compares that to the same approximation of the material_diameter defined in the material definition. 

See https://github.com/Ultimaker/Cura/issues/1441#issuecomment-281120017 and https://github.com/Ultimaker/Cura/issues/1308#issuecomment-277666289 for discussion

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -244,7 +244,13 @@ class ExtruderManager(QObject):
                 material = materials[0]
             preferred_material_id = machine_definition.getMetaDataEntry("preferred_material")
             if preferred_material_id:
-                search_criteria = { "type": "material",  "id": preferred_material_id}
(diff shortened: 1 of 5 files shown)


fieldOfView:
Follow-ups to this PR include a PR to fdm_materials to include a set of generic 1.75mm materials (PLA, ABS, CPE, Nylon), and an additional field in the machine settings to easily set the material diameter of the machine.

Ghostkeeper:
Devs, see CURA-3684.

PR #1686, opened 2017-04-12
Fix NaN values on material print settings tab

fieldOfView:
This PR fixes the print settings showing up as NaN:
NaN material settings

See https://ultimaker.com/en/community/41810-retraction-is-nan for discussion

### DIFF
--- a/resources/qml/Preferences/MaterialView.qml
+++ b/resources/qml/Preferences/MaterialView.qml
@@ -273,10 +273,10 @@ TabView
                     {
                         id: spinBox
                         anchors.left: label.right
-                        value: parseFloat(provider.properties.value);
+                        value: (provider.properties.value != "None") ? parseFloat(provider.properties.value) : 0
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3037.

resources/qml/Preferences/MaterialView.qml
```diff
@@ -273,17 +273,28 @@ TabView
                     {
                         id: spinBox
                         anchors.left: label.right
-                        value: parseFloat(provider.properties.value);
-                        width: base.secondColumnWidth;
+                        value: {
+                            if (!isNaN(parseFloat(materialPropertyProvider.properties.value)))
```
LipuFei: Perhaps call `parseFloat()` just once? For example:
```
parsedValue = parseFloat(materialPropertyProvider.properties.value);
if (!isNaN(parsedValue))
    ...
```

Issue #1049, opened 2016-10-17
USB connection handling proposal


fieldOfView:
A while back there was some discussion about how Cura handles USB-connected printers.

Current problems:
- Cura does not properly (auto-)detect all printers, and a manual port-selection is missing
- Cura tries connecting to all serial ports on startup, resetting all connected printers (and other Arduino-based peripherals)
- A USB connected printer is "seen" by all printer-instances in Cura as "the" USB-connected printer
- In case of multiple connected printers, there is no way to see which printer is connected to which serial port

Most of these problems would be solved if Cura did not automatically assume that any connected USB serial port could be a 3d printer that any of the printer instances could print to. Instead I think connecting a printer via USB should be a conscious step for the user. When adding a printer Cura should ask the connection details. These connection details are then stored with the printer (stack) configuration.

I have made a fairly rough mockup plugin of what this could look like:
https://github.com/fieldOfView/USBConnectionMockup
Install the plugin and add a new printer (or press the "Connect via USB" button on the Printers preference pane). Note that this mockup is fully non-functional at this point.

![connect_via_usb](https://cloud.githubusercontent.com/assets/143551/19434597/a69bf698-9466-11e6-9b63-a20b92bf66ab.gif)

This is inspired by how the connection with the UM3 or OctoPrint are handled; I think that in the future we will see more and more printers being available via other means than USB, so USB should be regarded as "just another way of connecting to the printer" instead of being assumed the default and only way.

Some further thoughts:
- Cura would not need to open the serial port until it is needed (eg when starting a print, uploading firmware, switching to the monitor). Once the serial port has been opened, it can remain open, so the print-monitor keeps working.
- Cura could store (a hash of) the response to a M115 gcode command to check if the printer is indeed the same as it was once connected to. If it detects another response it could invalidate the connection info and present the usb connection window again.

thopiekar:
Using M115 was also my idea in the past. You probably remember that we both researched on how printers answer to this command and the problem was, that all the printers answer to this command differently.
[...]

Issue #70, opened 2015-06-25
Current print is sabotaged when launching new Cura


Xisle:
Application Version: 15.05.96
Platform: OS X 10.9.5

Steps to Reproduce:
Launch new Cura while printer is printing

Actual Results:
Print fails -1

Expected results:
Cura would launch without effecting current print

awhiemstra:
This is actually more of a bug in the USB serial firmware. It resets the printer when you try to connect to it, even when not sending any commands. Since we try to detect what printers are connected on launch, this can trip up the ongoing USB print.

I do not yet know how to solve this apart from completely delaying any attempts to determine connected printers, which in my opinion is not really a nice solution.

nickthetait:
One option might be a shared list of COM ports which are available or currently in use. If this list has not been created (first instance of Cura to be started) then the normal printer detection is run. If the list exists, then don't let a new instance connect to printers which are being used.

Xisle:
The old Cura connects to the printer after choosing "Print". Isn't that still a viable option? Some people want to try Cura before they buy a printer or just to produce an .stl.

What benefit is there to printer querying on startup? Cura should perform slicing without one IMO.

nallath:
Yeah, but it would be nice to know if the device you found is an device that you can actually use. If you connect an arduino with the old Cura, it thinks its a 3D printer. The polling ensures that we know if there are printers connected.

Xisle:
That leaves anyone without a printer unable to use Cura. 

How do you determine if the Arduino connected is a printer or not? Can that be done after clicking Print? 

I personally have every printer you make added to Cura 15.04 for helping others on the forum and in chat. This is impossible with the new Cura :|

nallath:
No, it simply checks if there are printers. It will still be able to save to SD / G-code. I don't know why you think that isn't possible. It simply wants to know if the option of usb streaming is at all possible before doing anything.  

It can be done after you click the button. But the main point of what we do now is that we know before hand if there is a printer. I think its kinda bad user experience to see a 'print with usb' button, only to find out after you clicked it that it doesnt work. 

The only way we know that something is a printer is simply sending some commands to it and checkin its responses. If those are okay, we can safely assume its a 3D printer.

Xisle:
First of all, I resolved my problem of not being able to slice without a printer. The UM2 I had added to an earlier beta version was not allowing slicing - adding a new UM2 fixed it.
 
Issue #69  and Issue #70 are similar.

[...]

PR #1688, opened 2017-04-14
[WIP] Rework connection via USB

fieldOfView:
This PR aims to solve the following issues:

* Cura does not properly (auto-)detect all printers, and a manual port-selection is missing
* Cura tries connecting to all serial ports on startup, resetting all connected printers (and other Arduino-based peripherals)
* A USB connected printer is "seen" by all printer-instances in Cura as "the" USB-connected printer
* In case of multiple connected printers, there is no way to see which printer is connected to which serial port

It works off this proposal: https://github.com/Ultimaker/Cura/issues/1049. Also see https://github.com/Ultimaker/Cura/issues/70 for discussion.

Instead of automatically trying to connect to any USB serial device, this PR will present the user with a dialog to deliberately connect a printer to a serial port (or an automatic option, which still allows the user to scan all ports). It will also allow the user to disable connecting to the USB printer on startup, so Cura no longer resets the printer when starting.


fieldOfView:
I have created this premature PR mostly so it can get a JIRA issue and you know what's coming.

Ghostkeeper:
For devs, see CURA-3690.

awhiemstra:
Note that the unit tests are failing because you branched at a point where unit tests in master were broken. Merging master should fix that. :)

Ghostkeeper:
Can you let us know when you consider this finished, @fieldOfView?

fieldOfView:
Not for a while. I can close this in the mean time and reopen when I am done. Let me know what you prefer.

nallath:
We just had a research sprint, so we were a bit slow to respond. I think closing this and opening it when its done is the best option.

fieldOfView:
I am fairly happy with the interface work, but getting it all functional is taking more time than expected. I am retracting this PR for now, and will return once I have a working solution (hopefully in time for 2.7)

PR #1694, opened 2017-04-15
Rigid3D Zero2 Printer Profile

Rigid3D:
Adding printer profile for the second generation of the Rigid3D Zero printer.


nallath:
For our reference; CURA-3691

Ghostkeeper:
The provided platform mesh is quite heavy. Our guideline (or soft requirement) is to stay below 1MB. Could you reduce the resolution of that mesh a bit? Would you like me to do that for you?

Rigid3D:
If you like, you can transform. If not urgent, I can do it tonight.

Ghostkeeper:
It's not urgent :) The 2.6 release is still far away. I'll let you do it so you can check if the quality stays acceptable.

Myself I'd just load it into Blender and apply a decimate filter on the mesh, and tune it a bit so that the corners and inaccuracies fall nicely and the mesh looks reasonable.

Ghostkeeper:
Thanks!

PR #1698, opened 2017-04-18
CURA-3574: Use a toggle button mode switching

LipuFei:


resources/qml/Sidebar.qml
```diff
@@ -415,6 +417,54 @@ Rectangle
             anchors.top: parent.top
             anchors.left: parent.left
             width: parent.width
+        }*/
+
+        Text
+        {
+            id: toggleLeftText
+            anchors.right: modeToggleSwitch.left
+            anchors.rightMargin: 10
```
fieldOfView: Unthemeable number, please use UM.Theme.getSize()

resources/qml/Sidebar.qml
```diff
@@ -415,6 +417,54 @@ Rectangle
...
+            color: UM.Theme.getColor("toggle_active_text")
+            font: UM.Theme.getFont("default")
+        }
+
+        Switch
+        {
+            id: modeToggleSwitch
+            checked: false
+            anchors.right: toggleRightText.left
+            anchors.rightMargin: 10
```
fieldOfView: Unthemeable number, please use UM.Theme.getSize()

resources/themes/cura/styles.qml
```diff
@@ -8,6 +8,44 @@ import QtQuick.Controls.Styles 1.1
 import UM 1.1 as UM
 
 QtObject {
+    property Component toggle_button: Component {
+        SwitchStyle {
+            groove: Rectangle {
+                implicitWidth: 30
+                implicitHeight: 15
+                radius: 9
```
fieldOfView: Unthemeable numbers, please use UM.Theme.getSize()

resources/themes/cura/styles.qml
```diff
@@ -8,6 +8,44 @@ import QtQuick.Controls.Styles 1.1
...
+                    }
+                }
+                Behavior on border.color { ColorAnimation { duration: 50; } }
+                border.width: 1
+            }
+
+            handle: Rectangle {
+                implicitHeight: 15
+                implicitWidth: 15
+                radius: 9
```
fieldOfView: Unthemeable numbers, please use UM.Theme.getSize()

LipuFei:
@fieldOfView I've put those numbers into theme.

fieldOfView:
I can't say this is a big improvement. The hit-area for the toggle is extremely small compared to the previous solution(*). Please can the labels be clickable as well?

Style-wise I think this throws consistency out of the window. The control does not look or behave like any of the other control in the interface. Different colors, no change on hover, and the size does not fit will with other controls. In simple mode "Light" infill now seems to be recommended, and "Solid" seems to be the custom option. 

(*): Don't just take it from me, see Fitt's law: https://en.wikipedia.org/wiki/Fitts%27s_law ("This scientific law predicts that the time required to rapidly move to a target area is a function of the ratio between the distance to the target and the width of the target.")

Ghostkeeper:
Making the labels clickable is something I think we can all agree on. It's been said by two of my colleagues as well when they first saw this.

The consistency of this design is a contentious issue and has been under discussion for a while. I agree with fieldOfView that it's inconsistent and would prefer tabs, but the opposing viewpoint is that it's confusing to have 3 levels of tabs beneath each other with different styles. This is what we ended up with and I don't think it would change soon (until the next redesign, that is).

fieldOfView:
The switch does not have to be implemented as tabs to be consistent. The toggle switch can be made to match the design and behavior of other controls in the application. There are ways to separate the recommended/custom labels from the infill options.

LipuFei:
@fieldOfView @Ghostkeeper Made the labels clickable: 7db0f0ebb777651154f79b33ee91fd49107086b5

PR #1699, opened 2017-04-18
tweak start gcode (M109) for jellybox 

filipgoc:
swap M109 for M104 in the start gcode

- It’s necessary to preheat the nozzle fully before the auto bed leveling
procedure to loosen up leftover filament that may had hardened on the
nozzle.


Ghostkeeper:
Devs, see CURA-3698.

Ghostkeeper:
I can't test on the actual printer, but the resulting g-code output still looks like it's working all right. Thanks.

PR #1702, opened 2017-04-19
Only update the target bed temperature, without submitting it to the printer. To prevent race conditions.

daid:
This fix no longer sends a temperature update to the printer when Cura receives a bed temperature update from the printer.
We had race conditions where the bed was not shut off at the end of a printer, and Cura is the only possible source of this problem.
CURA-3702

### DIFF
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
@@ -283,10 +283,8 @@ class NetworkPrinterOutputDevice(PrinterOutputDevice):
     #
     #   /param temperature The new target temperature of the bed.
     def _setTargetBedTemperature(self, temperature):
-        if self._target_bed_temperature == temperature:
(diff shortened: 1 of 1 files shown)


nallath:
See: CURA-3703

plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
```diff
@@ -294,6 +294,15 @@ def _setTargetBedTemperature(self, temperature):
         put_request.setHeader(QNetworkRequest.ContentTypeHeader, "application/json")
         self._manager.put(put_request, data.encode())
 
+    ##  Updates the target bed temperature from the printer, and emit a signal if it was changed.
+    #
+    #   /param temperature The new target temperature of the bed.
+    def _updateTargetBedTemperature(self, temperature):
```
LipuFei: The first few lines in `_setTargetBedTemperature()` does the same thing here. Could you replace those code in `_setTargetBedTemperature()` with this function call as well?

PR #1710, opened 2017-04-21
Theme picker

victor9999:
Feature that allows to change themes


nallath:
For our reference; CURA-3714 

Thanks for making the pull request!

fieldOfView:
@rudowinger is going to be interested in this. How's that dark theme coming along @rudowinger?

awhiemstra:
My main comment with this is going to be that if we add yet another setting to the preferences, we really should split up the General preferences page.

Ghostkeeper:
That's not up to an external contributor to decide or bring up, really.

rudowinger:
Great! Can't wait to stop fearing about editing the main theme, also having some flexibility :)
@fieldOfView Great! Lost some time due to work and stuff but was able to optimize even more stuff, Now to find the new tickbox lines in the style...
dark

fieldOfView:
Ack! The new layerview slider panel does not use properly themed colors! How did I miss that?

I'll get that fixed.

awhiemstra:
@jackha actually already made a branch for that, we just never really got around to testing and integrating it properly.

awhiemstra:
The general/theme preference is actually already set in Uranium, see https://github.com/Ultimaker/Uranium/blob/master/UM/Qt/Bindings/Theme.py#L40

LipuFei:
@awhiemstra Ah, ok. @victor9999 then there just needs to be reset function :)

fieldOfView:
As a follow-up it would be nice if there were a ThemesModel that could be used instead of a hardcoded list, but that would probably live in Uranium

victor9999:
Added function to reset theme

PR #1711, opened 2017-04-21
Gcode keywords

victor9999:
Feature that adds additional gcode keywords


Ghostkeeper:
Devs, see CURA-3718.

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -442,6 +442,16 @@ def _onSlicingFinishedMessage(self, message):
         self.backendStateChange.emit(BackendState.Done)
         self.processingProgress.emit(1.0)
 
+        for line in self._scene.gcode_list:
+            replaced = line.replace("{print_time}", str(Application.getInstance().getPrintInformation().currentPrintTime.getDisplayString(DurationFormat.Format.ISO8601)))
+            replaced = replaced.replace("{filament_amount}", str(Application.getInstance().getPrintInformation().materialLengths))
+            replaced = replaced.replace("{filament_weight}", str(Application.getInstance().getPrintInformation().materialWeights))
+            # TODO: calculate filament cost
+            replaced = replaced.replace("{filament_cost}", "Not yet implemented")
```
Ghostkeeper: str(Application.getInstance().getPrintInformation().materialCosts)

Seems to work :)

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -442,6 +442,16 @@ def _onSlicingFinishedMessage(self, message):
...
+            replaced = replaced.replace("{jobname}", str(Application.getInstance().getPrintInformation().jobName))
+
+            self._scene.gcode_list[self._scene.gcode_list.index(line)] = replaced
+
```
Ghostkeeper: There is a better place for all this: `plugins.CuraEngineBackend.StartSliceJob._buildGlobalSettingsMessage`.

In there, we already create a few keywords such as `{time}`, `{date}` and `{day}`. You could add your new keywords to it very easily. What's more, that function performs the replacing only on the actual start and end g-codes rather than on all of the g-code, and does it in one query rather than a query for each line, which is much more efficient.

victor9999: @Ghostkeeper this is added here, because cura doesn't know these values before slicing process is completed

plugins/CuraEngineBackend/CuraEngineBackend.py
victor9999: The base feature was implemented in 2016, so this was not present then

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -442,6 +442,15 @@ def _onSlicingFinishedMessage(self, message):
...
         self.processingProgress.emit(1.0)
 
+        for line in self._scene.gcode_list:
+            replaced = line.replace("{print_time}", str(Application.getInstance().getPrintInformation().currentPrintTime.getDisplayString(DurationFormat.Format.ISO8601)))
+            replaced = replaced.replace("{filament_amount}", str(Application.getInstance().getPrintInformation().materialLengths))
+            replaced = replaced.replace("{filament_weight}", str(Application.getInstance().getPrintInformation().materialWeights))
+            replaced = replaced.replace("{filament_cost}", str(Application.getInstance().getPrintInformation().materialCosts))
+            replaced = replaced.replace("{jobname}", str(Application.getInstance().getPrintInformation().jobName))
+
+            self._scene.gcode_list[self._scene.gcode_list.index(line)] = replaced
```
LipuFei: You can use `enumerate()` in the `for` loop to get both the index and the content.

LipuFei:
Hi @victor9999 , the keywords you are replacing don't seem to exist... Could you elaborate more on those changes?

Ghostkeeper:
The keywords are new. They provide new functionality: A user can enter the keywords in the start- or end-g-code, and they'll get replaced by the appropriate information.

Ghostkeeper:
I'm a bit wary of the performance impact of this on large g-code. I'm going to test the feature now, so we'll see :)

Ghostkeeper:
I sliced a cube with 4000% infill, giving a g-code file of 160MB. The interface hiccupped a bit but that always happens when the front-end receives the g-code message. It didn't seem to be significantly longer. So the performance seems to be fine.

PR #1712, opened 2017-04-21
Reset machine

victor9999:
Allow reset machine to default settings


fieldOfView:
I'm not entirely sure what the functionality is trying to do. Is it supposed to remove all customisations made in the Machine Settings window? If so, wouldn't this button better be placed in the Machine Settings action?

An easier way to (programmatically) clear the definition_changes container is to simply call the clear() method of the container.

nallath:
I'm with @fieldOfView with regards to this one. The only way one can have custom machines if they have the changes action. That would also be a spot to have a "reset"

PR #1713, opened 2017-04-21
CURA-3634: Make prime blob optional

LipuFei:


resources/definitions/fdmprinter.def.json
```diff
@@ -3290,6 +3290,16 @@
             "description": "Adhesion",
             "children":
             {
+                "prime_enable":
+                {
+                    "label": "Enable Prime",
```
BagelOrb: I'm not sure whether this naming is good.
Priming is making the printer ready. This involves unretracting as well.

Did you have a talk with somebody about the best name for this setting?

Maybe Menno is the right guy?
Also Jeroen from support might have something to say about this.

resources/definitions/fdmprinter.def.json
```diff
@@ -3290,6 +3290,16 @@
...
+                    "description": "Whether to prime the filament before printing.",
```
BagelOrb: Please explain what benefit the user has by changing this setting.

resources/definitions/fdmprinter.def.json
```diff
@@ -3290,6 +3290,16 @@
             "description": "Adhesion",
             "children":
             {
+                "prime_poop_enable":
+                {
+                    "label": "Enable Prime Blob",
+                    "description": "Whether to prime the filament with a blob before printing.",
```
BagelOrb: Please explain what benefit the user has by changing this setting.

The description should explain to a user why and when he should change the setting.

It should explain the benefit of turning it off and the benefit of turning it off.

LipuFei:
@BagelOrb updated the setting name and the description.

PR #1732, opened 2017-04-24
Add Extruders to Context Menu and Refactor

awhiemstra:
This PR makes a number of changes to the context menu:

- It adds the ability to select what extruder is used for the selected objects.
- It merges the "per-object" and "global" context menus.
- The "per object" actions Center, Delete and Multiply have been changed to work on the entire selection.
- Extruder selection through the context menu can be undone/redone.


cura/CuraActions.py
```diff
@@ -1,10 +1,22 @@
...
 
 from UM.Event import CallFunctionEvent
 from UM.Application import Application
+from UM.Math.Vector import Vector
+from UM.Scene.Selection import Selection
+from UM.Operations.GroupedOperation import GroupedOperation
+from UM.Operations.RemoveSceneNodeOperation import RemoveSceneNodeOperation
+from UM.Operations.SetTransformOperation import SetTransformOperation
+
+from .SetParentOperation import SetParentOperation
```
nallath: Should we not use: from cura.SetParentOperation?

cura/CuraActions.py
```diff
@@ -23,5 +35,56 @@ def openBugReportPage(self):
...
+        operation.push()
+
+    ##  Multiply all objects in the selection
+    #
+    #   \param count The number of times to multiply the selection.
+    @pyqtSlot(int)
+    def multiplySelection(self, count: int) -> None:
+        job = MultiplyObjectsJob(Selection.getAllSelectedObjects(), count, 8)
+        job.start()
+        return
```
nallath: Return is not required (probably copy pasta issue)

cura/CuraActions.py
```diff
@@ -23,5 +35,56 @@ def openBugReportPage(self):
...
+            if group_node and group_node.callDecoration("isGroup") and group_node not in removed_group_nodes:
+                remaining_nodes_in_group = list(set(group_node.getChildren()) - set(nodes))
+                if len(remaining_nodes_in_group) == 1:
+                    removed_group_nodes.append(group_node)
+                    op.addOperation(SetParentOperation(remaining_nodes_in_group[0], group_node.getParent()))
+                    op.addOperation(RemoveSceneNodeOperation(group_node))
+        op.push()
+
+    @pyqtSlot(str)
+    def setSelectionExtruder(self, extruder_id: str) -> None:
```
nallath: The naming is different compared to the others; deleteSelection vs setSelectionExtruder. I'd say it should be osmething like "setExtruderForSelection"  or something like that.

cura/CuraApplication.py
```diff
@@ -1320,3 +1333,10 @@ def checkIsValidProjectFile(self, file_url):
         except Exception as e:
             Logger.log("e", "Could not check file %s: %s", file_url, e)
             return False
+
+    def _onContextMenuRequested(self, x, y):
```
nallath: Type hinting please (also for protected functions, as it's also there to protect us from mistakes)

cura/Settings/ExtrudersModel.py
```diff
@@ -1,12 +1,12 @@
 # Copyright (c) 2016 Ultimaker B.V.
 # Cura is released under the terms of the AGPLv3 or higher.
 
-from PyQt5.QtCore import Qt, pyqtSignal, pyqtProperty
+from PyQt5.QtCore import Qt, pyqtSignal, pyqtProperty, pyqtSlot
 
 import UM.Qt.ListModel
 from UM.Application import Application
 
-from cura.Settings.ExtruderManager import ExtruderManager
```
nallath: Why change the import statement?

resources/qml/Menus/ContextMenu.qml
```diff
@@ -0,0 +1,128 @@
...
+            SpinBox
+            {
+                id: copiesField
+                minimumValue: 1
+                maximumValue: 99
+            }
+        }
+    }
+
+    function findItemIndex(item)
```
nallath: Documentation; What does this function do?

resources/qml/Menus/ContextMenu.qml
```diff
@@ -0,0 +1,128 @@
...
+    {
+        for(var i in base.items)
+        {
+            if(base.items[i] == item)
```
nallath: Missing {}

cura/CuraActions.py
awhiemstra: I do not think we ever really agreed on something in that regard, but since the other import also uses from cura. I changed these to match that.

cura/CuraActions.py
awhiemstra: Yeah, this was also in the original method.

cura/CuraActions.py
awhiemstra: Fair point. Changed.

cura/CuraApplication.py
awhiemstra: Good point.

cura/Settings/ExtrudersModel.py
awhiemstra: I think I accidentally removed the import at some point then re-added it later.

resources/qml/Menus/ContextMenu.qml
awhiemstra: Fixed.

nickthetait:
"ability to select what extruder is used for the selected objects" This is a feature people have been asking for at least 3 years. And now it is finally here \o/

fieldOfView:
Note that this has been possible since 2.3 via per model settings, but it is less hidden now.

PR #1740, opened 2017-04-25
Add Extruder selection buttons to the Toolbar

awhiemstra:
This adds buttons to the toolbar that allow you to select what extruder the currently selected objects are printed with. They match the behaviour of the actions in the context menu.

Additionally, it removes extruder selection from per-object settings and changes per-object settings to only be enabled in advanced mode again.


resources/qml/ExtruderButton.qml
```diff
@@ -0,0 +1,86 @@
...
+
+        width: UM.Theme.getSize("default_margin").width
+        height: UM.Theme.getSize("default_margin").height
+
+        border.width: UM.Theme.getSize("default_lining").width
+        border.color: UM.Theme.getColor("lining");
+    }
+
+    // See line 51, same workaround
+    MouseArea
```
LipuFei: Check `Siderbar.qml` line 411-434, I had the same issue but found another solution. Maybe that one works here too.

resources/themes/cura/styles.qml
```diff
@@ -206,7 +206,10 @@ QtObject {
                     property bool down: control.pressed || (control.checkable && control.checked);
 
                     color: {
-                        if(control.checkable && control.checked && control.hovered) {
+                        if(control.customColor !== undefined && control.customColor !== null) {
+                            return control.customColor
+                        }
```
LipuFei: Coding style: `} else if (...)`?

resources/qml/Toolbar.qml
```diff
@@ -135,4 +156,15 @@ Item {
...
+    UM.SettingPropertyProvider
+    {
+        id: machineExtruderCount
+
+        containerStackId: Cura.MachineManager.activeMachineId
+        key: "machine_extruder_count"
+        watchedProperties: [ "value" ]
+    }
+
+    UM.I18nCatalog { id: catalog; name: "cura" }
```
LipuFei: (Not necessary) should we use the same coding style as above? (expand to multiple lines)

awhiemstra: I tend to use a single line for stuff that only has one or two properties. We do this in quite a few more QML files.

resources/qml/ExtruderButton.qml
awhiemstra: Interestingly enough, just removing the "checkable: true" fixes the issue with bindings breaking and still allows checked to work properly. So I guess I now know how to solve this properly.

resources/qml/Toolbar.qml
awhiemstra: However, in this case it turns out the catalog has become completely obsolete, so I removed it instead.

Appesteijn:
👍

PR #1741, opened 2017-04-25
Highlight text of the selected mode

LipuFei:
image
Updated to reflect 703f1c0


fieldOfView:
This makes the toggle switch itself rock back and forth a couple of pixels when switching because the widths of the texts change.

LipuFei:
@fieldOfView nice catch! :D I will polish it a little bit.

LipuFei:
@fieldOfView I did some refactoring and used Button for the texts, so we don't have the size changing issue. Also the text colour changes when mouse hovers.

PR #1743, opened 2017-04-25
Fix layer view menu items

fieldOfView:
This PR fixes styling issues with the text in the layer view options which are currently drawn in unthemed color (ie: system text default color). It build on a branch by @jackha, but adds a couple of improvements.

* Label elements can't use a TextFieldStyle
* Removed code duplication
* Add color swatch to extruders when applicable
* Use proper names for extruders from ExtrudersModel instead of "Extruder %d"


fieldOfView:
Issues in @awhiemstra's screenshot fixed (Labels were still using system colors because of incorrect use of style property)

nallath:
CURA-3730

Issue #1716, opened 2017-04-21
materials disapear until scrolling


MarcusWolschon:
In Cura 2.4 and 2.5 some custom materials are missing from the list in "manage materials" until you scroll down and scroll back up again.

Ghostkeeper:
Whoa, weird.

What operating system are you using?

Does it appear as a blank line in the material manager or is it just not listed until you scroll? If it's the first, we might be looking at a Qt bug. If it's the second, perhaps it's an update bug resulting from us asynchronously filling the list or something...?

MarcusWolschon:
Windows
They are just not listed until I scroll.
Since it's only custom materials that are affected, my guess is some update bug too. That the materials are added later but after adding them the model does not inform the view that there is an update.

fieldOfView:
See https://github.com/Ultimaker/Cura/pull/1757/commits/a727e4ffd09d365dfaa3f532c912a952f30e7089

PR #1757, opened 2017-04-29
Fix material uniqueness

fieldOfView:
This PR adds a "Create" button to the Materials pane. It duplicates generic PLA into a new material with a new GUID, so it is unique (instead of a material that is linked by its GUID to the original material). This allows users to properly create new materials.

Because people have been abusing the "Duplicate" button for a while now, this PR also adds a notice to materials that are linked to another material, and a button to unlink the material. This way unproperly created materials can be amended.

Current implementation:
image
Note that I changed the name of the duplicated material manually.


ex-nerd:
Thanks for adding the option to unlink.  I will probably still use "duplicate" because it's far easier than switching between modal screens to look up and copy values like print temperature and density whenever I want to just add a new cost/color from a brand I've purchased before.

Hopefully the linked/unlink stuff will avoid some confusion around the unexpected shared nature of the material GUIDs.  I notice the warning says "some ... properties".  Is there any way to show in the UI which properties are shared?

Ghostkeeper:
Devs, see CURA-3750.

cura/Settings/ContainerManager.py
```diff
@@ -694,6 +695,67 @@ def duplicateMaterial(self, material_id: str) -> str:
...
+        # This ensures that the inheritance goes well and all "cut up" subclasses of the xmlMaterial profile
+        # are also correctly created.
+        with open(containers[0].getPath(), encoding="utf-8") as f:
+            duplicated_container.deserialize(f.read())
+
+        duplicated_container.setMetaDataEntry("base_file", new_id)
+        duplicated_container.setMetaDataEntry("GUID", str(uuid.uuid4()))
+        duplicated_container.setName(catalog.i18nc("@label", "Custom Material"))
+
+        self._container_registry.addContainer(duplicated_container)
```
Ghostkeeper: I think you forgot to return something here. What it returns is also not documented.

Edit: Further down you check against this return value by comparing it with `""`. This currently seems to work because when a material is successfully created you get `None` which is not equal to the empty string. But still, please change it.

fieldOfView: I duplicated the behavior/code from duplicateMaterial().
Shall I fix that too?

fieldOfView:
Fixed the review issue.
Note that the functionality currently results in an error due to an issue on Uranium:
https://github.com/Ultimaker/Uranium/issues/249

fieldOfView:
> I notice the warning says "some ... properties". Is there any way to show in the UI which properties are shared?

As far as visible properties are concerned, it is the filament cost and filament weight. But more importantly, it is the GUID itself, which is used to identify the material eg in the UM3. So if you start with a PLA, and then change it with values for - say - PEEK, the printer will(should?) probably show all kinds of warnings that you are printing way too hot because it thinks you are printing with PLA.

cura/Settings/ContainerManager.py
Ghostkeeper: Sure, that'd be nice.

fieldOfView: Done

cura/Settings/ContainerManager.py
```diff
@@ -693,6 +696,70 @@ def duplicateMaterial(self, material_id: str) -> str:
...
+        return new_id
+
+    ##  Create a new material by cloning Generic PLA and setting the GUID to something unqiue
+    #   \return \type{str} the id of the newly created container.
+    @pyqtSlot(result = str)
+    def createMaterial(self) -> str:
+        # Ensure all settings are saved.
+        Application.getInstance().saveSettings()
+
+        containers = self._container_registry.findInstanceContainers(id="generic_pla")
```
fieldOfView: No, because then we would have no functional difference between Duplicate and Create.

Create should be used for people that want to create a material profile for - say - Colorfabb CopperFill or Colorfabb XT. What I could do is change the material type to eg "New material", so the user is forced to enter a material type.

fieldOfView: Both Brand and Material Type are now set to "Custom", inviting the user to enter the proper value.

jackha:
Great stuff. Looks very useful as new material creation is indeed very confusing.

I've pulled the branch and merged master into it, together with Uranium master. I have the following findings/questions:

- It seems that a custom material starts linked to itself? Is that intentional and is that the correct behavior?
- After duplicating PLA into PLA dub with material type PLAdub and I unlink it, the "unlink" is "not saved" after restart. I see 2 resulting generic_pla material files, one specific for the machine and one generic. In my case these are generic_pla+%232.xml.fdm_material and generic_pla+%232_ultimaker2_plus_0.4_mm.xml.fdm_material. The generic_pla carries the PLA guid, while the ultimaker2_plus variant has a new guid. 
- Unlinking Custom materials that are duplicated from the parent Custom material is also "not saved"
- When the brand or material type field is selected, I cannot change material by clicking in the list of materials.

fieldOfView:
I'll look into these issues. The not saving is strange. I set metadata of the containers, if that is not saving then there is something weird going on with not saving containers when their metadata is changed.

fieldOfView:
I fixed the issues found by @jackha, and fixed another issue I found while working on those fixes.

fieldOfView:
Also fixes https://github.com/Ultimaker/Cura/issues/1716

jackha:
Super! Thanks a lot ^^

teriperi:
Hello @fieldOfView ! I've verified CURA-3750 and I have good news! The CREATE button is present and works as described for Ultimaker and Custom FDM printers. However, for Other printers, I noticed that the create button is visible and enabled. When pressed, there is no feedback and nothing happens. Should the CREATE button be disabled or should a new custom material be created for Other printers?

fieldOfView:
What "Other" printers did you try?

fieldOfView:
I am guessing you tried one or more machines with 1.75 mm material. Cura now creates a new 2.85 mm material, which is subsequently filtered out from the display because it is not suitable for the current printer.

I'll fix this, but let me know what "Other" printers you tried, so I can see if there's not another issue too.

teriperi:
Ah! I have tried on 101 Hero, Zone 3d, Cartesio as random test. All three with 1.75mm. Then I saw German RepRap Neo and that one has 2.85mm and thus, the Create button works for that one. These are the only ones I've tried.

fieldOfView:
Fixed in https://github.com/Ultimaker/Cura/pull/1839

Issue #1759, opened 2017-04-29
include ALYA 3D printer in source code


nbennett1978:
Hi,
we are manufacturer of ALYA, a FFF printer for kids and youngsters. I'd like to ask to include the printer definitions into the source code. I tried to do that myself, but couldn't due to user permissions.
[alya3dp.def.json.txt](https://github.com/Ultimaker/Cura/files/966412/alya3dp.def.json.txt)

nbennett1978:
We used legacy cura (15.04) up until recently. The printer definition file i sent you was created with the generator link on cura's github repo- what exactly did you mean with your recommendation to update?

fieldOfView:
This is fixed now that #1761 has been merged. The Alya 3d printer will show up in the list of printers when adding a new printer in version 2.6 and up.

PR #1761, opened 2017-04-30
Add definition for Alya 3d printer

fieldOfView:
See https://github.com/Ultimaker/Cura/issues/1759

### DIFF
--- a/resources/definitions/alya3dp.def.json
+++ b/resources/definitions/alya3dp.def.json
@@ -1,0 +1,61 @@
+{
+    "id": "alya3dp",
+    "name": "ALYA",
+    "version": 2,
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "ALYA",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3759.

Ghostkeeper:
The nozzle offset has two dimensions: X and Y. They refer to the same dimensions as whatever counts as `X` and `Y` in the `G1` commands of the g-code. They are a relative offset that is compensated for when printing with multiple extruders. The zero-coordinate of the nozzle offset is meant to be the position of the head, and the position of the head is assumed to be able to move between the coordinates 0,0 and width,depth.

The same goes for the head shape and collision polygons. They are relative to the position of the head.

You don't need to set the nozzle offset for single-extrusion printers (in fact, you can't even do that). Not setting the machine_head_shape stuff will result in a possible collision with previously printed models if one-at-a-time mode is activated.

Issue #775, opened 2016-06-07
Cura [2.1.2] - Ultimaker 2 No multiple nozzle size


SamGuliver:
If you have an Ultimaker 2 and upgraded it with the Olson block, you cannot choose between nozzle sizes in the Software, if you have the right printer selected!
Needs to be fixed!!

fieldOfView:
Yeah, we actually had that at one point but unfortunately it got cut... Made me sad to see functional code getting removed.

What you could do is lie, and say you have an Ultimaker 2+. That will get you the nozzle selection, as well as material optimizations.

MarcusWolschon:
I can select my nozzle size for an UM2 (not +) extended.
As it doesn't come with an Olsson block as shipped, you may need to enter preferences
and make this setting visible.

bortek:
I installed Cura 2.3.0 for my "Ultimaker 2" and I cannot find where to set the nozzle size. If I change to Ultimaker 2+ for example then the option appears on the top where you can select material and nozzle size. But not for "Ultimaker 2". I have enabled all settings in the preferences still nothing. Is this a bug? I cannot use this version of Cura if I cannot set the nozzle size.

fieldOfView:
Check the Machine Settings on the Printers pane of the Preferences

bortek:
Oh dear it was hidden right there deep under the hood :) 👍  . Any plan to move it out to the surface ?

Ghostkeeper:
The Ultimaker 2 only had a 0.4mm nozzle, so others are not supported.

That said, we did have the nozzle size setting exposed as a normal setting for a while. We eventually removed it again because it is technically a machine setting, like gantry height or build volume width, that is not easily modified.

bortek:
I think most,  if not all,  um2 users have either olsson block or some other mods to change the nozzle. If you have um2 you would eventually come to the point where you experiment and want ti change the nozzle. That said I  still believe we should have a setting exposed. If not by default then perhaps cia a preference that would make it visible or hidden on the gui just like for um2+

MarcusWolschon:
With the new **Cura 2.4**

-  I can not select any nozzle sizes for UM2,
-  can not create a UM2+Olsson Block printer and
-  with the UM2+ selected, I can only select 0.25, 0.4, 0.6 and 0.8mm. However I do all structural parts with a 1.0mm nozzle that can not be selected.

nallath:
[...]

PR #1762, opened 2017-04-30
Add Upgraded Parts action for UM2 to add support for Olsson block

fieldOfView:
This PR adds a set of variants that just set the nozzle diameter and nozzle tip diameter. These variants are not used until the machine action sets the has_variants metadata entry.

See https://github.com/Ultimaker/Cura/issues/775#issuecomment-297284176 for discussion


nallath:
For our reference; CURA-3742

nallath:
Made a few more changes so it worked with cura specific stacks.

PR #1764, opened 2017-04-30
Fix hover effects

fieldOfView:
This PR adds hover effects to the new mode switch toggle and enabled support checkbox in simple mode. It also adds themeable colors to the mode switch toggle to make it fit more with the other controls.

It is an alternative to https://github.com/Ultimaker/Cura/pull/1741


nallath:
Our designers didn't like the colors as they are now, but as these are now themable, I think this can be merged.

PR #1765, opened 2017-05-01
CURA-3730: Hide Top/Bottom and Inner Wall legends in "Material Color"

LipuFei:
Don't show "Top/Bottom" and "Inner Wall" legends in "Material Color" in Layer View.

### DIFF
--- a/plugins/LayerView/LayerView.qml
+++ b/plugins/LayerView/LayerView.qml
@@ -273,17 +273,20 @@ Item
                     {
                         typesLegenModelNoCheck.append({
                             label: catalog.i18nc("@label", "Top / Bottom"),
-                            colorId:  "layerview_skin"
+                            colorId: "layerview_skin",
+                            visible: UM.Preferences.getValue("layerview/layer_view_type") != 0
                         });
                         typesLegenModelNoCheck.append({
                             label: catalog.i18nc("@label", "Inner Wall"),
-                            colorId:  "layerview_inset_x"
+                            colorId: "layerview_inset_x",
+                            visible: UM.Preferences.getValue("layerview/layer_view_type") != 0
                         });
                     }
                 }
 
                 Label {
                     text: label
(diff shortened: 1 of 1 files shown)


plugins/LayerView/LayerView.qml
```diff
@@ -127,6 +127,16 @@ Item
...
+
+                    typesLegenModelNoCheck.clear();
+                    typesLegenModelNoCheck.append({
+                        label: catalog.i18nc("@label", "Top / Bottom"),
+                        colorId:  "layerview_skin"
+                    });
+                    typesLegenModelNoCheck.append({
+                        label: catalog.i18nc("@label", "Inner Wall"),
+                        colorId:  "layerview_inset_x"
+                    });
```
Ghostkeeper: Why is this necessary? The `typesLegenModelNoCheck` is only initialised with these two settings, and never changed by anything afterwards. Clearing and re-appending its items should do nothing, right?

plugins/LayerView/LayerView.qml
```diff
@@ -292,7 +303,7 @@ Item
                         color: UM.Theme.getColor(model.colorId)
                         border.width: UM.Theme.getSize("default_lining").width
                         border.color: UM.Theme.getColor("lining")
-                        visible: view_settings.show_legend
+                        visible: (UM.Preferences.getValue("layerview/layer_view_type") != 0) && view_settings.show_legend
```
Ghostkeeper: This is now exactly the same as the parent, so this line has no effect any more. If the parent label is invisible, this rectangle will be too.

plugins/LayerView/LayerView.qml
Ghostkeeper: This `updateLegend` function is also called before the `componentCompleted`, causing these legend items to appear twice until you update for the first time.
doublelegenditems

LipuFei:
@Ghostkeeper fixed.

PR #1778, opened 2017-05-02
Added VCP serial device support for USB Printing on Windows.

rmcolbert:
This change adjusts the serial port filtering on Windows to also include ports defined as VCP. This is safer than disabling the filter entirely and should future-proof device detection across most 3rd party printers on Windows.


nallath:
There are a few stray visual studio files with your commit. Could you remove those? 

As for testing; I don't have a way to test it, but the change seem pretty safe to me.

rmcolbert:
Oops. Files removed.

Ghostkeeper:
Devs, see CURA-3766.

PR #1783, opened 2017-05-03
Make background color of toggle button themeable

randyxxl:
Background of toggle button gets same color as toolbar surrounding the button.

### DIFF
--- a/resources/qml/Sidebar.qml
+++ b/resources/qml/Sidebar.qml
@@ -348,6 +348,7 @@ Rectangle
 
     Rectangle {
         id: settingsModeSelection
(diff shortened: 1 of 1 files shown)


resources/qml/Sidebar.qml
```diff
@@ -348,6 +348,7 @@ Rectangle
 
     Rectangle {
         id: settingsModeSelection
+		color: UM.Theme.getColor("tool_panel_background")
```
fieldOfView: Code style: use spaces instead of tabs.

fieldOfView:
Good catch.

Alternative approaches: 
* use "transparent" as color
* use an Item instead of a Rectangle

PR #1787, opened 2017-05-04
Add smooth_spiralized_contours setting to control smoothing of contours when spiralizing.

smartavionics:
Smoothing the contours minimises the visibility of the z-seam and for most models is
probably a good idea. Where it is less good is when the model has a fine surface details
which will get mangled by the smoothing. So we should let the user decide whether they
want to smooth or not. The default is true as most of the time it's a good thing.

Relates to https://github.com/Ultimaker/CuraEngine/issues/507.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4565,7 +4565,17 @@
                     "type": "bool",
                     "default_value": false,
                     "settable_per_mesh": false,
-                    "settable_per_extruder": false
+                    "settable_per_extruder": false,
+                    "children":
+                    {
+                        "smooth_spiralized_contours":
+                        {
+                            "label": "Smooth Spiralized Contours",
+                            "description": "Smooth the spiralized contours to reduce the visibility of the Z seam (the Z-seam should be barely visible on the print but will still be visible in the layer view). Note that smoothing will tend to blur fine surface details.",
+                            "type": "bool",
+                            "default_value": true
+                        }
+                    }
                 }
             }
         },


Ghostkeeper:
Devs, see CURA-3771.

jackha:
@smartavionics Thanks! I added "enabled": "magic_spiralize" to the new option so it only shows when enabling spiralize.

smartavionics:
Hi Jack. Yes, that's fine, thanks for incorporating this setting.

resources/definitions/fdmprinter.def.json
```diff
@@ -4565,7 +4565,17 @@
                     "type": "bool",
                     "default_value": false,
                     "settable_per_mesh": false,
-                    "settable_per_extruder": false
+                    "settable_per_extruder": false,
+                    "children":
+                    {
+                        "smooth_spiralized_contours":
```
BagelOrb: This was a problem for command line slicing.

Child settings should control no more than a parent setting.

Parent settings are meant as a user-friendly way to edit the underlying real settings which are actually used by the engine.

This commit made command line slicing think there was no magic_spiralzie setting. The engine only uses the leaf settings of the setting tree.

smartavionics: Well, sorry, I did not understand this requirement. To be honest, I find the whole Cura UI programming & configuration scheme completely baffling. Could some kind person please fix this for me. Thanks, Mark

BagelOrb:
No need to apologize. That's why I explained what the pare t child
relationship means!

I don't know much about the frontend except for how settings are handled.
There's a lot of difficulty in the settings because we are trying to move
to a situation where we can predict the settings of a new material based on
a couple of material properties.

What do you find baffling exactly?


Op 10 mei 2017 22:03 schreef "Mark Burton" <notifications@github.com>:

*@smartavionics* commented on this pull request.
------------------------------

In resources/definitions/fdmprinter.def.json
<https://github.com/Ultimaker/Cura/pull/1787#discussion_r115838601>:

> @@ -4565,7 +4565,17 @@
                     "type": "bool",
                     "default_value": false,
                     "settable_per_mesh": false,
-                    "settable_per_extruder": false
+                    "settable_per_extruder": false,
+                    {
+                        "smooth_spiralized_contours":

Well, sorry, I did not understand this requirement. To be honest, I find
the whole Cura UI programming & configuration scheme completely baffling.
Could some kind person please fix this for me. Thanks, Mark

—
You are receiving this because you commented.
<https://github.com/Ultimaker/Cura/pull/1787#discussion_r115838601>, or mute
the thread

resources/definitions/fdmprinter.def.json
Ghostkeeper: BagelOrb has fixed this since then: https://github.com/Ultimaker/Cura/commit/4abcc8b3f15d7a3d4aa3e0ace2a9b4d8c0bcb46c

PR #1794, opened 2017-05-05
Time estimate per print feature

14bitVoid:
This PR lets cura give a time estimate for each of the print features, like support, infill, skin, etc. That lets you know where the printer spends most of its time, so you can adjust the speeds more effectively. For example you could see that the print takes 9h but 7h of that is skin, then changing the skin speed is better than changing the outer wall speed.

The normal time estimate looks the same but it only lists the details of each print feature in the tooltip of the normal time estimate. I can't make a screenshot because it's a tooltip.

Uranium: https://github.com/Ultimaker/Uranium/pull/247
CuraEngine: https://github.com/Ultimaker/CuraEngine/pull/512

### DIFF
--- a/cura/PrintInformation.py
+++ b/cura/PrintInformation.py
@@ -111,12 +111,14 @@ class PrintInformation(QObject):
     def materialCosts(self):
         return self._material_costs
 
-    def _onPrintDurationMessage(self, total_time, material_amounts):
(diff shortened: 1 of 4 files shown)


Ghostkeeper:
Devs, see CURA-3773.

nallath:
This is a pretty nice feature! Can't imagine why we didn't do this before...

Issue #1777, opened 2017-05-02
USB printing - Not working for Smoothieware firmware


Shaische:
There is an issue with printing via USB when attempting to print with a smoothieware firmware based printer. This issue didn't exist in Cura 1, so it would be good to compare the code to see what changed that is causing this as it has to do with how Cura 2 is communicating with the printer (what it is sending) based on the discussions I had with the smoothieware developers.

The original problem with Cura 2 was not being able to communicate with the printer at all and that was due to it sending empty command lines to the printer. Clicking "Print via USB" did not do anything at all. Jaime had been able to fix this, but not sure if it was merged into 2.5.0. 

After testing a nightly build that had the fix of not sending empty command lines, we were able to start the print. But, it would only get to the point where the hotend heated up, moved down to the bed to start to print and would only deposit maybe about 2mm worth of plastic before the printer froze in place and wouldn't continue further.

If anyone can fix this, we would be very grateful! We can test it for you.

nallath:
I've made a few changes outside of working hours, but as I only have a board without a full printer attached, it's hard to see what's going wrong. 

The code difference between Cura 1 and Cura 2 are pretty big. Also considering that cura 1 used python 2, whereas cura 2 used Python 3. The USB code of cura2 is based on that of the old one, but it's impossible to completely duplicate it.

ChrisTerBeke:
Is this still an issue?

Shaische:
Yes, for smoothieware it is still an issue.

nallath:
As Ultimaker doesn't use smoothieware, i don't think its likely that anyone from Ultimaker is going to spend time on fixing it. If someone from the comunity fixes it, we're more than happy to merge the pull request.

Issue #1434, opened 2017-02-17
Maker Select 3D/Cura 2.3/2.4  "The printer is not connected"


TTVert:
[Logs.zip](https://github.com/Ultimaker/Cura/files/783934/Logs.zip)

win10 64Bit, selected a Prusa i3 but also tried to create a custom printer to no avail, Works fine on Cura 15.0.4.6 and I confirmed I can communicate w/ COM3 at 115200. It seems others have this same issue and I'm not sure where to look. Looking at the log files in USBPrinting don't even mention it looking to any port. It appears this is all auto configured now but I cannot find a log indicating what it is trying to do or where it is failing.

I've attached the two log files I'm aware of that may have something to do with this but any other help would be appreciated.

Thanks
Dave

nallath:
Could you attach the logs?

Ghostkeeper:
It should be saying something like:

    2017-02-17 12:29:51,659 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [308]: Attempting to connect to /dev/ttyACM0
    2017-02-17 12:29:52,277 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [322]: Attempting to connect to printer with serial /dev/ttyACM0 on baud rate 115200
    2017-02-17 12:29:58,798 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [322]: Attempting to connect to printer with serial /dev/ttyACM0 on baud rate 250000
    2017-02-17 12:30:00,311 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [348]: Correct response for auto-baudrate detection received.
    2017-02-17 12:30:00,321 - INFO - USBPrinting.USBPrinterOutputDevice._connect [355]: Established printer connection on port /dev/ttyACM0

What version of Cura are you using? We haven't released any version 3.21.

TTVert:
Sorry is the skirt the dotted line or the purple?  I believe either would be fine.  I think i understand better now after some more reading. It seems you may be able to set a global printer config that will print a skirt around any think you print.  With this particular printer it homes X/Y and then prints a glob, retracts and drags that bit of filament to prevent curling and clean up any residue.  Is this though process correct?
Thanks
Dave

TTVert:
My mistake guys.  I had to zip them as they didn't support uploading the logs directly.  I attached them to the first post.  I certainly didn't see any USB printing lines in the logs when i looked.

[...]

Issue #1321, opened 2017-01-09
2.4 beta USB issues


tjcolejr:
I have 2 printers (one is a printer/cnc) connected via USB to my computer.  Cura does not identify them specifically by either name or com port.  There are just two lines in the drop down menu that say "Print to USB."  I figured the first one listed would be the lower com port number, but I was wrong.
Also, it would be wonderful if Cura did not automatically grab every connected printer upon startup.  If I am printing from an SD card and start Cura on my pc, which is connected via USB, the SD card print gets killed.   I can work around this by leaving Octoprint and Repetier connected to the printer and cnc respectively or just unplugging the USB cables before running Cura, but I'd rather have a checkbox in Cura to connect to USB on startup, so I can decide when I want Cura connected.

fieldOfView:
See https://github.com/Ultimaker/Cura/issues/1049

leutholl:
In USBPrinterOutputDeviceManager.py I changed the tty filter to be /dev/tty.* in order to also accept kits that list their tty als tty.* under /dev

change line 274 so that the else: statement looks like this
        else:
            if only_list_usb:
                base_list = base_list + glob.glob("/dev/tty.*") + glob.glob("/dev/ttyACM*") + glob.glob("/dev/cu.usb*")
                base_list = filter(lambda s: "Bluetooth" not in s, base_list) # Filter because mac sometimes puts them in the list
            else:
                base_list = base_list + glob.glob("/dev/ttyUSB*") + glob.glob("/dev/ttyACM*") + glob.glob("/dev/cu.*") + glob.glob("/dev/tty.usb*") + glob.glob("/dev/rfcomm*") + glob.glob("/dev/serial/by-id/*")
        return list(base_list)

    _instance = None

nallath:
If you have code changes, please make a pull request for it. That way others can also benefit from your changes.

leutholl:
Hi.

Yeah sorry. I'm sorry to say I don't know how all this pull request work... maybe somebody can pick my idea and make it in the next release?!

tjcolejr:
leutholl, thank you for your input.

ianpaschal:
No more work will be done on versions 2.x. If this is still occurring in versions 3.2 or newer, we can re-open this issue.

Ghostkeeper:
For release 2.6 we made the serial port detection use Python's built-in serial module which solved the issue that leutholl's proposed code change would solve as well. The rest should be handled in #1049.

Issue #1271, opened 2016-12-26
Missing driver (2.3.1 and 2.4 beta)


Ando120:
I was unable to connect my UM2+ printer with USB untill I installed 15.04.6
Then it installed some "Arduino... " driver. 
After that 2.3.1 and 2.4 is working too.

sedwards2009:
Which operating system and version are you using?

Ando120:
Windows 8.1 pro 64 bit

thopiekar:
Sounds like https://github.com/Ultimaker/cura-build/pull/96 is needed 😃

Ghostkeeper:
Nope, the drivers have not yet been updated.

ianpaschal:
No more work will be done on versions 2.x. If this is still occurring in versions 3.2 or newer, we can re-open this issue.

Ghostkeeper:
The drivers have been updated for 3.1. See https://github.com/Ultimaker/cura-build/pull/140

PR #1806, opened 2017-05-08
Fix USB printing with Marlin/Sprinter compatible printers

NightmareJoker2:
Fixes serial port enumeration with the latest version of the Windows 10 64-bit FTDI USB serial port adapter driver which will be automatically installed. Has been tested on Windows 10 1703 64-bit, only, using an ATmega2560 based GT2560 board.
Relevant issues: #1244, #1777, #1434, #1321, #736, #1049, #1271

An option to select the COM port or enumerate printer UUIDs and mapping them to printer configurations would probably be ideal (auto-detection breaks with multiple connected printers otherwise). See #1049.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -82,7 +82,7 @@ class USBPrinterOutputDeviceManager(QObject, OutputDevicePlugin, Extension):
 
     def _updateThread(self):
         while self._check_updates:
-            result = self.getSerialPortList(only_list_usb = True)
+            result = self.getSerialPortList(only_list_usb = False)
             self._addRemovePorts(result)
             time.sleep(5)
 


nallath:
I don't really like this fix as it changes how it works for all operating systems. It could be that the filter that is used for windows is wrong, but simply disabling it for everyone is not the way to go.
I think looking at what is being filtered is a far better solution. If you enable only_list_usb, it removes all devices that don't have "VCP" or "USBSER" in the result (which as far as I can see should work). What devices should according to you be added that would be filtered out otherwise?

fieldOfView:
Why are we not using ```serial.tools.list_ports.comports()``` provided by pySerial instead of our own (broken?) way of listing ports?

This seems to work fairly well for me:
https://github.com/fieldOfView/Cura/commit/4fad2a790bc466215dd459fd581e9253e3336ebc#diff-324e2c1917c374039e685b8ea6f44db3R254
@dorianmuthig, could you test the above approach with your printer?

nallath:
Honestly, i don't know. I think at some point i copied this from old Cura. If your patch works, we probably should just change it.

dorianmuthig:
The patch from @fieldOfView sort of works, but only if the printer serial port is the first one enumerated. I am not yet sure why, and I had to modify the code a bit to get COM3 as the first one in the array.
Generally though, I am in favor of using the enumerator [pyserial/pySerial](/pyserial/pyserial) already provides, instead of a poor man's port enumerator.

```
2017-05-09 04:51:43,539 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [308]: Attempting to connect to COM3
2017-05-09 04:51:43,589 - WARNING - UM.OutputDevice.OutputDeviceManager.removeOutputDevice [125]: Could not find output device with id COM3 to remove
2017-05-09 04:51:44,318 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [322]: Attempting to connect to printer with serial COM3 on baud rate 115200
2017-05-09 04:51:51,375 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [322]: Attempting to connect to printer with serial COM3 on baud rate 250000
2017-05-09 04:51:53,188 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [348]: Correct response for auto-baudrate detection received.
2017-05-09 04:51:53,210 - DEBUG - USBPrinting.USBPrinterOutputDevice._connect [348]: Correct response for auto-baudrate detection received.
2017-05-09 04:51:53,226 - INFO - USBPrinting.USBPrinterOutputDevice._connect [355]: Established printer connection on port COM3
2017-05-09 04:51:53,228 - INFO - USBPrinting.USBPrinterOutputDevice._listen [477]: Printer connection listen thread started for COM3
```

Filtering COM ports by whether they're attached via USB or not is ill-advised, since you can also attach an Arduino/Genuino compatible device via a real serial port (and by that virtue, a 3D printer). Granted, these days, people tend not to have one exposed to actually plug something into, but conversely various types of Bluetooth devices, cellular adapters and modems also expose USB serial port adapter devices, which you would fail to filter out on all platforms with this approach just the same. So, since you can't successfully check anyway, why bother?
Instead, enumerate all serial port devices and prompt them the question "Are you a 3D printer?" by sending it G-code `M115` and evaluating the response, which should be something like this:

```
FIRMWARE_NAME:Marlin 1.1.0 (Github)
SOURCE_CODE_URL:https://github.com/MarlinFirmware/Marlin
PROTOCOL_VERSION:1.0
MACHINE_TYPE:RepRap
EXTRUDER_COUNT:1
UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
```
or this:
```
FIRMWARE_NAME:Marlin 1.1.0 (Github)
SOURCE_CODE_URL:https://github.com/MarlinFirmware/Marlin
PROTOCOL_VERSION:1.0
MACHINE_TYPE:RepRap
EXTRUDER_COUNT:1
UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
Cap:EEPROM:1
Cap:AUTOREPORT_TEMP:1
Cap:PROGRESS:0
Cap:AUTOLEVEL:1
Cap:Z_PROBE:1
Cap:SOFTWARE_POWER:0
Cap:TOGGLE_LIGHTS:0
Cap:EMERGENCY_PARSER:1
```
(taken from http://marlinfw.org/docs/gcode/M115.html)

For the time being, disabling the check is perfectly reasonable, doesn't break anything, and involves the least amount of effort to get what you want.

fieldOfView:
Note that not every firmware supports M115. Smoothieware does not, for example.

One reason not to include all serial ports by default is that it would result in Cura opening all serial ports. We know that eg all (or most) Arduino-like devices reset when opening a port. It would also take longer to try all ports at all baudrates, with timeouts.

In #1688, which is based on #1049, I have added an option to include non-usb devices, so it will be up to the user.

fieldOfView:
> The patch from @fieldOfView sort of works, but only if the printer serial port is the first one enumerated. I am not yet sure why, and I had to modify the code a bit to get COM3 as the first one in the array.

Can you share a bit more information? What happens when you don't change the order? What change did you make?

dorianmuthig:
> Note that not every firmware supports M115. Smoothieware does not, for example.

Fully aware. Neither does GRBL. Though, if you want to do automatic detection, this is the only reliable method. Unless you allow users to select the port to use, which currently, you don't, you need to query all of the serial devices to determine what is connected, not arbitrarily decide the device you are looking for can't be connected to certain types of serial connectors, because you can't actually know that.

> We know that eg all (or most) Arduino-like devices reset when opening a port.

That is sort of annoying (especially with pre-heated printers which then cool down again), but considering you can only open a serial port once and need to close it or the application which opened it before being able to open it again, this shouldn't be an issue, if you're worried about resetting a device which is in use. I also fail to see how filtering devices in the manner you described would do anything about it.

> > The patch from @fieldOfView sort of works, but only if the printer serial port is the first one enumerated. I am not yet sure why, and I had to modify the code a bit to get COM3 as the first one in the array.
>
> Can you share a bit more information? What happens when you don't change the order? What change did you make? What port is "ïn front of" COM3 without your changes?

I have a real serial port as a header on my PC motherboard, which gets enumerated as COM1. Apparently the library tries to open it and then gets stuck, presumably, because nothing is attached to it. Basically I added two lines to skip `COM1`.
Of course, connecting two [MarlinFirmware/Marlin](/MarlinFirmware/Marlin) based printers never worked, and you only get to use the first one that is found.

Sineos:
Just my two cents: I perceive Curas habit of grabbing the COM port as highly annoying. Makes quick switches to pronterface or similar tools impossible without shutting down Cura.
I personally would prefer a manual Com Port selection and a manual Connect / Disconnect Button.

Also an error message when the Com port is unavailable (Windows can be very creative here) would be nice. This would avoid having pronterface open and wondering why the hell Cura does not connect to the printer.

ChrisTerBeke:
We're closing this PR for now as there is no consensus on the 'correct' implementation.

Ghostkeeper:
The PR is no longer applicable either since we switched to Python's serial tools rather than detecting COM ports ourselves.

PR #1816, opened 2017-05-11
Extend the zoomrange to cater for smaller and larger buildvolumes

fieldOfView:
This PR extends the zoomrange of the cameratool to better support printers with a buildvolume that is much larger or much smaller than the originally intended (UMO) printer.

See eg https://ultimaker.com/en/community/49326-cura-needs-to-zoom-out-further

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -619,7 +619,9 @@ class CuraApplication(QtApplication):
         camera.lookAt(Vector(0, 0, 0))
         controller.getScene().setActiveCamera("3d")
 
-        self.getController().getTool("CameraTool").setOrigin(Vector(0, 100, 0))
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3802.

fieldOfView:
Dang, we have to create 200 more issues before Cura 2.6 can be released ;-)

Ghostkeeper:
Oh why are you closing this? Trying to reach that 4000 number? :stuck_out_tongue:

fieldOfView:
Because I pressed the wrong button

Ghostkeeper:
It's perhaps slightly... extreme. But I'm going to merge this anyway, see what others think of it. We can always adjust it a bit.

PR #1817, opened 2017-05-11
Handle limit_to_extruder transparently

awhiemstra:
This adds logic to Extruder and Global stack's getProperty to handle the limit_to_extruder property if it is set. This works in a similar manner as the resolve property. This allows us to remove a bunch of calls to extruderValue(extruder, setting), which this PR also includes.

The main thing that remains to be done for this PR is that the unit test failures need looking in to, since the added checks for limit_to_extruder cause the unit tests to return unexpected data.

Note that there currently is an unrelated bug where setting the support extruder will not properly update the visual state of the extruders of its children. The actual underlying data _is_ properly updated however.


nallath:
Changes look fine. I will have a look at the unit tests.

awhiemstra:
Fixed the infinite recursion that could occur with settings that have limit_to_extruder set but not settable_per_extruder. This should now be ready to merge.

PR #1821, opened 2017-05-12
Just like magic_spiralize, smooth_spiralized_contours should not be settable per model/extruder.

smartavionics:
As simple as that.


Ghostkeeper:
I read the title of this pull request as starting with "Just like magic..." and was very excited.

BagelOrb:
Yup!
See https://github.com/Ultimaker/CuraEngine/blob/master/src/LayerPlan.cpp#L499

Filed this under the existing ticket CURA-3771

Issue #1832, opened 2017-05-16
negative values for extruder offset


maukcc:
You can not enter negative values for extruder offset in machine settings.
When you enter negative values manually in definition_changes, it will show them, but it will not use them.

fieldOfView:
Good catch. I'll make a PR.

maukcc:
Sorry, my brain was going in the wrong direction and not the machine.
It does use the negative values when you enter them manually.

PR #1833, opened 2017-05-16
Allow setting negative values on extruder offset x/y fields

fieldOfView:
This PR lets users enter negative values for extruder nozzle x/y offsets.

Fixes https://github.com/Ultimaker/Cura/issues/1832


Ghostkeeper:
Devs, see CURA-3825.

PR #1839, opened 2017-05-17
Fix create material for printers that use 1.75mm materials

fieldOfView:
This PR is a followup to #1757. It fixes the problem that if the user has a 1.75mm printer active, the Create button would clone the 2.85mm generic PLA which would then be invisible for the current machine.

The fix is to clone the generic PLA material for the current approximate material diameter instead.

### DIFF
--- a/cura/Settings/ContainerManager.py
+++ b/cura/Settings/ContainerManager.py
@@ -712,14 +712,14 @@ class ContainerManager(QObject):
         if not global_stack:
             return ""
 
-        approximate_diameter = round(global_stack.getProperty("material_diameter","value"))
-        containers = self._container_registry.findInstanceContainers(id="generic_pla*", approximate_diameter=approximate_diameter)
+        approximate_diameter = round(global_stack.getProperty("material_diameter", "value"))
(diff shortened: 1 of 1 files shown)


nallath:
CURA-3828

fieldOfView:
Don't worry, I'll get you to the CURA-4000 for the 2.6 release...

cura/Settings/ContainerManager.py
```diff
@@ -712,14 +712,21 @@ def createMaterial(self) -> str:
         if not global_stack:
             return ""
 
-        containers = self._container_registry.findInstanceContainers(id="generic_pla")
+        approximate_diameter = round(global_stack.getProperty("material_diameter","value"))
+        containers = self._container_registry.findInstanceContainers(id="generic_pla*", approximate_diameter=approximate_diameter)
```
Ghostkeeper: A few code style problems here. I fixed them for you in f65043ecc8f7b506464b31ba5995fac58b4d0358.

jackha:
Ahhh I missed the boat. In the process of reviewing and testing I heard my name a few times and now it's merged!

Ghostkeeper:
:boat:

Issue #1840, opened 2017-05-17
Conflicts between printer name/id and definition ids


fieldOfView:
When adding a new printer with the exact id of a machine definition, the printer can not be succesfully deserialised when restarting Cura.

~~This is probably a regression introduced by the CuraContainerStack work.~~
This reproduces in Cura 2.5 too.

Steps to reproduce:
* add a printer named "ultimaker3" (this exact spelling)
* close and restart Cura

Result:
The added "ultimaker3" printer is no longer listed. 
From the logs:
```
2017-05-17 13:39:26,655 - CRITICAL - UM.Settings.ContainerRegistry.load [175]: Found a container with a duplicate ID: ultimaker3
2017-05-17 13:39:26,657 - CRITICAL - UM.Settings.ContainerRegistry.load [176]: Existing container is <UM.Settings.DefinitionContainer.DefinitionContainer object at 0x0000026BBDFADA68>, trying to load <class 'cura.Settings.GlobalStack.GlobalStack'> from C:\Users\Aldo\AppData\Roaming\cura\master\machine_instances\ultimaker3.global.cfg
```

This became apparent when I added a 101Hero printer, which has the same name and id (whereas most other definitions have a different spelling between name and id, making this issue less apparent).

nallath:
I just tested this and it was already an issue with 2.5
I don't think we ever checked for globally unique names...

fieldOfView:
Shows you how many users use the 101Hero definition ;-)

> I don't think we ever checked for globally unique names
We really should.

Ghostkeeper:
Could reproduce this. It seems to be case sensitive (at least on Ubuntu).

fieldOfView:
Yes, it is case-sensitive on Windows too.

fieldOfView:
Same issue when naming a printer after a quality profile id (eg "draft").

PR #1843, opened 2017-05-18
Fix creating conflicting machine names

fieldOfView:
This PR ensures the name (id) of new printer instances does not conflict with definition ids or container instance ids.

Steps to reproduce:
* add a printer named "ultimaker3" (this exact spelling)
* add a printer named "draft" (this exact spelling)
* close and restart Cura

Result:
The added printers are no longer listed.

Fixes https://github.com/Ultimaker/Cura/issues/1840


Ghostkeeper:
Devs, see CURA-3833.

nallath:
As usual; thanks for the fix!

PR #1847, opened 2017-05-19
quality names matching file names

maukcc:


fieldOfView:
Hmm, this commit should be reverted: https://github.com/maukcc/Cura/commit/8a027340c55f9fc7c6137662e5eeedde0162cb77

maukcc:
I have no idea how to do that.

Ghostkeeper:
Also the first merge seems to have gone wrong. The UM3 AA0.8 TPU profiles are re-added by this pull request.

We renamed all profiles recently because of the trouble we were having on the low end with naming them. It was also worked through for other printers. I think we can let you rename it back, but I'll ask.

Ghostkeeper:
You copy the commit hash (a unique identifier for the commit) from Github and revert it on the command line like so:

    git revert 8a027340c55f9fc7c6137662e5eeedde0162cb77

maukcc:
Thanks for the info!

I do not mind to have the same quality names, but they need to be consistent.
at present one machine has "fine" for 0.1 and the other has "fine" for 0.2 layerheight.

Ghostkeeper:
Devs, see CURA-3842.

PR #1851, opened 2017-05-20
Wireframe view

BagelOrb:
A new view which only shows the edges and shows overhang on the edges.

I haven't gotten edge-overhang to be calculated, because each edge needs to know about both adjoining faces, which is not supported in the RenderBatch, I think.

Instead the edges are colored according to the normal face-overhang :/

The current implementation is only supported by version 4.1...

### DIFF
--- a/plugins/WireframeView/WireframeView.py
+++ b/plugins/WireframeView/WireframeView.py
@@ -6,6 +6,7 @@ import os.path
 from UM.Application import Application
 from UM.Math.Color import Color
 from UM.PluginRegistry import PluginRegistry
+from UM.Preferences import Preferences
 from UM.Event import Event
 from UM.View.View import View
 from UM.View.RenderBatch import RenderBatch
@@ -13,6 +14,8 @@ from UM.Scene.Iterator.BreadthFirstIterator import BreadthFirstIterator
 
 from UM.View.RenderBatch import RenderBatch
 from UM.View.GL.OpenGL import OpenGL
+
+import math
 
 
 ## View used to display the edges of objects.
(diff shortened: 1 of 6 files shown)


fieldOfView:
I don't mean to alarm you, but it looks like you're writing Python

thopiekar:
:fearful:

BagelOrb:
Nah it's just some scripting to get the shader c++ code to work

xD

Op 20 mei 2017 14:53 schreef "Thomas Karl Pietrowski" <
notifications@github.com>:

plugins/WireframeView/wireframe.shader
```diff
@@ -0,0 +1,91 @@
...
+    {
+        gl_Position = u_model_view_projection_matrix * a_vertex;
+    }
+
+fragment =
+    uniform lowp vec4 u_color;
+
+    void main()
+    {
+        gl_FragColor = u_color;
```
Ghostkeeper: This gives a silhouette of the object if the user has no OpenGL 4 support.

Ghostkeeper:
Devs, see CURA-3849.

plugins/WireframeView/wireframe.shader
BagelOrb: Yeah without geometry shaders the wireframe view is not possible unless you do computations on the CPU. 

Is there a way to disable the view if the GPU doesn't suport 4.1?

fieldOfView: See OpenGLContext.detectBestOpenGLVersion()
I think the best way would be to use that to check for the right OpenGL version in the __init__.py register() method and return an empty dict when the gl version is too low.

awhiemstra:
There is actually a very simple trick to do the same on all version of GL (and not just 4.1+): https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPolygonMode.xhtml

Also, we used to have a wireframe view back in the day, but at some point we removed it since it did not have a lot of purpose and it did not work anymore. I still see little reason to have it personally.

BagelOrb:
I was just trying some stuff out. If you think it shouldn't be included
than we can just leave it out. Maybe it's a nice one for the online plugin
repository. Nobody likes a sparsely populated plugin repository ;)

Op 22 mei 2017 16:59 schreef "awhiemstra" <notifications@github.com>:

LipuFei:
Expect for the OpenGL discussion, the rest looks fine. I have also added a legend panel for the WireframeView. **I have checked that the colors are correct, but they somehow look different on the legend and the build plate. @Ghostkeeper and I think it could be the lighting.**

image

plugins/WireframeView/WireframeView.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self._wireframe_shader = None
+
+    def beginRendering(self):
+        scene = self.getController().getScene()
+        renderer = self.getRenderer()
+
+        if not self._wireframe_shader:
+            self._wireframe_shader = OpenGL.getInstance().createShaderProgram(os.path.join(PluginRegistry.getInstance().getPluginPath("WireframeView"), "wireframe.shader"))
+            self._wireframe_shader.setUniformValue("u_color", Color(*Application.getInstance().getTheme().getColor("xray").getRgbF()))
```
fieldOfView: This is not very semantically correct, is it? This has nothing to do with "xray".

nallath:
I think we should keep the plugin seperate; It doesn't add any interesting or particularly usefull features. This is a perfect candidate once we have the plugin repo / online thing up & running.

plugins/WireframeView/WireframeView.py
LipuFei: I will create a separate set of colours for the Wireframe :)

LipuFei:
@fieldOfView @BagelOrb says he doesn't know what this view is for :P

PR #1863, opened 2017-05-23
2.5 nozzle size

Aurarum:
Hi, is there any chance to get the nozzle back into the settings?
it is not really convenient to recreate the printer json in order to change the nozzle size.


fieldOfView:
See "Machine Settings" in the "Printers" pane of the preferences

Aurarum:
am I missing something? we are talking about 2.5, right?
there is only configure Cura under preferences

fieldOfView:
In the preferences window, go to the Printers pane/page. Or from the printer dropdown in the sidebar, choose Manage Printers...

Aurarum:
where?
image

fieldOfView:
Select the active printer

fieldOfView:
Aw, wait. The active printer is a dual extrusion printer, and in 2.5 Machine Settings does not work with dual extrusion printers. If does in Cura 2.6

Aurarum:
oh ok. is there a Windows compiled version to download somewhere?

fieldOfView:
No, not yet

Aurarum:
thanks for that. so I would have to compile myself? right?

fieldOfView:
Well, it is python, it does not strictly need to be compiled. But yeah, you would need to download the source repositories and all the dependencies, and get it running from there. Not a trivial thing.

Aurarum:
thanks for your time. I haven't done this before thus the questions

awhiemstra:
Note that the 2.6 beta was just released: https://ultimaker.com/en/blog/49492-announcing-cura-26-beta
That should contain the new multi-extruder machine settings.

PR #1866, opened 2017-05-23
added 3Dator printer profile and printbed mesh

moritzmair:
Added json and stl for the 3Dator printer.


nallath:
Devs see CURA-3863

BagelOrb:
`"adhesion_type": { "default_value": "none" },` seems a bit odd. Are you sure your nozzle is primed correctly before starting to print?

This makes the Build Plate Adhesion setting in recommended settings not respond to user input.

moritzmair:
thanks for looking into this.
We wrote a custom start GCode that will prime the nozzle in a line on the printbed. That works much better then having a skirt.

BagelOrb:
Do note that the Build Plate Adhesion setting in recommended mode doesn't change if the user tries to check that setting.

I will merge this PR knowing that this problem exists...

theBohnanza:
Hi, i was expecting the Printer Profile to be seen in the new Cura Version. Is there still something, that has to be done?

fieldOfView:
The files were added to the "master" branch, after the branch for Cura 2.6 had been created. The printer profile will come with Cura 2.7

theBohnanza:
Alright. Thank you very much for the quick reply.

PR #1876, opened 2017-05-24
CURA-3810: Add material info button to show Ultimaker.com materials page

LipuFei:
Add a new info button for materials. When a user clicks on the button, it will open the materials information page on the Ultimaker website.


fieldOfView:
What about materials that Ultimaker has no information on?

Ghostkeeper:
This basically adds a button which is always visible. The page on ultimaker.com displays information on the material types, such as TPU, PLA, CPE. So this page won't give information about material types such as PETG which we don't test with. Is that important, you think?

fieldOfView:
Yes, I think it is important to consider that there are many, many more materials that people print with than the set that Ultimaker sells. Think of all the other flexibles, woodfills, metalfills, etc.

I think the fdmmaterial.xml spec needs to be updated with a url field for information, and the button should only show for materials that have a url.

resources/qml/SidebarHeader.qml
```diff
@@ -196,16 +196,63 @@ Column
...
+        {
+            width: parent.width * 0.05
+        }
+
+        Button
+        {
+            id: materialInfoButton
+            height: parent.height * 0.70
+            width: height
+            anchors.leftMargin: UM.Theme.getSize("default_margin").width
```
nallath: For the next time you can just set all anchors with; anchors.margins: UM.Theme.getSize("default_margin").width

fieldOfView:
This is making less sense by the commit.
"Only show material info button for multi-extrusion machines"
Wut? How is multi or single extrusion related to wanting to know about materials?

"take the space of the material info button when the active machine doesn't have multiple extruders"

So you're not showing something (why?), but take up space for it anyway? Like the sidebar is not tight for space enough as it is?

Note that my earlier comment still stands: there are LOTS of materials that Ultimaker has no information about.

Ghostkeeper:
> How is multi or single extrusion related to wanting to know about materials?

The page is about compatibility of certain combinations of materials, not general info about materials.

BagelOrb:
This is about how multiple materials work together in a multi-extruder machine. The pointed to web page lists which materials are verified to work with each other (for the UM3 only atm).

resources/qml/SidebarHeader.qml
```diff
@@ -243,6 +244,13 @@ Column
             }
         }
 
+        Text  // to take the space of the material info button when the active machine doesn't have multiple extruders
```
nallath: A single extrusion machine should also show it guess. It can have materials right?

Ghostkeeper: As explained on the main page of this pull request, the page is about which combinations of materials are viable, not about generic material information.

cura/CuraApplication.py
```diff
@@ -1362,3 +1362,7 @@ def _onContextMenuRequested(self, x: float, y: float) -> None:
                     node = node.getParent()
 
                 Selection.add(node)
+
+    @pyqtSlot(result=str)
+    def getCuraVersion(self):
+        return CuraVersion
```
Ghostkeeper: This is unnecessary. Take a look at how the version is displayed here: https://github.com/Ultimaker/Cura/blob/master/resources/qml/AboutDialog.qml#L43

resources/qml/SidebarHeader.qml
```diff
@@ -220,7 +220,10 @@ Column
             onClicked:
             {
                 // open the material URL with web browser
-                var url = "https://ultimaker.com/en/resources/23121-materials";
+                var version = CuraApplication.getCuraVersion();
+                var machineName = Cura.MachineManager.activeMachineDefinitionId;
+
+                var url = "https://ultimaker.com/materialcompatibility/" + version + "/" + machineName;
```
Ghostkeeper: This is unnecessary. Take a look at how the version is displayed here: https://github.com/Ultimaker/Cura/blob/master/resources/qml/AboutDialog.qml#L43

cura/Settings/MachineManager.py
```diff
@@ -474,6 +474,12 @@ def activeMachineId(self) -> str:
     def activeMachine(self) -> "GlobalStack":
         return self._global_container_stack
 
+    @pyqtProperty(str, notify = globalContainerChanged)
+    def activeMachineDefinitionId(self) -> str:
+        if self._global_container_stack and self._global_container_stack.definition:
+            return self._global_container_stack.definition.getId()
+        return ""
```
Ghostkeeper: Unnecessary. From QML you can obtain `Cura.MachineManager.activeMachine.definition.id`.

resources/themes/cura/theme.json
```diff
@@ -243,7 +243,12 @@
         "layerview_support_infill": [0,  255,  255, 255],
         "layerview_move_combing": [0,  0,  255, 255],
         "layerview_move_retraction": [128,  128,  255, 255],
-        "layerview_support_interface": [64, 192, 255, 255]
+        "layerview_support_interface": [64, 192, 255, 255],
+
+        "info_button_text": [255, 255, 255, 255],
+        "info_button_background": [127, 127, 127, 255],
+        "info_button_background_hover": [24, 41, 77, 255],
+        "info_button_background_active": [70, 84, 113, 255]
```
Ghostkeeper: Shouldn't this just always use the style of the normal buttons?

resources/qml/SidebarHeader.qml
```diff
@@ -196,16 +196,71 @@ Column
             }
 
             anchors.verticalCenter: parent.verticalCenter
-            width: parent.width * 0.45 - UM.Theme.getSize("default_margin").width
+            width: parent.width * 0.30
```
Ghostkeeper: I agree that it's silly to make space for a button that's not displayed on single-extrusion printers.

cura/Settings/MachineManager.py
LipuFei: I had the problem of accessing `MachineManager.activeMachine` when I used this:
```
QMetaProperty::read: Unable to handle unregistered datatype 'QObject' for property 'MachineManager::activeMachine'
file:///Cura/resources/qml/SidebarHeader.qml:225: TypeError: Cannot read property 'definition' of undefined
```

LipuFei:
@Ghostkeeper fixed.

cura/Settings/MachineManager.py
nallath: So in that case check if there is an active machine at all. If not, there is no id.

awhiemstra: The error is actually because we specified an incorrect type for the activeMachine property, it should be "QObject*" instead of "QObject".

awhiemstra:
Why is this planned to merge to 2.6? Isn't that feature frozen already?

LipuFei:
@Ghostkeeper Updated. This one requires this commit in Uranium: https://github.com/Ultimaker/Uranium/commit/b0834e9800a78fa831cfd613fa733437adea0b3e

Ghostkeeper:
Yes, 2.6 is feature frozen but this was overruled by higher up.

PR #1878, opened 2017-05-24
Get a list of available themes instead of the hardcoded list

fieldOfView:
This PR builds on https://github.com/Ultimaker/Uranium/pull/252 to show a list of available plugins instead of the current "list" that is hardcoded in GeneralPage.qml


Ghostkeeper:
Devs, see CURA-3873.

Issue #1273, opened 2016-12-27
Material Setting for Filament Diameter not recognized


1-prop-head:
Ultimaker 2 Ex (not plus),Original Filament Feeder,1.75mm ABS filament,Cura 2.4 Beta
Printer filament setting is ABS 1.75
Printer is as above.  When printing I need to set the Meterial Flow to 55% to be able to not grosely over extruding the filament

Ghostkeeper:
Hmm, you'd expect the flow to be reduced if no compensation were applied. Perhaps we're compensating double?

1-prop-head:
This is not closed in my opinion.  I Just loaded 2.5.0 (latest firmware 2.4ex) and with my UM2ex I still need to change material flow from 100% to 22% to not over extrude 1.75 abs filament. This was not the case with the firmware provided with the old version of cure (15...) which extrudes correctly at the default 100%

Ghostkeeper:
This was fixed via https://github.com/Ultimaker/Cura/pull/1890 in Cura 2.6, but it doesn't show as fixed because the commits were cherry-picked into the release.

1-prop-head:
What was the date on this and no it is not fixed 

Fixed or not in CURA 3.04 and all previous iterations I must use a material speed adjustment in the printer of 57% to not over extrude.

Ghostkeeper:
I think I misunderstood your initial question, sorry. I was thinking about the UM2E+ which has material profiles. The fix allowed the diameter in the material manager to be used for slicing. The UM2E has no material profiles so the fix in question did not apply.

Cura has no knowledge of what material is in that printer, be it 1.75mm or 2.85mm. It'll assume 2.85mm then, since that is the only option in the firmware. If you're slicing for a different diameter than the firmware thinks it has, I'd say that Flow is a fine setting to adjust. Since the E parameters for UltiGCode indicates filament length, not volume, you should be changing the material flow to 265% (=2.85^2 / 1.75^2) in theory. Why you had to compensate in the other direction I don't know.

If you think that the firmware needs to be adjusted to allow for 1.75mm diameter materials, you're looking for the repository of the [firmware of the UM2](https://github.com/Ultimaker/Ultimaker2Marlin), I'd say?

fieldOfView:
There should not be a material diameter field in Cura at all if you are using UltiGcode with the Ultimaker 2, because in that case Cura should be material-agnostic. 

[...]

PR #1890, opened 2017-05-26
Fix loading material diameter and GUID from XML file

### LINKED ISSUE #1886: 2.6.0beta Materials issues
Materials settings doesn't override some profile settings.
(works with inbuilt profiles, doesn't work with custom profiles)

I've created custom material and when selected -
- it doesn't override - hotted temp, filament diameter, retraction distance
- does override - bed temp, retraction speed, fan speed
- unable to test standby temp hence not configurable in print settings

Also regardless from the profile (custom or inbuilt) -
- in layer view mode with material colour selected it doesn't change preview to one in the Material config
- doesn't calculate cost of the print

OS - OSX 10.12.5 (clean install)

fieldOfView:
This PR fixes loading the material diameter from xml material files. Diameters were previously only shown, but not used as values.

Fixes #1886 

While I was at it, I also added loading the GUID into the material_guid setting. This setting would never be filled in.

### DIFF
--- a/plugins/XmlMaterialProfile/XmlMaterialProfile.py
+++ b/plugins/XmlMaterialProfile/XmlMaterialProfile.py
@@ -439,6 +439,8 @@ class XmlMaterialProfile(InstanceContainer):
         meta_data["base_file"] = self.id
         meta_data["status"] = "unknown"  # TODO: Add material verfication
 
+        global_setting_values = {}
+
         inherits = data.find("./um:inherits", self.__namespaces)
         if inherits is not None:
             inherited = self._resolveInheritance(inherits.text)
@@ -468,6 +470,9 @@ class XmlMaterialProfile(InstanceContainer):
                 continue
             meta_data[tag_name] = entry.text
 
+            if tag_name in self.__material_metadata_setting_map:
+                global_setting_values[self.__material_metadata_setting_map[tag_name]] = entry.text
+
         if "description" not in meta_data:
             meta_data["description"] = ""
 
@@ -480,18 +485,20 @@ class XmlMaterialProfile(InstanceContainer):
             tag_name = _tag_without_namespace(entry)
             property_values[tag_name] = entry.text
 
+            if tag_name in self.__material_properties_setting_map:
+                global_setting_values[self.__material_properties_setting_map[tag_name]] = entry.text
+
         meta_data["approximate_diameter"] = round(float(property_values.get("diameter", 2.85))) # In mm
         meta_data["properties"] = property_values
 
         self.setDefinition(ContainerRegistry.getInstance().findDefinitionContainers(id = "fdmprinter")[0])
 
         global_compatibility = True
-        global_setting_values = {}
         settings = data.iterfind("./um:settings/um:setting", self.__namespaces)
         for entry in settings:
             key = entry.get("key")
-            if key in self.__material_property_setting_map:
-                global_setting_values[self.__material_property_setting_map[key]] = entry.text
+            if key in self.__material_settings_setting_map:
+                global_setting_values[self.__material_settings_setting_map[key]] = entry.text
             elif key in self.__unmapped_settings:
                 if key == "hardware compatible":
                     global_compatibility = parseBool(entry.text)
@@ -510,8 +517,8 @@ class XmlMaterialProfile(InstanceContainer):
             settings = machine.iterfind("./um:setting", self.__namespaces)
             for entry in settings:
                 key = entry.get("key")
-                if key in self.__material_property_setting_map:
-                    machine_setting_values[self.__material_property_setting_map[key]] = entry.text
+                if key in self.__material_settings_setting_map:
+                    machine_setting_values[self.__material_settings_setting_map[key]] = entry.text
                 elif key in self.__unmapped_settings:
                     if key == "hardware compatible":
                         machine_compatibility = parseBool(entry.text)
@@ -573,8 +580,8 @@ class XmlMaterialProfile(InstanceContainer):
                     settings = hotend.iterfind("./um:setting", self.__namespaces)
                     for entry in settings:
                         key = entry.get("key")
-                        if key in self.__material_property_setting_map:
-                            hotend_setting_values[self.__material_property_setting_map[key]] = entry.text
+                        if key in self.__material_settings_setting_map:
+                            hotend_setting_values[self.__material_settings_setting_map[key]] = entry.text
                         elif key in self.__unmapped_settings:
                             if key == "hardware compatible":
                                 hotend_compatibility = parseBool(entry.text)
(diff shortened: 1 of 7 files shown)


fieldOfView:
PR is against master, but this fix may have to be cherry-picked to 2.6 since it fixes a long-standing bug and 2.6 has a lot of material-related fixes in it so it fits with the "theme".

Ghostkeeper:
Devs, see CURA-2822.

Ghostkeeper:
CURA-2822 had a link to this Github issue as well: https://github.com/Ultimaker/Cura/issues/1308
I think we have some duplicates around.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -439,6 +439,8 @@ def deserialize(self, serialized):
         meta_data["base_file"] = self.id
         meta_data["status"] = "unknown"  # TODO: Add material verfication
 
+        global_setting_values = {}
```
BagelOrb: Global?! You should be changing only the settings of the extruder in which this material is loaded!

This is either a bug or a rather unfortunate name for a variable.... Or is it?

LipuFei: Unfortunately this was the name we used... See line 489.

fieldOfView: Note that I moved this line of code from line 489, I did not introduce this variable name.

fieldOfView: "global" in this context is "global for all the containers that get created from a single xml file", not "global for all extrudertrains".

LipuFei: I think the "global" here means the settings in the "settings" section in the XML material file, which can applied to any machine. The settings in "settings/machine" sections are machine specific settings.

BagelOrb: Hmmmm interesting..... ^^

nallath: Global in this context should indeed be seen as "general"

fieldOfView:
Also see https://github.com/Ultimaker/Cura/issues/1273

plugins/XmlMaterialProfile/XmlMaterialProfile.py
LipuFei: The rest looks fine to me. @BagelOrb Do you still prefer to have this variable renamed?

BagelOrb: Words like global, general, all, overall fail to convey enough meaning. They all convey that it is unbounded in the field of subject without saying what the field of subject is. So in contrast to printer specific settings these global settings are not bounded to any application of the material.

The problem is that variable names should be self explanatory, while 'global' only has meaning wrt 'printer specific'. 

I don't know. Maybe this is just a philosophical excerpt.

fieldOfView: I have changed the word "global" into "common". While this does not fix the entire philosophical issue of not specifying the context of the word global/general/common, it does take away the confusion over the term "global" which has such a strong connotation in Cura.

BagelOrb:
Do I need to clean the cache or something?
I cannot get this to work...

BagelOrb:
I'm testing like so:
- open material manager
- create new material
- change density and diameter
- close cura
- start cura
- check new material in the manager

plugins/XmlMaterialProfile/XmlMaterialProfile.py
BagelOrb: Fair nuf

BagelOrb:
With the above test case failing this PR doesn't really add much value... At least before this commit it's clear that it doesn't work. With this PR the user has no idea the value reset when he reopened Cura....

Please fix the above issue? Pretty Please? :pray:

BagelOrb:
Also we have decided to merge this to 2.6, so either you or we will cherry-pick the commits into 2.6.

fieldOfView:
> At least before this commit it's clear that it doesn't work.

Can't argue with that reasoning...  :weary:

It turns out setting metadata for xmlmaterialprofiles was broken, which meant that changes did not get saved. I have also fixed the issue that changing the diameter in the material view did not also change it in the settings.

This PR now fixes deserialising, changing and serialising xmlmaterialprofiles.

BagelOrb:
I think I have disappearing custom materials with the new commits you made...

fieldOfView:
Repro steps?

Note that if you change a 2.85 mm material to a 1.75 mm material, it it disappears because it is no longer suitable for the current machine.

BagelOrb:
Could not deserialize any def.json file: 
`json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

BagelOrb:
Never mind. I just ffed up :O

BagelOrb:
Why is 1.75 mm material incompatible by default?

Does that hold for any machine?

fieldOfView:
1.75mm was just an example

BagelOrb:
I get that, but why is a material with a new diameter incompatible by default?

If I want to make a profile for the 1.2342535 mm filament I just bought I want to create a new profile and not have it disappear. Why would it disappear?

fieldOfView:
I fixed the case where the user changes the diameter of the currently active material. This now properly refreshes the list of materials to a list of materials with the same approximate diameter. This is the logical cause of the material diameter filtering that was introduced before.

Material management was very broken. With this PR it is much less broken. Remaining usability problems should (IMHO) be logged as a new issue so they can be fixed later.

> Why would it disappear?

Because it is not compatible with your current printer. You bought the wrong material.

BagelOrb:
Hah! Awesome!

Ghostkeeper:
Still some weird update bugs going on. When creating a new material and setting the diameter to 1.75mm (while using UM3) it didn't disappear (because it was active). Activating a different material then causes the new custom material to disappear in the material manager but not in the drop-down selection box in the side panel.

However this does make it a lot more stable so I'm cherry-picking this into 2.6. Hope I'm still on time for the 2.6.0 release.

Ghostkeeper:
Cherry-picked into 2.6.

fieldOfView:
> When creating a new material and setting the diameter to 1.75mm (while using UM3) it didn't disappear (because it was active)

Going forward Cura should probably show an Ok/Cancel dialog when changing the diameter in such a way that the approximate_diameter changes. The use-case that it currently works well with is changing the diameter from 2.85 to eg 2.9.

Ghostkeeper:
I'm working on a warning message that's being displayed when you change the diameter in such a way that it disappears. The warning message has an undo button. This should address that issue.

PR #1897, opened 2017-05-28
Create easyarts_ares.def.json

nliaudat:
EasyArts Ares printer definition

### DIFF
--- a/resources/definitions/easyarts_ares.def.json
+++ b/resources/definitions/easyarts_ares.def.json
@@ -1,0 +1,87 @@
+{
+    "id": "easyarts_ares",
+    "name": "EasyArts Ares",
+    "version": 2,
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "nliaudat",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3877.

resources/definitions/easyarts_ares.def.json
```diff
@@ -0,0 +1,87 @@
...
+        "wall_thickness": {
+            "default_value": 1
+        },
+        "top_bottom_thickness": {
+            "default_value": 1
+        },
+        "bottom_thickness": {
+            "default_value": 1
+        },
+        "material_print_temperature": {
```
Ghostkeeper: It's advisable to override `default_material_print_temperature` instead. Otherwise modifying the temperature in the materials doesn't work any more.

nliaudat:
Ok, I've suppressed "material_print_temperature" cause it's default temperature for PLA (200-210°).
Regards

Ghostkeeper:
Sorry that it took so long. Thank you for the addition!

PR #1909, opened 2017-05-31
quality updates

maukcc:
deleted speed_slowdown_layers from individual qualities as it is a
global setting


Ghostkeeper:
You're getting better at these pull requests. Thank you for the update!

PR #1919, opened 2017-06-02
 Cartesio acc. and jerk control

maukcc:
Added acceleration and jerk control for normal operation, for special
materials, and for multi material jobs.


Ghostkeeper:
Devs, see CURA-3908.

maukcc:
I have no idea why that high profile does not pass tests. It is the same as the normal profile.

fieldOfView:
Unfortunately the Jenkins setup results in a lot of false negatives. As an external contributor with no access to the actual build logs, I have started to just ignore the checks until I am told I actually broke something.

Ghostkeeper:
The tests were broken because we needed to upgrade PyQt to version 5.8 for the 2.7 release. And then we had tests running for the 2.6 release using PyQt 5.7 and tests running for the 2.7 release using PyQt 5.8, and so the build environments needed to be different for the two releases. This required some changes to the build system (and the CI system as a consequence). That broke the tests for a while.

Right now it looks like the CI system thinks your commit is fine.

PR #1935, opened 2017-06-05
Update russian translation.

RaD:


Ghostkeeper:
Devs, see CURA-3885 (the conventional 2.6 translations ticket).

resources/i18n/ru/cura.po
```diff
@@ -1508,7 +1510,7 @@ msgstr "Преобразование изображения..."
 #: /home/ruben/Projects/Cura/plugins/ImageReader/ConfigUI.qml:33
 msgctxt "@info:tooltip"
 msgid "The maximum distance of each pixel from \"Base.\""
-msgstr "Максимальная дистанция каждого пикселя от \"Основания\"."
+msgstr "Максимальная дистанция каждого пикселя от \"Основания.\""
```
Ghostkeeper: This is probably fine (meh, details) but feel free to use the grammar rules of your language as you see fit. In American English the period is placed inside the quotation marks. In other languages that doesn't have to be the case, so you don't need to conform to the original text in this case. For instance, British English places it outside of the quotation marks in this case.

Ghostkeeper:
These are complete and working well. Thank you for the update!

RaD:
Good!

Issue #1931, opened 2017-06-04
Layers display error with Python 3.6


sheinz:
There's a problem with showing layers in Layer Viewer on Arch Linux.
When scrolling down, layers are not shown.
There's a demo video with this issue
[https://www.youtube.com/watch?v=gb7YOO-DMB8](https://www.youtube.com/watch?v=gb7YOO-DMB8)
It might be a problem of versioning as on my Ubuntu machine everything seems fine.
Or it might be related to python version, as on Arch linux python version 3 is the default whereas on Ubuntu version 2 is still a default.

I understand that it might be hard to reproduce without specific OS, but if you guide me where to look I might be able to narrow down the problem.

smartavionics:
Hi, if you enable the layer view compatibility mode in settings do you get the old layer view working OK? If so, then the problem could be related to your OpenGL implementation (drivers, libs, etc.?)

When the new layer view arrived, I had to upgrade my hardware to be able to use it.

sheinz:
@smartavionics 
In compatibility mode I get somewhat working layer view but also wrongly segmented.
It seems like the issue is related to the new dictionary implementation of Python3.6.

sheinz:
The problem was indeed in relaying on dictionary being sorted.
I made some changes to explicitly sort the layers and it fixed the issue.

Please have a look at the patch
[cura_python3.6_dict_order_fix.patch.txt](https://github.com/Ultimaker/Cura/files/1050722/cura_python3.6_dict_order_fix.patch.txt)

smartavionics:
That's great. I'm not a Python programmer so I can't really comment on what you've done but hopefully a Cura dev will pick this up. Cheers, Mark.

nallath:
@Sheinz; Could you also make a pull request? That way your code will also be included in the release.

awhiemstra:
Huh. I have seen this issue before (I'm developing on Arch) but never attributed it to Python version. Nice find!.

Note that in regards to the patch, it is probably a better idea to use collections.OrderedDict in LayerDataBuilder (replace the {} dict construction) rather than sorting when we need to render.

sheinz:
@awhiemstra Yes. My first idea was the same. To replace the regular dictionaries with OrderedDict. But in python3.6 the regular dictionaries behaves exactly like OrderedDict, they preserve order of insertion. But in our case we do not need the order of insertion. We need the layers to be sorted by key(layer number).
[...]

PR #1943, opened 2017-06-06
Fix displaying layers with python3.6

sheinz:
This PR fixes https://github.com/Ultimaker/Cura/issues/1931


nallath:
For our reference; CURA-3917

nallath:
My apologies it took so long to merge this. Thanks for the contribution!

PR #1944, opened 2017-06-07
Updated helloBEEprusa printer profiles

marcosfg:
Upgraded helloBEEprusa profile to resolve some print quality and clogging issues.


nallath:
For our reference; CURA-3916

resources/definitions/helloBEEprusa.def.json
```diff
@@ -10,21 +10,28 @@
         "category": "Other",
         "platform": "BEEVERYCREATIVE-helloBEEprusa.stl",
         "platform_offset": [-226, -75, -196],
-        "file_formats": "text/x-gcode"
+        "file_formats": "text/x-gcode",
+		"machine_extruder_trains":
```
Ghostkeeper: You're mixing tabs with spaces here.

resources/definitions/helloBEEprusa.def.json
```diff
@@ -10,21 +10,28 @@
...
+        {
+            "0": "hBp_extruder_left",
+            "1": "hBp_extruder_right"
+        }
     },
 
     "overrides": {
+	    "machine_extruder_count": { "default_value": 2 },
```
Ghostkeeper: And here too.

resources/definitions/helloBEEprusa.def.json
```diff
@@ -10,21 +10,28 @@
...
         "machine_width": { "default_value": 185 },
         "machine_depth": { "default_value": 200 },
         "machine_height": { "default_value": 190 },
         "machine_heated_bed": { "default_value": true },
         "machine_center_is_zero": { "default_value": false },
-        "material_print_temperature": { "default_value": 220 },
+        "material_print_temperature": { "default_value": 200 },
         "material_bed_temperature": { "default_value": 60 },
         "material_diameter": { "default_value": 1.75 },
+		"line_width": { "default_value": 0.48 },
```
Ghostkeeper: More tabs.

Ghostkeeper: Are you sure that a line width of 0.48 with a nozzle size of 0.4 is going to work?

Well, your call I guess :)

marcosfg: Yes, we have been testing it with these parameters and its working better than with 0.4.

Thank you for fixings the tabs.

PR #1952, opened 2017-06-08
[WIP] Added settings for slow start bottom walls feature.

smartavionics:
Still very much WIP, the idea is that with some filaments, build plate adhesion can be marginal and when
starting a new bottom layer wall, the extrusion can fail to stick right at the start of the extrude. So by going
a bit slower for the first couple of mm, it should increase the chance of the filament sticking to the bed without
having to do the whole wall at that super slow speed.

See https://github.com/Ultimaker/CuraEngine/pull/531.


smartavionics:
Reworking this to make it more flexible.

PR #1961, opened 2017-06-11
Update Peopoly Moai definition

fieldOfView:
This PR adds common values for infill overlap and adhesion type to the Moai definition by request of Peopoly.


Ghostkeeper:
Devs, see CURA-3943.

PR #1978, opened 2017-06-15
Horizontal Ironing

Ghostkeeper:
This implements the basic ironing technique: Just horizontal ironing.

Ironing causes the nozzle to go over the top surface once again while extruding only very little material. This makes the top surface very smooth.

These settings are now in the Experimental category. As far as I'm concerned they can be put in the normal Shell category, since the technique is reliable.

See also the accompanying pull request in the engine: https://github.com/Ultimaker/CuraEngine/pull/534

Implements CURA-3903.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5423,7 +5423,7 @@
                     "description": "The amount of material, relative to a normal skin line, to extrude during sanding. Keeping the nozzle filled helps filling some of the crevices of the top surface, but too much results in overextrusion and blips on the side of the surface.",
                     "type": "float",
                     "unit": "%",
-                    "default_value": 30.0,
+                    "default_value": 10.0,
                     "minimum_value": "0",
                     "maximum_value_warning": "50",
                     "enabled": "sanding_enabled",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2310,6 +2310,21 @@
                             "limit_to_extruder": "top_bottom_extruder_nr",
                             "settable_per_mesh": true
                         },
+                        "acceleration_sanding":
+                        {
+                            "label": "Sanding Acceleration",
+                            "description": "The acceleration with which sanding is performed.",
+                            "unit": "mm/s²",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "minimum_value_warning": "100",
+                            "maximum_value_warning": "10000",
+                            "default_value": 3000,
+                            "value": "acceleration_topbottom",
+                            "enabled": "resolveOrValue('acceleration_enabled') and sanding_enabled",
+                            "limit_to_extruder": "top_bottom_extruder_nr",
+                            "settable_per_mesh": true
+                        },
                         "acceleration_support":
                         {
                             "label": "Support Acceleration",
@@ -2580,6 +2595,20 @@
                             "default_value": 20,
                             "value": "jerk_print",
                             "enabled": "resolveOrValue('jerk_enabled')",
+                            "limit_to_extruder": "top_bottom_extruder_nr",
+                            "settable_per_mesh": true
+                        },
+                        "jerk_sanding":
+                        {
+                            "label": "Sanding Jerk",
+                            "description": "The maximum instantaneous velocity change while performing sanding.",
+                            "unit": "mm/s",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "maximum_value_warning": "50",
+                            "default_value": 20,
+                            "value": "jerk_topbottom",
+                            "enabled": "resolveOrValue('jerk_enabled') and sanding_enabled",
                             "limit_to_extruder": "top_bottom_extruder_nr",
                             "settable_per_mesh": true
                         },
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5387,7 +5387,7 @@
                 },
                 "sanding_enabled":
                 {
-                    "label": "Enable Sanding",
+                    "label": "Enable Ironing",
                     "description": "Go over the top surface one additional time, but without extruding material. This is meant to melt the plastic on top further, creating a smoother surface.",
                     "type": "bool",
                     "default_value": false,
@@ -5396,8 +5396,8 @@
                 },
                 "sanding_pattern":
                 {
-                    "label": "Sanding Pattern",
-                    "description": "The pattern to use for sanding top surfaces.",
+                    "label": "Ironing Pattern",
+                    "description": "The pattern to use for ironing top surfaces.",
                     "type": "enum",
                     "options":
                     {
@@ -5411,8 +5411,8 @@
                 },
                 "sanding_line_spacing":
                 {
-                    "label": "Sanding Line Spacing",
-                    "description": "The distance between the lines of sanding.",
+                    "label": "Ironing Line Spacing",
+                    "description": "The distance between the lines of ironing.",
                     "type": "float",
                     "unit": "mm",
                     "default_value": 0.1,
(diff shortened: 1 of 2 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -2569,6 +2599,20 @@
                             "limit_to_extruder": "top_bottom_extruder_nr",
                             "settable_per_mesh": true
                         },
+                        "jerk_sanding":
+                        {
+                            "label": "Sanding Jerk",
+                            "description": "The maximum instantaneous velocity change while performing sanding.",
```
nallath: The description needs to change; In some places you call it sanding (here) and in other places you call it ironing. I think you need to pick one and stick with it. 
I'd also prefer it if the setting ids would also be consistent, but that is not as important.

Ghostkeeper: Yeah, I think I forgot them because they aren't in the same category as the rest of the ironing settings.

I fixed the setting names and descriptions here: https://github.com/Ultimaker/Cura/pull/1978/commits/a7ed0d011ad60f878629b86a1df6113718c0423a
I fixed the setting IDs here: https://github.com/Ultimaker/Cura/pull/1978/commits/092483cac5e7d6fb478502c3fca666d60627b905 and in the accompanying commit in CuraEngine: https://github.com/Ultimaker/CuraEngine/pull/534/commits/d927f7317a2601eb90f650e936377d3f7aa5b360

resources/definitions/fdmprinter.def.json
```diff
@@ -1970,6 +1970,21 @@
                             "limit_to_extruder": "top_bottom_extruder_nr",
                             "settable_per_mesh": true
                         },
+                        "speed_ironing":
+                        {
+                            "label": "Ironing Speed",
+                            "description": "The speed at which to pass over the top surface.",
```
BagelOrb: 'The speed *with* which'

BagelOrb: 'the top surface' > 'flat top surface*s*'

BagelOrb: ', when Ironing is enabled' or ', when performing ironing'

By comparison: you describe the jerk explicitly as the jerk 'with which ironing is performed'

resources/definitions/fdmprinter.def.json
```diff
@@ -1970,6 +1970,21 @@
...
+                            "type": "float",
+                            "unit": "mm/s",
+                            "default_value": 20.0,
+                            "value": "speed_topbottom * 20 / 30",
+                            "minimum_value": "0.001",
+                            "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
+                            "maximum_value_warning": "100",
+                            "enabled": "ironing_enabled",
+                            "limit_to_extruder": "top_bottom_extruder_nr",
```
BagelOrb: You've implemented all settings for a new feature type, so why is this limited to the top_bottom_extruder_nr, rather than an ironing_extruder_nr ?
Seems like a really small step to add that funcitonality..
[not blocking]

resources/definitions/fdmprinter.def.json
```diff
@@ -2296,6 +2311,21 @@
...
+                            "label": "Ironing Acceleration",
+                            "description": "The acceleration with which ironing is performed.",
+                            "unit": "mm/s²",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "minimum_value_warning": "100",
+                            "maximum_value_warning": "10000",
+                            "default_value": 3000,
+                            "value": "acceleration_topbottom",
+                            "enabled": "resolveOrValue('acceleration_enabled') and ironing_enabled",
```
BagelOrb: acceleration_enabled is not really a property which can be different between the extruders; it doesn't make sense to have a resolve function for it.

However, the value of the setting is already retrieved inconsistently, because sometimes it's just retrieved without calling resolveOrValue...
[not blocking]

resources/definitions/fdmprinter.def.json
```diff
@@ -5340,6 +5384,70 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
+                },
+                "ironing_enabled":
+                {
+                    "label": "Enable Ironing",
+                    "description": "Go over the top surface one additional time, but without extruding material. This is meant to melt the plastic on top further, creating a smoother surface.",
```
BagelOrb: There is a flow setting, so it is not `without extruding material`

It doesn't really make sense to say that melting causes the top surface to be more flat.

Maybe something like `Smooth out bumps and crevices on flat top parts by passing over the top skin lines again after they have been printed.`

resources/definitions/fdmprinter.def.json
```diff
@@ -5340,6 +5384,70 @@
...
+                },
+                "ironing_pattern":
+                {
+                    "label": "Ironing Pattern",
+                    "description": "The pattern to use for ironing top surfaces.",
+                    "type": "enum",
+                    "options":
+                    {
+                        "concentric": "Concentric",
+                        "zigzag": "Zig Zag"
```
BagelOrb: Why not all patterns which are available for top_bottom_pattern?!

Ironing is basically redoing the skin with less flow (and different direction), so I imagine that all the patterns must be available.

The lines type is different from the zig zag type, especially when the flow is more than zero!

resources/definitions/fdmprinter.def.json
```diff
@@ -5340,6 +5384,70 @@
...
+                    "limit_to_extruder": "top_bottom_extruder_nr",
+                    "settable_per_mesh": true
+                },
+                "ironing_line_spacing":
+                {
+                    "label": "Ironing Line Spacing",
+                    "description": "The distance between the lines of ironing.",
+                    "type": "float",
+                    "unit": "mm",
+                    "default_value": 0.1,
```
BagelOrb: Why so small?

I would expect you can iron very fast and get good results when using the full flat, so then the de fault `value` would be something like `machine_nozzle_tip_outer_diameter * 0.9`

resources/definitions/fdmprinter.def.json
```diff
@@ -5340,6 +5384,70 @@
...
+                    "minimum_value": "0.001",
+                    "maximum_value_warning": "machine_nozzle_tip_outer_diameter",
+                    "enabled": "ironing_enabled",
+                    "limit_to_extruder": "top_bottom_extruder_nr",
+                    "settable_per_mesh": true
+                },
+                "ironing_flow":
+                {
+                    "label": "Ironing Flow",
+                    "description": "The amount of material, relative to a normal skin line, to extrude during ironing. Keeping the nozzle filled helps filling some of the crevices of the top surface, but too much results in overextrusion and blips on the side of the surface.",
```
BagelOrb: It is very weird to make this value relative to the normal skin line width, rather than the ironing line spacing, because that means that changing the line distance influences over- and underextrusion.

BagelOrb: It's weird to talk about `Keeping the nozzle filled`; this setting is about extrusion, not about being unretracted. moreover it doesn't match the use of the terms `, but too much`; i.e. keeping the nozzle filled too much doesn't make sense.

Suggestion: `Keeping the nozzle filled` > `Extruding a bit during ironing`

resources/definitions/fdmprinter.def.json
```diff
@@ -5340,6 +5384,70 @@
...
+                    "minimum_value": "0",
+                    "maximum_value_warning": "50",
+                    "enabled": "ironing_enabled",
+                    "limit_to_extruder": "top_bottom_extruder_nr",
+                    "settable_per_mesh": true
+                },
+                "ironing_inset":
+                {
+                    "label": "Ironing Inset",
+                    "description": "A distance to keep from the edges of the model. Ironing all the way to the edge of the mesh may result in a jagged edge on your print.",
```
BagelOrb: Please explain to me and in the descriptions: what is the normal area which is being ironed? I thought it would have been (part of) the top skin area on that layer, but this description seems to suggest that it is larger than that, namely up to the outlines of the layer.

resources/definitions/fdmprinter.def.json
nickthetait: Sounds to me like "smoothing_enabled" is a more appropriate setting name

resources/definitions/fdmprinter.def.json
BagelOrb: When printing the skin with a .25 nozzle, it might be quite nice to do the ironing with a .8 nozzle.

PS: Do those nozzle's have different outer tip diameters?!

nickthetait:
Seems like a quite valuable and interesting technique!

BagelOrb:
It sure is. Results seem quite promising!

nallath:
Quite promising? They are frigging amazing!

PR #1994, opened 2017-06-19
Add a dark theme

fieldOfView:
This PR adds a dark theme, based on the work by @rudowinger. See https://github.com/Ultimaker/Cura/issues/1371 for the original screenshots

Current state:
image

image


nallath:
Shouldn't we make the grids a bit darker though? That seems like an awful lot of light for a dark theme.

fieldOfView:
image

Darker, but still showing some contrast with the background

nallath:
Much better! :)

LipuFei:
The background colour of material info button (the "i") is now the same as the panel... Could you change the button's colour to make it visible?
image

fieldOfView:
@LipuFei:
No, I can't, because of the way you built/styled the info button. You gave it the colors of the toolbar buttons.

I wonder why you introduced another style info buttons, when we already had an "i" button in the settings (the inherit-button).

LipuFei:
@fieldOfView It was required to look like that... which is different from the "i" buttons in the settings. I will try to find a better way to colour it then.

fieldOfView:
@LipuFei: theming is all about consistency. Don't add a new theme entry/rule for each new control, but instead look at what controls semantically should look and feel similar. 

In this case, the material info button should probably look and behave similar to the icon buttons in the custom settings. While the use of a textual "i" is elegant in the way that it does not require a new svg, the other buttons are implemented as svg images, so it makes sense from a theming point of view to also use an svg for this button (and to use the same colors and hover effects).

Also feel free to convey to the designers that I don't agree with the way they throw consistency out of the window.

nallath:
Yeah, that's an ongoing struggle. I also dislike the pre-heat button for suddenly following completely different design logic as the rest of the application.

BagelOrb:
:champagne:

nallath:
We do need to figure out a new color for the tooltips; The blue is a bit in your face.

fieldOfView:
Again I like the contrast of the tooltip color (it is called a "dark" theme, not "drab" theme ;-) ), but here's an alternative using Ultimaker dark-blue instead of Ultimaker cyan:

image

fieldOfView:
Note to self: the placeholder-text of the Search box is not themeable. To be fixed in a future PR.

rudowinger:
Awesome!!!!

Ghostkeeper:
I agree! That looks neat.

I don't know what will survive of this in the next version (since I shouldn't add that to 2.7) but here you go: 9241cfa5c6e7df071f5018d48a67ec76b0f863d3

PR #1995, opened 2017-06-20
Added z_seam_relative setting to determine whether z seam origin is relative to part or not.

smartavionics:
See https://github.com/Ultimaker/CuraEngine/pull/536.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1166,6 +1166,17 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "z_seam_relative":
+                {
+                    "label": "Z Seam Relative",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see CURA-3971.

resources/definitions/fdmprinter.def.json
```diff
@@ -1166,6 +1166,17 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "z_seam_relative":
+                {
+                    "label": "Z Seam Relative",
+                    "description": "When enabled, the z seam coordinates are relative to each part's centre. By default, the coordinates define an absolute position on the platform.",
```
BagelOrb: The description should be independent from the default value, because it might be different for each printer.
Suggested: "By default" => "When disabled"

also "platform" => "build plate"

smartavionics: Done!

Issue #2007, opened 2017-06-21
2.6 Filament Diameter and Nozzle Temp Ignored


DDeGonge:
I'm using Cura 2.6 on Windows, this is a clean install with no previous profiles. My machine is set to custom with 1.75mm filament. I have created a filament profile as well with 1.72mm as the mean diameter and 210*C printing temperature. I have tried using both versions of Marlin (standard and volumetric), but standard always under-extrudes and volumetric always over-extrudes.

When viewing the full list of slicing settings, I am able to see the proper filament diameter and temperature settings. However, when the model is sliced, there are 2 problems.

1. It always uses a print temperature of 200*C. I have tried using different baked in profiles as well as adjusting the value in the slicing settings by hand, but nothing seems to change the generated gcode print temperature.
2. It always slices assuming filament is 2.85mm diameter, but the preview estimation stats are correct. For example, a solid 2cm cube slices and when it is complete, the stats in the bottom right say it will use 2.2m of filament (I forget the actual value but it is close to 2.2). This is correct for 1.75mm filament. However, when importing the generated gcode into repetier host, it shows that only 0.8m of filament are actually being used. When printing, it shows that in fact 0.8m is accurate as the print is severely under-extruding. This is using the standard Marlin gcode flavor. Using volumetric seems to do the opposite and results in severe over-extrusion.

Increasing the flow multiplier in Cura to 270% fixes the problem and prints nicely, but this is a bit hacky and doesn't address the cause. Additionally, I still can not figure out any way to adjust the sliced print temperature.

Let me know if you need more info!

-Andrew

fieldOfView:
What printer definition do you use? One of the predefined ones, or a Custom FDM Printer?

DDeGonge:
Custom FDM

fieldOfView:
Both issues can be checked in the generated gcode. Could you post it somewhere?

The print temperature is set near the top of the gcode file. There should be a line that says eg ```M104 S209``` if you set the temperature to 209 degrees.

The total amount of extruded material can be check by scrolling to the bottom of the gcode file and checking the last G1 before the end gcode. The E value shows the last "position" in mm.

fieldOfView:
Could you try lying to Cura and setting the number of extruders to 2?

DDeGonge:
[Here is the gcode for a test part](https://drive.google.com/file/d/0BzkIVnYJoiSgR20wNERQSGNPRzQ/view?usp=sharing)

[...]

PR #2009, opened 2017-06-21
[2.6] Don't send extruder stack to CuraEngine when there is only 1 extruder

fieldOfView:
This PR fixes an issue which causes settings to be ignored if a Custom FDM Printer is used with the number of extruders set to 1. Fixes https://github.com/Ultimaker/Cura/issues/2007

When there is 1 extruder, the frontend stores all settings in the global stack. Sending an extruder stack confuses CuraEngine into using the values of the extruder stack, which results in defaults being used.

Basically, single extrusion Custom FDM Printers are broken at the moment, so if there were a point-release for Cura 2.6 it would be great if this fix could be included.

Steps to reproduce:
* Add a Custom FDM Printer
* Keep number of extruders to 1, keep diameter at 2.85
* Slice an object, and note down the estimated length of material.
* Change the diameter to 1.75 in the sidebar
* Note that after slicing the estimated length has changed
* Save the gcode file, and inspect the last G1 before the end gcode

Expected:
The E-value for the last G1 reflects the changed amount of material

Actual:
The E-value for the last G1 reflects the original amount of material, for 2.85 mm diameter.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -149,8 +149,9 @@ class StartSliceJob(Job):
             self._buildGlobalSettingsMessage(stack)
             self._buildGlobalInheritsStackMessage(stack)
 
-            for extruder_stack in ExtruderManager.getInstance().getMachineExtruders(stack.getId()):
-                self._buildExtruderMessage(extruder_stack)
+            if stack.getProperty("machine_extruder_count", "value") > 1:
+                for extruder_stack in ExtruderManager.getInstance().getMachineExtruders(stack.getId()):
+                    self._buildExtruderMessage(extruder_stack)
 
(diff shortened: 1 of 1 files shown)


nallath:
Thanks for the fast fix!

BagelOrb:
The problem seems fixed, but when I create a new Custon FDM Printer, the default material diameter is set to 1.75. Changing it to 2.85 doesn't have any effect on the material diameter in the side panel.

It seems to me that the material diameter should be removed from the machine settings...

fieldOfView:
> It seems to me that the material diameter should be removed from the machine settings...

We need a way to set the filament diameter the printer requires.

> the default material diameter is set to 1.75. Changing it to 2.85 doesn't have any effect on the material diameter in the side panel.

Hmm, the problem is that the diameter gets overridden by the currently active material...

PR #2018, opened 2017-06-22
Add tab navigation to sidebar

fieldOfView:
This PR adds tab-navigation between fields in the (custom) sidebar; it allows the user to tab/shift-tab between fields in the sidebar, making it quicker to make changes. Checkbox items can be operated with the spacebar, as can category headers. Comboboxes can be operated by the spacebar and/or up/down cursor keys.

### DIFF
--- a/resources/qml/Settings/SettingCategory.qml
+++ b/resources/qml/Settings/SettingCategory.qml
@@ -14,10 +14,11 @@ Button {
 
     style: UM.Theme.styles.sidebar_category;
 
-    signal showTooltip(string text);
-    signal hideTooltip();
+    signal showTooltip(string text)
+    signal hideTooltip()
     signal contextMenuRequested()
     signal showAllHiddenInheritedSettings(string category_id)
+    signal focusReceived()
 
     text: definition.label
     iconSource: UM.Theme.getIcon(definition.icon)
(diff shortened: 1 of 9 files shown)


fieldOfView:
I know there's a JIRA issue about keyboard friendliness, but I can't remember the issue nr. Probably tagged somehow with Oak & Morrow.

nallath:
CURA-1531

fieldOfView:
Technically "you didn't do this" ;-)

Thanks for finding the jira issue. I had guestimated it was around 1900, but the 1500's makes sense. There's a similar issue about touchscreen friendliness that I still want to do.

nallath:
If I have the focus of a category header, it "loops" between category headers. Is that intentional?

BagelOrb:
@nallath You mean that you cannot get out of the category headers any more?

fieldOfView:
@nallath please clarify. When I tab onto a category header, and press tab again, focus is moved to the next control (be it another header or a text/checkbox/combobox control)

fieldOfView:
Ah, I think I see. When you expand a category, new controls are created and these are then out of the taborder. So tabbing to the next control will tab to the next category header instead of the first control in the category. I hope that can be fixed easily.

nallath:
If it's something thats really hard to fix, i think we can still add the code as is. It's already a big improvement compared to the original state.

fieldOfView:
Instead of relying on QML to automagically understand what we want it to do, I now catch tab and backtab and focus the correct item.

nallath:
Thanks for the contribution!

ErikDeBruijn:
Nice one @fieldOfView ! Your continued contributions are making Cura more and more pleasant to work with for everyone!

PR #2029, opened 2017-06-25
Add DiscoEasy200 printer (dagoma.fr)

landru29:
Adding printer from https://dagoma.fr/


nallath:
For our reference; CURA-3966

nallath:
Thanks for the contribution!

Issue #2026, opened 2017-06-25
Line width problem in new layer view


smartavionics:
I have noticed that the new layer view shows some pretty weird line widths. Here's an example:

![screenshot_2017-06-25_17-39-07](https://user-images.githubusercontent.com/585618/27517970-3a5e1722-59cd-11e7-81eb-4fdf00b8b266.png)
The top layer has two walls close together and with wall overlap compensation enabled, the outer wall gets thinned. But look at the outer wall in the above image, the longish straight segments are all too thin and the corners are rather bumpy. The actual print looks absolutely fine so it must be a graphics issue. This is just one example, I could post lots of images of other walls whose widths (in the layer view) are silly.
My current theory is that the array of line widths sent to the UI are not synchronised with the vertices. So in ProcessSlicedLayersJob, if I add this line (after line 128) to rotate the array of line widths:

line_widths = numpy.roll(line_widths, -1);

I now get this:
![screenshot_2017-06-25_17-46-17](https://user-images.githubusercontent.com/585618/27518030-38f4d366-59ce-11e7-9dac-1b656673308d.png)
Which is much more sensible and other wall lines that had weird widths now look good too so that rotation of the array is good for the walls. Unfortunately, it causes the skins and infill to be replaced with very thin lines as shown below and I don't know what to do next.

![screenshot_2017-06-25_17-48-05](https://user-images.githubusercontent.com/585618/27518048-8f926a4e-59ce-11e7-8668-4f8524e04027.png)
Could someone who understands this python + GL stuff please look into this as I don't want to waste loads of time working out a fix if someone who is familiar with the code can do the same in a fraction of the time?

smartavionics:
OK, I figured out a fix so that now all lines are showing with sensible widths but I would be the first to admit that I am thrashing around in the dark here so it could well be that the real problem is elsewhere and this change is just compensating for a bug elsewhere?

nallath:
@jackha You made the original implementation of this, what do you think?

PR #2030, opened 2017-06-26
In geometry41core, take the line width/depth from the second vertex rather than the first.

smartavionics:
This change fixes a problem where a sequence of lines that have different widths are displayed
with the widths shifted by 1, i.e. line 1 had the width of line 2 and so on.

Fixes https://github.com/Ultimaker/Cura/issues/2026

### DIFF
--- a/plugins/LayerView/layers3d.shader
+++ b/plugins/LayerView/layers3d.shader
@@ -130,9 +130,9 @@ geometry41core =
             // fixed size for movements
             size_x = 0.05;
         } else {
-            size_x = v_line_dim[0].x / 2 + 0.01;  // radius, and make it nicely overlapping
+            size_x = v_line_dim[1].x / 2 + 0.01;  // radius, and make it nicely overlapping
         }
-        size_y = v_line_dim[0].y / 2 + 0.01;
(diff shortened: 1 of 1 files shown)


nallath:
@jackha Could you also have a look?

smartavionics:
A further example of the difference this PR makes, before..
screenshot_2017-06-26_09-21-24
After..
screenshot_2017-06-26_09-24-36

nallath:
@BagelOrb I think he found the issue you wanted to get fixed :)

nallath:
For our reference; CURA-3965

BagelOrb:
Yay!!!!

BagelOrb:
I hope we can merge this to 2.6, but I fear the worst :/

BagelOrb:
Yup. Too late for 2.6.1 :(

smartavionics:
One thing I don't understand is why the line type appears to be correct while the width is wrong. I would have thought that the type would have to come from the second vertex also?

nallath:
I just discussed this with @jackha, who also can't figure out why this would make any difference. Usually when unpacking it's the items at the end of the list that break.

 But i think it's a good idea to merge this, as it does fix a pretty annoying isuse.

Issue #1573, opened 2017-03-24
Feature Request: Horizontal Expansion on Initial layer


Labern2:
The squish factor that helps make prints stick to the glass also leaves the print with a sharp edge on the bottom.
If you could adjust the horizontal expansion and enter a - amount (eg -0.3) then this will remove the sharp edge and make the print square on the bottom. I have done tests with this by adding 2 stl's. One was the height of the Initial layer which I used per object settings to enter the -0.3 horizontal expansion. merged the 2 stl's and got a print with a square edge on the base.
This would allow a more aggressive initial layer and not distort the prints base to get strong adhesion.

Ghostkeeper:
That's basically what Brim does, isn't it?

Labern2:
No. It's kinda the opposite of brim. 
the initial layer is designed to squash onto the glass. The more squash the stronger it sticks to the glass.
the squashing action makes the line width wider and it ends up wider than the actual parts dimensions. If you print a part with skirt selected, you end up with a tiny brim from the squashed initial layer.
This can effect parts that need dimensional accuracy like an internal threaded hole and it makes the part sharp on the bottom.
But if you use horizontal expansion more as horizontal contraction on the initial layer you can remove the tiny brim or sharp edge.
Pictures shown are print placed upside down. One showing the sharp edge the other was with my modified horizontal expansion settings set to -0.3. Not perfect examples but illustrates my point.
![img_20170313_052712](https://cloud.githubusercontent.com/assets/17557978/24291734/14f93eac-10ef-11e7-9fb9-695961ea890f.jpg)
![img_20170313_052633](https://cloud.githubusercontent.com/assets/17557978/24291740/17277108-10ef-11e7-960e-63a794e443c6.jpg)

Ghostkeeper:
Ah, I see. It's to compensate elephant feet. Well, that could help. I'm making a feature ticket and we'll see what the planning allows.

Ghostkeeper:
Apparently the planning didn't allow it. Our product manager just said we aren't going to do this.

Dim3nsioneer:
But you would accept a proper pull request if someone from the community does it, right?

fieldOfView:
That is generally the case.

fieldOfView:
> But you would accept a proper pull request if someone from the community does it, right?

Let's test that theory.
It was fairly trivial to implement.

fieldOfView:
Feature has been implemented for Cura 2.7. This issue can be closed

[...]

PR #2031, opened 2017-06-26
Add a setting for Horizontal Expansion for the first layer

fieldOfView:
This PR adds functionality to change the Horizontal Expansion for the first layer only. This can be used to compensate for "elephant's foot".

See Ultimaker/Cura#1573 for discussion
Also see CuraEngine PR https://github.com/Ultimaker/CuraEngine/pull/540

### DIFF
--- a/cura/ConvexHullDecorator.py
+++ b/cura/ConvexHullDecorator.py
@@ -257,7 +257,11 @@ class ConvexHullDecorator(SceneNodeDecorator):
     #   \return New Polygon instance that is offset with everything that
     #   influences the collision area.
     def _offsetHull(self, convex_hull):
-        horizontal_expansion = self._getSettingProperty("xy_offset", "value")
+        horizontal_expansion = max(
+            self._getSettingProperty("xy_offset", "value"),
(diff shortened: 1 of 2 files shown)


nallath:
CURA-3968

nallath:
You also need to add it to the convexHullDecorator.

fieldOfView:
Aww, poop... There goes my "hey, it's a trivial change" feature.
I'll look into that.

fieldOfView:
Did not mean to close the PR...

Truely updating the convexhull with a 1st layer xy offset is quite involved. I would have to do a boolean to separate the first slice of the model, then offset and finally boolean the slice back. This sounds way too expensive. And it is not quite right either if you think about the squish of the first layer.

Here's my proposal for a "good enough" workaround: The convex hull should just be offset by the max value of xy_offset and xy_offset_layer_0. If xy_offset_layer_0 is smaller you don't loose anything, and if xy_offset_layer_0 is larger at least you don't get collisions.

fieldOfView:
Ready for review

nallath:
Thanks for the contribution!

PR #2056, opened 2017-07-01
Top most skin feature (CURA-3940)

BagelOrb:


resources/definitions/fdmprinter.def.json
```diff
@@ -921,6 +935,46 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "topmost_skin_extruder_nr":
+                {
+                    "label": "Roofing Extruder",
+                    "description": "The extruder train used for printing the top most skin. This is used in multi-extrusion.",
+                    "type": "optional_extruder",
+                    "default_value": "-1",
+                    "value": "-1",
```
Ghostkeeper: The value is superfluous. It's equal to the default value.

It's a bit weird that these `_extruder_nr`s have a string-type value, but that seems to go for all of them.

resources/definitions/fdmprinter.def.json
```diff
@@ -921,6 +935,46 @@
...
+                    "settable_per_mesh": true,
+                    "settable_per_extruder": false,
+                    "settable_per_meshgroup": true,
+                    "settable_globally": true,
+                    "enabled": "machine_extruder_count > 1"
```
Ghostkeeper: And if `topmost_skin_layer_count > 0` and `top_layers > 0`.

resources/definitions/fdmprinter.def.json
```diff
@@ -921,6 +935,46 @@
...
+                    "type": "enum",
+                    "options":
+                    {
+                        "lines": "Lines",
+                        "concentric": "Concentric",
+                        "zigzag": "Zig Zag"
+                    },
+                    "default_value": "lines",
+                    "limit_to_extruder": "topmost_skin_extruder_nr",
+                    "settable_per_mesh": true
```
Ghostkeeper: Disable if there is no top skin.

resources/definitions/fdmprinter.def.json
```diff
@@ -1999,6 +2053,20 @@
...
+                            "description": "The speed at which roofing layers are printed.",
+                            "unit": "mm/s",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
+                            "maximum_value_warning": "150",
+                            "default_value": 25,
+                            "value": "speed_print / 60 * 25",
+                            "limit_to_extruder": "topmost_skin_extruder_nr",
+                            "settable_per_mesh": true
```
Ghostkeeper: Disable if there is no top skin or roofing.

resources/definitions/fdmprinter.def.json
```diff
@@ -2339,6 +2407,21 @@
...
+                            "label": "Roofing Acceleration",
+                            "description": "The acceleration with which roofing layers are printed.",
+                            "unit": "mm/s²",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "minimum_value_warning": "100",
+                            "maximum_value_warning": "10000",
+                            "default_value": 3000,
+                            "value": "acceleration_print",
+                            "enabled": "resolveOrValue('acceleration_enabled')",
```
Ghostkeeper: Disable if there is no top skin or roofing.

resources/definitions/fdmprinter.def.json
```diff
@@ -2628,6 +2711,20 @@
...
+                        {
+                            "label": "Roofing Jerk",
+                            "description": "The maximum instantaneous velocity change with which roofing layers are printed.",
+                            "unit": "mm/s",
+                            "type": "float",
+                            "minimum_value": "0.1",
+                            "maximum_value_warning": "50",
+                            "default_value": 20,
+                            "value": "jerk_print",
+                            "enabled": "resolveOrValue('jerk_enabled')",
```
Ghostkeeper: Disable if there is no top skin or roofing.

resources/definitions/fdmprinter.def.json
```diff
@@ -699,6 +699,20 @@
...
+                        {
+                            "label": "Roofing Line Width",
+                            "description": "Width of a single line of the areas at the top of the print.",
+                            "unit": "mm",
+                            "minimum_value": "0.001",
+                            "minimum_value_warning": "0.1 + 0.4 * machine_nozzle_size",
+                            "maximum_value_warning": "2 * machine_nozzle_size",
+                            "default_value": 0.4,
+                            "type": "float",
+                            "value": "line_width",
```
Ghostkeeper: Should be `top_bottom_line_width`, right?

resources/definitions/fdmprinter.def.json
BagelOrb: fixed

resources/definitions/fdmprinter.def.json
BagelOrb: I don't see topmost skin as a subpart of normal skin. It's a separate feature on its own. It should therefore inherit from the line width like any other print feature line width setting.

resources/definitions/fdmprinter.def.json
BagelOrb: fixed

resources/definitions/fdmprinter.def.json
BagelOrb: fixed

resources/definitions/fdmprinter.def.json
BagelOrb: fixed

resources/definitions/fdmprinter.def.json
BagelOrb: v

resources/definitions/fdmprinter.def.json
BagelOrb: v

BagelOrb:
The top/bottom angles are used for the roofing.
I don't really see a reason to make an extra setting to control the roofing angles separately.

What reason(s) can you think of?

resources/definitions/fdmprinter.def.json
Ghostkeeper: Oh, but that's a bit strange right? Like how support roofing is part of support, this roofing should also be part of top skin.

Ghostkeeper:
For instance, you've set the line width to be so small that different angles aren't needed any more and it's faster to print pure diagonally every time.

resources/definitions/fdmprinter.def.json
BagelOrb: That comparison is skewed. Like support roofing is part of support so is roofing part of the mesh.
Support > bottom, infill, top, (walls)
mesh > walls, top/bottom, infill, roofing, (perimeter gaps, outline gaps)

BagelOrb: I could make it part of the skin conceptually, but then we'd need to introduce parent settings for all top/bottom and roofing settings.
mesh > skin, infill, walls, etc
skin > top/bottom, roofing

BagelOrb:
So maybe it's an idea to just make the roofing always diagonal without introducing a setting for it?

Ghostkeeper:
By default the roofing has a line width of 0.35mm, in which case he probably wants the alternation to continue as normal.

BagelOrb:
I don't get what you mean. What do you want me to do?

Ghostkeeper:
Shall we add a setting for the skin angles of the roofing that inherits from the normal skin angles, and keeps alternating based on absolute layer number?

BagelOrb:
I will make a setting for the angles, but it won't inherit, because then the roofing should be a part of the top/bottom skin everywhere, which means we'd have to introduce parent settings for both the top/bottom and the roofing, which introduces a lot of settings. Also we'd have to come up with a name for the parent of top/bottom and roofing.

Remember that the engine can only use leaf settings, so we cannot make roofing a child of top/bottom.

resources/definitions/fdmprinter.def.json
```diff
@@ -932,13 +946,63 @@
...
+                    "settable_per_mesh": true,
+                    "enabled": "topmost_skin_layer_count > 0 and top_layers > 0"
+                },
+                "topmost_skin_angles":
+                {
+                    "label": "Roofing Line Directions",
+                    "description": "A list of integer line directions to use when the roofing layers use the lines or zig zag pattern. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the traditional default angles (45 and 135 degrees).",
+                    "type": "[int]",
+                    "default_value": "[ ]",
+                    "enabled": "topmost_skin_pattern != 'concentric'",
```
Ghostkeeper: And `topmost_skin_layer_count > 0` and `top_layers > 0`.

resources/definitions/fdmprinter.def.json
```diff
@@ -932,13 +946,63 @@
...
+                    "limit_to_extruder": "topmost_skin_extruder_nr",
+                    "settable_per_mesh": true,
+                    "enabled": "topmost_skin_layer_count > 0 and top_layers > 0"
+                },
+                "topmost_skin_angles":
+                {
+                    "label": "Roofing Line Directions",
+                    "description": "A list of integer line directions to use when the roofing layers use the lines or zig zag pattern. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the traditional default angles (45 and 135 degrees).",
+                    "type": "[int]",
+                    "default_value": "[ ]",
```
Ghostkeeper: Perhaps it's a better idea to simply inherit from the skin angles, so we won't need to make any exception case for the empty list.

PR #2063, opened 2017-07-04
Green theme

BagelOrb:
I've made a green theme and I like it.
Been using this theme for the past week and haven't found any wrong colors any more the last few days.


fieldOfView:
image

I would choose either a darker color for the background of the tooltips, or a dark color for the tooltip text.

fieldOfView:
For reference and inspiration, here's another green-ish theme:
image

nallath:
Oh god that mint green for the tooltip.

fieldOfView:
I don't mind the mint green in itself (my teeth are getting whiter just looking at it), but the white text does not work well on top of it.

nickthetait:
I have a sudden urge to consume ice cream.

BagelOrb:
Is it because my theme is so hot?

BagelOrb:
I agree about the lightness of the description box (the rectangular arrow), but I've checked and the lightness value is the same as the normal box when you add up the r + g + b values.

I thought the human eye was better at discerning green colors, so this doesn't make any sense.

I'll look into making the color darker..

BagelOrb:
I've made the tooltip (and all thigns with the same color) darker. What do you think?

Ghostkeeper:
Devs, see CURA-4020.

nallath:
I'm not a big fan of adding a lot of themes. The current selection box isn't that well suited for a lot of themes. 
I can see the appeal of having the dark theme, but no so much in just switching around some colors.

fieldOfView:
Ideally, themes would be downloadable like plugins.

nallath:
Yeah, that would be a lot better.

Ghostkeeper:
We've decided to not merge this one. At least not for now.

Once we have a theme repository it would make sense to give the user more freedom to use this kind of themes, but up until then, just a hue shift of the current theme is not something we really need to provide for our users.

Issue #1977, opened 2017-06-15
Extruder Y-offset generates wrong disallowed area


maukcc:
Cura 2.5.99 master
OS Ubuntu 16.04

The Y offset is correct for the placement of the extruder, but false in placing the disallowed area on the build surface.
A positive value of the Y-offset should generate a disallowed area on the front of the build surface.

Switching from 2 nozzles to 1, remembers the offset and gives you a disallowed area

Let the machine settings dialog box make it clear that the X,Y,Z length settings are for maximum travel of the axis and not the bed size.

Ghostkeeper:
Fixed in fb1d7bc22363d3a3550748c1180d43ceba444ff0.

maukcc:
Fantastic ! that did the job. I have not tried printing, and will give it a go in a sec.

1 thing that is not correct, is that it still incorporates the disallowed area of extruders that are not there.
Our machine defenition has 4 extruders and standard only 2 are visible, but it respects the disallowed area of the other 2 as well.

Ghostkeeper:
Oh, are you sure they aren't used? Not for the platform adhesion or something? Or infill extruder?

We detect the used extruders here: https://github.com/Ultimaker/Cura/blob/master/cura/Settings/ExtruderManager.py#L416

maukcc:
I can not set them in Cura, as it will crash instantly. See: #1948

so we added it to the definition file:
        "wall_extruder_nr": { "default_value": 0 },
        "wall_0_extruder_nr": { "default_value": 0 },
        "wall_x_extruder_nr": { "default_value": 0 },
        "top_bottom_extruder_nr": { "default_value": 0 },
        "infill_extruder_nr": { "default_value": 0 },

But it still gives me disallowed areas generated by unused extruders.
With unused I do not mean that they are not used for this print job, but they are not selected in "machine settings". So extruder_count is at 2 and not 4.

maukcc:
Also setting the disallowed areas in printer.def.json goes in the wrong direction in Y axis.
If you give it a negative value, you would expect it to come to the front.

Ghostkeeper:
Oops, it seems I forgot one. Fixed: https://github.com/Ultimaker/Cura/commit/319559740dd8c8c2f9ffad637bee43cef9345073

Ghostkeeper:
> With unused I do not mean that they are not used for this print job, but they are not selected in "machine settings". So extruder_count is at 2 and not 4.

[...]

PR #2065, opened 2017-07-04
Cartesio update

maukcc:
Cartesio disalowed areas
Cartesio extruder offsets
Cartesio high quality profiles


nallath:
CURA-4008

resources/extruders/cartesio_extruder_2.def.json
```diff
@@ -14,7 +14,7 @@
             "maximum_value": "3"
         },
         "machine_nozzle_offset_x": { "default_value": 0.0 },
-        "machine_nozzle_offset_y": { "default_value": 60.0 },
+        "machine_nozzle_offset_y": { "default_value": 0.0 },
```
nallath: Are you sure about this? This way the disallowed area's are not displaced when different extruders are used.

maukcc: It is not perfect, but untill #1977 is resolved it is the safest way.

PR #2073, opened 2017-07-05
Refactor gcode flavors to better match the target firmwares

smartavionics:
Due to the incompatibility of legacy Marlin gcode with the gcode that is supported by RepRap firmwares (as used by the Duet family of controllers), this new gcode flavor has been created to support the latter.
I don't think it should be called "RepRap (Duet)" because that ties it to a particular product. Instead, I think it should be "RepRap (RepRap)" so that it can be applicable to all firmwares that are based on the RepRap source (that is until they diverge!)

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -303,9 +303,9 @@
                     "type": "enum",
                     "options":
                     {
-                        "RepRap (Marlin/Sprinter)": "RepRap (Marlin/Sprinter)",
-                        "RepRap (RepRap)": "RepRap (RepRap)",
-                        "RepRap (Volumatric)": "RepRap (Volumetric)",
+                        "RepRap (Marlin/Sprinter)": "Marlin",
+                        "RepRap (Volumatric)": "Marlin (Volumetric)",
+                        "RepRap (RepRap)": "RepRap",
                         "UltiGCode": "Ultimaker 2",
                         "Griffin": "Griffin",
                         "Makerbot": "Makerbot",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -304,6 +304,7 @@
                     "options":
                     {
                         "RepRap (Marlin/Sprinter)": "RepRap (Marlin/Sprinter)",
+                        "RepRap (RepRap)": "RepRap (RepRap)",
                         "RepRap (Volumatric)": "RepRap (Volumetric)",
                         "UltiGCode": "Ultimaker 2",
                         "Griffin": "Griffin",


nickthetait:
This just creates a new menu entry, but does not modify the output G-Code. Was that your intention?
Could you also describe a little bit about the incompatibility between these firmwares?

smartavionics:
Hi Nick, the companion engine PR is https://github.com/Ultimaker/CuraEngine/pull/553, take a look at that and you will see that it enables support for setting acceleration and jerk in the slicer which has not been possible before.

nickthetait:
Ah, gotcha. I was just missing a piece of information. Looks good :+1:

fieldOfView:
I don't think "RepRap (RepRap)" makes a lot of sense. How about something like "RepRap (2017)"?

smartavionics:
Well, to be honest, having "RepRap (Marlin/Sprinter)" which doesn't actually support RepRap firmware makes absolutely no sense whatsoever. In fact, it's completely misleading. What would be ideal would be that "RepRap" flavor produced gcode compatible with RepRap firmware and "Marlin" flavor produced gcode that is compatible with whatever version of Marlin Ultimaker is stuck with. However, it doesn't make sense to attempt to correct the name of the Marlin compatible flavor therefore we need a new name that people will understand that is targetting RepRap firmware.

So, as to your suggestion. "RepRap (2017)" doesn't give any hint of what firmware is supported, it could mean the 2017 version of Marlin, or anything. I think the new name has to indicate that it is targeting RepRap firmware. As long as it does that, I don't really care what it is. How about "Real RepRap"?

nickthetait:
"RepRap (firmware)" ?

fieldOfView:
>  However, it doesn't make sense to attempt to correct the name of the Marlin compatible flavor

Why not? If it better reflects the situation, why not have
* Reprap
* Marlin/Sprinter
* Marlin/Sprinter (Volumetric)
?

smartavionics:
Yes, that would be nice but then you would have to change all the printer definition files that current specify "RepRap (Marlin...)". And any user that had a custom file that contained that would need changing too. I think the breakage would be horrible.

fieldOfView:
No, you would only change the title, the key would stay the same.  There would be no breakage.

Note how the options for the machine_gcode_flavor have a key and a value. The value can be any string and it won't change a thing for what is stored in the definition or what CuraEngine sees. Also note the cute typo in one of the flavors that does not show in Cura but will be with us for the rest of time in CuraEngine and the definition files ;-)

smartavionics:
OK, I understand, so you would be looking to have something like this?

{
                        "Marlin": "RepRap (Marlin/Sprinter)",
                        "Marlin (Volumatric)": "RepRap (Volumetric)",
                        "RepRap": "RepRap (RepRap)",
                        "UltiGCode": "Ultimaker 2",
                        "Griffin": "Griffin",
                        "Makerbot": "Makerbot",
                        "BFB": "Bits from Bytes",
                        "MACH3": "Mach3",
                        "Repetier": "Repetier"
                    },

fieldOfView:
No, the other way around
```
{
"RepRap (Marlin/Sprinter)": "Marlin",
"RepRap (Volumetric)": "Marlin (Volumetric)",
"RepRap (RepRap)": "RepRap",
"UltiGCode": "Ultimaker 2",
"Griffin": "Griffin",
"Makerbot": "Makerbot",
"BFB": "Bits from Bytes",
"MACH3": "Mach3",
"Repetier": "Repetier"
},
```

smartavionics:
Oops, thanks, I will try that.

BagelOrb:
Thanks for the pre-work!
The frontend PR is ready to go!
:D

BagelOrb:
[accepted]

Ghostkeeper:
Devs, see CURA-4014.

fieldOfView:
Devs, you are too late ;-)

Ghostkeeper:
It's still going on here: https://github.com/Ultimaker/CuraEngine/pull/555

BagelOrb:
Devs, you are too late ;)

PR #2079, opened 2017-07-06
Speed up sidebar-settings scrolling

tsaaristo:
Scrolling the sidebar feels rather slow one line at a time (at least when using the scroll wheel; my Ubuntu laptop is able to use the "fling" mechanics with the touchpad).

This PR adjusts the scroll to be three lines/rows at once (the default scroll speed in Windows). If it seems a bit too fast, two lines (`50`) would be my fallback.

Comparison GIFs:


Ghostkeeper:
Devs, see CURA-4026.

Ghostkeeper:
@LipuFei and I are agreeing that it feels a bit too slow right now, and three times as fast seems to be the sweet spot. Let's merge this!

PR #2083, opened 2017-07-07
feat: print thin walls setting (CURA-4023)

BagelOrb:


BagelOrb:
Meant to do this in two commits, but my first commit failed and the code changes became part of the second commit...

resources/definitions/fdmprinter.def.json
```diff
@@ -1125,6 +1125,14 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "fill_outline_gaps": {
```
LipuFei: Shall we use the more or less than same name as the `label`?

LipuFei: Nevermind.

PR #2088, opened 2017-07-09
First layer line width Multiplier

### LINKED ISSUE #1165: No "First layer line width multiplier"
A couple of things:
Maybe there is a way around this, but I haven't found it in the past couple of months.
For most repraps and lower-end printers, or even higher end models and finicky filaments, you pretty much always need a fat first layer to get a good adhesion. Problem is, you can change the width of the bottom infill (solid) independently, but not the line width of the inner or outer walls. This leads to an issue where if you raise them to get the first layer down, the lines will be too big to print some small walls later on.

What I suggest is an isolated multiplier (or set of multipliers) for all lines on the first layer nothing else. This would let us poor folk finally get a reliable layer that won't "string up" and ruin everything, but importantly,  still make printing small walls a reality.

:)

14bitVoid:
Multiplies the line width on just the first layer.

for a description: https://github.com/Ultimaker/CuraEngine/pull/557
Fixes #1165


Ghostkeeper:
Devs, see CURA-4027.

MiLeung:
This setting only seems to have an effect when it's value is 100% or less.

Ghostkeeper:
For me it seems to work fine. I set it to 500% and the lines became humongous.

image

MiLeung:
Right, I meant only seemed to have an effect when actually printing the gcode. Anyway, I was able to alleviate the problem with a thicker first layer (0.3mm).

PR #2092, opened 2017-07-09
Add tooltips to Machine Settings and fix setting diameter

fieldOfView:
This PR adds tooltips to the Machine Settings dialog to clarify some of its settings. It also fixes setting the nominal diameter of the machine.

In 2.6.x, the material diameter set in the Machine Settings would always be overridden by the diameter set in the material container. In this PR, the material container is switched to an appropriate material when changing the material diameter field.

Most tooltips are directly from the property provider, though some tooltips are overridden for clarity in the context of the machine settings dialog.

The code has been restructured for less verbosity, using reusable components instead of repeating the same QML code over and over.

### DIFF
--- a/plugins/MachineSettingsAction/MachineSettingsAction.qml
+++ b/plugins/MachineSettingsAction/MachineSettingsAction.qml
@@ -135,27 +135,23 @@ Cura.MachineAction
                                     property string unit: catalog.i18nc("@label", "mm")
                                     property bool forceUpdateOnChange: true
                                 }
-                            }
 
-                            Column
-                            {
-                                Row
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
+
+                                Label
                                 {
-                                    spacing: UM.Theme.getSize("default_margin").width
-
-                                    Label
-                                    {
-                                        text: catalog.i18nc("@label", "Build Plate Shape")
-                                    }
-
-                                    Loader
-                                    {
-                                        id: shapeComboBox
-                                        sourceComponent: comboBoxWithOptions
-                                        property string settingKey: "machine_shape"
-                                        property bool forceUpdateOnChange: true
-                                    }
+                                    text: catalog.i18nc("@label", "Build Plate Shape")
                                 }
+
+                                Loader
+                                {
+                                    id: shapeComboBox
+                                    sourceComponent: comboBoxWithOptions
+                                    property string settingKey: "machine_shape"
+                                    property bool forceUpdateOnChange: true
+                                }
+
                                 Loader
                                 {
                                     id: centerIsZeroCheckBox
@@ -164,6 +160,7 @@ Cura.MachineAction
                                     property string settingKey: "machine_center_is_zero"
                                     property bool forceUpdateOnChange: true
                                 }
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
                                 Loader
                                 {
                                     id: heatedBedCheckBox
@@ -172,11 +169,10 @@ Cura.MachineAction
                                     property var settingKey: "machine_heated_bed"
                                     property bool forceUpdateOnChange: true
                                 }
-                            }
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
 
-                            Row
-                            {
-                                spacing: UM.Theme.getSize("default_margin").width
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
+                                Item { width: UM.Theme.getSize("default_margin").width; height: UM.Theme.getSize("default_margin").height }
 
                                 Label
                                 {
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Devs, see CURA-4028.

Ghostkeeper:
I just went through the code for CURA-4024 (faulty cache when starting Cura causes missing machines) and fixed all cases where we still used `stack.findContainer(type = "...")`: https://github.com/Ultimaker/Cura/commit/81e07b153071edb783d33f799785125cf58d6cf1

But I left the ones in MachineSettingsAction untouched in order to prevent a merge conflict here. In the unlikely case that this pull request doesn't get merged, we should apply those changes to the MachineSettingsAction as well.

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -221,44 +183,13 @@ Cura.MachineAction
...
-                                    onActivated:
-                                    {
-                                        machineGCodeFlavorProvider.setPropertyValue("value", flavorModel.get(index).value);
-                                        manager.updateHasMaterialsMetadata();
-                                    }
+                                    id: gcodeFlavorComboBox
+                                    sourceComponent: comboBoxWithOptions
+                                    property string settingKey: "machine_gcode_flavor"
+                                    property bool forceUpdateOnChange: true
+                                    property string afterOnActivate: "manager.updateHasMaterialsMetadata()"
```
nallath: Is this really the only way to do this? Dynamic evals scare me if the GUI needs to set them.

fieldOfView: I am open for suggestions.

fieldOfView: I'll have a look at this:
http://doc.qt.io/qt-5/qml-qtquick-loader.html#receiving-signals-from-loaded-objects

nallath: As far as I know you can just check for any property if it exists and call it if it does (pretty much the same as python does it)

fieldOfView: Thanks

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -189,7 +189,7 @@ Cura.MachineAction
                                     sourceComponent: comboBoxWithOptions
                                     property string settingKey: "machine_gcode_flavor"
                                     property bool forceUpdateOnChange: true
-                                    property string afterOnActivate: "manager.updateHasMaterialsMetadata()"
+                                    property string afterOnActivate: manager.updateHasMaterialsMetadata
```
nallath: Nitpicking; Can this still be a string if it's a function reference?

fieldOfView: Thanks for paying attention ;-)

It works as is, but I'll fix it.

nallath:
I also noticed another thing, not quite sure if this is intentional; The rest of cura also gives tooltips when hovering the labels. It took me a few seconds of confusing (and looking at the code if the tooltips worked at all) to figure out that it was only the textarea / dropdown that gives the tooltip.
Is this an intentional change? I don't feel that strongly about it, but i think it would be a tad bit more consistent to keep the behavior the same.

fieldOfView:
Machine Settings is "system-styled", so it cannot use the Setting*.qml controls. In order to get me some more freedom to layout the dialog, I initially separated the label from the control. I will look if I can easily unify them again without loosing too much layout flexibility.

fieldOfView:
@nallath: do you have any idea why I cannot get the "unit" property from a propertyProvider, even if I add it in watchedProperties? It always returns an empty string.

nallath:
Uh. No not really. As far as I know that should work. The only thing I can come up with is that the we use the SettingDefinitionsModel if we want to get the unit otherwise.

fieldOfView:
Ready for another round of scrutiny

plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -164,7 +164,7 @@ Cura.MachineAction
                                 id: gcodeFlavorComboBox
                                 sourceComponent: comboBoxWithOptions
                                 property string settingKey: "machine_gcode_flavor"
-                                property string label: catalog.i18nc("@label", "GCode flavor")
+                                property string label: catalog.i18nc("@label", "Gcode flavor")
```
nickthetait: It seems not to be very standardized. :cry: I personally prefer to use "G-Code".

Using current state of master branch (only the Cura repo) I see the following. This searches EVERYTHING including non-user visible text.

grep -r "GCode" | wc -l
495

grep -r "G-Code" | wc -l
184

grep -r "G-code" | wc -l
175

grep -r "gcode" | wc -l
594

grep -r "g-code" | wc -l
139

fieldOfView: I don't disagree with you. I actually had all occurrences changed to "G-code" locally, but then noticed fdmprinter.def.json (where most of the tooltips come from) is (mostly) standardised in "Gcode". I did not want to make changes to fdmprinter.def.json for this PR.

Both RepRapWiki and Wikipedia use "G-code".

Your greps are probably influenced by i18n files, and to non-userfacing text and variable names.

Ghostkeeper: > noticed fdmprinter.def.json (where most of the tooltips come from) is (mostly) standardised in "Gcode".

Probably due to preferences of the guy who writes these new settings ;) And lack of scrutiny with the guy who has to review that.

I believe I've always used "g-code" and capitalised that at the start of the sentence.

fieldOfView: Does the Ultimaker dictionary say anything about it? It probably should.

PR #2106, opened 2017-07-12
Fix creating a new definition_changes container

fieldOfView:
This PR fixes setting the upgrade option (heated bed) for the UMO, which got broken here:
https://github.com/Ultimaker/Cura/commit/81e07b153071edb783d33f799785125cf58d6cf1#commitcomment-23006662

```global_container_stack.definitionChanges``` always returns a container, ```global_container_stack.findContainer({"type": "definition_changes"})``` does not (because an empty container does not have the type "definition_changes".


nallath:
CURA-4040

PR #1999, opened 2017-06-21
Added additional filters for USB serial devices

rmcolbert:
Sanguinololu boards show up as /dev/cu.SLAB_USBtoUART and /dev/ttySLAB_USBtoUART. Tweaked first glob to pick up USB anywhere after tty and added an uppercase USB for /dev/cu.*USB*.

PR #2107, opened 2017-07-12
Use pyserial to list serial ports

fieldOfView:
This PR replaces the homebrew platform-specific serial port enumeration with functionality provided by pySerial. This prevents us from having to chase all iterations of tty|cu.\*usb|rfcomm\*, and should make serial port detection work with more printers.

This PR is a backport of a fix in the Lulzbot edition of Cura (where it has been extensively tested): https://code.alephobjects.com/rCT14f9814802f4384ef255327afc26ac3156e9539d. It has also been tested here: https://github.com/Ultimaker/Cura/pull/1999#issuecomment-314276734. It is an alternative to https://github.com/Ultimaker/Cura/pull/1999

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -91,6 +91,8 @@ class MachineManager(QObject):
 
         self._printer_output_devices = []
         Application.getInstance().getOutputDeviceManager().outputDevicesChanged.connect(self._onOutputDevicesChanged)
+        # There might already be some output devices by the time the signal is connected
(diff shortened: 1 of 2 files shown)


nallath:
CURA-4038

fieldOfView:
It would be nice if this could be tested in the beta. If it does not work out it could be reverted before the final release.

nallath:
Derp. I messed something up here. This should have been merged already. My bad.

PR #2108, opened 2017-07-12
Hide the extruder button for Custom FDM Printers with 1 extruder

fieldOfView:
This PR hides the extruder button in the toolbar if a Custom FDM Printer has its number of extruders set to 1 in Machine Settings. Before this PR, there would be a single extruder button to set the extruder of the currently selected object(s). For a single extrusion printer, this button serves no purpose.


nallath:
CURA-4039

PR #2109, opened 2017-07-12
Add platform mesh for the 101Hero & updated printer profile to use it.

rmcolbert:


nickthetait:
Do you work for the kickstarter/company or are you a user of the 101Hero?

rmcolbert:
I am a user of the 101Hero. The base profile for the printer already existed. I only added the mesh and updated the definition file to point to it.

Ghostkeeper:
Devs, see CURA-4046.

fieldOfView:
Note: the way to prevent commits from one feature to end up in the PR for another commit is to create a "feature branch" per feature you are working on. As a rule of thumb it is never a good idea to create a PR from your master branch.

nallath:
To keep the history a bit cleaner, i added the platform in a separate commit. I also saved the platform as binary, which decreases the size a bit. See dee09e15aed7ff4924188130ef9927cc262fcabd

Issue #1748, opened 2017-04-26
Text vs Label


fieldOfView:
Recently a number of changes were made to change Label components into Text components. I have yet to find the logic in which components were changed and which were not. Eg in https://github.com/Ultimaker/Cura/commit/2d14052f4e41f4346aa429439d46fe68182e9395 5 Labels were changed into a Text, but the 6th (which is a direct sibbling) was not.

Unfortunately the change of Label to Text in the parts of the interface mean that the text now looks asif it is rendered in another font:
![image](https://cloud.githubusercontent.com/assets/143551/25426097/58c86304-2a6e-11e7-8d64-d0d7696907cf.png)
If you look at the full sized screenshot, you will see that the texts below the "Connect via Network" button looks different than all the other text in the window. It is thicker (almost bold), looks slightly more condensed and smaller. 

It turns out the text is not rendered in another font, but using a different font rendering. This is how PyQt defines a Label:
https://github.com/pyqt/python-qt5/blob/master/PyQt5/qml/QtQuick/Controls/Label.qml#L79
The thing that sets the Label apart from Text are the color and the renderType. By using Text instead of Label, the texts are no longer rendered as Text.NativeRendering (but as Text.QtRendering instead).

Isn't the whole idea of the unthemed dialogs that they *do* use native rendering to fit in with the system theme?

I would argue that it is much better to use the same text rendering for all elements, or at least for all elements in one "theme".

The commit references "ugly fonts". Is there a screenshot somewhere?

nallath:
In some cases the rendering seems to fail for some operating systems. I don't have any screenshots, but the letters get warped somewhat.

fieldOfView:
I think I have seen what you describe, but this could always be fixed by making sure Labels were positioned on whole pixel positions. I'll look for an example to confirm we're talking about the same thing.

awhiemstra:
More important in my opinion, Text does not use platform style at all. This results in the configuration dialog now looking like this on my machine:

![screenshot_20170426_141621](https://cloud.githubusercontent.com/assets/9590386/25434057/1e434a88-2a8b-11e7-8cba-17734332d330.png)

fieldOfView:
Ah, yes I was afraid that might be the case too, since Label specifically sets the (default) color to the system color.

Here's an occurance of "ugly font rendering", is that what we're talking about? See the second column:
[...]

PR #2114, opened 2017-07-14
Fix system-themed text rendering

fieldOfView:
This PR fixes text rendering in system-themed qml by rounding fractional widths and heights.

Depending on window/dialog sizes, Label elements could be rendered at fractional pixel offsets. The Text.NativeRendering renderType does not handle that nicely, leading to ugly font rendering on Windows and OSX. In an attempt to fix this, some Labels were switched to Text (which uses Text.QtRendering rendertype). Unfortunately this leads to font-rendering that is distinctively non-native on Windows. Furthermore, Text does not use the system-themed colors used by Label, leading to unreadable text in dark-themed systems (black on black).

By rounding all widths/heights to whole pixels, Text.NativeRendering can be used without rendering problems.

Depends on https://github.com/Ultimaker/Uranium/pull/266
See #1748 for discussion.

Issue on dark system themes:
image

### DIFF
--- a/plugins/3MFReader/WorkspaceDialog.qml
+++ b/plugins/3MFReader/WorkspaceDialog.qml
@@ -87,18 +87,18 @@ UM.Dialog
                 {
                     text: catalog.i18nc("@action:label", "Printer settings")
                     font.bold: true
-                    width: parent.width / 3
+                    width: (parent.width / 3) | 0
                 }
                 Item
                 {
                     // spacer
                     height: spacerHeight
-                    width: parent.width / 3
+                    width: (parent.width / 3) | 0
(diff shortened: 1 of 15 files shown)


nallath:
CURA-4048

nallath:
The problem of this is that it's going to be easy to mess it up; If we forget somewhere, you're still going to have messed up fonts.

nallath:
Won't setting the renderType to Text.QtRendering fix the issue?

fieldOfView:
Yes, but that looks ugly on windows. It is a "paardenmiddel", and probably performs worse than native text rendering too.

PR #2129, opened 2017-07-17
Fix recently broken setting of gcode area value property on focus loss.

smartavionics:


fieldOfView:
Thanks. Too much late-night refactoring.

nallath:
As this is such a small fix, i'm not even going to bother with creating a jira issue.

PR #2141, opened 2017-07-20
Quarter cubic infill

BagelOrb:
frontend setting for https://github.com/Ultimaker/CuraEngine/pull/564

I  renamed tetrahedral to octet because that's its proper name.
Otherwise it would be very confusing because Quarter Cubic also contains a lot of tetrahedra


BagelOrb:
Devs, see CURA-4071

PR #2142, opened 2017-07-21
Force first

BagelOrb:
force a specific model to be the first to be printed with the current extruder on that layer

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2929,6 +2929,17 @@
                     "settable_per_extruder": true,
                     "settable_per_meshgroup": true
                 },
+                "force_first":
+                {
+                    "label": "Start Layer with this Mesh",
+                    "description": "Start each layer by printing this model rather than other models. In multi-extrusion this model will be printed first when we switch to the extruder with which this model is printed. This setting doesn't affect when we switch to a different extruder.",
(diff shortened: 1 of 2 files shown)


nickthetait:
An idea on what to name this feature: First item printed

Ghostkeeper:
Devs, see CURA-4073.

resources/definitions/fdmprinter.def.json
```diff
@@ -2929,6 +2929,17 @@
                     "settable_per_extruder": true,
                     "settable_per_meshgroup": true
                 },
+                "force_first":
+                {
+                    "label": "Start Layer with this Mesh",
+                    "description": "Start each layer by printing this model rather than other models. In multi-extrusion this model will be printed first when we switch to the extruder with which this model is printed. This setting doesn't affect when we switch to a different extruder.",
```
Ghostkeeper: You should add a note here how this behaves with respect to the start X/Y position of a layer, namely that this overrides the start X/Y (when applicable). Also in the start X/Y we should put a message that the setting may get overwritten by this setting.

BagelOrb:
I thought we were not going to do this PR.

Moreover, I implemented this 2 months ago. I don't really feel like working on it any more.

nallath:
If so; then close it.

PR #2160, opened 2017-07-25
LayerView: Set color for comp. mode text

thopiekar:
Before it was using the system's default, which is not that nicely visible, if you are using a dark theme.

Now it looks like this:
image

Belongs to no JIRA issue.


nallath:
Can't be bothered to make a proper Jira ticket for this. Looks fine!

PR #2162, opened 2017-07-25
FABtotum Personal Fabricator

krios-fabteam:
The purpose of this request is to add the support for the  FABtotum Personal Fabricator printer


PaoloP74:
@krios-fabteam: You should convert obj to stl. The size would be considerably lower.
.obj are no longer used.

fieldOfView:
Even when converted to binary STL, the platform weighs in at 1.9 MB which is way too large in my opinion. Can you create a less detailed version? Note that OBJ is fine, especially if you want to use UV coordinates. Just try to keep the file under 1 MB.

The quality profiles seem to reference (xml) materials that you did not include in the PR.

krios-fabteam:
@monolith-ultimaker  What is missing?

fieldOfView:
@krios-fabteam don't pay too much attention to the continuous integration thing. It has a tendency to not work in ways that are very opaque to us non-Ultimaker contributors.

PaoloP74 is right, the xml materials should go into a separate PR on the fdm_materials repo. I re-read my earlier response, and it does seem to suggest to add them here. Sorry about that.

krios-fabteam:
@fieldOfView i've just a created a PR  on the fdm_materials repo

Ghostkeeper:
Please also remove the XML materials from this pull request (and undo the change to .gitignore) then :)

Ghostkeeper:
Devs, see CURA-4128.

PaoloP74:
What is CURA-4128? A new branch? A reference for an external link?

fieldOfView:
CURA-4128 is a reference to the internal issue tracker that is only accessible by Ultimaker developers.

Ghostkeeper:
Also, that message means that your pull request is now finally in our sprint (sorry that it took so long). We'll be evaluating it now, but added machine definitions like this are usually a no-brainer to us so it'll probably just get accepted.

resources/quality/fabtotum/fabtotum_abs_black_high.inst.cfg
```diff
@@ -0,0 +1,25 @@
+[general]
+version = 2
+definition = fabtotum
+name = High Quality
+
+[metadata]
+type = quality
+setting_version = 1
```
LipuFei: The `setting_version` now needs to be 2. Could you change all of them to `2`?

krios-fabteam:
What is missing to approve the merge?

nallath:
Time on our end to actually test it. It will be added, it's just taking a bit more time. A number of our devs were / are going to be on holidays.

alekseisasin:
PaoloP74, I want to ask you make some changes: The fabtotum.def.json has attribute "category": "FABtotum", can you set it to the "Other", we keep added printers in separate group. 
Also after loading your printer type I see for the title "Head" drop box with empty value empty, if we don't have any type you can remove it at all, you can check other examples.

krios-fabteam:
@alekseisasin my last commit should have added your requests

PaoloP74:
@alekseisasin I'm not a FabTeam Developer but an independent maker. Ask @krios-fabteam.

fieldOfView:
Is it really necessary to have a set of qualities for each specific material/color combination?
Let me rephrase that: It is not necessary to have a set of qualities for each specific material/color combination. Instead it should be possible to create profiles for ```material = generic_pla_fabtotum``` and it should work for all PLA materials.

alekseisasin:
Looks now fine, can you also check "Add fabtotum materials" pull request, then I will merge them together.

alekseisasin:
Yes, it is, it will let set color of material to the displayed detail, what
makes it more attractive for the users than have one generic (gray) color.

fieldOfView:
@alekseisasin: no, it doesn't. The UM3 family does not have different quality profiles for each color material, but it has quality profiles for each of the generic profiles (for each variant).

If you merge this PR as is, you are you are setting a precedent and you will end up with a gazillion unnecessary profiles because all sorts of printers will have all sorts of quality profiles for all colors for all materials. That will be hell to maintain. And it already takes much too long to start Cura.

nallath:
Merging this in the current state is indeed not a good idea.

krios-fabteam:
@fieldOfView @alekseisasin Hi guys, 
did not understand if i have to do more changes to the PR for its approval

nallath:
@krios-fabteam; You have a load of profiles for every color of PLA out there, but all of them have the exact same settings. So instead, you should do the same as we already did for our PLA profiles; just make a generic one. Cura will default to generic if it can't find a specific profile.

krios-fabteam:
So i have to create a generic PLA profile but where to save it? in `resources/quality/fabtoum` or in `fmd_materials` repository?
I'm looking on your files and folders but i can't figure it out

alekseisasin:
If you created a generic profile then you should update both of them. Remove duplicated/unused material types and set generic type. 
You mentioned that you created PLA generic, but I see you use many times identical ABS profile. Would you please create same generic for ABS.

krios-fabteam:
I didn't create the PLA generic profile yet, could you link me to yours generic PLA profile so  I can use it for inspiration and see how it is done

Ghostkeeper:
Here's an example from the Ultimaker 2+: https://github.com/Ultimaker/Cura/blob/master/resources/quality/ultimaker2_plus/pla_0.4_normal.inst.cfg

The important line is this:

    material = generic_pla_ultimaker2_plus_0.4_mm

Since the coloured materials have the same material type as Generic PLA they will be used for every colour.

krios-fabteam:
thanks @Ghostkeeper 
but where should be located my generic pla profile, i didn't find `generic_pla_ultimaker2_plus_0.4_mm`

Ghostkeeper:
Yeah, it's quite confusing. Sorry. That actually points to a profile inside the `generic_pla` file, since such an XML file may contain different settings for different printers/variants. Basically, you could specify either `generic_pla` or a more specific sub-profile.

I suggest that you change your line to this:

    material = generic_pla

Ghostkeeper:
Now you have a lot of profile files that are exactly the same, so you can remove the duplicates :)

krios-fabteam:
Thanks guys for your patience and support

krios-fabteam:
Do you think this PR will be available with next Cura update?

PaoloP74:
Cura 2.7 is out.
Too late.

Ghostkeeper:
Yeah in general we don't add new features after the beta is released, so no new printer/material definitions either.

PaoloP74:
3.0? Spoiler? ;-)

PR #2164, opened 2017-07-26
Add settings for skipping some zags

LipuFei:
CURA-3957


resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "support_skip_some_zags":
+                {
+                    "label": "Skip Some ZigZags Connections",
```
BagelOrb: 1. It's not zalled 'ZigZags', but 'Zig Zag'
2. In the near future we're going to make any setting zig-zaggable; I already made the setting as part of my cross infill. It's called 'Connect Infill Lines', so it would be logical if a similar setting for support would be called 'Connect Support Infill Lines'.
3. This setting doesn't mention it's about support

Solution:
rename to 'Skip Some Support Infill Line Connections'

resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
...
+                    "description": "Skip some ZigZags connections to make the support structure easier to break.",
```
BagelOrb: 'ZigZags' -> 'support line'

'easier to break' -> 'easier to break away'

PS: note that the common term for this type of support is 'break-away support' in contrast to soluble support.

resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
...
+                    "type": "bool",
```
BagelOrb: I think you can leave out this whole setting by making a special case for support_zag_skip_count. You could make it such that `if support_zag_skip_count == 0 then don't skip any connections`.
The description of support_zag_skip_count would then also need to state that this works that way.

Please discuss with Ruben what you think is best.

resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
...
+                    "default_value": false,
+                    "enabled": "support_enable and (support_pattern == 'zigzag')",
+                    "limit_to_extruder": "support_infill_extruder_nr",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "support_zag_skip_count":
+                {
+                    "label": "ZigZag Connection Skip Count",
```
BagelOrb: rename to 'Support Line Connection Skip Cycle'

'Skin Count' suggests that this setting governs the amount of connections skipped, while it is about 1 in how many connections to skip.

Perhaps not 'skip cycle', but 'skip period' or 'skip periodicity'. Please discuss with Ruben.

resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
...
+                    "label": "Skip Some ZigZags Connections",
+                    "description": "Skip some ZigZags connections to make the support structure easier to break.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": "support_enable and (support_pattern == 'zigzag')",
+                    "limit_to_extruder": "support_infill_extruder_nr",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "support_zag_skip_count":
```
BagelOrb: Maybe it's better to make a parent setting `support_zag_skip_frequency` in 'mm^-1'.
The average amount of connections to be skipped per mm.

Breaking away support depends more on the size of the support chunk than on the amount of not-removed connections.
When changing the support infill density it would be nice to automatically also change the support_zag_skip_count by its inheritance function.

support_zag_skin_count could then have an inheritence function like `max(1, round(1 / support_zag_skip_frequency / support_line_distance))`

resources/definitions/fdmprinter.def.json
```diff
@@ -3379,6 +3379,30 @@
...
+                {
+                    "label": "ZigZag Connection Skip Count",
+                    "description": "Skip one in every N connection lines to make the support structure easier to break.",
```
BagelOrb: 'to break' --> 'to break away'

resources/definitions/fdmprinter.def.json
LipuFei: @BagelOrb I discussed with Ruben on this one. I thought it is not very clear what skip count = 0 means, so I made this boolean.

resources/definitions/fdmprinter.def.json
LipuFei: Discussed with Ruben and we think if we create a parent setting "support_zag_skip_count_per_mm" which is the "frequency" mentioned above, then this one can be called "Support Line Connection Skip Chunk Size" or similar since this will be automatically calculated from the "frequency".

resources/definitions/fdmprinter.def.json
LipuFei: Fixed in 26ec06e

resources/definitions/fdmprinter.def.json
LipuFei: Fixed in 26ec06e

resources/definitions/fdmprinter.def.json
LipuFei: Fixed in 26ec06e

resources/definitions/fdmprinter.def.json
LipuFei: Added in 02ba46b

resources/definitions/fdmprinter.def.json
BagelOrb: Okay.

resources/definitions/fdmprinter.def.json
BagelOrb: This is done 👍

resources/definitions/fdmprinter.def.json
```diff
@@ -3404,27 +3404,44 @@
...
                     "enabled": "support_enable and (support_pattern == 'zigzag') and support_skip_some_zags",
                     "limit_to_extruder": "support_infill_extruder_nr",
                     "settable_per_mesh": false,
-                    "settable_per_extruder": true
+                    "settable_per_extruder": true,
+                    "children":
+                    {
+                        "support_zag_skip_count":
+                        {
+                            "label": "Support Chunk Size",
```
BagelOrb: I think this new terminology ('support chunks') is good, but it is now applied inconsequently.
I propose:
Skip Some Support Line Connections -> Break Up Support In Chunks   (not sure about this one, please discuss with Ruben)
Support Line Skip Frequency -> Support Chunk Size
Support Chunk Size -> Support Chunk Line Count

resources/definitions/fdmprinter.def.json
BagelOrb: See new comment

resources/definitions/fdmprinter.def.json
LipuFei: Fixed in 8029efdffbfa610d213a236d662d0ca6ba52db44

resources/definitions/fdmprinter.def.json
```diff
@@ -3381,27 +3381,45 @@
...
+                "support_skip_zag_per_mm":
                 {
-                    "label": "ZigZag Connection Skip Count",
-                    "description": "Skip one in every N connection lines to make the support structure easier to break.",
-                    "type": "int",
-                    "default_value": 6,
-                    "minimum_value": "1",
-                    "minimum_value_warning": "3",
+                    "label": "Support Chunk Size",
+                    "description": "To skip one connection line in every N mm to make the support structure easier to break away.",
```
BagelOrb: You shouldn't start a description with "To do something"

Suggestion: "Leave out a connection between support lines once every N millimeter to make the support structure easier to break away.",

resources/definitions/fdmprinter.def.json
```diff
@@ -3381,27 +3381,45 @@
...
+                    "type": "float",
+                    "unit": "mm",
+                    "value": "support_line_distance * 5",
```
BagelOrb: I think a static value makes more sense here.
Maybe a default value of 10 mm would be better.

PR #2176, opened 2017-07-29
Add settings to specify z offset for initial layer and number of layers over which offset reduces to zero

smartavionics:
The rational behind this PR is that some filaments adhere to the build plate better if the extruder flies a little higher than the layer height. For example, when printing a 0.4mm thick initial layer using PETG it can ripple 
but if I increase the z height to 0.5mm (i.e. an offset of +0.1mm) the layer prints with no ripples and excellent adhesion to the build plate.
The PR includes another setting which is the number of layers over which the offset is tapered to zero. A value of 0 means that the offset is applied to all layers in the print.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4095,8 +4095,7 @@
                     "unit": "mm",
                     "type": "float",
                     "default_value": 0,
-                    "minimum_value": "-layer_height_0 * 0.5",
-                    "minimum_value_warning": "-layer_height_0 * 0.2",
+                    "minimum_value_warning": "0",
                     "maximum_value_warning": "layer_height_0",
                     "enabled": "resolveOrValue('adhesion_type') != 'raft'",
                     "settable_per_mesh": false,
@@ -4108,7 +4107,7 @@
                     "description": "When non-zero, the Z offset is reduced to 0 over that many layers. A value of 0 means that the Z offset remains constant for all the layers in the print.",
                     "type": "int",
                     "default_value": 0,
-                    "minimum_value": "0 if z_offset_layer_0 <= 0 or z_offset_taper_layers == 0 else max(1, round(5 * z_offset_layer_0 / layer_height + 0.5))",
+                    "minimum_value": "0",
                     "enabled": "resolveOrValue('adhesion_type') != 'raft' and z_offset_layer_0 != 0",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4090,7 +4090,7 @@
                 },
                 "z_offset_layer_0":
                 {
-                    "label": "Initial Layer Z Offset.",
+                    "label": "Initial Layer Z Offset",
                     "description": "The extruder is offset from the height of the first layer by this amount. It can be positive or negative. Some filament types adhere to the build plate better if the extruder is raised slightly.",
                     "unit": "mm",
                     "type": "float",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4088,6 +4088,28 @@
                     "settable_per_extruder": true,
                     "limit_to_extruder": "adhesion_extruder_nr"
                 },
+                "z_offset_layer_0":
+                {
+                    "label": "Initial Layer Z Offset.",
+                    "description": "The extruder is offset from the height of the first layer by this amount. It can be positive or negative. Some filament types adhere to the build plate better if the extruder is raised slightly.",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Devs, see issue CURA-4171.

alekseisasin:
In certain situation this setting will help improve the printing quality, but this setting also allows to the user move lower than Z = 0, which will/may lead to printer damage. In this case we should set "allowed range" or "warning" if the value is lower than min allowed.

smartavionics:
I have added some constraints to the setting values based on not allowing the gaps between the layers to be less than 80% of the normal layer height. Please re-test and let me know if further changes are required. Thanks.

resources/definitions/fdmprinter.def.json
```diff
@@ -4088,6 +4088,31 @@
...
                     "limit_to_extruder": "adhesion_extruder_nr"
                 },
+                "z_offset_layer_0":
+                {
+                    "label": "Initial Layer Z Offset",
+                    "description": "The extruder is offset from the normal height of the first layer by this amount. It can be positive (raised) or negative (lowered). Some filament types adhere to the build plate better if the extruder is raised slightly.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 0,
+                    "minimum_value": "-layer_height_0 * 0.5",
```
Ghostkeeper: There's no technical limitation of why this setting would have a minimum value, right?

resources/definitions/fdmprinter.def.json
```diff
@@ -4088,6 +4088,31 @@
...
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": false
+                },
+                "z_offset_taper_layers":
+                {
+                    "label": "Z Offset Taper Layers",
+                    "description": "When non-zero, the Z offset is reduced to 0 over that many layers. A value of 0 means that the Z offset remains constant for all the layers in the print.",
+                    "type": "int",
+                    "default_value": 0,
+                    "minimum_value": "0 if z_offset_layer_0 <= 0 or z_offset_taper_layers == 0 else max(1, round(5 * z_offset_layer_0 / layer_height + 0.5))",
```
Ghostkeeper: What's the reason for this minimum value of `5 * z_offset_layer_0 / layer_height + 0.5`? Unless it is physically impossible, I think just a minimum value of 0 would be fine.

resources/definitions/fdmprinter.def.json
smartavionics: @alekseisasin pointed out that it was possible for the user to specify values that would cause problems due to the inter-layer gaps being too small. So I tried to guard against that.

resources/definitions/fdmprinter.def.json
smartavionics: Same rational as above comment. Talk to @alekseisasin and decide which way you want to go.

alekseisasin:
smartavionics, my idea was to have constraints or limits, especially for lowest values. After discussing current settings we came to conclusion that we don't set minimum values for both settings, but still need to have warnings. 
Can you remove "minimum_value" from both settings and set z_offset_layer_0 "minimum_value_warning" only if value lower than zero.

smartavionics:
OK, I did that but I left the minimum_value for z_offset_taper_layers as 0 as you don't want that to be negative.

BagelOrb:
Hey Mark,

We had some internal discussion about this functionality and now we think the reason this feature works is probably a bit different than you think.

The rippling effect you are seeing might be due to over-extrusion on the first layer. This overextrusion might be the effect of an oversimplified model of an extruded line which Cura uses.

line_rounding_initial_layer svg

Because a printed line is never square (the simplified model Cura uses) the extruded lines are wider than they are computed to be. The difference is presumably larger for larger layer heights.

By printing the initial layer from a higher height than the line height of the simplified model you are effectively lowering the flow of the initial layer.
By lowering the flow the width of the line is altered until it falls more within the green square.

Note that by printing the initial layer from a higher height you are actually changing the thickness of the layer. It's not possible to print the initial layer such that it is actually hovering above the build plate ;)

You should therefore just increase the initial layer height, because that's what you want to happen.

In order to get the same flow-rate we think it would be best to introduce a new setting instead of the ones in this PR: Initial Layer Flow, which you would then set to something like 80%, based on the values you are currently using for the offset.

This new proposal is different from this PR in a couple of important ways:
- The new functionality could be configured per extruder
- There is no tapering

I'm not sure about why you would want tapering.
Setting it to zero means that your print will be higher than your input model by `z_offset_layer_0` mm.

Moreover, tapering doesn't make sense if the cause of the troubles is the difference between the initial layer thickness and the normal thickness.

-------------------------------------------
TL;DR

Please instead of this PR make a new setting **Initial Layer Flow**.

smartavionics:
Hi Tim, Understood, closing...

PR #2177, opened 2017-07-29
JSOn feat: raft smoothing setting (CURA-3854)

BagelOrb:


resources/definitions/fdmprinter.def.json
```diff
@@ -4102,6 +4102,19 @@
...
                     "settable_per_extruder": true
                 },
+                "raft_smoothing":
+                {
+                    "label": "Raft Smoothing",
+                    "description": "This setting control how much inner corners in the raft outline are rounded. Inward corners are rounded to a semi circle with a radius equal to the value given here. This setting also removes holes in the raft outline which are smaller than such a circle.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 5,
+                    "minimum_value_warning": "raft_interface_line_width",
```
LipuFei: @BagelOrb I think it should also have a minimum value 0. What do you think?

LipuFei: Fixed it for you in 63ea94068

PR #2180, opened 2017-07-30
Added cool_fan_min_at_height and cool_fan_min_layer settings.

smartavionics:
These specify the layer that the fans will start at the minimum speed. Below this layer, the
fans are off. Defaults to 0 so it will not change the existing behaviour unless a non zero
value is specified.

Engine stuff is at https://github.com/Ultimaker/CuraEngine/pull/574.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -3181,7 +3181,7 @@
                 "cool_fan_speed_0":
                 {
                     "label": "Initial Fan Speed",
-                    "description": "The speed at which the fans spin at the start of the print. In subsequent layers the fan speed is gradually increased up to the layer corresponding to Regular Fan Speed at Height.",
+                    "description": "The speed at which the fans spin when Initial Fan Speed at Height is reached. In subsequent layers the fan speed is gradually increased up to the layer corresponding to Regular Fan Speed at Height.",
                     "unit": "%",
                     "type": "float",
                     "minimum_value": "0",
@@ -3190,6 +3190,32 @@
                     "enabled": "cool_fan_enabled",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
+                },
+                "cool_fan_min_at_height":
+                {
+                    "label": "Initial Fan Speed at Height",
+                    "description": "The height at which the fans spin at Initial Fan Speed. The fans are off below this height.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 0,
+                    "value": "0",
+                    "minimum_value": "0",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true,
+                    "children":
+                    {
+                        "cool_fan_min_layer":
+                        {
+                            "label": "Initial Fan Speed at Layer",
+                            "description": "The layer at which the fans spin at Initial Fan Speed. If Initial Fan Speed at Height is set, this value is calculated and rounded to a whole number.",
+                            "type": "int",
+                            "default_value": 0,
+                            "minimum_value": "0",
+                            "value": "max(0, int(math.floor((cool_fan_min_at_height - resolveOrValue('layer_height_0')) / resolveOrValue('layer_height')) + 2))",
+                            "settable_per_mesh": false,
+                            "settable_per_extruder": true
+                        }
+                    }
                 },
                 "cool_fan_full_at_height":
                 {
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -3209,9 +3209,9 @@
                             "label": "Initial Fan Speed at Layer",
                             "description": "The layer at which the fans spin at Initial Fan Speed. If Initial Fan Speed at Height is set, this value is calculated and rounded to a whole number.",
                             "type": "int",
-                            "default_value": 0,
-                            "minimum_value": "0",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Seems like a nice feature. Our materials & processing team (who make and optimise Cura's Ultimaker profiles, among other things) are lusting for better control for the fan. Step in a good direction if you ask me.

Devs, see issue CURA-4172.

BagelOrb:
Sorry. We've been slacking on handling PRs lately. There's so many! :D

resources/definitions/fdmprinter.def.json
```diff
@@ -3191,6 +3191,32 @@
...
+                    "label": "Initial Fan Speed at Height",
+                    "description": "The height at which the fans spin at Initial Fan Speed. The fans are off below this height.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 0,
+                    "value": "0",
+                    "minimum_value": "0",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true,
+                    "children":
```
BagelOrb: Please add
````
"maximum_value": "(cool_fan_full_layer - 1) * layer_height + layer_height_0",
"enabled": "cool_fan_speed_min != cool_fan_speed_0",
````

File: resources/definitions/fdmprinter.def.json [hunk 2]
```diff
@@ -3191,6 +3191,32 @@
...
+                    {
+                        "cool_fan_min_layer":
+                        {
+                            "label": "Initial Fan Speed at Layer",
+                            "description": "The layer at which the fans spin at Initial Fan Speed. If Initial Fan Speed at Height is set, this value is calculated and rounded to a whole number.",
+                            "type": "int",
+                            "default_value": 1,
+                            "minimum_value": "1",
```
BagelOrb: Please add
````
"maximum_value": "cool_fan_full_layer",
"enabled": "cool_fan_speed_min != cool_fan_speed_0",
````

File: resources/definitions/fdmprinter.def.json [hunk 3]
```diff
@@ -3191,6 +3191,32 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "cool_fan_min_at_height":
+                {
+                    "label": "Initial Fan Speed at Height",
+                    "description": "The height at which the fans spin at Initial Fan Speed. The fans are off below this height.",
```
BagelOrb: What happens when cool_fan_full_layer equals cool_fan_min_layer??
Please add this to the description.

BagelOrb:
The functionality in this commit is a bit weird.

In earlier versions of Cura (Cura 15?) the initial fan speed couldn't be controlled. Instead it was always zero.
A fan speed of zero is as arbitrary as any other fan speed, so we made it controllable via the setting Initial Fan Speed.

Now in this PR you introduce a new fan speed of zero before the initial fan speed. Than means that the Initial Fan Speed is not initial any more - the print doesn't start with the Initial Fan Speed any more, so the name doesn't make sense any more.

Moreover you reintroduce a Fan Speed of zero, which was the main reason we introduced Initial Fan Speed in the first place.

If you want the first layer to start at a Fan Speed of zero I suggest you just set the Initial Fan Speed to zero.

If you want the fan speed to stay at zero for a couple of layers I think you should change the functionality you've implemented slightly.
Here's what I propose:
fan_speed_proposal

So instead of leaving the Fan Speed at zero before the Initial Fan Speed Height, leave it at the Initial Fan Speed.

smartavionics:
Tim, what I would like is to be able to specify that the fans to be off until a certain height is reached. At that height the fans should come on at some minimum rate which is gradually increased until the regular fan speed height is reached. The blue line above is the right shape for me!

BagelOrb:
That doesn't make sense to me. Why would you want a sudden change from zero to the initial fan speed at a given layer? What prompts this abrupt change in required fan speed?

smartavionics:
> That doesn't make sense to me. Why would you want a sudden change from zero to the initial fan speed at a given layer? What prompts this abrupt change in required fan speed?

OK, so I definitely want to have a region of the bottom of the print where the part cooling is completely off. I believe this reduces the chance of warping of corners, etc. Once up to a safe height, I want to be able to start feeding the fan in. Starting the fan at 0% is one possibility but some fans don't like being run at very low speeds (remember, not everyone has a Ultimaker) so I think it's better to allow people to be able to specify the initial fan speed (say, 10%, or whatever). That's my thinking behind the new settings.

BagelOrb:
> OK, so I definitely want to have a region of the bottom of the print where the part cooling is completely off. 

My thinking was that you should do that by specifying that the Initial Fan Speed is 0%

> some fans don't like being run at very low speeds

Hmm this makes it rather difficult.

According to my previous reasoning a fan speed of zero is as arbitrary as any other value. The reason you want a fan speed of zero for the first couple of layers might be exactly the same as somebody else wants a fan speed of 10% on those initial layers.

I think for now people with such fan troubles can just set the `cool_fan_full_layer` to be less than 10 more than `cool_fan_min_layer` so that the values below 10% will never be hit.

I think normal sequences of fan speeds would be like this:
```
0, 0, 0, 25, 50, 75, 100
30, 60, stay on 60
20, 20, 20, 20, 40, 60, 80, 100
0, 0, 0, 0, 20, 40, 60, 80, 100
```

The problem of less than 10% doesn't occur in any such normal sequence.

Ghostkeeper:
The jump in fan speed (as in the "current" line of BagelOrb's drawing) can also be achieved with the Tweak at Z post-processing script.

smartavionics:
Closing due to differences in opinion as to how it should work. I like it as it is and will continue to use it.

PR #2200, opened 2017-08-03
Adapt camera image to available space

fieldOfView:
This PR makes sure the camera image adapts to the available size in the window. Cura 2.6 did the same thing, but this was missing since the refactor of the camera image in Cura 2.7.

Steps to reproduce the issue: 
switch to the monitor view and make the window as small as it will go.
Result:
the image will be cut off, and could even overlap the topbar (making it hard to return to the "prepare" tab. 

The issue is even more of a problem for users of the OctoPrint plugin, where the camera image can be much larger than the UM3 image (eg 1920x1080). A separate fix for the OctoPrint plugin is in preparation, but it relies on this PR.

It would be nice if this could be cherry picked for Cura 2.7.

### DIFF
--- a/cura/PrinterOutputDevice.py
+++ b/cura/PrinterOutputDevice.py
@@ -144,6 +144,9 @@ class PrinterOutputDevice(QObject, OutputDevice):
         return self._control_item
 
     def _createControlViewFromQML(self):
+        if not self._control_view_qml_path:
+            return
+
         path = QUrl.fromLocalFile(self._control_view_qml_path)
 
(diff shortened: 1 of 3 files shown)


fieldOfView:
Humble bump

alekseisasin:
Devs, see CURA-4169

fieldOfView:
Is there any chance this could get cherrypicked into 2.7?

alekseisasin:
Yes, it will be added into 2.7

PR #2222, opened 2017-08-08
Update kemiq_q2_beta.def.json

kemiq3d:
Removed tabs and added spaces.


Ghostkeeper:
You didn't remove any tabs but just added spaces at the end of two lines. As it stands now, this change only makes those two lines disagree with our code style so I don't think we'll merge this one.

The current file in master doesn't have any tabs so I think that's all right.

PR #2223, opened 2017-08-08
Update kemiq_q2_gama.def.json

kemiq3d:
Removed tabs and added spaces.

### DIFF
--- a/resources/definitions/kemiq_q2_gama.def.json
+++ b/resources/definitions/kemiq_q2_gama.def.json
@@ -15,9 +15,8 @@
     },
 
     "overrides": {
-        "machine_name": { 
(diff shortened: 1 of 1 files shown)


kemiq3d:
We removed all the tab's and replaced with spaces. Why it failed again this time?

fieldOfView:
As far as I can tell the build failure has nothing to do with your commits, but the current "master" branch has issue (so any branches that are currently built that do not specifically fix the issues with "master" will fail)

Ghostkeeper:
These changes seem to already have been applied on the original PR. That PR has already replaced the tabs separately, so this PR is not needed any more.

PR #2233, opened 2017-08-09
Fix readme section about making definitions

fieldOfView:
This PR removes the link to definition generator and some incorrect instructions. The definition generator produces definitions that don't work with Cura 2.3 and newer. It also adds some instructions for multi-extrusion printers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Third party plugins
 
 Making profiles for other printers
 ----------------------------------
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Yeah I was thinking of doing this myself. You're taking the words out of my mouth, or rather, the code out of my fingers, so to say. Thanks!

PR #2258, opened 2017-08-11
Remove WirelessPrinting Plugin

probonopd:
[WirelessPrinting](https://github.com/probonopd/WirelessPrinting) now uses a subset of the OctoPrint protocol because this is supported by Cura by default now


Ghostkeeper:
Okido.

PR #2271, opened 2017-08-13
PluginBrowser: Send also OS and architecture

thopiekar:
Orientated a bit on the way how browsers are sending their strings.
No idea, whether there is a standard on this or not.
Needs to be discussed I guess..

### DIFF
--- a/plugins/PluginBrowser/PluginBrowser.py
+++ b/plugins/PluginBrowser/PluginBrowser.py
@@ -15,6 +15,7 @@ from PyQt5.QtQml import QQmlComponent, QQmlContext
 import json
 import os
 import tempfile
+import platform
 
 i18n_catalog = i18nCatalog("cura")
 
@@ -43,7 +44,14 @@ class PluginBrowser(QObject, Extension):
 
         self._is_downloading = False
 
-        self._request_header = [b"User-Agent", str.encode("%s - %s" % (Application.getInstance().getApplicationName(), Application.getInstance().getVersion()))]
+        self._request_header = [b"User-Agent",
+                                str.encode("%s\%s (%s %s)" % (Application.getInstance().getApplicationName(),
+                                                              Application.getInstance().getVersion(),
+                                                              platform.system(),
+                                                              platform.machine(),
+                                                             )
+                                          )
+                               ]
 
         # Installed plugins are really installed after reboot. In order to prevent the user from downloading the
         # same file over and over again, we keep track of the upgraded plugins.


Ghostkeeper:
Devs, see CURA-4246.

thopiekar:
So easily merged? :scream: 
I thought you would be more interested in sending custom headers for each info. This would make it more easy to parse the info across different Cura versions.
However, no problem that I made the first step with this PR :stuck_out_tongue_winking_eye:

alekseisasin:
It is first step, depends on it what will you do with this data. I wanted
extend Header data with "platform.version()" to know which windows version
is used. Also possible to add extra private data.

On 28 August 2017 at 18:11, Thomas Karl Pietrowski <notifications@github.com

fieldOfView:
In any case, a common user agent would be nice. Uranium uses one, Cura now uses another (and OctoPrintPlugin uses yet another one).

Ghostkeeper:
Should they all be using Cura's user agent?

thopiekar:
Well, I wouldn't do that. I expect that parsing the user agent string will be needed at the end. So yeah, on Cura's side it is nice to put everything together, but on the server side I can imagine more work needs to be done.

However, I guess this needs to be discussed with people who are working on the web applications at Ultimaker.

alekseisasin:
I don't think parsing will be an issue. Just choose delimiter like " ; "
and extend Header data.

plugins/PluginBrowser/PluginBrowser.py
```diff
@@ -43,7 +44,14 @@ def __init__(self, parent = None):
 
         self._is_downloading = False
 
-        self._request_header = [b"User-Agent", str.encode("%s - %s" % (Application.getInstance().getApplicationName(), Application.getInstance().getVersion()))]
+        self._request_header = [b"User-Agent",
+                                str.encode("%s\%s (%s %s)" % (Application.getInstance().getApplicationName(),
```
LipuFei: This should be `<product>/<product-version>`. The splash here is not right... I will fix it.

LipuFei: Fixed in 0358b1fca

LipuFei:
I think if we need extra data, that should be sent through the request body, not the HTTP headers. And the plugin server will be also a real (RESTful) service which processes requests based on the those data.

plugins/PluginBrowser/PluginBrowser.py
thopiekar: Oh, thanks! 👍

PR #2272, opened 2017-08-13
Added speed_print_layer_0_only_for_walls setting.

smartavionics:
This controls whether the speed_print_layer_0 is used for all features on layer 0 or just
the walls.

The rational behind this setting is that sometimes the layer 0 walls have to be printed very slowly to ensure that they adhere to the build plate well but the skin doesn't actually need to be printed as slowly. I think the skin can be printed faster because (a) the skin is made up of straight lines and not curves and (b) where the skin segments start, they are touching a wall and have something to anchor too as well as the build plate. If there is a lot of skin then it can take a long time to print at the speed required for good adhesion of the walls so I think being able to only slow down the walls is a good feature.
Engine PR is https://github.com/Ultimaker/CuraEngine/pull/583.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2250,6 +2250,14 @@
                             "maximum_value_warning": "300",
                             "settable_per_mesh": true
                         },
+                        "speed_print_layer_0_only_for_walls":
+                        {
+                            "label": "Only Use For Walls",
+                            "description": "Only use the Initial Layer Print Speed for the walls and not for the other features (e.g. skin).",
+                            "type": "bool",
+                            "default_value": false,
+                            "settable_per_mesh": true
+                        },
                         "speed_travel_layer_0":
                         {
                             "label": "Initial Layer Travel Speed",


Ghostkeeper:
Devs, see issue CURA-4247.

Ghostkeeper:
It might be more in line with the rest of Cura's setting to make this into a specialised first layer speed for walls.

smartavionics:
I think the best solution is to have separate speeds for walls and skin on the first layer. Slowing down the walls makes a huge difference with some filaments in terms of how well they "grab" the surface but once the walls are in place, the skin can go down quite a lot quicker. This can make quite a difference if the walls are being printed at, say, 20mm/S.

Ghostkeeper:
Yeah I agree, but usually we do that with a parent setting `speed_print_layer_0` and then a few child settings called `speed_walls_layer_0`, `speed_skin_layer_0`, etc. that inherit from `speed_print_layer_0` by default.

BagelOrb:
I think the boolean setting which controls how the setting Initial Layer Print Speed works is a bit weird.

I think an Initial Layer Wall Speed setting would be better.
This setting may be below Initial Layer Print Speed.

We don't need a parent setting for Initial Layer Wall Speed and Initial Layer non-wall Speed, just like Initial Layer Print Speed is besides Print Speed, rather than them having a parent Print Speed with two children Initial Layer Print Speed and Other Layer Print speed.

smartavionics:
I think I'll close this (and the matching engine PR) and put together a new PR that provides an Initial Layer Wall Speed setting.

PR #2274, opened 2017-08-14
Move printing time

Ghostkeeper:
The printing time is moved to a new location, in the sidebar rather than the print job statistics bit.


fieldOfView:
At some point I am going to have to stop caring about this and get on with my life.

Ghostkeeper:
Your feedback is still valued, @fieldOfView. My own concern with this was that the text "Print time" in between seems to belong to the material estimate (same font, with the Gestalt laws and such). I'd rather have removed that bit of text. But it was my word against the designer's.

PR #2279, opened 2017-08-15
Updated brazilian portuguese translations

Patola:


resources/i18n/ptbr/cura.po
```diff
@@ -805,7 +805,7 @@ msgstr "Carregando interface..."
 #, python-format
 msgctxt "@info 'width', 'depth' and 'height' are variable names that must NOT be translated; just translate the format of ##x##x## mm."
 msgid "%(width).1f x %(depth).1f x %(height).1f mm"
-msgstr ""
+msgstr "%(largura).1f x %(profundidade).1f x %(altura).1f mm"
```
Ghostkeeper: Oops, this shouldn't get translated. It says so in the context!

Patola: Sorry! Thanks for catching and correcting this.

PR #2295, opened 2017-08-18
Make it pass desktop-file-validate

probonopd:
```
cura.desktop: error: (will be fatal in the future): value "cura-icon.png" for key "Icon" in group "Desktop Entry" is an icon name with an extension, but there should be no extension as described in the Icon Theme Specification if the value is not an absolute path
```
__Solution:__ Remove `.png`. __PLEASE NOTE__ that for this to work, cura-icon.png must be located in one of the directories in which desktops search for icons, such as `/usr/share/icons/hicolor/128x128/apps/`. Please make sure it ends up there (also in the AppImage, too.)

```
cura.desktop: error: value "application/sla;application/vnd.ms-3mfdocument;application/prs.wavefront-obj;image/bmp;image/gif;image/jpeg;image/png" for string list key "MimeType" in group "Desktop Entry" does not have a semicolon (';') as trailing character
```
__Solution:__ Add `;`

Reference:
https://travis-ci.org/AppImage/AppImageHub/builds/266072586#L538-L539


awhiemstra:
Uh, are you running desktop-file-validate on the .in file? Because that is the only way "@CMAKE_INSTALL_FULL_DATADIR@/cura/resources/images/cura-icon.png would be a relative path, as @CMAKE_INSTALL_FULL_DATADIR@ should be replaced with "/usr/share/" by CMake, making it an absolute path matching the spec. Note that if you want to replace it by an icon-theme icon name, you should also make sure the icon is installed properly into the icon theme and not just change the icon name.

The other thing seems fine, though I find parser that require stuff like closing ; on a list like this a bit silly.

probonopd:
No, I was running it on the desktop file that is contained within the AppImage.

Ghostkeeper:
Validates correctly now!

And the icon still works.

PR #2308, opened 2017-08-23
Add relative_extrusion setting to special modes category.

smartavionics:
UI addition that goes with https://github.com/Ultimaker/CuraEngine/pull/579.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4923,7 +4923,7 @@
                 "relative_extrusion":
                 {
                     "label": "Relative Extrusion",
-                    "description": "Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps.",
+                    "description": "Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps. Irrespective of this setting, the extrusion mode will always be set to absolute before the start and end Gcode are output.",
                     "type": "bool",
                     "default_value": false,
                     "value": "machine_gcode_flavor==\"RepRap (RepRap)\"",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4925,7 +4925,8 @@
                     "label": "Use Relative Extrusion",
                     "description": "Use relative extrusion rather than absolute extrusion. Before, Cura always used absolute extrusion. Most printers can cope with either relative or absolute extrusion.",
                     "type": "bool",
-                    "default_value": "machine_gcode_flavor==\"RepRap (RepRap)\"",
+                    "default_value": false,
+                    "value": "machine_gcode_flavor==\"RepRap (RepRap)\"",
                     "enabled": true,
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4922,8 +4922,8 @@
                 },
                 "relative_extrusion":
                 {
-                    "label": "Use Relative Extrusion",
-                    "description": "Use relative extrusion rather than absolute extrusion. Before, Cura always used absolute extrusion. Most printers can cope with either relative or absolute extrusion.",
+                    "label": "Relative Extrusion",
+                    "description": "Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps.",
                     "type": "bool",
                     "default_value": false,
                     "value": "machine_gcode_flavor==\"RepRap (RepRap)\"",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4923,7 +4923,7 @@
                 "relative_extrusion":
                 {
                     "label": "Relative Extrusion",
-                    "description": "Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps. Irrespective of this setting, the extrusion mode will always be set to absolute before the start and end Gcode are output.",
+                    "description": "Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps. Irrespective of this setting, the extrusion mode will always be set to absolute before any Gcode script is output.",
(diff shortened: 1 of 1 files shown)


smartavionics:
Actually, I've screwed up again because even though the checkbox is checked by default, the setting value is false and I have to explicitly unset it and set it again manually for the setting value to be true. I just don't understand how the whole value, default_value mechanism works. Sorry. Perhaps someone will take pity on me and describe how it behaves.

resources/definitions/fdmprinter.def.json
```diff
@@ -4919,6 +4919,17 @@
                     "enabled": "magic_spiralize",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
+                },
+                "relative_extrusion":
+                {
+                    "label": "Use Relative Extrusion",
+                    "description": "Use relative extrusion rather than absolute extrusion. Before, Cura always used absolute extrusion. Most printers can cope with either relative or absolute extrusion.",
```
BagelOrb: The description of the setting shouldn't contain info on what old behavior Cura previously had.

I suggest: `Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it could lead to small differences in the amount of material being extruded for each line.`

resources/definitions/fdmprinter.def.json
```diff
@@ -4919,6 +4919,17 @@
                     "enabled": "magic_spiralize",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
+                },
+                "relative_extrusion":
+                {
+                    "label": "Use Relative Extrusion",
```
BagelOrb: I suggest to remove the word 'Use' here.

BagelOrb:
> Actually, I've screwed up again because even though the checkbox is checked by default, the setting value is false and I have to explicitly unset it and set it again manually for the setting value to be true. I just don't understand how the whole value, default_value mechanism works. Sorry. Perhaps someone will take pity on me and describe how it behaves.

I think you've found an interesting bug there!

The way it's supposed to work: the default_value is ignored by the frontend if theres a value function. If there is no value function specified the default_value is used.
The engine uses the default_value only, because the engine cannot evaluate the python formula.

BagelOrb:
I cannot reproduce your bug. Maybe the frontend was still using an outdated build the first time you sliced....

resources/definitions/fdmprinter.def.json
smartavionics: Small differences to what? Do you mean small differences compared to the amount extruded using absolute extrusion?

fieldOfView: Rounding errors in each relative extrusion amount add up. So at the end of the print, you are less sure of the actual amount of material extruded.

smartavionics: OK, I can understand that although I suspect the amount is tiny and would be completely hidden by the errors introduced by filament diameter variation and other mechanical effects. I just think that @BagelOrb's suggested blurb above is a little negative.

smartavionics:
Tim, there are two commits on this branch, the second works for me, the first has the problem I mentioned above. Perhaps you used the second version?

resources/definitions/fdmprinter.def.json
BagelOrb: Ok maybe instead of saying `small differences` it should say `tiny deviations`.

Just to be clear: the deviation is as small as half a step of the stepper motor of the extruder. Depending on the printer that might be something which has a visible effect on the print quality.

Maybe you are right and we can leave that part out.
I don't want to be too negative, but I'm trying to list the advantages and disadvantages of this setting. If it only has advantages it should always be turned on of course!

You can tweak the exact wording a bit to your liking. It's your PR of course ;)

BagelOrb:
Ah yes.
The problem with your initial commit is that you confused the default_value and the value function.

The `default_value` is directly a json value (e.g. `true`), while the `value` is a python function (e.g. `"True"`).
The `default_value` is always required, but if you want a function you should *also* introduce a `value` to have that function as a python formula.

resources/definitions/fdmprinter.def.json
smartavionics: How about:

Use relative extrusion rather than absolute extrusion. Using relative E-steps makes for easier post-processing of the Gcode. However, it's not supported by all printers and it may produce very slight differences in the printed object compared to using absolute E-steps.

smartavionics:
>The default_value is directly a json value (e.g. true), while the value is a python function (e.g. "True").
The default_value is always required, but if you want a function you should also introduce a value to have that function as a python formula.

Thanks, I think I finally understand, LOL!

resources/definitions/fdmprinter.def.json
BagelOrb: I think we're focusing too much on the detail.

You should just update your PR at some point with what you think is good.

As long as you don't do that I'll keep the discussion alive ;)
How about `However, it's not supported by all printers and it may produce very slight deviations in the amount of deposited material compared to absolute E-steps.`
Saying it has effect on the printed object is not specific enough in my opinion..

alekseisasin:
By default you set the setting is enabled. It may be a problem. For Ultimkaer printers we can disable it setting and make it invisible, but we have (support) many other printers in Cura and I do not know do they use RepRap flavour or not. After updating the Cura version they will may get strange result, because by default we add this M83. And I think it will be a problem.

smartavionics:
The RepRap people want to use relative extrusion mode. As Tim says, please read the discussions that have already take place.

BagelOrb:
@alekseisasin what are you talking about? It clearly says: `"default_value": false,` It's only enabled by default for RepRap printers, which aren't really out there because the RepRap flavor is new since 2.6 I think.

smartavionics:
And it also says `"value": "machine_gcode_flavor==\"RepRap (RepRap)\"",`

alekseisasin:
Tim was right I checked wrong value, the default_value is false, it is right

CCS86:
My "machine" in Cura is a UMO, which seems to disable the ability to access the relative extrusion mode setting.

Do I have any option to gain that control without recreating my machine as a custom type?

Ghostkeeper:
It's disabled by this commit: https://github.com/Ultimaker/Cura/commit/f50255f1c95eda418310421a99bab6b96509ffd8

However that commit doesn't document why it would need to be disabled. It pretends as if it's self-evident! It does refer to an internal issue though, were it mentions why this was disabled:

LipuFei:

> Tested with UM3, it didn't print correctly. My settings are as the following:
> * "Relative Extrusion" on
> * "Retract Before Outer Wall" on
> * "Support" on
> * Tested with Bridge_Test.stl

Asasin:

> For testing: Enable setting "Relative Extrusion" and run some program on Ultimaker2 and Ultimaker Original. If the will have wrong printing behaviour , then the setting should be invisible (adjust printer definition json file).
> 
> Do we want to add this change for Cura 3.0 ?

LipuFei:

> I have merged the PRs and disabled this setting for all UMs because it doesn't work. For testing, we should make sure that the "Relative Extrusion" setting is NOT available and printing with UM machines should still work as before.

So it sounds like they found that it didn't work for UM3 at least. Whether it was also tested on UMO and UM2 is inconclusive. If it went wrong on the UM3 it might also go wrong on the UMO and UM2 since there are very little changes in the Marlin parser or motion planner between those. However maybe some pre-parsing in the UM3 could also break this (like what it does for automatic bed levelling).

PR #2310, opened 2017-08-23
CURA-4186 Add stack context for function evaluation

LipuFei:
CURA-4186

### DIFF
--- a/cura/Settings/ExtruderStack.py
+++ b/cura/Settings/ExtruderStack.py
@@ -7,7 +7,7 @@ from UM.Decorators import override
 from UM.MimeTypeDatabase import MimeType, MimeTypeDatabase
 from UM.Settings.ContainerStack import ContainerStack
 from UM.Settings.ContainerRegistry import ContainerRegistry
-from UM.Settings.Interfaces import ContainerInterface
+from UM.Settings.Interfaces import ContainerInterface, PropertyEvaluationContext
 
 from . import Exceptions
 from .CuraContainerStack import CuraContainerStack
@@ -57,21 +57,30 @@ class ExtruderStack(CuraContainerStack):
     #   \throws Exceptions.NoGlobalStackError Raised when trying to get a property from an extruder without
     #                                         having a next stack set.
     @override(ContainerStack)
-    def getProperty(self, key: str, property_name: str) -> Any:
+    def getProperty(self, key: str, property_name: str, context: Optional[PropertyEvaluationContext] = None) -> Any:
         if not self._next_stack:
             raise Exceptions.NoGlobalStackError("Extruder {id} is missing the next stack!".format(id = self.id))
 
+        if context is None:
+            context = PropertyEvaluationContext()
+        context.pushContainer(self)
+
         if not super().getProperty(key, "settable_per_extruder"):
-            return self.getNextStack().getProperty(key, property_name)
+            result = self.getNextStack().getProperty(key, property_name, context)
+            context.popContainer()
+            return result
 
         limit_to_extruder = super().getProperty(key, "limit_to_extruder")
         if (limit_to_extruder is not None and limit_to_extruder != "-1") and self.getMetaDataEntry("position") != str(limit_to_extruder):
             if str(limit_to_extruder) in self.getNextStack().extruders:
-                result = self.getNextStack().extruders[str(limit_to_extruder)].getProperty(key, property_name)
+                result = self.getNextStack().extruders[str(limit_to_extruder)].getProperty(key, property_name, context)
                 if result is not None:
+                    context.popContainer()
                     return result
 
-        return super().getProperty(key, property_name)
+        result = super().getProperty(key, property_name, context)
+        context.popContainer()
+        return result
 
     @override(CuraContainerStack)
     def _getMachineDefinition(self) -> ContainerInterface:
--- a/cura/Settings/ExtruderStack.py
+++ b/cura/Settings/ExtruderStack.py
@@ -65,12 +65,14 @@ class ExtruderStack(CuraContainerStack):
             context = PropertyEvaluationContext()
         context.pushContainer(self)
 
-        if not super().getProperty(key, "settable_per_extruder"):
+        if not super().getProperty(key, "settable_per_extruder", context):
             result = self.getNextStack().getProperty(key, property_name, context)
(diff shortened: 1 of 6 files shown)


cura/Settings/ExtruderStack.py
```diff
@@ -65,12 +65,14 @@ def getProperty(self, key: str, property_name: str, context: Optional[PropertyEv
...
-        if not super().getProperty(key, "settable_per_extruder"):
+        if not super().getProperty(key, "settable_per_extruder", context):
             result = self.getNextStack().getProperty(key, property_name, context)
             context.popContainer()
             return result
 
-        limit_to_extruder = super().getProperty(key, "limit_to_extruder")
+        limit_to_extruder = super().getProperty(key, "limit_to_extruder", context)
+        if limit_to_extruder is not None:
+            limit_to_extruder = str(limit_to_extruder)
```
Ghostkeeper: I don't see why this is necessary?

cura/Settings/GlobalStack.py
```diff
@@ -109,9 +109,11 @@ def getProperty(self, key: str, property_name: str, context: Optional[PropertyEv
         context.pushContainer(self)
 
         # Handle the "limit_to_extruder" property.
-        limit_to_extruder = super().getProperty(key, "limit_to_extruder")
+        limit_to_extruder = super().getProperty(key, "limit_to_extruder", context)
+        if limit_to_extruder is not None:
+            limit_to_extruder = str(limit_to_extruder)
```
Ghostkeeper: Why is this necessary?

cura/Settings/PerObjectContainerStack.py
```diff
@@ -0,0 +1,49 @@
+from typing import Any, Optional
+
+from UM.Decorators import override
+from UM.Logger import Logger
+from UM.Settings.Interfaces import PropertyEvaluationContext
+from UM.Settings.ContainerStack import ContainerStack
+
+
+class PerObjectContainerStack(ContainerStack):
```
Ghostkeeper: This class should be located inside the per-object settings plug-in.

cura/Settings/PerObjectContainerStack.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @override(ContainerStack)
+    def getProperty(self, key: str, property_name: str, context: Optional[PropertyEvaluationContext] = None) -> Any:
+        if context is None:
+            context = PropertyEvaluationContext()
+        context.pushContainer(self)
+
+        # this import has to be here otherwise there will be a circular import loop
+        from cura.CuraApplication import CuraApplication
+        global_stack = CuraApplication.getInstance().getGlobalContainerStack()
```
Ghostkeeper: Try `Application.getInstance().getGlobalContainerStack()`. You won't have a circular import then.

cura/Settings/PerObjectContainerStack.py
```diff
@@ -0,0 +1,49 @@
...
+        # Handle the "limit_to_extruder" property.
+        limit_to_extruder = super().getProperty(key, "limit_to_extruder", context)
+        if limit_to_extruder is not None:
+            limit_to_extruder = str(limit_to_extruder)
+
+        # if this stack has the limit_to_extruder "not overriden", use the original limit_to_extruder as the current
+        # limit_to_extruder, so the values retrieved will be from the perspective of the original limit_to_extruder
+        # stack.
+        if limit_to_extruder == "-1":
+            if 'original_limit_to_extruder' in context.context:
```
Ghostkeeper: Use double quotes for strings.

cura/Settings/PerObjectContainerStack.py
```diff
@@ -0,0 +1,49 @@
...
+                result = global_stack.extruders[str(limit_to_extruder)].getProperty(key, property_name, context)
+                if result is not None:
+                    context.popContainer()
+                    return result
+            else:
+                Logger.log("e", "Setting {setting} has limit_to_extruder but is not settable per extruder!", setting = key)
+
+        result = super().getProperty(key, property_name, context)
+        context.popContainer()
+        return result
```
Ghostkeeper: Shouldn't this limit to extruder handling just always be applied, rather than only in per-object stacks? I'd have put it in the main ContainerStack class. Or better yet, `InstanceContainer` even.

cura/Settings/PerObjectContainerStack.py
```diff
@@ -17,6 +18,13 @@ def getProperty(self, key: str, property_name: str, context: Optional[PropertyEv
         from cura.CuraApplication import CuraApplication
         global_stack = CuraApplication.getInstance().getGlobalContainerStack()
 
+        # Return the user defined value if present, otherwise, evaluate the value according to the default routine.
+        if self.getContainer(0).hasProperty(key, property_name):
+            if self.getContainer(0)._instances[key].state == InstanceState.User:
+                result = super().getProperty(key, property_name, context)
+                context.popContainer()
+                return result
```
Ghostkeeper: User defined values are never formulas, so I don't see why this is necessary or why this is specific to per-object settings.

cura/Settings/PerObjectContainerStack.py
```diff
@@ -51,3 +51,16 @@ def getProperty(self, key: str, property_name: str, context: Optional[PropertyEv
...
+        super().setNextStack(stack)
+
+        # trigger signal to re-evaluate all default settings
+        for key, instance in self.getContainer(0)._instances.items():
+            # only evaluate default settings
+            if instance.state != InstanceState.Default:
+                continue
+
+            self._collectPropertyChanges(key, "value")
+        self._emitCollectedPropertyChanges()
```
Ghostkeeper: Settings on the stack should not get updated. They are fixed from the moment that they are added, and should not get re-evaluated after that.

cura/Settings/PerObjectContainerStack.py
LipuFei: Fixed in 37414e7bc7c85ee85ce47289a3268f1214bda2b3

cura/Settings/PerObjectContainerStack.py
LipuFei: Fixed in b84500f742db7ffb9352065027024b0a3f913ec6

PR #2321, opened 2017-08-25
Add further options to z_seam_type.

smartavionics:
Can now specify that the z-seam should be located in either the most concave region of the
outline, the most convex region or the most curved (concave or convex) region.

See https://github.com/Ultimaker/CuraEngine/pull/590.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1262,14 +1262,14 @@
                 "z_seam_corner":
                 {
                     "label": "Seam Corner Preference",
-                    "description": "Control whether corners on the model outline influence the position of the seam.",
+                    "description": "Control whether corners on the model outline influence the position of the seam. None means that corners have no influence on the seam position. Hide Seam makes the seam more likely to occur on an inside corner. Expose Seam makes the seam more likely to occur on an outside corner. Hide or Expose Seam makes the seam more likely to occur at an inside or outside corner.",
                     "type": "enum",
                     "options":
                     {
-                        "z_seam_corner_none":  "Ignore Corners",
-                        "z_seam_corner_inner": "Inner Corners",
-                        "z_seam_corner_outer": "Outer Corners",
-                        "z_seam_corner_any":   "Any Corner"
+                        "z_seam_corner_none":  "None",
+                        "z_seam_corner_inner": "Hide Seam",
+                        "z_seam_corner_outer": "Expose Seam",
+                        "z_seam_corner_any":   "Hide or Expose Seam"
                     },
                     "default_value": "z_seam_corner_inner",
                     "enabled": "z_seam_type != 'random'",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1272,7 +1272,7 @@
                         "z_seam_corner_any":   "Any Corner"
                     },
                     "default_value": "z_seam_corner_inner",
-                    "enabled": "z_seam_type == 'back' or z_seam_type == 'sharpest_corner'",
+                    "enabled": "z_seam_type != 'random'",
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1228,10 +1228,7 @@
                     {
                         "back": "User Specified",
                         "shortest": "Shortest",
-                        "random": "Random",
-                        "most_curvature_convex":  "Most Convex Region",
-                        "most_curvature_concave": "Most Concave Region",
-                        "most_curvature":         "Most Curved Region"
+                        "random": "Random"
                     },
                     "default_value": "shortest",
                     "limit_to_extruder": "wall_0_extruder_nr",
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1258,6 +1258,23 @@
...
+                "z_seam_corner":
+                {
+                    "label": "Z Seam Corner Preference",
+                    "description": "Control whether corners on the model outline influence the position of the Z-seam.",
+                    "type": "enum",
+                    "options":
+                    {
+                        "z_seam_corner_none":  "Ignore Corners",
+                        "z_seam_corner_inner": "Inner Corners",
+                        "z_seam_corner_outer": "Outer corners",
```
BagelOrb: `corners` should be camel cased. The capital letter is missing.

resources/definitions/fdmprinter.def.json
```diff
@@ -1258,6 +1258,23 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "z_seam_corner":
+                {
+                    "label": "Z Seam Corner Preference",
```
BagelOrb: Please remove `Z` from `Z Seam`. The seam algorithm is used for both Z-seam and retraction-seams..
The previous settings `Z Seam X`, `Z Seam Alignment` etc are kind of a misnomer.

resources/definitions/fdmprinter.def.json
```diff
@@ -1258,6 +1258,23 @@
...
+                    "type": "enum",
+                    "options":
+                    {
+                        "z_seam_corner_none":  "Ignore Corners",
+                        "z_seam_corner_inner": "Inner Corners",
+                        "z_seam_corner_outer": "Outer corners",
+                        "z_seam_corner_any":   "Any Corner"
+                    },
+                    "default_value": "z_seam_corner_inner",
+                    "enabled": "z_seam_type == 'back'",
```
BagelOrb: Why is it not enabled on 'shortest'??

smartavionics:
I have tried enabling the corner options with z-seam type shortest and it has very little effect. That may be because the distance being travelled is small so the dist/50 is very small. In this situation perhaps a smaller divisor would work better. To be honest, I am not convinced that it is really useful to be able to specify a corner preference when the seam type is shortest.

BagelOrb:
I think it would be a good idea to set the divisor to 50 for User Specified seam location and to 10 for Shortest.

resources/definitions/fdmprinter.def.json
```diff
@@ -1228,7 +1228,8 @@
                     {
                         "back": "User Specified",
                         "shortest": "Shortest",
-                        "random": "Random"
+                        "random": "Random",
```
BagelOrb: I think it would be a good idea to also rename `Shortest` to `Shortest Path` even though it's a bit unrelated to the feature you are introducing...

smartavionics:
I'm not using the corner stuff when type is shortest. To me, shortest means shortest, not sort of shortest.

BagelOrb:
Moreover the setting is used when the seam type is Shortest, whether it's enabled for the user or not. It should be enabled so that the user can tweak its influence.

BagelOrb:
@Devs see CURA-4244

BagelOrb:
> I'm not using the corner stuff when type is shortest.

True. In PathOrderOptimizer::123 you explicitly exclude SHORTEST....

> To me, shortest means shortest, not sort of shortest.

The same argument holds for User Specified seam location.

It's not about what the user would expect for a given set of settings, it's about what would be the ideal behavior and how to communicate that in the clearest way.

I think the perfect print would try to find the shortest path, but would also try to hide the seam in order to maximize the quality. There's a balance here between quality and print time. If we can find a way to hide the seam with a minimal detour we should definitely do that.
If you want we could rename 'Shortest' to something else, but the shortest path always took corners into consideration and I think we should keep it that way.

smartavionics:
It now supports the corner options for the closest seam type. We need to evaluate its behaviour.

resources/definitions/fdmprinter.def.json
```diff
@@ -1258,6 +1259,23 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "z_seam_corner":
+                {
+                    "label": "Seam Corner Preference",
+                    "description": "Control whether corners on the model outline influence the position of the seam.",
```
BagelOrb: I'd like the description to contain some info on when a user should choose which option.
`Ignoring corners means the seam can end up on almost flat regions in the model. In order to hide the seam set the value to Inner Corners. In order to make the seam easily accessible set the value to Outer Corners. Choosing Any Corner means that an inner corner or an outer corner will be chosen - whichever is closest.`

resources/definitions/fdmprinter.def.json
```diff
@@ -1258,6 +1259,23 @@
...
+                    "type": "enum",
+                    "options":
+                    {
+                        "z_seam_corner_none":  "Ignore Corners",
+                        "z_seam_corner_inner": "Inner Corners",
+                        "z_seam_corner_outer": "Outer Corners",
+                        "z_seam_corner_any":   "Any Corner"
+                    },
+                    "default_value": "z_seam_corner_inner",
+                    "enabled": "z_seam_type != 'random'",
```
BagelOrb: What if `z_seam_type == 'sharpest_corner' && z_seam_corner == 'z_seam_corner_none'`?
What would happen then?

resources/definitions/fdmprinter.def.json
BagelOrb: Actually i just talked to a colleague and we think you should rename the options to:
```
"z_seam_corner_none":  "Anywhere",
"z_seam_corner_inner": "Hide Seam",
"z_seam_corner_outer": "Expose Seam",
"z_seam_corner_any":   "Hide or Expose Seam"
```

BagelOrb:
What if z_seam_type == 'sharpest_corner' && z_seam_corner == 'z_seam_corner_none'?
What would happen then?

smartavionics:
> What if z_seam_type == 'sharpest_corner' && z_seam_corner == 'z_seam_corner_none'?
What would happen then?

It just uses the first point in the polygon as the start point. That's not really a valid combination but there's no way in the json to stop it being selected?

BagelOrb:
Hmm okay. No we don't have such functionality in the json :S

We could also prevent that behaviour by then letting the engine default back to shortest or default to hidden corners. I'm not sure.

PR #2322, opened 2017-08-25
Dark colours, new icons and sidebar margins

Ghostkeeper:
image
A new theme.

Developers, see CURA-4148.


fieldOfView:
My biggest beef is with the extruder tabs:
image
Because of the huge contrast between the viewport and the sidebar, the lack of contrast in the sidebar makes it almost "unreadable" which tab is selected. When I look really hard to discern the lining, it looks like the unselected tab is the selected tab because it looks like it lies higher (because of the highlight).

The settings (and categories) no longer have a mouse-over effect. This mouse-over effect was also used to show the "tab-focus" of a field. So tabbing between fields is now a low less usable because with many settings (and the categories) you simply cannot see what setting/category has the current focus.

The layer slider has no center track anymore. This is a crucial design element, because the idea is that you can click and drag it around.
image
Talking about tracks themselves... They too are an important design element because they show where the element starts/ends. I am not just talking about the layer slider (which really needs to show its top), but also about the progress bar in the message area (white on extremely close to white):
image

What's the idea with the Cura logo? Was it too visible before?
image
Also looks great in the About window...

Most of this was handled much, much better by @rudowinger's design of a dark theme that I implemented for Cura 2.7. Why wasn't that taken as a starting point?

The current design feels like the design team consistently took everything that I worked hard for and threw it out, _without improving it_. And that, to be honest, feels like a kick in the gut.

With the topbar I could say "this is not an improvement, but I can see where it is going". With this design I can only say "this is a big step back and I don't understand why".

Ghostkeeper:
I can answer a few of those remarks.

> When I look really hard to discern the lining, it looks like the unselected tab is the selected tab because it looks like it lies higher (because of the highlight).

The extruder tabs are very bad right now because there's another ticket that changes their layout. This is how it's supposed to be looking eventually:
image

> The settings (and categories) no longer have a mouse-over effect.

They should have a mouse-over effect, but it's the text that changes colour rather than the background.

> The layer slider has no center track anymore.

Similarly, the layer view panel is also pending a redesign. The colours were already changed but the panel itself wasn't changed yet. This is how it's supposed to look eventually:
image

> I am not just talking about the layer slider (which really needs to show its top), but also about the progress bar in the message area (white on extremely close to white):

This was also a critique of mine and that was eventually changed but also put in a separate ticket. Here's the current plan of how it should look eventually:
image

> What's the idea with the Cura logo? Was it too visible before?

It is going to be moved to the top left inside the dark bar:
image

> Most of this was handled much, much better by @rudowinger's design of a dark theme that I implemented for Cura 2.7. Why wasn't that taken as a starting point?
> The current design feels like the design team consistently took everything that I worked hard for and threw it out, without improving it. And that, to be honest, feels like a kick in the gut.

This I have no answer to. My best guess is that the UX designers didn't know about the dark theme in Cura or didn't start from Cura but from the LCD screen of the Ultimaker printers (as one of the reasons they named for this redesign was to make the design consistent with those LCDs).

Ghostkeeper:
Some other critique points that we received about this design, to summarise and aggregate:
* The contrast in the side bar is too low.
* It is too similar to the design of Adobe Illustrator and to the colour scheme of Makerbot.
* Favours design over usability.
* Looks boring/dull.
* The black design appeals to designers and hackers, not to the layman or a professional.
* It's not the same as Ultimaker's house style at the website and the hardware, etc.
* The spacing in between the settings is way too big.
* The link to the material compatibility webpage doesn't look like a link (and takes up too much real estate as well).

fieldOfView:
It would have helped to know this is a work in progress

fieldOfView:
Because of the increased spacing between elements, the "simple mode" side bar can now be too long for smaller windows, and draws "under" the save button area:

image

Speaking of the save button area, there is no visual difference between the enabled and disabled state of the buttons. The buttons don't stand out (easy to miss completely).

fieldOfView:
The viewmode dropdown is not going to fit there on smaller screens, especially with the Cura logo moved into the topbar too.
image

mennoberg:
As there is no insight in screen resolutions from Cura, we got the websites resolutions from Ultimaker.com. This data shows that about 3,5% of all visitors have a screen width lower than 1280. Within the dev team we agreed on primarily focussing for 1280x1024 and up but it also has to work on 1280x720 (2,13%).

Ghostkeeper:
Which dev team was that? Or did I miss that meeting? It probably would've been good to tell the guy who is going to implement it...

fieldOfView:
@mennoberg many users of Cura use an old computer set next to/near the 3d printer, exclusively for slicing/printing. Using website metrics may not be relevant.

Having said that, a 1280x800 minimal resolution for the optimal experience is probably ok, though it should not be broken/inaccessible for 1024x768 screens. Suboptimal: yes, maybe. Broken: no.

PR #2329, opened 2017-08-27
Fix setting the Gantry height setting in Machine Settings

fieldOfView:
This PR fixes setting the Gantry height setting in Machine Settings. Before this PR, the setting would be correctly added to the definition_changes container, but the buildvolume would not be updated. That means that Cura needs to be restarted, or the active stack needs to be switched before the buildvolume updates.

Fixes an issue uncovered here:
http://elb-v2-prod-615036279.eu-west-1.elb.amazonaws.com/en/community/50595-cura-27-beta?page=4&sort=#reply-187977

Would be nice if this could still make it in to 2.7, but I understand if it doesn't. There are workarounds for the bug.


nallath:
This change makes total sense. I'm not against chery-picking it to 2.7. Let me ask our PO if this is okay.

nallath:
As it's such a simple fix, i also chery-picked this to 2.7

Issue #1141, opened 2016-11-13
Speed overrides in materials


MarcusWolschon:
Some materials have maximum recommed printing speeds (bonus points for doing the same with travel speeds because a too large gap between printing and travel speeds will obviously cause quality issues).
This should be a part of the material profile to override the values from the printing profile
just like a maximum machine speed overrides the speeds in a printing profile that the machine is not capable of reaching.

filipgoc:
+1. I'm curious about this holistically. 
(Whether everything in material should override anything in quality https://github.com/Ultimaker/Cura/issues/1135)

Ghostkeeper:
The XML materials are not as powerful in what they can express as the rest of the profiles in Cura. Currently it defines only 7 settings (seen here: https://github.com/Ultimaker/Cura/blob/master/plugins/XmlMaterialProfile/XmlMaterialProfile.py#L532).This is because it was designed as an industry-wide standard.

We have plans to extend the profiles with a Cura namespace so that we can define our own settings in there.

filipgoc:
happy to hear the list is to be expanded a bit. 
(At least to add cool_fan_speed_max and cool_fan_speed_min (instead of just speed) etc. basically add categories of settings, instead of specific settings.)
(And I do agree with @MarcusWolschon that for many flexibles limiting max speeds make sense...)

ChrisTerBeke:
Do we want this? It might mean that something lower in the stack suddenly gets priority over something that's higher up in the stack?

nallath:
Preferably not. The data model is already complicated enough as it is.

ChrisTerBeke:
Agreed.

fieldOfView:
> It might mean that something lower in the stack suddenly gets priority over something that's higher up in the stack?

How would it do that? Please see (and preferably test) #2333, which implements the namespaced arbitrary cura setting concept mentioned by @Ghostkeeper. That enables setting speeds from the material profile. Any setting made there is neatly overridden by settings in higher containers (quality, quality changes and user). All fairly well behaved.

ChrisTerBeke:
@fieldOfView this behaviour is in the request of @MarcusWolschon ("This should be a part of the material profile to override the values from the printing profile"), which indeed goes against the whole stack setup so we probably won't do this. Loading Cura settings from an XML namespace is fine, but having them overrule the quality settings is not.

fieldOfView:
Right. It was getting confusing who was reacting to what.

[...]

PR #2333, opened 2017-08-29
Cura-specific settings in xmlmaterials

fieldOfView:
This PR adds functionality to include any Cura setting in an xml fdm material file. These Cura-specific settings are namespaced to not conflict with the fdm material xml spec.

A followup to this PR could enable the user to add settings to the material profile on the materials pane of the preferences. This UI is not part of this PR.

A material with a namespaced cura setting would look like this (note the namespace declaration at the top, and the `<cura.setting>` tag:
```
<?xml version="1.0" encoding="UTF-8"?>
<fdmmaterial xmlns="http://www.ultimaker.com/material" xmlns:cura="http://www.ultimaker.com/cura" version="1.3">
    <metadata>
        <name>
            <brand>Test</brand>
            <material>PLA</material>
            <color>Test Color</color>
        </name>
        <GUID>30a578af-5661-4149-8fbf-c82d8153a697</GUID>
        <version>1</version>
        <color_code>#ffc924</color_code>
        <description></description>
        <adhesion_info></adhesion_info>
    </metadata>
    <properties>
        <density>1.24</density>
        <diameter>2.85</diameter>
    </properties>
    <settings>
        <setting key="print temperature">200</setting>
        <setting key="heated bed temperature">60</setting>
        <setting key="standby temperature">175</setting>

        <cura:setting key="infill_sparse_density">13</cura:setting>

        <machine>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 2+"/>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 2 Extended+"/>

            <hotend id="0.25 mm" />
            <hotend id="0.4 mm" />
            <hotend id="0.6 mm" />
            <hotend id="0.8 mm" />
        </machine>

        <machine>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 2"/>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 2 Go"/>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 2 Extended"/>
            <setting key="standby temperature">150</setting>
            <setting key="processing temperature graph">
                <point flow="2" temperature="180"/>
                <point flow="10" temperature="230"/>
            </setting>
        </machine>

        <machine>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker Original"/>
            <setting key="standby temperature">150</setting>
            <cura:setting key="infill_sparse_density">14</cura:setting>
        </machine>

        <machine>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 3"/>
            <machine_identifier manufacturer="Ultimaker B.V." product="Ultimaker 3 Extended"/>
            <setting key="print cooling">100</setting>
            <hotend id="BB 0.4" />
            <hotend id="BB 0.8" />
            <hotend id="AA 0.25">
                <setting key="hardware compatible">yes</setting>
            </hotend>
            <hotend id="AA 0.4">
                <setting key="hardware compatible">yes</setting>
            </hotend>
            <hotend id="AA 0.8">
                <setting key="hardware compatible">yes</setting>
            </hotend>
        </machine>

    </settings>
</fdmmaterial>
```


nallath:
CURA-4253

maukcc:
Please make it so that it can be added under the machine specific details also, and not only as a general setting

fieldOfView:
@maukcc, that should already work. Doesn't it?

maukcc:
I do not know, I have not tested it.
I was just looking at the example, and saw it being used in the general settings.

fieldOfView:
Out of curiosity, is CURA-4253 in this sprint or in a future sprint?

In other words: *bump*

nallath:
Yeah it's in there, but it's a bit of a mess as to if it's something that we should add. The whole material profiles are a bit of a mess.

fieldOfView:
> The whole material profiles are a bit of a mess.

And this could be a first step towards making it more consistent with the rest of the profiles system. Unless you want to drop the xml profiles all the way.

maukcc:
+1 for dropping them.

nallath:
Well, we're not going to drop them all together as far as I know. I will try to get some concensus regarding this. The current implementation of Cura is bloated. A lot of the slowness in Cura is caused by the current material system.

Ghostkeeper:
I don't see us dropping the XML format any time soon. There's lots of plans for the near future for the material files. It's not going to become simpler either...

I see this sort of change as the only real solution that would make everybody happy, but some people see this as pollution of the files so apparently it doesn't.

fieldOfView:
Does it make sense if I keep this PR updated? If it stands no chance of being merged anyway, I would rather just close the PR.

ChrisTerBeke:
There is still a discussion about whether this should be supported. XML spec clearly states that it's extendable thus this could be added, but not everyone agrees...

fieldOfView:
Ok, I'll fix the conflict then...

fieldOfView:
For reference: https://github.com/Ultimaker/Cura/issues/1141#issuecomment-260236781

Ghostkeeper:
It's about to get a conflict with https://github.com/Ultimaker/Cura/pull/2867 as well. You might want to leave it for a while.

I think it's best to just plan a meeting in person with a few firmware guys about this. I'm sure @daid or @MarcoTvM has a good opinion too.

ChrisTerBeke:
I just did a quick unit test and the firmware breaks when adding Cura specific settings to an XML file. This means that we can't add this functionality until the firmware supports it.

fieldOfView:
So the firmware has an XML parser that does not understand namespaces? That's strange...

nallath:
Woah. That's weird...

daid:
Surprising, the code was written to ignore anything unknown. What exception is it throwing? (Element-tree xml is a bit annoying to use with namespaces)

fieldOfView:
Fixed conflicts with master

fieldOfView:
Fixed conflicts with master

Appesteijn:
To be clear, we need changes in the firmware for this PR and that is not likely to happen in the upcoming 4-6 months.

fieldOfView:
@Appesteijn, understood. I've set a reminder for a birthday party on august 29. There will be cake.

Ghostkeeper:
@Appesteijn Robin and I tested this a month ago in preparation for the model assistant issue. The firmware accepts this since version 4.0 apparently. So it's long been in there.

ianpaschal:
No merge conflicts but since this branch is 557 commits behind master, it doesn't work with the current Uranium. I pulled `Ultimaker/Cura master` into this `fieldOfView/Cura feature_xmlmaterials_cura_settings` but I'm still getting some hard crashes relating (I think) to single extrusion mode.

This will take some time to sort out...

ianpaschal:
Ok. I've been testing this. Number values seem to work like gravy but boolean (`yes`/`no`) values don't seem to have any effect. I've tried a few to try and make sure that it's not that they're being overwritten somehow but no luck.

Have you gotten this to work, @fieldOfView?

fieldOfView:
I had not tested anything but numeric values, but adding a bool setting works for me. I added the following line to generic_pla, and it successfully turned off retraction for the materials.

```        <cura:setting key="retraction_enable">False</cura:setting>````

**Comment by fieldOfView** [comment]
Godmode plugin is a good way to inspect if the internal material profiles get created in the way you expect:
https://github.com/sedwards2009/cura-god-mode-plugin

**Comment by ianpaschal** [comment]
> I had not tested anything but numeric values, but adding a bool setting works for me. I added the following line to generic_pla, and it successfully turned off retraction for the materials.

<cura:setting key="retraction_enable">False</cura:setting>

Not here unfortunately. Also, regardless, this should implement the same same boolean syntax as existing attributes which is `yes`/`no`

**Comment by fieldOfView** [comment]
I have fixed reading and writing yes/no as boolean values. True/False still work for reading, but they are always written as yes/no.

If this still does not work, please check with GodMode plugin if the value gets overwritten by another profile higher up the stack.

**Comment by ianpaschal** [comment]
Yeah I was using GodMode. I've tested again it seems to work. I'm not sure why I was having troubles yesterday.

File: cura/CuraApplication.py [hunk 1]
```diff
@@ -1368,6 +1380,58 @@ def mergeSelected(self):
...
+                transformation.setTranslation(zero_translation)
+                transformed_mesh = mesh.getTransformed(transformation)
+
+                # Align the object around its zero position
+                # and also apply the offset to center it inside the group.
+                node.setPosition(-transformed_mesh.getZeroPosition() - offset)
+
+            # Use the previously found center of the group bounding box as the new location of the group
+            group_node.setPosition(group_node.getBoundingBox().center)
+
```
ianpaschal: These changes don't seem to be related to material loading, but maybe I'm missing something. Can you explain why they're in the PR?

ianpaschal:
I think there's several files in this branch which need to have master pulled in.

fieldOfView:
Something went wrong merging in master. Just a minute...

fieldOfView:
Fixed the merge/master issue.

ianpaschal:
Great! In this case, merging!

fieldOfView:
@Appestein, scrap that cake in august. The birthdayparty is off... I’m having a beer instead now.

Appesteijn:
Nooooo, @ianpaschal  what did you do!!!!

filipgoc:
Loving this! Is this gonna make it into Cura 3.3 ?!

fieldOfView:
It’s in. Note that there is no UI for it whatsoever, but you can edit values into the xml files now

filipgoc:
Nice! This is a major improvement that I've not even hoped for anymore!

Now I can finally create functional material profiles and not a strange conglomeration of materials and quality and hope it falls in place.

PR #2343, opened 2017-08-30
Install cura icon on Linux

thopiekar:
This commit is needed for https://github.com/probonopd/Cura/commit/b9970d57b48b31f217f08fb652e5e39f1e55f3bb
Never tested it, but believing the author of the commit that this change will work.


nallath:
CURA-4265

nallath:
I'm not that worried about this, so I'm merging it. If it breaks anything, it should be easy enough to revert it.

PR #2352, opened 2017-08-31
Move Logo and Top Bar Icons

Ghostkeeper:
This moves the Cura logo to the top bar and removes the icons of the buttons in the top bar. The icon for the printer status is not removed but moved to the right.

image

Implements issue CURA-4213.


nallath:
Apart from this looking pretty ugly, it's according to the new design.

fieldOfView:
As long as the topbar is under redevelopment, can I make a case for changing the word "prepare"?

"Prepare" is a verb, and a verb affords an action. Pressing it does not evoke an action, but switches modality. The "Prepare" button in the top bar conflicts with the same name button in the sidebar (when autoslicing is turned off). That button does rightly have a verb as a caption. Come to think of it, that button would be better off with "Slice" as a caption (because that's what it does).

Proposed alternatives for the topbar button: "Settings" or "Setup" (yes, "setup" can also be a verb, but it is a noun as well).

nallath:
We already have a new new design.

LipuFei:
new-1
new-2
new-3

PR #2353, opened 2017-08-31
CURA-4211 New extruder buttons

LipuFei:
image


LipuFei:
Fixed the merge conflicts and some other things. This is the new look:
new-1
new-2

resources/qml/ExtruderButton.qml
```diff
@@ -54,22 +49,27 @@ Button
...
         color: model.color
 
-        width: UM.Theme.getSize("default_margin").width
-        height: UM.Theme.getSize("default_margin").height
+        width: UM.Theme.getSize("extruder_button_material_size").width
+        height: UM.Theme.getSize("extruder_button_material_size").height
+        radius: width / 2
+
+        border.width: 0
+        border.color: "transparent"
```
nallath: Do we really need to set color if the border if it's already at 0 size?

LipuFei: That can be gone.

maukcc:
Maybe also color the entire circel in which the number is when the extruder is actually used?

fieldOfView:
First of all, thanks for also testing this with 8 extruders.
Nitpick: there's a double border between extruder buttons that looks ugly.

PR #2356, opened 2017-09-01
Added speed_walls_layer_0, acceleration_walls_layer_0 and jerk_walls_layer_0 settings.

smartavionics:
These allow the user to control the speed/acceleration/jerk for walls on layer 0 separately
from other features (e.g. skin).

See https://github.com/Ultimaker/CuraEngine/pull/596.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2270,6 +2270,19 @@
                             "maximum_value_warning": "300",
                             "settable_per_mesh": true
                         },
+                        "speed_walls_layer_0":
+                        {
+                            "label": "Initial Layer Wall Speed",
+                            "description": "The speed used to print walls in the initial layer. Reducing this speed may increase adhesion to the build plate while still allowing the skin areas to be printed at a faster speed.",
+                            "unit": "mm/s",
+                            "type": "float",
+                            "default_value": 30,
+                            "value": "speed_print_layer_0",
+                            "minimum_value": "0.1",
+                            "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
+                            "maximum_value_warning": "300",
+                            "settable_per_mesh": true
+                        },
                         "speed_travel_layer_0":
                         {
                             "label": "Initial Layer Travel Speed",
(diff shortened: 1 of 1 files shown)


nallath:
CURA-4266

smartavionics:
Getting some very nice results with PETG by setting initial layer  walls to 15mm/S and the skin to 40mm/S.

Ghostkeeper:
I've discussed this PR with @nallath. We've decided not to accept this one.

There's currently a major problem with these changes: When you change the Initial Layer Print Speed, not knowing about the rest of the settings, the initial layer wall speed doesn't change along, so you essentially only change the speed of the bed adhesion, skin and support on that layer but somehow the walls don't change with it.

To solve such cases we typically have a parent setting (Initial Layer Print Speed) with a bunch of child settings that cover all print features (Initial Layer Wall Speed, Initial Layer Skin Speed, Initial Layer Support Speed, etc.). However that solution will introduce a load of new settings that are relatively useless to tweak. That basically doubles all of our speed settings.

In the future we'd like to have a structure to change all settings for the initial layer (or just for some specific part of the print). Until then, we don't want to merge this.

Unless there is a very strong argument that is yet unmade, of course.

smartavionics:
Well, it's your baby, you're in control. it works very well for me and I will keep using it, just bear it in mind when others gripe about their problems getting their walls to stick!

smartavionics:
Hello again. I wanted to revisit this topic because I have been using this code since the PR was created and it is very effective. On the first layer, I print the walls dead slow and the skin much faster and get perfect results. You mention above future work to facilitate the changing of all settings for the first layer, have you made any progress on this?

Ghostkeeper:
We haven't made any progress on that, no. Selecting separate settings per layer or for select features of the print is really complex so it's hard to plan that in.

fpapleux:
I really wish we could print walls at a different speed than the surface on first layer. This is really a drag.

smartavionics:
> I really wish we could print walls at a different speed than the surface on first layer. This is really a drag.

My Cura builds provide that option, they can be found at https://www.dropbox.com/sh/s43vqzmi4d2bqe2/AAADdYdSu9iwcKa0Knqgurm4a?dl=0. Please read the README.md file there for more details.

PR #2369, opened 2017-09-02
CuraApplication: Minor pep8 fix

vivekanand1101:


Ghostkeeper:
This shouldn't have been merged. PEP8 indeed specifies that named parameters should not have spaces around their `=` symbol, but [our own code style specification](https://github.com/Ultimaker/Meta/blob/master/code_conventions.md#spacing) overrules that. All binary operators should have spaces except `.`, `->`, `->*`, `.*`, `,` and `::`.

PR #2375, opened 2017-09-03
Optimise sending slice job

Ghostkeeper:
Some minor performance optimisations, and refactoring to pull some of the common code out to a separate function.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -250,9 +250,7 @@ class StartSliceJob(Job):
         settings = self._buildReplacementTokens(stack)
 
         #Also send the material GUID. This is a setting in fdmprinter, but we have no interface for it.
-        material_instance_container = stack.findContainer({"type": "material"})
-        if material_instance_container:
-            settings["material_guid"] = material_instance_container.getMetaDataEntry("GUID", "")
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -222,32 +247,24 @@ def _buildExtruderMessage(self, stack):
...
 
-        material_instance_container = stack.findContainer({"type": "material"})
+        settings = self._buildReplacementTokens(stack)
 
-        settings = {}
-        for key in stack.getAllKeys():
-            settings[key] = stack.getProperty(key, "value")
-            Job.yieldThread()
+        #Also send the material GUID. This is a setting in fdmprinter, but we have no interface for it.
+        material_instance_container = stack.findContainer({"type": "material"})
```
LipuFei: I think this can just be `stack.material`, which will be faster, and check if it is `empty_material`.

Ghostkeeper: Indeed, it can. That removes the need for some error handling. I've made that change here: https://github.com/Ultimaker/Cura/pull/2375/commits/e55bb1e25c94b0d2e78a53372116fa60cda99971

Ghostkeeper:
Developers, see issue CURA-4567 (nice number!).

Issue #1632, opened 2017-04-01
Memory leak in Cura 2.5 beta


mqudsi:
There seems to be a memory leak in one of the Cura menu calls, all I did was click on the `Settings` main menu and the application froze as the memory usage went all the way up to 8.xx GiB, before completing some odd calculation and then slowly ramping back down to 6xx MiB. I don't believe it is slicing-related as I have auto-slicing disabled (Cura 2.5 beta).

mqudsi:
[cura - Copy.zip](https://github.com/Ultimaker/Cura/files/887872/cura.-.Copy.zip)

nallath:
Autch. 80 mb logs. 

I do see some issues with auto calculated values causing a lot of overhead, but that's about it.

Ghostkeeper:
The logs show a division by 0 error towards the end:

    2017-04-01 12:54:33,246 - DEBUG - UM.Logger.logException [74]: Exception: An exception occurred in inherit function =0 if infill_sparse_density == 100 else math.ceil(round(top_thickness / resolveOrValue('layer_height'), 4))
    2017-04-01 12:54:33,247 - DEBUG - UM.Logger.logException [78]: Traceback (most recent call last):
    2017-04-01 12:54:33,248 - DEBUG - UM.Logger.logException [78]:   File "D:/2.5/build/inst/lib/python3.5/site-packages/UM\Settings\SettingFunction.py", line 71, in __call__
    2017-04-01 12:54:33,249 - DEBUG - UM.Logger.logException [78]:   File "<UM.Settings.SettingFunction (0x1f983ca64a8) =0 if infill_sparse_density == 100 else math.ceil(round(top_thickness / resolveOrValue('layer_height'), 4)) >", line 1, in <module>
    2017-04-01 12:54:33,250 - DEBUG - UM.Logger.logException [78]: ZeroDivisionError: float division by zero

Apparently layer height was set to 0.

For the rest I don't see much that could indicate a memory leak. Not that we put memory usage in the logs.

Could it be the log file itself?

mqudsi:
The division by zero is because of the other bug I filed where Cura loses focus in the middle of a change to an edit box (so 0.06^H8 becomes 0 then I need to re-select the edit box and add the 8); it's not related to the bug I think.

[...]

PR #2380, opened 2017-09-04
Prevent a possible memory leak while showing a stuck camera stream

fieldOfView:
This PR prevents a possible memory leak while showing a stuck camera stream by flushing the stream buffer and restarting the stream if the stream buffer gets unrealistically large.

Note: I don't have an UM3 to test this with. The 2000000 is an arbitrary number. I don't think any 800x600 jpg frame will realistically be larger than that.

Note 2: Interestingly I have not received any similar memory leak reports from OctoPrintPlugin users, though it uses the same code as UM3NetworkPrinting. It could be the specific version of mjpgstreamer in the UM3 that is acting up.

### DIFF
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
@@ -326,6 +326,9 @@ class NetworkPrinterOutputDevice(PrinterOutputDevice):
         return True
 
     def _stopCamera(self):
+        self._stream_buffer = b""
+        self._stream_buffer_start_index = -1
+
         if self._camera_timer.isActive():
             self._camera_timer.stop()
             if self._image_reply:
@@ -1164,6 +1167,12 @@ class NetworkPrinterOutputDevice(PrinterOutputDevice):
         if self._image_reply is None:
             return
         self._stream_buffer += self._image_reply.readAll()
+
+        if len(self._stream_buffer) > 2000000: # No single camera frame should be 2 Mb or larger
+            Logger.log("w", "MJPEG buffer exceeds reasonable size. Restarting stream...")
+            self._stopCamera() # resets stream buffer and start index
+            self._startCamera()
+            return
 
         if self._stream_buffer_start_index == -1:
             self._stream_buffer_start_index = self._stream_buffer.indexOf(b'\xff\xd8')


nallath:
CURA-4281

fieldOfView:
I referenced this pull in a number of issues, so these can be checked (and hopefully closed) when this is merged.

nallath:
It doesn't break anything, but i still notice a steady increase in memory :( I can't quite explain why this is an issue for the UM3 code and not for the octoprint.

fieldOfView:
Then I think it is not the mjpg stream afterall. This comment may have an important hint:
https://github.com/Ultimaker/Cura/issues/1632#issuecomment-295799962

nallath:
Ha! Found it. The culprit was indentation! 08766ee66f061ab35766b30199375ac0d966580f

fieldOfView:
Good find. I still say it can't hurt to clear the buffer and start_index when stopping the camera.

MFHFozzy:
I was running 2.7Beta, and when i installed 2.7 release, its crashing on startup. So i cant update this yet.

nallath:
Agreed. I'll merge this.

MFHFozzy:
Updated to 3.0. 
The good news is that Cura did not run out of memory...so far.
Ran a print overnight, with Cura minimized. Was still running in the morning. But, the camera was stuck (early on in the print, probably when i minimized it).
Running another print over night now, leaving cura on the camera page and not minimized.

MFHFozzy:
Same results. Camera frozen, but memory leak gone.

MFHFozzy:
Just did another print, but never had my monitors go into sleep mode.
Window never minimized either.
Camera worked fine all print, and memory held fast @ ~750K used.

SteveDC2:
I am getting memory leaks when connecting to Octoprint and displaying video.
Not sure if it is related to the above issue or not, but thought probably somewhat related so tagging on to this issue. 
If you would prefer me to start a new thread please just let me know.
Anyhow...
Cura 3.5.1 connecting to Octoprint 1.3.9. Windows 10
About 50% of the time I start a print with the "Print with OctoPrint" button after slicing the monitor window opens fine and starts displaying live video.
Monitoring memory usage with Task Manager shows Cura's memory utilization go up at about 15MB per second. Switching Cura to the "Prepare" view causes the increase to stop (but does not release the memory). Going back to the Monitor view then starts leaking memory again.
I have not been able to determine exactly what conditions cause the leak vs not leaking, but will try to see if I can find a pattern.

fieldOfView:
Yes, please start a new issue. Also please check if a new version of OctoPrint Connection is available in the Toolbox.

PR #2388, opened 2017-09-05
Added vertex delta k8800

velbn:
We've added the config for the K8800 machine

### DIFF
--- a/resources/definitions/vertex_delta_k8800.def.json
+++ b/resources/definitions/vertex_delta_k8800.def.json
@@ -1,0 +1,65 @@
+{
+    "id": "K8800",
+    "name": "Vertex Delta K8800",
+    "version": 1,
+    "inherits": "fdmprinter.json",
+    "machine_settings": {
+        "machine_width": {
+            "default": 200
+        },
+        "machine_height": {
+            "default": 225
+        },
+        "machine_depth": {
+            "default": 200
+        },
+        "machine_center_is_zero": {
+            "default": true
+        },
+        "machine_shape": {
+            "default": "elliptic"
+        },
+        "machine_nozzle_size": {
+            "default": 0.35
+        },
+        "machine_head_shape_min_x": {
+            "default": 0
+        },
+        "machine_head_shape_min_y": {
+            "default": 0
+        },
+        "machine_head_shape_max_x": {
+            "default": 0
+        },
+        "machine_head_shape_max_y": {
+            "default": 0
+        },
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, refer to issue CURA-4312.

alekseisasin:
velbn, Could you please make few changes in vertex_delta_k8800.def.json file:
-> The version should be 2;
-> "inherits": "fdmprinter", Not fdmprinter.json
-> you override settings, then define them within "overrides": {    ....    }
-> do you want to add some information for "metadata" : { }

velbn:
Hi I've updated the definition file

resources/definitions/vertex_delta_k8800.def.json
```diff
@@ -1,65 +1,62 @@
...
+      "default_value": 0
+    },
+    "machine_nozzle_offset_x_1": {
+      "default_value": 0
+    },
+    "machine_nozzle_offset_y_1": {
+      "default_value": 0
+    },
+    "machine_gcode_flavor": {
+      "default_value": "RepRap"
```
LipuFei: This value should be one of the following:
- `"RepRap (Marlin/Sprinter)"`
- `"RepRap (Volumetric)"`
- `"RepRap (RepRap)"`

resources/definitions/vertex_delta_k8800.def.json
```diff
@@ -1,65 +1,62 @@
...
-        "machine_end_gcode": {
-            "default": "; Vertex Delta end code\nM107 ; Turn off fan\nG91 ; Relative positioning\nT0\nG1 E-1 F1500; Reduce filament pressure\nM104 T0 S0\nG90 ; Absolute positioning\nG92 E0 ; Reset extruder position\nM300 S4000 P500\nM300 S3000 P500\nM300 S2000 P800\nG28\nM84 ; Turn steppers off"
-        }
+  "id": "K8800",
+  "name": "Vertex Delta K8800",
+  "version": 2,
+  "inherits": "fdmprinter",
+  "metadata": {
+    "manufacturer": "Velleman nv",
+	"file_formats": "text/x-gcode",
```
LipuFei: Please replace this `TAB` with `spaces`.

velbn:
Fixed the things =)

velbn:
One quick side question, when will this config be available? In the next release or when the user opens cura?

LipuFei:
Now that this is merged, it will be in the next release :)

PR #2398, opened 2017-09-07
Add setting to only iron the highest layer of a mesh

Ghostkeeper:
This can save some time if only the highest layer of your print needs to be smooth.

See also https://github.com/Ultimaker/CuraEngine/pull/603 for the implementation of this setting.


nallath:
Does this already have a Jira ticket?

LipuFei:
Ticket: CURA-4323

PR #2399, opened 2017-09-07
CURA-4212: Move LayerView to top right side

LipuFei:
image


nickthetait:
Can the LayerView box be minimized? Maybe this is outside of the scope of this work...

LipuFei:
@nickthetait It's not in the scope of this work... Sorry..

LipuFei:
layerview

PR #2416, opened 2017-09-11
Fix hover state of Monitor button in the topbar

fieldOfView:
This PR makes the Monitor button have the same behavior as the Prepare button, in regards to its hover state. Before this PR, the Monitor has no underline when hovered, while the Prepare button does.


ChrisTerBeke:
Good spot! Will review and merge tomorrow.

LipuFei:
This is so small I will merge it right away.

PR #2417, opened 2017-09-11
Draw grid lines using shader

fieldOfView:
This PR removes the grid line geometry from the buildplate, and instead draws grid lines using a glsl shader. This is much more efficient, especially for printers with large buildplates*. Nice anti-aliasing is also much cheaper to achieve using this shader than using line-geometry.

*: I am working on a printer that has a basically endless buildplate depth. The geometry-based approach that was recently introduced takes half a minute to just set up the geometry, whereas the proposed GLSL shader does it instantly.


LipuFei:
For Dev, see CURA-4317

nallath:
I'm getting a bit of a weird result in certain angles. 

weirdshaderresult
I think that overall, the change is definitely for the better.

fieldOfView:
Does it look/behave like z-fighting? I'll try to reproduce and fix this.

fieldOfView:
Should be better now

nallath:
Whoo, i'll test it again :)

fieldOfView:
That was quick!

Issue #2397, opened 2017-09-07
Cura 2.7.0 - Multiply group error


eloipr:
I'm running the Cura 2.7.0 AppImage on Ubuntu 16.04.
When I try to multiply a group I get the next error:
```
2017-09-07 15:40:04,279 - ERROR - UM.Logger.logException [76]: Exception: Job <cura.MultiplyObjectsJob.MultiplyObjectsJob object at 0x1891c8d8> caused an exception
2017-09-07 15:40:04,282 - ERROR - UM.Logger.logException [80]: Traceback (most recent call last):
2017-09-07 15:40:04,285 - ERROR - UM.Logger.logException [80]:   File "/build/2.7/build/inst/lib/python3/dist-packages/UM/JobQueue.py", line 125, in run
2017-09-07 15:40:04,288 - ERROR - UM.Logger.logException [80]:   File "/build/2.7/build/inst/lib/python3/dist-packages/cura/MultiplyObjectsJob.py", line 53, in run
2017-09-07 15:40:04,291 - ERROR - UM.Logger.logException [80]:   File "/build/2.7/build/inst/lib/python3/dist-packages/cura/ShapeArray.py", line 49, in fromNode
2017-09-07 15:40:04,294 - ERROR - UM.Logger.logException [80]: AttributeError: 'NoneType' object has no attribute 'getMinkowskiHull'
```
It looks like the group SceneNode doesn't have a ConvexHullDecorator.

nallath:
Odd, I just tried this and it works. Could you post some reproduction steps? 
I did the following;
1. Add 1 object
2. duplicate it
3. group both objects
4. Multiply the group

eloipr:
I've done the same steps as you.
I get the error when after grouping both objects I multiply the group without unselecting it. If after grouping the objects I unselect the group (click somewhere else) and then multiply it, it works.

I've done some debugging and when I unselect the group, the method __onChangeTimerFinished_ in PlatformPhysics.py is called and a ConvexHullDecorator is added to the group node. Then multiply works.

nallath:
Ah, it could be that i did unselect them in between. But if you've already found a fix, could you make a pull request for that? We also use the pull requests to give credits in our changelogs.

eloipr:
Ok, pull request created!

LipuFei:
Closed via #2418

PR #2418, opened 2017-09-12
add ConvexHullDecorator when grouping

eloipr:
When grouping nodes, add a ConvexHullDecorator to the group SceneNode.
When multiplying the group it was crashing because it didn't have the ConvexHullDecorator.


LipuFei:
Note that this fix is for https://github.com/Ultimaker/Cura/issues/2397

LipuFei:
For Dev, see CURA-4316

LipuFei:
Wrong dev number...

PR #2423, opened 2017-09-13
CURA-4182 Slider for quality profile

LipuFei:
slider_1
slider_2

### DIFF
--- a/cura/QualityManager.py
+++ b/cura/QualityManager.py
@@ -82,6 +82,17 @@ class QualityManager:
 
         return list(common_quality_types)
 
+    def findAllQualitiesForMachineAndMaterials(self, machine_definition: "DefinitionContainerInterface", material_containers: List[InstanceContainer]) -> List[InstanceContainer]:
+        # Determine the common set of quality types which can be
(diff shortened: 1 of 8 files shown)


fieldOfView:
No more custom profiles that can be used in Simple mode ? 👎

fieldOfView:
step-4

ChrisTerBeke:
Hi @fieldOfView. What we're trying to fix with this PR is that it was unclear to users what the impact of a profile was for quality vs speed. We think that custom profiles belong in the custom settings.

fieldOfView:
The power of having custom profiles (ie: Quality Changes profiles) in Recommended mode is that you could give a user a good working profile for a certain printer/material or even for a certain feature, without showing the complexity that is behind it. Eg a profile for printing vases, without having to hunt down vase mode.

awhiemstra:
Whoa, that is a lot of code just to do a slider with tick marks. Have you looked at how any of the QML controls implement this?

nallath:
Yeah, we did. It's not just the tick marks, it's also about disallowing certain parts of the slider and showing different messages on certain parts of the slider.

PR #2425, opened 2017-09-13
CURA-4224 New firmware notification

LipuFei:


plugins/FirmwareUpdateChecker/FirmwareUpdateCheckerJob.py
```diff
@@ -0,0 +1,86 @@
...
+            request = urllib.request.Request(self._url)
+            current_version_file = urllib.request.urlopen(request)
+            reader = codecs.getreader("utf-8")
+
+            machine_name = self._container.getId()
+
+            # If it is not None, then we compare between the checked_version and the current_version
+            # Now we just do that if the active printer is Ultimaker 3 or Ultimaker 3 Extended or any
+            # other Ultimaker 3 that will come in the future
+            if machine_name[0:11] == "Ultimaker 3":
```
fieldOfView: What happens with all other printers? Just nothing?

nallath:
Is there a reason we didn't do this with a machine action? Imho it makes way more sense to set this as part of a machine instead of doing it in an extension.

fieldOfView:
A machineaction would give this a button in the Printers preference pane, whereas this is meant to do its magic in the background similar to how Cura checks for a new version of itself on startup.

PR #2450, opened 2017-09-18
Fix missing setting_version after translation

### LINKED ISSUE #2117: Successful imported profile not showing up
Windows 10
Cura 2.6.2
WanHao i3

This looks like the same #1873 (closed) and possibly #2023. Im importing the following profile from WanHao Cura 16.01 and I get a "Profile Successfully Imported" message but the profile doesnt show in my list of profiles.

[WHLQ.zip](https://github.com/Ultimaker/Cura/files/1149802/WHLQ.zip)

ClemaX:
Resolves #2117.
https://github.com/Ultimaker/Cura/commit/0f5814e52f8b7ae2373dede03c16c0883589ba67 implemented a check against the "setting_version" metadata when importing profiles. Legacy profile translation does not add this entry, which leads to Cura failing to import the translated profile.

### DIFF
--- a/plugins/LegacyProfileReader/LegacyProfileReader.py
+++ b/plugins/LegacyProfileReader/LegacyProfileReader.py
@@ -141,4 +141,5 @@ class LegacyProfileReader(ProfileReader):
         profile.setDirty(True)
         profile.addMetaDataEntry("type", "quality_changes")
         profile.addMetaDataEntry("quality_type", "normal")
+        profile.addMetaDataEntry("setting_version", 2)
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is not the desired solution. You may _say_ that the settings are at version 2, but they aren't. They need to be upgraded to version 2 first. The legacy profile importer imports the profiles to the format they were in at Cura 2.1, but settings have been deleted since then, and renamed, and have their units changed. This is the reason for checking against the setting version.

PR #2480, opened 2017-09-25
Make Solid View a required plugin

fieldOfView:
This PR makes "Solid View" a required plugin. Without Solid View, Cura will start with a blank viewport and spew lots of critical errors because the "view" is None.

Note: There is no plugin with the id "MeshView".

Prevents https://ultimaker.com/en/community/51244-cura-27-3d-interface-is-gone


nallath:
Hmm, that was some damn old code.

Issue #2307, opened 2017-08-22
Aborting print -> auto home Y-X-Z end ?


Jhinta:
When printing with 2.6 and pressing abort, the print just wil halt
When printing with 2.7 and pressing abort it seems like its restarting start-code. ( home y-x and z)

Both 2.7 and 2.6 use the same start and end code.
Whats wrong here?

End-Code
`M104 S0 ;extruder heater off
M140 S0 ;heated bed heater off (if you have it)
G91 ;relative positioning
G1 E-1 F300 ;retract the filament a bit before lifting the nozzle, to release some of the pressure
G1 Z+0.5 E-5 X-20 Y-20 F9000 ;move Z up a bit and retract filament even more
G28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way
G90 ;absolute positioning
G1 Y190 F9000 ;use this line if you want the bed to move to the front. delete if not.
M84 ;steppers off`

Ghostkeeper:
Yes, I added this here: 6d2b46244f910cc5f09b07bdde1955bdcde77afb

Otherwise when you abort, it would keep the nozzle stuck on the print and make a mess. This was intended.

It's not the start or end code that it executes. It just homes the nozzle now.

Ghostkeeper:
Is there anything wrong with this approach?

Jhinta:
Yes, its also homes Z , it sould not do that .
only do home all but Z (Z save zone problems !)
And why would one do a hame z when there is a abort?

Jhinta:
Yes, its also homes Z , it sould not do that .
only do home all but Z (Z save zone problems, and when you have a big/multi project on border it will distroy your project) !)
And why would one do a home Z when there is a abort?

Jhinta:
more like this i think , move Z up a litle and then home Y & X
self._sendCommand("G1 Z+0.5 ")
self._sendCommand("G28 X Y")

nickthetait:
@Ghostkeeper this is a dangerous operation only for printers that home Z at the bottom

fieldOfView:
It is also an issue with Ultimaker Original, which homes at the bottom. I would say it is not necessary to home the z-axis at all.

A longer term solution would be to have an abort-gcode machine-setting in Cura (possibly along with pause-gcode and resume-gcode settings).

Ghostkeeper:
Thanks to @fieldOfView's fix, it now only homes X and Y, not Z.

[...]

PR #2481, opened 2017-09-25
refine abort sequence

krisrok:
#2307 homing the head with "G28" does not work on my printer for some reason. "G28 X" and "G28 Y" however do.
this fix should not break behaviour on other printers as "G28" is just a shortcut.


nallath:
Does it also work to use G28 X Y on your machine? That way it's a bit faster (eg it homes both of them at once)

fieldOfView:
In the long run I still vote for configurable abort, pause & resume gcode snippets.

krisrok:
@nallath nope, sadly that does not work. shitty firmware, i guess.

krisrok:
@fieldOfView yep of course. but until then...

PR #2483, opened 2017-09-26
Sorting imports and removing superfluous import

thopiekar:
.. Logger has been imported twice..

### DIFF
--- a/plugins/RemovableDriveOutputDevice/RemovableDrivePlugin.py
+++ b/plugins/RemovableDriveOutputDevice/RemovableDrivePlugin.py
@@ -1,15 +1,19 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
+# Python build-ins
 import threading
 import time
 
+# Uranium/Messaging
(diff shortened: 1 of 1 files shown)


plugins/RemovableDriveOutputDevice/RemovableDrivePlugin.py
```diff
@@ -1,15 +1,19 @@
 # Copyright (c) 2015 Ultimaker B.V.
 # Uranium is released under the terms of the AGPLv3 or higher.
 
+# Python build-ins
```
fieldOfView: built-ins?

thopiekar:
Well, that's how I call them 🙄
The modules, which normally come with the Python sources including the interpreter itself..

fieldOfView:
build vs built (present vs past perfect tense). I generally don't build Python, but I use builds that someone else built. And he/she built in the built-in modules in that build (probably while sitting in a building).

nickthetait:
Isn't english a lovely language :upside_down_face:

Ghostkeeper:
Still works. It seems that the Message is also an unused import, but I'll see to that myself (and the spelling).

PR #2484, opened 2017-09-26
Use screenScaleFactor to consistently adapt sizes to display pixel ratio

fieldOfView:
This PR attempts to solve the Hi DPI issues on both OSX and Windows. OSX apparently no longer needs the Screen.devicePixelRatio workaround when specifying pixel sizes, but Windows certainly does.

This PR is for testing an approach where instead of using Screen.devicePixelRatio, we use a Uranium-supplied screenScaleFactor. This context variable was introduced a couple of releases ago, but never used.

This PR requires https://github.com/Ultimaker/Uranium/pull/285

For additional discussion see https://github.com/Ultimaker/Cura/commit/7bc715a9bd5e61efcfef4b54f7d004b3f6248f5a#commitcomment-24539576

### DIFF
--- a/plugins/3MFReader/WorkspaceDialog.qml
+++ b/plugins/3MFReader/WorkspaceDialog.qml
@@ -13,7 +13,7 @@ UM.Dialog
     title: catalog.i18nc("@title:window", "Open Project")
 
     minimumWidth: 500 * screenScaleFactor
-    minumumHeight: 400 * screenScaleFactor
+    minimumHeight: 400 * screenScaleFactor
     width: minimumWidth
     height: minumumHeight
 
--- a/plugins/3MFReader/WorkspaceDialog.qml
+++ b/plugins/3MFReader/WorkspaceDialog.qml
@@ -12,11 +12,13 @@ UM.Dialog
 {
     title: catalog.i18nc("@title:window", "Open Project")
 
-    width: 500
-    height: 400
+    minimumWidth: 500 * screenScaleFactor
+    minumumHeight: 400 * screenScaleFactor
+    width: minimumWidth
+    height: minumumHeight
 
-    property int comboboxHeight: 15
-    property int spacerHeight: 10
+    property int comboboxHeight: 15 * screenScaleFactor
+    property int spacerHeight: 10 * screenScaleFactor
 
     onClosing: manager.notifyClosed()
     onVisibleChanged:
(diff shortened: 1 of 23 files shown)


Appesteijn:
Devs => CURA-4372

plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -12,11 +12,13 @@ UM.Dialog
 {
     title: catalog.i18nc("@title:window", "Open Project")
 
-    width: 500
-    height: 400
+    minimumWidth: 500 * screenScaleFactor
+    minumumHeight: 400 * screenScaleFactor
```
Appesteijn: Typo: minimumHeight

fieldOfView: Good catch; 5659ea0

fieldOfView:
I have included a fix for the text in the splash-screen, since that is semi-related.

Appesteijn:
It all looks good on MacOS with a High DPI screen

Appesteijn:
Looks good on Windows normal DPI screen, 100%. Only issue is the extruder tool top position.

Appesteijn:
Weird things in Windows10&200%
capture
capture2
capture3
capture4
capture5

fieldOfView:
The first two seem to be upstream issues (Qt), and are not caused by this PR. The others I will hunt down and fix.

resources/qml/Toolbar.qml
```diff
@@ -67,7 +67,7 @@ Item
             }
         }
 
-        Item { height: UM.Theme.getSize("default_margin").height; width: 1; visible: extruders.count > 0 }
+        Item { height: UM.Theme.getSize("default_margin").height; width: UM.Theme.getSize("default_lining"); visible: extruders.count > 0 }
```
ChrisTerBeke: This line causes the extruder tools to not have the correct top position. In order to fix it it should be changed to `UM.Theme.getSize("default_lining").width`. I've tested this locally but in order to simplify the merging process I'd be great if you can change that in your PR!

fieldOfView: Fixed, thanks

fieldOfView:
Has the PR been tested on Linux too?

Appesteijn:
Not yet, working on it

nallath:
Yup!

PR #2494, opened 2017-09-26
Add definitions for the Raise3D N2 series

fieldOfView:
This PR adds definitions for the Raise3D N2 printers (standard/extended and single/dual).

See https://ultimaker.com/en/community/51242-please-add-these-raise3d-printer-definitions for source and details.

### DIFF
--- a/resources/definitions/raise3D_N2_dual.def.json
+++ b/resources/definitions/raise3D_N2_dual.def.json
@@ -1,0 +1,95 @@
+{
+    "id": "raise3D_N2_dual",
+    "version": 2,
+    "name": "Raise3D N2 Dual",
+    "inherits": "fdmprinter",
+    "metadata": {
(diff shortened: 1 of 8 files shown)


resources/definitions/raise3D_N2_dual.def.json
```diff
@@ -0,0 +1,95 @@
...
+        {
+            "default_value": [
+                [ -75, 35 ],
+                [ -75, -18 ],
+                [ 18, 35 ],
+                [ 18, -18 ]
+            ]
+        },
+        "machine_min_cool_heat_time_window": {
+            "default_value": 3600
```
Ghostkeeper: Huh, so you basically never want it to go to stand-by temperature, right? Why did you do this? Seems like 5 minutes should be plenty for any PID regulator.

PR #2498, opened 2017-09-27
Fix infill slider rounding

ChrisTerBeke:
* Fixes using infill density values that are not a multiple of 10 in custom quality profiles (CURA-4381).
* Fixes sending infill density as integer to Python instead of auto parsing the string as float (CURA-4352).


fieldOfView:
Seems to fix the issue. 👍

diegopradogesto:
Everything seems to be OK. It's gonna be merged.

PR #2500, opened 2017-09-27
Move the quality indicator to the center when there is only one quality

### LINKED ISSUE #2485: When there is only 1 quality, Recommended mode always shows "slower"
With the new quality selection slider, if there is only a single available quality, the slider is always showing a "Slower" print speed. This will not look great when the user picks a 0.8mm nozzle for their Ultimaker 2+ (you would think that a 0.8mm slice at 0.2mm layer height is not "slow").

From a UX perspective, it would be better if the disabled "nub" that is visible when there is only a single layerheight is in the center (neither slower nor faster).

image

fieldOfView:
This PR fixes a UX issue where if only a single quality is available, it always shows a "Slower". If a single quality is available, the quality indicator (handle) is now moved to the center of the slider.

Fixes #2485
CURA-4383

### DIFF
--- a/resources/qml/SidebarSimple.qml
+++ b/resources/qml/SidebarSimple.qml
@@ -226,7 +226,7 @@ Item
                     Repeater
                     {
                         id: qualityRepeater
-                        model: qualityModel.availableTotalTicks > 0 ? qualityModel : 0
+                        model: qualityModel.totalTicks > 0 ? qualityModel : 0
 
                         Rectangle
                         {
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
There's an issue where if you change the material of the extruder back to something that 'unlocks' more values on the slider, the value doesn't update and the slider handle is in the middle between 2 tick marks (for example: PVA + PVA -> ABS + PVA).

fieldOfView:
Fixed

ChrisTerBeke:
Great! Will have a look first thing tomorrow morning and then merge.

ChrisTerBeke:
Changes are good, Thanks for this PR!

PR #2335, opened 2017-08-29
Hide adhesion & support options if they are disabled

fieldOfView:
This PR hides the adhesion & support options from the ~~~Simple~~~ Recommended mode when they are disabled by the machine definition. One example of a machine that disables the support options is the Peopoly Moai. This hides the support options from the ~~~Advanced~~~ Custom mode sidebar, but they are still there on the Recommended mode tab. This PR removes the settings from the Recommended mode tab if they are removed from the Custom mode tab.

PR #2503, opened 2017-09-27
Hide gradual infill if its setting is disabled

fieldOfView:
This PR hides the gradual infill checbox from the simple mode tab, if the "gradual_infill_steps" setting is disabled by the current machine definition.

Followup to https://github.com/Ultimaker/Cura/pull/2335


ChrisTerBeke:
Reviewing code right now.

PR #2507, opened 2017-09-28
CL-462

daid:
time sync to the printer if a printer has no valid system datetime. As any datetime is better then no datetime at all.
Note: Depends on EM-1896 for the new API endpoint.

### DIFF
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
@@ -27,6 +27,7 @@ import zlib
 
 from time import time
 from time import sleep
+from time import gmtime
 
 i18n_catalog = i18nCatalog("cura")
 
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
```diff
@@ -1114,6 +1115,11 @@ def _onFinished(self, reply):
                     else:
                         Logger.log("w", "Unable to save authentication for id %s and key %s", self._authentication_id, self._getSafeAuthKey())
 
+                    # Request 'system' printer data once, when we know we have authentication, so we know we can set the system time.
```
robinmdh: This file makes me sad, but I guess adding it here doesn't make it any worse, 
I assume the auth/verify is always called when opening a printer/starting up?

nallath: Yep, its always called. This file also makes me sad :(

plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
```diff
@@ -1133,6 +1139,27 @@ def _onFinished(self, reply):
...
+                except json.decoder.JSONDecodeError:
+                    Logger.log("w", "Received an invalid authentication request reply from printer: Not valid JSON.")
+                    return
+                if "time" in data and "utc" in data["time"]:
+                    try:
+                        printer_time = gmtime(float(data["time"]["utc"]))
+                        Logger.log("i", "Printer has system time of: %s", str(printer_time))
+                    except ValueError:
+                        printer_time = None
+                    if printer_time is not None and printer_time.tm_year < 1990:
```
nallath: If printer_time is none, it's also likely to be not valid right?

daid: Or the API returned something we don't understand. Figured I opt for the safe route of not handling "bad" data in any way other then doing nothing with it.

nallath: Fair enough.

PR #2224, opened 2017-08-08
Fix multiextruder menus

fieldOfView:
This PR fixes the application menus for materials and variants for multiextrusion printers. All extruder menus would always act on the currently active extruder only, instead of acting on the extruder suggested by the menu. 

**Reproduce steps:**
* Add a new UM3 printer to Cura, do not connect it
* Note that both extruders have PLA for material and that Extruder **1** is the selected extruder tab in the sidebar.
* From the Settings application menu, choose Extruder **2** -> Material -> ABS

**Expected result:**
The material of Extruder 2 (2nd tab) changes

**Actual result:**
The material of Extruder 1 (1st tab) changes instead

The PR fixes #1641, an issue that has been open since Cura 2.3. It may or may not have gotten a JIRA ticket nr.

PR #2513, opened 2017-09-28
[WIP] - Fix multiple extruder issues

diegopradogesto:
Adding some fixes to some problems that exist when switching number of extruders in multiextruder printers. The idea is to keep `active_extruder_index = -1` when there is just one extruder, and then the global stack is taken into account. Also, the `active_extruder_stacks` don't exist in the case we just have one extruder, so the global stack will be used to take the materials, qualities and so on.

There were some points in which the UI is setting the active extruder index value, think that it's consider to be done in python. So all these references were deleted and now it's done through signals.

Those fixes try to solve CURA-4386 and CURA-4379

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -41,7 +41,7 @@ class ExtruderManager(QObject):
     def __init__(self, parent = None):
         super().__init__(parent)
         self._extruder_trains = { } #Per machine, a dictionary of extruder container stack IDs. Only for separately defined extruders.
-        self._active_extruder_index = -1
+        self._active_extruder_index = -1 # Indicates the index of the active extruder stack. -1 means no active extruder stack
         self._selected_object_extruders = []
         Application.getInstance().globalContainerStackChanged.connect(self.__globalContainerStackChanged)
         self._global_container_stack_definition_id = None
@@ -74,15 +74,18 @@ class ExtruderManager(QObject):
         except KeyError:
             return 0
 
+    ##  Gets a dict with the extruder stack ids with the extruder number as the key.
+    #   The key "-1" indicates the global stack id.
+    #
     @pyqtProperty("QVariantMap", notify = extrudersChanged)
     def extruderIds(self):
-        map = {}
+        extruder_stack_ids = {}
         global_stack_id = Application.getInstance().getGlobalContainerStack().getId()
-        map["-1"] = global_stack_id
+        extruder_stack_ids["-1"] = global_stack_id
         if global_stack_id in self._extruder_trains:
             for position in self._extruder_trains[global_stack_id]:
-                map[position] = self._extruder_trains[global_stack_id][position].getId()
-        return map
+                extruder_stack_ids[position] = self._extruder_trains[global_stack_id][position].getId()
+        return extruder_stack_ids
 
     @pyqtSlot(str, result = str)
     def getQualityChangesIdByExtruderStackId(self, id: str) -> str:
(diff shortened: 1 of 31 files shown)


resources/qml/Menus/MaterialMenu.qml
```diff
@@ -70,13 +66,7 @@ Menu
             checkable: true
             checked: model.id == Cura.MachineManager.allActiveMaterialIds[ExtruderManager.extruderIds[extruderIndex]]
             exclusiveGroup: group
-            onTriggered:
-            {
-                var activeExtruderIndex = ExtruderManager.activeExtruderIndex;
-                ExtruderManager.setActiveExtruderIndex(extruderIndex);
-                Cura.MachineManager.setActiveMaterial(model.id);
-                ExtruderManager.setActiveExtruderIndex(activeExtruderIndex);
-            }
```
fieldOfView: There was a reason this code was in place. See https://github.com/Ultimaker/Cura/pull/2224. It looks like this commit breaks that again.

diegopradogesto: Oh, I will take a look at it. Thank you for the comment

diegopradogesto: I have reverted your workaround and add some comments. The other issues seem to work as well.

cura/Settings/ExtruderManager.py
```diff
@@ -41,7 +41,7 @@ class ExtruderManager(QObject):
     def __init__(self, parent = None):
         super().__init__(parent)
         self._extruder_trains = { } #Per machine, a dictionary of extruder container stack IDs. Only for separately defined extruders.
-        self._active_extruder_index = 0
+        self._active_extruder_index = -1
```
nallath: Some documentation would be nice; I assume that we use -1 as "not set" right?

cura/Settings/ExtruderManager.py
```diff
@@ -78,6 +78,7 @@ def extruderCount(self):
     def extruderIds(self):
         map = {}
```
nallath: Just noticed this; But map shadows a built in. Could this be re-factored to a better name?

cura/Settings/ExtruderManager.py
```diff
@@ -513,17 +514,32 @@ def getActiveExtruderStacks(self) -> List["ExtruderStack"]:
         global_stack = Application.getInstance().getGlobalContainerStack()
 
         result = []
+        machine_extruder_count = global_stack.getProperty("machine_extruder_count", "value")
+
+        if machine_extruder_count is 1:
```
nallath: is shouldn't be used in this case; should be == 
Also; document why we are aborting already.

cura/Settings/ExtruderManager.py
```diff
@@ -513,17 +514,32 @@ def getActiveExtruderStacks(self) -> List["ExtruderStack"]:
...
             self._global_container_stack_definition_id = global_container_stack.getBottom().getId()
             self.globalContainerStackDefinitionChanged.emit()
+
+        # If the global container changed, the number of extruders could be changed and so the active_extruder_index is updated
+        extruder_count = global_container_stack.getProperty("machine_extruder_count", "value")
+        if extruder_count > 1:
+            if self._active_extruder_index == -1:
+                self.setActiveExtruderIndex(0)
+        else:
+            if self._active_extruder_index > -1:
```
nallath: Might be a bit clearer if we explictly check for != -1 (As we stated that -1 is a special case)  (minor issue)

diegopradogesto:
Just added some minor changes and commented the code.

PR #2514, opened 2017-09-28
Add definition for Tevo Tarantula printer

TheAssassin:
Modified version of the Prusa i3's definition. While the i3 values work, the print time estimations were horribly imprecise (often the print takes 2x to 3x as long as Cura estimated). Therefore, I inspected the problem, and found a Reddit thread [1] about it. Instead of hacking my own installation, I just wrote a defition including the correct acceleration/jerk values that work with my firmware (custom build of Marlin based on https://github.com/JimBrown/MarlinTarantula/).

There's downsides however. I don't know how it performs with the stock firmware (although I doubt there'd be many differences). Furthermore, the `jerk_xy` is set to 4 in my definition, but that's only true for the `x`value,the Marlin firmware has a `y` jerk value of 7. That leads to minor differences in the printing time (e.g., just finished a print which Cura estimated with 13h 30m, but it took 12h 50m).

Feedback from stock firmware users is therefore highly welcome. Also, if someone could provide information about the jerk values and help improving those inaccuracies, that'd be awesome.

[1] https://www.reddit.com/r/3Dprinting/comments/5wz0xs/

### DIFF
--- a/resources/definitions/tevo_tarantula.def.json
+++ b/resources/definitions/tevo_tarantula.def.json
@@ -1,0 +1,73 @@
+{
+    "id": "tevo_tarantula",
+    "version": 2,
+    "name": "Tevo Tarantula",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "TheAssassin",
+        "manufacturer": "Tevo",
+        "file_formats": "text/x-gcode",
+        "icon": "icon_ultimaker2",
+        "platform": "prusai3_platform.stl"
+    },
+
+    "overrides": {
+        "machine_name": {
+            "default_value": "Tevo Tarantula"
+        },
+        "machine_heated_bed": {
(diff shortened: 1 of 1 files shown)


TheAssassin:
I'd love to see this get merged soon.

nallath:
Sorry for the delay, we're in a bit of a crunch while we are finishing up the 3.0 release. 

For our reference; CURA-4421

ChrisTerBeke:
The definition file looks OK and will be merged today. We cannot check print quality as we don't have this printer available, but I trust you've done testing with an actual Tevo machine?

ChrisTerBeke:
Also merged in the 3.0 branch, due for release on October 17th.

TheAssassin:
Yes, I own such a printer, and the print quality is really good as far as I can tell. Thanks for merging.

PR #2517, opened 2017-09-29
CL-460 do not downgrade Cura Connect printer to legacy

jackha:
https://jira.ultimaker.com:8443/browse/CL-460


sedwards2009:
The code appears to be correct. My only concern is whether it is a good idea to hang on to the old PrinterDevice object when normally they would be destroyed. Maybe it is safer just to remember the ID of the printers which were cluster printers.

jackha:
I've changed it to remember only names of cluster enabled printers.

PR #2521, opened 2017-09-29
Empty Materials and Print Cores or Nozzle

diegopradogesto:
Sometimes when loading a new project, the materials and print cores and nozzles appear as "empty". Even if the material, print cores or nozzles were properly selected, the UI doesn't update to the right value.

It also happens when creating a new "Custom FDM" or UM2+ or UM2E+:
**Reproduction Steps:** 
1. Add UM3.
2. Add UM2+
3. Check that material and nozzles, set to "empty"

This solves CURA-4377

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -354,11 +354,16 @@ class MachineManager(QObject):
         if containers:
             Application.getInstance().setGlobalContainerStack(containers[0])
 
+        self.activeQualityChanged.emit()
+        self.activeVariantChanged.emit()
+        self.activeMaterialChanged.emit()
+
     @pyqtSlot(str, str)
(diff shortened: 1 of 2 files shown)


cura/Settings/MachineManager.py
```diff
@@ -354,11 +354,16 @@ def setActiveMachine(self, stack_id: str) -> None:
         if containers:
             Application.getInstance().setGlobalContainerStack(containers[0])
 
+        self.activeQualityChanged.emit()
```
nallath: A better fix is to use __onInstanceContainersChanged(); That will do all this and check if there are errors (or just trigger the _onInstanceContainerChanged with any container.

diegopradogesto: Thanks, I don't realize about that. Already changed.

PR #2538, opened 2017-10-02
Update camera 'viewport_overlay' color > GREY

Jerooney:
This will apply to all UMO+, UM2, UM2+ and UM3 with old firmware users, which look at the camera view. The background changed from blue > grey.


nallath:
Aslo cherry-picked this to 3.0

PR #2541, opened 2017-10-02
CL-228 CL-229 authorization and fixup state display

jackha:
https://jira.ultimaker.com:8443/browse/CL-229

Changes:
- Printer is always authenticated. 
- The job status display now falls back to the printer status.

This solves:
- No authentication: This gets rid of Cura asking the printer for authentication and it solves the "Request Access" problem
- Job status: if the printer is in maintenance or unreachable, it now shows "Unknown" because of the string freeze
- When aborting a job, the printer is in system state "printing" and print job state "none". In the old situation this would render into "Available", now it is "Printing" (just like the web UI).

Open:
- System state "printing" and print job state "none" is going to be fixed in the firmware EM-1904.


plugins/UM3NetworkPrinting/NetworkClusterPrinterOutputDevice.py
```diff
@@ -111,6 +112,20 @@ def __init__(self, key, address, properties, api_prefix, plugin_path):
         temporary_translation3 = i18n_catalog.i18n("{printer_name} has finished printing '{job_name}'. Please collect the print and confirm clearing the build plate.") #When finished.
         temporary_translation4 = i18n_catalog.i18n("{printer_name} is reserved to print '{job_name}'. Please change the printer's configuration to match the job, for it to start printing.") #When configuration changed.
 
+    ##  No authentication, so requestAuthentication should do exactly nothing
+    @pyqtSlot()
+    def requestAuthentication(self, message_id = None, action_id = "Retry"):
+        Logger.log("d", "requestAuthentication for Cura Connect - nothing to be done")
```
nallath: No need to log this imho. I'd also like to recommend using the @override decorator.

PR #2544, opened 2017-10-02
Revert "CL-523: Made sure printers are present before printjobs to avoid referring to…"

jackha:
Reverts Ultimaker/Cura#2542


jackha:
Should have been merged into 3.0

nallath:
You know that there is such a thing as cherry-pick right?

PR #2551, opened 2017-10-03
Move text back to label and remove unused font

Appesteijn:
This should be merged together with an Uranium pull request:

https://github.com/Ultimaker/Uranium/pull/287

I replaced the qml Texts by Labels (yes @fieldOfView you were right) and in the Uranium PR the font sizes are then set to point size instead of pixel size. Pixel size seems to be messed up on high DPI screens.

### DIFF
--- a/resources/qml/SaveButton.qml
+++ b/resources/qml/SaveButton.qml
@@ -43,7 +43,7 @@ Item {
         }
     }
 
-    Text {
+    Label {
         id: statusLabel
         width: parent.width - 2 * UM.Theme.getSize("sidebar_margin").width
         anchors.top: parent.top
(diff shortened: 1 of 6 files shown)


fieldOfView:
> (yes @fieldOfView you we're right)

🍾 ;-)

ChrisTerBeke:
My last commit here removes all remaining occurrences of Text and makes sure that Labels never end up in a position that is not a whole pixel (causes font glitching on Windows).

fieldOfView:
If this still causes issues, there is another way fractional sizes can affect texts than just margins. If a label aligns with a sibling that has eg a fractional width, then the Label will still be on a fractional position. But lets see how builds with the current combination of fixes behave.

PR #2560, opened 2017-10-06
Updated brazilian portuguese strings

Patola:


Ghostkeeper:
As discussed via e-mail, I'll remove the language and country codes myself. They are not defined by the file format specification.

PR #2561, opened 2017-10-06
Fix the OpenGL shader for older (legacy?) OpenGL.

daid:
The patch to do the grid rendering from a shader wasn't done a 100% right, as the "legacy" shader wasn't compiling. I also fixed a warning that the intel driver is reporting.
After this fix the grid renders again on my Intel HD 3000 master-of-disaster rendering device.

@fieldOfView still love you ;-)

### DIFF
--- a/resources/shaders/grid.shader
+++ b/resources/shaders/grid.shader
@@ -31,10 +31,10 @@ fragment =
         vec4 minorGridColor = mix(u_plateColor, u_gridColor1, 1.0 - min(minorLine, 1.0));
 
         // Compute anti-aliased world-space major grid lines
-        vec2 majorGrid = abs(fract(coord / 10 - 0.5) - 0.5) / fwidth(coord / 10);
+        vec2 majorGrid = abs(fract(coord / 10.0 - 0.5) - 0.5) / fwidth(coord / 10.0);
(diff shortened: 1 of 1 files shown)


nallath:
CURA-4424

fieldOfView:
Thanks for fixing up my shoddy work @daid :rose:

daid:
@fieldOfView the fact that there are 2 implementations of each shader that are slightly different doesn't help. Took me a while to figure out that I was looking at the wrong shader code initially...

nallath:
Yeah. I tried to resolve the whole two shader implementation, but OSX seemed to think otherwise. This was the only way to get all platforms to work.

awhiemstra:
The problem is that the two GLSL dialects are so different that it becomes really hard to share the shader code. There is unfortunately no real way around that, because as soon as you use OpenGL Core Profile you lose support for the old version, and OSX only allows OpenGL > 3.0 using Core.

fieldOfView:
Discussion about OpenGl implementations aside, it would be nice if this were merged into 3.0.

fieldOfView:
Aww, this was merged into master but not cherrypicked to 3.0, so older GPUs show a white or black buildplate (depending on the GPU driver).

Any chance of a Cura 3.0.4?

PR #2564, opened 2017-10-07
slicing mode

BagelOrb:
Why don't I have push access to Cura any more?

This PR accompanies https://github.com/Ultimaker/CuraEngine/pull/620

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -643,6 +643,20 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
                 },
+                "slicing_mode":
+                {
+                    "label": "Slicing Mode",
+                    "description": "How to slice layers with diagonal surfaces. The areas of a layer can be generated based on where the middle of the layer intersects the surface (Middle). Alternatively each layer can have the areas which fall inside of the volume throughout the height of the layer (Exclusive) or a layer has the areas which fall inside anywhere within the layer (Inclusive). Exclusive retains the most details, Inclusive makes for the best fit and Middle takes the least time to process.",
+                    "type": "enum",
+                    "options":
+                    {
+                        "middle": "Middle",
+                        "exclusive": "Exclusive",
+                        "inclusive": "Inclusive"
+                    },
+                    "default_value": "middle",
+                    "settable_per_mesh": true
+                },
                 "line_width":
                 {
                     "label": "Line Width",
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -643,9 +643,9 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
                 },
-                "slicing_mode":
+                "slicing_tolerance":
                 {
-                    "label": "Slicing Mode",
+                    "label": "Slicing Tolerance",
                     "description": "How to slice layers with diagonal surfaces. The areas of a layer can be generated based on where the middle of the layer intersects the surface (Middle). Alternatively each layer can have the areas which fall inside of the volume throughout the height of the layer (Exclusive) or a layer has the areas which fall inside anywhere within the layer (Inclusive). Exclusive retains the most details, Inclusive makes for the best fit and Middle takes the least time to process.",
                     "type": "enum",
                     "options":


Ghostkeeper:
We set push access to be only allowed from the Cura team, because we were getting some commits in from the Connect project that caused instability. I think that's when it got changed.

Developers, see issue CURA-4459.

alekseisasin:
I think the "settable_per_extruder" should be false, because if I have two models with per model setting Slicing Tolerance  (Exclusive and Inclusive) then CuraEngine crashes:
"Backend quit with return code -11. Resetting process and socket. "

alekseisasin:
For UM3, whenever I change the Slice Tolerance in log file I see it is not changed and uses default = "middle". I think same value should be shared for all extruders.

BagelOrb:
I think it's a bit weird that you see a bug and conclude that it's probably because the use-case was never supported.

I've fixed the bug. It doesn't crash any more when you slice multiple models.

Do note that inclusive slices at z=0, while there is no model there because it is hovering the slightest above the build plate. This may cause there not to be a first layer. I think this is by design..

smartavionics:
I just tried this feature and when I use Exclusive with layer height and initial layer height set to 0.3 my first layer completely disappears! I have tried this with various models and they all seem to be affected. None of those models have previously shown this behaviour. Lowering the models by 0.01 into the build plate makes the first layer reappear.

ChrisTerBeke:
@smartavionics thanks for reporting. Now that it's in master (and still defaulting to 'middle') it's way easier for us to test on different models and with different modes.

@BagelOrb has this to do with your change of calculating the first layer z height?

BagelOrb:
This is by design. Because the initial layer goes from 0 to 0.3 and the model starts at 0.001 the initial layer should be empty.

It doesn't happen when the initial layer height is larger than the normal layer height, because the initial layer is sliced at `initial_layer_height` and `initial_layer_height - layer_height` instead of at `initial_layer_height` and at `0`. Basically it is a bug which helps a lot for normal values of the initia layer height,

It would be nice if such situations are solved, but it is not a problem of this PR; it is a problem of hovering models in general. I don't think there is a way to solve it without breaking other situations.

smartavionics:
The model starts at 0.001? Well, I tried 3 different models none of which have ever shown a problem before and with this PR they screw up. By design, eh?

BagelOrb:
Well, yes.

Shouldn't your initial layer height be larger than your normal layer height?

smartavionics:
I guess it depends... I often use initial layer height the same as the others. Who knows, maybe people even use an initial layer smaller than the others?

BagelOrb:
I don't know what to do about it. The problem should only occur if you set the setting to different values for different models. If it only removes some initial layers then there could be rounding problems.

It would be nice to have a model sink setting so that the models don't automatically drop on the build plate, but just below. I don't really see that as part of this feature though...

smartavionics:
As it is now, I think you're going to get quite a few people wondering where their first layer has gone.

BagelOrb:
I don't think so. I thought the bug only happens when your initial layer is
same or less than the normal layer height and you set the slicing tolerance
only for some objects - not for all.

Is my thinking wrong?

smartavionics:
In my brief tests I have only had one object at a time being sliced and each object I tried had the first layer missing. Hardly an exhaustive test (just 3 out of 3!)

BagelOrb:
Oh wait. CURA-4442 intentionally left in empty layers.
I think empty layers should be removed if you have drop on build plate enabled.

I'll create a new bug report for that.

smartavionics:
> I think empty layers should be removed if you have drop on build plate enabled.

Which I don't.

ChrisTerBeke:
@smartavionics I continued the discussion on empty layers on https://github.com/Ultimaker/Cura/issues/2659, but basically I think it should be a user setting (defaulting to the older behaviour of removing the empty layers from the bottom).

BagelOrb:
Look, the first layer being empty is an inherent property of exclusive slicing if the model is located at (or just above) the build plate. Exclusive slicing takes the intersection between the polygons at the top of a layer and the polygons at the bottom of a layer. The bottom of the initial layer is always empty because the frontend places the model just above the build plate.

This means that exclusive slicing - together with the frontend Drop on Build plate functionality - cause the first layer to be empty. 

I think you rightly say that that is not something you would want.

Now we can fix this in a couple of manners:
- slice the first layer inclusive while slicing the rest of the layers exclusive
- change the Drop on Build Plate functionality such that it places models just below the build plate (only works for initial layer height values the same or lower then normal layer height, but that's a separate bug)
- drop the model on the build plate in the engine

I think the first one is a bad option, because it violates the exclusive slicing tolerance principle. Your model might end up one layer too high.

smartavionics:
I vote for the option that surprises the users least!

BagelOrb:
I don't. I vote for the option which leads to the best results.

ChrisTerBeke:
Devs: CURA-4481 (this ticket describes adding a setting for this behaviour with a default to removing the layers).

Ghostkeeper:
How does making the first layer slice inclusively cause the model to become one layer too high? It's not like all layers are shifted or anything, right? Just that the other side of the edge is chosen in this edge case. So far, the first option sounds best to me.

The third option sounds okay-ish as well, if it only drops one layer specifically for this case.

BagelOrb:
If a model occupies z 0.9 to z 2.1 and all layer heights are 1, then it will have at least the layer from 1 to 2. If you slice the initial layer inclusively the layer from 0 to 1 will also be included, making the print 2 high, rather than something less than 1.2.

Ghostkeeper:
Ah yes, I see. The same problem also occurs with the old slicing mode when you have a model that occupies z0.4 to z1.6 and all layer heights are 1. Both 0.5 and 1.5 are inside the model so the print becomes 2 high instead of 1.2 or 1.

PR #2572, opened 2017-10-10
CURA-4412 - Layerview slider refactor

ChrisTerBeke:
Original issue: The layer position on the LayerView slidebar is incorrect when switching to layerview. Perhaps this is because the processing layers is done only once?

When solving this I found the QML part of this plugin quite hard to read, so I restructured it into separate re-usable components. See it as a first step towards more re-usability in the QML codebase.

Previous implementation of the layer slider only showed the layer number label for the handle that was lasted clicked on, but after discussing with Mark we found no good reason why this was implemented this way so now showing both labels at all times.


nallath:
@fieldOfView As you made the original implementation, you might want to have a look.

awhiemstra:
Note that you might want to have a look at http://doc.qt.io/qt-5/qml-qtquick-controls2-rangeslider.html rather than reimplementing it manually.

jackha:
It looks like the PR splits off the layer slider and layer slider labels from the layerview main qml. However, I have trouble finding the original bugfix in this big change.

ChrisTerBeke:
@jackha https://github.com/Ultimaker/Cura/pull/2572/files#diff-98b29e3ac669af8f2ffc7db5f09e55daR376

ChrisTerBeke:
@awhiemstra there is an added behaviour where you can also drag the area between the 2 slider handlers. That's not supported by the default range slider component.

fieldOfView:
When viewing the layers one at a time (ie: dragging the two handles all the way together and then dragging the middle part), the two layer number boxes overlap and show the same number, which looks somewhat silly.

It no longer seems to be possible to use the cursor keys to view one layer at a time, which is quite the regression.

ChrisTerBeke:
The cursor keys still work for me. Just select one of the text fields and use the up and down arrows...

fieldOfView:
Yes, the cursor keys work, but it is no longer possible to move the upper and lower handle at the same time with the cursor keys.

Also the when pushing the lower handle all the way up, it goes one step to far when using the cursor keys.

ChrisTerBeke:
Thanks, we'll get those things fixed.

ChrisTerBeke:
CURA-4432

ChrisTerBeke:
Added back active slider behavior in https://github.com/Ultimaker/Cura/commit/a7368e17518079219d22972d5b79d3e0045c9024. Was not able to reproduce your comment about the lower handle being able to go 1 step too far.

fieldOfView:
> Was not able to reproduce your comment about the lower handle being able to go 1 step too far.

That seems to be gone now.

Remaining unexpected behavior: the active handle only changes when clicking in the layer number box:
* move the lower handle half-way up
* click on the upper handle to activate it
* click in the layer number box
* move the upper handle down using the cursor keys
* click on the lower handle, this makes the layer number box move to the bottom handle
* press the cursor down key

Expected:
lower handle moves down, because that has the layer number box
Actual:
upper handle moves down

ChrisTerBeke:
Fixed in https://github.com/Ultimaker/Cura/commit/ea01d6ad01a686ef76a586413f760ff27a04f119

PR #2583, opened 2017-10-11
Cura-4358 Build Plate Adhesion default value takes the current value

ChrisTerBeke:
When BPA type is modified, and then the user changes Quality Profile, Cura will prompt user if they want to Keep or Discard Changes. However, the value for default is incorrect as displayed.

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -588,6 +588,46 @@ class ExtruderManager(QObject):
 
         return result
 
+    ##  Get all extruder values for a certain setting. This function will skip the user settings container.
+    #
+    #   This is exposed to SettingFunction so it can be used in value functions.
+    #
+    #   \param key The key of the setting to retrieve values for.
+    #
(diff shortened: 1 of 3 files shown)


ChrisTerBeke:
This PR is for other people than current core devs to comment on this change.

ChrisTerBeke:
Connected Uranium PR: https://github.com/Ultimaker/Uranium/pull/292.

fieldOfView:
Not just a better name; a description would be appreciated too

nallath:
There.

diegopradogesto:
What is done here is to use a context to avoid checking the user container stack when comparing the "default" and "customized" value of the settings. The initial problem was that when resolving the function in the build plate adhesion for getting the "default" value, the user container was used and then the value became the same as the "customized". Now the resolve functions such as getExtruderValue, getExtruderValues and resolveOrValue were remapped in order to use the context that avoids the first (or user) container. This fix also solves other similar problems in other settings, for instance "material_bed_temperature_layer_0".

PR #2586, opened 2017-10-11
Change Text to Label and cast all sizes to int

Appesteijn:
Together with Uranium PR: https://github.com/Ultimaker/Uranium/pull/293

CURA-4437

### DIFF
--- a/resources/qml/Preferences/MaterialView.qml
+++ b/resources/qml/Preferences/MaterialView.qml
@@ -108,15 +108,15 @@ TabView
                 {
                     width: scrollView.columnWidth;
                     height:  parent.rowHeight;
-                    spacing: UM.Theme.getSize("default_margin").width/2
+                    spacing: parseInt(UM.Theme.getSize("default_margin").width/2)
 
                     Rectangle
                     {
                         id: colorSelector
                         color: properties.color_code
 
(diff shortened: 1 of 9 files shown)


resources/qml/Preferences/MaterialView.qml
```diff
@@ -108,15 +108,15 @@ TabView
                 {
                     width: scrollView.columnWidth;
                     height:  parent.rowHeight;
-                    spacing: UM.Theme.getSize("default_margin").width/2
+                    spacing: parseInt(UM.Theme.getSize("default_margin").width/2)
```
sedwards2009: `Math.floor()`

fieldOfView: I agree that Math.floor() is better. I think parseInt will first cast the value to a string, and then parse it back to a number again. I used | 0 as shorthand for Math.floor() elsewhere, but that's a little obscure.

ChrisTerBeke: We found the | 0's indeed, but they're easily mistaken for || 0 (yay JS syntax). We can use Math.floor indeed.

Appesteijn: I ll replace all parseInt() with Math.floor()

PR #2588, opened 2017-10-11
CL-508 Show the printer unreachable status in the Cura Connect tab when needed.

sedwards2009:
See jira CL-508

Note: This PR recycles an existing string.

### DIFF
--- a/plugins/UM3NetworkPrinting/PrinterInfoBlock.qml
+++ b/plugins/UM3NetworkPrinting/PrinterInfoBlock.qml
@@ -39,7 +39,8 @@ Rectangle
                 return catalog.i18nc("@label:status", "Printing");
             case "idle":
                 return catalog.i18nc("@label:status", "Available");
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/PrinterInfoBlock.qml
```diff
@@ -165,6 +166,7 @@ Rectangle
                 anchors.right: printProgressArea.left
                 anchors.rightMargin: UM.Theme.getSize("default_margin").width
                 color: emphasisColor
+                opacity: printer != null && printer.status === "unreachable" ? 0.3 : 1
```
nallath: In what cases could the printer.status be anything but a string? eg; Why is it required to use strict comparison. As far as I see, it's actually better not to use strict comparison, as in quite a few cases the values can come from python (and thus we should assume duck typing, so if it pretends well enough to be a string, its fine!)

Issue #2069, opened 2017-07-04
Feature Request: New honeycomb infill pattern


zerspaner-gerd:
# Suggestion for new, further infill pattern!
###  **The honeycomb pattern**
![20161204_162424](https://user-images.githubusercontent.com/29805503/27842323-653cbf00-6109-11e7-9fbe-15faf008cedc.jpg)

- **It looks good**
![](https://www.simplify3d.com/wp-content/uploads/2015/06/Simplify3D-Infill-Options-2.jpg)

- There is a crucial difference to the other available infill variants, hardly overlapping lines

- Can be printed more solidly, more comfortably, more consistently, since there is no departure from the material flow

- Slightly more stability

BagelOrb:
So you mean the Full Honeycomb from S3D.
The fast honeycomb does have self-intersections..

I don't think it gives more stability, though. I think the Honeycomb is more squishy in the XY directions than triangles.

Actually I think you're the first person to give a holding argument for Honeycomb infill: the lack of overlapping lines!

DDDirk:
I have been playing with something similar, reasoning from the principle that circles (or spheres) are the strongest stuctures.
Here's a quick sketch (I really should learn to use Inkscape :) ):

![p1060106_rk_25](https://user-images.githubusercontent.com/13403306/27861195-160a45fc-6180-11e7-905d-671032b591dd.JPG)

A next row can then be added thus:

![p1060107_rk_25](https://user-images.githubusercontent.com/13403306/27861214-24fade00-6180-11e7-8b94-32f7aee9dd06.JPG)

Etc.

This is one continuous road, so no travel moves are needed. And there are no angles, which are often a problem in prints. But it may also mean the print speed can go up.
And there are more things one can do with this, such as making the lower 'circles' smaller, which will create a curving shape. Angles are a bit trickier, but can be done by making the circles increasingly smaller, then changing the direction and making the circles bigger again. 
And the sizes of the circles can vary, to make those near the outer walls smaller, to make the structure stronger there. There are loads of possible variations on this.

I suspect this will be strong (per print time) but not very stiff. And that may often be very desirable.

But on topic, of course this can be adapted to make a honeycomb pattern by making the 'circles' hexagonal. A major drawback is that whilst most sides are then double-walled, some sides will be single-walled, namely the middle ones in the first drawing.

nickthetait:
Can you think of a specific use case where this pattern is desired and no other patterns would suffice?

[...]

PR #2591, opened 2017-10-12
trihex infill

BagelOrb:
https://en.wikipedia.org/wiki/Trihexagonal_tiling

This is a very simple edit of Triangle infill: the same infill lines are just shifted a bit.

This pattern avoids the junctions where 3 lines meet. As such we can expect this pattern to fail less often.

Because it contains the same directions as triangle infill we can expect the structural properties to remain similar.

However, this hasn't been physically tested. At this moment we can't be 100% sure we can replace triangle infill with trihex infill.

Let's include it in Cura to test with it to see whether trihex infill is indeed as good as it theoretically seems to be.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1354,7 +1354,7 @@
                             "default_value": 2,
                             "minimum_value": "0",
                             "minimum_value_warning": "infill_line_width",
-                            "value": "0 if infill_sparse_density == 0 else (infill_line_width * 100) / infill_sparse_density * (2 if infill_pattern == 'grid' else (3 if infill_pattern == 'triangles' or infill_pattern == 'cubic' or infill_pattern == 'cubicsubdiv' else (2 if infill_pattern == 'tetrahedral' or infill_pattern == 'quarter_cubic' else (1 if infill_pattern == 'cross' or infill_pattern == 'cross_3d' else 1))))",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see ticket CURA-4476.

zerspaner-gerd:
> This pattern avoids the junctions where 3 lines meet.

Even better would be if no lines crossed at all. Just overlap.

> As such we can expect this pattern to fail less often.

Even less does this fail:
#2069

BagelOrb:
Yes, but this and triangle infill have only straight lines which means
little decelerations to do corners.

A pull request to implement honeycomb infill would be more than welcome I
think.

alekseisasin:
I printed this infill and I looks good. Theoretically it should be "stronger" and safer than Triangle (because 6 lines are joined at one point). I merge this request. Only one thing, how good is to have such a if statement for a new infill patterns in future?

"value": "0 if infill_sparse_density == 0 else (infill_line_width * 100) / infill_sparse_density * (2 if infill_pattern == 'grid' else (3 if infill_pattern == 'triangles' or infill_pattern == 'trihexagon' or infill_pattern == 'cubic' or infill_pattern == 'cubicsubdiv' else (2 if infill_pattern == 'tetrahedral' or infill_pattern == 'quarter_cubic' else (1 if infill_pattern == 'cross' or infill_pattern == 'cross_3d' else 1))))",

Ghostkeeper:
Yeah the formula there is not sustainable. Do you know of a better alternative though?

It's still not quite as bad as this one:
https://github.com/Ultimaker/Cura/blob/2be529668e3c0c55485369d922758a92ab0c6cbd/plugins/LegacyProfileReader/DictionaryOfDoom.json#L7

BagelOrb:
We should at least group all infill patterns which have * 2 in common.

It should be 
"value": "0 if infill_sparse_density == 0 else (infill_line_width * 100) / infill_sparse_density * (1 if ... else 2 if ... else 3 if ...)

PR #2594, opened 2017-10-12
CL-508 correct status msg when disconnected

sedwards2009:
Details in CL-508. This PR simply ensures that the 'unreachable' printer message is shown instead of the 'action required' one in the case where a printer disappears while waiting for a configuration/material change.


sedwards2009:
Damn it! This branch wants to pull in the other CL-555 related commit too.

PR #2603, opened 2017-10-14
Change renderType for TextInput elements

fieldOfView:
This PR changes the renderType of TextInput elements to be the same as other text elements (Label and native elements such as comboboxes).

Unlike Label elements, which are defined to use Text.NativeRendering, TextInput uses Qt text rendering by default. This can lead to differences in font rendering between input fields and other texts.

Probably fixes https://ultimaker.com/en/community/51334-cura-fonts-garbled-on-macos-high-sierra?page=1#reply-190665

Edit: it also fixes one more occurrence of a fractional pixel offset affecting the placement of text: the Recommended/Custom switch.

### DIFF
--- a/resources/qml/PrintMonitor.qml
+++ b/resources/qml/PrintMonitor.qml
@@ -442,6 +442,7 @@ Column
                 anchors.leftMargin: UM.Theme.getSize("setting_unit_margin").width
                 anchors.right: parent.right
                 anchors.verticalCenter: parent.verticalCenter
+                renderType: Text.NativeRendering
 
                 Component.onCompleted:
                 {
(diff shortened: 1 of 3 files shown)


fieldOfView:
Would be good if this could be cherry-picked to 3.0 still.

Appesteijn:
CURA-4452

Let's see tomorrow if this can be added for 3.0

ChrisTerBeke:
Code changes look good.

ChrisTerBeke:
@fieldOfView we're testing the changes now. If all is good it will be cherry picked into 3.0.

ChrisTerBeke:
Done.

PR #2612, opened 2017-10-16
Fix size of "beta" classifier on splash screen for HiDPI screens

fieldOfView:
This PR changes the size of secondary version "elements" (such as "BETA") to use a pixel size instead of a font size, so they scale correctly with the splash screen. This likely only shows on Windows.


nallath:
CURA-4458

PR #2619, opened 2017-10-18
initial layer flow

BagelOrb:
The initial layer has a different layer height and so the actual rounding of a line conforms to the naive square model CuraEngine uses in a different way.

In order to compensate for that you would want a different flow rate on the initial layer.


Ghostkeeper:
The round line model is all completely different anyway on the first layer if the build plate isn't 100% level or not exactly on the nozzle height. Still, also for that reason, it could be useful to have this.

Ghostkeeper:
Developers, see issue CURA-4506.

Ghostkeeper:
Hmm, I'm wondering what the reason is why you've set the minimum to 5%.

Issue #979, opened 2016-09-08
Print monitor/Print Controller UI discussion


kakaroto:
This is to have a discussion about the print monitor UI and/or whether we want to add control to it, or if we want a separate Print Controller tab instead.
This is to continue the discussion that was started in issue #70, more specifically, started at  https://github.com/Ultimaker/Cura/issues/70#issuecomment-245358634

To summarize: do we want to add controls to the print monitor. If Yes/No, what additional information or controls do we want to add to it. If no, do we want to add a Print Controller tab, similar to the print monitor, but only for manual control of the printer. 
Also, what UI design should be used, anyone has any suggestions/mockups/etc.. Let's discuss it before we implement it.

nallath:
I've already ensured that the architecture supports it. If we use PrinterOutPutDevice any printer that works with monitor tab (right now USB and the octoprint stuff made by @fieldOfView) 

I would like to have a way to add controls. I don't think that adding a third tab (next to monitor & settings) is a good idea. We could add a way to switch between monitoring & controling of sorts?

nickthetait:
How about hiding the controls once a print has started, then revealing them if it is paused or once the print completes?

fieldOfView:
I'll make some mockups after the weekend of what I have had in mind for a while.

tenten8401:
@nickthetait could always just grey them out instead of hiding completely.

kakaroto:
@tenten8401 I think the idea to hide them is to avoid the clutter while printing. Even if you're not printing, if you don't want to control the printer, then having the controls there will still clutter the window. If the user wants to take control, he could click the "control button" which would 'slide in' a controls drawer. That's the way I'd see it I guess.
I'm not sure it's clear so here's a little demo of how I imagined it : http://www.leotheme.com/demo/prestashop14x/?template=leo_gamegear
If you click on the little gear icons on the left hand side of the page, you see some controls slide out. 

That's kind of neat. I don't know if it would look good, or if it would be doable with qml, or if it's an absurd UI design idea, but we're here to discuss ideas, so... :)

kakaroto:
@fieldOfView did you make any progress on the mockups  ? 
I've been working a bit on this, and I've run into a couple of UI design concerns for the print monitor.

[...]

PR #2669, opened 2017-10-24
Jogging

fieldOfView:
This PR adds manual printhead/printbed control ("jogging") and is a partial implementation of the proposals in https://github.com/Ultimaker/Cura/issues/979#issuecomment-254503356

The implementation has been tested with OctoPrint (the master branch of the OctoPrintPlugin has some fixes for this), and should work for USB-connected printers too (because they have the same API). The PR hides the interface for Cura Connect clusters. I don't know if it works for single UM3 printers. If it doesn't, there is a convenient bool to set in the constructor of the NetworkPrinterOutputDevice, like this:
https://github.com/Ultimaker/Cura/blob/a9f2ad7ca348f974949d811b913f01e53d7ffa81/plugins/UM3NetworkPrinting/NetworkClusterPrinterOutputDevice.py#L106

This is what it currently looks like:
image


nallath:
Whoo! Nice work!

nallath:
CURA-4488

Appesteijn:
Is it difficult to add temp and speed control and make this work for UM3s?

fieldOfView:
I don’t have an UM3 to test this with, so I don’t know if it works or not.

The extruder/bed temps and extrusion functionality I would like to integrate with the extruder items and bed preheat above in a followup pr.

nallath:
Should be pretty easy to get it to work with an UM3. Do note that it can't / won't work with any printer that runs connect.

fieldOfView:
>  Do note that it can't / won't work with any printer that runs connect.

I have already added code that hides the controls for a Cura Connect cluster.

If it does not work for UM3 (has anyone tested?), please attach some logs

fieldOfView:
Hmm, NetworkPrinterOutputDevice does not expose an API to send Gcode directly. @nallath, I thought you said that functionality is available? Since the UM3 firmware is not publically available, can you point me in the right direction endpoint-wise?

nallath:
I don't think homing is available. But for the rest; all properties can accept a put which should update it.

fieldOfView:
For now, I have hidden the jogging UI for the UM3(Ex), since it does not have the API to support jogging and I am unable to add an implementation of that API without access to an UM3.

ChrisTerBeke:
Also works over USB (tested with UM2+ on Windows).

ChrisTerBeke:
@fieldOfView with the new cluster API manual control is not possible anyways, so I'd say it's outside the scope of this PR indeed.

mikeq3:
This is a good start. I really need the ability to preheat the nozzle and extrude or retract a specified length. Can we add a button to preheat the nozzle similar to the bed preheat button/field? Also an extrude and retract button with a length field?

Ghostkeeper:
That would be a nice addition.

But don't get your hopes up. If it doesn't work on the Ultimaker 3 there will be little chance nowadays that Ultimaker's employees get to spend their time on it. However it can't work for the Ultimaker 3 since that runs cluster firmware now which has an API that doesn't support it. You can't make the API support it either since the cluster firmware is closed software...

fieldOfView:
This PR is meant as a start. I plan to do follow up PRs, starting with preheating and extrude/retract.

robinmdh:
@fieldOfView The UM3 (stand alone) does support jogging, You can put the coordinates of the head.
We removed the debug gcode execution though.

fieldOfView:
@robinmdh, I know (jogging, but no homing). However, I don't have a UM3 to test with, and I would not want to commit untested code. I also heard that the Cura Connect stuff would interfere (https://github.com/Ultimaker/Cura/pull/2669#issuecomment-341136584), and again since I have no UM3, I have no idea about that.

Rudy-Raw:
I was wondering if consideration can be given to adding a jog control for the E step extrude and retraction?
Thank for any reply.

fieldOfView:
It has, but I gave priority to (pre-)heating extruders. There’s a PR for that in review. Once that merges, I’ll revisit extruding/retracting

Rudy-Raw:
That's great! Will be looking out for it.

PR #2677, opened 2017-10-25
Fix qml error when there is no connected printer

fieldOfView:
This PR fixes the following qml errors on startup:
```
.../Cura/resources/qml/Menus/NozzleMenu.qml:19: TypeError: Cannot read property 'clusterSize' of undefined
.../Cura/resources/qml/Menus/MaterialMenu.qml:19: TypeError: Cannot read property 'clusterSize' of undefined
.../Cura/resources/qml/Menus/MaterialMenu.qml:19: TypeError: Cannot read property 'clusterSize' of undefined
.../Cura/resources/qml/Menus/NozzleMenu.qml:19: TypeError: Cannot read property 'clusterSize' of undefined
```

Also fixes code style

### DIFF
--- a/plugins/UM3NetworkPrinting/UM3InfoComponents.qml
+++ b/plugins/UM3NetworkPrinting/UM3InfoComponents.qml
@@ -119,11 +119,17 @@ Item
             onClicked: manager.loadConfigurationFromPrinter()
 
             function isClusterPrinter() {
-                var clusterSize = Cura.MachineManager.printerOutputDevices[0].clusterSize
-                // This is a non cluster printer or the cluster it is just one printer
-                if (typeof clusterSize == "undefined" || clusterSize == 1)
-                    return false
-                return true
+                if(Cura.MachineManager.printerOutputDevices.length == 0)
+                {
+                    return false;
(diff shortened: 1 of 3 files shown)


nallath:
It seems there are some conflicts. Could you resolve them?

fieldOfView:
Conflicts resolved

plugins/UM3NetworkPrinting/UM3InfoComponents.qml
```diff
@@ -119,11 +119,17 @@ Item
             onClicked: manager.loadConfigurationFromPrinter()
 
             function isClusterPrinter() {
```
ChrisTerBeke: I'm not sure what the isClusterPrinter property is used for in the UI, but even a single printer running the cluster firmware should be treated as being in a cluster (since the API endpoints are different). If it's only UI fluffiness connected to this it's fine I guess.

fieldOfView: I'm not changing the logic here, just fixing the case where there is no connected printer at all.

PR #2684, opened 2017-10-26
Cura 4451 not support container

ChrisTerBeke:
Away with empty quality profiles for "Not supported" combinations! This PR adds a feature that dynamically inserts a "Not supported" empty instance container if there is no file for it.

By adding a quality type "not_supported" it also correctly filters on that type when switching between extruders. After doing that we found out that the list of profiles was different depending on the selected extruder, a bug (confirmed during architecture meeting) that has been in there since at least 2.6. It's now an intersect of the lists of both extruders, so you always see only available qualities.

There is some logic that filters on the active extruder when compiling a list of available materials, the intention is to move that to a single place (material or quality manager?) in the future. For now they're marked as TODO.

_p.s. it also adds a fix to the crash dialog that I needed to debug this feature._


resources/quality/ultimaker2_plus/um2p_pp_0.4_fast.inst.cfg
```diff
@@ -1,71 +0,0 @@
-[general]
```
diegopradogesto: I push back this file, that it seems to be deleted by mistake

cura/CrashHandler.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, exception_type, value, tb):
         self.exception_type = exception_type
         self.value = value
         self.traceback = tb
+        self.dialog = QDialog()
```
diegopradogesto: So the creation of the dialog should be removed from the _createDialog method, isn't it?

cura/Settings/ProfilesModel.py
```diff
@@ -155,14 +164,24 @@ def _recomputeItems(self):
 
         # Now all the containers are set
         for item in containers:
-            profile = container_registry.findContainers(id=item["id"])
+            profile = container_registry.findContainers(id = item["id"])
+
+            # when the profile is not supported
```
diegopradogesto: I suggest to better explain in the comments the difference between this case and the one that was introduced below.

ChrisTerBeke: The difference is a not existing profile (in the container registry) vs an existing profile that has the not supported flag (meta data field). Personally I think the `if not profile:` makes it quite clear. It handles cases where the profile does not exist in the container registry, something that is (or should?) be done for every case where we're fetching containers from the registry.

PR #2685, opened 2017-10-26
Container stack improvements

ChrisTerBeke:
This branch will contain changes to always have an extruder stack even when the printer only has 1 extruder. The idea is that this will simplify using the stack setup in the future and eliminate all the duplicated logic of `if extruder_count > 1` and similar.

Why already create this PR? Because it's a significant change and I'd like to keep track of the changes and allow others to comment on them as they happen.


nallath:
It has conflicts :(

ChrisTerBeke:
Yup, I'll merge that when I'm done making changes today.

ChrisTerBeke:
This is now ready for code review.

PR #2692, opened 2017-10-27
Added reload qml action

nallath:


ChrisTerBeke:
Is this for quick reloading during development?

fieldOfView:
Is it an idea to only enable this functionality when running Cura from source? Ie: when not frozen.

nallath:
It's for quick reloading. We could also enable this only when running it in debug, but i thought the whole shift+f5 to be obscure enough to not be used accidentally.

Ghostkeeper:
Perhaps we should trigger this upon changing the language or the theme?

Ghostkeeper:
Developers, see issue CURA-4539.

ChrisTerBeke:
@Ghostkeeper that would be nice, although I've seen ListModels being populated in Python from language-senstive strings (e.g. the menu structure). Those should be made up from IDs anyways IMO, but that needs to happen first.

nallath:
Not everything from the theme will be updated correctly; The background of the 3D view for instance won't correctly change. Only stuff handled by QML will change. I don't think we should use this as something a user can do just yet. I meant this to be a convenience for developers.

ChrisTerBeke:
I did a quick test and most components are properly reloaded. There are a lot of warnings while reloading however, and some functionality does not recover after the reload (like switching to layer view). Any thoughts about this?

alekseisasin:
It will definitely help us to simplify qml development process. But in release it might make errors, like after refreshing few times try to open any model, you will see that dialog box is opened few times. I suggest enable shortcut: "Shift+F5" only for master build.

diegopradogesto:
I agree that this can be used for debugging purposes but since there are some weird behaviors I would not let it be in a release. Once I refresh, I can not change the view anymore. Also after several reloads, the context menu and the modal windows become replicated.
image

PR #2696, opened 2017-10-28
New option for better sticking

mikko70:
I use glass and i use method for item better sticking in table.
I put down speed (about 30-60%) and drive bit near table.
I hope there is option for Build Plate adhesion. " Try stick better"  selection that drop down speed 50% and drop down about 0,05mm nozzle, 
But only skirt or brim, no first layer.


nallath:
Something went pretty wrong with your pull request. You're trying to merge master into 3.0 branch...

PR #2697, opened 2017-10-28
Feature add user agreement

Appesteijn:


plugins/UserAgreementPlugin/UserAgreement.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2015 Ultimaker B.V.
```
Ghostkeeper: What year is it?

(Goes for all of these files.)

plugins/UserAgreementPlugin/UserAgreement.py
```diff
@@ -0,0 +1,54 @@
...
+
+from PyQt5.QtQml import QQmlComponent, QQmlContext
+from PyQt5.QtCore import QUrl, QObject, pyqtSlot
+
+import os.path
+
+class UserAgreement(Extension, QObject):
+    def __init__(self, parent = None):
+        QObject.__init__(self, parent)
+        Extension.__init__(self)
```
Ghostkeeper: It's better to simply use `super()`. For some examples on how to do that with multiple inheritance, see [StackOverflow](https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance).

plugins/UserAgreementPlugin/UserAgreement.py
```diff
@@ -0,0 +1,54 @@
...
+    def showUserAgreement(self):
+        if not self._user_agreement_window:
+            self.createUserAgreementWindow()
+
+        self._user_agreement_window.show()
+
+    @pyqtSlot(bool)
+    def didAgree(self, userChoice):
+        if userChoice:
+            Logger.log("i", 'User agreed to the user agreement')
```
Ghostkeeper: Code style: Double quotes for strings.

plugins/UserAgreementPlugin/UserAgreement.qml
```diff
@@ -0,0 +1,64 @@
...
+    TextArea
+    {
+        anchors.top: parent.top
+        width: parent.width
+        anchors.bottom: buttonRow.top
+        text: ' <center><h3>DISCLAIMER BY ULTIMAKER</h3></center>
+                <p>PLEASE READ THIS DISCLAIMER CAREFULLY.</p>
+                <p>EXCEPT WHEN OTHERWISE STATED IN WRITING, ULTIMAKER PROVIDES ANY ULTIMAKER SOFTWARE OR THIRD PARTY SOFTWARE “AS IS” WITHOUT WARRANTY OF ANY KIND. THE ENTIRE RISK AS TO THE QUALITY AND PERFORMANCE OF ULTIMAKER SOFTWARE IS WITH YOU.</p>
+                <p>UNLESS REQUIRED BY APPLICABLE LAW OR AGREED TO IN WRITING, IN NO EVENT WILL ULTIMAKER BE LIABLE TO YOU FOR DAMAGES, INCLUDING ANY GENERAL, SPECIAL, INCIDENTAL OR CONSEQUENTIAL DAMAGES ARISING OUT OF THE USE OR INABILITY TO USE ANY ULTIMAKER SOFTWARE OR THIRD PARTY SOFTWARE.</p>
+                '
```
Ghostkeeper: I think it's better to read this from a text file than to hard-code it in here.

Does it need to be translated?

plugins/UserAgreementPlugin/UserAgreement.qml
```diff
@@ -0,0 +1,64 @@
...
+    {
+        id: buttonRow
+        anchors.bottom: parent.bottom
+        width: parent.width
+        anchors.bottomMargin: UM.Theme.getSize("default_margin").height
+
+        Button
+        {
+            anchors.right: parent.right
+            text: "I understand and agree"
```
Ghostkeeper: Make this translatable.

plugins/UserAgreementPlugin/UserAgreement.qml
```diff
@@ -0,0 +1,64 @@
...
+            onClicked: {
+                manager.didAgree(true)
+            }
+        }
+
+        Button
+        {
+            anchors.left: parent.left
+            text: "I don't agree"
```
Ghostkeeper: Make this translatable.

plugins/UserAgreementPlugin/UserAgreement.qml
```diff
@@ -0,0 +1,64 @@
...
+UM.Dialog
+{
+    id: baseDialog
+    minimumWidth: Math.floor(UM.Theme.getSize("modal_window_minimum").width * 0.75)
+    minimumHeight: Math.floor(UM.Theme.getSize("modal_window_minimum").height * 0.5)
+    width: minimumWidth
+    height: minimumHeight
+    title: "User Agreement"
+
+    TextArea
```
Ghostkeeper: Is there a way to prevent the user from modifying the text in the screen?

Appesteijn: 'readOnly: true;' is the only thing I can think of, but that is already in there

plugins/UserAgreementPlugin/UserAgreement.py
Appesteijn: done

plugins/UserAgreementPlugin/UserAgreement.py
Appesteijn: done

plugins/UserAgreementPlugin/UserAgreement.py
Appesteijn: done

plugins/UserAgreementPlugin/UserAgreement.qml
Appesteijn: done

plugins/UserAgreementPlugin/UserAgreement.qml
Appesteijn: done

plugins/UserAgreementPlugin/UserAgreement.qml
Appesteijn: Not sure if we can translate this legalese, let discuss tomorrow

ChrisTerBeke:
OMG, @Appesteijn is now a developer! :P

Appesteijn:
I will start worrying when you guys become product owners 😄

fieldOfView:
I own plenty products. My wife will say I own way too many products...

plugins/UserAgreementPlugin/UserAgreement.py
```diff
@@ -36,11 +35,11 @@ def showUserAgreement(self):
     @pyqtSlot(bool)
     def didAgree(self, userChoice):
         if userChoice:
-            Logger.log("i", 'User agreed to the user agreement')
+            Logger.log('i', 'User agreed to the user agreement')
             Preferences.getInstance().setValue("general/accepted_user_agreement", True)
             self._user_agreement_window.hide()
         else:
-            Logger.log("i", 'User did NOT agree to the user agreement')
+            Logger.log('i', 'User did NOT agree to the user agreement')
```
Ghostkeeper: Hehe, you changed this the wrong way around. We use double quotes for strings.

fieldOfView:
So, on a clean install, how many dialogs and messages does the user get when launching Cura for the first time?
* User agreement dialog
* Changelog dialog
* Add printer dialog
* (optional) network printing dialog or machine settings dialog
* Slice info message

All of these are shown as separate dialogs (unless I am mistaken). It would be nice if these could be unified into a single "onboarding wizard".

plugins/UserAgreementPlugin/UserAgreement.qml
fieldOfView: Does it need to be SHOUTED?

Ghostkeeper: Some parts of an end-user license agreement indeed has to be SHOUTED (or does it?) A good breakdown here: https://www.quora.com/Why-do-some-parts-of-legal-documents-appear-in-all-caps-e-g-EULAs-have-portions-written-in-caps-lock-Does-this-mean-something-in-a-legal-sense

oliv3r: Why not only have this annoying pop up on the cura download page? or will we show the bubble each time the terms change (even if this is not the first time?)

oliv3r: also, how valid are the EULA's in court?

fieldOfView: only valid when they're SHOUTED, apparently

Issue #2662, opened 2017-10-23
Cost of print behind print button


AnaGallego86:
Hello!

I just updated to Cura 3.0.

I decided to assign price for the different filaments to Cura gives me the price of the print.

If I decide to use PVA as support the price total estimation becomes to large, so that the liner appears right behind the PRINT OVER NETWORK button. I can read the total cost but you kind of have to guess a little bit.

![image](https://user-images.githubusercontent.com/33025296/31885919-15fa7fd0-b7ea-11e7-9739-5cb96aacce06.png)

Is it just me?

ChrisTerBeke:
Nope, I've seen that as well, but not really using the print cost feature though. Overall this part of the UI needs some re-thinking (e.g. why is the label "Print time" below the actual print time?

ChrisTerBeke:
Fixed in above referenced pull requests, due for next release.

Issue #2643, opened 2017-10-20
3.0 German Translation - Button text overlaps time information


nicolinux:
A german user reported this issue (Windows 7 64bit).
![overlap](https://user-images.githubusercontent.com/1069286/31828577-f2e04484-b5ba-11e7-8fd1-a84bf4a02706.png)

hannover345:
![2017-10-20_21h55_59](https://user-images.githubusercontent.com/32967836/31839749-bf14100e-b5e2-11e7-9d8c-ab62ca298209.png)
![2017-10-20_21h55_15](https://user-images.githubusercontent.com/32967836/31839762-c4fc0f12-b5e2-11e7-8f0f-0694fd615b1c.png)

Vandrasc:
Devs:4478

Ghostkeeper:
Is there a shorter German text we could use than "Speichern auf Wechseldatenträger"? Or maybe we should elide that as well just to be sure...

fieldOfView:
See #2700

hannover345:
I can suggest for Speichern auf Wechseldatenträger
Speichern auf SD/USB
or
Speichern auf SD

On my last Post, second image
"Open Octoprint..."  "Pausieren"  "Drucken Abbrechen"
may be you can change with    

"Octoprint..." "Pausieren"  "Druck Abrechen"

fieldOfView:
Problem is, we don't know if it is an sd-card. Could be a Compactflash card, or a floppy disk.

I am considering changing the "Open OctoPrint..." button to a graphical button (with a monochrome OctoPrint logo) so it takes up less space.

nicolinux:
Couldn't we just use "Speichern"? That's pretty general and could by anything compactflash card, floppy, hard disk...

ChrisTerBeke:
@nicolinux we use different button texts for outputting to disk or to removable drive (to indicate you've selected the correct output method).

Ghostkeeper:
[Wikipedia](https://de.wikipedia.org/wiki/Wechseldatentr%C3%A4ger) suggest "Wechselmedium" as alternative. Is that better, German-speaking folk?

ianpaschal:
This hasn't been touched in 3+ months and is marked as Fixed in JIRA so I'm marking it the same here and closing it.

Ghostkeeper:
Corrected the text in question to `Speichern auf Datenträger`, which was suggested by a colleague. It's a bit better now.

PR #2700, opened 2017-10-29
Cleanup save/monitor button area

fieldOfView:
This PR rearranges the time & cost specs in the save button area, and puts some more information in the tooltips for these elements

image

* Time specification has better formatting and now adds percentages
* Cost item has elide, shows only total cost and weight
* Cost item has a tooltip that shows further specification

This PR fixes https://github.com/Ultimaker/Cura/issues/2662 and https://github.com/Ultimaker/Cura/issues/2643 (CURA-4478)

Currently a WIP, because there are still a couple of QML binding loops that need fixing.

### DIFF
--- a/cura/PrintInformation.py
+++ b/cura/PrintInformation.py
@@ -56,6 +56,7 @@ class PrintInformation(QObject):
         self._material_lengths = []
         self._material_weights = []
         self._material_costs = []
+        self._material_names = []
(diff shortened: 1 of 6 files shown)


fieldOfView:
Ready for review

nallath:
CURA-4509

PR #2703, opened 2017-10-30
Stop SettingTextField from setting values when getting/losing the focus

### LINKED ISSUE #2694: Build plate temperatures thinks a change was made
Cura 3.0.3
Windows 10

If you click into the build plate temperatures field then click out with out making a change, Cura thinks a change was made as adds the circular reset arrow next to the field.

fieldOfView:
This PR stops SettingTextField from setting a new "literal" value when just tabbing in/out of focus by checking if the text has changed between receiving the focus and editingFinished.

Fixes #2694

### DIFF
--- a/resources/qml/Settings/SettingTextField.qml
+++ b/resources/qml/Settings/SettingTextField.qml
@@ -11,6 +11,14 @@ SettingItem
     id: base
     property var focusItem: input
 
+    property string textBeforeEdit
+    property bool textHasChanged
+    onFocusReceived:
+    {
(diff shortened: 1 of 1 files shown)


nallath:
CURA-4514

alekseisasin:
You are right, the "onTextChanged" is available in  QT 5.9, which requires "import QtQuick 2.9". For now I am not sure do we want to move from QT 5.8.2 to 5.9 version, because it requires more testing. I will apply onTextChanged if we decide to change the version.

Issue #2715, opened 2017-11-03
Cura should depend on Python 3, not 3.5.0


onitake:
Cura and the test suite have a hard dependency on Python 3.5.0, which is updated occasionally.
This causes problems when multiple Python interpreters are installed on a system, and especially when a system has a newer Python version available.

CMakeLists.txt and cmake/CuraTests.cmake should depend on Python 3 instead, so the main (or latest) Python interpreter is picked up by cmake.

onitake:
(Please disregard PR 2716, I accidentally pushed the wrong branch)

nallath:
Because of cross platform issues, we have to depend on at least python 3.5. @awhiemstra knows the specifics. From what I remember it had to do with MSVC issues.

onitake:
Ok, makes sense.
But does that have to be reflected in the in the Cmakefiles?
Wouldn't a simple notice in the README suffice?

awhiemstra:
find_package(PythonInterp 3.5.0) should find any Python equal to or greater than 3.5.0. At least that is what happens on my system, where it mentions
 
> -- Found PythonInterp: /bin/python3 (found suitable version "3.6.3", minimum required is "3.5.0")

So the version here indicates a minimum, any higher version should also be fine. If detecting the Python interpreter fails, that may point to a different issue.

onitake:
That's good, but it's still a problem if multiple Python versions are installed.

For example, if a system has Python 3.5 and 3.6, cmake will only find and run tests against 3.5. This will fail when Arcus or Savitar (which are bound to a single Python version) have been built for 3.6 instead.

Ghostkeeper:
That's not so much a problem with using a specific Python version but more with mixing up different Python interpreters.

I'd think that CMake's findPython script would find the same Python version every time though. Maybe there is a difference in the way that we call the find script? 3.5.0 instead of 3.5 or something?

onitake:
I understand your reasoning with respect to "find at least Python 3.5", but passing an explicit version to `find_package` seems like the wrong solution. While having multiple Python 3 interpreters installed is not common, it is not impossible either.

According to [the documentation](https://cmake.org/cmake/help/v3.0/module/FindPythonInterp.html), there is a way to pass a list of preferred Python versions, by setting the variable `Python_ADDITIONAL_VERSIONS` before calling `find_package`. That is not exactly "find the latest Python version that is at least 3.5", but would be close enough. The downside is that it requires manual updates to that list of Python versions.

[...]

PR #2717, opened 2017-11-03
Cmake script patch so they depend on Python 3 instead of 3.5

onitake:
This is a fix for #2715 .


Ghostkeeper:
We require at least 3.5 at least for the `typing` module and the better Pickle, so just 3 isn't going to be enough. 3.6 is allowed though and will be found on your system by our current CMake script, if installed.

PR #2743, opened 2017-11-06
Headless cura

LipuFei:


fieldOfView:
What does/will this do exactly? CuraEngine is basically headless Cura. 

Not starting the QML engine means a couple of plugins will not get initialised.

ChrisTerBeke:
It's about being able to use the settings container stack via command line / DBUS. You can use the engine yes, but you'll still need to manually input all the settings depending on machine/material/quality. This functionality will be used in some projects we're currently working on.

LipuFei:
@fieldOfView This headless cura always you to do high level operations, such as creating and changing your own machines and settings. This will be very convenient for automated testing and when you want to script certain steps like the following:

1. Add UM3
2. Select "Generic ABS" for Extruder 1
3. Select "Extra Fine" profile
4. Open model "20x20x20.stl"
5. Slice and save Gcode to "20x20x20.gcode`

Ghostkeeper:
Eventually we'll want to have some sort of command line interface to select profiles, but this is a good first step. You can get in via DBus.

Issue #2745, opened 2017-11-07
Fedora 26, appimage 3.0.4 


knuckleheadTech:
Cura from the repository works fine but I want to use 3.0.4 with the CR 10 settings. Running the appimage from a different computer (Xubuntu 14.04) works fine. But from on my Fedora 26 it starts to load then crashes as soon as the user agreement starts to come up. I never actually see the user agreement I just know from watching the other system run the appimage that it's the user agreement that crashes. Don't see any kind of warning, just kills off. 

Running from the terminal gives this output:

Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 72: non-double matrix element
Fontconfig error: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 72: non-double matrix element
Fontconfig warning: "/etc/fonts/conf.d/10-scale-bitmap-fonts.conf", line 80: saw unknown, expected number
Fontconfig warning: "/etc/fonts/conf.d/65-0-lohit-bengali.conf", line 32: unknown element "langset"
Fontconfig warning: "/etc/fonts/conf.d/69-gnu-free-sans.conf", line 24: unknown element "langset"
Fontconfig warning: "/etc/fonts/conf.d/69-gnu-free-serif.conf", line 24: unknown element "langset"
QObject::connect: Parentheses expected, signal MainWindow::
QObject::connect: Parentheses expected, signal MainWindow::
QObject::connect: Parentheses expected, signal MainWindow::
QObject::connect: Parentheses expected, signal MainWindow::
file:///tmp/.mount_Uzb2Us/usr/bin/resources/themes/cura-light/styles.qml:197:17: QML Item: Binding loop detected for property "width"
qml: TableViewSelection: index out of range
qml: TableViewSelection: index out of range
file:///tmp/.mount_Uzb2Us/usr/bin/plugins/plugins/UM3NetworkPrinting/DiscoverUM3Action.qml:282: TypeError: Cannot read property 'clusterSize' of null
./bin//cura.sh: line 11: 11489 Segmentation fault      (core dumped) cura $@

Appesteijn:
I made a pull request to fix this: https://github.com/Ultimaker/Cura/pull/2751

Appesteijn:
I made a pull request to fix this: https://github.com/Ultimaker/Cura/pull/2751

nallath:
This won't actually fix the crash. The typerror is just a warning.

[...]

PR #2751, opened 2017-11-08
Also check if base.selectedPrinter is null

Appesteijn:
Fix for https://github.com/Ultimaker/Cura/issues/2745

Not sure if 'undefined' is different from 'null' in qml

### DIFF
--- a/plugins/UM3NetworkPrinting/DiscoverUM3Action.qml
+++ b/plugins/UM3NetworkPrinting/DiscoverUM3Action.qml
@@ -276,12 +276,8 @@ Cura.MachineAction
                     width: parent.width
                     wrapMode: Text.WordWrap
                     text:{
-                        if (base.selectedPrinter == undefined || base.selectedPrinter == null)
-                        {
-                            return "";
-                        }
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
undefined means that a variable hasn't been initialised yet, null means it has and it explicitly has a null (empty) value. However comparing a value with null or undefined in a non-strict way (== rather than ===) both result in false, so these extra checks are not needed. Actually it would be better to use the short-hand `if (!base.selectedPrinter)` so we cover all cases, including a `false` value.

fieldOfView:
+1 for the shorthand version.
Also, first checking if base.selectedPrinter doesn't exists (and returning "" when so) and then checking if it does exist in the next statement is a bit redundant. 

I would be much surprised if this actually fixes the segfault in #2745. It might fix an innocent qml warning, but it does not fix the bigger issue.

LipuFei:
Agree with @fieldOfView that this has little chance fixing that segfault in #2745.

LipuFei:
@fieldOfView Thanks! Fixed in 78456d5b. I guess I need some coffee...

PR #2758, opened 2017-11-09
Improved mesh type UX (Per Model Settings)

fieldOfView:
This PR makes support_mesh, anti_overhang_mesh and cutting_mesh first class citizens of the Per Model Settings panel, helping their usability.

* A "Mesh Type" dropdown is added to Per Model Settings
* Non-printing meshtypes (anti_overhang_mesh and cutting_mesh) don't get a convex hull, so they can be moved around freely without pushing other models away
* Non-printing meshtypes are rendered transparently
* support_mesh models are rendered in the material color of the support_extruder
* Only one of the [support_mesh, anti_overhang_mesh and cutting_mesh] can be selected at a time
* When support_mesh or anti_overhang_mesh are selected, other per model settings are hidden (because they are not used anyway)

infill_mesh is not used in this PR at the moment; I have not yet found any functionality that can not be achieved with cutting_mesh.

This PR hopes to make the modifier meshes more discoverable and easier to use. They are a very powerful feature, worthy of more attention.

meshtypes


fieldOfView:
Pinging @BagelOrb for feedback on my assessment of infill_mesh vs cutting_mesh. If you set cutting_mesh, but don;t change the extruder, you can use the feature the same way as infill_mesh, right?

BagelOrb:
You can set the extruder irrespective of setting cutting mesh or infill
mesh. That's unrelated.

In short it's just that a cutting mesh also affects the shell.

fieldOfView:
I have added infill_mesh as an option to the dropdown

ChrisTerBeke:
This is freaking awesome.

fieldOfView:
Thanks. I'm hoping it brings out some freaking awesome functionality that has been hidden away in CuraEngine for a long time.

fieldOfView:
Infill mesh and cutting mesh are now rendered with the color associated with the selected extruder (still semi-transparent). Moar awesome.

BagelOrb:
image

BagelOrb:
Do cutting meshes and infill meshes now also have disabled the checking of whether the mesh falls inside of the build volume?? They don't need that..

fieldOfView:
Yes, they don't.

nickthetait:
Me when I see a useful feature but have no responsibility to maintain it 
slam-that-merge-button

BagelOrb:
Meh maintenance is low compared to added value for this one ;)

ChrisTerBeke:
We're currently in a research sprint, but I think this will get merged somewhere next week.

ChrisTerBeke:
@fieldOfView the extruder stack rewrite was merged to master and unfortunately introduced some conflicts with your branch. Do you mind having a look at those?

fieldOfView:
Done.

fieldOfView:
*subtlebump*

This PR has strings, so it would need to land before the beta I guess.

awhiemstra:
Agreed with the others that this is awesome.

Now we just need a panel with some basic shapes and we will have almost complete custom support ;)

BagelOrb:
Can't wait for this to be merged!!!

nallath:
I think the string freeze already happened :( I'm not on the Cura project anymore, so it's up to one of the other devs to actually merge it.

fieldOfView:
That's rather disappointing. Then again, the Cura team itself is making string freeze violations (https://github.com/Ultimaker/Uranium/pull/309/files#diff-8d9f25e632b89e5fc4d0a5380c049b64R173), so there might still be hope.

ChrisTerBeke:
@fieldOfView those two violations were pushed from 'higher up', we also weren't very happy about it. This PR is a bit different though since there's a few completely new strings. I would have loved for this to be merged already, but we had to prioritise and there were still a few blocking bugs for 3.1 beta that we needed to get out of the way :(

fieldOfView:
Well, there's always next year... :cry:

nallath:
sadcat

Appesteijn:
Chris just pointed me to this PR, awesome! This is 80% of the use case for manual support.
I only think it is still too hidden, you have to select a model, open custom settings, go to per model settings, add mesh type. In our usage data we see very few people use per model settings/
Is it possible to give this an easier to find place in the UI? For example as a tool and add a 1x1x1cm mesh that we load automatically when a user clicks on the tool-button?

ChrisTerBeke:
We could indeed make a special "support structure editing" tool and add a button in there to load a simple mesh that can be scaled/rotated to remove supports in certain areas (which is what most people ask for when it comes to "manual" supports). The tool can then also contain the object type selector that was extracted in this PR already.

BagelOrb:
Having data that people don't use per object settings that much doesn't mean the UI is unclear. It might very well be the case that people start using the per-object settigns a lot more now that some of it's main use cases are in a dropdown which is immediately visible.

There have been loads of talk about a custom support tool, but that's not relevant to this thread. Talk to Jack.

fieldOfView:
I think this PR and having actual manual supports are two different things. The current proposal can be seen as a stepping stone towards manual supports, but actually modelling supports in Cura (the frontend) is quite a step beyond. 

One step at a time, please.

Appesteijn:
You are missing my point. If we could extract this function into a separate 'tool' that also loads a cube when you click on it, people could easily remove auto generated support.
When I ask people why they want 'Manual Support', they usually say: 'So that I can remove parts of the auto generated support'.

BagelOrb:
I think @fieldOfView  and me are just saying that this PR is just one step; don't try to make it more than it is, because this one step is already rather big.

This step does trigger the imagination, but let's not make too hasty conclusions on what such a separate manual support tool should look like.

Ghostkeeper:
Painting 1mm voxel cubes is probably too much for Cura to handle. Scene nodes aren't optimized for that sort of usage. Cura starts to cough quite a lot when loading >30 models already.

I thought a lot about custom support because I planned to research it in the research sprint. The best solution is probably a tool to allow texture painting in combination with BagelOrb's texture processing branch. It's also the hardest solution because texture painting is notoriously difficult to implement. The second-best solution I could think of was subdividing the model and then vertex painting, which might have been feasible in 2 weeks with a lot of luck.

ChrisTerBeke:
@Appesteijn merge this as a first step or not?

Appesteijn:
If it is too difficult to do, then let's start with small steps. So yes merge it please!

ChrisTerBeke:
Played around with the available options and they all seem to work as expected. I think having some GIFs for users to see what you can actually do would be really helpful. We'll suggest that to marketing.

For the future we'll see how we can improve this. There are some ideas to be able to add a simple cube mesh so you don't have to load a 2nd model for simple support removal or modifiers. We can also see how this could be integrated with @jackha's model file list (show the mesh type, add the cube). We'll create a new internal ticket for this and let UX brainstorm a bit. @fieldOfView we appreciate your thoughts on this as well!

BagelOrb:
Yay!

nallath:
Well, loading 30 cubes shouldn't be a problem, as long as they don't show convex hull nodes. Those are what loading a bunch of small models slow-ish (that and collision checking is now N^2). But helper nodes don't suffer from either problem.

ChrisTerBeke:
By default we could set the mesh type to something else than printable, and maybe give them a different look to indicate they're helper nodes.

PR #2780, opened 2017-11-14
Cartesio profiles

maukcc:
updated prime tower
update PC profiles


maukcc:
changed the prime tower value

ChrisTerBeke:
There is a conflict between your branch and Cura master. Please fix that so we can merge.

Ghostkeeper:
Your printer still loads fine, just now with new settings. Updated for version 3.1.

PR #2782, opened 2017-11-14
New definition was again faulty

velbn:
Somehow moving the default value of the material diameter fixed the definition


nallath:
There are conflicts in your branch. Could you resolve them?

velbn:
Ok should be fixed now

fieldOfView:
As far as I can tell, this PR can be closed because the fix was already done in https://github.com/Ultimaker/Cura/commit/637d1a1da30ec312efcdd69696ed0b41910b18fe

ChrisTerBeke:
Looks the same indeed.

velbn:
Well i fixed it locally (removed the comma) but still it didn't worked. 
Somehow if i move the setting, it is fixed =/

Ghostkeeper:
The Vertex Delta K8800 seems to load fine on my side.

That commit https://github.com/Ultimaker/Cura/commit/637d1a1da30ec312efcdd69696ed0b41910b18fe seems like it actually fixed something, but this PR doesn't. I think we'll close this PR for that reason.

Maybe it was a caching issue? Perhaps you need to delete your cache? On Windows it's over at `C:\Users\<You>\AppData\Local\cura`. On Linux at `~/.cache/cura`. Not sure about MacOS.

PR #2783, opened 2017-11-14
CURA-4525 multi build plate WIP

jackha:
See: https://github.com/Ultimaker/Cura/pull/3029


nickthetait:
Care to show a screenshot in the current state?

jackha:
Thanks for the interest :-) Here's a pic. In the "object menu" you can also select files and use the context menu.

cura-4525-multi-build-plate

nallath:
Why make it a tool? That seems a bit weird to me (also; the tool menu is getting a bit cluttered.)

jackha:
The "final" UI is yet to be designed. Currently it indeed clutters with the tools.
A possible idea is to "flip the settings side around" and you get a nice and big space.

nallath:
Or a reserved bit of space in the main UI. I think this is interesting enough to take into acount. Having multiple workspaces will have such an impact on your workflow that it's almost impossible to do while fully remaining inside the scope of a plugin.

PR #2787, opened 2017-11-15
Prepare for Simulation View / Add line thickness and feedrates to the sliced models

diegopradogesto:
These changes prepare Cura for several things:
- Adapts the interface to include the Simulation View.
- Adapts the backend to receive data related with the thickness and feedrate from CuraEngine.
- Improves the GCodeReader by calculating the line widths of each segment

This PR needs to be applied together with the PR654 in CuraEngine: https://github.com/Ultimaker/CuraEngine/pull/654


awhiemstra:
Hmm, is it really a good idea to replace the Layer view instead of separating the two? It seems to me you are now busy overloading the layer view with features. Additionally, having the simulation view separate would allow for it to be a full simulation of the entire print, not just the current layer, which would mean you could remove the layer selection slider for example.

ChrisTerBeke:
@awhiemstra this decision was made in the Cura team after some discussion. If you have ideas about it, maybe ask @Appesteijn?

Ghostkeeper:
The decision was mostly made on the fact that the simulation view has all the functionality that the layer view has included by nature. There is no reason to use the layer view any more.

PR #2797, opened 2017-11-17
Add setting to determine whether to use firmware retraction

### LINKED ISSUE #1097: Feature Request: Explicit options for Firmware Retraction and Vol. Extrusion
Feature Request: Explicit options for Firmware Retraction and Vol. Extrusion

Not much else to say. Can we please have an explicit option to enable FR and Vol Extrusion? This would be the final nail in the coffin for me using Slic3r.

Ghostkeeper:
This allows printers to determine this for themselves. By default it's based on the g-code flavour in the same way as it is now implemented in the engine.

See also https://github.com/Ultimaker/CuraEngine/pull/656

Fixes #1097.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -318,6 +318,17 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_firmware_retract":
+                {
+                    "label": "Firmware Retraction",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-4579.

dintid:
This would be so awesome! 💯  :) Also sure it would make it easier for developers in the long 👍

andjan1964:
Yes. That would be great !!!!!! 😊👍

PR #2801, opened 2017-11-17
Depend on Python 3.5, but accept later versions

onitake:
Third try at a fix for #2715 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,10 @@ if(NOT ${URANIUM_SCRIPTS_DIR} STREQUAL "")
     CREATE_TRANSLATION_TARGETS()
 endif()
 
-find_package(PythonInterp 3.5.0 REQUIRED)
+find_package(PythonInterp 3 REQUIRED)
+if (PYTHON_VERSION_MAJOR LESS 3 OR PYTHON_VERSION_MAJOR EQUAL 3 AND PYTHON_VERSION_MINOR LESS 5)
+    message(FATAL_ERROR "At least Python 3.5 is required")
+endif()
 
 install(DIRECTORY resources
         DESTINATION ${CMAKE_INSTALL_DATADIR}/cura)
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
If I understood it correctly, the original find script already accepts versions later than 3.5. It's right here in the find module: https://github.com/Kitware/CMake/blob/53c6da8b9e549b07996b446981b4fce9577904a5/Modules/FindPythonInterp.cmake#L54
It works on Ubuntu 17.10 where I have only 3.6 installed, at least. It should be using the newest version that you have installed if you have multiple. So I don't see what this is going to change other than to customize the error message.

ChrisTerBeke:
Works fine for me on Windows as well (using Python 3.6).

onitake:
As I mentioned on #2715 : cmake fails to detect Python 3.6 if 3.5 is also installed. It is not possible to do this with `find_package` directly.
 
This PR will always detect the latest version, but makes sure it is at least 3.5.

Ghostkeeper:
I still can't reproduce the bug that this was supposed to fix :confused:
I have a Windows computer that has both Python 3.5 and Python 3.6 installed, and 3.6 is chosen by CuraEngine. Maybe the CMake find module doesn't work properly on Linux?

onitake:
Well... as stated in #2715, the CMake Python module does indeed behave differently on Windows and Linux. The actual detection code is different and a bit hard to understand. Fact is, if both Python 3.5 and a newer version are installed on a Linux system, FindPythonInterp will only detect 3.5, and fail to build when Cura's dependencies are only installed for the newer Python interpreter.

It would be best if there was a way of specifying a minimum version (which the version argument is not!), with consistent behaviour on all platforms.

But I agree that this is not a major issue for most users and that you don't want to accept a complicated workaround for something that is not a bug in Cura. If you think it's not worth it, I'll close the PR. The bug was already closed a while ago by @ChrisTerBeke .

onitake:
By the way, is your integration test broken?
If I click on the link, it reports full success, but the build is marked as failed on Github.

ChrisTerBeke:
Closing because it's not an issue in our build systems or for end users. Changes to this could cause unknown issues though.

PR #2732, opened 2017-11-05
Add a button to collapse the layerview legend

fieldOfView:
This PR adds a button to the layerview legend to collapse it, mitigating complaints that the layerview legends takes up too much space.

PR #2820, opened 2017-11-20
Simulation view UX tweaks

fieldOfView:
This PR adds the collapse button proposed in #2732 to the Simulation View. Compared to that PR, I have changed the collapsed legend to include the Color Scheme dropdown, so that functionality is retained even when collapsed.

The PR also cleans up some margins, and moves the play button up so it stands out a little less.

image
(new collapsed state)


ChrisTerBeke:
Awesome! @diegopradogesto check this :)

diegopradogesto:
Nice improvement! Very useful for people that never use the checkboxes in the legend and already know what the colours mean. Indeed the users with small screens will like this.
I also like the position of the button, but I don't know if the horizontal slider become very short.

fieldOfView:
Give it a go, confer with the IX theme, and let me know. If the consensus is the button is better left at the bottom, I'll move it back.

Personally I don't think it makes too much difference in the precision of the slider.

ChrisTerBeke:
Devs: CURA-4597

fieldOfView:
On the bright side: CURA-4547 can be closed

smartavionics:
I think the line slider needs left/right move buttons because you need to be able to move forward/backward with precision. Also, how does it cope with layers that contain very many lines (more lines than the slider precision)?

fieldOfView:
> you need to be able to move forward/backward with precision
> how does it cope with layers that contain very many lines

I don't think either issue can be solved by making the slider 36 pixels wider. How about step-by-step scrolling with the cursor keys?

smartavionics:
> Update: step-by-step scrolling is already implemented by @diegopradogesto. Click the slider nub once to activate it (makes it blue), and press cursor left/right.

OK, so that's good. What it really is missing then is the readout of the details for the current line. How craftware do it is nice (IMHO) because it shows you all the info you need right in front of your eyes. The problem with putting that info in another region of the screen is that you will need to move your visual focus from the info to the current line. Still if that's how it will be done, that would be better than no info at all.

screenshot_2017-11-21_17-10-07

diegopradogesto:
This is something that shouldn't be difficult because the information is there. I will try to do it in my free time.

fieldOfView:
Too bad this was not cherry-picked to 3.1...

ChrisTerBeke:
Might still happen for 3.1 stable? @Appesteijn

Appesteijn:
Ok, fine by me 😄

diegopradogesto:
DONE!

PR #2827, opened 2017-11-22
CuraApplication/UM.MeshBuilder: Removing _reloadMeshFinished

thopiekar:
If this commit gets merged into Uranium, then _reloadMeshFinished is not needed anymore.
https://github.com/Ultimaker/Uranium/commit/31bff5b9b9797fb5654ea8ed74a51fff42600e1d
The function has been moved, so reloading a mesh after a file has been modified is possible.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1151,7 +1151,7 @@ class CuraApplication(QtApplication):
             if file_name:
                 job = ReadMeshJob(file_name)
                 job._node = node
-                job.finished.connect(self._reloadMeshFinished)
+                job.finished.connect(node._reloadJobFinished)
                 job.start()
             else:
                 Logger.log("w", "Unable to reload data because we don't have a filename.")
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-4637.

thopiekar:
Related to https://github.com/Ultimaker/Uranium/pull/307

alekseisasin:
This feature might be useful, but not for all users. I think it should be customized some how and by default I would keep it disabled, otherwise users will complain about why model is updated. 
There is definitely other use cases if we will keep as default behavior.

alekseisasin:
we will close this pull request, because we do not have any business case for this feature.

Ghostkeeper:
The Jira ticket was closed because the misconception arose that this would send e-mails whenever something changed on disk. This is obviously not the case and I don't know where that came from. So I've cloned it and now it's on CURA-4912.

thopiekar:
Thanks @Ghostkeeper

cura/CuraApplication.py
```diff
@@ -1151,7 +1151,7 @@ def reloadAll(self):
             if file_name:
                 job = ReadMeshJob(file_name)
                 job._node = node
-                job.finished.connect(self._reloadMeshFinished)
+                job.finished.connect(node._reloadJobFinished)
```
Ghostkeeper: `SceneNode` has no function `_reloadJobFinished`. Pressing F5 now crashes Cura.

thopiekar: There was a PR months before. There I introduced or just renamed this signal.

Issue #2837, opened 2017-11-24
[3.1] Python Error after upgrade


Thisismydigitalself:
Win10/64, DX 12, Nvidia TI750, drv: 388.13.

![ice_screenshot_20171124-142536](https://user-images.githubusercontent.com/30949968/33210611-9f875604-d123-11e7-8c00-b9c109a6d236.png)

Thisismydigitalself:
[3.1B CrashLog.txt](https://github.com/Ultimaker/Cura/files/1501709/3.1B.CrashLog.txt)

alekseisasin:
Probably the problem is not related to your OpenGLVersion, from the log file I see the error occurs while initializing containter (it other word printer settings). 
I assume you have used custom printers in your previous cura version, if yes then can you provide additional information about them. To do this you can run old Cura load any model and save as project. (and share in github) The saved project will contain information of the used printer. This information might help us to understand better the issue.

Appesteijn:
Also reported at the forum: https://ultimaker.com/en/community/52135-ultimaker-cura-31-beta?page=1#reply-193559

Thisismydigitalself:
What should I be practicing to get this fixed?
Any file i should replace to get going?

fieldOfView:
Zip up your configuration folder (%APPDATA%\cura) and post it somewhere where it can be downloaded. Something is going wrong while loading your configuration, and we don't know what (yet).

The way to get it going would be to clear out the configuration folder (but leave the folder itself in place), but that will remove all your profiles so it should not be considered a solution.

Thisismydigitalself:
Unbelievable. 
Went into user folder, deleted my printer's *.inst.cfg file and Cura 3.1 is now up and running but without my printer which i will try to set again from scratch.

Thisismydigitalself:
Ok. this file is defiantly the culprit. 
[BW_current_settings.inst.zip](https://github.com/Ultimaker/Cura/files/1502396/BW_current_settings.inst.zip)

Thisismydigitalself:
I hope my findings will help fixing this bug.

![ice_screenshot_20171124-211558](https://user-images.githubusercontent.com/30949968/33222026-c34f9144-d15c-11e7-806a-fe9b914fe22c.png)

Drayson80:
Same error happens on my side - have also a custom printer/derivate of UMO printer profile...

alekseisasin:
[...]

PR #2838, opened 2017-11-24
Fix error when CrashHandler is opened before opengl is initialised

fieldOfView:
This PR fixes a Python error being thrown when CrashHandler is opened before opengl is initialised.

Contributes to #2837

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -126,13 +126,18 @@ class CrashHandler:
         return group
 
     def _getOpenGLInfo(self):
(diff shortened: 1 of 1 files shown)


Appesteijn:
Devs: CURA-4621

ChrisTerBeke:
We should probably cherry-pick this to 3.1 as well.

ChrisTerBeke:
I merged it to master, I'll cherry pick this tomorrow when I have some more time.

ChrisTerBeke:
Picked to 3.1 as well now.

PR #2856, opened 2017-11-26
Adapt gradient in Simulation view for better visualization of feedrates

diegopradogesto:
Due to the travel speed is normally very high compared with the other speeds in the model, the gradient is now gradual instead of linear in order to show more changes in the low values and less changes in the high values. In the pictures below, you can see the difference after the changes. Red colors are also avoided since it could indicate that there is an error.
gradient
gradient1

This PR also adapts the button styles and themes' colors. Play/Pause button is set to the left and reduce the size.


fieldOfView:
Could it be worthwhile to make the gradient "draggable" to change the bias? This could be implemented by creating a virtual "middle" step that can be dragged left/right/center, and effectively lets the user "zoom in" to a region of the gradient.

diegopradogesto:
CURA-4625

Issue #2855, opened 2017-11-26
BUG: post processing not applied when printing over network


UltiArjan:
firmware 4.0.1.20171023
cura 3.1 beta (but did not validate older versions)
pc: win 10
connection: wired network

Found what I think is a bug, no idea if it's specific for 3.1.

When using a post-processing g-code modifier, In my case "pause at height" on the UM3, it will not be applied when printing over the network directly from cura. If however you print via USB, or even if you first save the file somewhere and than print over network, the change will be applied.

fieldOfView:
Good catch. This is a regression introduced with Cura Connect. Should be fixed with the PR referenced above.

ChrisTerBeke:
For reference: https://community.ultimaker.com/topic/21065-um3-cura-304-pause-on-height-not-work/.

PR #2857, opened 2017-11-26
Allow postprocessing before sending data to the printer

fieldOfView:
This PR fixes PostProcessing plugins and possibly other actions on the gcode when printing to a print cluster (Cura Connect). See OutputDevice.py: "output device subclasses are completely free to implement writing however they want, though you should emit writeStarted and related signals whenever certain events happen related to the write process."

This PR fixes https://github.com/Ultimaker/Cura/issues/2855
Also see https://github.com/Ultimaker/Cura/commit/91e8ac6868762595a73363230234d223c1270753

You'll probably want to cherry-pick this to 3.1 as well.


ChrisTerBeke:
Good one. @nallath this might need to go in your refactor as well?

nallath:
Yep.

ChrisTerBeke:
Devs: CURA-4658.

PR #2864, opened 2017-11-27
desktop.in: Ultimaker Cura everywhere!

thopiekar:


ChrisTerBeke:
Are we sure this doesn't affect lookups from other applications?

thopiekar:
On Linux? No, this is only an menu entry - nothing else.

thopiekar:
Also we are overwriting this file in cura-build when creating the AppImage.
So this change is only visible in my Cura PPAs.

thopiekar:
Works:
image
image

Thanks @ChrisTerBeke

Issue #2810, opened 2017-11-19
Anonymous statistics (SliceInfo) should be opt-in not opt-out


onitake:
According to #750, this issue has been raised before, but was only partially addressed.

In its current version, Cura tells the user on first launch that Ultimaker is collecting anonymous slicing statistics and that this can be turned off in the settings.

While this is apparently better than before (where data would be collected without the user's consent), it may still be problematic for some people. If possible, there should be a link in the dialogue that allows disabling data collection immediately. The text doesn't even mention where said option is.

Personally, I would find it preferable to make this option opt-in instead of opt-out by changing the default to false and adapting the text accordingly.

From the feedback I received from fellow Debian maintainers, it seems like leaving this option as-is is not acceptable and will have to be changed for the Debian releases. We are currently discussing if the plugin should be removed altogether (which, I think, is unfair towards Ultimaker) or if we should just change the data collection default to `False`.

onitake:
Opt-in:
```diff
diff --git a/plugins/SliceInfoPlugin/SliceInfo.py b/plugins/SliceInfoPlugin/SliceInfo.py
index 0514c4da..2c3d6ba3 100755
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -36,11 +36,11 @@ class SliceInfo(Extension):
     def __init__(self):
         super().__init__()
         Application.getInstance().getOutputDeviceManager().writeStarted.connect(self._onWriteStarted)
-        Preferences.getInstance().addPreference("info/send_slice_info", True)
+        Preferences.getInstance().addPreference("info/send_slice_info", False)
         Preferences.getInstance().addPreference("info/asked_send_slice_info", False)
 
         if not Preferences.getInstance().getValue("info/asked_send_slice_info"):
-            self.send_slice_info_message = Message(catalog.i18nc("@info", "Cura collects anonymised slicing statistics. You can disable this in the preferences."),
+            self.send_slice_info_message = Message(catalog.i18nc("@info", "Cura may optionally collect anonymised slicing statistics. You can enable this feature in the preferences."),
                                                    lifetime = 0,
                                                    dismissable = False,
                                                    title = catalog.i18nc("@info:title", "Collecting Data"))
```

Thisismydigitalself:
Can i barge in and ask what "SliceInfo" includes?

fieldOfView:
[...]

Issue #337, opened 2015-09-05
Clicking on the telemetry sending message


Ghostkeeper:
Something that improves usability a bit, and is very common and intuitive: Clicking on a message takes you to an interface that has something to do with the message. In this case:

Version: 15.09.80
Platform: Windows 7
Qt: 5.4.1
PyQt: 5.4.1

Steps to reproduce:
- Delete your settings.
- Start Cura.
- A message appears that Cura automatically sends slicing info. Click on the body of the message.

Expected behaviour:
- The preferences menu is displayed where the option is to disable sending slicing info.

Observed behaviour:
- Nothing happens.

The same could be applied to other messages, such as that the model has automatically been resized when it was too big to be printed; open the scale tool, etc.

fieldOfView:
I like the idea. It will need some good thinking though, eg: should the aforementioned "the model has automatically been resized when it was too big to be printed" message link to the scale-tool or to the preference where this auto-sizing can be turned off?

Ghostkeeper:
Scale tool seems most intuitive to me, but that'd have to be evaluated individually for each type of message.

BagelOrb:
This idea is pretty okay, but I'm closing this because of overclouding in the Cura issues.

Please reopen if you don't agree.

Ghostkeeper:
I don't really think that overcrowding the github Cura issues is a problem. There's a search function to find stuff easily, and closing issues that aren't fixed doesn't help when you use the search function. We don't need to have an overview of stuff that's left to do here since we have our Jira tracker for that.

nallath:
Automagically sending crash reports somewhere could prove to be usefull.

onitake:
Sounds like a good idea to me, though I'd generally prefer a button instead. Makes the intent clearer and the message harder to miss.

nallath:
Doesn't the 3.1 beta already have a crash log sender?

BagelOrb:
I don't see how that's related to the issue.

ianpaschal:
@Ghostkeeper Newer versions have a button that takes you directly to the telemetry preference. Is this good enough to consider this closed, or were you thinking it's good if the actual message box is one giant button?

onitake:
@ianpaschal The only thing that bothers me a bit is that the link/button only opens the settings dialog but doesn't scroll down and/or highlight the setting. If you could implement that as well, it would be perfect. But I'm fine either way.

ianpaschal:
[...]

PR #2865, opened 2017-11-28
Add button to disable sending anonymous information

Ghostkeeper:
This is an idea I've had since the introduction of the thing. After discussion at https://github.com/Ultimaker/Cura/issues/2810 I decided to put some effort in to finally get it through. It's so easy to implement. Implements #337.


Ghostkeeper:
Developers, see issue CURA-4630.

onitake:
This did not end up in 3.1, correct?
Will it be in 3.2?

Ghostkeeper:
Yeah, this is in 3.2.

PR #2867, opened 2017-11-28
CURA-4243 Feature local container server WIP

jackha:
From CURA-4243 issue text:

Instead of shipping all printer profiles with each Cura release, I would like to make it possible for users to just download the printer profiles they really need.

Edit by Ghostkeeper for clarity: The intent of this change is to set up a repository of containers / profiles that loads only metadata at start-up and then lazily loads the rest only when needed, reducing start up time drastically. The intent is also to be able to have a remote server that implements the same interface later on.


cura/Settings/MachineManager.py
```diff
@@ -145,13 +145,13 @@ def printerOutputDevices(self):
...
     def totalNumberOfSettings(self) -> int:
-        return len(ContainerRegistry.getInstance().findDefinitionContainers(id="fdmprinter")[0].getAllKeys())
+        return len(ContainerRegistry.getInstance().findDefinitionContainers(id = "fdmprinter")[0].getAllKeys())
 
     def _onHotendIdChanged(self, index: Union[str, int], hotend_id: str) -> None:
         if not self._global_container_stack:
             return
 
-        containers = ContainerRegistry.getInstance().findInstanceContainers(type="variant", definition=self._global_container_stack.getBottom().getId(), name=hotend_id)
+        containers = ContainerRegistry.getInstance().findInstanceContainersMetadata(type = "variant", definition = self._global_container_stack.getBottom().getId(), name = hotend_id)
```
jackha: replace getBottom with definition, it makes the purpose a lot more clear

cura/Settings/ProfilesModel.py
```diff
@@ -53,7 +53,7 @@ def getInstance(cls) -> "ProfilesModel":
     def _fetchInstanceContainers(self):
         global_container_stack = Application.getInstance().getGlobalContainerStack()
         if global_container_stack is None:
-            return []
+            return {}, {}
```
jackha: Document what and why two dicts.

cura/Settings/QualityAndUserProfilesModel.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, parent = None):
     def _fetchInstanceContainers(self):
         global_container_stack = Application.getInstance().getGlobalContainerStack()
         if not global_container_stack:
-            return []
+            return {}, {}
```
jackha: Document what and why two dicts.

cura/Settings/UserProfilesModel.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, parent = None):
     def _fetchInstanceContainers(self):
         global_container_stack = Application.getInstance().getGlobalContainerStack()
         if not global_container_stack:
-            return []
+            return {}, {}
```
jackha: Document what and why two dicts.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -497,8 +496,9 @@ def deserialize(self, serialized):
 
         meta_data["approximate_diameter"] = str(round(float(property_values.get("diameter", 2.85)))) # In mm
         meta_data["properties"] = property_values
+        meta_data["definition"] = "fdmprinter"
 
-        self.setDefinition(ContainerRegistry.getInstance().findDefinitionContainers(id = "fdmprinter")[0])
+        self.setDefinition("fdmprinter")
```
jackha: This kind of thing makes me happy ^^

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -648,6 +645,163 @@ def deserialize(self, serialized):
...
+            "type": "material",
+            "status": "unknown", #TODO: Add material verification.
+            "container_type": XmlMaterialProfile,
+            "id": container_id
+        }
+
+        try:
+            data = ET.fromstring(serialized)
+        except:
+            Logger.logException("e", "An exception occurred while parsing the material profile")
```
jackha: Change log, add "metadata" somewhere.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -648,6 +645,163 @@ def deserialize(self, serialized):
         for container_to_add in containers_to_add:
             ContainerRegistry.getInstance().addContainer(container_to_add)
 
+    @classmethod
+    def deserializeMetadata(cls, serialized: str, container_id: str) -> List[Dict[str, Any]]:
```
jackha: Is it possible for deserialize to either use the result of deserializeMetadata or share some parsing code?

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -876,4 +877,22 @@ def _indent(elem, level = 0):
...
+
+#While loading XML profiles, some of these profiles don't know what variant
+#they belong to. We'd like to search by the machine ID and the variant's
+#name, but we don't know the variant's ID. Not all variants have been loaded
+#yet so we can't run a filter on the name and machine. The ID is unknown
+#so we can't lazily load the variant either. So we have to wait until all
+#the rest is loaded properly and then assign the correct variant to the
+#material files that were missing it.
+_with_missing_variants = []
+def _fillMissingVariants():
```
jackha: Maybe we can fill all variants afterwards, since we need this second pass anyway. This probably reduces complexity, you don't have to keep track of the ones missing, you can skip the check in the first pass.

fieldOfView:
This is shaping up to be a rewrite of @awhiemstra'ian proportions

cura/Settings/ContainerManager.py
```diff
@@ -317,34 +313,33 @@ def setContainerName(self, container_id, new_name):
...
     #
-    #   This effectively forwards to ContainerRegistry::findInstanceContainers.
+    # This effectively forwards to
+    # ContainerRegistry::findInstanceContainersMetadata.
     #
     #   \param criteria A dict of key - value pairs to search for.
     #
     #   \return A list of container IDs that match the given criteria.
     @pyqtSlot("QVariantMap", result = "QVariantList")
     def findInstanceContainers(self, criteria):
```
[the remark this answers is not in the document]
Ghostkeeper: It doesn't return the metadata, but only the IDs.

cura/Settings/ContainerManager.py
```diff
@@ -317,34 +313,33 @@ def setContainerName(self, container_id, new_name):
...
+            material_base_file = this_container["id"]
+            if "base_file" in this_container:
+                material_base_file = this_container["base_file"]
             # check all material container IDs with the same base
-            material_containers = self._container_registry.findInstanceContainers(base_file = material_base_file,
+            material_containers = self._container_registry.findInstanceContainersMetadata(base_file = material_base_file,
                                                                                   type = "material")
             if material_containers:
-                container_ids_to_check = [container.getId() for container in material_containers]
+                container_ids_to_check = [container["id"] for container in material_containers]
```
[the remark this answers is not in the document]
Ghostkeeper: Yes. We're searching for material profiles which have the same `base_file` as the `container_results`'s `base_file`. Naturally the container has the same `base_file` as itself.

cura/Settings/MachineManager.py
Ghostkeeper: I didn't use `getBottom`, that was already there. Anyway, I replaced all occurrences in this class of `getBottom()` with `definition`: 430f3dad27b2ad6e5d1eeddd258c045ef585fa08

cura/Settings/UserProfilesModel.py
Ghostkeeper: The return value is documented in the inherited documentation. I've documented the other return statement here: 097cc72d89882b0bb05fb2627f91f5d36dbf7042

cura/Settings/ProfilesModel.py
Ghostkeeper: The return value is documented in the inherited documentation. I've documented the other return statement here: 097cc72d89882b0bb05fb2627f91f5d36dbf7042

plugins/XmlMaterialProfile/XmlMaterialProfile.py
Ghostkeeper: I've tried, but then `deserialize` needs to basically go through each of these XML tags again and basically still use the same code structure, and then also match which metadata from `deserializeMetadata` to use for which subcontainer in `deserialize`. This made for less code.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
Ghostkeeper: The log already says that the error occurs in `deserializeMetadata`. The reason why this fails is an XML parsing error, so it doesn't matter whether it is in this one or the other.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
Ghostkeeper: We don't know which variants exist until after the calls to `deserializeMetadata`. I could remove the list and find them again but that would double the number of loops this `_fillMissingVariants` goes to (on average) because the variants that are already filled will also be handled again.

jackha: Ok let's keep it this way then :-)

fieldOfView:
Congratulations on the merge!

Having a machine_instance with a missing definition (ie: pointing to a definition that does not exist) now crashes Cura hard (throws a critical error and opens the crash handler, instead of logging an error and continuing as before).

Ghostkeeper:
I've put a try/except block around the load so that it now gives an ERROR in the log rather than crashing hard if a container can't be found: https://github.com/Ultimaker/Uranium/commit/dff1db09b9da57265e7573d95895b18ce0a9235d

PR #2868, opened 2017-11-28
Add experimental settings for adaptive layer heights - CURA-4524

ChrisTerBeke:
New experimental settings for adaptive layer heights in the engine. See https://github.com/Ultimaker/CuraEngine/pull/649 for details of this feature. Note that the default values might change depending on results from materials & processing.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5983,6 +5983,49 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
+                },
+                "adaptive_layer_height_enabled":
+                {
+                    "label": "Use adaptive layers",
+                    "description": "Adaptive layers computes the layer heights depending on the shape of the model.",
+                    "type": "bool",
+                    "default_value": false,
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": false,
+                    "settable_per_meshgroup": false
+                },
+                "adaptive_layer_height_variation":
+                {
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Let's keep the merged branch for future additions to this feature.

AxlePivot:
Addaptive layers is in Experimental features and hidden by default but it is turned on in the 0.16 Dynamic Quality option. This adaptive layers option should be moved into main layers. This was a facebook query, someone had ridges causing quality issue on a test cube as the layer heights changed and wasent aware of what 'Dynamic' setting actually did.
Cube
cube2

Ghostkeeper:
That just seems like a problem with badly optimized profiles instead. We can't really help you with that. You'll just need to reduce the Adaptive Layers Variation Step Size setting to get smoother transitions between thick and thin layers.

AxlePivot:
Yes, there is a print problem but the user (not me) didn't know the 0.16 layer height had the dynamic layer setting turned on because this was hidden in experimental section. They then couldn't work out why .16 was such a bad print compared to .2 layers. If the feature is on by default in 1.6 then it shouldn't be in experimental but listed with the layers setting?

Ghostkeeper:
Different printers have different profiles. I think you're talking about the profiles for the Creality printers? The developers didn't create those profiles. We merged them upon recommendation of Creality users, and the profiles were [scrutinised later](https://github.com/Ultimaker/Cura/issues/6106) as well. We can't print with third-party printers so we have to trust the community to provide us with working profiles. Apparently this profile is not working out for you, but that's not a bug with Adaptive Layers, is it?

PR #2872, opened 2017-11-28
add traditional chinese translation (zh_TW)

dinowchang:


Ghostkeeper:
I'm going to merge these into the master branch, then cherry-pick them to 3.1 so that they can be pulled along with our release.

PR #2874, opened 2017-11-29
Adding support for Malyan M200, Monoprice Select Mini V1 & V2

tylergibson:
This PR is adding support for the Malyan M200 and its sister printers the Monoprice Select Mini V1 and V2.  As part of this PR I've also renamed the Malyan M180 to have the consistent manufacturer_model naming convention used for other entries.

This adds the quality variants specific to the M200 for layer heights (multiples of .04375) and alternate nozzle configurations.  Credit for original authorship to Brian Corbino.


ChrisTerBeke:
These are all variants of the Wanhao Duplicator i3 right?

tylergibson:
@ChrisTerBeke No, that would be the Monoprice Select.  This is the "Mini".  The OEM is Malyan - http://malyansys.com/en/product/m200/ - although overwhelmingly they are marketed and sold (with some minor differences) by Monoprice.

ChrisTerBeke:
Devs: CURA-4655.

resources/quality/malyan_m200/malyan_m200_0.04375.inst.cfg
```diff
@@ -0,0 +1,22 @@
+[general]
+version = 2
+name = M1 Quality
+definition = malyan_m200
+
+[metadata]
+type = quality
+weight = 2
+quality_type = fine
+setting_version = 3
```
LipuFei: The `setting_version` is now 4 in the current master.

resources/quality/malyan_m200/malyan_m200_0.175.inst.cfg
```diff
@@ -0,0 +1,23 @@
+[general]
+version = 2
+name = M4 Quality
+definition = malyan_m200
+
+[metadata]
+type = quality
+weight = -1
+quality_type = fast
+global_quality = true
```
LipuFei: Some files have `global_quality = true` and some don't. I think all `global_quality = true` can be removed.

resources/definitions/malyan_m200.def.json
```diff
@@ -0,0 +1,90 @@
...
+    "inherits": "fdmprinter",
+    "metadata": {
+        "author": "Brian Corbino",
+        "manufacturer": "Malyan",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "platform": "malyan_m200_platform.stl",
+        "has_variants": true,
+        "has_variant_materials": false,
+        "has_materials": true,
```
LipuFei: There is no machine specific material profiles so, this should be `false`.

fieldOfView: Are you talking about ```has_materials```? Setting that to ```false``` will remove all materials. Were you thinking about ```has_machine_materials```?

LipuFei: I'll have a look. There is some wierd logic in fetching quality profiles which doesn't get anything if `has_materials` is set to `true` for this printer...

LipuFei:
And there are two files, one in `qualities` and one in `variants`, which have the same `quality/malyan_m200/malyan_m200_0.35.inst.cfg`. Unfortunately, this is not supported at the moment. They need to have different names. You can rename the qualities to, for example, `***mm.inst.cfg`.

resources/quality/malyan_m200/malyan_m200_0.175.inst.cfg
tylergibson: From MachineManager.py, global quality should only be set to true for the default quality profile on the machine.  This is the default quality profile, hence it being the only one set to global_quality=true
https://github.com/Ultimaker/Cura/blob/master/cura/Settings/MachineManager.py#L981

resources/definitions/malyan_m200.def.json
tylergibson: Yeah I was confounded by this as well.  Agree with fieldOfView, the machine does have materials but not machine specific materials.  Would be helpful if any of the metadata fields were documented ;)

tylergibson:
Updated based on feedback from @LipuFei

LipuFei:
@tylergibson There are conflicts... :(

tylergibson:
Rebased and resolved the conflict @LipuFei

alekseisasin:
@tylergibson I see that you are using very high precise number for the quality profiles, like layer_height = 0.04375. We will make some changes in cura UI in Recommended view and round these numbers up to two numbers after comma otherwise there is not enough space for all profiles and it looks bad. In Custom mode everything will be same and all numbers will be without any rounding.

tylergibson:
@alekseisasin as long as the values are persisted in calculations and gcode (and this is just a UI affordance), I think that's an acceptable compromise.

Ghostkeeper:
I've added a commit that properly rounds the layer heights. However to do that I had to do a merge of your branch into my later version of Cura which introduced a lot of commits into your branch. Because of the rebase that was done earlier, Github doesn't handle that properly unless a new Pull Request is made. You can see the actual differences made by this PR here: https://github.com/Ultimaker/Cura/compare/master...tylergibson:malyan_m200

Ghostkeeper:
There are a few problems that prevent us from merging this pull request still:
* The `setting_version` of the variants is still 3 instead of 4. This leads us to suspect you didn't test this on Cura 3.1 or master yet.
* The `global_quality` metadata property indicates that the quality profile should be put on the global stack instead of an extruder stack. Since you want to use settings that can be set per extruder, you should remove that metadata.
* If the printer `has_materials`, the profiles are filtered by material. (`has_machine_materials` indicates whether there are material profiles specific to this printer.) It should either be set to `false` or the quality profiles have to be made specific to a certain material. If the current quality profiles were designed for PLA I would set their material to PLA and let the rest of the materials be unsupported for now.

tylergibson:
@Ghostkeeper Thanks for the info re: metadata.  I'll run everything against 3.1 (I had been running against binary 3.0.4).  Will get this sorted and submit an updated PR.

Ghostkeeper:
Did you get around to updating your PR yet?

tylergibson:
@Ghostkeeper Sorry for the delay over the holidays.  I've simplified the definitions and tested everything against 3.1.  Added the generic material definitions that the MP Mini supports.

alekseisasin:
@tylergibson Are you sure that you made correct changes. I try run Cura with you changes and see bunch of Critical errors, I did not check why, but can you check again please ?

tylergibson:
@alekseisasin I can't actually test against master or 3.1 because your build-scripts and build-environment repos are both broken on high sierra (10.13.2) with xcode 9.  I spent half my day just trying to get the build scripts to actually run.  

I manually tested on both Windows and MacOS 3.1 binaries by copying the changed files into the local cache folder and running Cura 3.1 and slicing and printing.  No errors, no warnings.  Can you provide stderr or cura.log of the errors you're seeing?

resources/definitions/monoprice_select_mini_v2.def.json
```diff
@@ -0,0 +1,25 @@
...
+        "visible": true
+   },
+
+    "overrides": {
+        "machine_name": { "default_value": "Monoprice Select Mini V2" },
+        "adhesion_type": { "default_value": "brim" },
+        "retraction_combing": { "default_value": "noskin" },
+        "retraction_amount" : { "default_value": 2.5},
+        "retraction_speed" : { "default_value": 40},
+        "material_print_temperature_layer_0": { "value": "material_print_temperature + 5" },
```
LipuFei: An extra comma at the end. I think this is why it breaks.

LipuFei:
Hi @tylergibson, the only problem is that extra comma in one of the json files. I will merge this and fix it for you.

tylergibson:
@LipuFei THANK YOU!!

PR #2880, opened 2017-11-29
Hide material diameter when the machine has no materials

fieldOfView:
This PR hides the material diameter field from Machine Settings when the current machine does not have materials (eg Ultimaker 2 when using UltiGcode)

Contributes to https://github.com/Ultimaker/Cura/issues/1273


ChrisTerBeke:
Devs: CURA-4654.

PR #2882, opened 2017-11-30
Tweak layout and alignment of print time and cost estimation

### LINKED ISSUE #2712: Minor glitch on new costing popup
Price overflows onto new line, see here...

screenshot_2017-11-02_10-54-25

fieldOfView:
This PR tweaks the layout and alignment of print time and cost estimation tooltips, improving readability.

* table takes the full width and values are right-aligned
* time estimates are shortened
* values are right aligned
* feature/time/percentage are on a single line again
* handle long feature & material names gracefully
* totals are only shown when there are multiple extruders

Discussion:
In the current design it is fairly hard to associate the data with the label. Because there is a line of text between each datum, it is also harder to get an overview at a glance. Finally, this design is "punishing" wellbehaved (compact) languages for the chance that there may be a more verbose language.
image

 It would be better if the layout adapted itself (as a HTML table can). Here is the counter offer (this PR):

For languages that are nice and compact:
image

For languages that are fairly verbose (the exception, not the rule):
image

Note how in this design, the table uses the full width, and the times and percentages are actually right-aligned for better readability. In the verbose version, the data still align vertically with the colon of the feature name. I made the time display a bit more compact (based on the ISO8601 duration-format instead of the short format).

This PR also amends the cost specification tooltip, removing the totals when there is nothing to add up:
image
Same tooltip with multi-extrusion and long material names:
image

Fixes #2712

### DIFF
--- a/resources/qml/Sidebar.qml
+++ b/resources/qml/Sidebar.qml
@@ -350,21 +350,20 @@ Rectangle
                         var total_seconds = parseInt(base.printDuration.getDisplayString(UM.DurationFormat.Seconds))
 
                         // A message is created and displayed when the user hover the time label
-                        var content = catalog.i18nc("@tooltip", "<b>Time specification</b><br/><table>");
+                        var tooltip_html = "<b>%1</b><br/><table>".arg(catalog.i18nc("@tooltip", "Time specification"));
                         for(var feature in print_time)
                         {
                             if(!print_time[feature].isTotalDurationZero)
                             {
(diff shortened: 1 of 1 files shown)


nallath:
Nice! This is way better.

smartavionics:
Gets my vote, a definite improvement.

ChrisTerBeke:
Devs: CURA-4653.

ChrisRiddell:
Much Nicer.  :+1:

PR #2883, opened 2017-11-30
[regression] Fix material container for extruder when changing gcode flavor

fieldOfView:
This PR fixes updating material container for extruder when changing gcode flavor between Ultigcode and Marlin for the UM2(extended). Since single extrusion printers have a single extruder stack now, material container needs to be set on the extruder stack instead of the global stack.

This fixes the following scenario:
* Add a new UM2
* In Machine Settings, switch gcode flavor to Marlin
* The UM2 should now gain material selection, but Cura shows that the printer has no material set (red material selection dropdown).

Furthermore:
* Select a material
* Switch flavor back to Ultigcode (Ultimaker 2). Material selection is now removed
* Slice something
* Note that Cura still thinks the printer has a material (cost estimation in the lower right of the sidebar, which should not be available, because UM2 with Ultigcode is material agnostic).

This PR fixes these issues by setting/removing the material container on the extruder stack instead of on the global stack.

This PR is against the 3.1 branch, because it fixes a regression caused by the recent rework of single extrusion machines.

### DIFF
--- a/plugins/MachineSettingsAction/MachineSettingsAction.py
+++ b/plugins/MachineSettingsAction/MachineSettingsAction.py
@@ -188,9 +188,8 @@ class MachineSettingsAction(MachineAction):
             # In other words: only continue for the UM2 (extended), but not for the UM2+
             return
 
+        stacks = ExtruderManager.getInstance().getExtruderStacks()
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Good eye :)

Issue #2877, opened 2017-11-29
Adaptive layers - missing top of model


smartavionics:
Just playing with this interesting new option and I notice that using the default adaptive layer params, the top layer(s) are not displayed when adaptive layers is enabled:
![screenshot_2017-11-29_11-10-57](https://user-images.githubusercontent.com/585618/33372520-398a82c8-d4f6-11e7-8ee7-58b4b7d7a6e6.png)
Not enabled:
![screenshot_2017-11-29_11-09-00](https://user-images.githubusercontent.com/585618/33372529-43fcb6fe-d4f6-11e7-9ac1-47e92ef48696.png)

smartavionics:
The top skin layers are there so it looks like that the model has been truncated rather than just a display problem.

smartavionics:
Here's the model, it's large so I am scaling it to 10%.
[new_moai.zip](https://github.com/Ultimaker/Cura/files/1513587/new_moai.zip)

ChrisTerBeke:
Yeah there's some weird issue that sometimes it misses a few layers. Haven't been able to pin it down yet. Maybe you can have a quick look at the code changes to see if it's something we missed?

smartavionics:
I'm browsing the code to see if I can find the problem. One thing I did notice is that on line 137 of FffPolygonGenerator.cpp you are doing `adaptive_layer_heights->getLayers()` even if adaptive_layer_heights is null.

smartavionics:
I have fixed the top of the head being missing but have not fixed the potential null pointer reference that I mention above. I leave that to you to do.

smartavionics:
![screenshot_2017-11-29_23-12-12](https://user-images.githubusercontent.com/585618/33404048-c3c9be04-d55a-11e7-958e-9ea19072c6d3.png)

smartavionics:
Just printing the above moai with spiralization and adaptive layers. There's no way this would spiralize well without using thin layers, just look at that chin!!

![img_20171130_075051766_hdr](https://user-images.githubusercontent.com/585618/33419676-62fc99ae-d5a3-11e7-860b-d7b3fd55fca1.jpg)

smartavionics:
Well he almost came out perfect except that something terrible has happened to the start/end gcode inclusion which caused the end gcode not to be executed and so the hot end just sat there and burnt a hole in the top of his head!  

How did this happen:

````
;TIME_ELAPSED:1591.253817
G1 F1800 E-6
[...]

PR #2884, opened 2017-11-30
[CRITICAL] Prevent start & end gcode snippets being double encoded

fieldOfView:
This PR fixes a regression caused by https://github.com/Ultimaker/Cura/commit/1d6b7c71cbd3301915a0777404d0d6627373b979 which double encodes start/end gcode snippets, leading to non-valid start and end seuqences.

See https://github.com/Ultimaker/Cura/issues/2877#issuecomment-348113324 for discussion

Note: the issue does not manifest on UM2(+) or UM3 printers, because they don't use start/end gcode snippets. On all other printers, Cura outputs quite broken gcode at the moment.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -225,10 +225,10 @@ class StartSliceJob(Job):
         try:
             # any setting can be used as a token
             fmt = GcodeStartEndFormatter()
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
@Ghostkeeper ping

fieldOfView:
Thanks for the quick merge!

nickthetait:
One of the few times when a [CRITICAL] tag is worthy of being used

PR #2892, opened 2017-12-01
Fix simulation view render race (aka black cat issue!)

smartavionics:
Hi @diegopradogesto, I think this PR fixes the black cat problem. Does it make sense to you?


diegopradogesto:
Yes, it has a lot of sense. Let me double check on Monday with my colleagues using Linux and I will try to include it in the release.

smartavionics:
Since applying this fix, I haven't had a single black cat (or any other black display) whereas before I was getting them very often so I am happy that this fixes that race problem. Have a nice weekend!

PR #2896, opened 2017-12-02
Adding a factory for Qt

thopiekar:
Making use here of a PR in Uranium.

See: https://github.com/Ultimaker/Uranium/pull/313


fieldOfView:
To go with https://github.com/Ultimaker/Uranium/pull/313, I guess.

PR #2904, opened 2017-12-04
updated brazilian portuguese translations

Patola:


Ghostkeeper:
Looks good. I'll merge these.

Afterwards I'll reactivate the translation for Layer View since we renamed that back after the string freeze.

PR #2909, opened 2017-12-04
MachineManager: Adding activeMachineChanged signal

thopiekar:
Never tested it, but can imagine, that I will need it soon.

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -118,6 +118,7 @@ class MachineManager(QObject):
                                                 title = catalog.i18nc("@info:title", "Incompatible Material"))
 
     globalContainerChanged = pyqtSignal()  # Emitted whenever the global stack is changed (ie: when changing between printers, changing a global profile, but not when changing a value)
+    activeMachineChanged = pyqtSignal()
     activeMaterialChanged = pyqtSignal()
     activeVariantChanged = pyqtSignal()
     activeQualityChanged = pyqtSignal()
@@ -332,6 +333,7 @@ class MachineManager(QObject):
     def __onInstanceContainersChanged(self):
         self.activeQualityChanged.emit()
         self.activeVariantChanged.emit()
+        self.activeMachineChanged.emit()
         self.activeMaterialChanged.emit()
         self._updateStacksHaveErrors()  # Prevents unwanted re-slices after changing machine
         self._error_check_timer.start()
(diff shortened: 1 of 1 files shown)


fieldOfView:
Can you explain (document) when I would use activeMachineChanged over globalStackChanged?

LipuFei:
In this case, `activeMachineChanged` and `globalStackChanged` are the same thing. You are introducing a redundant signal, and because no one is emitting this new signal, all those activeMachine properties will never get a "changed" notification.

thopiekar:
Oh, I thought that globalContainerChanged is used around the whole code. So whenever a material or variant gets changed it will be triggered, too.
So actually I want to rename this signal 😅 .. Just to make it more self-explaining.. 😛

ChrisTerBeke:
I don't think we should rename the signal as globalContainerStackChanged is exactly what happens in the software. There's many places where it's used and will create many conflicts with the branches we're currently working on.

thopiekar:
No chance to mark it as deprecated and using "activeMachineChanged" instead? 🙄 
Currently I'm defining an API with my InterfacePlugin (for Windows Automation) and as soon as I know how to define and link events to Qt signals, I wouldn't like to keep this signal called like that.
Why I care about the name? - Well, because I want to create an algorithm which is generating the API depending on the instances inside Cura.
So there will be a list of instances and I would like to crawl through the instances, look for signals and connect them automatically with the same name to COM events.
Same with functions, where I would like to look for functions `isinstance(foo, method)` and same for variables.
Don't want to reinvent the wheel and redefine all the functions in the API, but using generation of these things create a mirror of the currently available things in Cura.
So at the end: `Cura structure` == `API structure`

Nobody wants to maintain this set of functions and variables (and later events), I have to define at the moment by hand. (Ugly work 🤢 )

thopiekar:
PS: And imagine that you know what that signal is. I didn't know it is only about the activeMachine and I'm sure other people will also get confused.

ChrisTerBeke:
How would that go then when we change things in Cura? Then your 'auto-copied' API would immediately break, with no way of adding backwards compatibility?

thopiekar:
Well, we already have a API enum for plugins and in the Cura instance you have the Cura version. Therefore enough you can check for in my point of view.

thopiekar:
@ChrisTerBeke: Here it is.. And you can override functions or add your wanted backwards-compatibility 😉  https://github.com/thopiekar/CuraInterfacePlugin/blob/18d3fd253344a109ac77a9cef7ba980e11be3e68/InterfaceHandler.py#L314

ChrisTerBeke:
Neat :)

Btw, we're working on something similar for a DBUS interface, maybe you can make the plugin repo name a bit more specific?

thopiekar:
Ok, forget about this code. Guess something inside the checks for the attributes is triggering Cura to reset 😅 

I know, this COM interface is actually wanted by Ultimaker. Having this COM server running is a requirement to connect other Windows applications to Cura and "remote control" it. Nope, need a working end concept here first, then there will be docs 😛

PR #2911, opened 2017-12-05
ExtruderManager: Adding getExtruderId(index)

thopiekar:
Since we have an getExtruderName, I think it is good to have a getExtruderId on the other hand.
This one is currently not needed by me, but can't be bad to have this option here.

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -128,6 +128,16 @@ class ExtruderManager(QObject):
     def activeExtruderIndex(self) -> int:
         return self._active_extruder_index
 
+    ##  Gets the extruder id of an extruder of the currently active machine.
+    #
+    #   \param index The index of the extruder whose id to get.
+    @pyqtSlot(int, result = str)
+    def getExtruderId(self, index):
+        try:
+            return list(self.getActiveExtruderStacks())[index].getId()
+        except IndexError:
+            return ""
+
     ##  Gets the extruder name of an extruder of the currently active machine.
     #
     #   \param index The index of the extruder whose name to get.


fieldOfView:
I don't think it is a good idea to add code that is not "yet" needed. That is sure to lead to bitrot.

thopiekar:
Sure. What would be better is to generate these functions. I mean, whenever I'm interested in a name, I also would like to have an id.

ChrisTerBeke:
We're planning on refactoring the whole machine manager setup (put everything in 1 manager which manages machine objects, machines will then have methods to get extruders etc.). @LipuFei might this be included in your idea?

LipuFei:
@ChrisTerBeke This is something that can done with the concept `Machine` when we start the refactoring.
@thopiekar Although many things have been fully thought about yet, but it is certain that the `ExtruderManager` will be gone after the refactoring.

thopiekar:
Ok, let's see then. There is no rush at the moment on these functions.
But it would be nice if you could care about choosing nice names. So we have something like `getMachineNames` and `getMachineNamesFromDefinitions` in one manager and people know then: Ah, the first one will get me all machines which I have in Cura and the second all machines, which are known by the definitions.
Reason why I would like to care more about names: https://github.com/Ultimaker/Cura/pull/2909#issuecomment-349285952

Ghostkeeper:
This function is specifically to expose the extruder ID to QML, but QML shouldn't ever have to use the extruder ID since the ID is not meant to be visible to the user.

thopiekar:
Ok, and what if an developer of an new plugin needs it? (Like I do?)
Also you can't expect that everybody is using QML 😬

Ghostkeeper:
If a developer of a plug-in needs it he should request it in Python.

QML is only for user interface. The visible stuff. The ID should not be visible. Whatever needs the ID should be implemented in Python, not in QML. Even more so if the user is not using QML it should be implemented in Python, not in QML.

fieldOfView:
I just found a usecase for getting the extruder ids in QML: if you want to use a SettingPropertyProvider for a specific extruder nr of the active printer.

Fortunately, Cura.ExtruderManager.extruderIds() exists, which provides access to all extruder ids of the active printer. So the functionality required already exists. 

As far as I can see there is no need for this PR. It can be closed.

ChrisTerBeke:
The whole extruder manager thing will get a big refactor next sprint, so then this PR will be outdated anyways.

PR #2926, opened 2017-12-07
Fix typo

gitname:


Ghostkeeper:
Maybe we should eventually change that text... It's been like 3.5 years since that rework was done.

PR #2928, opened 2017-12-07
Feature sidebar plugin target

ChrisTerBeke:
In order to dynamically set the main and sidebar views from plugins, the concept of workflow "stages" was transferred into code. Plugins can now add a stage to the workflow which will add a top bar menu entry. A stage plugin can set a main view overlay and a sidebar view. This greatly reduces the amount of logic that's needed in QML files. There are 2 new plugins added that are mandatory: PrepareStage and MonitorStage.

Uranium PR https://github.com/Ultimaker/Uranium/pull/315 is also needed for this.


fieldOfView:
Does that mean I can create a plugin that replaces the "Prepare" stage with a version that is virtually identical but does not have the materials link? Neat!

plugins/MonitorStage/MonitorStage.py
```diff
@@ -0,0 +1,73 @@
...
+        if not output_device:
+            return "tab_status_unknown"
+
+        if hasattr(output_device, "acceptsCommands") and not output_device.acceptsCommands:
+            return "tab_status_unknown"
+
+        if not hasattr(output_device, "printerState") or not hasattr(output_device, "jobState"):
+            return "tab_status_unknown"
+
+        # TODO: refactor to use enum instead of hardcoded strings?
```
diegopradogesto: I agree with this. Using enum or even a dict can keep things clearer instead of ifelse structures.

fieldOfView: Or even a switch statement.

ChrisTerBeke: @nallath something for your refactor?
@fieldOfView switch case is shorter, but still relies on exact string comparison. With an enum we can autocomplete/typehint and make sure we use the same string throughout Cura.

fieldOfView:
Unfortunately the OctoPrint monitor is broken now. The camera is not showing, and there is not even a dark viewport overlay. Any hints?

ChrisTerBeke:
I'm debugging some monitor things right now. There should be an overlay though when the plugin does not provide a main view in monitoring mode.

ChrisTerBeke:
Weird, the plugin correctly sets `self._monitor_view_qml_path` and that's all that's needed for the monitor view to pick up the qml component.

fieldOfView:
When switching between printers while on the monitor tab, I do get the OctoPrint camera, but it is all the way in the top left of the window instead of centered in the viewport. This leads me to believe that there is a viewport overlay but it has no width/height (but only for the OctoPrint monitor tab, sofar)

ChrisTerBeke:
Ah, the Loader `monitorViewComponent` does indeed not have a width and height property set. Let me fix that now.

ChrisTerBeke:
Can you check if baea9eb1e09d1f850a3c1e862aa63ce68ece3167 fixed it for you?

fieldOfView:
Not entirely. I get the viewport overlay, but the camera view is still missing. By switching between printers I can get it to appear. It is now centered horizontally, but centered at the top. It might still be getting a 0 height somewhere.

ChrisTerBeke:
I see the octoprint monitor component does not have a width and height set? You're probably relying on the parent having set the horizontal and vertical center. @nallath is also refactoring the whole printer output device structure so we need to keep that in mind as well.

fieldOfView:
Hmm, there seem to be more issues at stake here...

When starting up Cura, it always shows me the "not connected" (?) state in the topbar, even though in Manage Printers I see the printer is connected and available. The print monitor also says that the octoprint instance is connected. The topbar does not update untill I switch to/from another printer.

When the camera does show up (horizontally centered, vertically hugging the top with c4d7a33) and I resize the window, the camera image is suddenly stretched to fill the window.

The camera not showing up seems to be because the visible property of the MonitorItem is not changed when first opened (whereas it was before). That change is used to start/stop the video stream. Can you test with a "legacy" (non-cluster) UM3? That should show the same behavior because it has virtually the same MonitorItem.qml

ChrisTerBeke:
I'll try to find a non-cluster UM3 in the office, but they're hard to come by nowadays :P

The status icon update is a known thing, it will be fixed in the output device refactor which is due very soon (def. before any next release).

ChrisTerBeke:
Just heard the icon status was already fixed in the refactor branch. Also fixed some other small things a minute ago.

ChrisTerBeke:
@fieldOfView Jaime just did a legacy UM3 camera fix in his branch: https://github.com/Ultimaker/Cura/commit/ae629e2968fbbd251c779db82a806e8432d4dd30#diff-4e5173f3a1e2c7391b89268ab0ab05b8L9. The same update would be required for the OctoPrint plugin I guess.

fieldOfView:
I have fixed the visibility and position of the camera view in the OctoPrint plugin, but scaling is still broken. @nallath seems to have just taken out the scaling part (ie: when the viewport is too small to fit the image).

PR #2939, opened 2017-12-08
Getting logs earlier!

thopiekar:
Have currently the issue here, that when running Cura as a COM service, that the Cura.exe is popping up for (feels like) 1s or less and crashes.

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -2,12 +2,27 @@
 
 # Copyright (c) 2015 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
+
 import os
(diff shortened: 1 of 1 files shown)


thopiekar:
While writing this PR I found my logs.. At "%APPDATA%/Local", which is the old location.
```
usage: cura [-h] [--version] [--external-backend] [--disable-textures]
            [-qmljsdebugger QMLJSDEBUGGER] [--single-instance]
            [file [file ...]]
cura: error: unrecognized arguments: -Embedding
```

Appesteijn:
👍

PR #2960, opened 2017-12-10
Extra machine settings

Ghostkeeper:
This adds a few extra machine settings in order to be able to store more information about machines. These settings are primarily used by X3GWriter in order to output correct X3G. However I took care to select only settings that could feasibly be used for slicing in general. For instance, the steps per MM can be used to align g-code output to the step grid, which would improve quality and speed while reducing g-code size.

The settings are correctly configured for Malyan M180 as well. If we're going to use them for actual slicing we should set them correct for the Ultimaker printers too.

### DIFF
--- a/resources/definitions/m180.def.json
+++ b/resources/definitions/m180.def.json
@@ -25,8 +25,7 @@
             "default_value": true
         },
         "machine_nozzle_size": {
-            "default_value": 0.4,
-            "minimum_value": "0.001"
+            "default_value": 0.4
         },
         "machine_head_with_fans_polygon": {
             "default_value": [
@@ -35,6 +34,21 @@
                 [ 18, -18 ],
                 [ 18, 35 ]
             ]
+        },
+        "machine_max_feedrate_z": {
+            "default_value": 400
+        },
+        "steps_per_mm_x": {
+            "default_value": 93
+        },
+        "steps_per_mm_y": {
+            "default_value": 93
+        },
+        "steps_per_mm_z": {
+            "default_value": 1600
+        },
+        "steps_per_mm_e": {
+            "default_value": 92
         },
         "gantry_height": {
             "default_value": 55
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -606,6 +606,73 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_steps_per_mm_x":
+                {
+                    "label": "Steps per Millimeter (X)",
+                    "description": "How many steps of the stepper motor will result in one millimeter of movement in the X direction.",
+                    "type": "int",
+                    "default_value": 50,
+                    "minimum_value": "0.0000001",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "machine_steps_per_mm_y":
+                {
+                    "label": "Steps per Millimeter (Y)",
+                    "description": "How many steps of the stepper motor will result in one millimeter of movement in the Y direction.",
+                    "type": "int",
+                    "default_value": 50,
+                    "minimum_value": "0.0000001",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "machine_steps_per_mm_z":
+                {
+                    "label": "Steps per Millimeter (Z)",
+                    "description": "How many steps of the stepper motor will result in one millimeter of movement in the Z direction.",
+                    "type": "int",
+                    "default_value": 50,
+                    "minimum_value": "0.0000001",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "machine_steps_per_mm_e":
+                {
+                    "label": "Steps per Millimeter (E)",
+                    "description": "How many steps of the stepper motors will result in one millimeter of extrusion.",
+                    "type": "int",
+                    "default_value": 1600,
+                    "minimum_value": "0.0000001",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "machine_endstop_positive_direction_x":
+                {
+                    "label": "X Endstop in Positive Direction",
+                    "description": "Whether the endstop of the X axis is in the positive direction (high X coordinate) or negative (low X coordinate).",
+                    "type": "bool",
+                    "default_value": false,
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
+                "machine_endstop_positive_direction_y":
+                {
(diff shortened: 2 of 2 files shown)


Ghostkeeper:
Developers, see issue CURA-4696.

fieldOfView:
As long as these settings aren't used by curaengine, it might make more sense to define them in the printer definition instead of in fdmprinter.def.json. Or you could introduce a "x3gprinter.def.json" which inherits from fdmprinter.def.json. The M180 definition would then inherit from x3gprinter.def.json to get these settings. This way we can separate curaengine settings from settings introduced by plugins.

Ghostkeeper:
I considered that first, actually, but semantically that doesn't make sense. Steps per millimetre for each axis isn't only a property of printers that output to X3G. It's a property that applies to all Cartesian printers. (I considered making a subdefinition for Cartesian printers as well, but then all printers except Delta printers would have to change their inheritance.) Arguably, the direction of the end stop is not applicable to g-code since the g-code has no control for homing other than a "home" command. I'm willing to put those three in a separate subdefinition and put all X3G printers under that subdefinition, but I didn't think that'd be worth the complexity.

d235j:
Steps per mm may be useful to know in order for the slicer to determine at what delta-x (as well as speed) the printer loses significant precision and generate travel moves more carefully to avoid such loss of precision. But that's **very highly** hypothetical at this point.

ohrn:
Hi Ghostkeeper,

I've been toiling away for some time at creating definitions for two X3G based printers, ZYYX Plus and Pro.

These printers are already known to gpx so I went for the simpler solution of configuring only the gpx type:
https://github.com/ohrn/Cura/commit/cad3ae51482bfeb4a949796d2da62451d655d4c5
https://github.com/ohrn/X3GWriter/commit/1cee137f5c33ccd3a2d017ded07a63a67e88189e

It's not as comprehensive as this pull request of course, but I think it still has value as an optional easier way to configure printers that already exist in gpx, as it avoids the need to figure out and duplicate a lot of values from gpx.

If there is interest I can turn the above commits into a proper pull request, either on master or on top of this branch.

I'm still tuning the printer definitions but I guess I should have started submitting patches like this one a long time ago, perfect is the enemy of good etc. :)

/ohrn

d235j:
A major reason that I wanted to be able to specify these parameters in the printer definition is because if one changes out parts — e.g. the extruder — it would take a lot of effort to add a new gpx printer definition to gpx itself. (An issue with Makerbot Replicator printers is less than ideal torque on the extruder).

This way I can just modify the printer definition to match the specific hardware on the printer, without having to hack GPX.

ohrn:
I totally agree with that and I'm offering this as an additional simple configuration method, not as a replacement for this patch.

The ZYYX printers are mass produced turnkey machines, if the gpx settings don't work you send it back. :)

d235j:
Any thoughts/updates on this? Looking forward to be able to have proper X3G printer support!

ChrisTerBeke:
If you rebase this PR to 3.2 branch, I can have a look today?

alekseisasin:
I agree with @fieldOfView , if these settings are not used in CuraEngine they should be stored in the printer definition, otherwise it is not logical why we send all these settings to the CuraEngine without using them.

d235j:
The argument for having these around is that the steps per mm parameters and some others may be useful for the slicer. But if it is decided that is not necessary, that's fine.

@Ghostkeeper any updates on your end? Should I rebase this?

fieldOfView:
In the [Z Offset plugin](https://github.com/fieldofview/zoffsetplugin), I have found a way to add settings to definitionscontainers from inside a plugin. This could be used in the X3G writer plugin too.

The method currently uses access to private members of the SettingDefinition and DefinitionContainer classes - which is a bit naughty - but a couple of small public methods in those classes could make it more elegant.

ChrisTerBeke:
@fieldOfView we tried that once as a hack for material diameter in the extruder stack, when horribly wrong... problem is that the definition is supposed to be static (e.g. not changing after initial load). Some other parts of Cura are scared when that happens anyways :P

cc @LipuFei

fieldOfView:
It did take some fiddling to get it to work, but sofar it seems to be working for my plugin. I'll see what sort of feedback the plugin gets (just to be sure that it is not just working for me). 

I can imagine that it would be trickier with adding settings to extruder definitions that are already in global definitions.

d235j:
@fieldOfView I also did it here:
https://github.com/d235j/X3GWriter/blob/gpx-cfg-generation/X3GWriter.py

Again, messy...

ChrisTerBeke:
For reference: https://github.com/Ultimaker/Cura/commit/3927c70625108c71ba63045bc7d1dd6a595db8d0#diff-5cbee193d7820162884bae106e4dcb64L63.

(note that we added a `stack.definition.addDefinition()` method in Uranium, you could mis-use that if it works for your case?)

fieldOfView:
DefinitionContainer.addDefinition resulted in having my setting appear in the sidebar twice; once in the category I expected, and in the "root" of the settings tree next to the categories (or somehow in the Command Line Settings category, depending on where the SettingDefinitionsModel was shown).

fieldOfView:
> I also did it here: https://github.com/d235j/X3GWriter/blob/gpx-cfg-generation/X3GWriter.py

Ooh, that is "creative" in a sort of horrible way ;-)... The way you implemented it, all settings have steps_per_mm, endstop_is_max and steps_per_revolution properties (other properties being things like value, default_value, unit, minimum_value, etc).

d235j:
@fieldOfView ha yeah, I know full well this is pretty horrible (as I alluded to at https://github.com/Ghostkeeper/X3GWriter/pull/4#issuecomment-347563664 )

Unfortunately at the time that's what worked.

rpavlik:
FWIW, regarding knowing steps per mm, slic3r actually wants to know that as well as what microstepping the printer uses on the Z axis, so that it rounds your layer height to a multiple of a full step for more precise positioning.

LipuFei:
@Ghostkeeper Conflicts to resolve...

jackha:
I see the settings being sent to CuraEngine. There were no conflicts merging locally.
Merged!

resources/definitions/m180.def.json
```diff
@@ -36,6 +35,21 @@
                 [ 18, 35 ]
             ]
         },
+        "machine_max_feedrate_z": {
+            "default_value": 400
+        },
+        "steps_per_mm_x": {
```
jackha: Shouldn't this be machine_steps_per_mm_x ? And the same for the rest of these settings...
It's merged, it can be changed in master now :-)

LipuFei: Nice catch! Fixed in a477b8cdc

jackha: Nice fix ^^

Issue #2871, opened 2017-11-28
Layer view extremly slow Cura 3.0.4 win64


rekom:
I´m using cura 3.0.4 on win10 64bits with an AMD APU computer.
When in layer view, rotating the model is extremly slow. This is fixed with the option "force layer compatibility mode". 
In the windows monitor, I can see that the iGPU (there's no discrete gpu) is used at 100% in 3d. 
When i use an intel haswell iGPU, (also without discrete GPU) it runs ok.
Weirdly, in 3d benchmarks (3d Mark) AMD iGPU is faster than the intel one.

Could it be a driver problem?

ChrisTerBeke:
Do you have the latest video drivers for the discrete GPU installed?

rekom:
It´s the lasted drivers but i don´t have a discrete card, it´s just the iGPU rendering

rekom:
Just tested a luxmark benchmark in this case the AMD GPU is much more slower than the intel one...

fieldOfView:
This is the reason "force compatibility mode" is there; some GPUs support the required functionality for the modern layer view, but only at a snail's pace. Sometimes newer drivers help, but for some GPUs the hardware is just too slow. Fortunately there's a partial fallback that is much faster.

Ghostkeeper:
I hope that compatibility mode is sufficient for you. I've also just created a pull request that should almost double the frame rate of layer view: https://github.com/Ultimaker/Cura/pull/2970 (but that's maybe for next release).

rekom:
@Ghostkeeper Same machine Cura 3.1 with last AMD drivers (but i believe it didn´t change anything the APU is too old to get fix anymore0) and it works really really well. It´s not 100 fps/s but i can really use the ´normal´mode!

PR #2970, opened 2017-12-13
Layer view backface culling

Ghostkeeper:
This enables backface culling for layer view.
[Backface culling](https://en.wikipedia.org/wiki/Back-face_culling) effectively says that you only render one side of a triangle. If it's facing the other way, you don't render it. This is often used in games when they have watertight models, since you can't see the inside of the model anyway; they simply make all triangles face outwards. Backface culling then cuts the number of triangles to render in half because all the triangles at the back side of the model will be facing the other way and so won't be rendered.

In layer view I first had to fix the winding order but could then simply enable the backface culling that was already implemented. All the sides of the 3D tubes facing away from the camera won't be rendered, but you wouldn't see those anyway since they are on the other side of the tubes. The result: **Almost double the frame rate.**

Compatibility mode still works normally. Travel and retracted travel moves are rendered doubly now, a front and back side, to prevent them from being culled if you look at them from their back side. Other rendered objects are not touched. The only change other than performance should be that if you were to look from the inside of a tube (zoomed in very far), you won't be able to see the tube any more.

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -92,7 +92,7 @@ class SimulationPass(RenderPass):
 
         self.bind()
 
-        tool_handle_batch = RenderBatch(self._tool_handle_shader, type = RenderBatch.RenderType.Overlay)
+        tool_handle_batch = RenderBatch(self._tool_handle_shader, type = RenderBatch.RenderType.Overlay, backface_cull = True)
         head_position = None  # Indicates the current position of the print head
         nozzle_node = None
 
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
Made one more slight improvement, mostly for style. It's done now.

Developers, see issue CURA-4709.

rekom:
@Ghostkeeper Concerning my issue #2871 Same machine, Cura 3.1 with last AMD drivers (but i believe it didn´t change anything the APU is too old to get fix anymore) and it works really really well. It´s not 100 fps/s but i can really use the ´normal´ mode!

PR #2974, opened 2017-12-13
Network rewrite

nallath:
Octoprint is octoprint not yet fixed. 

The main difference is that the PrinterOutputDevice is no longer 1-1 connected with a single printer, but  1-*. So instead of a PrinterOutputDevice directly controling a single printer, it has a list of PrinterOutputModels that double as convenient qObjects, so they can also be used as models in QML.


cura/PrinterOutput/ExtruderOuputModel.py
```diff
@@ -0,0 +1,70 @@
...
+    def activeMaterial(self) -> "MaterialOutputModel":
+        return self._active_material
+
+    def updateActiveMaterial(self, material: Optional["MaterialOutputModel"]):
+        if self._active_material != material:
+            self._active_material = material
+            self.activeMaterialChanged.emit()
+
+    ##  Update the hotend temperature. This only changes it locally.
+    def updateHotendTemperature(self, temperature: int):
```
sedwards2009: Are you sure that temperature is an int and not a float? The printer rest API reports floats for the current temp.

cura/PrinterOutputDevice.py
```diff
@@ -25,662 +32,117 @@
...
-    def setProgress(self, progress):
-        if self._progress != progress:
-            self._progress = progress
-            self.progressChanged.emit()
+    @pyqtProperty(bool, notify=acceptsCommandsChanged)
+    def acceptsCommands(self):
+        return self._accepts_commands
+
+    ##  Set a flag to signal the UI that the printer is not (yet) ready to receive commands
+    def setAcceptsCommands(self, accepts_commands):
```
sedwards2009: Who is meant to use this? Is this a protected method for use by subclasses?

cura/PrinterOutput/ExtruderOuputModel.py
nallath: I copied the type from the old plugin, where it was also an int. I can't find back why it was set to be an int.

cura/PrinterOutput/PrinterOutputController.py
```diff
@@ -0,0 +1,46 @@
...
+from UM.Logger import Logger
+
+MYPY = False
+if MYPY:
+    from cura.PrinterOutput.PrintJobOutputModel import PrintJobOutputModel
+    from cura.PrinterOutput.ExtruderOuputModel import ExtruderOuputModel
+    from cura.PrinterOutput.PrinterOutputModel import PrinterOutputModel
+
+
+class PrinterOutputController:
```
sedwards2009: I'm trying to figure out the value of the `PrinterOutputController` and its relation to the `PrinterOutputModel`. Is the controller needed? Right now it just looks like the controller contains the verbs and the model contains the nouns, or more accurately, the data is in the model and the functions which operate on the data live in the controller. In a way OOP was about combining data with the functions that operate on it. i.e. objects have data and can do things.

nallath: Except that in this case, it would cause a slew of uneeded extra objects; In order for it to work you'd have a ClusterPrintjobOutputModel, ClusterExtruderOutputModel, LegacuUM3ExtruderOutputModel, etc. This way, i can keep the object itself as a proxy, which delegates the changing of a parameter to the controller (that is specific to the type of printer it's actually connected with)

cura/PrinterOutput/ExtruderOuputModel.py
nallath: Fixed in 005ba4ac5300d86a86fd86a1464bd5d645928cb5

cura/PrinterOutputDevice.py
nallath: You're right, it should be protected. It's again something I copied from the old code. Fixed in 80526893c277ce66853e0dae871ea3b0261a0f96

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -69,6 +69,11 @@ def __init__(self, serial_port, baud_rate = None):
     #   is ignored.
     #   \param kwargs Keyword arguments.
     def requestWrite(self, nodes, file_name = None, filter_by_machine = False, file_handler = None, **kwargs):
+        if self._is_printing:
+            return # Aleady printing
+
+        Application.getInstance().showPrintMonitor.emit(True)
```
ChrisTerBeke: Once the plugins are updated we can remove this emit right? It's only there for backwards compatibility (as we have the setStage() method now).

plugins/USBPrinting/USBPrinterOutputController.py
```diff
@@ -31,6 +31,17 @@ def homeHead(self, printer):
     def homeBed(self, printer):
         self._output_device.sendCommand("G28 Z")
 
+    def setJobState(self, job: "PrintJobOutputModel", state: str):
+        if state == "pause":
```
ChrisTerBeke: Would it be possible to use an enum for possible states instead of using hardcoded string everywhere this is used?

plugins/USBPrinting/USBPrinterOutputDevice.py
nallath: Yeah, i'll change it now.

nallath: see bd4797404df6761c84bf1b8b7f160957f8199805

fieldOfView:
Er... can I get an executive overview of what changed? I need to get started with updating the OctoPrint plugin.

I don't have an UM3, which makes it kind of hard to see what is going on and how things should work now.

nallath:
USB printing is also fixed up, so that should also give you a decent hint as to how to get it up & running again.
Most important change for you is that PrinterOutputDevice now has a list of printers. In order to send commands to a specific printer, you need a controller. The controller should be pretty self explanatory.

nallath:
So, after a slew of dusting & cleaning of that 'ol USB print code, I think i've managed to clean it up rather nicely. 
As far as I now see, this is nearing alpha / beta state.

fieldOfView:
This is what the print monitor looks like (for a non-connected UM3 here):
image

nallath:
Hmm, must be some mixup between the stage view & my rework. I'll have a look.

nallath:
@ChrisTerBeke: I thought the idea was that the entire sidebar would be unloaded once we switch? Now it just keeps certain selection boxes for nozzle and the like there, but just tries to draw something over it.

plugins/UM3NetworkPrinting/LegacyUM3OutputDevice.py
```diff
@@ -0,0 +1,614 @@
...
+        if self._authentication_failed_message:
+            self._authentication_failed_message.hide()
+        if self._authentication_succeeded_message:
+            self._authentication_succeeded_message.hide()
+        self._sending_gcode = False
+        self._compressing_gcode = False
+        self._authentication_timer.stop()
+
+    ##  Send all material profiles to the printer.
+    def sendMaterialProfiles(self):
```
fieldOfView: private?

nallath: Good point. Changed. 37461a7934fb71d6cacf9bdab776e4f0952805db

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,306 @@
...
+
+        self._api_prefix = ""
+        self._address = address
+        self._properties = properties
+        self._user_agent = "%s/%s " % (Application.getInstance().getApplicationName(), Application.getInstance().getVersion())
+
+        self._onFinishedCallbacks = {}
+        self._authentication_state = AuthState.NotAuthenticated
+
+        self._cached_multiparts = {}
```
sedwards2009: I've looked at the code below and I can't quite figure out what the purpose is of this dict. Is it needed to just hold a reference to the form part during the HTTP request, and stop the GC from killing it?

fieldOfView:
I see the WIP has been removed from the title of the issue, but the print monitor is still utterly broken in this branch (see screenshot above). That makes it fairly hard to start updating the OctoPrint plugin.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,306 @@
...
+            user = os.environ.get(name)
+            if user:
+                return user
+        return "Unknown User"  # Couldn't find out username.
+
+    def _clearCachedMultiPart(self, reply):
+        if id(reply) in self._cached_multiparts:
+            del self._cached_multiparts[id(reply)]
+
+    def put(self, target: str, data: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]):
```
sedwards2009: These HTTP related methods are only meant for subclasses. They should start with a single underscore.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,306 @@
...
+        self._last_request_time = time()
+        reply = self._manager.get(request)
+        if onFinished is not None:
+            self._onFinishedCallbacks[reply.url().toString() + str(reply.operation())] = onFinished
+
+    def delete(self, target: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]):
+        if self._manager is None:
+            self._createNetworkManager()
+        self._last_request_time = time()
+        pass
```
sedwards2009: Is this method complete? It seems to be half done.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,306 @@
...
+            del self._cached_multiparts[id(reply)]
+
+    def put(self, target: str, data: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]):
+        if self._manager is None:
+            self._createNetworkManager()
+        request = self._createEmptyRequest(target)
+        self._last_request_time = time()
+        reply = self._manager.put(request, data.encode())
+        if onFinished is not None:
+            self._onFinishedCallbacks[reply.url().toString() + str(reply.operation())] = onFinished
```
sedwards2009: The key generation code appears 5 separate times in this file, and this particular line appears 4 times. I think we can safely put it behind its own method.

I've got a few doubts about how this dict works. The keys aren't fantastic. (`str(reply.operation())` is just going to a be a `1`, `2`, or `3`). And this mechanism is built on the assumption that we will never have to requests to the same URL running at the same time. This seems reasonable, but doesn't make me feel comfortable.

sedwards2009: Is there a reason why we can't use the reply object as the key?

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+    def get(self, target: str, onFinished: Optional[Callable[[QNetworkReply], None]]) -> None:
+        if self._manager is None:
+            self._createNetworkManager()
+        request = self._createEmptyRequest(target)
+        self._last_request_time = time()
+        reply = self._manager.get(request)
+        if onFinished is not None:
+            self._onFinishedCallbacks[reply.url().toString() + str(reply.operation())] = onFinished
+
+    def delete(self, target: str, onFinished: Optional[Callable[[QNetworkReply], None]]) -> None:
```
sedwards2009: This method appears to be incomplete.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+            user = os.environ.get(name)
+            if user:
+                return user
+        return "Unknown User"  # Couldn't find out username.
+
+    def _clearCachedMultiPart(self, reply: QNetworkReply) -> None:
+        if id(reply) in self._cached_multiparts:
+            del self._cached_multiparts[id(reply)]
+
+    def put(self, target: str, data: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]) -> None:
```
sedwards2009: These HTTP related methods are only meant for subclasses. They should start with a single underscore.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+        if self._manager:
+            self._manager.finished.disconnect(self.__handleOnFinished)
+            #self._manager.networkAccessibleChanged.disconnect(self._onNetworkAccesibleChanged)
+            self._manager.authenticationRequired.disconnect(self._onAuthenticationRequired)
+
+        self._manager = QNetworkAccessManager()
+        self._manager.finished.connect(self.__handleOnFinished)
+        self._last_manager_create_time = time()
+        self._manager.authenticationRequired.connect(self._onAuthenticationRequired)
+        #self._manager.networkAccessibleChanged.connect(self._onNetworkAccesibleChanged)  # for debug purposes
```
sedwards2009: This line and its friend just here above can go. There is no `_onNetworkAccesibleChanged` either (and it is spelt incorrectly too).

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def __handleOnFinished(self, reply: QNetworkReply) -> None:
+        # Due to garbage collection, we need to cache certain bits of post operations.
+        # As we don't want to keep them around forever, delete them if we get a reply.
+        if reply.operation() == QNetworkAccessManager.PostOperation:
+            self._clearCachedMultiPart(reply)
```
sedwards2009: Ok, that comment explains a lot about what the multipart cache stuff is about. Can we come up with a better name than 'cached'? When I see the word cache I expect its contents to be reused, but here we are just using it to hold on to object and prevent GC.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+        request = self._createEmptyRequest(target, content_type=None)
+        multi_post_part = QHttpMultiPart(QHttpMultiPart.FormDataType)
+        for part in parts:
+            multi_post_part.append(part)
+
+        self._last_request_time = time()
+
+        reply = self._manager.post(request, multi_post_part)
+
+        self._cached_multiparts[id(reply)] = (multi_post_part, reply)
```
sedwards2009: You could just use the reply as the key. That *should* be fine.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
```diff
@@ -0,0 +1,387 @@
...
+
+    # This is a bit of a hack, as the notify can only use signals that are defined by the class that they are in.
+    # Inheritance doesn't seem to work. Tying them together does work, but i'm open for better suggestions.
+    clusterPrintersChanged = pyqtSignal()
+
+    def __init__(self, device_id, address, properties, parent = None):
+        super().__init__(device_id = device_id, address = address, properties=properties, parent = parent)
+        self._api_prefix = "/cluster-api/v1/"
+
+        self._number_of_extruders = 2
```
sedwards2009: Is this likely to be a problem in the future if UM releases a printer with more or less extruders?

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
```diff
@@ -0,0 +1,387 @@
...
+
+        self._print_jobs = []
+
+        self._monitor_view_qml_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "ClusterMonitorItem.qml")
+        self._control_view_qml_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "ClusterControlItem.qml")
+
+        # See comments about this hack with the clusterPrintersChanged signal
```
sedwards2009: I read the comment above, still don't understand the problem.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -0,0 +1,305 @@
...
+        max_chars_per_line = int(1024 * 1024 / 4)  # 1/4 MB per line.
+        byte_array_file_data = b""
+        batched_line = ""
+
+        for line in self._gcode:
+            if not self._compressing_gcode:
+                self._progress_message.hide()
+                # Stop trying to zip / send as abort was called.
+                return None
+            batched_line += line
```
sedwards2009: Repeatedly concatenating strings is incredibly slow in Python. It is a effectively a gotcha. I thought I had fixed code like this in Cura over a year ago, but now I see it here. Don't concat the lines, collect them in a list and do `"".join(my_lines)` in one go.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
```diff
@@ -0,0 +1,387 @@
...
+        # Keep a list of all completed jobs so we know if something changed next time.
+        self._finished_jobs = finished_jobs
+
+    def _update(self):
+        if not super()._update():
+            return
+        self.get("printers/", onFinished=self._onGetPrintersDataFinished)
+        self.get("print_jobs/", onFinished=self._onGetPrintJobsFinished)
+
+    def _onGetPrintJobsFinished(self, reply: QNetworkReply):
```
sedwards2009: This method is just too long. Splitting out the center part is probably the best way to go.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
```diff
@@ -0,0 +1,387 @@
...
+                    removed_job.assignedPrinter.updateActivePrintJob(None)
+                    removed_job.stateChanged.disconnect(self._printJobStateChanged)
+                    self._print_jobs.remove(removed_job)
+                    job_list_changed = True
+
+            # Do a single emit for all print job changes.
+            if job_list_changed:
+                self.printJobsChanged.emit()
+
+    def _onGetPrintersDataFinished(self, reply: QNetworkReply):
```
sedwards2009: This method hasn't reached 100 lines yet but it is damn close. Please split it.

plugins/UM3NetworkPrinting/DiscoverUM3Action.py
```diff
@@ -25,24 +28,24 @@ def __init__(self):
 
         Application.getInstance().engineCreatedSignal.connect(self._createAdditionalComponentsView)
 
-        self._last_zeroconf_event_time = time.time()
-        self._zeroconf_change_grace_period = 0.25 # Time to wait after a zeroconf service change before allowing a zeroconf reset
+        self._last_zero_conf_event_time = time.time()
 
-    printersChanged = pyqtSignal()
+        # Time to wait after a zero-conf service change before allowing a zeroconf reset
+        self._zero_conf_change_grace_period = 0.25
```
sedwards2009: I can't see what the units are here. Is `_zero_conf_change_grace_period_seconds` an improvement?

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: Nope, they are not. They can be used by the controllers as well.

cura/PrinterOutput/NetworkCamera.py
```diff
@@ -0,0 +1,119 @@
...
+        self._manager = None
+
+        self._started = False
+
+    def getImage(self):
+        return self._image
+
+    ##  Ensure that close gets called when object is destroyed
+    def __del__(self):
+        self.close()
```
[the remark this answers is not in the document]
nallath: Derp. It should have been .stop. Fixed in ef46f514970edca956f9942cdf35a1a19ad9e4dd

cura/PrinterOutput/ExtruderOuputModel.py
```diff
@@ -0,0 +1,70 @@
...
+
+from typing import Optional
+
+MYPY = False
+if MYPY:
+    from cura.PrinterOutput.PrinterOutputModel import PrinterOutputModel
+    from cura.PrinterOutput.MaterialOutputModel import MaterialOutputModel
+
+
+class ExtruderOutputModel(QObject):
```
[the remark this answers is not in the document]
nallath: I just added them as they were needed.

nallath:
The view being broken is not due to these changes, it's due to the stages in cura not being implemented in the right way.

plugins/UM3NetworkPrinting/LegacyUM3OutputDevice.py
```diff
@@ -0,0 +1,627 @@
...
+        self._number_of_extruders = 2
+
+        self._authentication_id = None
+        self._authentication_key = None
+
+        self._authentication_counter = 0
+        self._max_authentication_counter = 5 * 60  # Number of attempts before authentication timed out (5 min)
+
+        self._authentication_timer = QTimer()
+        self._authentication_timer.setInterval(1000)  # TODO; Add preference for update interval
```
sedwards2009: Does anyone care about the TODO?

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: As far as I know, the reply that we get back isn't the same object by definition. This was the only way I could come up with would work.
You are indeed right that we can't first send a get to "zomg", which is sent to function 1 and then send a get to "zomg" and send it to function 2.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: It can be removed, as we never really use it. Wasn't sure if we needed to leave it as is or just empty.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: These were copied from the old stuff. But fixed them now.

fieldOfView:
> The view being broken is not due to these changes

The fact remains that the view works in the master branch, but not in yours.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: I'm open for suggestions.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: I couldn't find any guarantees in the documentation regarding this. As this was also how it was done in ye' old version, i thought it best to leave it as is.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
nallath: No; look at the name of the file. It only supports um3. This is also copied code from the old version.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: Except that in that case the next few lines won't work. So if you can think of a way to do this and be able to know the length of the result, feel free.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
nallath: How would that make it more readable? It just makes it harder to read imho.

plugins/UM3NetworkPrinting/DiscoverUM3Action.py
nallath: Same as last; splitting the code won't make it more reusable, but make changing it harder.

nallath:
True. But the extruder selection shouldn't even be there if you switch between the stages. These changes simply show that that didn't go well.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -178,7 +178,7 @@ def put(self, target: str, data: str, onFinished: Optional[Callable[[Any, QNetwo
         if onFinished is not None:
             self._onFinishedCallbacks[reply.url().toString() + str(reply.operation())] = onFinished
 
-    def get(self, target: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]):
+    def get(self, target: str, onFinished: Optional[Callable[[QNetworkReply], None]]) -> None:
```
nallath: This change is incorrect; the _onFinished functions need to be a callable with an any type as first param and a QNetwork reply as second.

nallath:
@ChrisTerBeke Is working on a fix for the monitor tab strangeness.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
nallath: Properties cant use notify from parent classes.

ChrisTerBeke:
I just pushed a possible fix to master branch (as it was a generic issue). The sidebar header component would always be visible, I've added a check to hide it when monitoring mode is active. It's a quick fix, later on we intent to pull all monitoring logic outside of the settings sidebar and have a completely separate QML file for the monitoring stage.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -178,7 +178,7 @@ def put(self, target: str, data: str, onFinished: Optional[Callable[[Any, QNetwo
         if onFinished is not None:
             self._onFinishedCallbacks[reply.url().toString() + str(reply.operation())] = onFinished
 
-    def get(self, target: str, onFinished: Optional[Callable[[QNetworkReply], None]]) -> None:
+    def get(self, target: str, onFinished: Optional[Callable[[Any, QNetworkReply], None]]) -> None:
```
sedwards2009: These changes are incorrect. The signature of a *bound* method in Python doesn't have the `self` parameter.
```
class Foo:
    def bar(self, some_int: int) -> None:
        # blah blah blah

my_foo = Foo()
bar_method = my_foo.bar
```

`bar_method` is now effectively a function which takes one parameter, an int. The `self` part of the original method is "pre-filled".

nallath: Except that all type hinting completely trip up if you don't set it like that, which defeats the purpose of the hinting. So we can write it correctly and train users to ignore the type hinting or write it incorrect and train people to trigger at the warnings.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
sedwards2009: The code sticks the `reply` in the value of the dict (actually in a tuple). There is no reason why you couldn't also use `reply` as a key too. `reply` is a real Python object and can do the same things, it just happens to have a reference to a C++ counter part.

fieldOfView:
It still looks rather broken to me. Here's an UM2+, not connected on the left, connected over USB to the right:
image

* The "The printer is not connected." label looks very out of place in the not connected state.
* When connected, the whole connected state is missing
* There's a ghost Extruder (probably caused by the "every single extruder printer is a multiextruder printer" rework, but I did not see it before)
* There's an empty button next to the Abort button.

I would not get the 🎉 out yet...

plugins/UM3NetworkPrinting/DiscoverUM3Action.py
sedwards2009: I think your comment landed on the wrong part of the page here.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
sedwards2009: Because then it would be divided into higher level blocks with give a higher level view/summary of the process. Also, when the reader wants to know the details of how a step is done they can see that smaller piece of code in isolation. Right now I have no summary, I have to read through the lines and figure out what each block is trying to do.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
sedwards2009: Which application is being tripped up? mypy should be fine with it.

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
nallath: I'm not a big fan of splitting functions like this up (I do think one should strive to keep functions short, but I don't really agree with the notion that long functions are bad per-se). I'll try to find a natural place to split it up.

resources/qml/PrinterOutput/HeatedBedBox.qml
```diff
@@ -0,0 +1,352 @@
...
+                }
+            }
+        }
+        Rectangle //Input field for pre-heat temperature.
+        {
+            id: preheatTemperatureControl
+            color: !enabled ? UM.Theme.getColor("setting_control_disabled") : showError ? UM.Theme.getColor("setting_validation_error_background") : UM.Theme.getColor("setting_validation_ok")
+            property var showError:
+            {
+                if(bedTemperature.properties.maximum_value != "None" && bedTemperature.properties.maximum_value <  Math.floor(preheatTemperatureInput.text))
```
sedwards2009: ouch. Do we really expose Python `None` as the string "None" and not null?

ChrisTerBeke: I guess that's PyQt's doing?

plugins/UM3NetworkPrinting/ClusterUM3OutputDevice.py
sedwards2009: thanks.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
sedwards2009: Ugly solutions to ugly problems deserve ugly names. I'm thinking of the dict being `_multiPartsWhichMustNotBeGCed` with a `_releaseReplyMultiParts()` method.

ChrisTerBeke:
Unfortunately the multi-buildplate merge caused some conflicts with this branch.

sedwards2009:
Names are hard but I really like "_kept_alive..." or "_keep_alive...".

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
nallath: pycharm, which runs mypy.

sedwards2009: Pycharm has its own checker, it doesn't run mypy. Mypy expects the first parameter to be omitted from the signature.

ChrisTerBeke:
* There's merge conflicts
* There are still UI issues as pointed out by Aldo.

When will these be solved?

cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -0,0 +1,101 @@
...
+        if self._time_total != new_time_total:
+            self._time_total = new_time_total
+            self.timeTotalChanged.emit()
+
+    def updateTimeElapsed(self, new_time_elapsed):
+        if self._time_elapsed != new_time_elapsed:
+            self._time_elapsed = new_time_elapsed
+            self.timeElapsedChanged.emit()
+
+    def updateState(self, new_state):
```
fieldOfView: This whole file lacks documentation, but can you please at least document what are the accepted states?

cura/PrinterOutput/PrinterOutputModel.py
```diff
@@ -0,0 +1,240 @@
...
+
+            if print_job is not None:
+                print_job.updateAssignedPrinter(self)
+            self._active_print_job = print_job
+
+            if old_print_job is not None:
+                old_print_job.updateAssignedPrinter(None)
+            self.activePrintJobChanged.emit()
+
+    def updateState(self, printer_state):
```
fieldOfView: Please document what are the accepted states.

ChrisTerBeke:
Unfortunately there's some conflicts again.

fieldOfView:
Shouldn't the plugin API be increased?

fieldOfView:
The Topbar icon does not change when the AcceptsCommands property of the active printer changes

fieldOfView:
The OctoPrintPlugin now also has a branch that at least works with the network rewrite:
https://github.com/fieldOfView/OctoPrintPlugin/compare/master...network_rewrite

At the moment it is still subclassing PrinterOutputDevice instead of NetworkedOutputDevice, but it is a working wip. One smallish thing I can't seem to figure out (yet) is why the topbar is not showing the correct status (it is always showing the "unknown" status icon, as if the outputdevice does not AcceptsCommands.

fieldOfView:
I don't have an UM3, so I cannot test this, but is it correct that there is a _printer_ state "idle" and a _job_ state "ready"? That's bound to lead to confusion (especially since the two states are so intertwined; there is a job state "offline", wtf?). Could we settle for one of the two?

resources/qml/PrinterOutput/ExtruderBox.qml
```diff
@@ -0,0 +1,201 @@
...
+            anchors.left: parent.left
+            anchors.top: parent.top
+            anchors.margins: UM.Theme.getSize("default_margin").width
+        }
+
+        Label //Target temperature.
+        {
+            id: extruderTargetTemperature
+            text: Math.round(extruderModel.targetHotendTemperature) + "°C"
+            //text: (connectedPrinter != null && connectedPrinter.hotendIds[index] != null && connectedPrinter.targetHotendTemperatures[index] != null) ? Math.round(connectedPrinter.targetHotendTemperatures[index]) + "°C" : ""
```
fieldOfView: Doh! Commented-out code!

resources/qml/PrinterOutput/ExtruderBox.qml
```diff
@@ -0,0 +1,201 @@
...
+                        base.hideTooltip();
+                    }
+                }
+            }
+        }
+        Label //Temperature indication.
+        {
+            id: extruderTemperature
+            text: Math.round(extruderModel.hotendTemperature) + "°C"
+            //text: (connectedPrinter != null && connectedPrinter.hotendIds[index] != null && connectedPrinter.hotendTemperatures[index] != null) ? Math.round(connectedPrinter.hotendTemperatures[index]) + "°C" : ""
```
fieldOfView: More commented-out code

resources/qml/PrinterOutput/ExtruderBox.qml
```diff
@@ -0,0 +1,201 @@
...
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+
+Item
+{
+    property alias color: background.color
+    property var extruderModel
+    property var position: index
+    //width: index == machineExtruderCount.properties.value - 1 && index % 2 == 0 ? extrudersGrid.width : Math.floor(extrudersGrid.width / 2 - UM.Theme.getSize("sidebar_lining_thin").width / 2)
```
fieldOfView: Commented-out code.

ChrisTerBeke:
@nallath can you check if my conflict fixing is correct?

ChrisTerBeke:
@nallath we now have 3.2 branches, so this can be merged to master when your team is done testing.

fieldOfView:
So this is not going to be part of 3.2 then?

ChrisTerBeke:
@fieldOfView no, we don't feel comfortable with that yet, as there are many other big changes in 3.2 as well that need system testing.

fieldOfView:
As an added incentive, here's a branch that adds hotend preheating on top of @nallath's rewrite:
https://github.com/Ultimaker/Cura/compare/network_rewrite...fieldOfView:feature_preheat_extruder
image

I'll make a PR out of it once the network_rewrite is merged.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
Ghostkeeper: You could sum up the lengths of the lines manually, like so (pseudocode-ish):

    batch_length = 0
    batch = []
    for line in self._gcode:
        if not self._compressing_code:
            # Stop trying to zip / send as abort was called. Bla bla.
        if batch_length + len(line) > max_chars_per_line:
            #Close up the batch.
            batched_lines = "\n".join(batch)
            byte_array_file_data += self._compressDataAndNotifyQt(batched_lines)
            batch = []
            batch_length = 0
        batch.append(line)
        batch_length += len(line)

resources/qml/PrinterOutput/HeatedBedBox.qml
Ghostkeeper: PyQt should convert it to `null`, I thought.

PR #3005, opened 2017-12-18
Prevent CrashHandler creating a QWidget before QApplication is created

fieldOfView:
This PR prevents crashhandler from crashing (and hiding the actual exception) when it is invoked before the QApplication has been initialized (eg when there is an syntax error in a py file).

```
QWidget: Must construct a QApplication before a QWidget
Fatal Python error: Aborted

Current thread 0x000012c8 (most recent call first):
  File "...\UM\Cura\cura\CrashHandler.py", line 56 in __init__
  File "cura_app.py", line 47 in exceptHook
```

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -53,7 +53,7 @@ class CrashHandler:
         self.exception_type = exception_type
         self.value = value
         self.traceback = tb
-        self.dialog = QDialog()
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Had this one, couldn't find out after why, thanks!

Patola:
Thank you!

PR #3009, opened 2017-12-18
Added infill_enable_travel_optimization setting.

smartavionics:
Enabling this can greatly reduce the travel time during the printing of infill but it can
be expensive to compute so now the user can decide whether to use it or not.


smartavionics:
Companion to https://github.com/Ultimaker/CuraEngine/pull/671.

jackha:
Devs, see: CURA-4725

PR #3015, opened 2017-12-19
Feature tree support review

jackha:
only for reviewing, will be removed after


resources/definitions/fdmprinter.def.json
```diff
@@ -5262,7 +5377,22 @@
                             "enabled": "support_enable and (support_pattern == 'zigzag') and support_skip_some_zags",
                             "limit_to_extruder": "support_infill_extruder_nr",
                             "settable_per_mesh": false,
-                            "settable_per_extruder": true
+                            "settable_per_extruder": true,
+                            "children": {
+                                "support_zag_skip_count": {
```
jackha: The setting is in itself?

Ghostkeeper: Eh, yes. I don't know why this branch now duplicates this setting?

fieldOfView:
this one can be closed now, right?

Issue #1969, opened 2017-06-12
Tab Through Fields


CCS86:
Hi guys, would it be a pain to allow tabbing through the setting fields? 

It would really be nice to tab through when setting up a print!

fieldOfView:
I agree that would be nice. I remember a couple of small complications from when I last tried to implement this. For example all the controls will need a "has keyboard focus" state. Not all controls respond to keyboard input at the moment (eg checkboxes, section headers).

DDDirk:
Keyboard control is more than just nice. At the moment my RSI is really bad, to the extent that it really hinders me in my work (reduced control over finger movements). Tabbing through the fields is just one thing. Simply using the menu in a standard way is something that would also help. (So all menus accessible with Alt-<key> and moving through the menu structure with the arrow keys.)

nallath:
This has been implemented.

CCS86:
@nallath , which version? 

I just installed 2.6.2, and don't have that functionality.

fieldOfView:
Cura 2.7, which has not yet been released. I expect a beta to be available in a week or so

CCS86:
Is it possible to change the behavior so that the current field value is selected when tabbing I to the field?

This would make it so much more efficient to tab through, quickly keying values, instead of having to delete every current entry value first.

fieldOfView:
I have made a PR (#3017) to implement this. It is something you can try/test for yourself by editing resources/qml/Settings/SettingTextField.qml (it is a single line addition)

CCS86:
Awesome, thanks @fieldOfView , that worked perfectly!

PR #3017, opened 2017-12-19
Select text when gaining focus

fieldOfView:
This PR changes the textfield to select its contents when gaining focus. This is especially usefull when tabbing through fields.

See https://github.com/Ultimaker/Cura/issues/1969#issuecomment-352632035

### DIFF
--- a/resources/qml/Settings/SettingTextField.qml
+++ b/resources/qml/Settings/SettingTextField.qml
@@ -17,6 +17,7 @@ SettingItem
     {
         textHasChanged = false;
         textBeforeEdit = focusItem.text;
+        focusItem.selectAll();
     }
 
     contents: Rectangle


Ghostkeeper:
Developers, see issue CURA-4764.

CCS86:
In testing this feature I am running into issues. 

I'm not sure the steps to repeat, as I just went to test and am not having the issue.

Basically, with some fields, if you start typing a new value, let's say line width. Cura will interrupt you and start slicing. I might try to type "0.45". Sometimes it takes the "0" and tries to slice. Sometimes "0.4". Sometimes the whole thing.

ChrisTerBeke:
Honestly the longer I use the settings after this change was made the more annoying it is to me. Often I just want to replace the last character of the text field instead of starting from scratch. Not sure how others feel about this?

fieldOfView:
The behavior is fairly consistent with how other applications work (at least on Windows). There's always the right cursor key to move the cursor to the end of the selected text before you start typing.

CCS86:
@ChrisTerBeke, I feel the same. At first, I thought I was just unlearning old habits. Now I actually think it's annoying.

My wish: 

-tabbing between fields would automatically highlight the text. 

-clicking into a field acted more like "normal" windows: single-click places a cursor, click and drag highlights a portion of text, double click highlights the entire value.

fieldOfView:
I have made a new PR that implements what @CCS86 suggests. See #3196.

PR #3030, opened 2017-12-22
Reduce amount of movements after print for Tevo Tarantula

TheAssassin:
At the moment, the base plate is moved to the end stops, which moves the head to the front of the base plate (at least for stock printers, but most likely for modified ones, too), making the object inaccessible. This has been fixed by moving the base plate to the back after this step. That eliminates the need for the X/Y homing, hence it got removed from the after print code.

Also, before these changes, the "steppers off" command was pretty pointless, as the final move was performed after this step. This is not necessarily a problem, but can be annoying when trying to move the base plate or extruder after the print by hand.

All in all some useful changes.

Owners of the printer should replace the "after print" code of their existing printers in their settings with this:

```
M104 S0 ;extruder heater off
M140 S0 ;heated bed heater off (if you have it)
G91 ;relative positioning
G1 E-1 F300  ;retract the filament a bit before lifting the nozzle, to release some of the pressure
G1 Z+0.5 E-5 X-20 Y-20 F9000 ;move Z up a bit and retract filament even more
G90 ;absolute positioning
G1 Y200 F3600 ;move extruder out of the way by moving the baseplate to the front for easier access to printed object
M84 ;steppers off

### DIFF
--- a/resources/definitions/tevo_tarantula.def.json
+++ b/resources/definitions/tevo_tarantula.def.json
@@ -66,7 +66,7 @@
             "default_value": "G21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 ;move X/Y to min endstops\nG28 Z0 ;move Z to min endstops\nG1 Z15.0 F9000 ;move the platform down 15mm\nG92 E0 ;zero the extruded length\nG1 F200 E3 ;extrude 3mm of feed stock\nG92 E0 ;zero the extruded length again\nG1 F9000\n;Put printing message on LCD screen\nM117 Printing..."
         },
         "machine_end_gcode": {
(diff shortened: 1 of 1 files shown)


thetobby:
@TheAssassin  Do you know if this could be relevant for Tevo Black Widow, just collected some community gcode? (I do not have that printer)  #3057

TheAssassin:
Hey @thetobby, sorry, I don't have access to such a printer. As the setup of such a printer looks familiar to the default Tarantula one (except for some additional stiffness due to all the additional profiles and probably better connectors), I'd say, it's worth giving a try. The question is, what does your printer currently do after print? My printers' default behavior was that it was returning to the X/Y end stops (a.k.a. home those axes), which moved the extruder to the front left corner (well, it moved the build plate, but the effect is the same). If you have some similar annoying behavior (moving the build plate to the front to be accessible should just be the default behavior, which was the main intent for me to write a custom config based on the Prusa i3 one, aside from having more precise time estimations), then the code might be useful for you.

thetobby:
Sadly I do not currently have access to Tevo ( or any 3D printers) either, so I guess it is best to not change the suggested Gcode of the Black widow (my pull request). Thanks for the reply! :)

nallath:
For our reference; CURA-4755

PR #3031, opened 2017-12-22
Feature headless

ChrisTerBeke:
This PR is a WIP for running Cura 100% headless. The goal is to separate all GUI related stuff and don't initialize those when running in headless (server) mode.

For now the PR can be reviewed. More things might be changed later when I'm ready to start building the Docker container.

Things this PR does:

* More all component creation of QML to a single method
* Split the run function for GUI and non-GUI mode, no QML things will start when running headless
* Add some optional flags to several methods to prevent UI interaction being needed (e.g. confirm dialogs)
* Add a new event when file loading is completed (including auto-centering and arranging)
* Add some debug logging for configuration changes which are also useful in normal mode
* Set convex hull color to a default value if QML and Theme are not enabled

The new [CuraCloudPlugin](https://github.com/Ultimaker/CuraCloudPlugin) makes use of these changes.


cura/CuraApplication.py
```diff
@@ -661,6 +661,10 @@ def run(self):
         self._volume = BuildVolume.BuildVolume(self.getController().getScene().getRoot())
         Arrange.build_volume = self._volume
 
+        # initialize info objects
```
ChrisTerBeke: +1

cura/CuraApplication.py
```diff
@@ -652,12 +653,47 @@ def preRun(self):
...
+        self._print_information = PrintInformation.PrintInformation()
+        self._cura_actions = CuraActions.CuraActions(self)
+
+        self._started = True
+        self.exec_()
+
+    ##  Run Cura without GUI elements and interaction (server mode).
+    def runWithoutGUI(self):
+        self._use_gui = False
+        self.closeSplash()
```
jackha: it seems that the way self._splash is assigned there isn't a splash screen when you arrive here. And closeSplash doesn't do anything.

fieldOfView:
The CuraCloudPlugin mentioned in the OP is private, I guess.

PR #3033, opened 2017-12-22
CURA-4708 Fix definition_changes containers for extruder stacks

LipuFei:
Please note that this doesn't fix the problems in the MachineSettings dialog. That will be done separately.


cura/Settings/CuraContainerRegistry.py
```diff
@@ -430,11 +430,32 @@ def addExtruderStackForSingleExtrusionMachine(self, machine, extruder_id):
...
+        definition_changes_id = self.uniqueName(extruder_stack.getId() + "_settings")
+        definition_changes_name = definition_changes_id
+        definition_changes = InstanceContainer(definition_changes_id)
+        definition_changes.setName(definition_changes_name)
+        definition_changes.addMetaDataEntry("setting_version", CuraApplication.SettingVersion)
+        definition_changes.addMetaDataEntry("type", "definition_changes")
+        definition_changes.addMetaDataEntry("definition", extruder_definition.getId())
+
+        # move definition_changes settings if exist
+        for setting_key in ("machine_nozzle_size", "material_diameter"):
```
Ghostkeeper: I think it's better to let this depend on the `settable_per_extruder` property of the setting. If it's settable per extruder it should be in the extruder stack, right?

If this is not possible due to certain settings that are an exception, we might make a special case for the settings that shouldn't be moved. If that gets too much and it's all impossible, at least document why these two settings are special.

ChrisTerBeke:
There are conflicts now. Also are the new changes solving the reason why the PR was denied?

PR #3056, opened 2017-12-28
duplikate

thetobby:
I am apparantly blind, new pull request on the way.


thetobby:
Regarding monoprice:  Similar hardware to Malyan M180, but the other hardware has very minimal profile and no mesh exists #2874  I suggest adding this as one and depreciate m180.def.json or consider merging.

PR #3058, opened 2017-12-29
Updating to reflect use of "Marlin" instead "RepRap (Marlin/Sprinter)"

thetobby:
GUI is showing and currently defaulting to Marlin in _MachineSettingsAction_ 
This had my tricked for a good while while trying to debugging my printer config as to why it did not show _RepRap (Marlin/Sprinter)_ as my and others definition says. 

I have updated all the remaining configuration and GUI with translation to reflect _Marlin_ instead of _RepRap (Marlin/Sprinter)_ so everything is in sync.


ChrisTerBeke:
As this changes an actual stored value there is also a change required in the version update scripts so previously stored machines don't break after this change.

thetobby:
Thanks for your reply @ChrisTerBeke  I am just starting and do not know any advance programming yet, out trying to help with what I can. Sadly creating a version update script for this is beyond my current capabilityes.  Is there anyone else who can make this?

ChrisTerBeke:
We probably can, but I don't think it's high on our prio list as keeping the old value (and using that in your own definition as well) is simpler and less error prone.

thetobby:
May I temporarly change the Gui back to reflect the old name _RepRap (Marlin/Sprinter)_ until such efforts can be made? I think the current name is confusing unless we are doing this whole name change everywhere. With the current gui use of _Marlin_ most user manuals shows wrong compared to what the user is presented with in the GUI.  Meaning the "name change" damage is already done for the users.

ChrisTerBeke:
Personally I think it's weird anyways that the actual stored value (setting name) is the same as the pretty name for the UI, but that's something that's hard to change now.

fieldOfView:
Unless there is a matching change in CuraEngine, changing the key will break things quite badly. Changing the display string through the translation (in english only) will only be more confusing. 

The user-facing "RepRap (Marlin/Sprinter)" string was changed for a reason; it is incorrect and confusing (@smartavionics can tell you more about that).

smartavionics:
Hi, yes, to better support real RepRap firmware which is used on machines with Duet controllers I added the RepRap UI string so the user can choose between Marlin and RepRap (and others, of course). I did this without changing the setting keys so that existing configs would not be broken. I appreciate that the key strings are not very good now but without updating existing configs I don't think they can be changed.

thetobby:
This PR is bad and I will close it, thanks for the explanation @fieldOfView  
@smartavionics I think we should keep using the old user-facing "RepRap (Marlin/Sprinter)" string since most user manuals are written with that in mind and personally I should the UI should match the config. I made a PR to change it back. Please continue discussion in #3060 if it is up to discussion :)

fieldOfView:
> I think we should keep using the old user-facing "RepRap (Marlin/Sprinter)" string since most user manuals are written with that in mind

I think the manuals should be updated.

> personally I think the UI should match the config

I agree to an extend, but this is very low priority because the issue is not user-facing. Making the change has a high probability of messing things up.

thetobby:
> personally I think the UI should match the config
> I agree to an extend, but this is very low priority because the issue is not user-facing. Making the change has a high probability of messing things up.

Changing config to match UI has been proved to be difficult in this discussion.
But changing back the UI text to match config should be pretty easy as far as I can tell #3060  should be enough? So the main/only question is should we have the name change in UI or not not. Most manuals I find online are based on older versions. Offcourse everyone should update their manuals, but why make life difficult unless there is a compelling reason to use a different name in UI than config.

fieldOfView:
If a program must always follow the documentation of older versions, bugs can never be fixed. We should not put an error back in "until the guides are updated", because when we put that error back in the manuals/guides will never be updated. I vote against #3060

smartavionics:
Here's my take on this: The firmware flavour setting choices each have a label like string that is visible to the user in the UI and a non-visible key string that is used to identify the setting value to the engine. 

Ideally, the label and the key would use similar words (i.e. for the Marlin flavour gcode, the key should be Marlin). However, due to the nature of software development the key and label strings have now diverged somewhat.

It's easy to change the label strings because such changes only affect what is visible in the UI (ignoring translation to other languages) but changing the keys is difficult. Not because the engine code needs changing, that's probably a 5 minute job but because of all the machine defs that contain sections like this example from one of my own files:

````
    "overrides": {
        "machine_gcode_flavor": {
            "default_value": "RepRap (RepRap)"
        },
       ...
    }
````

OK, the machine defs that are packaged in the release could all be changed to use new key values but not the users' machine defs.

Is that right? If so, I don't think the keys can be changed now without causing much grief.

thetobby:
@smartavionics if i understand you correctly perhaps adding support for both old and new key in engine for a transition period could would be an option?

smartavionics:
Hi @thetobby, the engine could easily support multiple keys that represent the same gcode flavour so you could introduce a new tag/label pair that, for example, produces the same gcode as the existing Marlin flavour. Whether that is actually beneficial and whether such a scheme would be "blessed" by the Cura guardians is not clear.

Ghostkeeper:
It's unnecessary to update the translation files. This is done automatically after the string freeze (around the release of the beta).

The change of the key in fdmprinter requires:
* A version upgrade entry. This is a bit of work but easy.
* A change in CuraEngine. This is a one-liner then; just change the string.

Leaving both options for the key in only introduces more legacy code structures that'll haunt us forever, I predict.

But what about Sprinter firmware though? The g-code flavour in question should also support Sprinter, so that would be unclear for Sprinter machines. I don't really agree with the change as a whole, actually.

PR #3059, opened 2017-12-29
Revert to use of RepRap (Marlin/Sprinter) in English GUI

thetobby:
Untill full changover from RepRap (Marlin/Sprinter) to Marlin is ready in #3058


thetobby:
Did a mistake with the pull request

PR #3060, opened 2017-12-29
Revert to use of RepRap (Marlin/Sprinter) in English GUI

thetobby:
Untill full changover from RepRap (Marlin/Sprinter) to Marlin is ready in #3058 
Much more userfriendly and the naming will be correct with most user manuals published

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -302,7 +302,7 @@
                     "type": "enum",
                     "options":
                     {
-                        "RepRap (Marlin/Sprinter)": "Marlin",
+                        "RepRap (Marlin/Sprinter)": "RepRap (Marlin/Sprinter)",
                         "RepRap (Volumetric)": "Marlin (Volumetric)",
                         "RepRap (RepRap)": "RepRap",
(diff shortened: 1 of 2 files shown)


nallath:
For our reference; CURA-4756

ChrisTerBeke:
I don't see why the UI label and the actual in-code value must be the same here?

thetobby:
@ChrisTerBeke 
Someone has changed the UI label from _RepRap (Marlin/Sprinter)_ til _Marlin_ I wish to revert that.

It has been discussed in more detail in #3058 but basically, my argument is that having Marlin in just the UI label has a negative effect as most user manuals and peoples experience are based from the old label. I do not see any reason to change the UI label in the first place to _Marlin_  from the in-code value. Personally, I actually though my printer profile was bugged until I confirmed that the UI has another display name than in-code value. So this is a suggestion to revert to same name in UI as in-code/older Cura.  Alternative #3058  is to change all in-code values to Marlin, if this is indeed the correct name to use. The counter-arguments from fieldofview is that people should update their manuals to Marlin, problem solved. I agree if there is a reason it should be named Marlin in UI and therefore offered both #3058  should this PR be wrong.

PR #3063, opened 2017-12-30
Remove ID from definition files

thetobby:
Finishing the cleaning up like #2867 did.
Contributes to issue CURA-4243.


ChrisTerBeke:
Makes sense to remove these indeed, it'll prevent confusion.

Ghostkeeper:
They were removed everywhere, but since then new definitions have been merged that had the ID again. We should pay attention before merging them.

PR #3068, opened 2017-12-31
Multiextruder gcode snippet replacement patterns

fieldOfView:
This PR allows specifying the extruder stack to use for each replacement pattern. The replacement pattern is (optionally) extended with an extruder_nr: {setting_key,extruder_nr}. The extruder_nr can either be a number (-1 for the global stack, 0 or higher for an extruder), or a setting keyword such as support_extruder_nr etc. If an extruder_nr is not specified, the value is taken from the same stack as the current implementation.

* ```{material_standby_temperature}``` in start gcode gets the (resolved) setting value from the default stack (the first used extruder), as before
* ```{material_standby_temperature, 1}``` gets the value from the 2nd extruder (extruder nrs start by 0)
* ```{material_standby_temperature, support_extruder_nr}``` gets the value from whatever extruder is set as support extruder

Contributes to #1296 (see https://github.com/Ultimaker/Cura/issues/1296#issuecomment-354556760 and on)

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -258,6 +258,10 @@ class StartSliceJob(Job):
         result["date"] = time.strftime("%d-%m-%Y")
         result["day"] = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"][int(time.strftime("%w"))]
 
+        initial_extruder_stack = Application.getInstance().getExtruderManager().getUsedExtruderStacks()[0]
+        initial_extruder_nr = initial_extruder_stack.getProperty("extruder_nr", "value")
+        result["initial_extruder_nr"] = initial_extruder_nr
+
         return result
 
     ##  Replace setting tokens in a piece of g-code.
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Good solution. No regression as the default behaviour stays the same, and adding an extruder index as stack indicator seems like the simplest solution to me.

stuartpb:
I want this functionality more than I care about having it implemented the particular way that I'm describing it, so I don't want this to be a blocker without an alternative implementation to pull, but I *really* think this should be done by importing all variables from all stacks with prefixes (either using named prefixes or nested replacement for the role-based stack selection) instead of introducing a compound replacement construct, as I described [here](https://github.com/Ultimaker/Cura/issues/1296#issuecomment-354571388).

The simple answer as to why I think it should work this way is that [this is how Simplify3D does it](https://forum.simplify3d.com/viewtopic.php?f=8&t=1959&p=7574) (numbered prefixes and role variables), it's [how Slic3r does it](https://github.com/alexrj/Slic3r/wiki/FAQ) (numbered suffixes and nested role variables), and it's [how MatterControl does it](http://wiki.mattercontrol.com/Custom_G-Code_Variables) (role prefixes). There are enough translation headaches involved in developing profiles across different slicer environments, and it'd be nice to have one less pointlessly divergent thing to deal with.

nallath:
For our reference; CURA-4757

PR #3075, opened 2018-01-01
Add option to UM3NetworkPlugin to allow preventing print queue buildup

sameer:
## Use case
We have a problem in our print shop where users will accidentally send a print over network to a printer that's already printing. 

When we were using USBs, there was a physical barrier to sending prints. Now, because prints can be queued when sent over network, our users don't realize that they are queuing prints to a printer that's already printing. They think the print failed so they try to send it to another printer, which is problematic.

## What this PR does
Added an option in the cfg file to prevent queuing of network prints on Cura. It's set to False by default, so it will not affect normal usage.

When set to True, if the print is not finished, it will prevent the user from sending a print and tell them that  a print is already in progress. I did partial translation (German and French).

### DIFF
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevice.py
@@ -865,6 +865,14 @@ class NetworkPrinterOutputDevice(PrinterOutputDevice):
             self._error_message.show()
             return
 
+        # Check if the printer is already printing, if user wants to prevent queuing, stop queuing.
+        if not self._print_finished and self._prevent_queue:
+            self._error_message = Message(
+                i18n_catalog.i18nc("@info:status",
+                                   "Sending new jobs blocked, a print job is currently in progress."))
+            self._error_message.show()
+            return
+
         # Indicate we're starting a new write action, is set back to True at the end of this method
         self._write_finished = False
 
--- a/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevicePlugin.py
+++ b/plugins/UM3NetworkPrinting/NetworkPrinterOutputDevicePlugin.py
@@ -54,6 +54,8 @@ class NetworkPrinterOutputDevicePlugin(QObject, OutputDevicePlugin):
         self._preferences = Preferences.getInstance()
         self._preferences.addPreference("um3networkprinting/manual_instances", "") #  A comma-separated list of ip adresses or hostnames
         self._manual_instances = self._preferences.getValue("um3networkprinting/manual_instances").split(",")
+        self._preferences.addPreference("um3networkprinting/prevent_queue", False) # Whether to prevent queuing of prints over network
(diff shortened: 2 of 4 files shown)


ChrisTerBeke:
As we're doing a complete rewrite of the networking (and other output devices) I think this would have to be rebased after we have merged that. CC @nallath

ChrisTerBeke:
Also, the legacy UM3 doesn't support queueing, so that wouldn't be a problem to begin with. Cura Connect (new UM3 networking solution) is supposed to work as queue, so changing the behaviour for that would be unexpected.

fieldOfView:
> Cura Connect (new UM3 networking solution) is supposed to work as queue, so changing the behaviour for that would be unexpected.

I guess it would be nice to be able to have a choice.

ChrisTerBeke:
As Cura Connect is quite an important new asset for Ultimaker, I doubt that product level people will agree with that. I'll bring it up in the stand-up though and report back in this thread.

ChrisTerBeke:
Discussed with several devs and product owner, and we don't feel like this is something Ultimaker wants to add as it 'disabled' the primary feature of Cura Connect.

fieldOfView:
One could argue that it 'enhances' the primary feature of Cura Connect for more environments. But I don't have a Cura Connect enabled printer either way...

sameer:
So is there absolutely no way to prevent queuing when printing over network?

Our print shop is in a university and offers free printing, so we operate on a "first come first served" model. We tried queuing before but the queue grew to be 3+ days long and was really problematic.

We need to prevent queuing somehow, preferably without returning to USBs.

ChrisTerBeke:
In the web interface of Cura Connect you can re-order the queue (place on top) and remove jobs. I don't see how that functionality doesn't allow you to prevent getting a huge queue which must be printed entirely before reaching the desired print job.

fieldOfView:
The problem, as outlined in the OP, is that there is no flood control. Especially in places such as schools and fablabs, users can now send copious amounts of print jobs, essentially launching a denial of service attack on the printer(s) because each print job has the potential to run for many hours, if not days.

sameer:
@ChrisTerBeke Sorry, the issue is not that the queue is getting too large and one print job needs to be completed before the others. The issue is that since we offer *free* printing to university students/professors, the queue quickly fills with prints, leaving them constantly running. 

With queuing, we had issues where students were printing things like Pokemon sculptures which, while cool, are hard to justify for filament purchase orders. A lack of queuing has forced the prioritization of prints that are important.

What @fieldOfView said is also true. There are 3 UM3s in our shop and we would like to avoid wearing them down quickly.

ChrisTerBeke:
@sameer thanks for clarifying your problem. Indeed that's a use case that Cura Connect does not support at the moment, but we are thinking about ways to fix that that fit into our software vision (besides disabling queueing all together) since we had multiple requests to support different organizational structures. I can't go into much details right now as we're still figuring out what cases to support and how. We'll use your comments as input for that internal conversation.

sameer:
@ChrisTerBeke thank you, I really appreciate it!

PR #3082, opened 2018-01-03
QtQuick controls 2.0 for settings sidebar

ChrisTerBeke:
This PR updates the slower UI parts to QtQuick 2.0. We're still ironing out some details and probably leave other parts of the UI alone for now (as updating all dialogs etc. is a lot of work to do at once and don't give any visible improvements).
 
Next to speed improvements, QtQuick controls 2.0 drops support for theme styles, so we can't use the QML file in the theme resources directory to style generic components.


awhiemstra:
> Next to speed improvements, QtQuick controls 2.0 drops support for theme styles, so we can't use the QML file in the theme resources directory to style generic components.

This is not strictly true, you can set the environment variable " QT_QUICK_CONTROLS_STYLE" to a path containing a custom style. Assuming the environment variable gets set before creating controls, it should be possible to point this at a directory in the theme. See http://doc.qt.io/qt-5/qtquickcontrols2-styles.html for details about the styling mechanism of QQC2.

ChrisTerBeke:
But then you can only set 1 "global" style?

I personally think it's cleaner to move the styles into re-usable components (button component, extruder dropdown component, etc) in the resources/qml directory and just load those components in the rest of the UI.

LipuFei:
@ChrisTerBeke I think we can use higher QtQuick versions in QML files

image

PR #3089, opened 2018-01-05
Support Mesh enhancements

fieldOfView:
This PR provides the following enhancements to support meshes:
* Allow adding (only) support-specific settings for support meshes in Per Model Settings
* Render support meshes with a vertical stripe to tell them apart from normal meshes
* Hide the Support Mesh/Anti overhang mesh/... option if the machine definition disables the settings

This PR is a continuation of #2758


nallath:
CURA-4775

ChrisTerBeke:
@alekseisasin relevant for your bugfix ticket?

LipuFei:
Hi @fieldOfView , there is a bug in the engine. It crashes when "Print as support" is enabled for a model. We will test this PR after that's fixed.

PR #3114, opened 2018-01-09
Fix updating icon in the top bar

fieldOfView:
This PR makes sure events such as onAcceptsCommandsChanges get connected if an outputdevice has been added before the monitorstage is initialized. Without these connections, an outputdevice that uses eg setAcceptCommands to change its state will not show an updated icon in the topbar. 

This PR applies to #2974


fieldOfView:
@nallath, this is for you

ChrisTerBeke:
Seems useful.

PR #3123, opened 2018-01-11
Added SeeMeCNC printers w/ Reprap firmware

pouncingiguana:
Definition and platform files added for 2 new SeeMeCNC Printers:

SeeMeCNC Artemis: seemecnc_artemis.def.json, artemis_platform.stl
SeeMeCNC Rostock Max V3.2: seemecnc_v32.def.json, rostock_platform.stl


nallath:
For our reference; CURA-4818

Ghostkeeper:
The Rostock platform file is a bit heavy. Our guideline is to not go above 1MB per platform file. Could you trim it down a bit?

Or maybe you'd like us to do that for you? I usually use Blender's Decimate filter.

pouncingiguana:
If you're able to knock down the size, please do. Thanks!

Ghostkeeper:
I did a decimate in Blender with 60%. It ended up at about 800kB. I wasn't able to see any difference at all, but I think it probably reduced the roundness of the screw holes and such at the bottom.

PR #3134, opened 2018-01-15
Fix for CURA-4789

ianpaschal:


Ghostkeeper:
For posterity, CURA-4789 was a bug where the build plate was not cleared when loading a project file.

fieldOfView:
Posterity thanks you

PR #3137, opened 2018-01-16
Feature: Printers with different build plates

diegopradogesto:
This PR adds the possibility to set different build plates into a printer and modify settings depending on that.


resources/definitions/fdmprinter.def.json
```diff
@@ -154,6 +154,21 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_buildplate_type":
+                {
+                    "label": "Build Plate Material",
+                    "description": "The material of the build plate installed on the printer.",
+                    "default_value": "glass",
```
ChrisTerBeke: this means the engine G-code will now always output build plate type "glass" as comment even though most 3rd party printers use another type of plate?

resources/qml/SidebarHeader.qml
```diff
@@ -314,6 +314,62 @@ Column
...
+        height: visible ? UM.Theme.getSize("sidebar_lining_thin").height / 2 : 0
+        color: UM.Theme.getColor("sidebar_lining_thin")
+    }
+
+    //Buildplate row
+    Item
+    {
+        id: buildplateRow
+        height: UM.Theme.getSize("sidebar_setup").height
+        visible: Cura.MachineManager.hasVariantBuildplates && !sidebar.monitoringPrint && !sidebar.hideSettings
```
ChrisTerBeke: This will be refactored in the machine manager refactor I guess? (e.g. globalVariants and extruderVariants?)

resources/definitions/fdmprinter.def.json
diegopradogesto: Yes, that is right. It it is worthy, the default value can be changed.

resources/qml/SidebarHeader.qml
diegopradogesto: Yes, probably.

cura/Settings/MachineManager.py
```diff
@@ -1222,6 +1267,53 @@ def hasVariants(self) -> bool:
             return Util.parseBool(self._global_container_stack.getMetaDataEntry("has_variants", False))
         return False
 
+    @pyqtProperty(bool, notify = globalContainerChanged)
+    def hasVariantBuildplates(self) -> bool:
+        if self._global_container_stack:
```
LipuFei: I think this is supposed to be:
```
if not self._global_container_stack:
    return False
```

LipuFei: Ah, sorry. The next line is split into two on GitHub review... Ignore this one.

PR #3142, opened 2018-01-17
Updated definition and added the qualities

velbn:
Hi

We've updated the definition of our Vertex Delta 3D printer and also added the qualities.
We've also added our materials to this repo => https://github.com/Ultimaker/fdm_materials

Thanks!


nallath:
For our reference; CURA-4819

resources/definitions/vertex_delta_k8800.def.json
```diff
@@ -1,12 +1,15 @@
 {
+  "id": "K8800",
```
Ghostkeeper: Remove this ID. We're not adding that here any more.

velbn:
Ok done

PR #3144, opened 2018-01-17
Feature preheat extruder

fieldOfView:
This PR adds per-extruder preheat buttons to the print monitor. Preheating works the same as bed preheating; a timer is started to cool the heaters after a certain period, to prevent material degrading in an idle hot extruder.

image

The code is tested on USB connected printers, as well as OctoPrint.

Update:
The PR also includes some fixes for the network_rewrite:
* Bed preheat button did not disable when a print is started
* Tooltips were not pixelaligned, leading to bad text rendering
* The end of a (bed) preheat period could set the temperature to 0 while printing


ChrisTerBeke:
Cool!

fieldOfView:
After #3144 (extruder preheating), the USBPrinterOutputController and the OctoPrintOutputController are 99.9% identical; the only difference is the class name. I would think it would be a good thing to create one common functional controller that can be shared between USBPrinter, OctoPrintPlugin and all the derivatives of OctoPrintPlugin (there's a Repetier Host variation floating around), as well as a future version of the Doodle3d output device.

What say you I move the USBPrinterOutputController from the USBPrinter plugin to cura.PrinterOutput.GenericOutputController ?

ChrisTerBeke:
Having a basic/generic output controller would be nice I guess, other output devices can inherit it or at least copy it as an example.

fieldOfView:
> Having a basic/generic output controller would be nice I guess

That's all I needed to hear ;-)

nallath:
Was this added to JIRA?

fieldOfView:
Subtle bump

Ghostkeeper:
No, it wasn't added yet. It is now.

Developers, see issue CURA-4949.

hashtagdavi:
Hi. Ive been searching for a way to enable this. not having the ability to preheat extruder like you can with the heat bed is very frustrating. In monitor tab, either with or without octoprint, there is no option to do so with Cura. 

is this a fix that your working on , which can give/enable the ability to preheat extruder(s)?! as i have read, i seem to understand that it is?

 If that is so, is it possible to use the fix, and if so, how?

please and thank you! any help would be greatly appreciated.

fieldOfView:
Fixed conflicts with master.

@hashtagdavi, this is not something that is easily added to Cura 3.2.1. You will have to wait until a release with this functionality is made, or run Cura from source (see the wiki), though the latter is not very straightforward to set up.

alekseisasin:
@fieldOfView I have a question, why did you move functionality from USBPrinterOutputController to GenericOutputController ?

ChrisTerBeke:
@alekseisasin We forsee multiple plugins/output devices needing the generic controller functionality. Right now the USB output controller, UM3 networking controller and octoprint controller are largely the same code. By making a generic one that contains all default features (controls, pre-heat, status updates, etc) we can re-use that logic. It also makes refactorings or updates easier.

ChrisTerBeke:
(in the future we might wanna re-write the other output controllers to also use the generic one as base)

alekseisasin:
If we are talking about Generic type then _now/later I _ would expect that all printers inherit it's methods. For now Generic is a specific printer controller, same as ClusterUM3PrinterOutputController.
Having GenericController is fine, I don't have nothing against it, only later I assume we will create again extra USBPrinterOutputController with some unique methods which we don't need for ClusterUM3PrinterOutputController.

alekseisasin:
Anyway, I tested this solution for UM2 and it works fine.

ChrisTerBeke:
USBPrinterOutputController and ClusterUM3PrinterOutputController could then inherit GenericOutputController, but if we need to make changes for all those can be done in the generic one. That's just a common software design pattern.

alekseisasin:
Yep, this was my question. We could steel keep USBPrinterOutputController and Generic. But now we do not have USBPrinterOutputController.

fieldOfView:
OctoPrintPlugin, and other plugins that derive from it, would use the exact same controller as USBPrinting.

fieldOfView:
See https://github.com/Ultimaker/Cura/pull/3144#comment-358608247

fieldOfView:
OctoPrintPlugin now also uses the GenericOutputController (without needing to subclass it):
https://github.com/fieldOfView/OctoPrintPlugin/commit/d067de44a9743fcba19499e1de3010dc9d513db6

PR #3150, opened 2018-01-18
Create ChangeFilamentAtLayer.py

A-J-Bauer:
A post processing script for Cura. Makes it easy to change the filament at a given layer before this layer is printed.
The Marlin command M600 is used which requires an LDC display.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/ChangeFilamentAtLayer.py
+++ b/plugins/PostProcessingPlugin/scripts/ChangeFilamentAtLayer.py
@@ -1,0 +1,118 @@
+from ..Script import Script
+from UM.Logger import Logger
+from PyQt5.QtWidgets import QMessageBox
+
+# Written 2018 by A.J.Bauer
+#  - Tested with Cr-10 printer using stock firmware, gcode flavor in Cura set to "Marlin"
+#  - M600 since Marlin 1.1.0, requires LCD display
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Seems like this does the same thing as the ColorChange script we already have. What's the difference?

A-J-Bauer:
The layer actually corresponds to the layer as seen by the user in the Simulation / Layer View.

The ColorChange script assumes that the layer can be found by data[ layer_num - 1 ], that is by using an index on the the data structure. But several tests showed that this doesn’t always correspond to the current layer. That’s why I calculate the layer at which to insert the M600 Marlin command by looking at the G commands z coordinate and taking into account that the first layer height can be of different height.

Ghostkeeper:
Perhaps we can then just edit the current script. We should also probably rename it...

A-J-Bauer:
In the meantime I changed the script because I needed to print with different materials (PETG on top of PLA). Now the script has just 4 input parameters ("Change at layer", "Print temperature", "Heat bed temperature" and "Change pos relative Z"). The name is ChangeFilamentAtLayer to indicate that a different filament can be used at a given layer. If interested let me know and I create another pull request.

Ghostkeeper:
If it works better or easier than the previous version, I'm interested to see it in action :)

PR #3154, opened 2018-01-19
Enable setting per mesh group settings when printing One at a Time

fieldOfView:
This PR fixes Per Model Settings to allow adding settings such as Print Temperature when the Print Sequence is set to One at a Time. When printing One at a Time, the Per Model Settings apply to mesh groups. 

Lots of settings that cannot be set on a per-mesh basis can be set on a per-meshgroup basis.

* Set Print Sequence to "One at a time"
* Check the settings that can be added to a model using Per Model Settings

Expected:
Printing temperature can be set

Actual:
Printing temperature is not listed as a per model setting that can be added.


fieldOfView:
Not sure why this makes a test fail

Ghostkeeper:
Developers, see issue CURA-4921. Oooh, we're approaching 5000!

fieldOfView:
Well, I still have 2 open PRs that don't have a JIRA issue attached to them, so I am doing all I can to reach that 5000 ;-)

alekseisasin:
@fieldOfView , Do I understand correct using your PR, I should get extra setting in list like: Printing temperature. If I load any model and check the setting list I still do not see it. I expect to find under Material group, but only listed: Flow and Initial Layer Flow.

fieldOfView:
@alekseisasin: Did you set Print Sequence to "One at a Time"? 
What printer do you use?

alekseisasin:
I followed next steps: Added Um2, in special adjusted "Print sequence" to One at a Time, loaded simple model and I expected to find "Printing temperature" in per model settings.

fieldOfView:
The UM2 has no printing temperature in Cura, unless you change the gcode flavor. This PR does not intend to change that; that's the nature of Ultigcode.

alekseisasin:
Ok, then how I can set set Print Sequence to "One at a Time", so far I see this settings becomes available if "enabled": "extruders_enabled_count == 1",

fieldOfView:
Add any other single extrusion printer than the UM2 or UM2+.

Ghostkeeper:
I think it's broken in the meanwhile. The temperature settings are really not visible.
screenshot from 2018-03-15 10-34-50

fieldOfView:
I'll check it out.

fieldOfView:
It turns out searching for the temperature was resetting the filter to not showing the additional settings. This is fixed now.

plugins/PerObjectSettingsTool/PerObjectSettingsPanel.qml
```diff
@@ -395,11 +394,33 @@ Item {
...
+                new_filter["settable_per_meshgroup"] = true;
             }
+            else
+            {
+                new_filter["settable_per_mesh"] = true;
+            }
+
+            if(filterInput.text != "")
+            {
+                new_filter["i18n_label"] = "*" + filterInput.text;
```
Ghostkeeper: Doesn't there also need to be an asterisk after the input? Like:

    "*" + filterInput.text + "*"

fieldOfView: SettingView and SettingVisibilityPage both filter with a single prepended "*".

LipuFei:
Hi @fieldOfView , FYI, certain per-object settings are not working in one-at-a-time mode. Here is a list of what we have found out so far:
- Initial layer height
- Printing Temperature Initial Layer and possibly other temperature settings
- **Line width works** somehow

This is a bug in CuraEngine. For example:
- see https://github.com/Ultimaker/CuraEngine/blob/master/src/commandSocket.cpp#L329, the mesh group settings are probably ignored by the engine
- also check https://github.com/Ultimaker/CuraEngine/blob/master/src/Preheat.cpp#L17, the temperature settings are only fetched from the extruder trains, not from the objects

Note that this is not a complete list of buggy files related to this bug.

fieldOfView:
A quick fix for 3.3 would be to set the settings that don't work to ```"settable_per_meshgroup": false``` in fdmprinter.def.json.

PR #3175, opened 2018-01-24
Add bridge settings to experimental category.

smartavionics:
Not yet a merge candidate, just making this visible so that people can try it out, post comments, etc.


smartavionics:
Hi Team, this has now matured sufficiently that it could be worth evaluating. Enjoy!

Engine PR is https://github.com/Ultimaker/CuraEngine/pull/688.

Ghostkeeper:
Developers, see issue CURA-67. (Whew, that's an old one!)

smartavionics:
I'll just say a few words about the latest addition to the settings (bridge_wall_coast). The idea behind this setting is that it gives the user a way of controlling how far the nozzle should coast immediately before it prints a bridge wall line. Why is the coasting required? Well, unless the printer firmware implements some kind of extruder pressure lookahead (i.e. pressure advance on the Duets), what happens is that the nozzle zooms up to the start of the bridge section at normal wall speed/extrusion rate and then as it goes past the edge and starts to slow down the speed/flow there's still sufficient pressure in the nozzle to  push the filament way down below the bridge level. By coasting before the bridge is reached, the pressure in the nozzle is already dropping before it reaches the bridge section and the bridge line comes out flatter. The implementation tries to take into account the wall print speed and flow and calculates a coasting distance but because the amount of coasting required can vary depending on printer hw/fw, settings, etc. that's why this new setting has been created. Setting it to 0 will completely disable the coasting.

ianpaschal:
I'm not able to get the intended results. I thought I was getting it with the sides but then tested by rotating the model and adding more "legs" and bridges.

screen shot 2018-02-28 at 12 41 16

Am I missing something here? If using the coasting settings is correct then maybe I need to double check that I'm using the right (your) build of CuraEngine.

smartavionics:
Hi Ian, it's a bit hard to tell from your image but I can see the coasting happening on the walls just to the right of the leg nearest the bottom. If you set the bridge wall/skin flows to something really small (i.e < 30) then it becomes obvious in the layer view as to which is bridge and which isn't. Also, you can use the feedrate display mode and then the bridge sections should be obviously slow.

BTW, just a few minutes ago I committed a bug fix.

smartavionics:
I haven't made any changes to the code that that computes the best angle for the bridge skin lines, that's still original.

Can you post the model, so I can play with it?

smartavionics:
Update: that bug fix was in the engine, not the UI.

ianpaschal:
[This is my test model I made.](https://www.dropbox.com/sh/tuhzu6fordhaatz/AACQw0xD9J1TIGaEca39fikwa?dl=0)

smartavionics:
The problem was that the skin over the 4 bridge regions was one area so it could not have different directions for the 2 pairs of sub-areas. By reducing the amount of bottom skin expansion to 0.5mm, I got it to do the right thing in terms of skin line directions:

screenshot_2018-02-28_12-54-02

ianpaschal:
Ah that makes sense. That would explain why before lunch I got it to work using only 2 legs.

I will continue testing later this afternoon and tomorrow morning (picked up something else for right now).

smartavionics:
Hi @ianpaschal, I am making a few changes to the bridging code which should be done by Monday. Could you please continue your testing using the updated PRs next week? Thanks.

smartavionics:
Hello Ian, I have made quite a few changes to the engine code and the settings. Please resume your testing after updating to the latest revisions.

I spent the weekend coding and printing what seems like hundreds of bridges! I have learnt that there's a lot more to it than simply tweaking the speed/flow/fan of the first layer above the air. Therefore, I have added the ability to specify the speed/flow/fan for the next skin layer because if you print that layer using the normal skin settings, the bridge can sag more.

My testing is done using PLA at 200deg, 0.4mm nozzle, 0.2mm layer height, 0.5mm walls.

My current strategy is to print the walls and the skin of the first layer very slowly (15mm/s) with full fan and 50% flow. This produces nice tight thin extrusions across the gap. On the next layer, the walls are printed normal speed (this could be changed but I'm not sure if it would help). The skin is printed slowly again with 125% flow at 90 deg to the bridge direction with 0% fan. This makes a relatively strong layer. The reason the fan is 0% is because when the fan is on it blows the air around the nozzle directly onto the thin first layer extrusions and they tend to sag immediately before the second layer is printed which doesn't help. If a third skin is to be printed, it will use the normal speed/flow/fan but the direction is aligned with the bridge. Any further skin layers get printed using normal skin settings. I think that 4 skin layers work OK.

I'm still not completely happy with the end result because when further layers of infill are printed on top of the bridge skin layers you do get some sagging. I expect that it is impossible with PLA to achieve a bridge with zero sagging but I shall keep experimenting and thinking about this.

I realise that there is a lot of settings but given that it is an experimental feature, only after people have experimented with it will we know which settings are useful and which are not and we can rationalise the settings in the future.

Anyway, please carry on the testing when you can.

PS - here's a simple test bridge. It takes a while to print and uses more filament than a simple bridge across two pillars but I have found that unless the pillars are attached to a base, the tension in the bridge tends to pull them off the build plate and so to get a really accurate idea of how the print is working out you need to ensure the pillars stay upright. For quick tests, you can always lower the model 2mm into the  build plate to remove the base. I could give the base some holes and that would save filament.

[multi-bridge.zip](https://github.com/Ultimaker/Cura/files/1779066/multi-bridge.zip)

smartavionics:
Hi @ianpaschal , I have added yet more settings and tweaked some default values. As I mentioned above, I think it's a good idea at this stage to provide as much flexibility as possible so that people can experiment to get the best results. Later, we can simplify if necessary.

Please use the updated engine PR also.

Here's an updated version of my test bridge.

[multi-bridge.zip](https://github.com/Ultimaker/Cura/files/1784633/multi-bridge.zip)

smartavionics:
Here's a couple of pics of the bottom of a test print. The bridge undersides look a bit rough as the first layer lines get distorted when the second layer goes down but the important points are that, overall, there isn't a huge amount of droop in the middle of the bridge and also that the bonding between those bridge skins is pretty good so the bridge is reasonably strong.

img_20180306_112931898

img_20180306_113004287_hdr

ianpaschal:
I've done some more tests as has @jackha so we are merging this.

PR #3187, opened 2018-01-25
Custom G-code commands for USB printing

ChrisTerBeke:
A start in adding functionality to send custom G-code commands for USB printing.


ChrisTerBeke:
It's now functional and styled, so ready to be reviewed and hopefully merged.

**Features**
* Send any custom G-code command over USB (or any output device that accepts manual controls).
* The text field is cleared after sending the command so a new command can be entered.
* Sending the command can be done by hitting enter (we could add a button but I feel like using enter is way quicker).
* There's a popup text when hovering.

**Screenshots**
screen shot 2018-01-26 at 22 10 12

screen shot 2018-01-26 at 22 13 38

Tested on MacOS so far, but should work on any OS since it's re-using the send_gcode function that already existed.

ChrisTerBeke:
@fieldOfView feel free to have a look or improve the UI :)

fieldOfView:
It feels like a bit of a clutch to have a raw gcode field for everybody. Personally I would rather have the ability to specify a number of gcode commands (or snippets) for certain actions than having to memorize those gcode commands.

Could this be implemented more pluginable? Something where like MachineActions a plugin can register a qml control to be added where you now have the Send G-code section? That way multiple buttons can be added for specific actions (eg turn on/off the leds on an UM2).

fieldOfView:
I think people will be asking for a way to see responses to g-code that is sent if you give them a g-code field like this.

cura/PrinterOutput/PrinterOutputModel.py
```diff
@@ -90,6 +90,10 @@ def homeHead(self):
     def homeBed(self):
         self._controller.homeBed(self)
 
+    @pyqtSlot(str)
+    def sendCustomCommand(self, command: str):
```
fieldOfView: How about "sendRawCommand" or "sendGcodeCommand" instead? Slightly more reusable in (other) plugins

resources/qml/PrinterOutput/ManualPrinterControl.qml
```diff
@@ -429,6 +428,120 @@ Item
...
+                // Input field for custom G-code commands.
+                Rectangle
+                {
+                    id: customCommandControl
+
+                    // state
+                    visible: printerModel != null ? printerModel.canPreHeatBed: true
+                    enabled: {
+                        if (printerModel == null) {
+                            return false // Can't preheat if not connected
```
fieldOfView: comment does not apply here

resources/qml/PrinterOutput/ManualPrinterControl.qml
```diff
@@ -101,29 +100,29 @@ Item
...
-                return true
-            }
-
-            if (activePrintJob.state == "printing" || activePrintJob.state == "resuming" || activePrintJob.state == "pausing" || activePrintJob.state == "error" || activePrintJob.state == "offline")
-            {
-                return false; //Printer is in a state where it can't react to manual control
-            }
-            return true;
-        }
+//        enabled:
```
fieldOfView: lots of commented-out code, which needs removing

resources/qml/PrinterOutput/ManualPrinterControl.qml
```diff
@@ -429,6 +428,120 @@ Item
...
+
+            Row
+            {
+                // Input field for custom G-code commands.
+                Rectangle
+                {
+                    id: customCommandControl
+
+                    // state
+                    visible: printerModel != null ? printerModel.canPreHeatBed: true
```
fieldOfView: canPreHeatBed has nothing to do with this functionality. canControlManually would be more appropriate, though a separate "ability" might be introduced (canSendRawGcode ?)

Ghostkeeper:
Developers, see issue CURA-4879.

cura/PrinterOutput/PrinterOutputModel.py
alekseisasin: I really like the idea of sending Raw G-code only it is such a master mode which might break machine by making small typo.

fieldOfView: It would be great if this functionality could be a downloadable plugin.

ChrisTerBeke: Right now that would mean overwriting the entire monitor stage? There is no QML target area to place a button there.

fieldOfView: The proper way to do this would probably involve creating a model of monitor plugins that can each provide a QML file.

LipuFei:
@ChrisTerBeke Conflicts to resolve...

ChrisTerBeke:
If it get's merged like this (only adding the input field for now) I can fix those. If more features are needed before it's merge-worthy I don't think I have time to work on it.

ChrisTerBeke:
@LipuFei I've fixed the merge conflicts
@fieldOfView I've incorporated your feedback

resources/qml/PrinterOutput/ManualPrinterControl.qml
ChrisTerBeke: Actually it needed to be un-commented, it disabled the state checking for development to see the components while no USB printer was attached :P

resources/qml/PrinterOutput/ManualPrinterControl.qml
ChrisTerBeke: Added this.

cura/PrinterOutput/PrinterOutputModel.py
ChrisTerBeke: Renamed to sendRawCommand. The other thing I can't really do until there is a QML target area for it.

resources/qml/PrinterOutput/ManualPrinterControl.qml
```diff
@@ -429,6 +427,120 @@ Item
...
+                    visible: printerModel != null ? printerModel.canSendRawGcode: true
+                    enabled: {
+                        if (printerModel == null) {
+                            return false // Can't preheat if not connected
+                        }
+                        if (!connectedPrinter.acceptsCommands) {
+                            return false // Not allowed to do anything
+                        }
+                        if (connectedPrinter.jobState == "printing" || connectedPrinter.jobState == "pre_print" || connectedPrinter.jobState == "resuming" || connectedPrinter.jobState == "pausing" || connectedPrinter.jobState == "paused" || connectedPrinter.jobState == "error" || connectedPrinter.jobState == "offline") {
+                            return false // Printer is in a state where it can't react to pre-heating
```
Ghostkeeper: Doesn't matter if it can react to pre-heating or not, though.

It's probably wise to put this code somewhere in a property at the top of the file. We currently repeat this check at several places (case in point, you copy-pasted it from the pre-heating code). Giving the root item a check like this can prevent that:

    property var notIdling: connectedPrinter.jobState == "printing" || connectedPrinter.jobState == "pre_print" || connectedPrinter.jobState == "resuming" || connectedPrinter.jobState == "pausing" || connectedPrinter.jobState == "paused" || connectedPrinter.jobState == "error" || connectedPrinter.jobState == "offline"

(or maybe something more semantic like `connectedPrinter.jobState != "idle"`)

And then use `if (notIdling)` instead of this check here and at line 123.

Ghostkeeper: In the interest of development speed I think I'll just make this change myself for you to help out. It's just a style thing after all.

PR #3196, opened 2018-01-27
Only select all text when tabbing through fields, not when clicking a field

fieldOfView:
This PR tweaks the behavior introduced in #3017, which intended to select all text when tabbing through text input fields. In #3017 text is also selected when the control is clicked. Some users have reported that the latter is unwanted.

This PR makes sure the text selection is not triggered when the input field gains the focus by pressing on the input field. I have made the PR against the 3.2 branch, since it fixes behavior introduced in 3.2; some would call the current behavior in 3.2 a regression.

See https://github.com/Ultimaker/Cura/issues/1969#issuecomment-352632035 for the original comment, and https://github.com/Ultimaker/Cura/pull/3017#issuecomment-358357664 and https://community.ultimaker.com/topic/21149-introducing-ultimaker-cura-32-beta/?tab=comments#comment-197552 for the complaints

### DIFF
--- a/resources/qml/Settings/SettingTextField.qml
+++ b/resources/qml/Settings/SettingTextField.qml
@@ -13,11 +13,17 @@ SettingItem
 
     property string textBeforeEdit
     property bool textHasChanged
+    property bool focusGainedByClick: false
     onFocusReceived:
     {
         textHasChanged = false;
         textBeforeEdit = focusItem.text;
-        focusItem.selectAll();
+
+        if(!focusGainedByClick)
+        {
+            // select all text when tabbing through fields (but not when selecting a field with the mouse)
+            focusItem.selectAll();
+        }
     }
 
     contents: Rectangle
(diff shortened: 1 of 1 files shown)


hoegge:
I'd like to oppose the reversion of this. Selecting the whole field is standard and makes sense, it is much faster to retype the full number than starting to select, delete individual parts etc. Don't change normal be behavior to fit a slow typist. Thanks. He can still click once more with the mouse to select the single digit or part he wants to edit. One thing you should change though: 
Right now the 2nd click with the mouse selects the digits to the right of the decimal point instead of selecting nothing and putting the cursor to the right. It should do the latter

fieldOfView:
A decision is needed from the Ultimaker UX team.

fieldOfView:
Would you prefer I close this and create a new PR for the master branch, or keep this one and then one of you merge the commit into master instead of 3.2, if it gets approved.

nallath:
For our reference; CURA-4922

ChrisTerBeke:
@fieldOfView could you rebase this to master? Then it'll get added for 3.3.

fieldOfView:
I will resubmit this against master.

PR #3203, opened 2018-01-29
Plugin browser

ianpaschal:
Re-styled plugin browser. Some small tweaks are needed but it is fully functioning.

### DIFF
--- a/plugins/PluginBrowser/PluginBrowser.py
+++ b/plugins/PluginBrowser/PluginBrowser.py
@@ -22,6 +22,8 @@ import platform
 import zipfile
 import shutil
 
+from cura.CuraApplication import CuraApplication
+
 i18n_catalog = i18nCatalog("cura")
 
 # Architecture thoughts:
@@ -76,6 +78,11 @@ class PluginBrowser(QObject, Extension):
         self._plugins_metadata = []
         self._plugins_model = None
 
+        # Can be 'installed' or 'availble'
+        self._view = "available"
+
+        self._restart_required = False
+
         self._dialog = None
         self._restartDialog = None
         self._download_progress = 0
@@ -112,6 +119,8 @@ class PluginBrowser(QObject, Extension):
     pluginsMetadataChanged = pyqtSignal()
     onDownloadProgressChanged = pyqtSignal()
     onIsDownloadingChanged = pyqtSignal()
+    restartRequiredChanged = pyqtSignal()
+    viewChanged = pyqtSignal()
 
     @pyqtSlot(result = str)
     def getLicenseDialogPluginName(self):
@@ -245,6 +254,8 @@ class PluginBrowser(QObject, Extension):
         self.pluginsMetadataChanged.emit()
 
         self.openRestartDialog(result["message"])
+        self._restart_required = True
+        self.restartRequiredChanged.emit()
         # Application.getInstance().messageBox(i18n_catalog.i18nc("@window:title", "Plugin browser"), result["message"])
 
     @pyqtSlot(str)
@@ -254,7 +265,22 @@ class PluginBrowser(QObject, Extension):
         self._newly_uninstalled_plugin_ids.append(result["id"])
         self.pluginsMetadataChanged.emit()
 
+        self._restart_required = True
+        self.restartRequiredChanged.emit()
+
         Application.getInstance().messageBox(i18n_catalog.i18nc("@window:title", "Plugin browser"), result["message"])
+
+    @pyqtSlot(str)
+    def enablePlugin(self, plugin_id):
+        self._plugin_registry.enablePlugin(plugin_id)
+        self.pluginsMetadataChanged.emit()
+        Logger.log("i", "%s was set as 'active'", id)
+
+    @pyqtSlot(str)
+    def disablePlugin(self, plugin_id):
+        self._plugin_registry.disablePlugin(plugin_id)
+        self.pluginsMetadataChanged.emit()
+        Logger.log("i", "%s was set as 'deactive'", id)
 
     @pyqtProperty(int, notify = onDownloadProgressChanged)
     def downloadProgress(self):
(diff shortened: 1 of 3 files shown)


ianpaschal:
Fixed quit button and improved QML to always use `UM.Theme.getColor()` (for theme support)

plugins/PluginBrowser/PluginBrowser.py
```diff
@@ -221,55 +279,72 @@ def cancelDownload(self):
...
-                "name": metadata["label"],
-                "version": metadata["version"],
-                "short_description": metadata["short_description"],
-                "author": metadata["author"],
-                "already_installed": self._checkAlreadyInstalled(metadata["id"]),
-                "file_location": metadata["file_location"],
-                "can_upgrade": self._checkCanUpgrade(metadata["id"], metadata["version"])
-            })
-        self._plugins_model.setItems(items)
+        print("Updating plugins model...", self._view)
```
diegopradogesto: Can you remove this print out or make it smaller? Is a very long piece of text in the logs

plugins/PluginBrowser/PluginBrowser.py
```diff
@@ -221,55 +279,72 @@ def cancelDownload(self):
...
+        self._plugins_model = PluginsModel(self._view)
+        # self._plugins_model.update()
+
+        # Check each plugin the registry for matching plugin from server
+        # metadata, and if found, compare the versions. Higher version sets
+        # 'can_upgrade' to 'True':
+        for plugin in self._plugins_model.items:
+            if self._checkCanUpgrade(plugin["id"], plugin["version"]):
+                plugin["can_upgrade"] = True
+                print(self._plugins_metadata)
```
diegopradogesto: This one is the longest, sorry.

plugins/PluginBrowser/PluginBrowser.py
```diff
@@ -290,7 +365,11 @@ def _onRequestFinished(self, reply):
             if reply_url == self._api_url + "plugins":
                 try:
                     json_data = json.loads(bytes(reply.readAll()).decode("utf-8"))
+
+                    # Add metadata to the manager:
                     self._plugins_metadata = json_data
+                    print(self._plugins_metadata)
```
diegopradogesto: Also this one. If it is worthy, then use Logger.log instead

ianpaschal:
Just deleted and committed to master since the branch was deleted and it couldn't affect any other code.

fieldOfView:
I have a couple of issues with the new plugin browser. Please see this as constructive criticism (though it may sound like I disagree with every little detail).

I'll start with functional issues. I see no reason why the plugin browser, and especially the overview of the installed plugins can't be in preferences on the old location. Either that, or remove the Plugins pane of the preferences; the placeholder is a clutch at best; Plugins -> Installed plugins... is fairly broken now.

Opening the dialog takes over a second for me, until data is refreshed. The dialog should open immediately, possibly showing a progressbar indicating that data is still being downloaded. That way at least I can switch to the "Manage" tab to eg disable a plugin.

There's some functionality missing: The "Install" "tab" misses a Browse... button to browse for a plugin to install plugins that are not in the "store". The "Manage" tab no longer shows plugins that could not be loaded (which is important info for troubleshooting). The use of whitespace is quite wasteful, showing only about 6 plugins at a time, which makes the "Manage" tab rather more unwieldly.

Then there's the theming. Unfortunately the current implementation uses a hybrid between the Cura and the "system" theme. It should use the system theme, like all dialogs. The implementation of the "Cura" theme has some issues by itself: 
* No mouseover/mousedown clues
* Wrong hyperlink color (especially unreadable in dark theme)
* Wrong font (no font specified, which would be ok for the system theme, but not for the Cura theme)
* Wrong listview outline (dark theme)

Theming is not just about looks though: The dialog is currently not keyboard aware; if you used the "system" theme you would get that almost automatically. The QML uses a lot of pixel sizes that are not DPI aware, so there will be a stark difference between Hi DPI and "normal" screens. Because you use Text elements instead of Label elements (like the rest of Cura), the text is rendered differently than all other text in Cura.

ianpaschal:
Thank you for your feedback Aldo.

I'll do my best to address your feedback "point-by-point-ish".

Regarding the location and the new message in settings. Yes, it's true, there are better ways for how to hand the redirection from settings to the new plugin browser, but this is something that still needs to be discussed and in the mean time, this works and is less confusing than having it in two places, or simply removing it without a warning.

Regarding the refresh time. This is a valid point and I like the idea of a progress bar, although the loading time is what it always was. Maybe "Manage" should be first in the meantime.

Regarding the browse button. We'd like to emphasize the usage of a future, broader plugin repository system, as well as the functionality that you can simply drop plugins into Cura to install them. You can still always manually clone a git repo (or download the folder and drag it) into the plugin directory as well for the "power users." I would not agree with you that functionality is really missing.

Regarding the theming: This is true. To be honest, I am not familiar with QML and I'm still getting the hang of how it works and how specifically components and styles are implemented in Cura/Uranium. There's very little documentation online about how things should in QML and also very little internal documentation either. Labels instead of Text? They seem to do pretty much exactly the same thing and there's no explanation why one should be used over another. Even now your remark doesn't help to clarify that. As for pixel sizes, these *are* used all over the place, with the exception of with window sizes. Reusing the standard button components would help with mouse cues. Normally I'd be in favor of "as native as possible" but since the "system" theme does not actually look native, instead it just looks like "the UI designer just forgot about these."

Finally, regarding the whitespace: I'm sorry you find it wasteful. Everyone else who I've spoken to has considered it a significant step up in aesthetics, so props to the Cura team designer (Luke).

P.S. :/ emojis are not super constructive. Cura routinely commits bug fixes straight to master and in this case it was literally just removing a print() statement.

awhiemstra:
> Regarding the browse button. We'd like to emphasize the usage of a future, broader plugin repository system, as well as the functionality that you can simply drop plugins into Cura to install them. You can still always manually clone a git repo (or download the folder and drag it) into the plugin directory as well for the "power users." I would not agree with you that functionality is really missing.

Dragging and dropping plugins is not discoverable however, so you are basically completely hiding that functionality. So to most users, yes, this functionality will be missing.

> Regarding the theming: This is true. To be honest, I am not familiar with QML and I'm still getting the hang of how it works and how specifically components and styles are implemented in Cura/Uranium. There's very little documentation online about how things should in QML and also very little internal documentation either. 

There indeed is missing documentation, mostly about which style should be used in which context. The general rule in Cura was always to use system styling for dialogs and most things "above" the window and use Cura styling for anything contained within the main view. 

> Labels instead of Text? They seem to do pretty much exactly the same thing and there's no explanation why one should be used over another. Even now your remark doesn't help to clarify that. 

Label is a subclass of Text that applies some system-specific styling. That is the major difference. 

> As for pixel sizes, these are used all over the place, with the exception of with window sizes. 

Are they? As far as I know, the sizes from the Theme are used in most places, which are based on font size. I'd be rather worried if a lot of hardcoded pixel sizes were added recently.

> Reusing the standard button components would help with mouse cues. Normally I'd be in favor of "as native as possible" but since the "system" theme does not actually look native, instead it just looks like "the UI designer just forgot about these."

Maybe on OSX. On other platforms the looks are different, specifically because platform styling is uses. And the differences are especially noticable when using a dark system theme.

fieldOfView:
To start with the 😕 reaction, I was confused why the branch was merged to master, before a proper review was done; how can a branch get lost in a distributed versioning system such as github? That's all.

My first QML was on the Cura project too, and I am sure I have offended the "Cura rules" a lot more than you did. That's what code reviews are for. I considered putting my comments as inline comments in the code here, but thought that would have turned out (even more?) pedantic.

Dragging .curaplugin files onto the application works, but has no affordance; you have to know you can do this (or be instructed to), it is not something you are likely to find out. Having a "Browse...." (or similar) button on the GUI that is all about installing plugins is a lot more discoverable. 

Labels vs Text, see this issue: https://github.com/Ultimaker/Cura/issues/1748.
Regarding the hi-dpi vs lo-dpi scaling, see https://github.com/Ultimaker/Cura/pull/2484

About the whitespace being wasteful. I am not arguing over looks. I am comparing the old and the new GUI. The task of disabling the USB Printing plugin has become more mentally complex now that fewer plugins fit in the ui, and the alphabetical ordering has become less obvious due to the disconnect between one plugin name and the next.

ianpaschal:
Ah ok. I think you'll have to ask others more senior than me in the Cura team why. Although I think the gist was that it met the requirements listed in Jira for changes as well as was done with the understanding that we'd be doing a lot more reworking and expanding plugins.

@awhiemstra Because of Aldo's feedback, last night I thought about writing a new page in confluence about how UM.Theme should be used. Knowing why "Labels over Text" would also make a helpful addition. Regarding the sizing I tried to use UM.Theme.getSize() wherever possible, although I didn't know there were reusable buttons as they were not used in another part of Cura which was my first experience with QML. Could also be in a Wiki here and available to all since this _is_ FOSS.

@fieldOfView That's reassuring to hear you also violated rules haha. To be honest, inline comments work quite well. I have a lot on my plate today but I'll try to take a look at the scaling and label vs text issues you linked.

@both: I'll look into adding the import button back in, although I think we'd like to move away from using it anyway.

fieldOfView:
>> the "system" theme does not actually look native, instead it just looks like "the UI designer just forgot about these."
>
>Maybe on OSX. On other platforms the looks are different, specifically because platform styling is uses. And the differences are especially noticable when using a dark system theme.

Come to think of it, this may have changed in QtQuick Controls 2:http://docs.huihoo.com/qt/5.x/qtquickcontrols2-differences.html

 For me, that would be reason enough *not* to use QtQuick Controls 2 when native styling is required.

fieldOfView:
> That's reassuring to hear you also violated rules haha.

Hey, I implemented materials... in Uranium! You've got nothing to be worried about ;-)

PR #3211, opened 2018-01-31
Added Support Eraser Plugin

ianpaschal:
This is the first draft of the Support Eraser tool. The plugin creates a cube mesh which has the `anti_overhang_mesh` decorator causing it to block the creation of support material for overhangs within its volume. This distinction is necessary to avoid confusing this behavior from actually erasing a _portion_ of a support structure.

Some (non-necessary) improvements could be made such as:

- Better graphical style
- Discussion of whether the cube should be able to pass through the build plate or not
- Possible improvements to the tool's icon
- Placing the cube at a cursor click location

### DIFF
--- a/plugins/SupportEraser/SupportEraser.py
+++ b/plugins/SupportEraser/SupportEraser.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
 from UM.Tool import Tool
 from PyQt5.QtCore import Qt, QUrl
 from UM.Application import Application
--- a/plugins/SupportEraser/SupportEraser.py
+++ b/plugins/SupportEraser/SupportEraser.py
@@ -1,0 +1,68 @@
+from UM.Tool import Tool
+from PyQt5.QtCore import Qt, QUrl
+from UM.Application import Application
+from UM.Event import Event
+from UM.Mesh.MeshBuilder import MeshBuilder
+from UM.Operations.AddSceneNodeOperation import AddSceneNodeOperation
+from UM.Settings.SettingInstance import SettingInstance
+from cura.Scene.CuraSceneNode import CuraSceneNode
+from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
+from cura.Scene.BuildPlateDecorator import BuildPlateDecorator
+from cura.Settings.SettingOverrideDecorator import SettingOverrideDecorator
+
+import os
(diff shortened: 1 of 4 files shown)


plugins/SupportEraser/__init__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2015 Ultimaker B.V.
```
nallath: It's 2018, welcome to the future! :)

ianpaschal: Good point. I copied `__init__.py` from the original Mirror tool.

[Personally I'm not a fan of adding this kind of information into code](https://hackerboss.com/get-rid-of-templates/), partially for this reason, but since I guess Ultimaker has its standards:

Should these files (including the `SupportEraser.py` file where I forgot it) read "2018"? Or does Ultimaker do some kind of "2015-2018"? Not sure if it's per file/per plugin/or should match for everything which is distributed in the Cura application.

nallath: I don't know. Someone started it, i just copied it. I tend to just use the year in which i wrote it.

plugins/SupportEraser/SupportEraser.py
```diff
@@ -0,0 +1,71 @@
...
+        node.addDecorator(SettingOverrideDecorator())
+        node.addDecorator(BuildPlateDecorator(active_build_plate))
+        node.addDecorator(SliceableObjectDecorator())
+
+        stack = node.callDecoration("getStack") #Don't try to get the active extruder since it may be None anyway.
+        if not stack:
+            node.addDecorator(SettingOverrideDecorator())
+            stack = node.callDecoration("getStack")
+
+        print(stack)
```
diegopradogesto: Remove debug printouts or use the logger

plugins/SupportEraser/SupportEraser.py
```diff
@@ -0,0 +1,71 @@
...
+        if event.type == Event.ToolActivateEvent:
+
+            # Load the remover mesh:
+            self._createEraserMesh()
+
+            # After we load the mesh, deactivate the tool again:
+            self.getController().setActiveTool(None)
+
+    def _createEraserMesh(self):
+        # Selection.clear()
```
diegopradogesto: Remove unused code

plugins/SupportEraser/SupportEraser.py
ianpaschal: Fixed and committed to master! Thanks Diego.

PR #3216, opened 2018-02-01
Enhance Dagoma profiles

dagoma3d:
Hi,

Here are some additional stuff regarding Dagoma printers.
- DiscoEasy200 : Add some default parameters to replicate the behavior we are using Cura By Dagoma.
- Neva : Add the needed stuff (profile and mesh).

Regards,
Orel from Dagoma


Appesteijn:
👍

PR #3218, opened 2018-02-01
CURA-3710 Add preset setting visibility files

LipuFei:


cura/CuraApplication.py
```diff
@@ -348,57 +351,19 @@ def __init__(self, **kwargs):
 
         preferences.setDefault("local_file/last_used_type", "text/x-gcode")
 
-        preferences.setDefault("general/visible_settings", """
```
ChrisTerBeke: Huge +1 for removing these lines :P

LipuFei: Huge +1 x2

LipuFei: Oh, it's my PR... Somehow I thought this is Ian's....

ianpaschal: I'll take it.

tools/check_preset_settings.py
```diff
@@ -0,0 +1,126 @@
...
+                        if setting_key not in setting_tree_dict[key]:
+                            if setting_key not in incorrect_settings:
+                                incorrect_settings[setting_key] = {"seen_in": [],
+                                                                   "should_be_in": self._should_setting_be_in(setting_tree_dict, setting_key)}
+
+                            incorrect_settings[setting_key]["seen_in"].append(key)
+
+                # show results
+                print("==========================================")
+                if incorrect_settings or incorrect_settings:
```
ianpaschal: This should be `if incorrect_settings or incorrect_sections`, right?

LipuFei: Nice catch!

tools/check_preset_settings.py
```diff
@@ -0,0 +1,126 @@
...
+        # parse the definition file
+        setting_tree_dict = self._parse_definition_file(self._fdmprinter_def_path)
+
+        has_invalid_files = False
+
+        # go through all the preset settings files
+        for root_dir, _, filenames in os.walk(self._preset_settings_dir):
+            for filename in filenames:
+                file_path = os.path.join(root_dir, filename)
+                print("Validating [%s] ..." % file_path)
```
fieldOfView: This should use Logger.log()

tools/check_preset_settings.py
```diff
@@ -0,0 +1,126 @@
...
+                    for setting_key in parser[key]:
+                        if setting_key not in setting_tree_dict[key]:
+                            if setting_key not in incorrect_settings:
+                                incorrect_settings[setting_key] = {"seen_in": [],
+                                                                   "should_be_in": self._should_setting_be_in(setting_tree_dict, setting_key)}
+
+                            incorrect_settings[setting_key]["seen_in"].append(key)
+
+                # show results
+                print("==========================================")
```
fieldOfView: debug statement?

LipuFei: This is an independent script which validates those files and outputs a report in console. That's why it's all prints and not using any cura or uranium modules.

LipuFei: I plan to put this script in CI some time later so it will become one of the checks we do on commits. Other checks/scripts can also be included in the future.

fieldOfView: Ok, my bad.

fieldOfView:
Might I suggest a different name for the resource folder instead of "preset_setting_visibility_groups"? The current name stands out from the other folders: ```setting_visibility```

"setting_visibility" is much shorter, and it opens the possibility of having user-defined sets in a folder by the same name in the configuration folder in the future (those would not be "presets"). Removing the "groups" also makes sense; the other folders are not "quality_profiles", "material_profiles", "variant_definitions" etc, but the much shorter "quality", "material" and "variants".

resources/qml/Preferences/SettingVisibilityPage.qml
```diff
@@ -85,14 +93,95 @@ UM.PreferencesPage
...
+                        index = i;
+                        break;
+                    }
+                }
+
+                visibilityPreset.currentIndex = index
+            }
+
+            id: visibilityPreset
+            width: 150
```
fieldOfView: This should be made DPI-aware

resources/qml/Preferences/SettingVisibilityPage.qml
```diff
@@ -85,14 +93,95 @@ UM.PreferencesPage
...
             }
 
             placeholderText: catalog.i18nc("@label:textbox", "Filter...")
 
             onTextChanged: definitionsModel.filter = {"i18n_label": "*" + text}
         }
 
+        ComboBox
+        {
+            property int customOptionValue: 100
```
fieldOfView: Wouldn't -1 be a saver value to use?

cura/CuraApplication.py
```diff
@@ -348,57 +351,19 @@ def __init__(self, **kwargs):
...
-                prime_tower_position_y
-            meshfix
-            blackmagic
-                print_sequence
-                infill_mesh
-                cutting_mesh
-            experimental
-        """.replace("\n", ";").replace(" ", ""))
+        setting_visibily_preset_names = self.getVisibilitySettingPresetTypes()
+        preferences.setDefault("general/visible_settings_preset", setting_visibily_preset_names)
```
fieldOfView: Let me get this straight: "general/visible_settings_preset" is a preference, that never gets set (so it never gets saved to the preference file), but it does have a default value that is set to a dict so the preference can be used to get at that dict from within QML? 

Does anyone mind if I clean this up a bit and make a model out of these setting visibility presets?

ianpaschal: Go for it. I think it would be appreciated.

fieldOfView: Work in progress here: https://github.com/Ultimaker/Cura/pull/3274

PR #3227, opened 2018-02-03
Fix the detection of multiextruder replacement patterns

fieldOfView:
This PR fixes the detection of multiextruder replacement patterns. Since #3068, replacement patterns can take the form of {setting_nr, extruder_nr}. This form was not being recognised when determining if CuraEngine should prepend its own preheat sequence.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -5,6 +5,7 @@ import numpy
 from string import Formatter
 from enum import IntEnum
 import time
+import re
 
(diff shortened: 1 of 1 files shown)


fieldOfView:
Would you prefer I close this and create a new PR for the master branch, or keep this one and then one of you merge the commit into master instead of 3.2?

LipuFei:
CURA-4920
@fieldOfView No need. We will merge this for 3.2.

fieldOfView:
Cool! I got a bit confused because the blogpost for the 3.2 stable release has already been published.

Issue #3133, opened 2018-01-14
Feature Suggestion: Keep Selected Post-Processing rules in place after shutdown and restart of Cura


MB3Y:
**Application Version**
3.1
**Platform**
Win10

**Steps to Reproduce**
Set some Pros-Processing rules.
Use them in the selected model.
Notice these rules keep active, even when clearing the buildplate and selecting a new model.
Shutdown Cura.
Restart Cura.

**Actual Results**
Post-Processing rules are not active anymore.
If you want to use them again, you have to create them again.

**Expected results**
Make Post-Processing rules persistant over restarts of Cura.
As a flagable  option maybe in the Post-Processor itself.

**Additional Information**
Idea came up since I use the post-processor to compensate/as workaround for printer mis-interpretations of standard GCode. As such I should keep them rules active for any slicejob. See also #2918.

MB3Y:
Wouldn't it be possible to keep all happy by letting people the choice of keeping the settings permanent or not? Intermediate step could also be to store the formerly used rules such that you can select previous rules instead of forming them from scratch again?

________________________________

alekseisasin:
Actually what you are asking is to store Post-Processing settings. To save them to external location, I think it is doable, only to support for different versions might be tricky. (I mean if you save in cura 3.1 then it will work for 3.1 without guarantee for the next version)

ChrisTerBeke:
Duplicate of #2796 and #2854.

Issue #2854, opened 2017-11-26
Feature Request: Store applied postprocessors also in Print Profile


Drayson80:
the postprocessors and its settings shall be attached/stored in print profiles too, to be ensure to have everything the same setup.

Ghostkeeper:
This is a duplicate of #2796.

Issue #2796, opened 2017-11-17
Plugin Settings not Saved


thesec:
Using post processing extensions I am all but getting used to have to set up these from scratch with every new project especially if it is just the model that I exchange or even just slightly tweaking a previously saved project.

Please save the extension/s' settings with the project file
or
at least save them when leaving the app.

fieldOfView:
This is by design. I never agreed with the design either.

The idea was that there were too many users that forgot they had postprocessing options enabled from the last session.

thesec:
... on the other hand all other is thrown in and even tried to mix machine settings from different projects ... all but consequent and all but good design with consistency being one of the base pillars of such.

One might think of integrating the post processing (and extension settings) with the extruder settings (being able to apply them per each ...) and as such making them transparent to the users and hence not risiking that they are forgotten about.

Getting this in as an option to advanced users likely to tweak more would also be an option.

Ghostkeeper:
We can't apply the post-processing settings for each extruder. They act on the final g-code, so they can even apply stuff cross-extruder, such as replace all `T1` commands by `M117 Do a little dance!`.

There's two main use cases for post-processing scripts: To fix some weirdness in the user's firmware that's incompatible with a preset g-code flavour (Search and Replace), and to make a tweak in the model (Tweak at Z or Pause at Height). For the first it would be handy to save it. For the second it wouldn't be handy at all. An option to save the post-processing script with the current printer can be added, but it'll have to be optional. I'll see what we can do about it.

thesec:
Thank you @Ghostkeeper .

Ghostkeeper:
See also https://github.com/nallath/PostProcessingPlugin/issues/46

thesec:
I see. Four+ totally separate settings' storages (Cura preferences, Printer settings, Extensions and the print settings) are just too much to handle and keep an eye on or even keep on mind ...

It would already be of great help to have all settings saved on exiting the application:
As a sample and reference I would like to point to Photoshop's extensions aka Filters where every single one's last settings are stored, despite affecting and being applicable on different part of the workflow, each.

ChrisTerBeke:
It seems this was fixed in #3229.

Issue #2103, opened 2017-07-11
Feature request: Search and Replace, Multiple Gcode Lines in Replace Frame


PM2189:
When Using the "Search and Replace" Post Processing Script it would be useful if it were possible to type in multiple lines of Gcode in the Replace frame. The main benefit would be the ability to add comments to the end of each line.

zerspaner-gerd:
To the info one can replace with `\n` several lines
Example:
![image](https://user-images.githubusercontent.com/29805503/28096572-21889d6e-66aa-11e7-832c-8ec59fe0bfab.png)

Ghostkeeper:
This just got removed from our planning by Cura's project manager because it was older than 12 weeks.

PR #3229, opened 2018-02-04
Persistent Post-Processing Scripts

Ghostkeeper:
This change will cause the added post-processing scripts to be saved in the user's configuration. Post-processing scripts were one of the few things that did reset when you restart Cura. No longer!

Post-processing scripts are added on a per-printer basis instead of globally. This choice was made because the scripts are quite often printer-dependent, compensating for broken firmware or adding small features to the g-code in the same way as the normal settings.
As such, the post-processing scripts will be removed when you switch to a different printer, and reappear if you switch back. You could view them as global settings.

Implementation details: The post-processing scripts that were added and the settings that belong to them are stored in the metadata of the global stack (as serialised config files). They can never cause a name clash with existing settings (which is a requirement also because you can add multiple of the same script). Their order is retained. It also does not require any version upgrade because the metadata is optional. If the metadata entry is missing it will cause the script list to be empty. If a post-processing script is changed in later versions to have a different setting name, that setting will reset to its default value unless a version upgrade deals with the change, just like our normal settings.


Ghostkeeper:
This implements https://github.com/Ultimaker/Cura/issues/2796, https://github.com/nallath/PostProcessingPlugin/issues/46, https://github.com/Ultimaker/Cura/issues/2854 and https://github.com/Ultimaker/Cura/issues/3133.

Ghostkeeper:
Developers, see issue CURA-4623.

Drayson80:
Damn, I want this feature so heavy since a long time... PLEASE make it into CURA3.3!!

jackha:
merge conflict: loadAllScripts needs to be rewritten :-(

Ghostkeeper:
The merge conflict is resolved now.

Kriechi:
This introduces a bug where newlines in the replace string are mangled/removed/lost.
ref #2103

Example:
  * Search: `;LAYER:`
  * Replace: `M118 P4 S"foobar"\n;LAYER:`
  * No regex

Not save & restart Cura. The newline will be missing. Or the UI simply refuses to display the whole string altogether end empties the field.

Ghostkeeper:
Thanks for the bug report. I've fixed that now in master, so it should be fixed in the next release.

thelastWallE:
Can we maybe have just a checkbox in the printsetting menu, there we can enable or disable all postprocessing altogether?
Or a set of settings which one can enable if he wants to print a temp tower.
To have to stacking the ChangeatZ Scripts on top of each other and change a number of settings is not very userfriendly. Then one would only have to set the bandwidth of layers for this temperature, startlayer for the first change, step-increment for the temperature and number of changes for the whole tower.
Anmerkung 2019-12-07 133938

Ghostkeeper:
> Can we maybe have just a checkbox in the printsetting menu, there we can enable or disable all postprocessing altogether?

The currently advised workflow (since post-processing scripts are stored with the printer) if you use post-processing scripts that are specific to a print rather than a printer, you can switch them on or off by having two printers and switching between the two. This way you won't have to go into the post-processing scripts list and delete all of them. The use case is small though, since if your post-processing scripts are specific to the print rather than the printer, then there's little reason to store them for the next print anyway; you'll need to remove them for the next print usually, unless you're printing many of the same thing but with small adaptations.

> Or a set of settings which one can enable if he wants to print a temp tower.

You mean the prime tower? There's a set of settings for that.

> To have to stacking the ChangeatZ Scripts on top of each other and change a number of settings is not very userfriendly. Then one would only have to set the bandwidth of layers for this temperature, startlayer for the first change, step-increment for the temperature and number of changes for the whole tower.

Which workflows are user friendly is a function of both the size of the use case (how often is it used) and complexity of the system. We've not yet implemented per-layer settings but we're keeping it open as a possibility for the future in our architecture. For now, ChangeAtZ is a passable workaround but even there we don't see a lot of usage yet.

If you actually have an issue with this, please open up a ticket and fill in the template!

PR #3230, opened 2018-02-04
Update zh_TW translation for 3.2

dinowchang:


Ghostkeeper:
Thanks. See https://github.com/Ultimaker/Uranium/pull/326

PR #3270, opened 2018-02-09
Fix rounding errors for garbled text

Appesteijn:
I went through all divisions in the qml files. Hope I didn't miss anything.

CURA-4941


fieldOfView:
I think the height in the settingcategory is the one

::thumbs_up::

ianpaschal:
This looks okay to me re: kerning and pixelation, but since I'm not in today I can't test on a non-HD display.

Also there are still some alignment issues but I've just checked and they were already there pre Noto Sans so that's a separate tweak needed somewhere.

PR #3274, opened 2018-02-09
Setting Visibility Presets sidebar menu

fieldOfView:
This PR adds a dropdown menu next to the search field in the custom sidebar to (even more quickly) switch between setting visibility presets. The menu also has an option to quickly switch to showing all settings (and then switching back to a more manageable set). It builds on the work done in #3218 

image

The PR moves the handling of setting visibility presets out of CuraApplication into a (singleton) model named SettingVisibilityPresetsModel. All code from #3218 is refactored to be contained in the Cura project instead of being split between Cura and Uranium. The code is prepared for allowing users to create their own setting visibility presets (by storing cfg files in the configuration folder), though this is not exposed to the user in the UI.


Ghostkeeper:
Let us know if you consider this ready!

fieldOfView:
The functionality that is currently missing is also missing in the current implementation: switching back to the custom set of visible items. But at the very least the conflicts have to be resolved. I'll try to get to that friday.

fieldOfView:
This PR is now ready for review

Appesteijn:
Tested an look very nice (and no crashes :) ) Code quality is up to the devs.

cura/CuraApplication.py
```diff
@@ -974,6 +882,7 @@ def registerObjects(self, engine):
         qmlRegisterType(MachineNameValidator, "Cura", 1, 0, "MachineNameValidator")
         qmlRegisterType(UserChangesModel, "Cura", 1, 0, "UserChangesModel")
         qmlRegisterSingletonType(ContainerManager, "Cura", 1, 0, "ContainerManager", ContainerManager.createContainerManager)
+        qmlRegisterSingletonType(SettingVisibilityPresetsModel, "Cura", 1, 0, "SettingVisibilityPresetsModel", SettingVisibilityPresetsModel.createSettingVisibilityPresetsModel)
```
LipuFei: @fieldOfView Just a note that we are reaching the limited number of singletons QML allows us to use, which is 30. We need to be careful not to add more singletons to QML...

fieldOfView: Should we start using some sort of UM.Models model?

ChrisTerBeke: Whut, there is a limit to that?!

Models like these don't need to be singletons per-se of course.

PR #3320, opened 2018-02-15
Set renderType to Text.NativeRendering for qtquick controls 2 Labels

fieldOfView:
This PR fixes font rendering of setting labels in the sidebar on affected OS X installations. Apparently, the qtquickcontrols 2 Label no longer defaults to using renderType: Text.NativeRendering
This causes text to render with QtRendering which looks subtly different on Windows and succinctly broken on some OSX installations.

With this PR the sidebar is now rendered with a single renderer again, making for a more consistent look on all systems and avoiding the broken QtRendering on OS X alltogether.

### DIFF
--- a/resources/qml/Settings/SettingCategory.qml
+++ b/resources/qml/Settings/SettingCategory.qml
@@ -78,6 +78,7 @@ Button
                 verticalCenter: parent.verticalCenter;
             }
             text: definition.label
+            renderType: Text.NativeRendering
             font: UM.Theme.getFont("setting_category")
             color:
             {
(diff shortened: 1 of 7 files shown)


ianpaschal:
@Appesteijn I could test this except that I never had the problem in the first place. Shall we create a build and post it in the comments as you have done before?

fieldOfView:
To emphasize the difference, I have switched to a theme that uses the default system font on Windows. Left is the current master, right is after this fix:
image

If you inspect the image closely you will see that the in the right version, setting label text is rendered more "crisp" and matches the input field text rendering.

Appesteijn:
So this improves the texts anyway, nice! Lets merge it and cherry-pick it into 3.2. 
I'll create a build then so people can test it.

fieldOfView:
Note that I seem to have missed one more Label; the Print Setup Label looks like it is still being rendered with Qt Rendering.

Appesteijn:
See my last commit on the 3.2 branch. I'm making a 3.2.2 build to send to testers.

PR #3325, opened 2018-02-15
Add a button to open the configuration folder.

fieldOfView:
This PR adds a button on the early crash dialog to open the configuration folder, and clarifies where configuration backups are stored.


Appesteijn:
👍

PR #3341, opened 2018-02-17
Post stretch algorithm fixes

electrocbd:
Extract line width from current extruder nozzle size
Change default values of stretch parameters from 0.08 to 0.1

Using the Cura version 3.2.1, the old global "line_width" parameter is not valid anymore (it reports 0.35 instead of 0.53 in my case). To fix this issue, I use the nozzle size of the current extruder.
By the way, I pushed small refinements...

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/Stretch.py
+++ b/plugins/PostProcessingPlugin/scripts/Stretch.py
@@ -12,6 +12,7 @@ import numpy as np
 from UM.Logger import Logger
 from UM.Application import Application
 import re
+from cura.Settings.ExtruderManager import ExtruderManager
 
 def _getValue(line, key, default=None):
     """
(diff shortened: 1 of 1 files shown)


nallath:
For our reference; CURA-4986

plugins/PostProcessingPlugin/scripts/Stretch.py
```diff
@@ -282,7 +283,7 @@ def wideCircle(self, orig_seq, modif_seq):
         dmin_tri is the minimum distance between two consecutive points
         of an acceptable triangle
         """
-        dmin_tri = self.line_width / 2.0
+        dmin_tri = 0.5
```
ianpaschal: This should not be hard coded to a fixed size. It makes more sense for this to depend on the print settings. Also, considering lines are often, say, 0.4mm, and this used to use 1/2 of that, bumping it up by 150% is a big jump!

electrocbd: In fact, this value is more linked to CuraEngine computation precision than to nozzle size or line width. The post stretch algorithm tries to find in which direction the nozzle turns when extruding the material. If it turns right,  the algorithm moves the nozzle to the left. If it turns left, it moves it to the right, to compensate the tendency of the plastic to take short cuts.
If several successive points are close, CuraEngine tends to drop several points, to speed up the exchange with the instrument and, sometimes, rounds positions. The consequence is that the post stretch algorithm fails to find the correct direction of the turn, and the movement is not stable (erratic movements).
To have smooth movements, the algorithm looks 0,5mm ahead. It is like looking far away to estimate next turn instead of looking at the front tire of the vehicle ;-)

ianpaschal: Ah ok. We were misinterpreting what that variable does. Could possibly use a better name but thank you for your explanation.

In this case, we'll go ahead with testing this and then merging it.

Ghostkeeper:
The difference is kind of hard to detect! The defaults are changed, which is easy. I could only find a difference in the moving of completely surrounded walls when I set the expansion distance to be very large. Result seems better now. Thanks for the small fix!

PR #3346, opened 2018-02-18
Various bug fixes for USB/serial attached printers

### LINKED ISSUE #1777: USB printing - Not working for Smoothieware firmware

slundell:
[edit]
I might have foobared this PR, thought I could send it as a separate, but I might have missed something. Please let me know if I should submit it in another way.

This PR contains various fixes in handling of USB/Serial printers, see commit messages for more info. With these fixes Cura should play nice with these printers. Tested on a Cetus3D with TinyFab smoothieware controller. 
[/edit]

Do not send new G-CODEs when the RX buffer is filling up. This is in … an attempt to fix #1777 where smoothieware based printers would halt in the middle of the print.
Tested on a tinyfab smoothieware board on a Cetus3D extended on USB. With this change, the print completes, without it it hangs in the middle of the print.


fieldOfView:
Perhaps you could document the magic number 256.

slundell:
The sarcasm was received. 
The error happened around 12Kb waiting in buffer. 256 was well under that, but still enough to give the host/client some slack in timing. 

`elif not self._serial.in_waiting:`
works well in my testing, I do not know if it may have side effects on other printers.

fieldOfView:
I did not intend sarcasm, sorry if I came across that way.

nallath:
For our reference; CURA-4985

slundell:
After some more research I have found that this is not correct way to fix the issue. It has to do with how smoothie and Cura treats empty gcode. Cura sends \n before gcodes and \n after. Smoothie sees this as an empty gcode followed by the actual gcode, hence answering with two ok\r\n . Cura interpretes the first ok\r\n as an ok for the sent gcode, when it is actually not. This results in that Cura consumes roughly half of what smoothie sends, hence the in buffer is filling up. When smoothie requests a resend, the communication gets flooded because cura sees the first ok\r\n in the buffer, and sends an other gcode, and then smoothie sends a new resend request. This continues until cura has consumed all the ok\r\n in front of the resend request. I assume the printer stops because it gets starved on gcodes and the communication is flooded so that it cannot get new ones.

I do not know if there is a standard for this kind of communication. Should a \n be sent before every gcode? Should an empty line be answered with ok\r\n ?

Easy fix is just to remove line 269:
`self._serial.write(b"\n")`
in UsbPrinterOutputDevice.py

If that is the correct way, depends on if there is a standard regarding this, and what that says.

slundell:
@nallath: Can you please explain what that is in reference to?

fieldOfView:
It is a reference to the internal issue tracker used at Ultimaker.

LipuFei:
@slundell There are conflicts...

Ghostkeeper:
The conflict is pretty trivial to solve.

The code looks fine to me and the Ultimaker printers still print all right via USB. I can't test the actual issue you were having because we have no printers with Smoothieware here, so we leave that responsibility to you. Thanks for the fix!

Shaische:
@Ghostkeeper Jaime should have one of our boards I believe. It is a smootheiboard that we sent him to fix this issue, perhaps he can test it? Otherwise we would be happy to test a compiled version.

Ghostkeeper:
Sure, just let me know what operating system you need it for.

Shaische:
Need it for Mac. If you can provide a compiled version, that would be great. Thank you!

Shaische:
Hi, any update on this?

Shaische:
@Ghostkeeper Any update on the Mac version to test?

fieldOfView:
Give it a week or 2...

Shaische:
Just tested the new 3.3 beta and it's been fixed! You can now print on smoothieware. However, when clicking pause print, the print pauses but when attempting to resume, nothing happens. Any ideas? Otherwise everything else seems to work!

Ghostkeeper:
Sorry I didn't come back to you. It got lost in a mountain of notifications.

When pausing via USB printing, Cura just pauses to send new g-code commands. It's possible that the Smoothieware firmware sort of times out on this then?

Jogging the printer while paused is explicitly disallowed. It says in the code that the printer is then not in a state where it can accept commands. This is probably not true though, but jogging the print head while paused could be a very bad idea if printing in relative mode. Then again, it is an explicit user action that can hardly happen by accident...

PR #3360, opened 2018-02-20
CURA-4790 Added weight properties to theme

ianpaschal:


ianpaschal:
Merging so Terri can build it.

PR #3368, opened 2018-02-20
Cartesio extruder change update

maukcc:
extruder change update:
- wiping nozzles
- temperature updates


Appesteijn:
CURA-5007

resources/quality/cartesio/abs/cartesio_0.25_abs_high.inst.cfg
```diff
@@ -23,7 +23,7 @@ infill_sparse_density = 40
 infill_pattern = grid
 
 material_print_temperature_layer_0 = =material_print_temperature + 5
-material_initial_print_temperature = =material_print_temperature
+material_initial_print_temperature = =material_print_temperature_layer_0 + 5
```
Ghostkeeper: I'll merge this for now I think, but the initial print temperature is the initial print temperature after an extruder switch, which doesn't only happen on the first layer. It'll pre-heat to this initial print temperature before the switch, then heat up to the normal print temperature after the switch.

It's a bit weird to have that depend on the initial layer height. It's fine for now, but it's a bit weird.

Ghostkeeper:
Looks okay for the rest. Thanks!

resources/quality/cartesio/abs/cartesio_0.25_abs_high.inst.cfg
maukcc: it was done because of #3382

PR #3411, opened 2018-02-28
setting to disable omitting retractions in support

BagelOrb:
Some users report that omitting the retractions causes print failures.
It might be machine/material dependant, so we need a user setting for this.

Omitting retractions was implemented somewhere in 2.5 or something, but it was never a setting.
We've since gotten some critique on this.


BagelOrb:
CuraEngine implementation: https://github.com/Ultimaker/CuraEngine/pull/709

Appesteijn:
devs: CURA-5106

PR #3425, opened 2018-03-02
M&p setting optimization

THeijmans:
Reverted combing settings (noskin lead to a lot of extra ooze), fixed the minimum travel for TPU and Connected our infill lines.


Ghostkeeper:
Seems to work fine. Thanks!

PR #3436, opened 2018-03-05
CURA-4400 Feature enable disable extruder

jackha:
Manually disable extruder(s).
Right click on a extruder tab -> context menu -> 'enable' & 'disable'


cura/Scene/CuraSceneNode.py
```diff
@@ -72,9 +78,32 @@ def getDiffuseColor(self) -> List[float]:
             1.0
         ]
 
+    def collidesWithBbox(self, check_bbox):
```
alekseisasin: Add documentation for this method

cura/Scene/CuraSceneNode.py
```diff
@@ -72,9 +78,32 @@ def getDiffuseColor(self) -> List[float]:
...
+        bbox = self.getBoundingBox()
+
+        # Mark the node as outside the build volume if the bounding box test fails.
+        if check_bbox.intersectsBox(bbox) != AxisAlignedBox.IntersectionResult.FullIntersection:
+            return True
+
+        return False
+
+    def collidesWithArea(self, areas):
```
alekseisasin: Add documentation for this method

cura/Scene/CuraSceneNode.py
Ghostkeeper: In particular it's interesting to know what happens in edge cases when they exactly align. This is important for the whole 214.9mm vs. 215mm build volume thing.

cura/Settings/MachineManager.py
```diff
@@ -690,6 +696,30 @@ def getDefinitionByMachineId(self, machine_id: str) -> str:
         if containers:
             return containers[0].definition.getId()
 
+    ##  Update extruder number to a valid value when the number of extruders are changed, or when an extruder is changed
+    def correctExtruderSettings(self):
+        extruder_count = self._global_container_stack.getProperty("machine_extruder_count", "value")
+
+        # reset all extruder number settings whose value is no longer valid
+        for setting_instance in self._global_container_stack.userChanges.findInstances():
```
Ghostkeeper: Also check for the setting instances in the quality_changes profiles.

cura/Settings/MachineManager.py
```diff
@@ -690,6 +696,30 @@ def getDefinitionByMachineId(self, machine_id: str) -> str:
...
+                continue
+
+            if not self._global_container_stack.getProperty(setting_key, "type") in ("extruder", "optional_extruder"):
+                continue
+
+            old_value = self._global_container_stack.userChanges.getProperty(setting_key, "value")
+            if int(old_value) >= extruder_count:
+                self._global_container_stack.userChanges.removeInstance(setting_key)
+                Logger.log("d", "Reset setting [%s] because its old value [%s] is no longer valid", setting_key, old_value)
+            if not self._global_container_stack.extruders[str(old_value)].isEnabled:
```
Ghostkeeper: Better use `or` for these two cases. The code below is exactly the same.

resources/qml/Settings/SettingExtruder.qml
```diff
@@ -72,6 +72,24 @@ SettingItem
...
+        Binding
+        {
+            target: control
+            property: "currentIndex"
+            value:
+            {
+                if(propertyProvider.properties.value == -1)
+                {
+                    // TODO: accidently the extruder position is also the index. fix it
+                    return Cura.MachineManager.defaultExtruderPosition;
```
Ghostkeeper: It's a bit hard to predict, but I think that if you set the extruder setting manually to the default (usually 0) this will cause the setting to be marked as changed even though it hasn't. The reset arrow will appear but won't do anything. The profile will be marked as containing user changes.

cura/Settings/SettingOverrideDecorator.py
```diff
@@ -90,10 +90,13 @@ def getActiveExtruderPosition(self):
...
         return self._is_non_printing_mesh
 
+    def evaluateIsNonPrintingMesh(self):
+        return any(bool(self._stack.getProperty(setting, "value")) for setting in self._non_printing_mesh_settings)
+
     def _onSettingChanged(self, instance, property_name): # Reminder: 'property' is a built-in function
         # Trigger slice/need slicing if the value has changed.
-        if property_name == "value":
-            self._is_non_printing_mesh = any(bool(self._stack.getProperty(setting, "value")) for setting in self._non_printing_mesh_settings)
+        if property_name == "value" and instance in self._non_printing_mesh_settings:
```
alekseisasin: This is quite interesting, it works fine for general settings. But if you set setting per model setting, I expect cura starts slicing. Can you check it?

cura/Settings/MachineManager.py
```diff
@@ -804,6 +805,12 @@ def updateNumberExtrudersEnabled(self):
     def defaultExtruderPosition(self):
         return self._default_extruder_position
 
+    ##  This will fire the propertiesChanged for all settings so they will be updated in the front-end
+    def forceUpdateAllSettings(self):
+        property_names = ["value", "resolve"]
```
Ghostkeeper: I'd think you would only need to emit the functions that call the `extruderValue` or `extruderValues` functions. Check for that instead of emitting all 400-something settings to be changed. It may be fast now, but in the future this can potentially slow it down a lot.

plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -269,9 +278,15 @@ def isCancelled(self):
     #   \return A dictionary of replacement tokens to the values they should be
     #   replaced with.
     def _buildReplacementTokens(self, stack) -> dict:
+        default_extruder_position = Application.getInstance().getMachineManager().defaultExtruderPosition
```
alekseisasin: you do need it here because you use it in one place. 
value = Application.getInstance().getMachineManager().defaultExtruderPosition

resources/qml/Settings/SettingExtruder.qml
jackha: It will mark it as changed, but I don't think that's a problem.
If the user explicitly sets the extruder, then IMO it can be regarded as a user setting and it will not change upon disabling / enabling extruders (unless that specific extruder has been turned off, then the setting is removed)

cura/Settings/ContainerManager.py
```diff
@@ -334,10 +334,12 @@ def clearUserContainers(self) -> None:
 
         # Go through global and extruder stacks and clear their topmost container (the user settings).
         for stack in ExtruderManager.getInstance().getActiveGlobalAndExtruderStacks():
-            container = stack.getTop()
+            container = stack.userChanges
             container.clear()
             send_emits_containers.append(container)
 
+        Application.getInstance().getMachineManager().correctExtruderSettings()
```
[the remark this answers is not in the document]
jackha: It was for the old implementation which uses user changes to force some extruder, but I removed it now.

cura/Settings/ExtrudersModel.py
```diff
@@ -183,11 +188,13 @@ def __updateExtruders(self):
                 item = {
                     "id": extruder.getId(),
                     "name": extruder.getName(),
+                    "enabled": extruder.isEnabled,
                     "color": color,
                     "index": position,
                     "definition": extruder.getBottom().getId(),
                     "material": extruder.material.getName() if extruder.material else "",
                     "variant": extruder.variant.getName() if extruder.variant else "",  # e.g. print core
+                    "stack": extruder,
```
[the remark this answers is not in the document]
jackha: From QML you have direct access to the extruder.

ianpaschal:
This is groovy. Jack has made all requested changes and I haven't managed to find any new bugs testing it. Mergin'!

PR #3447, opened 2018-03-06
New Definition

uBuild-3D:
New definition for our printer.

Thanks in advance.


resources/definitions/ubuild-3d_mr_bot_280.def.json
```diff
@@ -0,0 +1,49 @@
+{
+    "id": "ubuild-3d_mr_bot_280",
```
ChrisTerBeke: the ID field can be left out, we take that from the filename.

resources/definitions/ubuild-3d_mr_bot_280.def.json
```diff
@@ -0,0 +1,49 @@
...
+    "name": "uBuild-3D Mr Bot 280",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "uBuild-3D",
+        "manufacturer": "uBuild-3D",
+        "category": "Other",
+        "file_formats": "text/x-gcode",
+        "icon": "icon_uBuild-3D",
+        "platform": "mr_bot_280_platform.stl",
```
ChrisTerBeke: I don't see this STL file in the pull request, please add it in the resources/meshes folder.

resources/definitions/ubuild-3d_mr_bot_280.def.json
```diff
@@ -0,0 +1,49 @@
...
+        "has_materials": true
+    },
+
+    "overrides": {
+        "machine_name": { "default_value": "Mr Bot 280" },
+        "machine_heated_bed": { "default_value": true },
+        "machine_width": { "default_value": 275 },
+        "machine_height": { "default_value": 275 },
+        "machine_depth": { "default_value": 275 },
+        "machine_center_is_zero": { "default_value": false },
```
ChrisTerBeke: The default value for this is false, so it doesn't have to be added here.

uBuild-3D: I'll fix it asap

resources/definitions/ubuild-3d_mr_bot_280.def.json
fieldOfView: The icon line can be removed too. It is not used, but if it were used it would refer to an icon file which does not exist.

resources/definitions/ubuild-3d_mr_bot_280.def.json
```diff
@@ -0,0 +1,49 @@
...
+        "material_diameter": { "default_value": 1.75 },
+        "material_bed_temperature": { "default_value": 70 },
+        "machine_nozzle_size": { "default_value": 0.4 },
+        "layer_height": { "default_value": 0.2 },
```
fieldOfView: Do not specify a default layer_height, but instead specify a preferred quality. See eg https://github.com/Ultimaker/Cura/commit/05c034558dfe88e4c3ca6e12b023ff1291baea27

Ghostkeeper:
We currently seem to have a definition for `ubuild-3d_mr_bot_280` and there is a conflict. Was this change duplicate? It looks like this PR is outdated.

PR #3463, opened 2018-03-09
[Feature] Sync Button

diegopradogesto:
This PR enables the possibility to synchronize the current machine instance in Cura to one of the configurations available in the group you are connected to.
It also shows the list of printers ordered by "Network enabled printers" and "Local printers".
The name of the group now becomes the name of the host printer and so is the label in the machine selector.
The "Printers" management page is also grouped following the same structure.
In case a group has more than one type of printers, a new "Printer type" selector appears. In that case, when selected, a new container stack is created under-the-hood. However, not duplicate printers appear in the list since there is a new metadata entry called "hidden" that indicates when a printer should or not be shown in the list. The intention of this is to avoid the user confusion.

I'm open to reviews, advises and possible improvements.


ianpaschal:
You're a hero Diego! 🥇

plugins/MonitorStage/MonitorStage.py
```diff
@@ -92,46 +72,3 @@ def _updateSidebar(self):
...
-        if output_device.activePrinter.activePrintJob.state in ["ready", ""]:
-            return "tab_status_connected"
-
-        if output_device.activePrinter.activePrintJob.state == "paused":
-            return "tab_status_paused"
-
-        if output_device.activePrinter.activePrintJob.state == "error":
-            return "tab_status_stopped"
-
-        return "tab_status_unknown"
```
ianpaschal: I made these same changes this morning with Luke but wasn't fast enough to push them before you did. Question though: Do we want to delete it or just comment it out? This is functionality we may want reuse in the near future related to sync.

I mean I guess we could copy it from earlier commits but that can be kind of a pain.

resources/themes/cura-light/icons/buildplate.svg
```diff
@@ -0,0 +1,17 @@
...
+        <g id="Printer-status-icon" transform="translate(-33.000000, -366.000000)" stroke="#000000">
+            <g id="icn_buildplate" transform="translate(33.000000, 367.000000)">
+                <polyline id="Stroke-6823" points="0 5 7 8 14 5"></polyline>
+                <polyline id="Stroke-6823-Copy" points="0 7 7 10 14 7"></polyline>
+                <polyline id="Stroke-6823-Copy" points="0 9 7 12 14 9"></polyline>
+                <polygon id="Stroke-6824" points="0 2.75 7 0 14 2.75 7 5.5"></polygon>
+            </g>
+        </g>
+    </g>
+</svg>
```
ianpaschal: We should clean up our SVGs at some point but for now I guess it's not an issue.

cura/Machines/VariantManager.py
```diff
@@ -78,6 +83,22 @@ def initialize(self):
 
             variant_dict[variant_name] = ContainerNode(metadata = variant_metadata)
 
+            # If the variant is a buildplate then fill also the buildplate map
+            if variant_type == VariantType.BUILD_PLATE:
```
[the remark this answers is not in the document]
fieldOfView: ```==``` and ```is``` in python are as different as ```==``` and ````===``` in javascript. So you would use each as appropriate for what you mean.
https://dbader.org/blog/difference-between-is-and-equals-in-python

cura/PrinterOutput/ConfigurationModel.py
```diff
@@ -0,0 +1,66 @@
...
+        info += "Extruders: [\n"
+        for configuration in self._extruder_configurations:
+            info += "   " + str(configuration) + "\n"
+        info += "]"
+        if self._buildplate_configuration is not None:
+            info +=  "\nBuildplate: " + self._buildplate_configuration
+        return info
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
```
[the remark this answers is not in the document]
fieldOfView: This is used extensively all over Uranium:
https://github.com/Ultimaker/Uranium/search?utf8=%E2%9C%93&q=__eq__&type=

cura/CuraApplication.py
```diff
@@ -961,6 +960,7 @@ def registerObjects(self, engine):
         qmlRegisterType(BrandMaterialsModel, "Cura", 1, 0, "BrandMaterialsModel")
         qmlRegisterType(MaterialManagementModel, "Cura", 1, 0, "MaterialManagementModel")
         qmlRegisterType(QualityManagementModel, "Cura", 1, 0, "QualityManagementModel")
+        qmlRegisterType(MachineManagementModel, "Cura", 1, 0, "MachineManagementModel")
```
ChrisTerBeke: This is new in this Cura, so the version number should be higher. Check the other grouped version numbers (below in this method) as example.

cura/Machines/VariantManager.py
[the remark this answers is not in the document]
diegopradogesto: My intention was to compare if they are equals so I use the same solution as in other places: https://github.com/Ultimaker/Cura/blob/91e3e6b50c8625d4b18ee76035ea084be9d1c253/cura/Machines/VariantManager.py#L101

cura/PrinterOutput/ConfigurationModel.py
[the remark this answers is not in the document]
diegopradogesto: I can create a different function and use it for both __eq__ and __hash__ functions, but I don't think it makes a real difference.

cura/PrinterOutput/ConfigurationModel.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __str__(self):
+        info =  "Printer type: " + self._printer_type + "\n"
+        info += "Extruders: [\n"
+        for configuration in self._extruder_configurations:
+            info += "   " + str(configuration) + "\n"
+        info += "]"
+        if self._buildplate_configuration is not None:
+            info +=  "\nBuildplate: " + self._buildplate_configuration
+        return info
```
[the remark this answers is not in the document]
diegopradogesto: This is not used to show in QML, it was just useful for me to use it while debugging. I can remove it if it's not used anymore.

cura/PrinterOutput/ExtruderConfigurationModel.py
```diff
@@ -0,0 +1,49 @@
...
+    def hotendID(self):
+        return self._hotend_id
+
+    def __str__(self):
+        if self._material is None or self._hotend_id is None or self.material.type is None:
+            return "No information"
+        return "Position: " + str(self._position) + " - Material: " + self._material.type + " - HotendID: " + self._hotend_id
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
```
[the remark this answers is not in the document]
diegopradogesto: Same answer

plugins/MonitorStage/MonitorStage.py
diegopradogesto: The decision was to remove it. We can always revert this commit and go back to the previous code.

cura/CuraApplication.py
diegopradogesto: All the models introduced in the MM refactor are also new and all of them have the same version number. I agree, but then we should change them all.

cura/Settings/MachineManager.py
```diff
@@ -160,6 +170,42 @@ def _onOutputDevicesChanged(self) -> None:
...
+            self._current_printer_configuration.extruderConfigurations.append(extruder_configuration)
+
+        self._current_printer_configuration.buildplateConfiguration = self._global_container_stack.getProperty("machine_buildplate_type", "value") if self._global_container_stack.variant != self._empty_variant_container else None
+        self.currentConfigurationChanged.emit()
+
+    @pyqtSlot(QObject, result = bool)
+    def matchesConfiguration(self, configuration: ConfigurationModel) -> bool:
+        # print(configuration)
+        # print(self._current_printer_configuration)
+        # print("%%%%%%%%", configuration == self._current_printer_configuration)
```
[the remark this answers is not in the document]
diegopradogesto: Removed in da54c93fda6a9c795020607d69718e292524cfb6

cura/PrinterOutput/PrinterOutputModel.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, output_controller: "PrinterOutputController", number_of_extru
         # Update the printer configuration every time any of the extruders changes its configuration
         for extruder in self._extruders:
             extruder.extruderConfigurationChanged.connect(self._updatePrinterConfiguration)
+        self.typeChanged.connect(self._updatePrinterConfiguration)
```
ChrisTerBeke: Maybe use self.machineTypeChanged? Makes it a bit more explicity what type we're talking about.

cura/Machines/Models/MachineManagementModel.py
```diff
@@ -0,0 +1,90 @@
...
+        self._update()
+
+    ##  Private convenience function to reset & repopulate the model.
+    def _update(self):
+        items = []
+        # Remove all connections
+        for container in self._local_container_stacks:
+            container.nameChanged.disconnect(self._onContainerNameChanged)
+        for container in self._network_container_stacks:
+            container.nameChanged.disconnect(self._onContainerNameChanged)
```
jackha: I think it's better not to connect to models dynamically like this, or alter some models in any way in the _update function. Instead, connect to some generic signal for name changed (create it if needed).

cura/Machines/VariantManager.py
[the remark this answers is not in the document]
jackha: Looks fine IMO. Capitals indicate that the variable is used as a constant.

cura/PrinterOutput/ConfigurationModel.py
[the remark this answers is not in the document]
jackha: Small style thingy: I would have make a list of strings in info, then return "\n".join(info).

cura/PrinterOutput/ConfigurationModel.py
```diff
@@ -0,0 +1,66 @@
...
+        extruder_hash = hash(0)
+        first_extruder = None
+        for configuration in self._extruder_configurations:
+            extruder_hash ^= hash(configuration)
+            if configuration.position == 0:
+                first_extruder = configuration
+        if first_extruder:
+            extruder_hash &= hash(first_extruder)
+
+        return hash(self._printer_type) ^ extruder_hash ^ hash(self._buildplate_configuration)
```
jackha: The orring and anding look a bit hacky. Instead of orring and anding hashes, you can consider hashing a tuple of hashes.

cura/PrinterOutput/ExtruderOutputModel.py
```diff
@@ -68,3 +74,13 @@ def updateHotendID(self, id: str):
         if self._hotend_id != id:
             self._hotend_id = id
             self.hotendIDChanged.emit()
+
+    @pyqtProperty(QObject, notify = extruderConfigurationChanged)
+    def extruderConfiguration(self):
+        return self._extruder_configuration
+
+    def _updateExtruderConfiguration(self):
```
jackha: Why is this updating done in a separate step? Instead of setting the position, while setting self._position, material while setting self._active_material and hotendID while setting _hotend_id? Then it is never out of sync and maybe you don't need the signal.

cura/PrinterOutput/PrinterOutputModel.py
```diff
@@ -73,6 +82,15 @@ def updateType(self, type):
             self._type = type
             self.typeChanged.emit()
 
+    @pyqtProperty(str, notify = buildplateChanged)
+    def buildplate(self):
+        return self._buildplate_name
+
+    def updateBuildplate(self, buildplate_name):
```
jackha: Naming is a bit vague, maybe something like setBuildplateName?

cura/PrinterOutput/PrinterOutputModel.py
```diff
@@ -238,3 +256,14 @@ def canControlManually(self):
...
             return self._controller.can_control_manually
         return False
+
+    # Returns the configuration (material, variant and buildplate) of the current printer
+    @pyqtProperty(QObject, notify = configurationChanged)
+    def printerConfiguration(self):
+        return self._printer_configuration
+
+    def _updatePrinterConfiguration(self):
+        self._printer_configuration.printerType = self._type
```
jackha: Same comment as above about variables being able to be out of sync.

cura/Settings/MachineManager.py
```diff
@@ -306,12 +349,25 @@ def setActiveMachine(self, stack_id: str) -> None:
...
     @staticmethod
-    def getMachine(definition_id: str) -> Optional["GlobalStack"]:
+    def getMachine(definition_id: str, metadata_filter: Dict[str, str] = None) -> Optional["GlobalStack"]:
         machines = ContainerRegistry.getInstance().findContainerStacks(type = "machine")
         for machine in machines:
             if machine.definition.getId() == definition_id:
-                return machine
+                if metadata_filter:
+                    pass_all_filters = True
+                    for key in metadata_filter:
```
jackha: actually I think you can let findContainerStacks do this extra filtering as well. In that case you just pass the metadata_filter along

cura/PrinterOutput/PrinterOutputModel.py
diegopradogesto: I've just used a previously created signal. I agree in changing the name. Done in 54882402adcf00a7cd6a844173d34ef975f9656b

cura/PrinterOutput/ConfigurationModel.py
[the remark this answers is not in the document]
diegopradogesto: Done in b3f9679a5fddd94bc6ca1e3ddccc221ad911f2a4

cura/Machines/Models/MachineManagementModel.py
diegopradogesto: Thanks for the advice. I remove the connections to signals since I think I don't need those. Done in fa2b5f141c9c87cc533e7cf3f355450c470a832a

resources/qml/SidebarHeader.qml
```diff
@@ -380,74 +435,6 @@ Column
         }
     }
 
-    // Material info row
```
jackha: I suppose removing this is part of the design?

resources/themes/cura-light/icons/buildplate.svg
jackha: Apparently having groups and such (<g ...>) sometimes cause trouble for recoloring. Not an issue now.

cura/PrinterOutput/PrinterOutputModel.py
diegopradogesto: I rename it in dfb903fb8113b2694085d95f80d3982ab4707a16

cura/PrinterOutput/ExtruderOutputModel.py
diegopradogesto: I remove the listeners and set the values when the material and hotend are set. Done in dfb903fb8113b2694085d95f80d3982ab4707a16

cura/PrinterOutput/PrinterOutputModel.py
diegopradogesto: Same as above :-)

cura/Settings/MachineManager.py
diegopradogesto: Yes, I was thinking about it and at the end I did it differently. Changed and done in 180139090a085be88ce4aa2e8b5ff2712589917f

cura/Settings/MachineManager.py
```diff
@@ -988,6 +1062,69 @@ def _updateMaterialWithVariant(self, position: Optional[str]):
                 self._setMaterial(position, new_material)
                 continue
 
+    ##  Given a printer definition name, select the right machine instance. In case it doesn't exist, create a new
+    #   instance with the same network key.
+    @pyqtSlot(str)
+    def switchPrinterType(self, machine_name):
```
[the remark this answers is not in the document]
diegopradogesto: Yep!

resources/qml/Menus/PrinterMenu.qml
```diff
@@ -1,37 +1,60 @@
...
+import QtQuick.Controls.Styles 1.4
 
 import UM 1.2 as UM
 import Cura 1.0 as Cura
 
 Menu
 {
-    id: menu;
+    id: menu
+//    TODO Enable custom style to the menu
```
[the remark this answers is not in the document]
diegopradogesto: Yes, I started but the decision was to leave it like this for now. This will be done in the nearly future!

resources/qml/SidebarHeader.qml
diegopradogesto: Let's remove it and give it a try. If people complain we can always put it back.

resources/qml/Menus/PrinterStatusIcon.qml
```diff
@@ -0,0 +1,25 @@
...
+import QtQuick 2.2
+
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+Item {
+    property var status: "disconnected"
+    width: childrenRect.width
+    height: childrenRect.height
+    UM.RecolorImage {
```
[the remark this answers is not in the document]
diegopradogesto: Done in 855df814d83b368907e7c537bbdd18b161f6ffb5

cura/Machines/VariantManager.py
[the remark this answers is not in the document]
ianpaschal: @fieldOfView Ahhhh ok! Good to know.

PR #3472, opened 2018-03-11
Disable Support Eraser if anti_overhang_mesh is disabled

fieldOfView:
This PR disabled the Support Eraser tool for machines that have the "anti_overhang_mesh" setting disabled.

To test, a machine definition would have the following snippet:
```
        "anti_overhang_mesh": {
            "enabled": false
        },
```


fieldOfView:
I will include this in #3492

PR #3474, opened 2018-03-11
Added default material diameter to SeeMeCNC printer definitions

pouncingiguana:
All SeeMeCNC printers will only print with 1.75 diameter filament. Added this to the definitions of both printers.


Appesteijn:
👍

PR #3484, opened 2018-03-13
add parameters Print Infill Every & Print Support Infill Every

guillemavila:
To the user:
Change the setting Infill Layer Thickness to Print Infill Every (# layers), which seems to be more understandable. As the value of Infill Layer Thickness has to be multiple of the layer height it makes it easier to the user to set the setting.
Same for Support Infill Layer Thickness

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1665,6 +1665,21 @@
                     "limit_to_extruder": "infill_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "infill_sparse_layer":
+                {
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Devs: CURA-5110

Ghostkeeper:
To be consistent with the functionality of the rest of the settings, please make this a child setting of [Support] Infill Layer Thickness, and change CuraEngine to use the new layer count setting.

ianpaschal:
This needed some changes but never got them, nor did it get the accompanying PR in CuraEngine so I'm closing it.

PR #3485, opened 2018-03-13
Cura 5059 reset icon visibility

jackha:
Fixes: When modifying a setting value, the reset icon appears. If you put the value back manually to the original value, the reset icon will still remain. This is observed on Settings that are boolean and Limit to Extruder only... but it may be possible that there are more...


cura/Machines/MachineErrorChecker.py
```diff
@@ -0,0 +1,184 @@
...
+        self._error_keys_in_progress = set()
+        self._need_to_check = False
+        self.needToWaitForResultChanged.emit()
+
+        global_stack = self._machine_manager.activeMachine
+        if global_stack is None:
+            Logger.log("i", "No active machine, nothing to check.")
+            return
+
+        self._stacks_to_check = deque([global_stack] + list(global_stack.extruders.values()))
```
jackha: To me, it was not very clear what the "work stack" actually it. It seems like it's looping the stacks_to_check, then keys_to_check, then it refills keys_to_check. I think it's better to make a single deque with tuples (stack, key) that represent a single item to check and you can add them all here. In checkStack you only fetch one item and do its work, instead of refilling keys_to_check in some cases.

PR #3492, opened 2018-03-14
Support Eraser improvements

fieldOfView:
This PR will improve the Support Eraser tool in a number of ways.

The original implementation of the support eraser has a rather big usability issue. Other than the other Tools in the toolbar, it is an action that is performed once when clicking the icon. All the other Tools are actually modalities: you click the tool and then manipulate the selected models. This is confounded by the placement of created cubes in the center of the buildplate; in all likelyness the model itself obscures the added cubes. When first using the tool I had clicked the tool at least 10 times before I discovered as many cubes had been added under my model.

This PR makes the Support Eraser modal, like the other tools. Once the tool has been activated, the user can click on the model to place anti-support-cubes on the clicked location:
anti_support

Picking is implemented with an "on demand" renderpass to render the "depth" of the scene. This PickingPass is implemented as a generic class that could be useful in more places in Cura. It provides functionality to pick a location at up to micron accuracy and a range of up to 16 meter.

TODO:
* ~~~Put the created anti-support-cubes into a group with the clicked model so (a) they don't drop to the buildplate when that option is enabled and (b) they move with the object when it is transformed~~~
* ~~~Click an added anti-support-cube to remove it again~~~
* ~~~Ignore press when nothing is selected~~~
* Allow dragging the anti-support-cube to scale it
* A panel with instructions (?)

PR #3505, opened 2018-03-16
Further improvements of the Support Eraser

fieldOfView:
This PR is a small followup to #3492, which parents the added meshes to the clicked node instead of putting both the added meshes and the parent into a group. This prevents an extra convex hull from being drawn, and a wireframe cage from appearing. It does require a small change in PlatformPhysics, or otherwise the added mesh would still drop down to the buildplate.

### DIFF
--- a/cura/PlatformPhysics.py
+++ b/cura/PlatformPhysics.py
@@ -71,7 +71,7 @@ class PlatformPhysics:
             # Move it downwards if bottom is above platform
             move_vector = Vector()
 
-            if Preferences.getInstance().getValue("physics/automatic_drop_down") and not (node.getParent() and node.getParent().callDecoration("isGroup")) and node.isEnabled(): #If an object is grouped, don't move it down
+            if Preferences.getInstance().getValue("physics/automatic_drop_down") and not (node.getParent() and node.getParent().callDecoration("isGroup") or node.getParent() != root) and node.isEnabled(): #If an object is grouped, don't move it down
                 z_offset = node.callDecoration("getZOffset") if node.getDecorator(ZOffsetDecorator.ZOffsetDecorator) else 0
                 move_vector = move_vector.set(y = -bbox.bottom + z_offset)
 
--- a/plugins/SupportEraser/SupportEraser.py
+++ b/plugins/SupportEraser/SupportEraser.py
@@ -5,6 +5,7 @@ import os
 import os.path
 
 from PyQt5.QtCore import Qt, QTimer
+from PyQt5.QtWidgets import QApplication
 
 from UM.Math.Vector import Vector
 from UM.Tool import Tool
@@ -34,7 +35,7 @@ class SupportEraser(Tool):
     def __init__(self):
         super().__init__()
         self._shortcut_key = Qt.Key_G
-        self._controller = Application.getInstance().getController()
+        self._controller = self.getController()
 
         self._selection_pass = None
         Application.getInstance().globalContainerStackChanged.connect(self._updateEnabled)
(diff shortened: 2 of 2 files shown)


ianpaschal:
Yeah this feels a lot nicer to me. Do you think so too, or were you just humoring me, haha

The only last change I'd recommend is that if I select a support blocker and then click the support tool, and click it again it, removes. Then if I click another support blocker with the support tool still selected, it selects that one instead of removing it. To me, the tool should stay selected like other tools and and then remove any cubes you click like... bam bam bam rapid fire, vs. click to select and click again to erase.

fieldOfView:
I  agree this feels nicer. It was just the patch to PlatformPhysics that stopped me before.

I'll try to reproduce your findings to see if I can fix it; rapid-fire was the idea.

ianpaschal:
Here I will make a quick gif.

kapture 2018-03-16 at 14 31 20

fieldOfView:
Fixed that. Also, ctrl-clicking a node will now switch to the translate tool so it there is a quicker way to reposition added nodes.

ianpaschal:
giphy-downsized

Tested and works really great. Thanks so much Aldo!

ianpaschal:
@fieldOfView This may need some more changes. @Vandrasc has discovered a few issues:

1. When grouping and then ungrouping a mesh with blockers, the blockers will fall to the build plate. This is expected but possibly confusing to users who don't know what the feature is doing under the hood.
2. When importing and then ungrouping a project with blockers, the non-blocker meshes will just... vanish. This is of course not the expected behavior, even given what's under the hood.

This gives us two options as far as I can tell:

A. Fix the bug with importing and leave the ungrouping behavior as expected but possibly unintuitive (and see what our users have to say about it).
B. Use another mechanic to "stick" the blockers to their parent.

Personally B sounds more correct since the grouping feature is quite solid/stable when used for its intended purpose and I feel funny "fixing" it so it can do something that I don't think it was originally intended for.

What do you think? Do you know of another way to stick the blockers to their parent without putting them in a group?

fieldOfView:
1. is not expected (I did not expect it), but is caused by grouping not respecting the pre-existing parenting.
2. sounds to me like project saving or loading is broken

I have been considering adding a "FloatDecorator" to anti-overhang, infill and cutting meshes that exempts them from dropping to the buildplate. That would mitigate the issue, but wont make behavior less confusing. I feel strongly that the parenting in the current implementation is the best way to stick the objects to the model.

I will see if I can fix 1 properly, by fixing the grouping/ungrouping code. Expecto pee-errum.

ianpaschal:
1. Oh. Well I expected it. I know a lot of other software which prevents you from making a group consisting only of a group and figured Cura was just ignoring it and then performing the ungroup on the parent and its blockers.
2. Yeah that's a tricky one. Grouping = normally very reliable, project loading = not so much.

The FloatDetector would help with dropping to build plate, but wouldn't when moving the parent to the side a bit. If I were doing this in Three.js, it's a simple matter of adding the children to their parent in the scene graph so that their positions in the scene are relative to the parent, and then if the parent moves they move too. But that's separate from "grouping."

Do you think that could be done with Cura? I will take a look at the grouping code and if that's what grouping does already, in that case I might suggest that we add/use something like an "isGroup" decorator that lets the user ungroup things and not apply it to the relationship between a normal mesh and anti overhang meshes.

fieldOfView:
>  it's a simple matter of adding the children to their parent in the scene graph 

That is exactly what the code does now.

fieldOfView:
Here's a WIP branch that fixes the ungrouping issues:
https://github.com/Ultimaker/Cura/compare/master...fieldOfView:fix_grouping

I am checking to see if I can also fix the issue that multiplying a model that has support erasers attached does not copy the support erasers. If that is a similarly simple fix I will include both in one PR.

ianpaschal:
> That is exactly what the SupportEraser code does now.
Interesting. So I take it grouping adds a new container style node which the children are then added to?

Either way, thanks for working on a fix. You'll make Mark very happy if this all gravy by 3.3.

fieldOfView:
> You'll make Mark very happy

That's what we do this for, right?

Note: I'm not working on importing projects. That's someone else's mess.

ianpaschal:
> That's what we do this for, right?

Yes. Our benevolent dictator. Praise be.

> Note: I'm not working on importing projects. That's someone else's mess.

Yeah of course. But I think if the issue of ungrouping support blockers from their parent isn't possible in the first place, that bug will won't occur anyway.

fieldOfView:
See https://github.com/Ultimaker/Cura/pull/3524

PR #3506, opened 2018-03-16
Retraction minimum travel update

THeijmans:
Changed  it to 2 or 3 times line width, similar as with the 0.4mm profiles.  Was 5mm standard for all profiles with the 0.8mm printcore, except TPU.


ChrisTerBeke:
Our unit tests test if the UM3 and UM3 extended variants are the same. If they are not supposed to anymore, we should disable that test (otherwise the CI will keep failing).

Ghostkeeper:
The variants for UM3 and UM3E should still be the same, I think.

PR #3507, opened 2018-03-16
add Copy all values to all extruders

guillemavila:
Context menu option when right clicking in a value.
So the user has both 'copy value to all extruders' and 'copy all values to all extruders'.

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -550,17 +550,18 @@ class MachineManager(QObject):
             if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
                 extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
 
-    ## Copy the value of all settings of the current extruder to all other extruders as well as the global container.
+    ## Copy the value of all manually changed settings of the current extruder to all other extruders.
     @pyqtSlot()
     def copyAllValuesToExtruders(self):
-        for key in self._active_container_stack.userChanges.getAllKeys():
-            new_value = self._active_container_stack.getProperty(key, "value")
-            extruder_stacks = [stack for stack in ExtruderManager.getInstance().getMachineExtruders(self._global_container_stack.getId())]
+        extruder_stacks = list(self._global_container_stack.extruders.values())
+        for extruder_stack in extruder_stacks:
+            if extruder_stack != self._active_container_stack:
+                for key in self._active_container_stack.userChanges.getAllKeys():
+                    new_value = self._active_container_stack.getProperty(key, "value")
 
-            # check in which stack the value has to be replaced
-            for extruder_stack in extruder_stacks:
-                if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
-                    extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
+                    # check if the value has to be replaced
+                    if extruder_stack.getProperty(key, "value") != new_value:
+                        extruder_stack.userChanges.setProperty(key, "value", new_value)
 
     @pyqtProperty(str, notify = activeVariantChanged)
     def activeVariantName(self) -> str:
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -560,8 +560,7 @@ class MachineManager(QObject):
                     new_value = self._active_container_stack.getProperty(key, "value")
 
                     # check if the value has to be replaced
-                    if extruder_stack.getProperty(key, "value") != new_value:
-                        extruder_stack.userChanges.setProperty(key, "value", new_value)
+                    extruder_stack.userChanges.setProperty(key, "value", new_value)
 
     @pyqtProperty(str, notify = activeVariantChanged)
     def activeVariantName(self) -> str:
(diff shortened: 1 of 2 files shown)


Appesteijn:
Devs: CURA-5117

ChrisTerBeke:
Copying all the values to the global stack seems dangerous, as there is a specific difference between an extruder stack and the global stack (otherwise they would have been the same class).

@LipuFei how would this behave with the machine manager rewrite?

LipuFei:
I agree with @ChrisTerBeke that this seems dangerous. Note that there are also auto-calculated values which may differ between Extruder because they have different nozzle sizes, material diameters, material types, etc. In this case, literally copying all values including the auto-calculated ones to all extruders can produce bad results.

LipuFei:
And this has no impact on the machine manager rewrite.

guillemavila:
Yep, changed to replace only user changed values, avoiding machine_settings and not settable_per_extruder settings.
This keeps all calculated values, nozzle sizes etc

ChrisTerBeke:
Cool. The CI tests seem to fail though now.

cura/Settings/MachineManager.py
```diff
@@ -550,6 +550,18 @@ def copyValueToExtruders(self, key: str):
             if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
                 extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
 
+    ## Copy the value of all settings of the current extruder to all other extruders as well as the global container.
+    @pyqtSlot()
+    def copyAllValuesToExtruders(self):
+        for key in self._active_container_stack.userChanges.getAllKeys():
+            new_value = self._active_container_stack.getProperty(key, "value")
+            extruder_stacks = [stack for stack in ExtruderManager.getInstance().getMachineExtruders(self._global_container_stack.getId())]
```
LipuFei: It's better to use the global stack to get all extruders, so we don't depend on the `ExtruderManager`, which is something we want to remove later. You can do it like this:
```
extruder_stacks = list(self._global_container_stack.extruders.values())
```

cura/Settings/MachineManager.py
```diff
@@ -550,6 +550,18 @@ def copyValueToExtruders(self, key: str):
...
+
+            # check in which stack the value has to be replaced
+            for extruder_stack in extruder_stacks:
+                if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
+                    extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
```
LipuFei: Remove the `TODO` if it's not applicable.

cura/Settings/MachineManager.py
```diff
@@ -550,6 +550,18 @@ def copyValueToExtruders(self, key: str):
             if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
                 extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
 
+    ## Copy the value of all settings of the current extruder to all other extruders as well as the global container.
+    @pyqtSlot()
+    def copyAllValuesToExtruders(self):
+        for key in self._active_container_stack.userChanges.getAllKeys():
```
LipuFei: I think you can simplify/optimize the function this way:
```
extruder_stacks = [all extruders which is not the currently active one]
for key in active_extruder's userchanges keys:
    for extruder_stack in extruder_stacks:
        copy values over
```
This way, you don't need to reconstruct the `extruder_stacks` list in each iteration.

cura/Settings/MachineManager.py
```diff
@@ -550,6 +550,18 @@ def copyValueToExtruders(self, key: str):
             if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
                 extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
 
+    ## Copy the value of all settings of the current extruder to all other extruders as well as the global container.
```
LipuFei: Doc doesn't seem correct. Not the global container, only the extruders.

cura/Settings/MachineManager.py
```diff
@@ -550,17 +550,18 @@ def copyValueToExtruders(self, key: str):
...
-                if extruder_stack != self._active_container_stack and extruder_stack.getProperty(key, "value") != new_value:
-                    extruder_stack.userChanges.setProperty(key, "value", new_value)  # TODO: nested property access, should be improved
+        extruder_stacks = list(self._global_container_stack.extruders.values())
+        for extruder_stack in extruder_stacks:
+            if extruder_stack != self._active_container_stack:
+                for key in self._active_container_stack.userChanges.getAllKeys():
+                    new_value = self._active_container_stack.getProperty(key, "value")
+
+                    # check if the value has to be replaced
+                    if extruder_stack.getProperty(key, "value") != new_value:
```
LipuFei: Hi @guillemavila , I don't know if this is intended, but please be aware that, with this check, if two extruders happen to have the same values, but one is auto-calculated instead of being modified by the user, that value will not be copied. As a result, if later you change something in that extruder, the auto-calculated value may get changed as well, and you may not have the same values across all extruders.

If you want to make sure that all extruder must have the same values, compare with `extruder_stack.userChanges` instead.

guillemavila: Yes, well seen! I removed the check so that it forces all user changed values even if they are the same (auto calculated or not)

PR #3516, opened 2018-03-17
[Feedback wanted] Alternative USB transfer modes

slundell:
When printing over USB the G-code is streamed to the printer which executes them roughly as they arrive. The printer is then dependent on the host computer. If there are problems with the host computer or the interface, the print stops. This is an attempt to explore other means of transferring G-code to the printer. In this PR two modes are supported, the standard "streaming" way, and also "upload" by sending all codes at once. I also plan on implementing standard file transfer to the SD card.

I would like feedback on:
* Is there interest at all for this functionality?
* Is this way a sensible way to implement it?
* Is there a way to find which removable drive corresponds to a specific printer?
* When trying to update the printing progress (commented out in this PR) I get thread errors since I cannot update the UI from the current thread. I cannot get why I get this error, as AFAIK this should be the same way as was previously done.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -58,12 +58,17 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
 
         self._all_baud_rates = [115200, 250000, 230400, 57600, 38400, 19200, 9600]
 
+        self._supports_gcode_upload = None # use None as unknown
+        self._gcode_transfer_mode = GCodeTransferMode.unknown
+        self._is_uploading = False
+
         # Instead of using a timer, we really need the update to be as a thread, as reading from serial can block.
         self._update_thread = Thread(target=self._update, daemon = True)
 
         self._update_firmware_thread = Thread(target=self._updateFirmware, daemon = True)
 
         self._last_temperature_request = None  # type: Optional[int]
+        self._last_progress_request = None  # type: Optional[int]
 
         self._is_printing = False  # A print is being sent.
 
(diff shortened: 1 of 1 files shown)


fieldOfView:
Sounds nice, but what firmware supports this?

slundell:
I don't know really. I know smoothie does, but I don't know which others may. There might also be variations of this functionality in other firmwares. The idea is that the plugin should transperantly auto detect what transfer mode is supported in order of preference. First, regular OS file writing directly to a removable device corresponding to the printer (need ideas on a method to detect this). Second, the upload method implemented here, it is slow, but good in the sense that once transferred the printer can operate without the host computer. As a last resort, the current "streaming" method should be used.

ChrisTerBeke:
I don't think Marlin nor Repetier work like this. Even Repetier Host uses the streaming method with their own firmware. For Smoothieware I'm not sure, not familiar with it, but also not used that much anymore. Would it handle SD writing when there is no SD card in the printer properly? How to relate executed command responses with the original command for integrity checking?

marcosfg:
I believe it is very useful, namely if we want to use power failure detection and other features that require a "stand alone printer". 

Marlin allows the transfer of the gcode for the SD card through serial port, however it is extremely slow. I tested it in octoprint, I do not have the values but I believe It was transferred at ~10kbits/s.

Smootieware, implements both serial port and mass storage over USB. The mass storage interface allows write speeds of about 100kbits/s. Still not fast, but usable. After the file is transferred to the sd card, is possible to start the print and check the status of the print, like any other serial printer.

It is important to note that, after starting a print the media should be ejected. Recent OS usually like to continuously sync the contents of the attached drives, those tasks are "heavy" for a microcontroller.

There is a specific configuration for using smoothieware boards with octoprint to transfer files over mass storage interface, instead of default the serial port method:
https://github.com/foosel/OctoPrint/wiki/Setup-OctoPrint-with-Smoothie

Ghostkeeper:
Is it possible to detect if the connected printer supports this?

marcosfg:
By looking at the device descriptor it is possible to identify that a the connected device has both CDC class interface and USB mass storage. In this case, it can be assumed that we have a compatible printer connected.

Ghostkeeper:
Hmm, similar question: If the printer supports USB mass storage, won't the operating system detect this and won't Cura then display "Save to Removable Drive"?

marcosfg:
correct, in fact when we connect a smoothieware based board Cura detects the mass shortage device and allows the transfer the file to the printer SD card.

Currently there is no difference between connecting a usb flash drive from connection a USB MSD printer. By looking at the USB descriptor it would be possible to distinguish and show a different option for both.

There are however some limitations (in my opinion):

1. Operating systems usually have a transfer cache of a couple MBs. By default they assume that when the last GCode data is transfer to the cache the transfer is over... However, this last bytes (Mbytes) can still take a while to transfer. It is important to ensure synchronisation of the files transfer. If this does not happen, Gcode files my not be completely transferred...

2. After finishing transfer, we should send the GCode commands to the printer to start the print. I believe currently it is not possible to start a print from the internal memory using Cura, but it should be quite simple.

3. Monitor plugin should ask the printer the status of the print process and show updated information. Again, nothing extraordinary...

Ghostkeeper:
So our desired behaviour would be to:
* Somehow differentiate between a USB stick and a printer that allows USB mass storage to store prints on-device.
* If it's a USB MSD printer, let it behave as if it's a USB printer (instead of a USB stick that Cura says it is now). But still transfer files using the printer's file system rather than bare g-codes via baud.

This PR currently doesn't behave like that, but this gives a final answer to Slundell's initial question and request for discussion.

Appesteijn:
Closing this to keep things nice and tidy. Feel free to reopen when the discussion becomes more concrete.

PR #3524, opened 2018-03-19
Fix support eraser grouping

fieldOfView:
This PR fixes issues with grouping and meshes created with the support eraser tool.

See https://github.com/Ultimaker/Cura/pull/3505#issuecomment-374141862:
> @Vandrasc has discovered a few issues:
>
>  1.  When grouping and then ungrouping a mesh with blockers, the blockers will fall to the build plate. This is expected but possibly confusing to users who don't know what the feature is doing under the hood.
>  2.  When importing and then ungrouping a project with blockers, the non-blocker meshes will just... vanish. This is of course not the expected behavior, even given what's under the hood.

With this PR, support eraser meshes are no longer ungrouped. The PR also fixes multiplying meshes that have support eraser meshes parented to them.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1401,6 +1401,10 @@ class CuraApplication(QtApplication):
                 group_parent = node.getParent()
                 children = node.getChildren().copy()
                 for child in children:
+                    # Ungroup only 1 level deep
+                    if child.getParent() != node:
+                        continue
+
                     # Set the parent of the children to the parent of the group-node
                     op.addOperation(SetParentOperation(child, group_parent))
 
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1383,6 +1383,12 @@ class CuraApplication(QtApplication):
         group_node.setPosition(center)
         group_node.setCenterPosition(center)
 
+        # Remove nodes that are directly parented to another selected node from the selection so they remain parented
+        selected_nodes = Selection.getAllSelectedObjects().copy()
+        for node in selected_nodes:
+            if node.getParent() in selected_nodes and not node.getParent().callDecoration("isGroup"):
+                Selection.remove(node)
+
         # Move selected nodes into the group-node
         Selection.applyOperation(SetParentOperation, group_node)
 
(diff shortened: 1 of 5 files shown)


Appesteijn:
Devs: CURA-5121

ianpaschal:
Bad news. I'm getting the same behavior as before. In both places.

ianpaschal:
**Case 1:**
ungroup

**Case 2:**
bad import

fieldOfView:
~~~Are you sure you checkout out my branch? That's not what I am getting.~~~

Oh, wait, now I see. You first select all objects (which includes the parented objects) and then group them. This destroys the parenting.

fieldOfView:
This should fix both cases. Note that the project saved before will have incorrect grouping; please create a new project to test.

ianpaschal:
First one is fixed. Yay!

Second one still seems broken. I did create a new project. Same behavior as before with the disappearing base mesh.

Going to investigate further.

fieldOfView:
Can you attach a project?

ianpaschal:
[new testing.3mf.zip](https://github.com/Ultimaker/Cura/files/1825412/new.testing.3mf.zip)

ianpaschal:
I just noticed that the imported support blockers also have that old problem of only deleting on the 2nd click.

fieldOfView:
It is also interesting that when importing, the convex hull is wrong and the support meshes render in a lighter shade of grey (wut?).

ianpaschal:
Ah yes! There was a 3rd thing that I couldn't think of while writing it. That was it.

fieldOfView:
Ready for another round of testing.

The expected behavior with the tool is that when you deselect everything (click on the buildplate), the tool will not add/remove a support blocker. Instead that first click is used to make a selection. This is because when there is no selection you cannot activate another tool, so deselecting all objects would get you in a state where you cannot further manipulate the objects without adding or removing a support blocker. As long as there is a selection, the support blocker tool adds/removes support blockers on every click for me.

ianpaschal:
Yup! Looks great! Thank you Aldo!

PR #3529, opened 2018-03-20
Prime tower updates

THeijmans:
These settings should increase the reliability of PVA prime towers, especially in combination with CPE.

### DIFF
--- a/resources/variants/ultimaker3_bb0.8.inst.cfg
+++ b/resources/variants/ultimaker3_bb0.8.inst.cfg
@@ -10,6 +10,7 @@ hardware_type = nozzle
 
 [values]
 acceleration_enabled = True
+acceleration_prime_tower = =math.ceil(acceleration_print * 200 / 4000)
 acceleration_print = 4000
 acceleration_support = =math.ceil(acceleration_print * 2000 / 4000)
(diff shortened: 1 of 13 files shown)


resources/variants/ultimaker3_bb0.8.inst.cfg
```diff
@@ -39,7 +40,8 @@ material_print_temperature = =default_material_print_temperature + 10
 material_standby_temperature = 100
 multiple_mesh_overlap = 0
 prime_tower_enable = False
-prime_tower_purge_volume = 1
+prime_tower_purge_volume = 2
+prime_tower_wall_thickness = 2.2
```
Ghostkeeper: Maybe something like `=prime_tower_line_thickness * 3`?

THeijmans: Unfortunately Cura does not handle this well, as it will do weird short line pieces instead of three walls with that setting (bugged thin walls). =prime_tower_line_width * 3 + 0.1  would work though.

Ghostkeeper: I'll keep it this way for now, but maybe you can try again using the Try Multiple Line Thicknesses setting. It's an ongoing test to fix those broken walls.

PR #3545, opened 2018-03-23
[3.3] Support Eraser related fixes

fieldOfView:
This PR fixes some issues related to the Support Eraser, caused by parenting support eraser meshes to the picked node.

* Support Erasers added to scaled meshes now get the same size as those added to non-scaled meshes
* When multiplying a support eraser mesh, the parent node is multiplied instead (to make them behave the same way as groups)
* When placing support erasers near the bottom of a mesh, they would push the object up off the buildplate.

~~~In addition this PR fixes a longer standing bug when multiplying a node that has per object settings. These settings would not be copied over.~~~ This has been fixed in master (c7a3d33411a51f4fe284bfc45f0bba178636cc27)


fieldOfView:
Also check out https://github.com/Ultimaker/Uranium/pull/334, which is not dependent on this PR but fixes the last known issue (at least known to me)

fieldOfView:
Added a fix for when placing support erasers near the bottom of a mesh, they would push the object up off the buildplate (see https://github.com/Ultimaker/Cura/issues/3636#issuecomment-379684424).

Is this PR on the radar for the 3.3 release? It does not seem to have a JIRA ticker nr. yet.

Vandrasc:
@ianpaschal, can we still add this? See CURA-5209

fieldOfView:
Note that this was merged into master, because at the time I created the PR the 3.3 branch had not yet been created. This should probably be cherry-picked into 3.3.

LipuFei:
@fieldOfView Done in #3685

PR #3584, opened 2018-03-28
Add ZYYX Agile printer.

ohrn:
Hi!

This pull adds the ZYYX Agile printer by Magicfirm Europe. If it's not to late I'd love if it can be merged into the upcoming 3.3 release, otherwise master will do.

The printer uses X3G and the config is set up to work with the rpavlik:machine-type-metadata branch that is currently in a pull request to X3GWriter.

The profile was developed with input from Magicfirm. It also works with the older ZYYX+ printer.


ohrn:
Not sure I understand why the Jenkins build failed.

ianpaschal:
I'm not sure either. It's all green when looking at the build report... 😂

ohrn:
Hi!

Did you test the original push or the updated branch that includes the "Add global quality profiles." commit?

Adding the global qualities resolved a crash in 3.3 that I'm pretty sure is the same as the one you posted (I have to go back and double check). It didn't crash in 3.2 which is why I missed it at first.

ohrn:
Or, if it was the latest branch, did you have the patch to fdm_materials? I think it may be a bit funky after I added the preferred material setting.

ohrn:
@Appesteijn I've now double checked and the crash happened because you're missing the zyyx_pro_pla material, apply the https://github.com/Ultimaker/fdm_materials/pull/31 patch to fdm_materials an it'll work.

The crash happens when preferred_material references an non existing material, a bug worth fixing on it's own.

ohrn:
I took a stab at fixing the crash.

The easy solution of returning None from MaterialManager.getDefaultMaterial if the preferred material doesn't exists seems to make Cura select some random material, not good.

What would be nice is to walk up the inheritance chain of definitions and pick the first existing preferred material but I can't figure out how to do that. It doesn't seem that the inherits value is kept after the json files have been read.

Appesteijn:
I'll check it first thing tomorrow

Appesteijn:
👍 
Only a backplate is missing, but that is not blocking for this PR.
screen shot 2018-04-19 at 09 12 53

ohrn:
Thanks!

Issue #3563, opened 2018-03-26
[3.2.1] Fails to display second extruder's temperature


kirilledelman:
I've a modified Prusa Mk3 with 2 extruders. Since it's not in the list of printers, I've added it as Custom FDM printer with 2 extruders. The Monitor tab shows bed temperature and first extruder's temperature correctly, but not the second. Digging around, I found in Cura's log the following error message:
```
Could not parse hotend temperatures from response: b'ok T:11.56 /0.00 B:12.00 /0.00 T0:11.56 /0.00 T1:12.00 /0.00 @:0 B@:0 @0:0 @1:0\n'
```
in file/function USBPrinting.USBPrinterOutputDevice._listen

I've added some exception logging code and see it's triggering `list index out of range` in the code where it's looping over all matches to "T(\d*): ?([\d\.]+) ?\/?([\d\.]+)?" regular expression.

It looks to me like, it's finding my printer's `T0: temp/targetTemp` and calls `self._setHotendTemperature( 0, temp )` fine for first extruder, but calling it for the second triggers this exception. It's as if internally, inside `self._setHotendTemperature` the number of extruders is still 1, and it's hitting some array's boundaries. I couldn't find `def _setHotendTemperature` anywhere myself.

I was unable to rebuild Cura on MacOS (using cura-build) packages myself. Dependencies hell. But I can hack the py scripts sitting inside app bundle.

Anyone has any ideas what else I can try?

kirilledelman:
@slundell, @Ghostkeeper, do you guys have any ideas? 

I know I am running the last public release (3.2.1), not the down-to-the-wire version, but maybe you have some insight as to why `_setHotendTemperature` call with `1` for extruder, for an FDM printer with 2 extruders results in `list index out of range` exception.

I tried replacing just the USBPrinting plugin folder in my Cura with the latest, but of course, the API has changed and it didn't work. I'm so far unable to compile Cura myself on a Mac, stuck on some dependency, and running it on a Mac from source results in exceptions as well.

ianpaschal:
Possibly related to #3311? Basing this on the association between 3rd party, usb printing, and temperature readout.

GMagician:
This issue is also reported here: #3361

fieldOfView:
Note to self:
https://github.com/Ultimaker/Cura/issues/3361#issuecomment-367131150

ianpaschal:
Yeah marking as duplicate and closing.

kirilledelman:
[...]

PR #3597, opened 2018-03-29
[3.3] Fix getting multiple extruder temperatures from USB printer

fieldOfView:
This PR fixes the USB Printing plugin incorrectly handling multiple extruder temperatures. Zipping the extruders with the matches puts the temperature of the first extruder in both extruders, because the temperatures for the first extruder are returned twice for the M105 request; once without the tool number, and once with.

Could fix https://github.com/Ultimaker/Cura/issues/3563


Ghostkeeper:
This won't make it in time for 3.3, I'm afraid. Sorry.

Ghostkeeper:
Developers, see issue CURA-5259.

Issue #3050, opened 2017-12-27
bugs for Cura3.1


warnant:
**Application Version**

**Platform**

**Qt**

**PyQt**

**Display Driver**

**Steps to Reproduce**

**Actual Results**

**Expected results**

**Additional Information**

warnant:
- Cura 3.1 :
- Original+

Problèmes :
1-	Nivellement du plateau : la tête d’impression se positionne trop haut par rapport au plateau.
2-	Pas de curseur visible permettant de paramétrer la finesse d’impression (configuration de la couche/hauteur de la couche)
 
Luc Warnant

ChrisTerBeke:
Please fill out the issue template and post your comments in English, otherwise we might not take the time to process the issue.

warnant:
- Ultimaker Original +
- Cura 3.1

Problems: 
1- Level build plate : the print head is positioned too high in relation to the plate.
2- No visible slider to set the fineness of printing (Print setup / layer height)

Luc Warnant

ChrisTerBeke:
The Ultimaker original requires manual levelling of the hardware, guides for this can be found on the Ultimaker website. In Cura you must setup a first printer to see the settings appear in the right sidebar. Could you send a screenshot of your current situation?

warnant:
Hello Chris :-)
1- The first adjustments were made manually.
Leveling works in the previous version of Cura. The print head is positioned correctly
![image](https://user-images.githubusercontent.com/34890276/34529245-9efcac8c-f0ab-11e7-81cf-ef3fb6667850.png)
In version 3.1, the print head is too high in relation to the platter.
![image](https://user-images.githubusercontent.com/34890276/34529288-bbae6f14-f0ab-11e7-9770-5ad0b3adc251.png)
2- Original + is selected but the cursor is not present.
In the previous version of Cura, the cursor is present.
![image](https://user-images.githubusercontent.com/34890276/34529330-d6213de0-f0ab-11e7-9c4b-b6e24b9fc31f.png)

Lu :c)

ChrisTerBeke:
For 1 I'll have to ask some colleagues. It might be that the wrong settings are used somehow?
Nr 2 is actually expected because you have changes to the preset profile that belongs to one of these slider ticks. You can see that by hovering over the reset icon (back round arrow) in front of the slider. You can also click that reset button to discard your changes and use one of the preset profiles that belong to the slider ticks.

Ghostkeeper:
[...]

PR #3601, opened 2018-03-30
Fix sending commands in rapid succession over USB

fieldOfView:
This PR fixes sending commands in rapid succession over USB. When sending multiple commands using sendCommand(), it can happen that a command is sent before a previous command is fully received.

Probably contributes to #3050

### DIFF
--- a/cura/PrinterOutput/GenericOutputController.py
+++ b/cura/PrinterOutput/GenericOutputController.py
@@ -58,9 +58,7 @@ class GenericOutputController(PrinterOutputController):
         self._output_device.sendCommand("G90")
 
     def homeHead(self, printer):
-        self._output_device.sendCommand("G28 X")
-        self._output_device.sendCommand("G28 Y")
-        self._output_device.sendCommand("G28 Z")
+        self._output_device.sendCommand("G28 XY")
 
     def homeBed(self, printer):
         self._output_device.sendCommand("G28 Z")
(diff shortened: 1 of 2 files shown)


alekseisasin:
I tested this solution and it does not work for UM Original+. Has same issue.
I think we need make a special command for UM Original+ firmware 'Marlin V1', -> (G28 XYZ)

fieldOfView:
I don't have an UMO+, so I cannot test this. 

It still baffles me how changing the head homing to also home the bed would fix this (for any printer). Does homing the bed work properly using the jogging controls on the UMO+?

How is the UMO+ firmware different from the UM2?

fieldOfView:
Note that even if the PR does not fix the #3050 issue, it still fixes a possible overflow in the serial connection to USB printers, so I would still argue it needs to be merged. Perhaps @nallath could add his feedback, since het did a refactor of the USB printing code.

alekseisasin:
UM+ does not recognize 'X' or 'Y' in G28. For this reason we had before only 'G28' .  
I do not mind to merge your pull request, because it might help to prevent USB commands overflow. I have wish to solve the initial issue.
The firmwares are:
UM2  firmware name is "Marlin Ultimaker2"
UM+  firmware name is "Marlin V1"

fieldOfView:
What repo does the UMO+ firmware use? It is strange that G28 accepts the X & Y arguments just fine on my UMO and on my UM2.

LipuFei:
@fieldOfView Conflicts to resolve...

fieldOfView:
@LipuFei Done...

PR #3618, opened 2018-04-04
Backend of the Toolbox

diegopradogesto:


plugins/Toolbox/resources/qml/ToolboxViewDownloads.qml
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2018 Ultimaker B.V.
 // PluginBrowser is released under the terms of the LGPLv3 or higher.
 
-import QtQuick 2.2
+import QtQuick 2.7
```
ianpaschal: Are we sure this won't be an issue for people running PyQt 5.8? I'm not sure what you're developing on (5.8 vs 5.10 or whatever).

PR #3620, opened 2018-04-04
[3.3] Prevent preheat timeout from occurring during a print

fieldOfView:
This PR prevents preheat timers (bed, extruder) resetting the temperature to 0 after a print has started. Starting a print should stop all preheat timers, without setting the temperatures back to 0.

Steps to reproduce the issue:
* Use printer connected with USB
* Preheat bed
* Start print

After a couple of minutes of printing, the bed temperature will be set back to 0, and the print will detach from the buildplate.

### DIFF
--- a/cura/PrinterOutput/GenericOutputController.py
+++ b/cura/PrinterOutput/GenericOutputController.py
@@ -152,3 +152,17 @@ class GenericOutputController(PrinterOutputController):
         for extruder in self._preheat_hotends:
             self.setTargetHotendTemperature(extruder.getPrinter(), extruder.getPosition(), 0)
         self._preheat_hotends = set()
+
+    # Cancel any ongoing preheating timers, without setting back the temperature to 0
+    # This can be used eg at the start of a print
+    def stopPreheatTimers(self):
+        if self._preheat_hotends_timer.isActive():
+            for extruder in self._preheat_hotends:
+                extruder.updateIsPreheating(False)
+            self._preheat_hotends = set()
+
+            self._preheat_hotends_timer.stop()
+
+        if self._preheat_bed_timer.isActive():
+            self._active_printer.updateIsPreheating(False)
+            self._preheat_bed_timer.stop()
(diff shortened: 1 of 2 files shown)


fieldOfView:
Note that I have made this PR against 3.3; I think the issue is fairly critical and should be fixed before the 3.3 release.

kirilledelman:
I'm having this issue big-time. Would not using preheat function from Cura (but say, preheating the printer manually) not result in this behavior?

fieldOfView:
> Would not using preheat function from Cura (but say, preheating the printer manually) not result in this behavior?

That should work, but please test and let us know.

Note to devs: this fixes a fairly critical bug in Cura 3.3. Can we have a JIRA ticket nr?

diegopradogesto:
Here you are: CURA-5227 

Let's see if we can have it for 3.3

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -96,6 +96,9 @@ def requestWrite(self, nodes, file_name = None, filter_by_machine = False, file_
         if self._is_printing:
             return  # Aleady printing
 
+        # cancel any ongoing preheat timer before starting a print
+        self._printers[0].stopPreheatTimers()
```
LipuFei: This should actually be:
`self._printers[0].getController().stopPreheatTimers()`

cura/PrinterOutput/GenericOutputController.py
```diff
@@ -152,3 +152,17 @@ def _onPreheatHotendsTimerFinished(self):
...
+    def stopPreheatTimers(self):
+        if self._preheat_hotends_timer.isActive():
+            for extruder in self._preheat_hotends:
+                extruder.updateIsPreheating(False)
+            self._preheat_hotends = set()
+
+            self._preheat_hotends_timer.stop()
+
+        if self._preheat_bed_timer.isActive():
+            self._active_printer.updateIsPreheating(False)
```
LipuFei: And this should be
`self._preheat_printer.updateIsPreheating(False)`

LipuFei:
@fieldOfView OK, ignore that one. It's a problem with my GCode flavour. I will merge this and fix those 2 parts I mentioned.

LipuFei:
Fixed in 340916179

PR #3631, opened 2018-04-07
[3.3] Update zh_tw translation for cura 3.3

dinowchang:
Traditional Chinese translation for cura 3.3 was done.


Ghostkeeper:
Hi Dinowchang. I see that this PR is based on master rather than 3.3. We can't merge that into our 3.3 build because then we'd be merging unstable new features into the 3.3 build that shouldn't be in there. Could you rebase your branch off of 3.3 so that the changes from master won't be included?

dinowchang:
OK. I will rebase it to 3.3

Issue #3576, opened 2018-03-27
[3.3.0-beta] Cura crashes on first startup while parsing custom variant file


Spirou42:
**Application Version**
official 3.3b

**Platform**
macOS 10.13.3

**Display Driver**
AMD FirePro D700 6 GB

**Printer**
Ultimaker 3

**Steps to Reproduce**
Have a custom variant based on ultimaker3_aa04.inst.cfg from cura 3.2.1 which is missing the hardware_type entry in the [metadata] section.
start cura 3.3b for the first time (3.3 directory in the cura configuration directory)

**Actual Results**
crash 

**Expected results**
ignore the  "formerly correct" file, start up and inform the user, that there is a problem with one of the custom configuration files.

**Additional Information**
by the way, the fonts look much much better now.

ianpaschal:
I have a strong suspicion of the problem, but can you share your custom variant file so we can check for sure?

Spirou42:
Jep no problem, also have attached them to the crash report.
[CustomVariantFiles.zip](https://github.com/Ultimaker/Cura/files/1855477/CustomVariantFiles.zip)
The crash disappears, if i add *hardware_type* = *nozzle* in the [metadata] section 

btw, what other hardware_types are valid entries? I only have found he value *nozzle* inside the applications bundle.

jackha:
Yes this looks like a problem of the setting_version, it should have been increased. 
Devs: CURA-5175

Spirou42:
Thanks for the quick answer Diego.

PR #3638, opened 2018-04-09
Fix a crash when a variant has no hardware_type

fieldOfView:
This PR prevents a crash (and logs a warning instead) when a variant has no hardware_type. A hardware_type of "nozzle" is assumed (as all pre-3.3 variants are nozzle variants).


Vandrasc:
Related to this https://github.com/Ultimaker/Cura/issues/3576?

fieldOfView:
Yes, this fixes #3576, especially if the variant is in the application bundle instead of in the configuration folder. The variant upgrade will not work on variants in the application bundle.

LipuFei:
For dev, JIRA ticket: CURA-5293

PR #3642, opened 2018-04-09
Printer Definitions for Rigid3D Mucit

Rigid3D:
Adding Rigid3D Mucit definitions and build platform mesh


diegopradogesto:
Devs. CURA-5257

PR #3650, opened 2018-04-11
Fix gcode reader bug that can create a gap at the start of a spiralized layer

smartavionics:
The gcode reader assumed that each layer starts with a move to the initial position but
for spiralized code that isn't true because the previous layer always ends up in the right
location. So we now start each layer with a fake move to the end position of the previous layer.
This won't actually cause a real move to occur but it ensures that the first line segment
in the new layer has the correct initial point.

Without this PR:

screenshot_2018-04-11_08-15-55

With this PR:

screenshot_2018-04-11_08-18-10

I admit that this PR hasn't received extensive testing.

### DIFF
--- a/plugins/GCodeReader/FlavorParser.py
+++ b/plugins/GCodeReader/FlavorParser.py
@@ -368,6 +368,7 @@ class FlavorParser:
                     layer_number = int(line[len(self._layer_keyword):])
                     self._createPolygon(self._current_layer_thickness, current_path, self._extruder_offsets.get(self._extruder_number, [0, 0]))
                     current_path.clear()
+                    # start the new layer at the end position of the last layer
                     current_path.append([current_position.x, current_position.y, current_position.z, current_position.f, current_position.e[self._extruder_number], LayerPolygon.MoveCombingType])
 
                     # When using a raft, the raft layers are stored as layers < 0, it mimics the same behavior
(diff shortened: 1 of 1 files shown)


smartavionics:
Closed by mistake during spring cleaning! Reopening...

plugins/GCodeReader/FlavorParser.py
```diff
@@ -368,6 +368,8 @@ def processGCodeStream(self, stream):
                     layer_number = int(line[len(self._layer_keyword):])
                     self._createPolygon(self._current_layer_thickness, current_path, self._extruder_offsets.get(self._extruder_number, [0, 0]))
                     current_path.clear()
+                    # start the new layer at the end position of the last layer
+                    current_path.append([current_position.x, current_position.y, current_position.z, current_position.f, current_position.e[self._extruder_number], LayerPolygon.MoveCombingType])
```
diegopradogesto: I would also add this line when the gcode has a "changing tool" code.

smartavionics: Can you do that, please? I am not that familiar with Python or this code so I will probably get it wrong.

diegopradogesto:
Devs. CURA-5275

plugins/GCodeReader/FlavorParser.py
diegopradogesto: Yes, I'll do it 😄

PR #3651, opened 2018-04-11
No! G90 does not command absolute extrusion, M82 does that.

smartavionics:
See https://duet3d.dozuki.com/Wiki/Gcode#Section_G90_Set_to_Absolute_Positioning

### DIFF
--- a/plugins/GCodeReader/RepRapFlavorParser.py
+++ b/plugins/GCodeReader/RepRapFlavorParser.py
@@ -18,10 +18,10 @@ class RepRapFlavorParser(FlavorParser.FlavorParser):
             self._is_absolute_extrusion = False
 
     ##  Set the absolute positioning
-    #   RepRapFlavor code G90 sets position of X, Y, Z, and E to absolute
+    #   RepRapFlavor code G90 sets position of X, Y, Z to absolute
+    #   For absolute E, M82 is used
     def _gCode90(self, position, params, path):
         self._is_absolute_positioning = True
-        self._is_absolute_extrusion = True
         return position
 
     ##  Set the relative positioning


fieldOfView:
Slightly more descriptive PR texts would be useful.

smartavionics:
> Slightly more descriptive PR texts would be useful.

Well, the title really says it all, doesn't it?

ianpaschal:
It's factually accurate but I think what @fieldOfView means is that ideally the PR would say exactly what the code does. I actually thought this was an angry bug report when I saw it in my notifications.

smartavionics:
The title gives a hint as to what the problem is, the cited web page describes how G90 behaves and explicitly says that it doesn't affect the extrusion mode. The change in the PR is utterly trivial. Where's the problem? Why are you guys giving me a hard time?

ianpaschal:
I'm not really meaning to give you a hard time. I make a lot of commits with (hopefully) humorous messages. I'm just commenting because I'm sifting through GitHub notifications. 😜

fieldOfView:
It simply helps to mention that this fixes something that happens when reading gcode, specifically with the reprap firmware.

diegopradogesto:
I used this as source of truth: http://reprap.org/wiki/G-code#G90:_Set_to_Absolute_Positioning
Now I'm confused.

smartavionics:
It says there: _All coordinates from now on are absolute relative to the origin of the machine. (This is the RepRap default.)_ which, to me, implies (but doesn't explicitly state) that it's referring to XYZ and not E.

diegopradogesto:
Yes, it could be, but if you read in the G91 (http://reprap.org/wiki/G-code#G91:_Set_to_Relative_Positioning) it explicitly says: _Note: RepRapFirmware latest revision firmware uses M83 to set the extruder to relative mode: extrusion is NOT set to relative by ReprapFirmware on G91: only X,Y and Z are set to relative._

I will do more investigations on this before changing it.

Kriechi:
I'm an active user of RRF with a DuetWifi board.
The canonical resource for RRF g-codes is this wiki page: https://duet3d.dozuki.com/Wiki/Gcode#Section_G90_Set_to_Absolute_Positioning
Which contains this warning:

> Note: RepRapFirmware uses M82 to set the extruder to absolute mode: extrusion is NOT set to absolute using G90

and 
> Note: RepRapFirmware uses M83 to set the extruder to relative mode: extrusion is NOT set to relative using G91

diegopradogesto:
I've checked that the 3 maintained repositories for RepRapFirmware are dc42, chrishamm and dcnewman's (as pointed in http://reprap.org/wiki/RepRap_Firmware#Compiling_from_Source).
Then I checked the code and all of them behave the same, not using G90 for absolute extrusion. As you can see in:

- https://github.com/dc42/RepRapFirmware/blob/eb65356bab5a76f3f0e8873a8de2cd97840f9878/src/GCodes/GCodes2.cpp#L318
- https://github.com/chrishamm/RepRapFirmware/blob/690bcc643bd3333248db8e95d1488b3c5f6bc2cc/src/GCodes/GCodes2.cpp#L227
- https://github.com/dcnewman/RepRapFirmware-RADDS/blob/269aadf44f3d23c62f3c3701df530ae41d32d565/src/GCodes.cpp#L2603

I approve the changes.

smartavionics:
That's very thorough of you, I am impressed!

PR #3653, opened 2018-04-11
Now divides line width by 2 as this gives a visually better result.

smartavionics:
Without this PR:

screenshot_2018-04-11_15-40-25

With this PR:

screenshot_2018-04-11_15-34-36

### DIFF
--- a/plugins/GCodeReader/FlavorParser.py
+++ b/plugins/GCodeReader/FlavorParser.py
@@ -160,7 +160,8 @@ class FlavorParser:
         # Area of the printed line. This area is a rectangle
         Ae = dVe / dX
         # This area is a rectangle with area equal to layer_thickness * layer_width
-        line_width = Ae / layer_thickness
+        # 11/04/2018 - added divide by 2 to obtain visually correct line widths
+        line_width = Ae / layer_thickness / 2
 
         # A threshold is set to avoid weird paths in the GCode
         if line_width > 1.2:


plugins/GCodeReader/FlavorParser.py
```diff
@@ -160,7 +160,8 @@ def _calculateLineWidth(self, current_point, previous_point, current_extrusion,
         # Area of the printed line. This area is a rectangle
         Ae = dVe / dX
         # This area is a rectangle with area equal to layer_thickness * layer_width
-        line_width = Ae / layer_thickness
+        # 11/04/2018 - added divide by 2 to obtain visually correct line widths
```
fieldOfView: I don't think adding the date in this comment will add anything to the readability of this fix; there's always git blame. I am much more interested in knowing why dividing by 2 is the correct thing to do here (is it a radius vs diameter thing?)

smartavionics: > I am much more interested in knowing why dividing by 2 is the correct thing to do here (is it a radius vs diameter thing?)

I guess it could be. I think what is happening is that without the div by 2 the limit test below is coming out true and so the width gets set to 0.35? which would explain why the walls and the infill lines have the same width and everything is too narrow.

smartavionics:
Perhaps with this PR, the lines come out a little fat compared to what you see after the part has been sliced but I think it is better than them being so obviously wrong.

diegopradogesto:
This is not a good fix. I was checking and the problem is caused by a wrong filament_diameter. I'm working now on it. I will share with you and you can check if it works for you.

smartavionics:
Good, I thought that this wasn't the real problem.

diegopradogesto:
Check #3659

PR #3659, opened 2018-04-12
Adjust filament diameter to calculate line width in the gcode parser

diegopradogesto:


fieldOfView:
I was thinking the same but it makes sense to me now; you would have the same issue when you print that gcode with the “wrong” printer setup

diegopradogesto:
The diameter taken into account is the one that is set up in the current printer. There is no information in the gcode about the diameter so we can't guess the diameter. That's why we show this message:

> "Make sure the g-code is suitable for your printer and printer configuration before sending the file to it. The g-code representation may not be accurate."

smartavionics:
OK, that's fair enough.

smartavionics:
But... (there's always a 'but'), that info is available in the gcode file, e.g:

> ;SETTING_3 th = 0.5\\nmaterial_diameter = 1.75\\nmaterial_flow_layer_0 = 100\\nm

smartavionics:
Is it worth trying to look for that setting? Probably not, I think the warning is good enough

diegopradogesto:
This setting is just there if the material diameter was changed, otherwise it doesn't appear, so we still need to get the value from somewhere if the gcode was created with the "by default" filament diameter.
The best solution is to save the filament diameter in the header, but then if we use multiple extruders we need to set several values. It will be also confusing if the user opens the gcode (generated for 1.75mm) in Cura in a printer that uses 2.85mm filament and it looks good but when trying to print it's awful.

PR #3663, opened 2018-04-12
Feature material marketplace

ianpaschal:
Not yet ready for merge, but ready to start reviewing!

Please test in conjunction with uranium/CURA-4644-package-reader


cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+            management_dict = json.loads(f.read(), encoding = "utf-8")
+
+            self._installed_package_dict = management_dict["installed"]
+            self._to_remove_package_set = set(management_dict["to_remove"])
+            self._to_install_package_dict = management_dict["to_install"]
+
+            Logger.log("i", "Package management file %s is loaded", self._package_management_file_path)
+
+    def _saveManagementData(self) -> None:
+        with open(self._package_management_file_path, "w", encoding = "utf-8") as f:
```
Ghostkeeper: You need to check for the configuration file lock before writing this file, too. Otherwise you will get \x00 characters due to corrupt files when people are running multiple instances of Cura at the same time.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        if not os.path.exists(self._package_management_file_path):
+            Logger.log("i", "Package management file %s doesn't exist, do nothing", self._package_management_file_path)
+            return
+
+        with open(self._package_management_file_path, "r", encoding = "utf-8") as f:
+            management_dict = json.loads(f.read(), encoding = "utf-8")
+
+            self._installed_package_dict = management_dict["installed"]
+            self._to_remove_package_set = set(management_dict["to_remove"])
+            self._to_install_package_dict = management_dict["to_install"]
```
Ghostkeeper: You will get a crash if this file is corrupt. Better catch the `KeyError`s in this part.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+
+            Logger.log("i", "Package management file %s is loaded", self._package_management_file_path)
+
+    def _saveManagementData(self) -> None:
+        with open(self._package_management_file_path, "w", encoding = "utf-8") as f:
+            data_dict = {"installed": self._installed_package_dict,
+                         "to_remove": list(self._to_remove_package_set),
+                         "to_install": self._to_install_package_dict}
+            data_dict["to_remove"] = list(data_dict["to_remove"])
+            json.dump(data_dict, f)
```
Ghostkeeper: We should add a version number to this so that we can still read it after we need to change the data format of this file.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+            Logger.log("i", "Package management file %s is saved", self._package_management_file_path)
+
+    # (for initialize) Removes all packages that have been scheduled to be removed.
+    def _removeAllScheduledPackages(self) -> None:
+        print("Will purge", self._to_remove_package_set)
```
Ghostkeeper: Use the logger for this (or remove the debug print).

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        self._saveManagementData()
+
+    # Checks the given package is installed. If so, return a dictionary that contains the package's information.
+    def getInstalledPackageInfo(self, package_id: str) -> Optional[dict]:
+        if package_id in self._to_remove_package_set:
+            return None
+
+        if package_id in self._to_install_package_dict:
+            package_info = self._to_install_package_dict[package_id]["package_info"]
+            package_info["is_bundled"] = False
```
Ghostkeeper: So this `getInstalledPackageInfo` getter changes the metadata of a package. I think that is unexpected, right?

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        if package_id in self._installed_package_dict:
+            package_info = self._installed_package_dict.get(package_id)
+            package_info["is_bundled"] = False
+            return package_info
+
+        for section, packages in self.getAllInstalledPackagesInfo().items():
+            for package in packages:
+                if package["package_id"] == package_id:
+                    package_info = package
+                    return package_info
```
Ghostkeeper: Just `return package` then?

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+
+        return None
+
+    def getAllInstalledPackagesInfo(self) -> dict:
+        installed_package_id_set = set(self._installed_package_dict.keys()) | set(self._to_install_package_dict.keys())
+        installed_package_id_set = installed_package_id_set.difference(self._to_remove_package_set)
+
+        managed_package_id_set = set(installed_package_id_set) | self._to_remove_package_set
```
Ghostkeeper: Don't need to cast `installed_package_id_set` to a set again. It's already a set.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        # Also get all bundled plugins
+        all_metadata = self._plugin_registry.getAllMetaData()
+        for item in all_metadata:
+            plugin_package_info = self.__convertPluginMetadataToPackageMetadata(item)
+            # Only gather the bundled plugins here.
+            package_id = plugin_package_info["package_id"]
+            if package_id in managed_package_id_set:
+                continue
+
+            plugin_package_info["is_bundled"] = True if plugin_package_info["author"]["name"] == "Ultimaker B.V." else False
```
Ghostkeeper: Not all bundled plug-ins have the author "Ultimaker B.V." Look at the g-code reader for instance.

I think this check needs to be removed. We will want to merge plug-ins by other people in the future too.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        # Check the delayed installation and removal lists first
+        if package_id in self._to_remove_package_set:
+            self._to_remove_package_set.remove(package_id)
+            has_changes = True
+
+        # Check if it is installed
+        installed_package_info = self.getInstalledPackageInfo(package_info["package_id"])
+        to_install_package = installed_package_info is None  # Install if the package has not been installed
+        if installed_package_info is not None:
+            # Compare versions and only schedule the installation if the given package is newer
```
Ghostkeeper: I think this check should be performed in the GUI or the API, not in the back-end. If we want to support downgrading plug-ins in the future this would prevent us from doing that.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        if to_install_package:
+            Logger.log("i", "Package [%s] version [%s] is scheduled to be installed.",
+                       package_id, package_info["package_version"])
+            # Copy the file to cache dir so we don't need to rely on the original file to be present
+            package_cache_dir = os.path.join(os.path.abspath(Resources.getCacheStoragePath()), "cura_packages")
+            if not os.path.exists(package_cache_dir):
+                os.makedirs(package_cache_dir, exist_ok=True)
+
+            target_file_path = os.path.join(package_cache_dir, package_id + ".curapackage")
+            shutil.copy2(filename, target_file_path)
```
Ghostkeeper: This is not atomic, so you need to acquire the configuration file lock before doing this, or you'll have corrupt files if two instances of Cura are writing at the same time.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+
+            self._to_install_package_dict[package_id] = {"package_info": package_info,
+                                                         "filename": target_file_path}
+            has_changes = True
+
+        self._saveManagementData()
```
Ghostkeeper: If Cura ever crashes in the future before this file is written, the JSON will be out of date. For instance, if there is a PermissionError or something. Maybe you'll have to put this in a `try:` `finally:` block?

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+            Logger.log("i", "Attempt to remove package [%s] that is not installed, do nothing.", package_id)
+            return
+
+        # Remove from the delayed installation list if present
+        if package_id in self._to_install_package_dict:
+            del self._to_install_package_dict[package_id]
+
+        # If the package has already been installed, schedule for a delayed removal
+        # if package_id in self._installed_package_dict:
+        #     self._to_remove_package_set.add(package_id)
```
Ghostkeeper: Remove commented code.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        # "Or rather don't because sometimes packages are not making it into the
+        # dict I guess." - Ian
```
Ghostkeeper: TODO: Clear this up.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        self._to_remove_package_set.add(package_id)
+
+        self._saveManagementData()
+        self.installedPackagesChanged.emit()
+
+    # Removes everything associated with the given package ID.
+    def _purgePackage(self, package_id: str) -> None:
+        print("Purging",package_id)
```
Ghostkeeper: Use a logger or remove the debug print.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        from cura.CuraApplication import CuraApplication
+        dirs_to_check = [
+            Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer),
+            Resources.getStoragePath(CuraApplication.ResourceTypes.QualityInstanceContainer),
+            os.path.join(os.path.abspath(Resources.getDataStoragePath()), "plugins"),
+        ]
+
+        for root_dir in dirs_to_check:
+            package_dir = os.path.join(root_dir, package_id)
+            print(package_dir)
```
Ghostkeeper: Remove debug print.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+            if os.path.exists(package_dir):
```
Ghostkeeper: Don't we need to check for subdirectories?

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        print("Purging",package_id)
+        # Get all folders that need to be checked for installed packages, including:
+        #  - materials
+        #  - qualities
+        #  - plugins
+        from cura.CuraApplication import CuraApplication
+        dirs_to_check = [
+            Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer),
+            Resources.getStoragePath(CuraApplication.ResourceTypes.QualityInstanceContainer),
+            os.path.join(os.path.abspath(Resources.getDataStoragePath()), "plugins"),
```
Ghostkeeper: This is missing the resource path to definitions. Maybe we should just check the data storage path?

cura/CuraPackageManager.py
ianpaschal: In this case we need a new mechanic for determining which plugins are bundled. Open to ideas.

Ghostkeeper: You could check if their `__init__.py` file is located in one of the resource directories? If it is, then it is not bundled. If it isn't, then it is bundled. Not entirely obvious in code, but it is correct.

plugins/Toolbox/src/PackagesModel.py
```diff
@@ -0,0 +1,101 @@
...
+        if filter_dict != self._filter:
+            self._filter = filter_dict
+            self._update()
+
+    @pyqtProperty("QVariantMap", fset = setFilter, constant = True)
+    def filter(self) -> Dict[str, str]:
+        return self._filter
+
+    # Check to see if a container matches with a regular expression
+    def _matchRegExp(self, metadata, property_name, value):
```
Ghostkeeper: Maybe we should consider moving this function and `_matchString` to `ListModel` by this point...

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+                            self._metadata["authors"].append(package["author"])
+
+                    for author in self._metadata["authors"]:
+                        if "package_count" not in author:
+                            author["package_count"] = 0
+
+                        for package in self._metadata["packages"]:
+                            if package["author"]["name"] == author["name"]:
+                                author["package_count"] += 1
+                                author["type"] = package["package_type"]
```
Ghostkeeper: If an author creates packages of multiple types, this will select an arbitrary type for the author?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+                    # Create packages model with all packages:
+                    if not self._models["packages"]:
+                        self._models["packages"] = PackagesModel(self)
+                    self._metadata["packages"] = json_data["data"]
+                    self._models["packages"].setMetadata(self._metadata["packages"])
+                    self.metadataChanged.emit()
+
+                    # Create authors model with all authors:
+                    if not self._models["authors"]:
+                        self._models["authors"] = AuthorsModel()
```
Ghostkeeper: This is always constructed in the `__init__` function, so it will never hit this `if` statement.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+                                author["package_count"] += 1
+                                author["type"] = package["package_type"]
+                                if "icon_url" in package:
+                                    author["icon_url"] = package["icon_url"]
+
+                    self._models["authors"].setMetadata(self._metadata["authors"])
+                    self.metadataChanged.emit()
+
+                    if not self._models["materials_showcase"]:
+                        self._models["materials_showcase"] = AuthorsModel(self)
```
Ghostkeeper: The `AuthorsModel` is always constructed in the `__init__` function, so it will never hit this `if` statement.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+
+                    # Check for errors:
+                    if "errors" in json_data:
+                        for error in json_data["errors"]:
+                            Logger.log("e", "%s", error["title"])
+                        return
+
+                    # Create packages model with all packages:
+                    if not self._models["plugins_showcase"]:
+                        self._models["plugins_showcase"] = PackagesModel()
```
Ghostkeeper: The `PackagesModel` is always constructed in the `__init__` function, so it will never hit this `if` statement.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+                    self._models["materials_showcase"].setMetadata(self._metadata["materials_showcase"])
+
+                    # This part is also needed for comparing downloaded packages to
+                    # installed packages.
+                    self._models["packages"].setMetadata(self._metadata["packages"])
+                    self._models["packages"].setFilter({"type": "plugin"})
+
+                    self.metadataChanged.emit()
+
+                    self.setViewPage("overview")
```
Ghostkeeper: Wait, when the package list is updated, we always go to the overview page? Why?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+        if license_content is not None:
+            self.openLicenseDialog(package_info["package_id"], license_content, file_path)
+            return
+
+        self.install(file_path)
+        return
+
+
+    # Getter & Setters for Properties:
+    # --------------------------------------------------------------------------
```
Ghostkeeper: Missing type hinting for these setters.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+    @pyqtProperty(str, fset = setViewSelection, notify = viewChanged)
+    def viewSelection(self):
+        return self._view_selection
+
+
+
+    # Expose Models:
+    # --------------------------------------------------------------------------
+    # TODO: Maybe replace this with simply exposing self._models to Qt and then
+    # setting model: toolbox.models.foobar instead of toolbox.foobarModel
```
Ghostkeeper: Actually I think it is fine like this. Just my opinion though. This way you don't need a separate Models object.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+    @pyqtProperty(QObject, notify = metadataChanged)
+    def materialsInstalledModel(self):
+        return self._models["materials_installed"]
+
+
+
+    # Filter Models:
+    # --------------------------------------------------------------------------
+    @pyqtSlot(str, str, str)
+    def filterModelByProp(self, modelType, filterType, parameter):
```
Ghostkeeper: Code style: use_underscores_for_parameters.

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
```diff
@@ -0,0 +1,43 @@
...
+# Cura is released under the terms of the LGPLv3 or higher.
+
+import configparser #To parse preference files.
+import io #To serialise the preference files afterwards.
+
+from UM.VersionUpgrade import VersionUpgrade #We're inheriting from this.
+
+
+##  Upgrades configurations from the state they were in at version 3.2 to the
+#   state they should be in at version 3.3.
```
Ghostkeeper: Update copy-pasted documentation: Version 3.3 to 3.4.

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
```diff
@@ -0,0 +1,43 @@
...
+        return format_version * 1000000 + setting_version
+
+    ##  Upgrades instance containers to have the new version
+    #   number.
+    def upgradeInstanceContainer(self, serialized, filename):
+        parser = configparser.ConfigParser(interpolation = None)
+        parser.read_string(serialized)
+
+        # Update version number.
+        parser["general"]["version"] = "4"
```
Ghostkeeper: If all we do is updating the version number, why do we need to upgrade at all?

resources/themes/cura-light/theme.json
```diff
@@ -317,9 +321,11 @@
     },
 
     "sizes": {
+        "base_unit": [1.0, 1.0],
```
Ghostkeeper: I still think that this is fairly useless... :confused: Especially now that when resizing this means resizing several buttons in the toolbox but nothing else.

resources/themes/cura-light/theme.json
```diff
@@ -317,9 +321,11 @@
...
         "window_minimum_size": [70, 50],
         "window_margin": [1.0, 1.0],
         "default_margin": [1.0, 1.0],
+        "double_margin": [2.0, 2.0],
```
Ghostkeeper: Rather choose a name like "wide margin". With the name "double margin" it always has to be twice the "default margin", so we might as well just hardcode `default_margin * 2` then. "wide margin" allows for some flexibility.

resources/themes/cura-light/theme.json
```diff
@@ -433,6 +439,10 @@
         "objects_menu_size": [20, 40],
         "objects_menu_size_collapsed": [20, 17],
         "build_plate_selection_size": [15, 5],
-        "objects_menu_button": [0.3, 2.7]
+        "objects_menu_button": [0.3, 2.7],
+
+        "toolbox_thumbnail_small": [6.0, 6.0],
+        "toolbox_thumbnail_medium": [8.0, 8.0],
+        "toolbox_thumbnail_large": [12.0, 12.0]
```
Ghostkeeper: Maybe it's an idea to merge this with the icon size, like the one we use for the top bar. If that is possible in the design constraints, at least...

plugins/Toolbox/resources/qml/ToolboxInstalledTile.qml
```diff
@@ -195,10 +188,13 @@ Item
...
-            anchors.left: updateButton.left
-            anchors.right: updateButton.right
-            anchors.top: updateButton.bottom
-            anchors.topMargin: 4
+            anchors
+            {
+                left: updateButton.left
+                right: updateButton.right
+                top: updateButton.bottom
+                topMargin: Math.floor(UM.Theme.getSize("base_unit") / 4)
```
Ghostkeeper: Use the default margin?

resources/themes/cura-light/theme.json
ianpaschal: I know. But the alternative is to just move what's basically in the code (`unit * x`) into the theme which makes the already massive file even longer and harder to use. Plus then all these arbitrary names need to be thought up. I know they shouldn't be arbitrary in theory but let's be honest: it's pretty much impossible to make more descriptive terms out of all these "spacing lining margin default whatevers"

resources/themes/cura-light/theme.json
ianpaschal: Fair point. Will change.

plugins/Toolbox/resources/qml/ToolboxInstalledTile.qml
ianpaschal: Too big. Needs to be 0.25 units, default margin is 1. Could define yet another margin on a case by case basis but don't really see the point of polluting the theme with that.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+            "materials_showcase": None
+        }
+
+        # Data:
+        self._metadata = {
+            "authors": [],
+            "packages": [],
+            "plugins_showcase": [],
+            "plugins_installed": [],
+            # TODO: Replace this with a proper API call:
```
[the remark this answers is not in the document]
ianpaschal: Can't until the API call is possible from Cloud's end. 😉

plugins/Toolbox/src/Toolbox.py
ianpaschal: Yes. Unfortunately this is the super hacky stuff that will be replaced by actually just making an API call for authors. But since it doesn't exist, I used this method to get a pretty-much-always accurate list.

plugins/Toolbox/src/Toolbox.py
ianpaschal: Yup. Artifact from before I decided "why not just create it on init?"

plugins/Toolbox/src/Toolbox.py
ianpaschal: Yup. Artifact from before I decided "why not just create it on init?"

plugins/Toolbox/resources/qml/ToolboxInstalledTile.qml
fieldOfView: >  don't really see the point of polluting the theme with that.

But that's the whole point of theming. By hardcoding the 4, you basically make it unthemeable.

I would argue that if just one element needs this adjusted margin, then the design is using margins inconsistently.

ianpaschal: >I would argue that if just one element needs this adjusted margin, then the design is using margins inconsistently.

Luke is quite good about keeping things consistent. This is one spot in the code but many places in the interface.

What I mean about polluting the theme is what I see as making it too granular. Since the theme doesn't cascade and can't inherit element-specific measurements from more general ones, we end up having to make a trade off between "what if I want to generally change one size, then I have to try to figure out everywhere where that size is used and update 10 different values in the theme because each one is defined separately" which is annoying to some vs. "what if I want to make everything a random size untethered from everything else, but I can't because everything is proportional to each other" which would be annoying to others.

I'd love to theme Cura but I won't be investing the time into it because I find it incredibly bothersome because it's so granular. I can change these to all have their own definition but I don't think it's definitely not UM.Theme "more useful."

My plan for the research sprint is to re-write how lots of UI stuff is handled so that it can have cascading/inheritance, comments, x-ref'd values, etc.

resources/themes/cura-light/theme.json
ianpaschal: I don't think so because they're not really "icons" per say. They're also way bigger than any 12-20 px icon in the UI.

cura/Utils/VersionTools.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2018 Ultimaker B.V.
```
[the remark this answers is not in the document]
ianpaschal: Fixed in fbf7317

plugins/GCodeGzReader/GCodeGzReader.py
```diff
@@ -3,19 +3,19 @@
...
+
 ##  A file reader that reads gzipped g-code.
 #
 #   If you're zipping g-code, you might as well use gzip!
 class GCodeGzReader(MeshReader):
 
     def __init__(self):
         super().__init__()
-        self._supported_extensions = [".gcode.gz"]
+        self._supported_extensions = [".gz"]
```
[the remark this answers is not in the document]
ianpaschal: Good question. @LipuFei?

plugins/Toolbox/plugin.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Toolbox",
+    "author": "Ultimaker B.V.",
+    "version": "1.0.0",
+    "api": 4,
+    "description": "Find, manage and install new cura packages."
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 7f622eb

plugins/Toolbox/resources/qml/ToolboxAuthorPage.qml
```diff
@@ -0,0 +1,130 @@
...
+                leftMargin: UM.Theme.getSize("default_margin").width
+                right: parent.right
+                rightMargin: UM.Theme.getSize("double_margin").width
+                bottomMargin: UM.Theme.getSize("default_margin").height
+            }
+            text: details.name
+            font: UM.Theme.getFont("large")
+            wrapMode: Text.WordWrap
+            width: parent.width
+            height: UM.Theme.getSize("base_unit") * 2
```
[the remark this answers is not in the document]
ianpaschal: Not in my opinion. Then your type size starts to screw up your grid layout.

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
```diff
@@ -0,0 +1,142 @@
...
+                leftMargin: UM.Theme.getSize("default_margin").width
+                right: parent.right
+                rightMargin: UM.Theme.getSize("double_margin").width
+                bottomMargin: UM.Theme.getSize("default_margin").height
+            }
+            text: details.name
+            font: UM.Theme.getFont("large")
+            wrapMode: Text.WordWrap
+            width: parent.width
+            height: UM.Theme.getSize("base_unit") * 2
```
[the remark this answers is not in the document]
ianpaschal: Nah. From above:

> Not in my opinion. Then your type size starts to screw up your grid layout.

More specifically, typography 101: your type's position shouldn't be determined by its size.

I would change it to use the lineHeight property but if it works the result is the same and if not, it makes a mess of things.

resources/themes/cura-light/theme.json
ianpaschal: Fixed in 668a929

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
```diff
@@ -0,0 +1,142 @@
...
+            width: UM.Theme.getSize("base_unit").width * 12
+            Label
+            {
+                text: details.version
+                font: UM.Theme.getFont("very_small")
+                color: UM.Theme.getColor("text")
+            }
+            Label
+            {
+                text: Qt.formatDateTime(details.last_updated, "dd MMM yyyy")
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 668a929

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
```diff
@@ -0,0 +1,142 @@
...
+    Rectangle
+    {
+        id: header
+        anchors
+        {
+            left: sidebar.right
+            right: parent.right
+            rightMargin: UM.Theme.getSize("double_margin").width
+        }
+        height: UM.Theme.getSize("base_unit").height * 12
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 668a929

plugins/Toolbox/resources/qml/ToolboxAuthorPage.qml
```diff
@@ -0,0 +1,130 @@
...
+            id: values
+            anchors
+            {
+                top: title.bottom
+                left: properties.right
+                leftMargin: UM.Theme.getSize("default_margin").width
+                topMargin: UM.Theme.getSize("default_margin").height
+            }
+            spacing: Math.floor(UM.Theme.getSize("default_margin").height/2)
+            width: UM.Theme.getSize("base_unit").width * 12
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 668a929

plugins/Toolbox/resources/qml/ToolboxAuthorPage.qml
```diff
@@ -0,0 +1,130 @@
...
+    Rectangle
+    {
+        id: header
+        anchors
+        {
+            left: sidebar.right
+            right: parent.right
+            rightMargin: UM.Theme.getSize("double_margin").width
+        }
+        height: UM.Theme.getSize("base_unit").height * 12
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 668a929

plugins/Toolbox/plugin.json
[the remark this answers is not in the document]
ianpaschal: Fixed in 668a929

cura/Utils/VersionTools.py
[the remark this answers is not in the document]
ianpaschal: Fixed in

resources/themes/cura-light/theme.json
ianpaschal: Fixed in 9b556e9

cura/Utils/VersionTools.py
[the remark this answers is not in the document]
fieldOfView: So VersionTools.py can be removed, right?

cura/Utils/VersionTools.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Make sure that two versions have the same number of parts. For missing parts, just append 0s.
+    for parts in (version1_parts, version2_parts):
+        for _ in range(max_part_length - len(parts)):
+            parts.append(0)
+
+    # Compare the version parts and return the result
+    result = 0
+    for idx in range(max_part_length):
+        result = version1_parts[idx] - version2_parts[idx]
```
jackha: Aha! So the result is not necessarily one of [-1, 0, +1] as the description says.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,613 @@
...
+        self._download_reply = None
+        self._download_request = None
+        self.setDownloadProgress(0)
+        self.setIsDownloading(False)
+
+
+
+    # Handlers for Network Events
+    # --------------------------------------------------------------------------
+    def _onNetworkAccesibleChanged(self, accessible: int):
```
jackha: Typo in function name, should be accessible.

plugins/Toolbox/src/Toolbox.py
jackha: Yup. Artifact from before @ianpaschal decided "why not just create it on init?"

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,613 @@
...
+
+        else:
+            # Ignore any operation that is not a get operation
+            pass
+
+    def _onDownloadProgress(self, bytes_sent: int, bytes_total: int):
+        if bytes_total > 0:
+            new_progress = bytes_sent / bytes_total * 100
+            self.setDownloadProgress(new_progress)
+            if new_progress == 100.0:
```
jackha: Better have a habit of not comparing floats with an exact number using ==. In this case I'd rather use just bytes_sent and bytes_total.

plugins/GCodeGzReader/GCodeGzReader.py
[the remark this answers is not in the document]
LipuFei: Should be `.gcode.gz`.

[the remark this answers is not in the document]
LipuFei: Fixed in cbc45bffd

plugins/GCodeGzReader/GCodeGzReader.py
```diff
@@ -3,19 +3,19 @@
 
 import gzip
 
-from io import TextIOWrapper
-
+from UM.Platform import Platform
```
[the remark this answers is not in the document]
LipuFei: Removed in 48eda4d0d

plugins/Toolbox/src/Toolbox.py
LipuFei: Fixed in 3ea80c7c7

plugins/Toolbox/src/Toolbox.py
ianpaschal: Fixed in ec8b955

cura/Utils/VersionTools.py
ianpaschal: This file isn't used anymore. Per an earlier suggestion from Ruben we use UM.Version.

Removed from code earlier, but file removed in 332b7fb

plugins/Toolbox/src/PackagesModel.py
ianpaschal: I'm going to leave it since the filtering may actually be removed eventually once more specific API calls are possible.

cura/Utils/VersionTools.py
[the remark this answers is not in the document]
ianpaschal: Removed in 332b7fb

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+from UM.Resources import Resources
+
+from cura.Utils import VersionTools
+
+
+class CuraPackageManager(QObject):
+
+    # The prefix that's added to all files for an installed package to avoid naming conflicts with user created
+    # files.
+    PREFIX_PLACE_HOLDER = "-CP;"
```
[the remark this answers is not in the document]
ianpaschal: @LipuFei

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        self._installAllScheduledPackages()
+
+    # (for initialize) Loads the package management file if exists
+    def _loadManagementData(self) -> None:
+        if not os.path.exists(self._package_management_file_path):
+            Logger.log("i", "Package management file %s doesn't exist, do nothing", self._package_management_file_path)
+            return
+
+        with open(self._package_management_file_path, "r", encoding = "utf-8") as f:
+            management_dict = json.loads(f.read(), encoding = "utf-8")
```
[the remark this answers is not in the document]
ianpaschal: @LipuFei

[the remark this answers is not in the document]
ianpaschal: @LipuFei

cura/CuraPackageManager.py
ianpaschal: @LipuFei

cura/CuraPackageManager.py
ianpaschal: @LipuFei

plugins/Toolbox/resources/qml/ToolboxAuthorPage.qml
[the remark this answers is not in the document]
Ghostkeeper: But now your type size starts to screw up the readability since text will start to overlap if it's too large.

resources/themes/cura-light/theme.json
```diff
@@ -433,6 +441,24 @@
...
-        "objects_menu_button": [0.3, 2.7]
+        "objects_menu_button": [0.3, 2.7],
+
+        "toolbox_thumbnail_small": [6.0, 6.0],
+        "toolbox_thumbnail_medium": [8.0, 8.0],
+        "toolbox_thumbnail_large": [12.0, 12.0],
+        "toolbox_footer": [1.0, 4.5],
+        "toolbox_footer_button": [8.0, 2.5],
+        "toolbox_showcase_spacing": [1.0, 1.0],
+        "toolbox_header_tab": [8.0, 4.0],
```
Ghostkeeper: I'm thinking that the toolbox header theming should be merged with the top bar theming. They should look/feel the same, so we can enforce that.

ianpaschal: You're killing me here Ruben! After all that conversation I added this so that this part of the UI is individually theme able if the user wants to theme it differently.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+
+
+    # Make API Calls
+    # --------------------------------------------------------------------------
+    def _makeRequestByType(self, type):
+        Logger.log("i", "Toolbox: Requesting %s metadata from server.", type)
+        request = QNetworkRequest(self._request_urls[type])
+        request.setRawHeader(*self._request_header)
+        self._network_manager.get(request)
+    # TODO: Request authors and request material showcase
```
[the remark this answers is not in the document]
ianpaschal: It is now!

Fixed in 0100794

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+            "plugins_installed": PackagesModel(self),
+            "materials_showcase": AuthorsModel(self),
+            "materials_installed": PackagesModel(self)
+        }
+
+        # These properties are for keeping track of the UI state:
+        # ----------------------------------------------------------------------
+        # View category defines which filter to use, and therefore effectively
+        # which category is currently being displayed. For example, possible
+        # values include "plugin" or "material", but also "installed".
```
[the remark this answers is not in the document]
ianpaschal: Meh I'm going to leave it. I also don't like arbitrarily using enums for things that don't really have an objective order to them.

plugins/Toolbox/resources/qml/ToolboxAuthorPage.qml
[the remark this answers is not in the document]
ianpaschal: Can't sacrifice the entire integrity of the grid for one line of text. If you want to make the text giant, you should also allot more space to that element.

Sorry but this is a change I'm not making. I don't think I was hired because I'm such a self-taught programmer but because I have formal training and professional experience as a designer. I'm not about to shoot Typography 101 in the face just for 1 less line of code. 😛

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,357 @@
...
+        archive = zipfile.ZipFile(filename, "r")
+
+        temp_dir = tempfile.TemporaryDirectory()
+        archive.extractall(temp_dir.name)
+
+        from cura.CuraApplication import CuraApplication
+        installation_dirs_dict = {
+            "materials": Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer),
+            "quality": Resources.getStoragePath(CuraApplication.ResourceTypes.QualityInstanceContainer),
+            "plugins": os.path.join(os.path.abspath(Resources.getDataStoragePath()), "plugins"),
```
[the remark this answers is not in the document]
ianpaschal: @LipuFei

cura/CuraPackageManager.py
ianpaschal: @LipuFei

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
ianpaschal: We can delete the whole version upgrade file right, @LipuFei? Or is there another reason you added it?

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
ianpaschal: No idea. Why did you add it @LipuFei?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+                self._download_reply.downloadProgress.disconnect(self._onDownloadProgress)
+                self._download_reply.abort()
+                self._download_reply = None
+
+    def _onRequestFinished(self, reply):
+
+        if reply.error() == QNetworkReply.TimeoutError:
+            Logger.log("w", "Got a timeout.")
+            # Reset everything.
+            self.setDownloadProgress(0)
```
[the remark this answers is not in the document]
ianpaschal: Fixed bb7f9fe

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+            self.setIsDownloading(False)
+            if self._download_plugin_reply:
+                self._download_plugin_reply.downloadProgress.disconnect(self._onDownloadProgress)
+                self._download_plugin_reply.abort()
+                self._download_plugin_reply = None
```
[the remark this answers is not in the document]
ianpaschal: Fixed bb7f9fe

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,609 @@
...
+    # Handlers for Network Events
+    # --------------------------------------------------------------------------
+    def _onNetworkAccesibleChanged(self, accessible):
+        if accessible == 0:
+            self.setDownloadProgress(0)
+            self.setIsDownloading(False)
+            if self._download_reply:
+                self._download_reply.downloadProgress.disconnect(self._onDownloadProgress)
+                self._download_reply.abort()
+                self._download_reply = None
```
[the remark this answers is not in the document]
ianpaschal: Fixed bb7f9fe

cura/CuraPackageManager.py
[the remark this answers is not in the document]
LipuFei: Changed in cff4dc8a0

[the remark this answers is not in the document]
LipuFei: Used lock in 1fb2edace

cura/CuraPackageManager.py
LipuFei: Used lock in 1fb2edace

cura/CuraPackageManager.py
LipuFei: Fixed in 51abd3232

cura/CuraPackageManager.py
LipuFei: Now it's checking all subdirs in the data storage dir for package removal in 7ee085fe5

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
LipuFei: To trigger a version upgrade so we can move and rename the existing custom profiles, materials, etc. to have their own namespaces (dirs).

cura/CuraPackageManager.py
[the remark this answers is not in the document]
LipuFei: Fixed in 863d67ae6 and https://github.com/Ultimaker/Uranium/commit/0c50d7c9d33954ef297fc6bdd83fc8140f388084

plugins/Toolbox/resources/qml/ToolboxInstalledTile.qml
```diff
@@ -0,0 +1,219 @@
...
+            right: pluginActions.left
+            rightMargin: UM.Theme.getSize("default_margin").width
+        }
+        Label
+        {
+            text:
+            {
+                if (model.author_email)
+                {
+                    return "<a href=\"mailto:"+model.author_email+"?Subject=Cura: "+model.name+"\">"+model.author_name+"</a>"
```
Ghostkeeper: We are reviewing these e-mails and names to test for code injection, right? :stuck_out_tongue:

plugins/Toolbox/resources/qml/ToolboxLicenseDialog.qml
```diff
@@ -0,0 +1,72 @@
...
+        TextArea
+        {
+            id: licenseText
+            anchors.top: licenseTitle.bottom
+            anchors.bottom: parent.bottom
+            anchors.left: parent.left
+            anchors.right: parent.right
+            anchors.topMargin: UM.Theme.getSize("default_margin").height
+            readOnly: true
+            text: licenseDialog.licenseContent != null ? licenseDialog.licenseContent : ""
```
Ghostkeeper: `licenseContent` can never be `null`, right?

plugins/Toolbox/resources/qml/ToolboxLoadingPage.qml
```diff
@@ -0,0 +1,22 @@
...
+
+Rectangle
+{
+    id: page
+    width: parent.width
+    height: parent.height
+    color: "transparent"
+    Label
+    {
+        text: "Fetching packages..."
```
Ghostkeeper: Needs to be translatable.

plugins/Toolbox/resources/qml/ToolboxShadow.qml
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2018 Ultimaker B.V.
+// Toolbox is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.2
+
+Rectangle
+{
+    property bool reversed: false
```
Ghostkeeper: Did you intend to do something with this? It seems to always be `false`.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
```diff
@@ -6,19 +6,20 @@ import QtQuick.Controls 1.4
...
         frameVisible: false
-        anchors.fill: base
+        anchors.fill: detailList
         style: UM.Theme.styles.scrollview
         Column
         {
             anchors
             {
-                right: base.right
+                right: detailList.right
```
diegopradogesto: QML says: "Cannot anchor to an item that isn't a parent or sibling". It must be `parent.right`

plugins/Toolbox/src/Toolbox.py
```diff
@@ -368,14 +368,17 @@ def _onRequestFinished(self, reply: QNetworkReply):
                         # HACK: Eventually get rid of the code from here...
                         if type is "plugins_showcase" or type is "materials_showcase":
                             self._metadata["plugins_showcase"] = json_data["data"]["plugin"]["packages"]
+                            self._models["plugins_showcase"].setMetadata(self._metadata["plugins_showcase"])
                             self._metadata["materials_showcase"] = json_data["data"]["material"]["authors"]
+                            self._models["materials_showcase"].setMetadata(self._metadata["materials_showcase"])
+                            print("MATERIAL SHOWCASE", self._metadata["materials_showcase"])
```
diegopradogesto: Remember to remove print commands.

plugins/Toolbox/resources/qml/ToolboxShadow.qml
ianpaschal: Yes. It's possible to overwrite when creating an instance of it. I think in some place I set "reversed: true" when using an instance.

plugins/Toolbox/resources/qml/ToolboxLicenseDialog.qml
ianpaschal: I think it can. It typically is None unless there is a license which is only the case with some plugins, but never with materials.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2018 Ultimaker B.V.
+// Toolbox is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.2
+import QtQuick.Controls 1.4
+import QtQuick.Controls.Styles 1.4
+import UM 1.1 as UM
+
+Item
+{
```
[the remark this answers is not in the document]
ianpaschal: I can try and double check but generally if there is two ways to do it, I did the only one that worked. To me, the hallmark of QML is "stuff that looks like it should work but just doesn't." This is the result of that.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
```diff
@@ -0,0 +1,37 @@
...
+    ScrollView
+    {
+        frameVisible: false
+        anchors.fill: detailList
+        style: UM.Theme.styles.scrollview
+        Column
+        {
+            anchors
+            {
+                right: detailList.right
```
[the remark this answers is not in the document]
diegopradogesto: I also commented in this line. QML says: "Cannot anchor to an item that isn't a parent or sibling". It must be anchored to parent.

plugins/Toolbox/src/PackagesModel.py
```diff
@@ -43,13 +43,14 @@ def _update(self):
...
         for package in self._metadata:
+            print(package["author"])
             items.append({
                 "id":                package["package_id"],
                 "type":              package["package_type"],
                 "name":              package["display_name"],
                 "version":           package["package_version"],
-                "author_id":         package["author"]["author_id"],
-                "author_name":       package["author"]["display_name"],
+                "author_id":         package["author"]["author_id"] if "author_id" in package["author"] else package["author"]["name"],
```
diegopradogesto: Does the filed "name" in "author" exist? Otherwise it will crash.

cura/CuraPackageManager.py
```diff
@@ -245,22 +245,17 @@ def removePackage(self, package_id: str) -> None:
...
+        # Iterate through all directories in the data storage directory and look for sub-directories that belong to
+        # the package we need to remove, that is the sub-dirs with the package_id as names, and remove all those dirs.
+        data_storage_dir = os.path.abspath(Resources.getDataStoragePath())
+
+        for root, dir_names, _ in os.walk(data_storage_dir):
+            for dir_name in dir_names:
+                package_dir = os.path.join(root, dir_name, package_id)
+                if os.path.exists(package_dir):
+                    Logger.log("i", "Removing '%s' for package [%s]", package_dir, package_id)
                 shutil.rmtree(package_dir)
```
diegopradogesto: Should this be within the previous `if` statement?

plugins/Toolbox/src/PackagesModel.py
ianpaschal: This will be changed with an update I have made to the packages repo which is awaiting merge. The logic here got a little convoluted because plugins used an `author_name` field while the packages API used: `author.display_name`

plugins/Toolbox/resources/qml/ToolboxShadow.qml
ianpaschal: Just checked. It's overwritten for the "restart bar" shadow.

plugins/Toolbox/resources/qml/ToolboxInstalledTile.qml
ianpaschal: Yes. Right now it's manually updated but this is eventually a job for whatever interface we build for this.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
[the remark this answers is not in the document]
ianpaschal: Eh I don't remember exactly but I couldn't get it behave properly without, as I recall.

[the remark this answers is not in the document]
ianpaschal: Well removing it breaks the layout.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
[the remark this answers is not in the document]
ianpaschal: Yeah I just checked. I think I have it because I needed to be able to make the scroll view fill to something.

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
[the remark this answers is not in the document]
ianpaschal: But I've changed it to anchor to the parent.

plugins/Toolbox/resources/qml/ToolboxLicenseDialog.qml
ianpaschal: Oh you are correct. This check needs to exist though, which is why it's *also* in Python. Removing it here. Fixed in 62aeb6c

plugins/Toolbox/resources/qml/ToolboxLoadingPage.qml
ianpaschal: Fixed in 62aeb6c

plugins/Toolbox/resources/qml/ToolboxFooter.qml
```diff
@@ -0,0 +1,68 @@
...
+Item
+{
+    id: footer
+    width: parent.width
+    anchors.bottom: parent.bottom
+    height: visible ? Math.floor(UM.Theme.getSize("toolbox_footer").height) : 0
+    Label
+    {
+        visible: toolbox.restartRequired
+        text: "You will need to restart Cura before changes in plugins have effect."
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 62aeb6c

plugins/Toolbox/resources/qml/ToolboxDownloadsGrid.qml
```diff
@@ -0,0 +1,42 @@
...
+import QtQuick.Controls 1.4
+import QtQuick.Controls.Styles 1.4
+import QtQuick.Layouts 1.3
+import UM 1.1 as UM
+
+Column
+{
+    // HACK: GridLayouts don't render to the correct height with odd numbers of
+    // items, so if odd, add some extra space.
+    height: grid.model.items.length % 2 == 0 ? childrenRect.height : childrenRect.height + 3 * UM.Theme.getSize("wide_margin").height
```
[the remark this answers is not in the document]
ianpaschal: Fixed in 62aeb6c

plugins/Toolbox/resources/qml/ToolboxDetailList.qml
ianpaschal: Fixed in 62aeb6c

cura/CuraPackageManager.py
ianpaschal: Fixed in 62aeb6c

plugins/Toolbox/src/Toolbox.py
ianpaschal: Fixed in 62aeb6c

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,515 @@
...
+    # Make API Calls
+    # --------------------------------------------------------------------------
+    def _makeRequestByType(self, type: str):
+        Logger.log("i", "Toolbox: Requesting %s metadata from server.", type)
+        request = QNetworkRequest(self._request_urls[type])
+        request.setRawHeader(*self._request_header)
+        self._network_manager.get(request)
+
+    @pyqtSlot(str)
+    def startDownload(self, url: str):
```
Ghostkeeper: This function needs to cancel any ongoing downloads, or fail or block if there are any ongoing downloads.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,515 @@
...
+        self.setDownloadProgress(0)
+        self.setIsDownloading(False)
+
+
+
+    # Handlers for Network Events
+    # --------------------------------------------------------------------------
+    def _onNetworkAccessibleChanged(self, accessible: int):
+        if accessible == 0:
+            self.resetDownload()
```
Ghostkeeper: What does the value of `0` mean? Could you document this function?

plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
```diff
@@ -0,0 +1,104 @@
...
+            Label
+            {
+                id: info
+                text:
+                {
+                    if (model.description.length > 50)
+                    {
+                        return model.description.substring(0, 50) + "..."
+                    }
+                    return model.description
```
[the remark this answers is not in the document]
ianpaschal: Fixed in c2ed2a4

plugins/Toolbox/src/Toolbox.py
```diff
@@ -0,0 +1,515 @@
...
+                        # TODO: Make multiple API calls in the future to handle this
+                        if type is "packages":
+                            self._models[type].setFilter({"type": "plugin"})
+                        if type is "authors":
+                            self._models[type].setFilter({"package_types": "material"})
+
+                        self.metadataChanged.emit()
+
+                        if self.loadingComplete() is True:
+                            self.setViewPage("overview")
```
Ghostkeeper: Why do we switch to the overview page if a package download is completed?

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
Ghostkeeper: Update copy-pasted documentation: Version 3.3 to 3.4.

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
```diff
@@ -0,0 +1,43 @@
...
+
+##  Upgrades configurations from the state they were in at version 3.2 to the
+#   state they should be in at version 3.3.
+class VersionUpgrade33to34(VersionUpgrade):
+
+    ##  Gets the version number from a CFG file in Uranium's 3.2 format.
+    #
+    #   Since the format may change, this is implemented for the 3.2 format only
+    #   and needs to be included in the version upgrade system rather than
+    #   globally in Uranium.
```
Ghostkeeper: More copy-pasted documentation is wrong.

plugins/Toolbox/resources/qml/ToolboxLicenseDialog.qml
Ghostkeeper: But if there is no license, the `ToolboxLicenseDialog` can never be shown, I was assuming.

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+    def _saveManagementData(self) -> None:
+        # Need to use the file lock here to prevent concurrent I/O from other processes/threads
+        container_registry = self._application.getContainerRegistry()
+        with container_registry.lockFile():
+            with open(self._package_management_file_path, "w", encoding = "utf-8") as f:
+                data_dict = {"installed": self._installed_package_dict,
+                             "to_remove": list(self._to_remove_package_set),
+                             "to_install": self._to_install_package_dict}
+                data_dict["to_remove"] = list(data_dict["to_remove"])
+                json.dump(data_dict, f)
```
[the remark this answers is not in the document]
LipuFei: Added in e47f47d8c

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+        if package_id in self._installed_package_dict:
+            package_info = self._installed_package_dict.get(package_id)
+            package_info["is_bundled"] = False
+            return package_info
+
+        for section, packages in self.getAllInstalledPackagesInfo().items():
+            for package in packages:
+                if package["package_id"] == package_id:
+                    package_info = package
+                    return package_info
```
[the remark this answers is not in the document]
LipuFei: Fixed in d4cf944e1

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+        self._saveManagementData()
+
+    # Checks the given package is installed. If so, return a dictionary that contains the package's information.
+    def getInstalledPackageInfo(self, package_id: str) -> Optional[dict]:
+        if package_id in self._to_remove_package_set:
+            return None
+
+        if package_id in self._to_install_package_dict:
+            package_info = self._to_install_package_dict[package_id]["package_info"]
+            package_info["is_bundled"] = False
```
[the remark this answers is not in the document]
LipuFei: Removed in 5cc5f32a7

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+                    package_info = package
+                    return package_info
+
+        return None
+
+    def getAllInstalledPackagesInfo(self) -> dict:
+        installed_package_id_set = set(self._installed_package_dict.keys()) | set(self._to_install_package_dict.keys())
+        installed_package_id_set = installed_package_id_set.difference(self._to_remove_package_set)
+
+        managed_package_id_set = set(installed_package_id_set) | self._to_remove_package_set
```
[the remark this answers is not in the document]
LipuFei: Done in 09af97634

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+        if to_install_package:
+            Logger.log("i", "Package [%s] version [%s] is scheduled to be installed.",
+                       package_id, package_info["package_version"])
+            # Copy the file to cache dir so we don't need to rely on the original file to be present
+            package_cache_dir = os.path.join(os.path.abspath(Resources.getCacheStoragePath()), "cura_packages")
+            if not os.path.exists(package_cache_dir):
+                os.makedirs(package_cache_dir, exist_ok=True)
+
+            target_file_path = os.path.join(package_cache_dir, package_id + ".curapackage")
+            shutil.copy2(filename, target_file_path)
```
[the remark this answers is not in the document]
LipuFei: Done in 49649e3d4

cura/CuraPackageManager.py
```diff
@@ -0,0 +1,360 @@
...
+
+            self._to_install_package_dict[package_id] = {"package_info": package_info,
+                                                         "filename": target_file_path}
+            has_changes = True
+
+        self._saveManagementData()
```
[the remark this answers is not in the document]
LipuFei: Done in 49649e3d4

plugins/Toolbox/resources/qml/ToolboxBackColumn.qml
```diff
@@ -0,0 +1,69 @@
...
+        {
+            toolbox.viewPage = "overview"
+            toolbox.filterModelByProp("packages", "type", toolbox.viewCategory)
+        }
+        style: ButtonStyle
+        {
+            background: Rectangle
+            {
+                color: "transparent"
+            }
```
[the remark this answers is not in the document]
ianpaschal: Not useless. QML renders white by default which will fuck up dark themes.

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
LipuFei: Fixed in ad3317a99

plugins/VersionUpgrade/VersionUpgrade33to34/VersionUpgrade33to34.py
LipuFei: Fixed in ad3317a99

plugins/Toolbox/src/Toolbox.py
ianpaschal: That's not what this does. This is loading the data. Switching from loading page.

plugins/Toolbox/src/Toolbox.py
ianpaschal: No idea. :/ That's something that wasn't changed by me. I think @sedwards2009 or @nallath wrote that a long time ago.

I could investigate in the network manager docs but that would have been pretty low on my priorities.

plugins/Toolbox/src/Toolbox.py
ianpaschal: Don't think it really does as it can't be called in the first place. Since it doesn't induce any bugs I won't be changing that for now.

plugins/Toolbox/src/PackagesModel.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def setMetadata(self, data):
+        self._metadata = data
+        self._update()
+
+    def _update(self):
+        items = []
+
+        for package in self._metadata:
+            print(package["author"])
```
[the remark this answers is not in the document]
LipuFei: Removed in 554ebb661

plugins/Toolbox/resources/qml/ToolboxBackColumn.qml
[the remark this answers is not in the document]
fieldOfView: Use background: Item {} then. That does not render white.

ChrisTerBeke:
Getting the same thing. Could be quite easily solved by using plain Python to download the package instead of the super weird Qt network manager.

ianpaschal:
I tried to avoid changing as much as possible and since the PluginBrowser used the QTNetworkManager, I continued using that. QNetworkRequest.RedirectPolicyAttribute was the only way Lipu and I could get the Toolbox to handle the redirect, without, I guess, writing our own solution.

fieldOfView:
Another way around this is not to put redirecting urls in the db.

ianpaschal:
But muh analytics!

ianpaschal:
Actually I think @ChrisTerBeke has a different reason for doing this.

ChrisTerBeke:
The redirect on the server side is done for tracking downloads indeed before redirecting to the actual storage location of the package file on the CDN. Any decent HTTP client should support this, but again, it's Qt, so it's not....

fieldOfView:
Well, currently you can either have your analytics, or have a working Cura. I know what I would choose.

Talking about CDNs, is there any way the system could work without giving my data to Google? Google does not need to know I opened Cura (and they do now, because Cura gets thumbnails from a google CDN). Can't Ultimaker self-host this?

ChrisTerBeke:
Google doesn't know anything. We're using GCP storage with is completely free of the usual tracking, since we're paying customers of that service. We even have a signed contract with Google that includes clauses like this.

fieldOfView:
Ok, please be very upfront about that when this feature lands in a public build.

ChrisTerBeke:
@Appesteijn Can you check if this (lawfully) needs to be added to the EULA or some other place where the user can see it?

Ghostkeeper:
Looks like the RedirectPolicyAttribute problem was solved by using a different redirect policy when you have Qt version 5.8: https://github.com/Ultimaker/Cura/blob/57ca41291a8859c4f0d0156c25d609e759e62e1e/plugins/Toolbox/src/Toolbox.py#L367-L372

ChrisTerBeke:
Yup, I added that after getting crashes when running from source using 5.8 :)

PR #3674, opened 2018-04-16
Fix and update Korean Translation

jbkim:


resources/i18n/ko_KR/cura.po
```diff
@@ -5,66 +5,48 @@
...
-#: /home/ruben/Projects/Cura/plugins/ModelChecker/ModelChecker.py:66
-#, python-brace-format
-msgctxt "@info:status"
-msgid ""
-"Some models may not be printed optimally due to object size and chosen material for models: {model_names}.\n"
-"Tips that may be useful to improve the print quality:\n"
-"1) Use rounded corners.\n"
-"2) Turn the fan off (only if there are no tiny details on the model).\n"
-"3) Use a different material."
-msgstr "일부 모델은 개체 크기 및 다음 모델에 대해 선택한 재료로 인해 최적으로 인쇄되지 않을 수 있습니다. {model_names}.\n인쇄 품질을 향상하는 데 유용한 팁:\n1) 둥근 모서리를 사용합니다.\n2) 팬을 끕니다(모델에 대해 세부 정보가 거의 없는 경우에만).\n3) 다른 재료를 사용합니다."
```
Ghostkeeper: :( It looks like you've translated Cura 3.2 instead of Cura 3.3. These translations are outdated!

jbkim: Hmm.  Ok, I'll check it again.

Ghostkeeper:
Looks like this was a lot of work. Many improvements were made! Thank you.

We've looked through these and found a few places where you replaced a translation with the English original. We fixed those here: https://github.com/Ultimaker/Cura/commit/132efba2727ecea2e05ef2fc03943898c81f6f50
I hope you'll agree.

jbkim:
Most translation is correct but there are some misspelled Korean word and wrong translation.

PR #3688, opened 2018-04-19
Felix profile

kerog777:
Adding profile info for the Felix Tec4 printer.  Got OK from vendor as long as I make it clear that this is a user contribution and not an "official" profile.  I did that by marking author as my email address and putting a "modified by" note in the start and end gcode comments.

I've tested the profile and verified that it works OK.  I'm still fuzzy on what's supposed to go in the version fields in the .json files, though.


diegopradogesto:
Devs. CURA-5256

kerog777:
I already have preferred variant in the metadata section, don't I?

kerog777:
Can you explain the version number?  You are saying it has to be something
in particular, but I have no idea why...

fieldOfView:
> I already have preferred variant in the metadata section, don't I?

It needs to be ```preferred_variant_name``` in Cura 3.3 and newer, and it must specify the name instead of the id.

> Can you explain the version number? You are saying it has to be something in particular, but I have no idea why...

The version and setting_version must match the other files in the same folder, so Cura knows the definition is up to date with what it expects. The exact data in definitions files like this can change between versions of Cura.

resources/variants/felixtec4_0.25.inst.cfg
```diff
@@ -6,8 +6,9 @@ definition = felixtec4dual
 [metadata]
 author = kerog777
 type = variant
-setting_version = 1
+setting_version = 4
 
 [values]
+hardware_type = nozzle
```
diegopradogesto: The `hardware_type = nozzle` must be in the `[metadata]` section, not in the `[values]`. You can take a look at some of the built-in variants, such as https://github.com/Ultimaker/Cura/blob/master/resources/variants/ultimaker_s5_aa04.inst.cfg

resources/definitions/felixtec4dual.def.json
```diff
@@ -0,0 +1,59 @@
...
+        "platform_offset": [-128,0, 102],
+        "machine_extruder_trains":
+        {
+            "0": "felixtec4_dual_extruder_0",
+            "1": "felixtec4_dual_extruder_1"
+        },
+        "file_formats": "text/x-gcode",
+        "has_variants": true,
+        "has_materials": true,
+        "preferred_variant_name": "felixtec4_0.35",
```
LipuFei: It's a bit confusing but `preferred_variant_name` is actually the name. So here it should be `0.25 mm`

LipuFei: Ah, sorry, it's "0.35 mm"

PR #3691, opened 2018-04-21
Only open USB ports if the printer understands gcode.

ohrn:
Hi!

This fix prevents the USB print manager from needlessly tying up the serial port on printers that do not support gcode.

The current behavior with extended baud rate probing is harmless (at least on the ZYYX Agile) but annoying. I don't know your policy on which bugs are severe enough to go into the release, feel free to put this on master if that is more appropriate.


fieldOfView:
Note that the port will still be opened, but printing over USB will just not be available.

ohrn:
@fieldOfView  Hehe, you had me confused there for a second.

Anyway, in case anyone wonders why I didn't just set supports_usb_connection to false on the affected printers I'm investigating adding USB print support for X3G printers.

This fix is the first small result of that.

Ghostkeeper:
Developers, see issue CURA-5340.

PR #3699, opened 2018-04-23
PP formula based temperatures fix

THeijmans:
Fixed the temperatures so they are calculated from the material file. No more exclamation marks.


ChrisTerBeke:
It breaks the CI though...

THeijmans:
I overlooked the AA0.25 variant profiles, but the 0.8mm profiles were already like this.

PR #3765, opened 2018-05-02
Fix toolbox appearance

fieldOfView:
This PR fixes a number of minor issues in the new Toolbox:
* improves thumbnail rendering
* fixes text rendering (OSX and on dark themes)
* fixes binding loop and other warnings

### DIFF
--- a/plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
@@ -36,8 +36,8 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode
-                text: styleData.value
+                elide: styleData.elideMode || Text.ElideNone
+                text: styleData.value || ""
                 color: UM.Theme.getColor("text")
                 font: UM.Theme.getFont("default_bold")
             }
@@ -55,8 +55,8 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode
-                text: styleData.value
+                elide: styleData.elideMode || Text.ElideNone
+                text: styleData.value || ""
                 color: UM.Theme.getColor("text_medium")
                 font: UM.Theme.getFont("default")
             }
@@ -67,8 +67,8 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode
-                text: styleData.value
+                elide: styleData.elideMode || Text.ElideNone
+                text: styleData.value || ""
                 color: UM.Theme.getColor("text_medium")
                 font: UM.Theme.getFont("default")
             }
--- a/plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
@@ -36,7 +36,7 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode || Text.ElideNone
+                elide: Text.ElideRight
                 text: styleData.value || ""
                 color: UM.Theme.getColor("text")
                 font: UM.Theme.getFont("default_bold")
@@ -55,7 +55,7 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode || Text.ElideNone
+                elide: Text.ElideRight
                 text: styleData.value || ""
                 color: UM.Theme.getColor("text_medium")
                 font: UM.Theme.getFont("default")
@@ -67,7 +67,7 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode || Text.ElideNone
+                elide: Text.ElideRight
                 text: styleData.value || ""
                 color: UM.Theme.getColor("text_medium")
                 font: UM.Theme.getFont("default")
--- a/plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
@@ -1,7 +1,7 @@
 // Copyright (c) 2018 Ultimaker B.V.
 // Toolbox is released under the terms of the LGPLv3 or higher.
 
-import QtQuick 2.2
+import QtQuick 2.3
 import QtQuick.Controls 1.4
 import QtQuick.Controls.Styles 1.4
 import UM 1.1 as UM
@@ -33,6 +33,7 @@ Item
             height: UM.Theme.getSize("toolbox_thumbnail_medium").height
             fillMode: Image.PreserveAspectFit
             source: details.icon_url || "../images/logobot.svg"
+            mipmap: true
             anchors
             {
                 top: parent.top
--- a/plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
@@ -135,6 +135,7 @@ Item
                 }
                 font: UM.Theme.getFont("very_small")
                 color: UM.Theme.getColor("text")
+                linkColor: UM.Theme.getColor("text_link")
                 onLinkActivated: Qt.openUrlExternally(link)
             }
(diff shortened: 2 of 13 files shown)


ChrisTerBeke:
You're too fast man! :P

ChrisTerBeke:
@ianpaschal this fixes that binding loop warning that's endless :P

plugins/Toolbox/resources/qml/ToolboxLicenseDialog.qml
```diff
@@ -42,7 +42,7 @@ UM.Dialog
             anchors.right: parent.right
             anchors.topMargin: UM.Theme.getSize("default_margin").height
             readOnly: true
-            text: licenseDialog.licenseContent
+            text: licenseDialog.licenseContent || ""
```
ianpaschal: I don't think this one is needed as licenseDialog is already set to an empty string if none exists and is therefore not shown. Shouldn't need to check for it again in QML.

plugins/Toolbox/resources/qml/ToolboxDownloadsShowcaseTile.qml
```diff
@@ -1,15 +1,15 @@
...
+import QtQuick 2.3
 import QtQuick.Controls 1.4
 import QtQuick.Controls.Styles 1.4
 import UM 1.1 as UM
 
 Item
 {
     width: UM.Theme.getSize("toolbox_thumbnail_large").width
-    height: childrenRect.height
+    height: thumbnail.height + packageName.height
```
ianpaschal: Out of curiosity, why change this?

plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
```diff
@@ -34,10 +34,11 @@ Item
             Image
             {
                 anchors.centerIn: parent
-                width: UM.Theme.getSize("toolbox_thumbnail_small").width - 26
-                height: UM.Theme.getSize("toolbox_thumbnail_small").height - 26
+                width: UM.Theme.getSize("toolbox_thumbnail_small").width - UM.Theme.getSize("default_margin").width
```
ianpaschal: Good catch. Only thing is that default margin is 13px so this should be minus `"wide_margin"`

plugins/Toolbox/resources/qml/ToolboxDetailTile.qml
```diff
@@ -80,8 +80,8 @@ Item
             {
                 background: Rectangle
                 {
-                    implicitWidth: 96
-                    implicitHeight: 30
+                    implicitWidth: 96 * screenScaleFactor
```
ianpaschal: These should actually use a standardized button class. Obviously that's on me as this is a remnant of my early Plugin Browser tweaks but if we're going to change it; let's change it properly.

ianpaschal: Do you think we should use an existing button style? Or make a custom one (i.e. a `"toolbox_button"` style) so that it can be more independently modified?

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
```diff
@@ -33,6 +33,7 @@ Item
             height: UM.Theme.getSize("toolbox_thumbnail_medium").height
             fillMode: Image.PreserveAspectFit
             source: details.icon_url || "../images/logobot.svg"
+            mipmap: true
```
ianpaschal: Nifty!

plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
```diff
@@ -36,8 +36,8 @@ Item
             Label
             {
                 anchors.verticalCenter: parent.verticalCenter
-                elide: styleData.elideMode
-                text: styleData.value
+                elide: styleData.elideMode || Text.ElideNone
```
ianpaschal: styleData is something built into QT's table views. is there a reason to add the `|| Text.ElideNone`? Seems like if styleData is also not sufficient for some reason (although I'm pretty sure it will *always* be set) that we should just be able to use `Text.ElideNone` in all cases?

ianpaschal:
Sorry I did a proper review now and I have some remarks/questions.

plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
fieldOfView: > (although I'm pretty sure it will always be set)

Apparently it isn't. Queue ubiquitous "Qt is garbage" comment...
I think styleData can become undefined if a model "up the stream" becomes undefined.

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
fieldOfView: It does require a slightly newer qtquick (2.3 vs 2.2), but that seems to be available with pyqt 5.8 (and newer)

plugins/Toolbox/resources/qml/ToolboxDownloadsShowcaseTile.qml
fieldOfView: childrenRect is a nice concept, but it leads to binding errors when doing too many creative anchorings (ie: a child anchor to the parent when the parent size depends on the positioning of the children).

fieldOfView: childrenRect is safe to use for eg columns and rows because those don't rely on anchoring.

plugins/Toolbox/resources/qml/ToolboxDetailTile.qml
fieldOfView: How about "toolbox_action_button"?

plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
ianpaschal: Haha you know me so well.

But in that case let's just always set it to use Text.ElideRight. The data that will be here is quite predictable.

plugins/Toolbox/resources/qml/ToolboxDetailPage.qml
ianpaschal: Yes we use 2.3 for some other things as well as I recall.

plugins/Toolbox/resources/qml/ToolboxDetailTile.qml
ianpaschal: Sure. But you think a separate button style is preferable to using the existing ones? (I'm really asking because I feel constantly on the fence about which UI elements should be able to be independently styled and which ones should be linked to a default style.)

plugins/Toolbox/resources/qml/ToolboxDownloadsShowcaseTile.qml
ianpaschal: Good to know.

plugins/Toolbox/resources/qml/ToolboxDetailTile.qml
fieldOfView: Fixed in 89322b9

plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
fieldOfView: Fixed in 1f675f5

plugins/Toolbox/resources/qml/ToolboxCompatibilityChart.qml
fieldOfView: Fixed in 4e30507

fieldOfView:
I'll see if I can fix the remaining QML warnings in a followup PR (but feel free to beat me to it)

plugins/Toolbox/resources/qml/ToolboxDownloadsShowcaseTile.qml
ChrisTerBeke: Unless your column/row items have a variable/collapsible height :P

ianpaschal:
Tuesday we have an exploratory test session planned so I'm not going to try and fix any bugs until we do that and make a comprehensive list.

Also I'm now in the thick of handling packages-as-upgrades-to-bundled-resources-without-destroying-existing-installations.

PR #3785, opened 2018-05-07
CURA-5164 Make application initialization and start up more clear

jackha:
- Create SingleInstance class to handling single instance stuff.
 - Instead of calling getInstance() everywhere, initialize each object
   explicitly in order when application starts and getInstance()s do not
   create instances any more and they merely return the created
   instances.
 - Only set initial values in construtor functions __init__(). Move the
   initialization of context-aware (i.e. things that depend on other
   things) to separate functions.
 - Split application creation and initialziation into several steps and
   them should be called explicitly in the correct order.


cura/CuraApplication.py
```diff
@@ -725,8 +695,12 @@ def run(self):
 
         self.exec_()
 
+    def __setUpSingleInstanceServer(self):
+        if self._use_single_instance:
```
jackha: Seems like a bit of an overkill to put startServer in a function and check _use_single_single two times (it is also checked before calling this function)

cura/CuraPackageManager.py
```diff
@@ -27,8 +27,8 @@ def __init__(self, parent = None):
         self._plugin_registry = self._application.getPluginRegistry()
 
         #JSON files that keep track of all installed packages.
-        self._user_package_management_file_path = None
-        self._bundled_package_management_file_path = None
+        self._user_package_management_file_path = None #type: str
```
ChrisTerBeke: Should be `# type: Optional[str]`

cura/CuraPackageManager.py
```diff
@@ -27,8 +27,8 @@ def __init__(self, parent = None):
...
+        self._bundled_package_management_file_path = None #type: str
```
ChrisTerBeke: Should be `# type: Optional[str]`

PR #3786, opened 2018-05-07
Backup Manager and API

ChrisTerBeke:
* Add a new API class that acts as a version-safe layer between Cura code and plugins (discussed earlier on, just making a start on it since it's a new touching point between a plugin and Cura).
* Adds a backup manager to manage backups (doh!).
* Adds a backup representation that will contain some business logic for actually reading/writing the backup data.
* Implements creating and restoring backups (including the preferences file on Linux).
* Added a way to set the packages API root URL using `CuraVersion.py`.


ChrisTerBeke:
@Appesteijn FYI

Appesteijn:
CURA-5360

ChrisTerBeke:
Note: reason for using docstring here instead of the other inline commenting syntax we use in the rest of Cura is because that automatically works with any Python documentation tool (https://wiki.python.org/moin/DocumentationTools). That is handy if we want to generate API docs for plugin authors.

cura/Api/Backups.py
```diff
@@ -0,0 +1,32 @@
...
+# Cura is released under the terms of the LGPLv3 or higher.
+from cura.Backups.BackupsManager import BackupsManager
+
+
+class Backups:
+    """
+    The backups API provides a version-proof bridge between Cura's BackupManager and plugins that hook into it.
+
+    Usage:
+        from cura.Api import CuraApi
```
ianpaschal: I would capitalize API in all cases where possible. e.g. from `cura.API import CuraAPI`

ChrisTerBeke: Fixed.

cura/Backups/Backup.py
```diff
@@ -0,0 +1,156 @@
...
+        # When restoring a backup on Linux, we move it back.
+        if Platform.isLinux():
+            preferences_file_name = CuraApplication.getInstance().getApplicationName()
+            preferences_file = Resources.getPath(Resources.Preferences, "{}.cfg".format(preferences_file_name))
+            backup_preferences_file = os.path.join(version_data_dir, "{}.cfg".format(preferences_file_name))
+            Logger.log("d", "Copying preferences file from %s to %s", preferences_file, backup_preferences_file)
+            shutil.copyfile(preferences_file, backup_preferences_file)
+
+        # Ensure all current settings are saved.
+        CuraApplication.getInstance().saveSettings()
```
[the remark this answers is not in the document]
LipuFei: I will check more then. Got some weird errors after some backup + restore...

LipuFei:
Seem to work now. @ChrisTerBeke Fixed that data saving after factory reset problem in cfd1b7b

Ghostkeeper:
Note that the reason why we are using documentation _above_ the function/class is to be able to use Doxygen to generate our code documentation. It would be appropriate to convert the documentation in here to Doxygen's format as well.

ChrisTerBeke:
Do whatever you want with it, it's part of Cura anyways now that it's merged.

Personally I hate this commenting style  because IDEs don't support it well (no auto-completion or missing argument warnings) and Doxygen also supports normal Python doc blocks nowadays, plus everyone else in the Python community is using it.

I'm going with the flow in libCharon to have a unified style though, but I'm wasting so much time manually typing all the comment blocks....

Ghostkeeper:
Doxygen supports normal Python `__doc__` but doesn't support tags in it like `\param`.

PR #3817, opened 2018-05-15
Only trigger settings filter on enter or after 500ms timeout

ChrisTerBeke:
This makes the settings filtering field way more bearable on slower machines. It now only triggers when you hit enter, leave the field or after you stop typing for 500ms.


jackha:
Awesome, thanks @ChrisTerBeke !

PR #3832, opened 2018-05-18
CURA-5357 drag and drop packages

ianpaschal:


LipuFei:
Use a simpler approach instead: https://github.com/Ultimaker/Cura/pull/3845

PR #3846, opened 2018-05-22
CURA-5370 Small refactor for Arranger: 

jackha:
Fixed CURA-5370 bug: arranger didn't find a location while it fits fine. This was caused by brims taken into account double. The arranger used an offset convex hull, but the disallowed areas already had a "brim"* in.
- arranger uses disallowed areas without brim* 
- faster, 
- better results when "Arrange all"
- cleanup, 
- more unit tests, 
- take actual build plate size in Arranger instances
- make x and y consistent (numpy arrays start with y first in general), they were used mixed but it wasn't seen because previously the build size was always a square
- use *brim to determine minimum offset between objects

*brim = brim / skirt / other

### DIFF
--- a/cura/Arranging/Arrange.py
+++ b/cura/Arranging/Arrange.py
@@ -18,17 +18,20 @@ LocationSuggestion = namedtuple("LocationSuggestion", ["x", "y", "penalty_points
 #   good locations for objects that you try to put on a build place.
 #   Different priority schemes can be defined so it alters the behavior while using
 #   the same logic.
+#
+#   Note: Make sure the scale is the same between ShapeArray objects and the Arrange instance.
 class Arrange:
     build_volume = None
 
-    def __init__(self, x, y, offset_x, offset_y, scale= 1.0):
-        self.shape = (y, x)
-        self._priority = numpy.zeros((x, y), dtype=numpy.int32)
-        self._priority_unique_values = []
-        self._occupied = numpy.zeros((x, y), dtype=numpy.int32)
+    def __init__(self, x, y, offset_x, offset_y, scale= 0.5):
         self._scale = scale  # convert input coordinates to arrange coordinates
-        self._offset_x = offset_x
-        self._offset_y = offset_y
+        world_x, world_y = int(x * self._scale), int(y * self._scale)
+        self._shape = (world_y, world_x)
+        self._priority = numpy.zeros((world_y, world_x), dtype=numpy.int32)  # beware: these are indexed (y, x)
+        self._priority_unique_values = []
+        self._occupied = numpy.zeros((world_y, world_x), dtype=numpy.int32)  # beware: these are indexed (y, x)
+        self._offset_x = int(offset_x * self._scale)
+        self._offset_y = int(offset_y * self._scale)
         self._last_priority = 0
         self._is_empty = True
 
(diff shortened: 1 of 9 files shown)


cura/Arranging/Arrange.py
```diff
@@ -61,13 +64,17 @@ def create(cls, scene_root = None, fixed_nodes = None, scale = 0.5, x = 220, y =
 
         # If a build volume was set, add the disallowed areas
         if Arrange.build_volume:
-            disallowed_areas = Arrange.build_volume.getDisallowedAreas()
+            disallowed_areas = Arrange.build_volume.getDisallowedAreasNoBrim()
```
ianpaschal: Is there a reason for not simply altering the `getDisallowedAreas()` function? Seems not so intuitive to add a second method with almost the same name doing almost the same thing.

cura/Arranging/Arrange.py
```diff
@@ -104,15 +111,15 @@ def findNodePlacement(self, node, offset_shape_arr, hull_shape_arr, step = 1):
     def centerFirst(self):
         # Square distance: creates a more round shape
         self._priority = numpy.fromfunction(
-            lambda i, j: (self._offset_x - i) ** 2 + (self._offset_y - j) ** 2, self.shape, dtype=numpy.int32)
+            lambda j, i: (self._offset_x - i) ** 2 + (self._offset_y - j) ** 2, self._shape, dtype=numpy.int32)
```
ianpaschal: Why flip `i` and `j`?

cura/Arranging/ArrangeObjectsAllBuildPlatesJob.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2017 Ultimaker B.V.
```
ianpaschal: Update year

cura/Arranging/Arrange.py
```diff
@@ -18,17 +18,20 @@
 #   good locations for objects that you try to put on a build place.
```
ianpaschal: Need copyright and license at top of file

cura/Arranging/ArrangeObjectsAllBuildPlatesJob.py
```diff
@@ -17,6 +18,7 @@
 from typing import List
 
 
+##  Do an arrangements on a bunch of build plates
```
ianpaschal: Bunch of build plates? Not a very clear comment

cura/Arranging/Arrange.py
ianpaschal: As far as I can tell, this is the only place `getDisallowedAreas()` is used.

jackha: It's used a few times in BuildVolume itself for showing which areas are disallowed. The annoying thing is that the *brim size setting is taken into account. Hmmmm maybe in the (near) future we can get rid of that as well.

cura/Arranging/Arrange.py
jackha: In some places the _priority and _occupied were used as (x, y) and in other places it was (y, x). I found the error when I made the Arranger arrange on a rectangular raster. I use (y, x) everywhere now (in the Arranger) because of a few reasons:
- If you print the table, it will have a good orientation for you, this is the main reason
- This is more theoretical and it has become a habit for some (big) tables. Having "y" as rows makes you calculate the memory location of an item as `y*row_size+x`. If you loop through it from "top left to right, then top to bottom" it may be faster than the other way around. For (x, y) loops over these tables I would place the y loop on the outside for this reason.

PR #3848, opened 2018-05-22
Select models after loading

### LINKED ISSUE #356: Auto-select model after loading
When a model is loaded, it seems logical to me to select it immediately, such that the scaling, rotation, etc. tools can immediately be used.

What are your thoughts on this?

fieldOfView:
This PR selects models after loading them into Cura. This makes Cura behave like other applications, including legacy Cura.

Fixes #356

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1572,6 +1572,8 @@ class CuraApplication(QtApplication):
             self.callLater(self.openProjectFile.emit, file)
             return
 
+        Selection.clear()
+
         f = file.toLocalFile()
         extension = os.path.splitext(f)[1]
         filename = os.path.basename(f)
@@ -1695,6 +1697,8 @@ class CuraApplication(QtApplication):
 
             node.callDecoration("setActiveExtruder", default_extruder_id)
             scene.sceneChanged.emit(node)
+
+            Selection.add(node)
 
         self.fileCompleted.emit(filename)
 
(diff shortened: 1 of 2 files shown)


fieldOfView:
I feel this should just be the default and only option. Let me know if you want a preference instead.

ianpaschal:
Sorry Aldo, as I said we discussed it as a team (including our UX-ers) and decided selecting models after loading was not a default behavior we wanted.

Personally I actually do want it, so if you were inclined to make a preference with the default as "off" we could discuss that.

fieldOfView:
I thought it was decided not to put time towards it.

I'll create a preference.

ianpaschal:
Well yes. We're trying to move away from the "We'll get to it, _someday..._" attitude and instead be decisive so we said, "Are we going to do this?", discussed it a bit, voted "No." and therefore closed the issue. That doesn't change the fact that we're _always_ open to pull requests. Hypothetically there are cases where the PR to fix a closed issue might also be closed because we feel it's just really a bad idea, but that's not the case here.

As always, thank you for your contributions!

fieldOfView:
Done

ianpaschal:
Personally this makes me very happy. Will make a JIRA ticket and review ASAP.

ianpaschal:
Devs, JIRA: CURA-5390

ianpaschal:
Works like a charm. ❤️

PR #3854, opened 2018-05-25
Fix configurationmenu text

fieldOfView:
This PR fixes some minor issues with the ConfigurationMenu ("sync button"):

* Fix text rendering on some OSX systems
* Fix untranslatable text

### DIFF
--- a/resources/qml/Menus/ConfigurationMenu/SyncButton.qml
+++ b/resources/qml/Menus/ConfigurationMenu/SyncButton.qml
@@ -13,7 +13,7 @@ Button
     id: base
     property var outputDevice: null
     property var matched: updateOnSync()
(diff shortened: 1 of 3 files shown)


resources/qml/Menus/ConfigurationMenu/SyncButton.qml
```diff
@@ -13,7 +13,7 @@ Button
     id: base
     property var outputDevice: null
     property var matched: updateOnSync()
-    text: matched == true ? "Yes" : "No"
+    text: matched == true ? catalog.i18nc("@label:extruder label", "Yes") : catalog.i18nc("@label:extruder label", "No")
```
diegopradogesto: We replaced the text for an icon, so this text is never used. But good catch!

PR #3855, opened 2018-05-26
Fix for detecting USB connection to Malyan M200 and Monoprice Mini

jorgegarciadev:
Hello, just installed Ultimate Cura 3.3.1 and wasn't able to connect my Malyan M200 via USB. Debugging the problem I found that the answer from the printer during the `AutoDetectBaudJob` was:

`b'ok N0 P15 B15 T:28.8 /0.0 B:24.8 /0.0 T0:28.8 /0.0 @:0 B@:0\n'`

So it didn't pass the check. Changing the check to `if b"ok " in line and b"T:" in line` fixed it.

I don't have any other printers to test if the fix breaks the check for them, but I think they will be fine

Jorge.

### DIFF
--- a/plugins/USBPrinting/AutoDetectBaudJob.py
+++ b/plugins/USBPrinting/AutoDetectBaudJob.py
@@ -58,7 +58,7 @@ class AutoDetectBaudJob(Job):
 
                 while timeout_time > time():
                     line = serial.readline()
-                    if b"ok T:" in line:
+                    if b"ok " in line and b"T:" in line:
                         successful_responses += 1
                         if successful_responses >= 3:
                             self.setResult(baud_rate)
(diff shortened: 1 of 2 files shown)


fieldOfView:
You may also want to check out https://github.com/Ultimaker/Cura/blob/master/plugins/USBPrinting/USBPrinterOutputDevice.py#L308 so temperatures get shown in the monitor UI,

jorgegarciadev:
Thanks, didn't realise about that. Now the temperatures are shown correctly.

jackha:
Hi @jorgegarciadev, thanks for contributing! Sorry from our side that it takes so long.
Devs: CURA-5504

jorgegarciadev:
No problem, thanks for answering.

ElizabethGreene:
Thanks for committing this.  I found the same issue today and was seconds from submitting a similar/inferior patch. :)

PR #3863, opened 2018-05-28
Make headers non-transparent

fieldOfView:
This PR fixes a cosmetic issue in the Toolbox:
image
Notice how the items in the table shine through the header, which is not a great look.


diegopradogesto:
Small fix, it makes sense. Thanks!
It looks OK.
image

PR #3878, opened 2018-06-03
Feature update old plugins

Appesteijn:
Check if plugins are installed through the 'old' plugin browser and are not yet known in the package manager. If so, show them as installed and updatable.
Once a user updates them through the new Toolbox they are correctly known in the package manager.

### DIFF
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -169,6 +169,27 @@ class Toolbox(QObject, Extension):
             "materials_showcase": QUrl("{base_url}/showcase".format(base_url=self._api_url))
         }
 
+    OLD_PLUGINS = ['3DPrinterOS',
+                   'BarbarianPlugin',
+                   'CuraSolidWorksPlugin',
+                   'MakePrintablePlugin',
+                   'OctoPrintPlugin',
+                   'OrientationPlugin',
+                   'ZOffsetPlugin',
+                   'cura-siemensnx-plugin']
+
+    # check for plugins that were installed with the old plugin-browser
+    def _isOldPlugin(self, plugin_id) -> bool:
+        if plugin_id in self.OLD_PLUGINS and plugin_id in self._plugin_registry.getInstalledPlugins():
+            Logger.log('i', 'Found a plugin that was installed with the old plugin browser: %s', plugin_id)
+            if not self._package_manager.isPackageInstalled(plugin_id):
+                Logger.log('i', 'Plugin was not found in package.json: %s', self._package_manager.getInstalledPackageInfo(plugin_id))
+                return True
+        return False
+
+
+
+
     # Get the API root for the packages API depending on Cura version settings.
     def _getCloudAPIRoot(self) -> str:
         if not hasattr(cura, "CuraVersion"):
@@ -325,6 +346,9 @@ class Toolbox(QObject, Extension):
     # --------------------------------------------------------------------------
     @pyqtSlot(str, result = bool)
     def canUpdate(self, package_id: str) -> bool:
+        if self._isOldPlugin(package_id):
+            return True
+
         local_package = self._package_manager.getInstalledPackageInfo(package_id)
         if local_package is None:
             return False
@@ -355,7 +379,7 @@ class Toolbox(QObject, Extension):
 
     @pyqtSlot(str, result = bool)
     def isInstalled(self, package_id: str) -> bool:
-        return self._package_manager.isPackageInstalled(package_id)
+        return self._package_manager.isPackageInstalled(package_id) or self._isOldPlugin(package_id)
 
     @pyqtSlot(str, result = bool)
     def isEnabled(self, package_id: str) -> bool:
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -238,6 +238,23 @@ class Toolbox(QObject, Extension):
         dialog = Application.getInstance().createQmlComponent(path, {"toolbox": self})
         return dialog
 
+
+    def _convertPluginMetadata(self, plugin: dict) -> dict:
+        formatted = {
+            "package_id": plugin["id"],
+            "package_type": "plugin",
+            "display_name": plugin["plugin"]["name"],
+            "package_version": plugin["plugin"]["version"],
+            "sdk_version": plugin["plugin"]["api"],
+            "author": {
+                "author_id": plugin["plugin"]["author"],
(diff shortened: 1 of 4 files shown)


Appesteijn:
Not sure why CURA-5357.... is in this PR

plugins/Toolbox/src/Toolbox.py
```diff
@@ -169,6 +169,27 @@ def _onAppInitialized(self) -> None:
...
+                   'BarbarianPlugin',
+                   'CuraSolidWorksPlugin',
+                   'MakePrintablePlugin',
+                   'OctoPrintPlugin',
+                   'OrientationPlugin',
+                   'ZOffsetPlugin',
+                   'cura-siemensnx-plugin']
+
+    # check for plugins that were installed with the old plugin-browser
+    def _isOldPlugin(self, plugin_id) -> bool:
```
ChrisTerBeke: Missing typing for `plugin_id` (`plugin_id: str`)

plugins/Toolbox/src/Toolbox.py
```diff
@@ -169,6 +169,27 @@ def _onAppInitialized(self) -> None:
...
+        if plugin_id in self.OLD_PLUGINS and plugin_id in self._plugin_registry.getInstalledPlugins():
```
ChrisTerBeke: Why check in 2 places? Shouldn't only checking the plugin registry be enough?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -325,6 +346,9 @@ def getRemotePackage(self, package_id: str) -> Optional[Dict]:
     # --------------------------------------------------------------------------
     @pyqtSlot(str, result = bool)
     def canUpdate(self, package_id: str) -> bool:
+        if self._isOldPlugin(package_id):
```
ChrisTerBeke: I assume from this that regardless of version, we want people to update plugins from the old plugin browser to the new packages?

Appesteijn:
Tried to cherry-pick into a new branch based of 3.4 but gives the same two commits

plugins/Toolbox/src/Toolbox.py
Appesteijn: If it is in the plugin_registry I assume it has been loaded correctly.

plugins/Toolbox/src/Toolbox.py
Appesteijn: yes indeed

plugins/Toolbox/src/Toolbox.py
Appesteijn: fixed

plugins/Toolbox/src/Toolbox.py
ChrisTerBeke: But why do we need `OLD_PLUGINS`? If it's in the plugin registry that should be enough right?

ChrisTerBeke:
I mean making a new branch from master, then `git cherry-pick 30a05e6ba6e7af6b498276bcc5b62229c4dd3821` and `git cherry-pick 0e11a165a3a39ec74f1884041e6f3a5734ed817b`. That should give a branch with only your commits on it. I think that extra commit is here because it's on the 3.4 branch but not in master yet (which also isn't a problem I think, so it can be merged).

diegopradogesto:
Shouldn't the base branch be 3.4?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -169,6 +169,27 @@ def _onAppInitialized(self) -> None:
...
         }
 
+    OLD_PLUGINS = ['3DPrinterOS',
+                   'BarbarianPlugin',
+                   'CuraSolidWorksPlugin',
+                   'MakePrintablePlugin',
+                   'OctoPrintPlugin',
+                   'OrientationPlugin',
+                   'ZOffsetPlugin',
+                   'cura-siemensnx-plugin']
```
ianpaschal: Forcing the user to upgrade is not a terrible solution but still not as clean as just using the sdk_version tag as it's meant to be used. Now the user will still have to upgrade the existing plugin despite them having the same code.

So unfortunately in my mind this doesn't do much other than make the code messy.

We COULD scan the plugins dir though and check if any plugin folders (which are IDs) match something on the server and then offer an upgrade. That would be a bit cleaner than hard-coding these 8 IDs into the toolbox.

Appesteijn:
Yes indeed

plugins/Toolbox/src/Toolbox.py
```diff
@@ -355,7 +379,7 @@ def canDowngrade(self, package_id: str) -> bool:
 
     @pyqtSlot(str, result = bool)
     def isInstalled(self, package_id: str) -> bool:
-        return self._package_manager.isPackageInstalled(package_id)
+        return self._package_manager.isPackageInstalled(package_id) or self._isOldPlugin(package_id)
```
ianpaschal: Even plugins which are not installed will now appear as such with an update button. Not ideal and kind of misleading.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -169,6 +169,27 @@ def _onAppInitialized(self) -> None:
...
+                   'CuraSolidWorksPlugin',
+                   'MakePrintablePlugin',
+                   'OctoPrintPlugin',
+                   'OrientationPlugin',
+                   'ZOffsetPlugin',
+                   'cura-siemensnx-plugin']
+
+    # check for plugins that were installed with the old plugin-browser
+    def _isOldPlugin(self, plugin_id: str) -> bool:
+        if plugin_id in self.OLD_PLUGINS and plugin_id in self._plugin_registry.getInstalledPlugins():
```
ianpaschal: I'm going to change this so that like the other upgrading code, the "update available" designation is made based on the call from the server.

The new logic will look like: "for each plugin from the server, if installed in plugin registry and not installed in package manager, _isOldPlugin = true"

ianpaschal:
I made some changes. Need someone else to review.

Gist of it is:

- As part of updating the installed package models, also update the list of `_old_plugin_ids`.
- This is done by checking everything in the plugin registry for a corresponding plugin in the package manager. If it's not found, it's a plugin sans package.
- The canUpgrade method exposed to QT now returns true if that package ID is in the `_old_plugin_ids` list.
- Installing a plugin triggers `_updateInstalledModels()` which means the `_old_plugin_ids` list will be updated as well.

Appesteijn:
It doesn't work when I update plugins in 3.3.1, update to 3.4 and go to the Toolbox. Plz test next time

PR #3891, opened 2018-06-05
Add settings for prime tower infill to fdmprinter.def.json

### LINKED ISSUE #5330: Cura 3.6 filling error of the second extruder
Hello, looks like I have found an error that most likely goes from version to version of the program.
If you turn off the first extruder in the slicer and leave only the second, then the infilling of the model works fine and changes depending on the percentage. But if the program is closed and opened the next time, the infilling section in "custom" stops responding. But at the same time the infilling section in "Recomended" still works.
The same happens to the infilling pattern.

**Application Version**
version 3.6 , 4.0

**Platform**
windows 8.1

**Printer**
delta printer 

**Steps to Reproduce**

turn on the second extruder,
turn off the first extruder
close the program
open the program
load model
change infilling in the "custom" section
try slicing with new filling 

**Actual Results**
Infilling remains on default

**Expected results**
Infilling must change 

**Additional Information**
if you do it with the first extruder, everything is fine. The problem occurs if you turn off the first extruder and use only the second one

sjoerdtimmer:
Closes #5330

Add the settings required to print a less dense prime tower.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5051,6 +5051,46 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
                 },
+                "prime_tower_sparse_on_nochange":
+                {
+                    "label": "Use sparse infill where possible",
+                    "description": "If enabled, the prime tower will be printed using a grid infill when the extruder does not change between layers. Mostly usefull for n-in-1-out printers where a relatively large prime tower is needed and the unused extruders don't ooze.",
+                    "type": "bool",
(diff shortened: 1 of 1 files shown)


jackha:
Devs: CURA-5500
https://github.com/Ultimaker/CuraEngine/pull/789

ianpaschal:
@sjoerdtimmer Spoke again with colleagues today and we've come to the decision that we're not going to merge this. Just recently, we wrote a whole new implementation for prime towers, which is not really compatible with this code.

Sorry about that.

PR #3893, opened 2018-06-05
Enhance Dagoma printers profiles

dagoma3d:
Hi,

Here is a pull request to enhance Dagoma printers profiles.
What has been done:
- Refine existing profiles.
- Add our new Neva Magis printer.
- Add printer-specific quality profiles.

Regards,
Orel from Dagoma

### DIFF
--- a/resources/definitions/dagoma_neva_magis.def.json
+++ b/resources/definitions/dagoma_neva_magis.def.json
@@ -1,6 +1,6 @@
 {
-    "id": "dagoma_discoeasy200",
-    "name": "Dagoma DiscoEasy200",
+    "id": "dagoma_neva_magis",
+    "name": "Dagoma NEVA",
     "version": 2,
     "inherits": "fdmprinter",
     "metadata": {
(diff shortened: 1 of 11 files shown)


Appesteijn:
I see two the same printers when I run with this PR:
screen shot 2018-06-21 at 16 24 35

jackha:
Devs, see: CURA-5499

resources/definitions/dagoma_neva_magis.def.json
```diff
@@ -0,0 +1,81 @@
+{
+    "id": "dagoma_neva_magis",
+    "name": "Dagoma NEVA",
```
diegopradogesto: This printer has the same name as the other, that's why it is appearing twice in the list.

Appesteijn:
@dagoma3d We would like to add your profile updates, could you add the extruder definitions?

resources/definitions/dagoma_discoeasy200.def.json
```diff
@@ -1,4 +1,5 @@
 {
+    "id": "dagoma_discoeasy200",
```
Ghostkeeper: Remove this ID. It's not used.

Vandrasc:
Hi @dagoma3d, the PR is currently still in our sprint, which means we can add the changes as soon as all conflicts have been resolved. 
Thank you!

PR #3901, opened 2018-06-06
Limit flicking the Toolbox scrollviews to vertical only

fieldOfView:
This PR fixes the scrolling panes in the Toolbox window so they can only be "flicked" vertically, like other scrollviews in Cura. "Flicking" is where, instead of dragging the scrollbar or using mousewheel/multitouch scrolling, you just click and drag around the contents of the scrolling pane. Before this PR, the toolbox scrolling panes could be dragged horizontally as well as vertically, which looks weird.


jackha:
Devs: CURA-5501

Appesteijn:
👍

PR #3911, opened 2018-06-08
Add Creality CR-10S machine definition

mkroman:
The CR-10S is functionally equivalent to the CR-10, but since the definitions support inheritance I figured I'd add it for completeness sake.

It's not entirely the same - The CR-10S has a different board and has dual Z-rods, but that shouldn't matter for this definition.


Appesteijn:
CURA-5493, tested == ok

PR #3920, opened 2018-06-12
CURA-5450 remove concentric 3d

LipuFei:


plugins/VersionUpgrade/VersionUpgrade34to40/VersionUpgrade34to40.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+import configparser
+import io
+
+from UM.VersionUpgrade import VersionUpgrade
+
+
+##  Upgrades configurations from the state they were in at version 3.3 to the
```
jackha: Update comments

jackha: 5b688fc

PR #3922, opened 2018-06-12
definition for Tevo Tornado printer

nean-and-i:
This new definition is a variation of CR-10 definition but with Tevo Tornado specifics as well as a long tested and proofed start and end gcode.


ChrisTerBeke:
The filename is missing the .def.json extension.

nean-and-i:
Can someone please provide more detailed information about that specific errors?
It looks like this jenkins test is testing quite a different codebase and non-related to what I try to push here!
This is just a printer definition json, without any magic or new/divergent functionality.
It works in former version 3.3., it succeeded 3 different json validation tests.
I'd like to ask what exactly is wrong so I can fix it.

ChrisTerBeke:
The issue is in our CI system, not your profile. Please rebase this PR to master as we're not accepting any PRs for 3.4 anymore this close to a stable release.

PR #3923, opened 2018-06-13
Update MachinesPage.qml

ShenRen:
The size of the form is initialization, so that the text is not stacked together.
给定窗体大小初始化值，避免文字堆叠在一起.

### DIFF
--- a/resources/qml/Preferences/MachinesPage.qml
+++ b/resources/qml/Preferences/MachinesPage.qml
@@ -109,6 +109,10 @@ UM.ManagementPage
         UM.Dialog
         {
             id: actionDialog
+            //NOTE:(RenShen): 给定窗体初大小始化值，避免文字堆叠在一起
+            width:800;
+            height:600;
+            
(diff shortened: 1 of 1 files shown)


resources/qml/Preferences/MachinesPage.qml
```diff
@@ -109,6 +109,10 @@ UM.ManagementPage
         UM.Dialog
         {
             id: actionDialog
+            //NOTE:(RenShen): 给定窗体初大小始化值，避免文字堆叠在一起
```
ChrisTerBeke: Please remove this comment, attribution is already visible in GitHub (and Git in general). We also want to keep all comments in English.

fieldOfView:
Note that this will break the dialog on Windows HiDPI screens.

Perhaps you can share a screenshot that shows the problem, and a better fix can be made.

ShenRen:
Ok！ I will remove this comment.

PR #3934, opened 2018-06-14
Create tevo_tornado.def.json

nean-and-i:
This new definition is a variation of CR-10 definition but with Tevo Tornado specifics as well as a long tested and proofed start and end gcode.


Appesteijn:
CURA-5494, tested == ok

PR #3964, opened 2018-06-23
Updated pt_BR translations for 3.4

Patola:


Vandrasc:
Devs see ticket CURA-5497.

Patola:
pytest-main and code-style failed? How is that possible? I just re-checked all strings to see if I could find missing or extra quotes or something like that that could trigger an integration error, but everything seems to be ok. What should I do to get this PR accepted?

Ghostkeeper:
The failing unit tests are not your mistake. Sorry for the confusion. It's something that we have to resolve ourselves.

PR #3982, opened 2018-06-26
Fix a bright line/gap between the viewport and sidebar due to rounding

fieldOfView:
This PR fixes a bright line/gap between the viewport and the sidebar that can appear due to rounding issues or when resizing the window when using a dark theme.


Appesteijn:
Could not really test this, but it looks like no regression on MacOS

PR #3986, opened 2018-06-27
Clarify Profile Override preference options

fieldOfView:
This PR changes the wording of the Profile Override options on the General preferences page. It also makes the "Always ask me this" wording consistent.

image


Appesteijn:
CURA-5531

Appesteijn:
👍

Issue #3987, opened 2018-06-27
firmware update not working i was trying to upgrade my ender 3 frimware it was lagging out


chrisbeaudoin:
**Application Version**

**Platform**

**Printer**

**Steps to Reproduce**

**Actual Results**

**Expected results**

**Additional Information**

alekseisasin:
@chrisbeaudoin , First of all could you please fill in the issue form correctly, How the "ender 3 firmware" is related to Cura?

chrisbeaudoin:
Sorry for not filling the form correctly ok im trying to update my ender 3 firmware with the cura software like the company was telling me to do when I go to cura I go to manage printer update firmware I choose the .hex file but the loading bar doesn’t move the printer company said to contact you guys thasnks for your time


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

alekseisasin:
@chrisbeaudoin, It is still hard to reproduce the issue without having your printer. What you can do, try to update firmware in different Cura version, like latest 3.4, or 3.3 or 3.2 etc. We want to be sure that it worked before.
I think the vendor of your printer also should have another way to update printer firmware.

fieldOfView:
Duplicate of #3731

PR #3999, opened 2018-06-29
Fix firmware upload on Windows

### LINKED ISSUE #3731: [3.3.1] Can't upload firmware
**Application Version**
3.3.1

**Platform**
Windows 10

**Display Driver**

**Printer**
Custom Prusa MK2S DIY

**Steps to Reproduce**
Select printer -> manage printer -> upload firmware -> "select and confirm"

**Actual Results**
Uploading window appears but no progress

**Expected results**
Upload firmware

**Additional Information**
On 3.2 version it was working

fieldOfView:
This PR fixes firmware uploading on Windows by correctly creating a local path from a url-encoded path. 

On Windows, local file URLs start with file:/// (three forwards slashes). Instead of hardcoding a fix for Windows only, this PR uses QUrl to handle this properly.

Fixes #3731 and #3987

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -15,7 +15,7 @@ from cura.PrinterOutput.GenericOutputController import GenericOutputController
 from .AutoDetectBaudJob import AutoDetectBaudJob
 from .avr_isp import stk500v2, intelHex
 
-from PyQt5.QtCore import pyqtSlot, pyqtSignal, pyqtProperty
+from PyQt5.QtCore import pyqtSlot, pyqtSignal, pyqtProperty, QUrl
 
 from serial import Serial, SerialException, SerialTimeoutException
 from threading import Thread, Event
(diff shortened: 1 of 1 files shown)


fieldOfView:
Further testing showed that the default firmware (if any, for the selected printer) uploaded ok, but a custom firmware did not upload. That's because the default firmware location is supplied as a local file, whereas the custom firmware location is supplied QUrl encoded. Both are now handled correctly.

fieldOfView:
subtle bump
This has no Jira issue yet.

ChrisTerBeke:
Devs: CURA-5588 :)

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -127,8 +127,11 @@ def showFirmwareInterface(self):
 
     @pyqtSlot(str)
     def updateFirmware(self, file):
-        # the file path is qurl encoded.
-        self._firmware_location = file.replace("file://", "")
+        # the file path could be qurl encoded.
+        if file[:7] == "file://":
```
LipuFei: Minor: I think `file.startswith("file://")` is better.

fieldOfView: Fixed: 846c55a99d0486bea3623e47412c36d0ab71f0fe

PR #4014, opened 2018-07-03
CURA-5491 pause at height

jackha:
There was a bit more involved than I thought, so now making a PR and a branch for it.
- For UM3 / Griffin it only does M10 ; pause
- For UM2 / UltiGCode it skips the temperature settings and replaces the retracts with M10 and M11 if the retract_amount > 0

Tested both and it seems to work fine.


ChrisTerBeke:
> making a PR and a branch for it :tada:

PR #4022, opened 2018-07-04
Remove ID from Raise3D definition

Ghostkeeper:
These IDs are not used and should never be added again.


Appesteijn:
Tested == OK

diegopradogesto:
Why was this removed? This extruder is being used by the `raise3D_N2_plus_single`. If those must be removed, then should we remove them in all the extruders created in https://github.com/Ultimaker/Cura/commit/aba3795d0b51b3ca8fa5307bd380a0a8fade6aff?

diegopradogesto:
I know, but then for consistency we should remove from all the extruder and machine definition files.

fieldOfView:
Indeed you should.

PR #4038, opened 2018-07-06
CURA-5545 Fix 3MF Reader

LipuFei:


LipuFei:
This is in QA but not merged...

PR #4067, opened 2018-07-12
CURA-5562 Avoid reseting slice status when sSwitch to layer view 

diegopradogesto:


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -490,7 +491,9 @@ def _onSceneChanged(self, source: SceneNode) -> None:
                 if mesh_data and mesh_data.getVertices() is None:
                     return
 
-            build_plate_changed.add(source_build_plate_number)
+            # There are some SceneNodes that do not have any build plate associated, then do not add to the list.
+            if source_build_plate_number is not None:
+                build_plate_changed.add(source_build_plate_number)
```
jackha: Good find! So the `build_plate_changed.discard(None)` had a use anyway ^^ (I forgot what the use case was)

diegopradogesto: YES, XDDDDDDD

PR #4070, opened 2018-07-13
CURA-5384 confirm upon exit

LipuFei:


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -88,6 +88,25 @@ def __init__(self, serial_port: str, baud_rate: Optional[int] = None) -> None:
...
+    # to exit. If so, it will show a confirmation before
+    def _checkActivePrintingUponAppExit(self) -> None:
+        application = CuraApplication.getInstance()
+        if not self._is_printing:
+            # This USB printer is not printing, so we have nothing to do. Call the next callback if exists.
+            application.triggerNextExitCheck()
+            return
+
+        application.setConfirmExitDialogCallback(self._onConfirmExitDialogResult)
+        application.showConfirmExitDialog.emit("USB printing is in progress")
```
alekseisasin: In ticked was another message:
"A USB print is in progress, closing Cura will stop this print. Are you sure?"

cura/TaskManagement/OnExitCallbackManager.py
```diff
@@ -0,0 +1,69 @@
...
+            self._is_all_checks_passed = True
+
+            # Tell the application to exit
+            self._application.callLater(self._application.closeApplication)
+
+    # This is the callback function which an on-exit callback should call when it finishes, it should provide the
+    # "should_proceed" flag indicating whether this check has "passed", or in other words, whether quiting the
+    # application should be blocked. If the last on-exit callback doesn't block the quiting, it will call the next
+    # registered on-exit callback if available.
+    def onCurrentCallbackFinished(self, should_proceed: bool = True) -> None:
```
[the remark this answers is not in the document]
LipuFei: The is the part that I didn't like: this should be called whenever an on-exit callback finishes. Because an on-exist callback can do a lot of things which can involve waiting, we cannot just call the callback and get a result directly. An example is the USB printing check: It needs to show a confirmation dialog and wait for the user's choice, and then it knows whether we should exit or not.

plugins/USBPrinting/USBPrinterOutputDevice.py
LipuFei: Actually, this string should be translated. I will fix this.

LipuFei: Fixed in 49bd249819924f0cf1f0881c646fc01fd102f6f9

LipuFei:
@alekseisasin You pushed your changes to the wrong branch

PR #4084, opened 2018-07-16
CURA-5578 Change the Cartesio variants to use the new ones

diegopradogesto:
Change all the quality profiles and the default variant for Cartesio printers. The VersionUpgrade is also adapted to change the old names to the new ones.

This PR goes together with https://github.com/Ultimaker/fdm_materials/pull/38


diegopradogesto:
FYI @maukcc

PR #4088, opened 2018-07-16
Add generic material packages to materials page

ianpaschal:
Contributes to CURA-5510

### DIFF
--- a/plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
@@ -9,8 +9,8 @@ import UM 1.1 as UM
 
 Item
 {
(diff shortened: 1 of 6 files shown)


plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml
```diff
@@ -104,8 +104,18 @@ Item
             switch(toolbox.viewCategory)
             {
                 case "material":
-                    toolbox.viewPage = "author"
-                    toolbox.filterModelByProp("packages", "author_id", model.id)
+
+                    // If model has a type, it must be a package
+                    if (model.type)
```
diegopradogesto: I don't like this if-statement to differentiate whether the model is a PackageModel or an AuthorsModel. However after discussing with @ianpaschal , this is fine for now. Probably this will change in the future if we have more models in the Toolbox.

PR #4097, opened 2018-07-19
Adjust size of layer number depending on the maximum number of layers

diegopradogesto:
This was already adjusted before, but in Windows I noticed that when the number of layers is less than 10, the width of the TextField that shows the layer number in the LayerView is too small, so this
commit adjusts the size.

Please test in macOS and Linux with number of layers less than 10, less than 100, less than 1000 and less than 10000 at least, since it can have different sizes.


Ghostkeeper:
Developers, see issue CURA-5685.

ianpaschal:
screen shot 2018-08-31 at 13 41 08

4th character doesn't show all the way, and the slider tab doesn't change size.

diegopradogesto:
In Windows looks like this:
image

IMO it has a lot of blank space. But I can live with it if I can see the layer numbers.

PR #4102, opened 2018-07-20
Add winbo_sh105 machine definition

linjinc:
Adding winbo SH105 new machine profiles to Cura

Can I add my machine configuration to cura? Thank you

PR #4106, opened 2018-07-21
I want to add Winbo machine definitions and extruder definitions to Cura

milkok:
I want to add Winbo machines to Cura, including 4 machines.
Can you put these definitions in the offical installer? Thanks a lot!
[definitions.zip](https://github.com/Ultimaker/Cura/files/2215828/definitions.zip)
[extruders.zip](https://github.com/Ultimaker/Cura/files/2215829/extruders.zip)


diegopradogesto:
I would say that this is not the correct way to create a PR for adding new machine definitions. I would be very happy to add these definitions, but doing it that way, some of the members should pick this data and create a proper PR first. If you can do it, then is better and quicker.

I also notice that the PR #4102 also tries to add the Winbo SH105. What definition should prevail?

By the way, this PR intends to merge `master` into `3.4`, that is something we can't ever do. So I will close this PR just in case somebody is tempted to create a mess in our branches.

diegopradogesto:
Hi @milkok if you want us to add these definitions to `master`, please create a proper PR. And, of course, if you need help, don't hesitate to ask.

PR #4119, opened 2018-07-24
Create winbo_mini2.def.json

milkok:


resources/definitions/winbo_mini2.def.json
```diff
@@ -0,0 +1,138 @@
...
+        "machine_gcode_flavor": { "default_value": "Marlin" },
+        "machine_max_feedrate_x": { "default_value": 250 },
+        "machine_max_feedrate_y": { "default_value": 200 },
+        "machine_max_feedrate_z": { "default_value": 40 },
+        "machine_acceleration": { "default_value": 3000 },
+        "gantry_height": { "default_value": 75 },
+        "machine_extruder_count": { "default_value": 1 },
+        "machine_start_gcode": { "default_value": "G21\nG90\nM82\nM107\nG28 X0 Y0\nG28 Z0\nG1 F1000 Z3\nG1 F4000 X0\nG1 F4000 Y0\nG1 F1000 Z0.2\nG92 E0\nG1 F1000 X30 E8\nG92 E0\nM117 Printing." },
+        "machine_end_gcode": { "default_value": "M104 S0\nM140 S0\nG92 E2\nG1 E0 F200\nG28 X0 Y0\nM84 X Y E" },
+        "material_diameter": { "default_value": 1.75 },
```
diegopradogesto: This setting should be defined in the extruder definition.

resources/definitions/winbo_mini2.def.json
```diff
@@ -0,0 +1,137 @@
+{
+    "version": 2,
+    "name": "Winbo FDM-Mini2",
```
diegopradogesto: I forgot this in the previous review. We are not allowing to use **FDM** in the printer names. You can choose for instance **FFF** standing for Fused Filament Fabrication. We recently change our **Custom FDM Printer** to be **Custom FFF Printer**. Of course you can also choose another different name that doesn't contain the acronym.

resources/definitions/winbo_mini2.def.json
```diff
@@ -0,0 +1,137 @@
...
+        "file_formats": "text/x-gcode",
+        "supports_usb_connection": true,
+        "machine_extruder_trains":
+        {
+            "0": "winbo_mini2_extruder"
+        }
+    },
+
+    "overrides": {
+        "machine_name": { "default_value": "Winbo FDM-Mini2" },
```
diegopradogesto: Same here.

resources/definitions/winbo_mini2.def.json
fieldOfView: That's what they name their printers! You can't require them to rename their products, can you?
https://www.winbo.top/3d-printers/

ChrisTerBeke: FDM is trademarked by Stratasys though, that's why we had to change it, courtesy of our legal council....

ChrisTerBeke: The filename is also without fdm, so I guess Winbo Mini2 would do?

diegopradogesto: @fieldOfView I know that I can't require to change their name. Of course I'm not saying that they must change the product name. However now we can't accept those names due to legal reasons. If they can just change how the product shows in Cura, then it's fine. Otherwise, we can't accept this PR.

PR #4117, opened 2018-07-24
Create winbo_superhelper105.def.json

milkok:
PR #4102 is created by my workmate.It will be followed by me.Thanks for reminding.

PR #4121, opened 2018-07-24
Create winbo_superhelper105_extruder.def.json

milkok:


resources/extruders/winbo_superhelper105_extruder.def.json
```diff
@@ -0,0 +1,19 @@
+{
+    "id": "winbo_superhelper105_extruder",
```
diegopradogesto: We don't need this `id`, we use the name of the file as `id`.

resources/extruders/winbo_superhelper105_extruder.def.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+
+    "overrides": {
+        "extruder_nr": {
+            "default_value": 0,
+            "maximum_value": "2"
+        },
+        "machine_nozzle_offset_x": { "default_value": 0.0 },
+        "machine_nozzle_offset_y": { "default_value": 0.0 }
+    }
```
diegopradogesto: As I commented in PR #4117 , you need to specify here the `material_diameter` and `machine_nozzle_size` if they are different than the defaults (2.85mm and 0.4mm, respectively)

PR #4123, opened 2018-07-24
Create winbo_mini2_extruder.def.json

milkok:


resources/extruders/winbo_mini2_extruder.def.json
```diff
@@ -0,0 +1,19 @@
+{
+    "id": "winbo_mini2_extruder",
```
diegopradogesto: We don't use this, the name of the file itself is the `id`.

resources/extruders/winbo_mini2_extruder.def.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+
+    "overrides": {
+        "extruder_nr": {
+            "default_value": 0,
+            "maximum_value": "2"
+        },
+        "machine_nozzle_offset_x": { "default_value": 0.0 },
+        "machine_nozzle_offset_y": { "default_value": 0.0 }
+    }
```
diegopradogesto: You need to specify here the `material_diameter` and `machine_nozzle_size` if they are different than the defaults (2.85mm and 0.4mm, respectively)

PR #4124, opened 2018-07-24
Create winbo_dragonl4_extruder.def.json

milkok:


resources/extruders/winbo_dragonl4_extruder.def.json
```diff
@@ -0,0 +1,19 @@
+{
+    "id": "winbo_dragonl4_extruder",
```
diegopradogesto: Don't need this, the `id` is the name of the file.

resources/extruders/winbo_dragonl4_extruder.def.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+
+    "overrides": {
+        "extruder_nr": {
+            "default_value": 0,
+            "maximum_value": "2"
+        },
+        "machine_nozzle_offset_x": { "default_value": 0.0 },
+        "machine_nozzle_offset_y": { "default_value": 0.0 }
+    }
```
diegopradogesto: Need to define here the `material_diameter` and `machine_nozzle_size`, instead of in the machine definition.

PR #4128, opened 2018-07-25
 Inner Wall Speed changed to be an average speed between Wall Speed and Infill Speed

H4xus:
Previously speed_wall_x (Inner Wall Speed) was calculated by multiplying speed_wall (Wall Speed) by two. This meant that if a user raised their base Wall Speed, it could result in unexpectedly high Inner Wall Speed, especially if the Inner Wall Speed setting is not visible. The description says that inner walls are usually best between Outer Wall Speed and Infill Speed, so I've modified the definition to reflect this. 

This is my first time contributing to Cura, so I apologize if I'm overlooking something. This may need to be changed in multiple places, but I know this affects the Custom FDM Printer profile, which is the only one I use. I'm using this formula now and it seems to work well. Someone with more expertise may have room for improvement though.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2471,7 +2471,7 @@
                                     "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
                                     "maximum_value_warning": "150",
                                     "default_value": 60,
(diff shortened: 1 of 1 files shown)


H4xus:
I will open another pull request to more accurately match the description. I used speed_wall instead of speed_wall_0 for Outer Wall Speed.

PR #4130, opened 2018-07-25
Updating Inner Wall Speed calculation for Custom FDM Printer profile

H4xus:
Value was previously calculated by multiplying Outer Wall Speed by two. This works when the Wall speed is the default of 30 with a print speed of 60, but can cause unexpectedly high print speeds if the user manually raises Wall Speed, especially if the Inner Wall Speed setting is hidden. For example, if Wall Speed is set to 50, Inner Wall Speed would become 100.

I've updated the formula to reflect the definition. This has been working well in my testing.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2471,7 +2471,7 @@
                                     "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
                                     "maximum_value_warning": "150",
                                     "default_value": 60,
-                                    "value": "speed_wall * 2", //Let's think about how to better calculate inner wall speed and limit it to print speed
+                                    "value": "(speed_wall + speed_infil) / 2", 
                                     "limit_to_extruder": "wall_x_extruder_nr",
                                     "settable_per_mesh": true
                                 }
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2471,7 +2471,7 @@
                                     "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
                                     "maximum_value_warning": "150",
                                     "default_value": 60,
-                                    "value": "(speed_wall + speed_infill) / 2", 
+                                    "value": "(speed_wall_0 + speed_infill) / 2", 
                                     "limit_to_extruder": "wall_x_extruder_nr",
                                     "settable_per_mesh": true
                                 }
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2471,7 +2471,7 @@
                                     "maximum_value": "math.sqrt(machine_max_feedrate_x ** 2 + machine_max_feedrate_y ** 2)",
                                     "maximum_value_warning": "150",
                                     "default_value": 60,
-                                    "value": "speed_wall * 2",
+                                    "value": "speed_wall * 2", //Let's think about how to better calculate inner wall speed and limit it to print speed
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I don't agree with this change because it changes the defaults for dozens of printers for quite a significant setting, and because it violates the expectation that these child settings depend on their parent settings. People make the Outer Wall Speed setting visible because they want to tune it separately from the Inner Wall Speed, but when you change it now the Inner Wall Speed also changes. How about this:

* We double the Wall Speed setting.
* We halve all the settings that inherit from Wall Speed (including Outer Wall Speed and Inner Wall Speed).

This should have no effect on the actual g-code, but makes the Wall Speed setting appear higher so that the user is not surprised by the actual speed of the printer. Only the Outer Wall Speed will be slower than what the user indicated.

H4xus:
I can see your point. Using ``speed_wall_0`` would be inheriting a value from another child instead of a parent, and the results are not intuitive. This also increase print times. I simply used that formula because of the description in fdmprinter.def.json. 

It took me a little while to look through what inherits what, but I think your solution may be good. The one thing would clarify is instead of actually doubling Wall Speed through multiplication, we can simply make ``speed_wall`` equal to ``speed_print`` instead of ``speed_print / 2.`` (effectively doubling)

So to summarize, ``speed_wall`` would just become the same value as ``speed_print`` instead of being divided by two ``speed_wall_0`` (Outer Wall Speed) would then be divided by two, and ``speed_wall_x`` (Inner Wall Speed) would inherit its value from ``speed_wall``.

I believe that would effectively limit unexpected behavior that may come from raising wall value, while allowing users to customize child wall speeds without affecting each other. 

Potential issues I'm unsure about:

- Users may get confused about the sudden doubling in ``speed_wall`` value
- Need to confirm that nothing in g-code generation relies on `speed_wall` itself. If you're printing several walls thick, are those all considered Inner or Outer Walls? That's something I don't know off-hand, but can look into if you're not sure. The only wall speed related references I see in fdmprinter,json are related to bridge settings which rely on ``speed_wall_0``. I believe this change could actually prevent possible unexpected behavior there too.

Am I following you correctly Ghostkeeper?

Ghostkeeper:
You seem to be following me :smiley:

> The one thing would clarify is instead of actually doubling Wall Speed through multiplication, we can simply make `speed_wall` equal to `speed_print` instead of `speed_print / 2`. (effectively doubling)

That's what I had in mind :)

> Users may get confused about the sudden doubling in `speed_wall` value

Yeah that's true. I think it would be acceptable in this case. But what's perhaps worse is that there might be profiles that change the `speed_wall` setting and those will have to be halved, lest we inadvertently double the wall printing speed for some printers.

> If you're printing several walls thick, are those all considered Inner or Outer Walls?

There's always one Outer Wall. The rest are Inner Walls. In layer view, the outer walls are coloured red and the inner walls are coloured green if you select the "Line Type" colour scheme.

> Need to confirm that nothing in g-code generation relies on `speed_wall` itself.

We have a rule that CuraEngine itself is only allowed to use leaf settings that don't have any child settings, so the engine itself doesn't use the setting directly. In the front-end you can see in the tooltip of the setting which other settings rely on it, and that is indeed only the Outer Wall Speed and Inner Wall Speed settings so we're all good on this.

Ghostkeeper:
Developers, see issue CURA-5691.

H4xus:
Got it. I should have some time tomorrow or the next day to look through the existing profiles for speed_wall that would need doubling. 

I've updated the profile with what I'm currently testing with.

alekseisasin:
I suppose we also need to update one of the values for:

**ultimaker2_plus.def.json**
Ultimaker 2+
"speed_wall_x": { "value": "speed_wall" }

For the rest, I did not see conflicts.

PR #4132, opened 2018-07-25
CURA-5538 Fix extruder convex hull for one at a time mode

LipuFei:
In one at a time mode, when multiple extruders or not the 1st extruder is used, the convex hull of the head and fan is not correct and can cause collision into a printed model.


cura/Scene/ConvexHullDecorator.py
```diff
@@ -229,15 +230,53 @@ def _compute2DConvexHull(self):
...
+                    point[1] += offset[1]
+
+            apply_offset(offset_polygon, [-offset_x, -offset_y])
+
+            def union(rect1, rect2):
+                bl = [min(rect1[0][0], rect2[0][0]), min(rect1[0][1], rect2[0][1])]
+                tl = [min(rect1[1][0], rect2[1][0]), max(rect1[1][1], rect2[1][1])]
+                tr = [max(rect1[2][0], rect2[2][0]), max(rect1[2][1], rect2[2][1])]
+                br = [max(rect1[3][0], rect2[3][0]), min(rect1[3][1], rect2[3][1])]
+                return [tl, bl, tr, br]
```
alekseisasin: What is it - **tl, bl, tr, br** ?

LipuFei: top-left
bottom-left
top-right
bottom-right

LipuFei: There were some mistakes. Fixed in 42b86086c

PR #4137, opened 2018-07-26
Cura 5595 add custom button to menu

alekseisasin:


ChrisTerBeke:
What does this look like in the application?

alekseisasin:
menu_actions = []
        menu_actions.append("sidebarMenuItemOnClickHander")

        data = {
            "name": "Settings Guid",
            "iconName": "help-contents",
            "actions": menu_actions,
            "menu_item": MenuItemHandler(self)
        }
        CuraApplication.getInstance().addSidebarCustomMenuItem(data)

I think the icon_name should be updated. If the plugin is outside Cura (user private folder), then the icon might be empty,

ianpaschal:
I think this is really nice to have. It should also be part of the Cura API imo. Would be good for authors to have a way to interact with the menus in a reliable way.

Also, just to clarify, this is about the right click context menu right?

ianpaschal:
^That's what I mean. I think this PR should only be merged if it's done that way... the "right" way. So that it's usable in many scenarios and more forward compatible.

ianpaschal:
Aleks, to answer your question in standup:

API in this context just basically means wrapper functions which plugin owners can always use and count on it doing the same thing. Under the hood, we might change the implementation a bit, but as long as the wrapper functions (or... "interface" functions...) stay the same, plugins won't become broken.

Because this is a useful feature for plugin authors, I think it should be implemented via the Cura Plugin API.

ChrisTerBeke:
You can check the already existing API for the backups to see how it could be implemented.

cura/CuraApplication.py
```diff
@@ -1730,3 +1734,11 @@ def _onContextMenuRequested(self, x: float, y: float) -> None:
     @pyqtSlot()
     def showMoreInformationDialogForAnonymousDataCollection(self):
         cast(SliceInfo, self._plugin_registry.getPluginObject("SliceInfoPlugin")).showMoreInfoDialog()
+
+
+    def addSidebarCustomMenuItem(self, menu_item: list) -> None:
```
ianpaschal: Should be a `dict` here right?

cura/Settings/SidebarCustomMenuItemsModel.py
```diff
@@ -0,0 +1,41 @@
...
+
+from UM.Qt.ListModel import ListModel
+from PyQt5.QtCore import pyqtSlot, Qt
+
+
+class SidebarCustomMenuItemsModel(ListModel):
+    NameRole = Qt.UserRole + 1
+    ActionsRole = Qt.UserRole + 2
+    MenuItemRole = Qt.UserRole + 3
+    MenuItemIconNameRole = Qt.UserRole + 5
```
ianpaschal: Should use lowercase and underscores (`menu_item_icon_name_role`)

cura/Settings/SidebarCustomMenuItemsModel.py
```diff
@@ -0,0 +1,41 @@
...
+        self.addRoleName(self.MenuItemIconNameRole, "iconName")
+        self._updateExtensionList()
+
+    def _updateExtensionList(self)-> None:
+        from cura.CuraApplication import CuraApplication
+        for menu_item in CuraApplication.getInstance().getSidebarCustomMenuItems():
+
+            self.appendItem({
+                    "name": menu_item["name"],
+                    "iconName": menu_item["iconName"],
```
ianpaschal: Use proper casing

cura/Settings/SidebarCustomMenuItemsModel.py
```diff
@@ -0,0 +1,41 @@
...
+                    "actions": menu_item["actions"],
+                    "menu_item": menu_item["menu_item"]
+                })
+
+    @pyqtSlot(str, "QVariantList", "QVariantMap")
+    def callMenuItemMethod(self, menu_item_name: str, menu_item_actions: list, kwargs: Any)-> None:
```
ianpaschal: Space before `->`

cura/Settings/SidebarCustomMenuItemsModel.py
```diff
@@ -0,0 +1,41 @@
...
+        self.addRoleName(self.MenuItemRole, "menu_item")
+        self.addRoleName(self.MenuItemIconNameRole, "iconName")
+        self._updateExtensionList()
+
+    def _updateExtensionList(self)-> None:
+        from cura.CuraApplication import CuraApplication
+        for menu_item in CuraApplication.getInstance().getSidebarCustomMenuItems():
+
+            self.appendItem({
+                    "name": menu_item["name"],
```
ianpaschal: Single indent

resources/qml/Settings/SettingView.qml
```diff
@@ -561,6 +561,28 @@ Item
...
                 }
 
+                Instantiator
+                {
+                    id: customMenuItems
+                    model: Cura.SidebarCustomMenuItemsModel { }
+                    MenuItem
+                    {
+                        text: model.name
+                        iconName: model.iconName
```
ianpaschal: See other comment, should be `icon_name`

ianpaschal:
@alekseisasin I made some code style requests to you. I also added an API layer which should be called by the plugin instead of accessing Cura directly. See the "usage" comment in the API file. Pleas try integrating this in your plugin. Also, just a side note, make sure you check your plugin for typos and such. I see you make the menu item "Settings Guid" instead of "Settings Guide".

@ChrisTerBeke What do you think about this API file? I think it's pretty straight forward but maybe you have some tips.

ianpaschal:
Ok @alekseisasin, the final syntax to use is:

`api.interface.settings.addContextMenuItem(data)`

Issue #3875, opened 2018-06-02
Cura should never start a perimeter on an overhang


VanessaE:
I'm printing the vertical version of this part:

https://www.thingiverse.com/thing:2563185

As you can see, it consists of one horizontal beam, two vertical beams, and 45° diagonal beams forming an X.  This X, thus, has overhangs that face each other, when sliced.

No matter how I tell Cura to position the seam, one of the diagonal bits will have the seam placed right on one of its outermost corners (or too close to it).  For example, here is "user specified" mode with the target coordinates set to {100,600}, which inexplicably puts the right island's seam on the overhang:

![image](https://user-images.githubusercontent.com/1733537/40870985-6c00adb2-6601-11e8-83d0-27a11bc01fcc.png)

Shortest has the same effect, the right island's seam starts on the overhang:

![image](https://user-images.githubusercontent.com/1733537/40870988-81497762-6601-11e8-92df-5c2e7929dd47.png)

Sharpest Corner, the right island is fine now, but now the left island has its seam starting on the overhang:

![image](https://user-images.githubusercontent.com/1733537/40870997-a8945cba-6601-11e8-85a0-4cfb98316952.png)

I guess I don't need to show Random mode.

The print moves proceed counter-clockwise from the starting point (at least in Sharpest Corner mode).

The design of this part makes it impossible to rotate it to any orientation that would put all of the seams in desirable positions, because of the facing overhangs created by the X shape.  Anyways, you're not allowed to rotate this part to arbitrary angles - it's a skew calibration object, so it must be printed parallel to either X or to Y (depending on which of X/Z or Y/Z is being calibrated).

Basically, Cura should never -- in any seam placement mode -- start a perimeter on an overhanging corner, whether it's the exterior perimeter, or the corresponding interior ones.  I realize an interior perimeter is usually 100% supported by the lines under it, but it's still on an overhanging corner.  Although there is some plastic there, there's nowhere near enough thickness to ensure a rigid base for each new layer.

In the actual print, as expected, this causes the affected corner to curl, warping the whole diagonal bit after a while.  Warp it enough and the nozzle crashes into it, forcing it down.  Too much further, and something will crash into it that will knock the part off the plate (like fan ducts).  The bit with the seam placed properly is perfect (or as much so as my bot can do right now).  There's plenty of cooling, and I'm printing slow with this part (30 mm/s).

[...]

PR #4142, opened 2018-07-26
Improve processing of overhanging walls

smartavionics:
The rational behind this PR is to separate the handling of overhung walls from the bridge handling. This allows overhung walls to be detected (and potentially printed with different speed) without having the bridge code enabled.

Detecting overhung walls is beneficial because it allows the wall start vertex to be moved if it is in an overhang region. Without doing that, poor print quality may result.

See https://github.com/Ultimaker/Cura/issues/3875 for a (bad tempered) discussion of this problem!

This PR make the following changes:

1 - changes the bridge_wall_min_length, bridge_wall_coast, wall_min_flow, wall_min_flow_retract settings to be retrieved from the current mesh (now available) rather than the current extruder which was always just a hack to get those values.

2 - removes the bridge_wall_max_overhang setting as it is no longer required.

3 - adds the wall_overhang_angle and wall_overhang_speed_factor settings to provide improved overhung wall processing independent of the bridge code.

Engine PR is https://github.com/Ultimaker/CuraEngine/pull/814


Ghostkeeper:
Developers, see issue CURA-5633.

PR #4144, opened 2018-07-27
Added the "Within Infill" combing option that functions identically to the original "Not in Skin" option.

smartavionics:
See https://github.com/Ultimaker/CuraEngine/pull/781

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -3344,13 +3344,14 @@
                 "retraction_combing":
                 {
                     "label": "Combing Mode",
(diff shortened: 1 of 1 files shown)


smartavionics:
Thanks for merging this.

I hope you realise that the change to the no skin combing behaviour means that people who are currently using no skin combing will see different results (hopefully better!). If they want the old behaviour then they have to use this new within infill setting. So I am wondering if you want to add something to your version upgrade script to automatically change the value of the retraction_combing setting in existing profiles from "noskin" to "infill"  so that the user doesn't see a change in behaviour when they upgrade?

alekseisasin:
I have noted in the ticket that the old implementation is "infill": "Within Infill".
by the way, it is a good notice, indeed we probably need to apply the change in project loading. Tomorrow I will discuss it with other colleagues.

alekseisasin:
@smartavionics , we made the change to the version upgrade, now after loading an old project the retraction_combing = "noskin" will become "infill"

smartavionics:
Good. That should avoid people getting a shock when their prints come out different to expected.

PR #4149, opened 2018-07-27
- Fixed extruder behaviour to force filament diameter to 1.75

kaleidoscopeit:
I realized that, after a config reset, the automatic upgrade of the machine definition, in order to support multiple extruder, will not keep the filament diameter to 1.75 and fall to the cura default (2.85). I've made a profile upgrade in order to keep that setting.

Futhermore, I've modified the z max speed.


LipuFei:
Hi @kaleidoscopeit , could you provide reproduce steps for this problem? I see that Deltacomb 3D only has 1 extruder. What do you mean by supporting multiple extruders? And We did that to specifically fix the nozzle size / material diameter issue. Thanks.

kaleidoscopeit:
Hi, I've tested this configuration by decompressing CURA 4.3.1 AppImage on Ubuntu 16.04, then replaced the configuration files and reset the local settings by deleting ~/.config/cura. I wrote "By supporting multiple extruder" because I guessed that the authomated gneration of the separated extruder definition was on that puprose. Thank you in advance. :)

resources/extruders/deltacomb_extruder_0.def.json
```diff
@@ -10,7 +10,6 @@
 
     "overrides": {
         "extruder_nr": { "default_value": 0 },
-        "machine_nozzle_size": { "default_value": 0.4 },
-        "material_diameter": { "default_value": 1.75 }
```
LipuFei: `material_diameter` should also be there in the extruder definitions.

resources/definitions/deltacomb.def.json
```diff
@@ -55,6 +56,7 @@
         "support_use_towers" :                   { "default_value": false },
         "jerk_wall_0" :                          { "value": "30" },
         "jerk_travel" :                          { "default_value": 20 },
-        "acceleration_travel" :                  { "value": 10000 }
+        "acceleration_travel" :                  { "value": 10000 },
+        "machine_max_feedrate_z" :               { "default_value": 150 }
```
LipuFei: Is this setting change also needed? At least this is not part of the diameter fix.

LipuFei:
I'll fix the missing info in the extruder definition.

LipuFei:
Fixed in e48c24e1d

kaleidoscopeit:
Thank you so much Lipu for your fix, you have been very kind!

resources/definitions/deltacomb.def.json
kaleidoscopeit: Yes It's needed. I forgot it in a previous commit. will force the Z speed to 150 because in delta machines can handle it better

PR #4159, opened 2018-07-30
Fix the scrolling when the cursor is hovering the combobox settings

diegopradogesto:
Previously, when I hover a combobox setting (e.g. Build Plate Adhesion Type) the scrolling doesn't work.

### DIFF
--- a/resources/qml/Settings/SettingComboBox.qml
+++ b/resources/qml/Settings/SettingComboBox.qml
@@ -20,13 +20,6 @@ SettingItem
 
         anchors.fill: parent
 
-        MouseArea
-        {
-            anchors.fill: parent
-            acceptedButtons: Qt.NoButton
-            onWheel: wheel.accepted = true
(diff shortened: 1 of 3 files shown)


fieldOfView:
Now you can no longer use the scrollwheel (or scroll multitouch-gestures) to change combobox-settings. 

In most (if not all) applications, the mousewheel acts on the control the mouse pointer hovers over. This includes eg nested scrollable divs. 

I would argue this change is counter intuitive.

diegopradogesto:
The scrolling in the ComboBoxes was not working at all. I was trying different things and I couldn't make it work. I prefer then to be able to use the scroll to scroll the list of the settings, instead of get stuck when I hover the setting.
Open to solutions.

fieldOfView:
The previous scrolling in comboboxes must have been broken since the qtcontrols 2 upgrade then.

PR #4164, opened 2018-07-31
Add the active build plate name to default (not user specified) job name

jackha:


diegopradogesto:
Code looks OK. However I see that in case the user rename the project, the different buildplates don't use different names. So the problem is still there.

diegopradogesto:
@jackha can you take a look to my commit and test if you like the behavior?

cura/PrintInformation.py
```diff
@@ -311,18 +311,24 @@ def _updateJobName(self):
...
             else:
-                job_name = base_name
+                self._job_name = base_name
+
+        # In case there are several buildplates, a suffix is attached
+        if self._multi_build_plate_model.maxBuildPlate > 0:
+            connector = "_#"
+            suffix = connector + str(self._active_build_plate + 1)
+            if connector in self._job_name:
+                self._job_name = self._job_name.split(connector)[0] # get the real name
```
jackha: Isn't this making it "too fancy"?

cura/PrintInformation.py
```diff
@@ -311,18 +311,24 @@ def _updateJobName(self):
...
+                    self._job_name = self._abbr_machine + "_" + base_name
             else:
-                job_name = base_name
+                self._job_name = base_name
+
+        # In case there are several buildplates, a suffix is attached
+        if self._multi_build_plate_model.maxBuildPlate > 0:
+            connector = "_#"
+            suffix = connector + str(self._active_build_plate + 1)
+            if connector in self._job_name:
```
jackha: I would use self._job_name.endswith(connector)

cura/PrintInformation.py
jackha: On the other hand, if you use _# already...

diegopradogesto: This if statement will only affect if the builtiplexer feature is enabled.

I did use a connector that is not a common character for naming.

cura/PrintInformation.py
diegopradogesto: But the self._job_name can be something like name_#2, or name_#3, so in that case I need to look also for the active buildplate number.

PR #4170, opened 2018-08-01
CURA-5613 Set the focus in the upper handle in the layer view

diegopradogesto:
Previously it was set to the text field. That makes it easier to release the focus by clicking in the main view and use the arrow keys to rotate the view.

### DIFF
--- a/plugins/SimulationView/LayerSlider.qml
+++ b/plugins/SimulationView/LayerSlider.qml
@@ -67,6 +67,10 @@ Item {
         activeHandle = handle
     }
 
+    // make sure the upper handle is focussed when pressing the parent handle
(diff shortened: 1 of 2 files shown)


ianpaschal:
Note, this now contributes to CURA-5613

ianpaschal:
I think this is for the best. It's also inline with typical OS behavior: text boxes only are affected by arrow keys once you give them focus.

fieldOfView:
Does this also solve the issue that if you are in Layer View and you are making changes in the sidebar (with auto slicing enabled), that the focus can "stolen" from the sidebar while typing in values?

ianpaschal:
I didn't think about that, but yes probably.

PR #4176, opened 2018-08-02
Fix the sliders that went out of range when using the arrow keys

diegopradogesto:
This PR tries to fix this problem:

- When using the arrow keys to move up/down the layers or left/right the paths, the slider went out of range, as you can see in this picture:
image

It's more notifiable when there are just a few layers (or paths) and you use the SHIFT combination to move 10 by 10.

### DIFF
--- a/plugins/SimulationView/LayerSlider.qml
+++ b/plugins/SimulationView/LayerSlider.qml
@@ -68,11 +68,7 @@ Item {
     }
 
     function normalizeValue(value) {
-        if (value > sliderRoot.maximumValue)
-            return sliderRoot.maximumValue
-        else if (value < sliderRoot.minimumValue)
-            return sliderRoot.minimumValue
-        return value
+        return Math.min(Math.max(value, sliderRoot.minimumValue), sliderRoot.maximumValue)
     }
 
     // slider track
--- a/plugins/SimulationView/LayerSlider.qml
+++ b/plugins/SimulationView/LayerSlider.qml
@@ -40,31 +40,39 @@ Item {
 
     property bool layersVisible: true
 
-    function getUpperValueFromSliderHandle () {
+    function getUpperValueFromSliderHandle() {
         return upperHandle.getValue()
     }
 
-    function setUpperValue (value) {
+    function setUpperValue(value) {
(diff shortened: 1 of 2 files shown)


plugins/SimulationView/LayerSlider.qml
```diff
@@ -40,33 +40,41 @@ Item {
...
+    function setActiveHandle(handle) {
         activeHandle = handle
     }
 
+    function normalizeValue(value) {
+        if (value > sliderRoot.maximumValue)
+            return sliderRoot.maximumValue
+        else if (value < sliderRoot.minimumValue)
+            return sliderRoot.minimumValue
+        return value
```
ianpaschal: I would just use the JavaScript classic: `return Math.min(Math.max(value, upper), lower);`

ianpaschal: I avoid `else if`s whenever possible. I don't think they should be in any language. 😛

plugins/SimulationView/PathSlider.qml
```diff
@@ -47,6 +48,14 @@ Item {
         rangeHandle.width = handle.x - sliderRoot.handleSize
     }
 
+    function normalizeValue(value) {
+        if (value > sliderRoot.maximumValue)
+            return sliderRoot.maximumValue
+        else if (value < sliderRoot.minimumValue)
+            return sliderRoot.minimumValue
+        return value
+    }
```
ianpaschal: Same here.

plugins/SimulationView/LayerSlider.qml
diegopradogesto: Fair enough. But I would rather use `return Math.min(Math.max(value, lower), upper)` 😛

Done in 1a189ba4f5be0d0a377ddcdd8671e0a59b5d7533

plugins/SimulationView/PathSlider.qml
diegopradogesto: Done in 1a189ba4f5be0d0a377ddcdd8671e0a59b5d7533

ianpaschal:
I forgot about it. 😎 🤑 💯

Issue #1696, opened 2017-04-16
rounding errors in Cura


mypeopoly:
Hi,
I was talking with Tim Kuipers on Facebook and he informed me that Cura rounds the calculation to the integer micron in a discussion about using Cura for SLA printing.  He is concerned the accumulated errors may be up to 10 microns. I just want to learn more about the inner working of Cura in this area.  What is the best place to dig in and learn more

Ghostkeeper:
Start with the `utils/LinearAlg2D.cpp` file. The accumulated errors he mentions are the result of CuraEngine working with integer math. The integer math is in there: It has to do some vector math over there, and multiplies distances with small-ish direction vectors. This is the result of most of CuraEngine's inaccuracies, as far as I know.

mypeopoly:
@Ghostkeeper  I found those files in the Curaengine section. How do I go about testing them? Do I complie the Curaengine and replace the one in my Cura 2.4?
I have some programming experience from math course works but not much with C++

Ghostkeeper:
A good programmer would write unit tests to test them. We already have some tests in the `tests/utils/PolygonUtilsTest.cpp` file and such that should serve as an example. Basically, you'd write a function that has assertions in it. You test the assertion using `CPPUNIT_ASSERT_MESSAGE` or other such functions that the CppUnit library provides. Then you register those functions in the accompanying header file. If you then compile CuraEngine (including tests) using CMake, it'll produce additional executables that will yield an `output.xml` file when ran, with the results of the test.

A less experienced programmer would just look for the main function in `src/main.cpp`, which is the entry point of the program. There you can just call the Engine's functions manually, and print the results into `std::cout`. Put a `return 0;` after your calls so that you don't run the rest of the engine every time.

If you make tests in the first way we can consider merging them and we'll run those tests regularly to confirm that nothing breaks. If you test things in the second way we can't, obviously.

mypeopoly:
@Ghostkeeper 
Thanks for such detailed reply. I am very inexperienced when it comes to cpp so it would take some times to dig into to that.  Perhaps let me share with you where I became interested in this topic and like to find out to know if I am attacking the right problem.

As mentioned, we are using Cura to generate toolpaths for laser SLA printer.  One thing we noticed is that in some situations, Cura may skip some of the small features of an object. We were wondering if this is due to rounding errors in which the small features are rounded to zero.

An example is here:
here is an STL for a ring with support built in:
[...]

PR #4177, opened 2018-08-02
Add user editable snap distance

cubiq:
This PR requires an update to CuraEngine as well ( https://github.com/Ultimaker/CuraEngine/pull/818 )

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5625,6 +5625,19 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "snap_distance":
+                {
+                    "label": "Snap Distance",
+                    "description": "Snap distance between polygons in mm. Controls the polygons that may be filtered out when the model is sliced. Lower values lead to higher resolution mesh at the cost of slicing time. It is meant mostly for high resolution SLA printers and very tiny 3D models with a lot of details.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 1.0,
+                    "minimum_value": "0.001",
+                    "minimum_value_warning": "0.05",
+                    "maximum_value_warning": "1.0",
+                    "settable_per_mesh": true,
+                    "settable_per_extruder": false
+                },
                 "meshfix_maximum_resolution":
                 {
                     "label": "Maximum Resolution",


Ghostkeeper:
I think we need to give this a better name and a more neutral description. This adjusts a filter that removes polygons with a circumference less than the "snap distance". So Snap Distance is not a good name because it doesn't snap to anything at all. What about Minimum Circumference? Or if you can find something more elegant...

The description mentions that it's suitable for SLA printers. I'd prefer that to be more neutral so that other types of printers may use this too. The first sentence of the description is also inaccurate. It's not between polygons.

Yeah, I think we should adjust that variable name in the engine as well.

cubiq:
Thanks @Ghostkeeper for evaluating my PR!

So if I get it right, it removes polygons inscribed in a circumference smaller than snap distance (I guess diameter). To simplify how about "Minimum Polygon size" as a name and then add a suitable description?

Ghostkeeper:
"Size" is indeed better. It's a bit vague but perhaps intuitive enough.

I'm still thinking that the setting is perhaps a bit too detailed for a user to be fiddling with. It may be applicable for SLA but Cura is primarily intended for FFF. This piece of code was intended as a sort of pre-filter as well to not work with degenerate shapes that would fall away in the first inset anyway. Perhaps it's more logical to hard-code this to always be equal to the outer wall line width times 4 (which would remove pieces smaller than a square of one outer wall wide).

I'll discuss this with the Cura team tomorrow though, see what they think about it.

ianpaschal:
My opinion is that we're already deep in the deep end when it comes to settings users probably don't need to mess with; after all, that's why we have recommended mode.

Considering what a small and clean tweak this is, and removes a magic hard-coded value from the engine (a `TODO` anyway, apparently) I'm in favor of a merge, @Ghostkeeper.

ianpaschal:
See my comments in https://github.com/Ultimaker/CuraEngine/pull/818

mypeopoly:
Hi, this is Mark from Peopoly.  I was the one who started a post about rounding error in Cura (https://github.com/Ultimaker/Cura/issues/1696) after Tim Kuipers mentioned it in a Facebook group.   In the attempt to preserve as much polygons in Cura for SLA printing, our tests led us to the snapdistance variable.  We felt this could be a useful tweak for other users who need more polygons.  It worked quite well and we created a special version of Cura with this variable modded.

Matteo is a Peopoly Moai user and he was interested about getting as more resolution from Cura while minimizing artifacts.  After some tests, we would able found combinations of snapdistance / resolution that improve the print results while keeping the resolution high.  It worked so well that we felt this would likely to be useful for other users. This led to the pull request.  

If there is a concern about users being confused about what this "snap distance" does or enter a number that is not practical, perhaps we have a drop down box with 2 options, one is for Default Cura, and the other is Higher Polygon (0.1) without giving any option to manually enter a value.

jackha:
Discussed this in the team. We renamed it to Minimum polygon circumference and merged it :-) Thanks @cubiq

PR #4184, opened 2018-08-03
CURA-5570 More decision points in material and quality lookup!

LipuFei:
Hope we can still understand this after a month or so...


cura/Machines/MaterialManager.py
```diff
@@ -186,52 +186,80 @@ def initialize(self):
...
-                    variant_node = machine_node.children_map[variant_name]
-                    if root_material_id in variant_node.material_map:  # We shouldn't have duplicated variant-specific materials for the same machine.
-                        ConfigurationErrorMessage.getInstance().addFaultyContainers(root_material_id)
-                        continue
-                    variant_node.material_map[root_material_id] = MaterialNode(material_metadata)
-                else:
-                    # Add this container id to the wrong containers list in the registry
-                    Logger.log("w", "Not adding {id} to the material manager because the variant does not exist.".format(id = material_metadata["id"]))
-                    self._container_registry.addWrongContainerId(material_metadata["id"])
+        root_material_id = material_metadata["base_file"]
```
ChrisTerBeke: Would Cura crash if any of these keys are not in material_metadata for whatever reason?

fieldOfView:
It would be great if, instead of hardcoding "NOZZLE" and "BUILDPLATE", it would be possible to specify/register other variant types too.

cura/Machines/MaterialManager.py
LipuFei: It will crash, and it will also crash everywhere because `base_file` is a must-have metadata in a material profile. If this is missing, it will be a very serious bug.

LipuFei:
@fieldOfView That's true, but I don't know if future options will still be "variants"... That's what annoys me. Actually, `variant_type` is here only for that sanity check to make sure that the variant defined the material exists. If that check is no longer needed, then `variant_type` can be gone too.

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -693,74 +693,38 @@ def deserialize(self, serialized, file_name = None):
...
-                        for entry in settings:
-                            key = entry.get("key")
-                            if key in self.__unmapped_settings:
-                                if key == "hardware compatible":
-                                    buildplate_compatibility = self._parseCompatibleValue(entry.text)
-                                elif key == "hardware recommended":
-                                    buildplate_recommended = self._parseCompatibleValue(entry.text)
-                            else:
-                                Logger.log("d", "Unsupported material setting %s", key)
+                        _, buildplate_unmapped_settings_dict = self._getSettingsDictForNode(buildplate)
```
alekseisasin: _, - is very strange name for variable

ChrisTerBeke: That means you wanna ignore that variable (but have no control over the fact that it comes back from the called method). It's actually a very common practice in tuples or method args.

cura/Machines/QualityManager.py
```diff
@@ -258,40 +248,46 @@ def getQualityGroups(self, machine: "GlobalStack") -> dict:
...
-            #   2. machine-variant-specific qualities if exist
-            #   3. machine-material-specific qualities if exist
-            #   4. machine-specific qualities if exist
-            #   5. generic qualities if exist
+            #   1. machine-nozzle-buildplate-and-material-specific qualities if exist
+            #   2. machine-nozzle-and-material-specific qualities if exist
+            #   3. machine-nozzle-specific qualities if exist
+            #   4. machine-material-specific qualities if exist
+            #   5. machine-specific qualities if exist
+            #   6. generic qualities if exist
```
ianpaschal: Maybe an idea to put this in the top of the file since it's sort of the crux of adding build plates, right?

ianpaschal:
This is pretty gnarly but I don't think there's a better implementation for the current architecture.

PR #4189, opened 2018-08-03
Only show material packages when clicking an author

ianpaschal:
Contributes to CURA-5604

> *Please do **not** delete this branch after merging. Save it for future Toolbox related changes and PRs.*

### DIFF
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -783,7 +783,16 @@ class Toolbox(QObject, Extension):
         self._models[modelType].setFilter({ filterType: parameter })
         self.filterChanged.emit()
 
-    @pyqtSlot()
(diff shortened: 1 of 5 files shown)


plugins/Toolbox/src/Toolbox.py
```diff
@@ -783,7 +783,15 @@ def filterModelByProp(self, modelType: str, filterType: str, parameter: str):
         self._models[modelType].setFilter({ filterType: parameter })
         self.filterChanged.emit()
 
-    @pyqtSlot()
+    @pyqtSlot(str, "QVariantMap")
+    def setFilters(self, modelType: str, filterDict: dict):
```
LipuFei: Variables names should be lower cases.

LipuFei: Changed in 1bacdabf3

PR #4200, opened 2018-08-06
JSON fix: only enable skin settings when there is skin

BagelOrb:
before: (truncated)
image

after:
image


BagelOrb:
Note that there's a little bug which causes parent settings to be shows of invisible child settings which match the search term. That is irrelevant to this PR, though.

alekseisasin:
@BagelOrb , If you do not mind I can check the fix and commit to master then you do not need to solve merge conflict.

PR #4207, opened 2018-08-07
CURA-5578 Handle alternative_ids in XML material

LipuFei:
Companion PR: https://github.com/Ultimaker/fdm_materials/pull/41

### DIFF
--- a/plugins/XmlMaterialProfile/XmlMaterialProfile.py
+++ b/plugins/XmlMaterialProfile/XmlMaterialProfile.py
@@ -1,15 +1,15 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
+from collections import namedtuple
 import copy
 import io
-import json #To parse the product-to-id mapping file.
-import os.path #To find the product-to-id mapping.
+import json
+import os
 import sys
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, List, Optional, cast
 import xml.etree.ElementTree as ET
 from typing import Dict
-from typing import Iterator
 
 from UM.Resources import Resources
 from UM.Logger import Logger
@@ -21,9 +21,13 @@ from UM.ConfigurationErrorMessage import ConfigurationErrorMessage
 
 from .XmlMaterialValidator import XmlMaterialValidator
 
+HotendNameAndAlternativeNamesResult = namedtuple("HotendNameAndAlternativeNamesResult",
+                                                 ["main_name", "alternative_name_list"])
+
+
 ##  Handles serializing and deserializing material containers from an XML file
 class XmlMaterialProfile(InstanceContainer):
-    CurrentFdmMaterialVersion = "1.3"
+    CurrentFdmMaterialVersion = "1.4"
     Version = 1
 
     def __init__(self, container_id, *args, **kwargs):
@@ -41,7 +45,7 @@ class XmlMaterialProfile(InstanceContainer):
     #   \return The corresponding setting_version.
     @classmethod
     def xmlVersionToSettingVersion(cls, xml_version: str) -> int:
-        if xml_version == "1.3":
+        if xml_version == "1.4":
             return CuraApplication.SettingVersion
         return 0 #Older than 1.3.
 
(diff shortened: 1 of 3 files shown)


plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -41,7 +45,7 @@ def __init__(self, container_id, *args, **kwargs):
     #   \return The corresponding setting_version.
     @classmethod
     def xmlVersionToSettingVersion(cls, xml_version: str) -> int:
-        if xml_version == "1.3":
+        if xml_version == "1.4":
             return CuraApplication.SettingVersion
         return 0 #Older than 1.3.
```
ianpaschal: Don't forget to update comments!

plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -801,6 +809,42 @@ def deserialize(self, serialized, file_name = None):
         for container_to_add in containers_to_add:
             ContainerRegistry.getInstance().addContainer(container_to_add)
 
+    @classmethod
+    def __getHotendNameandAlternativeNamesFromNode(cls, node, machine_id: str, check_if_variants_exist: bool) -> HotendNameAndAlternativeNamesResult:
```
ianpaschal: Capitalize And

plugins/XmlMaterialProfile/XmlMaterialProfile.py
ianpaschal: Don't forget to update comments.

plugins/XmlMaterialProfile/XmlMaterialUpgrader.py
```diff
@@ -5,10 +5,19 @@
...
 
 
+CARTESIO_VARIANT_ID_TO_NEW = {
+    "0.25 mm": {"id": "0.25mm thermoplastic extruder",
+                "alternative_ids": "0.25 mm"},
+    "0.4 mm": {"id": "0.4mm thermoplastic extruder",
+               "alternative_ids": "0.4 mm"},
+    "0.8 mm": {"id": "0.8mm thermoplastic extruder",
+               "alternative_ids": "0.8 mm"},
+}
```
ianpaschal: Also related to below: This feels super messy to me to add all this 3rd party conversion stuff into our code base. Is this not something we can wait until 4.0 to do (or rather, not do, and be ok with breaking changes)?

ChrisTerBeke: Wasn't the plan to put the actual alternative_ids in the material files?

LipuFei: @ChrisTerBeke The plan was to add `alternative_ids`. This is XML material file upgrade, which also needs to upgrade the old files to have `alternative_ids`. Since we are already doing this, it's better to update the id as well.

@ianpaschal I discussed with @diegopradogesto and this can break for, for example, UM3 with a custom PVA because generic_pva has those missing variants of cartesio, so we think it would be nice to have it fixed, although I totally agree with you that this is super messy not to mention that we are using the XML schema version as the `settings_version` of the XML material.

LipuFei: If you guys think it's better not to have this "fix", I'm ok with it.

ChrisTerBeke: Mm this is getting out of hand :P In that case I'd say skip this fix and just say that IDs are supposed to be immutable.

ianpaschal: Yeah I don't think IDs should be mutable. I also think the human-readable thing we do everywhere is really bad practice because we end up with all sorts of IDs that don't reflect the data anymore but are just super confusing.

These should have immutable UUID v4 or whatever.

alekseisasin: I agree that ID should be immutable otherwise let name it different. The CARTESIO_VARIANT_ID_TO_NEW also should be removed, If we want to make fix then I see only create the profile which does not have any specific information. Otherwise what if CARTESIO will not make a fix then these code lines will be forever in Cura.

diegopradogesto: I'm also OK with not fixing it. The only problem is that users will see the configuration errors even if they don't have Cartesio printers, just because this `machine` section is in a lot of material profiles, so for instance `generic_pva` (just to mention one).
We did a mistake approving the original PR that caused this, but now we need to agree in a solution, either live with it or fix it in a horrible way.

diegopradogesto:
In any case, we were talking about adding some GUID to all containers so this won't happen again.

Issue #3959, opened 2018-06-20
Mesh Type Modify settings Overrides Extruder Choice


AbeFM:
As per: https://github.com/Ultimaker/CuraEngine/issues/798

Model printed in one material (extruder #1), when using a modify-settings mesh, will print with the material of the modifying mesh's extruder selection instead of the original object.

**Application Version**
3.4b

**Platform**
win10

**Steps to Reproduce**
Load a model. Add a support blocker. Press "move", select blocker. Resize, reposition.
Select 'per model settings', choose mesh type "modify settings".
Selected "compensate...overlaps" (doubt this matters)

**Actual Results**

![image](https://user-images.githubusercontent.com/42385/41619814-6643df66-73bc-11e8-932e-9401a8a960d1.png)

**Expected results**
Object should continue to print with its own extruder. If the mesh is to modify this, it should be a selection under modify settings.

**Additional Information**
You're wasting my glue, son!

fieldOfView:
I would say this is expected behavior. If anything it should be impossible to change the mesh type of a support blocker.

diegopradogesto:
As I commented in the other thread, this is the expected behavior. I get your point that maybe the "support blocker" doesn't make really sense to assign it to a extruder, but since it's never printed, I don't see why we should not assign one.
When you load an stl and you choose "modify the settings" using "per object settings" it is assigned automatically (and always) to the default extruder, the same way when you load the model you want to print is assigned to the default extruder.

AbeFM:
The other case is, that you chose your object to have the extruder you would like, then you add this modifier, and it doesn't default to where you were.

It would be like if the modifier set the outer walls always to 1, or top pattern always to ZigZag. Great if that happens to be what you want instead of what you had selected, but I feel making changes that are non-explicit confuses people.

Ghostkeeper:
So how would you indicate that part of your mesh needs to be printed with a different extruder, then?

AbeFM:
A pull down menu that says which extruder to print with, 
![image](https://user-images.githubusercontent.com/42385/43816434-a41d0292-9a89-11e8-82b3-5bbfa5fddf12.png)
without editing it too heavily, one can imagine that it says "Print Modified Objects With:" or something more clever. :-)

It's being broken off so it seems possible/easy and something I would use. Something similar could also be used to change supports from inside and under the object, coarsely.

fieldOfView:
[...]

PR #4210, opened 2018-08-08
Fix rendering support eraser cubes

fieldOfView:
This PR fixes rendering support eraser/blocker cubes by creating meshes that have per-vertex normals.

MeshBuilder.addCube creates cubes that don't have per-vertex normals. Geometry like that does not render properly in most shaders used in Cura. This becomes obvious when a user changes the "Mesh Type" using Per Model Settings.

Contributes to https://github.com/Ultimaker/Cura/issues/3959 by making the effect of changing a support blocker mesh to another type more visible and predictable.

### DIFF
--- a/plugins/SupportEraser/SupportEraser.py
+++ b/plugins/SupportEraser/SupportEraser.py
@@ -25,6 +25,8 @@ from cura.Scene.BuildPlateDecorator import BuildPlateDecorator
 
 from UM.Settings.SettingInstance import SettingInstance
 
+import numpy
+
 class SupportEraser(Tool):
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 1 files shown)


diegopradogesto:
Nice! I notice this error but didn't have the time to check it and we thought that was not really important to fix it. Now you did it and we appreciate that.

This is how it looks now when changing a support blocker cube to a "print as support" cube:
image

And this is how it looked before:
image

diegopradogesto:
BTW, @fieldOfView we added Trimesh to our build system. We want to start playing with it as you suggested, trying to fix some annoying bugs. We also plan to use the library to completely remove the MeshBuilder, since it has nice implementations for that as well.

fieldOfView:
I am already working on a “MeshTools” which checks watertightness, splits compound objects and fixes normals (among other things)

Issue #3481, opened 2018-03-13
Support Direction (Angle)


darkdvd:
I posted something on the forum Ultimaker:
https://community.ultimaker.com/topic/21946-support-direction-angle/

ChrisTerBeke:
It seems indeed that there is no infill direction parameter for support infill. I've flagged this ticket with the correct tags.

darkdvd:
Merci Chris !

darkdvd:
I suggest adding a function to the supports management in CURA: "Support Direction".

By default, the supports are printed along the X and Y axes, regardless of the angle of the object.
(Note that the object is positioned by default along the X and Y axes as well).

In certain cases, the geometry of the object to be printed poses a problem during printing: this is particularly the case for objects intended to be molded (The walls of these objects generally have angles of 2 ° in order to optimize demolding, sometimes it is the design of the piece that uses very tight angles ...).

I took as an example the piece that hides the linear bearings of the UM3: this piece is intended to be made of molded plastic (1905-H).

![support_angle_01](https://user-images.githubusercontent.com/16816760/37388252-0cb72bb8-2760-11e8-9070-ba25ec5c6a5e.JPG)

The front of this object has an angle of 2 °.
In my example, I print the object vertically to limit the amount of supports. 

![support_angle_02](https://user-images.githubusercontent.com/16816760/37388277-2468d2a2-2760-11e8-8d2a-20d864481c1c.JPG)

Problem # 1: this wall at 2 ° will be badly printed because the displacement of the head will not be precise enough to restore the wall which must be perfectly rectilinear.

Solution # 1: rotate the object at 45 °. The angle of this wall will be no longer 2 ° but 47 °, which changes everything. 
  
![support_angle_03](https://user-images.githubusercontent.com/16816760/37388294-329a3be0-2760-11e8-83d4-f108d7bc19c8.JPG)

Problem # 2: the supports remain positioned on the X and Y axes and they overflow the object and the printing will necessarily be a failure.

Solution # 2: give the possibility to change the angle of the supports.
In this case, if the object is turned at 45 °, it is necessary to rotate the supports by 45 °, which is not possible at present (or I missed something ....). 
Note #1 : The object contained in the STL has been modified to be printed vertically to limit the number of supports.
Note #2 : I use zigzag supports because they are (far) the best support in terms of finishing and easier to remove.
Note #3 : I am french...

ChrisTerBeke:
[...]

PR #4212, opened 2018-08-08
Add setting for Support Infill Line Direction

fieldOfView:
This PR makes the orientation of infill lines settable. It fixes https://github.com/Ultimaker/Cura/issues/3481

Also see the matching PR for the engine: https://github.com/Ultimaker/CuraEngine/pull/819


fieldOfView:
Fixed the link in the PR. Can't fix the one that was sent out in the notifications.

smartavionics:
What I mean is that your commit contains the following description which doesn't match my idea of what the setting is doing:

````
"description": "The minimum angle of overhangs for which support is added. At a value of 0° all overhangs are supported, 90° will not provide any support.",
````

**Comment by fieldOfView** [comment]
Ah, right, I’ll fix that.

**Comment by diegopradogesto** [comment]
Together with the PR in CuraEngine: CURA-5635

File: resources/definitions/fdmprinter.def.json [hunk 1]
```diff
@@ -3873,6 +3873,19 @@
...
+                "support_infill_angle":
+                {
+                    "label": "Support Infill Line Direction",
+                    "description": "Orientation of the infill pattern for supports. The support infill pattern is rotated in the horizontal plane.",
+                    "unit": "°",
+                    "type": "float",
+                    "minimum_value": "-90",
+                    "maximum_value": "90",
+                    "default_value": 0,
+                    "enabled": "support_enable",
```
alekseisasin: Could you please disable the setting if _Support Pattern_ is Concentric or Cross support.

fieldOfView: Good catch. I've also disabled the setting if the Support Density is set to 0 (in analogy to Infill Line Directions.

fieldOfView:
I have widened the range of allowed angles (from -90, 90 to -180,180). Even though the difference is very subtle with the current patterns, there is a different (with zigzag) and with future patterns the difference may be more dramatic.

PR #4216, opened 2018-08-09
Adjust maximum zoom to printer size

### LINKED ISSUE #3990: Zoom is limited for tall printer dimensions -- cannot zoom to fit
**Application Version**
3.4

**Platform**
MacOS 10.13.5

**Printer**
Custom printer: 500(x)x500(y)x1800mm(z)

**Steps to Reproduce**
Create custom printer with 500 x 500 x 1800 mm dimensions.
machine setup

**Actual Results**
Does not show full volume.

**Expected results**
Should be able to zoom out to show full volume--often called zoom to fit.

**Additional Information**
Prototyping a tall printer for a project. Created a machine setup that is 500mm(x) x 500mm(y) x 1800mm(z). I appears you are limited to har far out you can zoom.

Would assume you could zoom to fit.

[cone.3mf.zip](https://github.com/Ultimaker/Cura/files/2142687/cone.3mf.zip)

Standard 3d view
3d_view

3d view zoomed out to the max. Cannot see full object.
3d_view_max_zoom_out

front view
front_view

front view zoom out to max.
front_view_max_zoom_out

front view panned all the way to the top. Can only see a partial section of the object--never in full.
front_view_panned_to_top

top view panned -- Can see the object full, but not very useful.
top_view_panned

Ghostkeeper:
Previously the maximum distance you could zoom out was 2000. Now it's variable depending on how large your printer is. I've put it to 5 so that it works out to be approximately the same maximum zoom level for normal sized printers (like 25 cube build volume approximately).

This should make it possible to zoom out completely for large sized printers.

Fixes #3990.

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -3,6 +3,7 @@
 
 from cura.Scene.CuraSceneNode import CuraSceneNode
 from cura.Settings.ExtruderManager import ExtruderManager
+from UM.Application import Application #To modify the maximum zoom level.
 from UM.i18n import i18nCatalog
 from UM.Scene.Platform import Platform
 from UM.Scene.Iterator.BreadthFirstIterator import BreadthFirstIterator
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Developers, see issue CURA-5629.

diegopradogesto:
This is the maximum distance for a custom printer of 1m x 1m x 1m:
image

And this is the maximum distance before the changes:
image

I was wondering if there something we can do to set the initial view of the buildplate after restart, depending on the printer size? This is how it looks like when just restarted in a custom printer of 1m x 1m x 1m:
image

PR #4244, opened 2018-08-16
Feature favorite materials

ianpaschal:
This pull request incorporates CURA-5162 and CURA-5378 into one feature addition: favorited materials.

In broad strokes it makes the following changes:

- Adds a preferences setting for favorite material IDs
- In the materials drop down in the sidebar, the generic materials are now in their own group
- Favorite materials also appear in the drop down
- In the materials preferences page, the materials list also includes favorite materials
- Brand and material type sections can be collapsed and expanded, note that this isn't preserved across restarts (but it could be in the future)
- Clicking the blue star adds or removes a material from the favorites section
- This material list is now made of several components, plus other refactoring of the QML in that area

Happy reviewing! Ask me if you have any questions.


cura/Machines/Models/FavoriteMaterialsModel.py
```diff
@@ -0,0 +1,69 @@
...
+        self._extruder_manager = CuraApplication.getInstance().getExtruderManager()
+        self._material_manager = CuraApplication.getInstance().getMaterialManager()
+
+        self._machine_manager.activeStackChanged.connect(self._update) #Update when switching machines.
+        self._material_manager.materialsUpdated.connect(self._update) #Update when the list of materials changes.
+
+        self._update()
+
+    def _update(self):
+        Logger.log("d", "Updating {model_class_name}.".format(model_class_name = self.__class__.__name__))
```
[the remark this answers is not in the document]
ianpaschal: Yeah in hind sight I'm going to reorganize these models a bit anyway so they're reused for the tree structure in the Materials Page as well.

resources/qml/Preferences/Materials/MaterialsList.qml
```diff
@@ -0,0 +1,347 @@
+// Copyright (c) 2018 Ultimaker B.V.
```
ChrisTerBeke: It would be nice if we could split up this huge qml file into a few more simple components.

ianpaschal: It could be. I originally tried to just make a single cascading tree component but Qt did not like my recursive antics so I settled for just splitting the "material slot" piece out. I could split out the fold-out headers too though. Will do that tomorrow.

ianpaschal: Done.

ianpaschal: Also fixed that unfavoriting bug you found. 😉

cura/Machines/MaterialManager.py
```diff
@@ -194,6 +198,11 @@ def initialize(self):
 
         self.materialsUpdated.emit()
 
+        favorites = self._application.getPreferences().getValue("cura/favorite_materials")
```
LipuFei: I would move the favorites to a different place and `MaterialManager` only handles essential material lookup.

cura/Machines/Models/BaseMaterialsModel.py
```diff
@@ -2,45 +2,64 @@
...
         super().__init__(parent)
-        self._application = Application.getInstance()
+
+        from cura.CuraApplication import CuraApplication
+
+        self._application = CuraApplication.getInstance()
+
+        # Make these managers available to all material models
+        self._container_registry = self._application.getInstance().getContainerRegistry()
+        self._extruder_manager = self._application.getExtruderManager()
```
LipuFei: No need to use `ExtruderManager`.

cura/Machines/MaterialManager.py
LipuFei: We decided to keep this in. It's not so coupled with other things.

cura/Machines/Models/BaseMaterialsModel.py
LipuFei: Removed in 33f495442

Issue #4259, opened 2018-08-19
Request : Tool Change Enhancement


Weruminger:
**1. Fan control**
It would be helpfull to set "P" parameter for G-Code "M106" to the actual used Tool Index by CURA:
sample :
```
; some gcode ...
T0 
; some gcode ...
M106 P0 S127
; some gcode ...
M106 P0 S0 ; this can be set in Machine settings in extruder0 "End" code  
T1 
M106 P1 S127 ; current fan speed value can be set in Machine settings in extruder1 "Start" code
; some gcode ...
M106 P1 S255 
: 
; end of gcode ...
M106 P0 S0 ; this can be set in Machine settings in machine "End" code  
M106 P1 S0 ; this can be set in Machine settings in machine "End" code  
```
**2. Machine Retraction**
Using G10, G11 and G10 S1, G11 S1 for retraction and tool change retraction will produce a more universal g-code (e.g. running on equal machines with different tool change technique.
also the g-code becomes smaller and more readability or post processing possibility. 

**Application Version**
CURA 3.4.1 
request for future version

**Platform**
Windows 10 / X64
request is platform independend (g-code engine)

**Printer**
DIY I3 Mendel based on CTC I3 Clone 
upgraded with a pure magnetic tool chain machanism (w/o magnetic solenoid)
[Construction](https://github.com/Weruminger/Magnetic_Parking_Extruder.git)
request is usefull for any printer with a tool change upgrade
[MPE_TEST_Color2.zip](https://github.com/Ultimaker/Cura/files/2300264/MPE_TEST_Color2.zip)

**Steps to Reproduce**
NONE

**Actual Results**
NONE

**Expected results**
NONE

**Additional Information**

smartavionics:
This is related to something I did a little while ago which was to be able to specify the M106/M107 P value as a setting. It didn't get merged into cura because most printers are happy with the default value of 0. If I made that setting capable of being specified per-extruder, would that do what you want?

smartavionics:
Just to clarify - you don't want the tool index to be used for the M10[6/7] P value because machines that have multiple extruders but only one print cooling fan want to use the same P value for all tool indices. So you need to have the print cooling fan index as a setting that can be set per extruder and which has a default value of 0 (for compatibility with how it works now). Agreed?

Weruminger:
[...]

PR #4267, opened 2018-08-21
Add machine_extruder_cooling_fan_number setting to specify print cooling fan for each extruder.

smartavionics:
As discussed in https://github.com/Ultimaker/Cura/issues/4259.


alekseisasin:
I have tested your pull request and it worked well. We need some external tests and then we might close the discussion ticket.

PR #4268, opened 2018-08-21
Added extruder setting field for cooling fan number.

smartavionics:
Provide easy way of setting the per-extruder print cooling fan number (see #4267).


smartavionics:
I made this a separate PR in case you wish to use the setting but don't want to add a field to the machine settings dialog.

PR #4270, opened 2018-08-22
Fix fx reset button

alekseisasin:
CURA-5646


cura/Settings/MachineManager.py
```diff
@@ -1531,3 +1531,31 @@ def _updateUponMaterialMetadataChange(self) -> None:
         with postponeSignals(*self._getContainerChangedSignals(), compress = CompressTechnique.CompressPerParameterValue):
             self.updateMaterialWithVariant(None)
             self._updateQualityWithMaterial()
+
+    @pyqtSlot(QObject)
+    def calculateSettingValue(self, propertyProvider):
```
LipuFei: This is also not good because now `MachineManager` depends on the PropertyProviders which are UI modules. It will be very difficult to separate the core functionality out after this.

nallath:
I've fixed the issue in 054a49c63de38423bb3ee6e6b1146be6ebd09e55

PR #4273, opened 2018-08-22
add PauseAtHeight post-processing script for RRF

Kriechi:
This PR adds a post-processing script to add a Pause-At-Height step for RepRapFirmware, which inserts a M226 at the specified height (layer).
This works for RRF, commonly used with DuetWifi, Duet Ethernet, and Duet Maestro boards (also some RADDS-based ones).


Ghostkeeper:
At some point we must consider merging all of these variants into one big pause script that chooses the correct implementation based on the g-code flavour or something.

For now, developers, see issue CURA-5683.

PR #4275, opened 2018-08-22
Clean up firmware updating

fieldOfView:
This PR cleans up firmware updating:

* factors updating firmware out of main USBPrinterOutputDevice class so that class becomes easier to read
* adds a FirmwareUpdater class that can be subclassed by other outputdevices (eg OctoPrint) to also offer firmware updating
* factors a hardcoded list of known firmware into metadata in printer definitions, which opens the way for packages that include definitions + firmware files
* removes an unnecessary QMLSingleton
* moves UpgradeFirmwareMachineAction out of UltimakerMachineActions
* moves FirmwareUpdateWindow to the UpgradeFirmwareMachineAction
* adds typing

This PR is a first step into cleaning up USB printing, so I can start implementing #1049 again.


cura/FirmwareUpdater.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+from PyQt5.QtCore import QObject, QUrl, pyqtSignal, pyqtProperty
```
ChrisTerBeke: It would be nice if this class is inside the PrinterOutput folder instead of in the root cura folder.

ChrisTerBeke:
@nallath you're probably interested in this concerning your monitor rewrite (e.g. a NetworkClusterUpdater implementation?).

cura/FirmwareUpdater.py
ChrisTerBeke: It's sending something to a connected printer (just not a print job). Given that it's tightly coupled with the other printer output functionality I think it's nice to have it in the same place. The root folder is all over the place :P

fieldOfView: PrinterOutputDevice.py also isn't in PrinterOutput. Shall I move that too?

Appesteijn:
6/6 Test #6: code-style .......................***Failed   12.12 sec

fieldOfView:
This PR changes 21 files. Just saying "***Failed" without any further detail is about as helpful as a screen door on a submarine. Thanks Jenkins.

Appesteijn:
I thought it was better than 'continuous-integration/jenkins/pr-merge — This commit has test failures'
You should be able to run the tests locally with 'ctest --verbose' in the build directory

fieldOfView:
I noticed this PR does not have a JIRA issue yet. Is it being considered for the next release?

Appesteijn:
You are right, CURA-5701. We are finishing up for the 3.5 beta this might not make it.

cura/Settings/MachineManager.py
```diff
@@ -1531,3 +1533,35 @@ def _updateUponMaterialMetadataChange(self) -> None:
         with postponeSignals(*self._getContainerChangedSignals(), compress = CompressTechnique.CompressPerParameterValue):
             self.updateMaterialWithVariant(None)
             self._updateQualityWithMaterial()
+
+    ##  Get default firmware file name if one is specified in the firmware
+    @pyqtSlot(result = str)
+    def getDefaultFirmwareName(self):
```
LipuFei: I guess this function can be moved to `GlobalStack`? Prevent `MachineManager` from uncontrollable growth.

resources/definitions/ultimaker_original.def.json
```diff
@@ -18,7 +18,9 @@
         "machine_extruder_trains":
         {
             "0": "ultimaker_original_extruder_0"
-        }
+        },
+        "firmware_file": "MarlinUltimaker-{baudrate}.hex",
+        "firmware_hbk_file": "MarlinUltimaker-HKB-{baudrate}.hex"
```
LipuFei: Looks like it should be `hkb` because of `HKB`.

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,82 @@
...
+        self._output_device = output_device
+
+        self._update_firmware_thread = Thread(target=self._updateFirmware, daemon=True)
+
+        self._firmware_view = None
+        self._firmware_location = None
+        self._firmware_progress = 0
+        self._firmware_update_state = FirmwareUpdateState.idle
+
+    def updateFirmware(self, file):
```
nallath: Missing typing

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,82 @@
...
+        # the file path could be url-encoded.
+        if file.startswith("file://"):
+            self._firmware_location = QUrl(file).toLocalFile()
+        else:
+            self._firmware_location = file
+        self.showFirmwareInterface()
+        self.setFirmwareUpdateState(FirmwareUpdateState.updating)
+        self._update_firmware_thread.start()
+
+    def _updateFirmware(self):
```
nallath: Missing typing

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,82 @@
...
+        raise NotImplementedError("_updateFirmware needs to be implemented")
+
+    def cleanupAfterUpdate(self):
```
nallath: I think you get the point by now. We want some typing :)

cura/PrinterOutput/PrinterOutputController.py
```diff
@@ -55,3 +57,9 @@ def homeHead(self, printer: "PrinterOutputModel"):
 
     def sendRawCommand(self, printer: "PrinterOutputModel", command: str):
         Logger.log("w", "Custom command not implemented in controller")
+
+    canUpdateFirmwareChanged = Signal()
+    def setCanUpdateFirmware(self, can_update_firmware: bool):
```
nallath: Missing return type -> None

cura/Settings/MachineManager.py
```diff
@@ -1531,3 +1533,35 @@ def _updateUponMaterialMetadataChange(self) -> None:
...
+        # Check if there is a valid global container stack
+        if not self._global_container_stack:
+            return ""
+
+        # The bottom of the containerstack is the machine definition
+        machine_id = self._global_container_stack.getBottom().id
```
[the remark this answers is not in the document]
fieldOfView: As a matter of fact, the variable wasn't really used.
Fixed in 9a98341

cura/Settings/MachineManager.py
```diff
@@ -1531,3 +1533,35 @@ def _updateUponMaterialMetadataChange(self) -> None:
...
+        machine_has_heated_bed = self._global_container_stack.getProperty("machine_heated_bed", "value")
+
+        baudrate = 250000
+        if platform.system() == "Linux":
```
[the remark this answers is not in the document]
fieldOfView: Fixed in 9a98341

cura/Settings/MachineManager.py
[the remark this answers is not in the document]
fieldOfView: Fixed in 9a98341

fieldOfView: Good point, fixed in bc52830

cura/Settings/GlobalStack.py
```diff
@@ -200,6 +202,31 @@ def getHasVariants(self) -> bool:
...
+            # Linux prefers a baudrate of 115200 here because older versions of
+            # pySerial did not support a baudrate of 250000
+            baudrate = 115200
+
+        # If a firmware file is available, it should be specified in the definition for the printer
+        hex_file = self.getMetaDataEntry("firmware_file", None)
+        if machine_has_heated_bed:
+            hex_file = self.getMetaDataEntry("firmware_hbk_file", hex_file)
+
+        if hex_file:
```
ChrisTerBeke: This would be simpler to read by using an early return

```python
if not hex_file:
    Logger.log(...)
    return ""

try:
....
```

cura/PrinterOutput/FirmwareUpdater.py
fieldOfView: Fixed the typing in a573a59

resources/definitions/ultimaker_original.def.json
fieldOfView: "hbk" is "heated bed kit", and the filename refers to files in the cura-binary-data repo. Thanks for spotting the typo.

fieldOfView:
I am getting a lot of MYPY errors in the ConvexHullDecorator, but I did not touch that code nor does it touch my code, so I am guessing those CI failures are not my fault,

alekseisasin:
ConvexHullDecorator is not related to your ticket. We need to fix it.

nallath:
The convex hull stuff is fixed in the account branch. Once that's merged, these issues should be gone.

nallath:
Unfortunately, it doesn't work. I found the following issues (and stopped at some point)
1. the super().__init__ of firmwareUpdater is never called
2. firmwareUpdater doesn't have a self.ResourceTypes (should be CuraApplication.ResourceTypes)
3. firmware updater doesn't have a ._serial_port. (when changing that to ._output_device._serial_port, it also breaks, because then you're trying to open more than one port).

fieldOfView:
Back to the drawing board. I don't often have a printer connected through USB either, so I must have broken some things while cleaning up after testing it myself.

fieldOfView:
I'm working on another QML cleanup. I'll reopen once I have properly tested this. Shouldn't take more than a day.

nallath:
No problem. I'm happy you're doing this in the first place :)

fieldOfView:
I'm wondering if UpgradeFirmwareMachineAction should become a plugin of its own, instead of being part of UltimakerMachineActions. The other plugins in UltimakerMachineActions are fairly Ultimaker specific, but upgrading firmware is relevant for almost all printers.

What do you think?

nallath:
Yeah, i think that's a good idea. It used to be that we only had specific (um) firmwares but the upload custom (and the fact that BQ also added a few firmware versions) put an end to that.

fieldOfView:
Now much more betterer

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self._output_device = output_device
+
+        self._update_firmware_thread = Thread(target=self._updateFirmware, daemon=True)
+
+        self._firmware_location = ""
+        self._firmware_progress = 0
+        self._firmware_update_state = FirmwareUpdateState.idle
+
+    def updateFirmware(self, file: Union[str, QUrl]) -> None:
```
nallath: file shadows built in (i'd suggest renaming it to file_location or something like that (This also makes it a bit clearer that it's a path location not an actual file handle)

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @pyqtProperty(float, notify = firmwareProgressChanged)
+    def firmwareProgress(self) -> float:
+        return self._firmware_progress
+
+    @pyqtProperty(int, notify=firmwareUpdateStateChanged)
+    def firmwareUpdateState(self) -> "FirmwareUpdateState":
+        return self._firmware_update_state
+
+    def _setFirmwareUpdateState(self, state) -> None:
```
nallath: state is missing a type.

cura/PrinterOutput/FirmwareUpdater.py
```diff
@@ -0,0 +1,78 @@
...
+        if self._firmware_update_state != state:
+            self._firmware_update_state = state
+            self.firmwareUpdateStateChanged.emit()
+
+    # Callback function for firmware update progress.
+    def _onFirmwareProgress(self, progress, max_progress = 100) -> None:
```
nallath: progress & max_progress are missing types.

cura/PrinterOutput/GenericOutputController.py
```diff
@@ -1,35 +1,36 @@
...
 class GenericOutputController(PrinterOutputController):
-    def __init__(self, output_device):
+    def __init__(self, output_device: "PrinterOutputDevice") -> None:
         super().__init__(output_device)
 
         self._preheat_bed_timer = QTimer()
         self._preheat_bed_timer.setSingleShot(True)
         self._preheat_bed_timer.timeout.connect(self._onPreheatBedTimerFinished)
-        self._preheat_printer = None
+        self._preheat_printer = None #type: Optional[PrinterOutputModel]
```
nallath: nitpicking; pep sez there needs to be 2 spaces before an inline coment and a single space after #

cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -91,7 +91,7 @@ def updateOwner(self, owner):
     def assignedPrinter(self):
         return self._assigned_printer
 
-    def updateAssignedPrinter(self, assigned_printer: "PrinterOutputModel"):
+    def updateAssignedPrinter(self, assigned_printer: Optional["PrinterOutputModel"]):
```
nallath: More nitpicking; Return type should also be None.

plugins/FirmwareUpdater/FirmwareUpdaterMachineAction.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self._active_output_device = None #type: Optional[PrinterOutputDevice]
+        self._active_firmware_updater = None #type: Optional[FirmwareUpdater]
+
+        CuraApplication.getInstance().engineCreatedSignal.connect(self._onEngineCreated)
+
+    def _onEngineCreated(self) -> None:
+        CuraApplication.getInstance().getMachineManager().outputDevicesChanged.connect(self._onOutputDevicesChanged)
+
+    def _onContainerAdded(self, container) -> None:
```
nallath: Missing type for container (I think that ContainerInterface should be sufficient here)

plugins/FirmwareUpdater/FirmwareUpdaterMachineAction.qml
```diff
@@ -0,0 +1,191 @@
...
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+
+Cura.MachineAction
+{
+    anchors.fill: parent;
+    property bool printerConnected: Cura.MachineManager.printerConnected
+    property var activeOutputDevice: printerConnected ? Cura.MachineManager.printerOutputDevices[0] : null
+    property var canUpdateFirmware: activeOutputDevice ? activeOutputDevice.activePrinter.canUpdateFirmware : false
```
nallath: nitpicking; property bool is a tiny bit faster (less conversion required)

plugins/FirmwareUpdater/FirmwareUpdaterMachineAction.qml
```diff
@@ -0,0 +1,191 @@
...
+            width: parent.width
+            text: catalog.i18nc("@title", "Update Firmware")
+            wrapMode: Text.WordWrap
+            font.pointSize: 18
+        }
+        Label
+        {
+            width: parent.width
+            wrapMode: Text.WordWrap
+            text: catalog.i18nc("@label", "Firmware is the piece of software running directly on your 3D printer. This firmware controls the step motors, regulates the temperature and ultimately makes your printer work.")
```
nallath: I'm not a 100% sure if it can happen, but if we set a width, we should also set elide, right?

plugins/FirmwareUpdater/FirmwareUpdaterMachineAction.qml
```diff
@@ -0,0 +1,191 @@
...
+            wrapMode: Text.WordWrap
+            text: catalog.i18nc("@label", "The firmware shipping with new printers works, but new versions tend to have more features and improvements.");
+        }
+
+        Row
+        {
+            anchors.horizontalCenter: parent.horizontalCenter
+            width: childrenRect.width
+            spacing: UM.Theme.getSize("default_margin").width
+            property var firmwareName: Cura.MachineManager.activeMachine.getDefaultFirmwareName()
```
nallath: Same nitpicking as above; it's probably a tad bit faster if we use property string (also for readability, since then it's clear it's always a string)

plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -107,9 +98,16 @@ def _onConfirmExitDialogResult(self, result: bool) -> None:
             application = CuraApplication.getInstance()
             application.triggerNextExitCheck()
 
+    @pyqtSlot(str)
+    def updateFirmware(self, file: Union[str, QUrl]) -> None:
```
nallath: Shadows built-in file (see previous comment)

plugins/FirmwareUpdater/FirmwareUpdaterMachineAction.qml
fieldOfView: No, because wrapMode is set to Text.WordWrap, so this Label renders multiple lines and will not elide.

cura/PrinterOutput/FirmwareUpdater.py
fieldOfView: Fixed in a36deea

cura/PrinterOutput/FirmwareUpdater.py
fieldOfView: Fixed in a36deea

cura/PrinterOutput/GenericOutputController.py
fieldOfView: Fixed in a36deea

cura/PrinterOutput/PrintJobOutputModel.py
fieldOfView: Fixed in a36deea

plugins/USBPrinting/USBPrinterOutputDevice.py
fieldOfView: Fixed in a36deea

Ghostkeeper:
Code looks nice. I hope we'll be able to merge this.

Ghostkeeper:
My test fails though. I connected an Ultimaker 2 via a USB cable and started an automatic firmware update. I got the following error:

    2018-10-11 17:08:27,551 - DEBUG - [Thread-4] USBPrinting.AutoDetectBaudJob.run [49]: Checking /dev/ttyACM0 if baud rate 250000 works. Retry nr: 0. Wait timeout: 3
    2018-10-11 17:08:28,319 - DEBUG - [MainThread] cura.AutoSave._onTimeout [48]: Autosaving preferences, instances and profiles
    2018-10-11 17:08:29,067 - DEBUG - [Thread-4] USBPrinting.AutoDetectBaudJob.run [79]: Detected baud rate 250000 on serial /dev/ttyACM0 on retry 0 with after 0.01 seconds.
    2018-10-11 17:08:29,134 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/ruben/Projects/Cura/resources/qml/Preferences/MachinesPage.qml:162: TypeError: Cannot read property 'connectionText' of null
    2018-10-11 17:08:29,141 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/ruben/Projects/Cura/resources/qml/Menus/ConfigurationMenu/PrintCoreConfiguration.qml:73: TypeError: Cannot read property 'name' of null
    Exception in thread Thread-12:
      File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
        self.run()
      File "/usr/lib/python3.6/threading.py", line 864, in run
        self._target(*self._args, **self._kwargs)
      File "/home/ruben/Projects/Cura/cura/../plugins/USBPrinting/AvrFirmwareUpdater.py", line 25, in _updateFirmware
        hex_file = intelHex.readHex(self._firmware_location)
    AttributeError: 'AvrFirmwareUpdater' object has no attribute '_firmware_location'
    
    2018-10-11 17:08:33,512 - INFO - [Thread-11] USBPrinting.USBPrinterOutputDevice._setFirmwareName [293]: USB output device Firmware name: Marlin Ultimaker2

It looks like this field was moved to FirmwareUpdater but not updated here.

fieldOfView:
This PR is cursed. CURSED, I tells ya.

I broke it while making requested changes, but it should be fixed again now.

PR #4288, opened 2018-08-25
bq Hephestos2: Getting a rid of custom g-code

thopiekar:
* Getting a rid of M800:
This removes M800 completely. The problem is basically that M800 sets the temperature to 200deg.
As a result we tell the firmware to heat up to e.g. 205deg, then M800 kicks in and sets the temperature to 200deg and then we heat up to 205deg again.
Therefore I looked into the firmware and extracted all g-code lines except the one which tells to go to 200deg.

* Getting independent of bq's fork and therefore adding compatibility to Marlin
We use standard g-code, which is safe to be used with both firmware.

Note: This change might be also interesting for other bq products, but can't test on them. Therefore it is up to bq and/or the owners of these products to contribute and correct this here, as I did.

### DIFF
--- a/resources/definitions/bq_hephestos_2.def.json
+++ b/resources/definitions/bq_hephestos_2.def.json
@@ -17,8 +17,8 @@
 
     "overrides": {
         "machine_name": { "default_value": "BQ Hephestos 2" },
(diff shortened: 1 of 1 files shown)


Appesteijn:
Checked and looks good 👍

thopiekar:
Great, thanks!

PR #4299, opened 2018-08-29
Remove the 40 char maximum on printer address textfield

Appesteijn:


Appesteijn:
We got a request from a customer to make it possible to enter more that 40 characters in the hostname/url field

PR #4300, opened 2018-08-29
change .curaproject.3mf to just use .3mf

Appesteijn:
CURA-5650

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -85,14 +85,6 @@ class ThreeMFWorkspaceReader(WorkspaceReader):
     def __init__(self) -> None:
         super().__init__()
 
-        MimeTypeDatabase.addMimeType(
-            MimeType(
-                name="application/x-curaproject+xml",
-                comment="Cura Project File",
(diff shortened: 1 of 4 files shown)


ChrisTerBeke:
Will this break opening existing .curaproject.3mf files, or is the OS smart enough to detect that?

Appesteijn:
It should not. The 'opening filter' is also adjusted to .3mf and .curaproject.3mf ends in .3mf

plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -89,7 +89,7 @@ def __init__(self) -> None:
             MimeType(
                 name="application/x-curaproject+xml",
                 comment="Cura Project File",
-                suffixes=["curaproject.3mf"]
+                suffixes=["3mf"]
```
LipuFei: I don't think this will work. It either overwrites the MimeType `3MFReader` registers or the other way around.

The project file detection is done with `checkIsValidProjectFile()` in `CuraApplication.py`, and regardless wether the given file is a "3mf" or not, it always performs the check. I think it's better removing this mime type registration to avoid that overwriting problem.

Appesteijn:
Did changes

PR #4301, opened 2018-08-29
CL-893 CL-894 Cura connect ui rework

sedwards2009:
https://jira.ultimaker.com:8443/browse/CL-893

https://jira.ultimaker.com:8443/browse/CL-894


Appesteijn:
Code style still fails

plugins/UM3NetworkPrinting/ClusterMonitorItem.qml
```diff
@@ -25,93 +25,85 @@ Component
...
+            anchors.fill: parent
+            camera: OutputDevice.activeCamera
         }
 
         onVisibleChanged:
         {
-            if (!monitorFrame.visible)
+            if (monitorFrame != null && !monitorFrame.visible)
             {
                 // After switching the Tab ensure that active printer is Null, the video stream image
```
sedwards2009: This comment is slightly out of date.

cura/PrintJobPreviewImageProvider.py
```diff
@@ -0,0 +1,27 @@
...
+
+from UM.Application import Application
+
+
+class PrintJobPreviewImageProvider(QQuickImageProvider):
+    def __init__(self):
+        super().__init__(QQuickImageProvider.Image)
+
+    ##  Request a new image.
+    def requestImage(self, id, size):
```
sedwards2009: Can we have some types plz?

cura/PrinterOutput/ExtruderConfigurationModel.py
```diff
@@ -1,56 +1,67 @@
...
 
+from cura.PrinterOutput.MaterialOutputModel import MaterialOutputModel
+
 
 class ExtruderConfigurationModel(QObject):
 
     extruderConfigurationChanged = pyqtSignal()
 
-    def __init__(self):
+    def __init__(self, position: int = -1):
```
sedwards2009: This needs a `-> None`.

cura/PrinterOutput/ExtruderConfigurationModel.py
```diff
@@ -1,56 +1,67 @@
...
         self._position = position
 
     @pyqtProperty(int, fset = setPosition, notify = extruderConfigurationChanged)
-    def position(self):
+    def position(self) -> int:
         return self._position
 
-    def setMaterial(self, material):
-        self._material = material
+    def setMaterial(self, material: Optional[MaterialOutputModel]):
```
sedwards2009: Needs a `-> None`.

cura/PrinterOutput/ExtruderOutputModel.py
```diff
@@ -12,64 +12,61 @@
...
-        self._printer = printer
+        self._printer = printer  # type: PrinterOutputModel
         self._position = position
-        self._target_hotend_temperature = 0 # type: float
-        self._hotend_temperature = 0 # type: float
-        self._hotend_id = ""
-        self._active_material = None  # type: Optional[MaterialOutputModel]
-        self._extruder_configuration = ExtruderConfigurationModel()
-        self._extruder_configuration.position = self._position
+        self._target_hotend_temperature = 0  # type: float
```
sedwards2009: It is a good idea to use a real float here. 0.0

cura/PrinterOutput/ExtruderOutputModel.py
```diff
@@ -12,64 +12,61 @@
...
             return self._printer.canPreHeatHotends
         return False
 
-    @pyqtProperty(QObject, notify = activeMaterialChanged)
+    @pyqtProperty(QObject, notify = extruderConfigurationChanged)
     def activeMaterial(self) -> Optional["MaterialOutputModel"]:
-        return self._active_material
+        return self._extruder_configuration.activeMaterial
 
     def updateActiveMaterial(self, material: Optional["MaterialOutputModel"]):
```
sedwards2009: Needs a `-> None`.

cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -29,6 +36,48 @@ def __init__(self, output_controller: "PrinterOutputController", key: str = "",
...
+    def updatePreviewImage(self, preview_image: Optional[QImage]):
+        if self._preview_image != preview_image:
+            self._preview_image = preview_image
+            self.previewImageChanged.emit()
+
+    @pyqtProperty(QObject, notify=configurationChanged)
+    def configuration(self) -> Optional["ConfigurationModel"]:
+        return self._configuration
+
+    def updateConfiguration(self, configuration: Optional["ConfigurationModel"]) -> None:
```
sedwards2009: This method is really a setter not an updater. It hangs on to the config which is given to it. It doesn't really update itself based the config.

plugins/UM3NetworkPrinting/ClusterControlItem.qml
```diff
@@ -25,217 +30,663 @@ Component
...
+                                    text: modelData.activePrintJob != null ? modelData.activePrintJob.owner : ""
+                                    font: UM.Theme.getFont("default")
+                                    opacity: 0.6
+                                    width: parent.width
+                                    elide: Text.ElideRight
+                                }
+
+                                function switchPopupState()
+                                {
+                                    popup.visible ? popup.close() : popup.open()
```
sedwards2009: Just use a normal `if()` here and leave the `?:` for expressions.

LipuFei:
Code style is failing.

plugins/UM3NetworkPrinting/PrintJobInfoBlock.qml
```diff
@@ -0,0 +1,359 @@
...
+                font: UM.Theme.getFont("default_bold")
+                text:
+                {
+                    if(printJob.assignedPrinter == null)
+                    {
+                        return catalog.i18nc("@label", "Waiting for: first available")
+                    }
+                    else
+                    {
+                        return catalog.i18nc("@label", "Waiting for: ") + printJob.assignedPrinter.name
```
sedwards2009: Ideally this would use formatting for the i18n string and not assume so much about how every language should build up this bit of text. Also it really should be combined with the  i18n above. The "Waiting for: " bit is meant to be the same although the translators might not realise it.

plugins/UM3NetworkPrinting/PrintJobInfoBlock.qml
```diff
@@ -0,0 +1,378 @@
...
+                    {
+                        if(printJob.state == "error")
+                        {
+                            return catalog.i18nc("@label", "Waiting for: Unavailable printer")
+                        }
+                        return catalog.i18nc("@label", "Waiting for: First available")
+                    }
+                    else
+                    {
+                        return catalog.i18nc("@label", "Waiting for: ") + printJob.assignedPrinter.name
```
sedwards2009: Ideally this would use formatting for the i18n string and not assume so much about how every language should build up this bit of text. Also it really should be combined with the i18n above. The "Waiting for: " bit is meant to be the same although the translators might not realise it.

cura/PrinterOutput/PrintJobOutputModel.py
nallath: It is an updater, since set<something> is used to actually call the linked device. So if you want to update the info (which is what this does), updateConfiguration is required.

plugins/UM3NetworkPrinting/ClusterControlItem.qml
nallath: This was copied from the other cura code.

cura/PrinterOutput/PrintJobOutputModel.py
ianpaschal: I am not going to change this one.

I don't think the semantics are that important here. The question of overwrite vs. merging data isn't really addressed (we don't have two functions to handle them differently), and replacing a value can be a form of updating. Plus I think its nice that the updaters/setters all follow the same format (makes it easy to guess what the function is called).

plugins/UM3NetworkPrinting/PrintJobInfoBlock.qml
ianpaschal: I see what you're saying but I'm not experienced enough with i18nc to know what to do about it.

diegopradogesto:
It would be nice if there is some structure within the plugin, such as storing the images in a folder called `resources\images` and the qml files in `resources\qml` or something like that. There are now a lot of files all in the same folder.

ianpaschal:
@diegopradogesto This is true, although considering it's all going to be re-written to fit the new Cura style in the next few months (and hopefully use more modular re-usable) I wasn't going to invest time in breaking it apart. But I agree. Can't keep 700-line QML files forever.

ianpaschal:
Also I'm afraid that splitting things up without having a really firm grasp of how it fits together (seeing as I didn't write it) might jeopardize getting it into the 3.5 beta before the end of the day. So I'm going to merge it.

Issue #495, opened 2015-10-20
OS X Fullscreen application


Xisle:
Cura 15.09.91
OS X 10.9.5

Cura basically dies when put in fullscreen mode (ctrl-command-F). The application goes all white and doesn't seem to recover even after exiting fullscreen mode.

Ghostkeeper:
My condolences.

TMHogenhout:
I found this as well on Windows 7 (only then it goes black..).

awhiemstra:
Since there are several major upstream bugs relating to fullscreen we have temporarily disabled the fullscreen option. Once the upstream bugs are resolved we can re-enable it.

PR #4307, opened 2018-08-30
Re-Enable fullscreen shortcut and add menu on ViewMenu

KangDroid:
Referring https://github.com/Ultimaker/Cura/issues/495 issue, thought fullscreen won't work at all, but reverting https://github.com/Ultimaker/Cura/commit/a94961d28bebc9bfbc1c7981227f8a254896d7b3 this commit worked for macOS High Sierra 10.13.6. 

Tested with python3 cura_app.py (Running from source) - MBA(2014). 

Actually todo is this: actually fullscreen with COMMAND(CTRL+Command+F) works, but not with

this one(Green button). It just expand its size, but not going in to full screen mode.


resources/qml/Actions.qml
```diff
@@ -75,6 +75,7 @@ Item
     Action
     {
         id:toggleFullScreenAction
+		shortcut: StandardKey.FullScreen;
```
fieldOfView: Please replace tabs with spaces

resources/qml/Menus/ViewMenu.qml
```diff
@@ -73,4 +73,5 @@ Menu
     MenuSeparator {}
 
     MenuItem { action: Cura.Actions.expandSidebar; }
+	MenuItem { action: Cura.Actions.toggleFullScreen; }
```
fieldOfView: Please replace tabs with spaces

resources/qml/Actions.qml
KangDroid: Re-pushed again. Thanks for review!

fieldOfView:
Works on Windows too (using F11). Personally, I would put a `MenuSeparator {}` above the menuitem.

KangDroid:
@fieldOfView Yeah, I've put MenuSeparator{} above fullscreen one.


As I mentioned in TODO, is there any way to put cura a fullscreen mode by 
 ?

PR #4311, opened 2018-08-31
Add keyboard accessibility for Settings>Profiles>Manage Profile...

lf-:
This menu item is missing a key shortcut, so let's add one.

Unfortunately I haven't been able to test this due to issues getting Cura to run with system versions of Uranium-git. Localization of some sort builds with cmake/make fine. Please tell me if I'm misunderstanding any part of the i18n system with these changes.


diegopradogesto:
FYI @alekseisasin

alekseisasin:
Hi @lf- , do not worry about this, we already have a fix: https://github.com/Ultimaker/Cura/pull/4326 .It will be already in the next release.

PR #4313, opened 2018-08-31
Build models using reserved tags

ianpaschal:
Contributes to CURA-5670

Note: When testing this, you won't see any plugins in the plugins showcase because no plugins are currently tagged as "showcase". This has to be fixed in packages repository, not Cura.

### DIFF
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -83,7 +83,7 @@ class Toolbox(QObject, Extension):
             "plugins_available":   PackagesModel(self),
             "plugins_installed":   PackagesModel(self),
             "materials_showcase":  AuthorsModel(self),
-            "materials_available": PackagesModel(self),
+            "materials_available": AuthorsModel(self),
             "materials_installed": PackagesModel(self),
             "materials_generic":   PackagesModel(self)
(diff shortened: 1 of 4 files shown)


ianpaschal:
As mentioned to @ChrisTerBeke, this also means that the showcase API end-point used in Cura 3.4 will not be used in Cura 3.5+.

ChrisTerBeke:
@ianpaschal CI fails ;)

ianpaschal:
Oh right, I knew that would happen, then walked away from my computer, and when I get back couldn't remember what was that thing I was going to do. So I committed anyway. 😎

plugins/Toolbox/src/Toolbox.py
```diff
@@ -517,7 +517,7 @@ def getNumberOfInstalledPackagesByAuthor(self, author_id: str) -> int:
     @pyqtSlot(str, result = int)
     def getTotalNumberOfPackagesByAuthor(self, author_id: str) -> int:
         count = 0
-        for package in self._metadata["materials_available"]:
+        for package in self._metadata["packages"]:
```
diegopradogesto: This method is just used for the materials, so the "number of packages" means in fact "number of material packages". We should keep the code, but maybe change the name of the method or add documentation to avoid confusion.

PR #4316, opened 2018-09-01
Add Creality Ender-3 machine definition

stelgenhof:
At the moment Cura does not have machine definition available for the Creality Ender-3 printer. Commonly, the Creality CR-10 profile is used instead for the Ender-3 with making a few alterations. 

This PR contains a proper machine definition for the Creality Ender-3 and (simple) Ender-3 platform mesh. The machine definition contains custom start and end G-code that is an accumulation of various sources and tests and maybe somewhat opinionated. These can be adjusted of course if needed.


Appesteijn:
👍

diegopradogesto:
Every printer needs to add also a extruder definition file.

stelgenhof:
@diegopradogesto Oh you're right. I looked at printer definitions made by Prusa Research for their printers and they didn't use an extruder definition file. Since these printers have only one extruder, I thought it wasn't necessary.

I'll create a new PR with the extruder definition file.

PR #4326, opened 2018-09-04
Cura 5642 remove shortkeys from translation

alekseisasin:
The Alt keys, that are normally used for controlling the menu items, don't work properly. Very notifiable in other languages such as German.
The case is actually doe to the same shortcut key appears more than one time in the same scope.


LipuFei:
@alekseisasin I moved the check to a standalone script and added it as a pre-check in CI. The changes seem to be good except we still have one conflict in the French translation:
```
The following messages have the same shortcut key 't':
 - [resources/i18n/fi_FI/cura.po] L   4025 : [{'message': '&Tiedosto', 'shortcut_key': 't'}]
 - [resources/i18n/fi_FI/cura.po] L   4097 : [{'message': '&Toolbox', 'shortcut_key': 't'}]
```

LipuFei:
Sorry, not French. Finnish.

PR #4346, opened 2018-09-09
Found a dead line while reading some code

thopiekar:


nallath:
Good find!

thopiekar:
Thanks. My pleasure 😉

PR #4349, opened 2018-09-10
Get the generic materials

diegopradogesto:
Since the tag-filtered calls to the api don't work yet, we need to hack this part as well.


Ghostkeeper:
Do the tag filtered API calls work by now?

PR #4353, opened 2018-09-10
Insert gcode at layer change

wporter82:
I needed the ability to manually send commands on layer change and couldn't find a solution so I made this one up to fill that need.


Ghostkeeper:
As seen on Reddit!

Developers, see issue CURA-5713.

Kriechi:
@wporter82 I made a similar script to inject some instructions for timelapse creation with the Duet & RepRapFirmware. I think there is a good overlap - so maybe we can merge our approaches together! Let me know what you think!

wporter82:
@Kriechi I have added the option for parking the head to be optional as well as looked at what your scripts were doing for the Duet. I added the wait for moves to finish but for the picture taking command, the default I have set is to use the Marlin code M240 but can be changed by the user to something like what you have.

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -0,0 +1,99 @@
...
+        feed_rate = self.getSettingValueByKey("park_feed_rate")
+        park_print_head = self.getSettingValueByKey("park_print_head")
+        x_park = self.getSettingValueByKey("head_park_x")
+        y_park = self.getSettingValueByKey("head_park_y")
+        trigger_cmd = self.getSettingValueByKey("trigger_cmd")
+        pause_length = self.getSettingValueByKey("pause_length")
+
+        gcode_to_append = self.putValue(M = 400) + ";Wait for moves to finish\n"
+        if park_print_head:
+	        gcode_to_append = self.putValue(G = 90) + ";Absolute positioning\n"
```
LipuFei: This line and the next in this `if` block contain <TAB>s as indents.

alekseisasin:
@wporter82 , I have tried _Before_ and _After_ and all of them has the same output, did not see any difference.

alekseisasin:
@wporter82 , do you still want to add this script?

wporter82:
I agree with @dector in that this can cause problems if considerations are not taken. I've considered writing a complete plugin just for time lapse instead of going this route of basically find/replace. If this appears to cause more possible problems than it helps solve, I would think it should be abandoned for a better solution.

alekseisasin:
Right, a solution which might introduce another issue is not a good scenario.
If will be no more effort done for this ticket then I suggest closing this issue.

plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def getSettingDataString(self):
+        return """{
+            "name": "Insert at layer change",
+            "key": "InsertAtLayerChange",
+            "metadata": {},
+            "version": 2,
+            "settings":
+            {
+                "insert_loc":
```
diegopradogesto: I would not abbreviate the setting name:
```suggestion
                "insert_location":
```

plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
```diff
@@ -0,0 +1,55 @@
...
+            # Check that a layer is being printed
+            lines = layer.split("\n")
+            if ";LAYER:" in lines[0]:
+                in_layer = True
+            else:
+                in_layer = False
+
+            if in_layer:
+                index = data.index(layer)
+                if self.getSettingValueByKey("insert_loc") == "before":
```
diegopradogesto:
```suggestion
                if self.getSettingValueByKey("insert_location") == "before":
```

plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
```diff
@@ -0,0 +1,55 @@
...
+        gcode_to_add = self.getSettingValueByKey("gcode_to_add") + "\n"
+        for layer in data:
+            # Check that a layer is being printed
+            lines = layer.split("\n")
+            if ";LAYER:" in lines[0]:
+                in_layer = True
+            else:
+                in_layer = False
+
+            if in_layer:
```
diegopradogesto: Suggestion: This can be simplified by using `if ";LAYER:" in lines[0]:` and remove the 4 previous layers.

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def getSettingDataString(self):
+        return """{
+            "name": "Time Lapse",
+            "key": "TimeLapse",
+            "metadata": {},
+            "version": 2,
+            "settings":
+            {
+                "trigger_cmd":
```
diegopradogesto: Do not abbreviate setting name:
```suggestion
                "trigger_command":
```

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }"""
+
+    def execute(self, data):
+        in_layer = False
+        feed_rate = self.getSettingValueByKey("park_feed_rate")
+        park_print_head = self.getSettingValueByKey("park_print_head")
+        x_park = self.getSettingValueByKey("head_park_x")
+        y_park = self.getSettingValueByKey("head_park_y")
+        trigger_cmd = self.getSettingValueByKey("trigger_cmd")
```
diegopradogesto:
```suggestion
        trigger_command = self.getSettingValueByKey("trigger_command")
```

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -0,0 +1,100 @@
...
+        pause_length = self.getSettingValueByKey("pause_length")
+        gcode_to_append = ";TimeLapse Begin\n"
+
+        if park_print_head:
+            gcode_to_append += self.putValue(G = 1, F = feed_rate, X = x_park, Y = y_park) + ";Park print head\n"
+        gcode_to_append += self.putValue(M = 400) + ";Wait for moves to finish\n"
+        gcode_to_append += trigger_cmd + ";Snap Photo\n"
```
diegopradogesto:
```suggestion
        gcode_to_append += trigger_command + ";Snap Photo\n"
```

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -0,0 +1,100 @@
...
+        gcode_to_append += ";TimeLapse End\n"
+        for layer in data:
+            # Check that a layer is being printed
+            lines = layer.split("\n")
+            if ";LAYER:" in lines[0]:
+                in_layer = True
+            else:
+                in_layer = False
+
+            if in_layer:
```
diegopradogesto: Suggestion: This can be simplified by using `if ";LAYER:" in lines[0]:` and remove the 4 previous layers.

diegopradogesto:
@wporter82 are you handling the suggested changes? No rush, only to know if we need to keep this in mind or not.

diegopradogesto:
I did some changes here 93966281708a1072fe4ae7f4efbfc3e54a179707 and merged it since it was taking a lot of lead time.

Ghostkeeper:
We might want to add the option to retract during the move towards the park location as well.

PR #4356, opened 2018-09-12
Changes to Dagoma printers profiles.

dagorel:
Hi all,

Here is a small pull request to change some stuff around Dagoma printers profiles.

What is done:
- Rename Dagoma Magis profile.
- Remove machine_gcode_flavor definition as the default value fits the
needs.

Regards,
Orel


nallath:
By changing these files, you also need to ensure that the upgrade script can handle this. Since the name of the files is actually used as ID, this change will mean that everyone that had the machine added will lose their configurations without it (because it's still referring to the dagoma_neva_magis inside their configurations)

dagorel:
Hi again,

After a quick look, i guess it is in https://github.com/Ultimaker/Cura/blob/master/plugins/VersionUpgrade/VersionUpgrade34to35/VersionUpgrade34to35.py

Could you confirm please?

Regards,
Orel

dagorel:
Actually your solution is the best one for 3.5.
I will revert back file names for this release and will come back to change them for the next one.

Regards,

nallath:
Right, I didn't realize that no-one had the profiles in the first place! My bad for not checking this. 

If there is no-one with the profile in the first place, there is no upgrade needed.

nallath:
As far as I'm concerned you can.

nallath:
I've cherry-picked your commit to 3.5 (093c5487bff2509797d14f5ebba1ffb956ee98cb). Could you check if everything is correct now?

dagorel:
> I've cherry-picked your commit to 3.5 ([093c548](https://github.com/Ultimaker/Cura/commit/093c5487bff2509797d14f5ebba1ffb956ee98cb)). Could you check if everything is correct now?

Hi @nallath,
I've just realized i forgot to rename the associated quality files...
That's the purpose of the last commit related to this pull request : [d996e42](https://github.com/Ultimaker/Cura/pull/4356/commits/d996e42f51b124b7a27e3dccd973f565e08cc755).

nallath:
Right. Just saw it as well. Fixed that by hand.

dagorel:
Thank you very much for your support ;)

nallath:
Since all the changes are in, this pull request can be closed.

PR #4363, opened 2018-09-13
Feat: support brim

BagelOrb:
image
image

I had some issues with my print, so I decided to fix the issues with a sweat and cool little pull request.

1) control the interaction between brim and support

2) generate a brim under support with the support extruder

new settings:
Enable Support Brim
Support Brim Width
Support Brim Line Count
and
Brim Replaces Support

This PR are the frontend settings for: https://github.com/Ultimaker/CuraEngine/pull/849

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4538,6 +4538,17 @@
                         }
                     }
                 },
+                "brim_replaces_support":
+                {
+                    "label": "Brim Replaces Support",
+                    "description": "Enforce brim to be printed around the model even if that space would otherwise be occupied by support. This replaces some regions fo the first layer of supprot by brim regions.",
+                    "type": "bool",
+                    "default_value": true,
+                    "enabled": "resolveOrValue('adhesion_type') == 'brim' and support_enable",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true,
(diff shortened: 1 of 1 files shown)


BagelOrb:
I had an adhesion problem similar to this situation:
image

The second material wouldn't stick to the brim of the first material.

With the new settings the adhesion is greatly improved:
image

BagelOrb:
I think the support brim should be enabled by default and Brim Replaces Support should be disabled by default.

It used to be the case that the brim replaced support. This was so because if a model has support all around then it could be the case that it didn't receive any brim, which could lead to the part detaching from the build plate. However, if we now enable support brim by default we don't have that problem any more and the support brim ensures that there is always brim around the model.

moreover the support had quite bad adhesion in general, so that is now also improved.

BagelOrb:
I'm not sure whether the support should still get a brim on the outisde with the adhesion extruder, but that is something we can figure out some other time.

LipuFei:
@BagelOrb I "stole" your branch to a local branch `feat_support_brim` because Harma wants to have a build to test. :P

Appesteijn:
CURA-5760

rburema:
There're some small typo's in a description, but pushing to your fork will be more trouble than that's worth, so I'll merge this, then fix those in master.

BagelOrb:
Yay!

PR #4366, opened 2018-09-13
CL-1045 Cura connect improvements

ianpaschal:
DO NOT MERGE YET!

Just opening to start to view the diffs and review.

[CL-1045](https://jira.ultimaker.com:8443/browse/CL-1045)

### DIFF
--- a/cura/PrinterOutput/PrintJobOutputModel.py
+++ b/cura/PrinterOutput/PrintJobOutputModel.py
@@ -12,6 +12,8 @@ if TYPE_CHECKING:
     from cura.PrinterOutput.PrinterOutputModel import PrinterOutputModel
     from cura.PrinterOutput.ConfigurationModel import ConfigurationModel
 
+from cura.PrinterOutput.ConfigurationChangeModel import ConfigurationChangeModel
+
 
 class PrintJobOutputModel(QObject):
     stateChanged = pyqtSignal()
@@ -24,6 +26,7 @@ class PrintJobOutputModel(QObject):
     configurationChanged = pyqtSignal()
     previewImageChanged = pyqtSignal()
     compatibleMachineFamiliesChanged = pyqtSignal()
+    configurationChangesChanged = pyqtSignal()
 
     def __init__(self, output_controller: "PrinterOutputController", key: str = "", name: str = "", parent=None) -> None:
         super().__init__(parent)
@@ -41,6 +44,7 @@ class PrintJobOutputModel(QObject):
         self._preview_image_id = 0
 
         self._preview_image = None  # type: Optional[QImage]
+        self._configuration_changes = []    # type: List[ConfigurationChangeModel]
 
     @pyqtProperty("QStringList", notify=compatibleMachineFamiliesChanged)
     def compatibleMachineFamilies(self):
@@ -148,3 +152,13 @@ class PrintJobOutputModel(QObject):
     @pyqtSlot(str)
     def setState(self, state):
         self._output_controller.setJobState(self, state)
+
+    @pyqtProperty("QVariantList", notify=configurationChangesChanged)
+    def configurationChanges(self) -> List[ConfigurationChangeModel]:
+        return self._configuration_changes
+
+    def updateConfigurationChanges(self, changes: List[ConfigurationChangeModel]) -> None:
+        if len(self._configuration_changes) == 0 and len(changes) == 0:
+            return
+        self._configuration_changes = changes
+        self.configurationChangesChanged.emit()
--- a/plugins/UM3NetworkPrinting/resources/qml/CameraButton.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/CameraButton.qml
@@ -19,7 +19,7 @@ Rectangle
     UM.RecolorImage
     {
         id: icon
-        width: parent.width - 1
+        width: parent.width / 2
         height: width
         anchors.verticalCenter: parent.verticalCenter
         anchors.horizontalCenter: parent.horizontalCenter
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -652,7 +652,14 @@ Component
                                         hoverEnabled: true
                                         onClicked:
                                         {
-                                            OutputDevice.setActiveCamera(modelData.camera)
+                                            if (OutputDevice.activeCamera !== null)
+                                            {
+                                                OutputDevice.setActiveCamera(null)
+                                            }
+                                            else
+                                            {
+                                                OutputDevice.setActiveCamera(modelData.camera)
+                                            }
                                         }
                                     }
                                 }
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -337,7 +337,7 @@ Component
                                 {
                                     id: printJobName
                                     text: modelData.activePrintJob != null ? modelData.activePrintJob.name : ""
-                                    font: UM.Theme.getFont("default_bold")
+                                    font: UM.Theme.getFont("default")
                                     anchors.left: parent.left
                                     anchors.right: contextButton.left
                                     anchors.rightMargin: UM.Theme.getSize("default_margin").width
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -380,7 +380,7 @@ Component
                                 {
                                     id: printJobName
                                     text: modelData.activePrintJob != null ? modelData.activePrintJob.name : ""
-                                    font: UM.Theme.getFont("default")
+                                    font: UM.Theme.getFont("default_bold")
                                     anchors.left: parent.left
                                     anchors.right: contextButton.left
                                     anchors.rightMargin: UM.Theme.getSize("default_margin").width
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -223,7 +223,7 @@ Component
                                     width: parent.width
                                     elide: Text.ElideRight
                                     font: UM.Theme.getFont("default")
-                                    opacity: 0.6
+                                    color: UM.Theme.getColor("monitor_secondary_text")
                                 }
                             }
 
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -638,19 +638,24 @@ Component
 
                             style: ProgressBarStyle
                             {
+                                property var remainingTime:
+                                {
+                                    if(modelData.activePrintJob == null)
+                                    {
+                                        return 0
+                                    }
+                                    /* Sometimes total minus elapsed is less than 0. Use Math.max() to prevent remaining
+                                        time from ever being less than 0. Negative durations cause strange behavior such
+                                        as displaying "-1h -1m". */
+                                    var activeJob = modelData.activePrintJob
+                                    return Math.max(activeJob.timeTotal - activeJob.timeElapsed, 0);
+                                }
                                 property var progressText:
                                 {
                                     if(modelData.activePrintJob == null)
                                     {
                                         return ""
                                     }
-
-                                    /* Sometimes total minus elapsed is less than 0. Use Math.max() to prevent remaining
-                                        time from ever being less than 0. Negative durations cause strange behavior such
-                                        as displaying "-1h -1m". */
-                                    var activeJob = modelData.activePrintJob
-                                    var remainingTime = Math.max(activeJob.timeTotal - activeJob.timeElapsed, 0);
-
                                     switch(modelData.activePrintJob.state)
                                     {
                                         case "wait_cleanup":
@@ -663,6 +668,7 @@ Component
                                         case "sent_to_printer":
                                             return catalog.i18nc("@label:status", "Preparing")
                                         case "aborted":
+                                            return catalog.i18nc("@label:status", "Aborted")
                                         case "wait_user_action":
                                             return catalog.i18nc("@label:status", "Aborted")
                                         case "pausing":
@@ -687,7 +693,24 @@ Component
 
                                 progress: Rectangle
                                 {
-                                    color: UM.Theme.getColor("primary")
+                                    color:
+                                    {
+                                        var state = modelData.activePrintJob.state
+                                        var deactiveStates = [
+                                            "pausing",
+                                            "paused",
+                                            "resuming",
+                                            "wait_cleanup"
+                                        ]
+                                        if(deactiveStates.indexOf(state) > -1 && remainingTime > 0)
+                                        {
+                                            return UM.Theme.getColor("monitor_secondary_text")
+                                        }
+                                        else
+                                        {
+                                            return UM.Theme.getColor("primary")
+                                        }
+                                    }
                                     id: progressItem
                                     function getTextOffset()
                                     {
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -775,7 +775,6 @@ Component
                                         text: progressText
                                         anchors.verticalCenter: parent.verticalCenter
                                         color: progressItem.width + progressLabel.width < control.width ? "black" : "white"
-                                        }
                                         width: contentWidth
                                         font: UM.Theme.getFont("default")
                                     }
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -696,13 +696,13 @@ Component
                                     color:
                                     {
                                         var state = modelData.activePrintJob.state
-                                        var deactiveStates = [
+                                        var inactiveStates = [
                                             "pausing",
                                             "paused",
                                             "resuming",
                                             "wait_cleanup"
                                         ]
-                                        if(deactiveStates.indexOf(state) > -1 && remainingTime > 0)
+                                        if(inactiveStates.indexOf(state) > -1 && remainingTime > 0)
                                         {
                                             return UM.Theme.getColor("monitor_secondary_text")
                                         }
--- a/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
@@ -15,7 +15,7 @@ Component
     {
         id: base
         property var lineColor: "#DCDCDC" // TODO: Should be linked to theme.
-
+        property var shadowRadius: 5 * screenScaleFactor
         property var cornerRadius: 4 * screenScaleFactor // TODO: Should be linked to theme.
         visible: OutputDevice != null
         anchors.fill: parent
@@ -82,6 +82,8 @@ Component
 
             ListView
             {
+                id: printer_list
+                property var current_index: -1
                 anchors
                 {
                     top: parent.top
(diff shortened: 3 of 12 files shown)


nallath:
Is it intentional that this is merged into master (eg; Will not be in 3.5?)

ianpaschal:
@nallath Eh, either way it needs to be cherry picked one way or another right?

fieldOfView:
3.5 can be merged into master, but master should not be merged into 3.5

So from 3.5 to master, no cherry-picking is necessary.

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
```diff
@@ -389,6 +424,14 @@ Component
                                         radius: 0.5 * width
                                         color: UM.Theme.getColor("viewport_background")
                                     }
+                                    contentItem: Label
+                                    {
+                                        text: contextButton.text
+                                        color: UM.Theme.getColor("monitor_text_inactive")
+                                        font.pixelSize: 25
```
sedwards2009: Is screen scaling needed here?

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -2,13 +2,119 @@ import QtQuick 2.2
...
 
 import UM 1.3 as UM
 
 
 Item
 {
     id: base
+
+    function haveAlert() {
+        return printJob.configurationChanges.length !== 0;
```
nallath: JavaScript calls are slower than direct assignments in properties.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -27,6 +133,9 @@ Item
     Rectangle
     {
         id: background
+
+        height: (cardHeight + alertHeight()) * screenScaleFactor
```
nallath: Using dynamic heights is strongly discouraged by Qt, since it's much slower. Anchors are the prefered solution.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -2,13 +2,119 @@ import QtQuick 2.2
...
+
+            if (name.slice(-extension.length) === extension) {
+                name = name.substring(0, name.length - extension.length);
+            }
+        }
+
+        return name;
+    }
+
+    function isPrintJobForcable(printJob) {
```
nallath: Why is this logic in the QML?

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -2,13 +2,119 @@ import QtQuick 2.2
...
+            case 'aluminum':
+                translationText = catalog.i18nc("@label", "Aluminum");
+                break;
+            default:
+                translationText = null;
+        }
+        return translationText;
+    }
+
+    function formatPrintJobName(name) {
```
nallath: This shouldn't be in QML in the first place (This should be in the model, where the printJob object is created). This can also be done dynamicly without hardcoding extentions.
It's also a legacy thing, so document it (newer printers don't show this)

plugins/UM3NetworkPrinting/resources/qml/CameraButton.qml
```diff
@@ -0,0 +1,47 @@
+import QtQuick 2.3
+import QtQuick.Controls 1.4
+import QtQuick.Controls.Styles 1.3
+import QtQuick.Controls 2.0 as Controls2
+import QtGraphicalEffects 1.0
+
+import UM 1.3 as UM
+import Cura 1.0 as Cura
+
+Rectangle
```
nallath: I'm not sure if I initially built it like that, but this should just be a button. You can set the label of a button to be an image and the background to be an item. That will do what you want with a fraction of the code (and probably with more optimisations possible for the QML engine)

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
nallath: Also, shouldn't the config changes just load an item if there is a config change? This is what loaders are intended to do and it also ensures that items that are not needed don't get created.

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -406,6 +407,11 @@ def deleteJobFromQueue(self, print_job_uuid: str) -> None:
         # is a modification of the cluster queue and not of the actual job.
         self.delete("print_jobs/{uuid}".format(uuid = print_job_uuid), on_finished=None)
 
+    @pyqtSlot(str)
+    def forceSendJob(self, print_job_uuid: str) -> None:
```
nallath: See my previous comment about removing the config changes and switch to warning. If you change that, you can actually add a overrideWarning function to the printJobOutputModel, which can then be implemented by the configurationOutputController to send this.

cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -147,4 +151,14 @@ def updateState(self, new_state):
...
+        self._output_controller.setJobState(self, state)
+
+    @pyqtProperty("QVariantList", notify=configurationChangesChanged)
+    def configurationChanges(self) -> List[ConfigurationChangeModel]:
+        return self._configuration_changes
+
+    def updateConfigurationChanges(self, changes: List[ConfigurationChangeModel]) -> None:
+        if len(self._configuration_changes) == 0 and len(changes) == 0:
+            return
+        self._configuration_changes = changes
```
nallath: Missing check to see if something changed. The current implementation will ensure that for every printjob with config changes, you will get an emit every 2 seconds (forcing a redraw in qml)

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -574,6 +580,16 @@ def _updatePrintJob(self, print_job: PrintJobOutputModel, data: Dict[str, Any])
         if not status_set_by_impediment:
             print_job.updateState(data["status"])
 
+        print_job.updateConfigurationChanges(self._createConfigurationChanges(data["configuration_changes_required"]))
+
+    def _createConfigurationChanges(self, data: List[Dict[str, Any]]) -> List[ConfigurationChangeModel]:
+        result = []
+        for change in data:
+            result.append(ConfigurationChangeModel(type_of_change=change["type_of_change"],
```
nallath: Why re-create them every update? I'm a bit afraid that this wil wreak havoc on the garbage collect for larger clusters.

cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -147,4 +151,14 @@ def updateState(self, new_state):
 
     @pyqtSlot(str)
     def setState(self, state):
-        self._output_controller.setJobState(self, state)
\ No newline at end of file
+        self._output_controller.setJobState(self, state)
+
+    @pyqtProperty("QVariantList", notify=configurationChangesChanged)
+    def configurationChanges(self) -> List[ConfigurationChangeModel]:
```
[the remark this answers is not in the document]
nallath: If this is behavior that you really, really want in here, I think it's better to create a ClusterPrintJobOutputModel and put it in there.

resources/themes/cura-light/theme.json
```diff
@@ -321,7 +321,12 @@
         "favorites_header_hover": [245, 245, 245, 255],
         "favorites_header_text": [31, 36, 39, 255],
         "favorites_header_text_hover": [31, 36, 39, 255],
-        "favorites_row_selected": [196, 239, 255, 255]
+        "favorites_row_selected": [196, 239, 255, 255],
+
+        "monitor_text_inactive": [154, 154, 154, 255],
+        "monitor_background_inactive": [240, 240, 240, 255],
```
nallath: If you're adding these, it's also up to you to ensure that the other built-in monitor options also use these variables.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
sedwards2009: I didn't see much point in rewriting it for something which is going to thrown away and replaced in a month or two.

cura/PrinterOutput/PrintJobOutputModel.py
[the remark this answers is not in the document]
sedwards2009: I did discuss this with Ruben and he seemed ok with it. Most of the plugin code will be replaced sometime in the near future, but I can see how it would be worth the effort of keeping stuff out of Cura core where possible.

A subclass of `PrintJobOutputModel` sounds like a good solution.

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
sedwards2009: Re-creating on every update is much simpler less buggy than some kind of funky in-place update. We are not talking about many objects and you have to keep in mind that when doing anything in Python it creates and destroys objects like a madman. I strongly suspect that recreating is just faster than doing it in place.

sedwards2009: For example, that `for ... in` loop creates an iterator object and terminates when the collection creates an exception object and throws it. That is just one simple loop.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
fieldOfView: But historically, hasn't that been so for all of Cura?
(ducks for cover)

nallath: It's not going to be replaced in a month or two as far as I know. 

Right now, I don't think these changes should be put in 3.5, since they add new features to the printJobOutput model.

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
nallath: Well, sure, but this is also creating a bunch of QObjects and matching properties. So it's doing a tad bit more. Especially since the update isn't checking if a signal needs to be emitted (so it's constantly triggering the re-drawing)

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
ianpaschal: This is already being fixed in my branch from this branch where I'm rewriting the Print Info Block from scratch to be much cleaner. Don't spend more time reviewing this for now.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
ianpaschal: This is already being fixed in my branch from this branch where I'm rewriting the Print Info Block from scratch to be much cleaner. Don't spend more time reviewing this for now.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
ianpaschal: This is already being fixed in my branch from this branch where I'm rewriting the Print Info Block from scratch to be much cleaner. Don't spend more time reviewing this for now.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
ianpaschal: This is already being fixed in my branch from this branch where I'm rewriting the Print Info Block from scratch to be much cleaner. Don't spend more time reviewing this for now.

plugins/UM3NetworkPrinting/resources/qml/CameraButton.qml
ianpaschal: Nope. You initially built it twice in two places so this was already an optimization. It took a little too much extra work to make it truly reusable since we need to make a reusable button component for Cura 4.0 anyway. This was just a quick little refactor.

resources/themes/cura-light/theme.json
ianpaschal: Yeah I've been replacing hardcoded colors and sizes when I see them... but someone didn't use any from the start haha. 😛

ianpaschal:
Closing since @sedwards2009 merged the relevant stuff into 3.5, which will later be merged back to master.

alekseisasin:
Delete branch?

fieldOfView:
Please make sure that all Labels Rectangles have a (themed) color, so things don't break with the dark theme (or when there is a dark system theme which makes Labels without a color light). Also keep rounding fractional sizes to integer values or we get OSX font rendering issues again. And finally, keep things DPI aware...

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
```diff
@@ -257,8 +285,16 @@ Component
...
+                                color:
+                                {
+                                    if(modelData.state == "disabled")
+                                    {
+                                        return UM.Theme.getColor("monitor_lining_inactive")
+                                    }
+                                    return UM.Theme.getColor("viewport_background")
+                                }
+                                // UM.Theme.getColor("viewport_background")
+                                height: 1
```
fieldOfView: UM.Theme.getSize("default_lining") is at least DPI aware...

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
```diff
@@ -357,21 +400,13 @@ Component
...
-                                    }
+                                    popup.visible ? popup.close() : popup.open()
                                 }
 
                                 Controls2.Button
                                 {
                                     id: contextButton
                                     text: "\u22EE" //Unicode; Three stacked points.
-                                    font.pixelSize: 25
                                     width: 35
```
fieldOfView: not DPI aware

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
fieldOfView: Yes, this should use pointSize instead of pixelSize

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
```diff
@@ -488,19 +549,20 @@ Component
...
                                             {
                                                 id: point
-                                                height: 13
-                                                width: 13
+                                                height: 14 * screenScaleFactor
+                                                width: 14 * screenScaleFactor
                                                 color: UM.Theme.getColor("setting_control")
                                                 transform: Rotation { angle: 45}
                                                 anchors.right: bloop.right
+                                                anchors.rightMargin: 24
```
fieldOfView: Not DPI aware

plugins/UM3NetworkPrinting/resources/qml/ClusterControlItem.qml
```diff
@@ -693,11 +751,28 @@ Component
...
+                                        else
+                                        {
+                                            return UM.Theme.getColor("primary")
+                                        }
+                                    }
                                     id: progressItem
                                     function getTextOffset()
                                     {
-                                        if(progressItem.width + progressLabel.width < control.width)
+                                        if(progressItem.width + progressLabel.width + 16 < control.width)
```
fieldOfView: That 16 is not DPI aware (and why isn't this using default_marging instead of arbitrary 16?

plugins/UM3NetworkPrinting/resources/qml/ClusterMonitorItem.qml
```diff
@@ -50,7 +50,7 @@ Component
             anchors.left: queuedPrintJobs.left
             anchors.top: parent.top
             anchors.topMargin: 2 * UM.Theme.getSize("default_margin").height
-            anchors.leftMargin: 3 * UM.Theme.getSize("default_margin").width
+            anchors.leftMargin: 3 * UM.Theme.getSize("default_margin").width + 5
```
fieldOfView: That 5 is not DPI aware

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -3,20 +3,128 @@ import QtQuick.Dialogs 1.1
...
 Item
 {
     id: base
+
+    function haveAlert() {
+        return printJob.configurationChanges.length !== 0;
+    }
+
+    function alertHeight() {
+        return haveAlert() ? 230 : 0;
```
fieldOfView: DPI awareness? You should have it.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -3,20 +3,128 @@ import QtQuick.Dialogs 1.1
...
+            var typeOfChange = printJob.configurationChanges[i].typeOfChange;
+            if (typeOfChange === 'material_insert' || typeOfChange === 'buildplate_change') {
+                forcable = false;
+            }
+        }
+
+        return forcable;
+    }
+
+    property var cardHeight: 175
```
fieldOfView: It would be much more readable if you factor in the DPI awareness here.

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -168,7 +291,6 @@ Item
             {
                 id: contextButton
                 text: "\u22EE" //Unicode; Three stacked points.
-                font.pixelSize: 25
                 width: 35
```
fieldOfView: Not DPI aware

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -186,6 +308,14 @@ Item
                     radius: 0.5 * width
                     color: UM.Theme.getColor("viewport_background")
                 }
+                contentItem: Label
+                {
+                    text: contextButton.text
+                    color: UM.Theme.getColor("monitor_text_inactive")
+                    font.pixelSize: 25
```
fieldOfView: Should use pointSize instead to be... DPI aware

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -288,19 +436,20 @@ Item
...
                         {
                             id: point
-                            height: 13
-                            width: 13
+                            height: 14 * screenScaleFactor
+                            width: 14 * screenScaleFactor
                             color: UM.Theme.getColor("setting_control")
                             transform: Rotation { angle: 45}
                             anchors.right: bloop.right
+                            anchors.rightMargin: 24
```
fieldOfView: DPI awareness? 

Also, I am getting a deja-vu here. Code duplication?

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -374,28 +523,141 @@ Item
...
+        // Alert / Configuration change box
+        Rectangle
+        {
+            height: alertHeight() * screenScaleFactor
+
+            anchors.left: parent.left
+            anchors.right: parent.right
+            anchors.bottom: parent.bottom
+
+color: "#ff00ff"
```
fieldOfView: Codestyle: indentation.
Also, that color should probably be themeable, or is this one of your "Rectangle-that-should-be-an-Item" elements with a debug color?

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -374,28 +523,141 @@ Item
...
+                    {
+                        id: collapseIcon
+                        width: 15
+                        height: 15
+                        sourceSize.width: width
+                        sourceSize.height: height
+
+                        // FIXME
+                        source: base.collapsed ?  UM.Theme.getIcon("arrow_left") : UM.Theme.getIcon("arrow_bottom")
+                        color: "black"
```
fieldOfView: That's not going to work in dark mode

plugins/UM3NetworkPrinting/resources/qml/PrintJobInfoBlock.qml
```diff
@@ -374,28 +523,141 @@ Item
...
+                        Layout.fillWidth: true
+                    }
+
+                }
+
+                Rectangle
+                {
+                    Layout.fillHeight: true
+                    Layout.fillWidth: true
+color: "red"
```
fieldOfView: More debug Rectangles, I guess

fieldOfView:
> Yup. Got * screenScaleFactor and Math.floor() everywhere.

Nope. You don't.

ianpaschal:
> Nope. You don't.

Yup. I do. In the open pull request which is being reviewed. This PR was closed 2 days ago.
🙄

Issue #3994, opened 2018-06-28
Cura 3.4 Pauses print after heating and again after bed leveling


leeb65:
**Application Version**
3.4.0

**Platform**
Windows 10, version 1803

**Printer**
CTC Prusa i3 Pro B DIY
Modified with Marlin 1.1.8 and 3D Touch bed levelling probe.

**Steps to Reproduce**
Load STL, Prepare and print

**Actual Results**
The printer heats up bed and nozzle, then on the printer screen it says press button to resume, I then have to press the button twice and the printer goes through the bed levelling and stops again with the message Click to Resume. I press the button and it prints as usual.

**Expected results**
The printer should heat up, level the bed and print without interaction.

**Additional Information**
I have run the same print through Octopi and it runs as expected, so it is not the gcode.

[PI3_Toilet_Seat_Bumper_v1-0.zip](https://github.com/Ultimaker/Cura/files/2144739/PI3_Toilet_Seat_Bumper_v1-0.zip)

[cura_Log.zip](https://github.com/Ultimaker/Cura/files/2144944/cura_Log.zip)

alekseisasin:
Hi @leeb65, Unfortunately we do not have CTC Prusa i3 Pro B DIY in our office and I do not know how to test. Can you run the gcode from USB stick? Also, did you have same issue in Cura 3.3.1?

leeb65:
Hi, No I did not have the problem until I updated to Cura 3.4.0 yesterday. I thought it might be my printer, so I connected it to Octopi and saved the gcode from Cura, copied it to Octopi server and set the print off from there. It ran from start to end without a problem.

alekseisasin:
I know we did fix for the auto bed leveling (https://github.com/Ultimaker/Cura/issues/3808), but this is a different issue. (probably if you know how to search in github they were related to ticket CURA-4844)

neowinx:
I have to report that I have this same issue with my Geeetech Prusa Pro W with BL Touch sensor.

After updating Cura to 3.4.0 it pauses at these same spots (after heating and leveling) displayin a "click to resume" message.

Would do some testing to see if this problem only occurs with marlin 1.1.8 with BLTouch.

**Application Version**
3.4.0

**Platform**
Windows 10, version 1803

**Printer**
Geeetech Prusa i3 Pro W
Modified with Marlin 1.1.8 and 3D Touch bed levelling probe.

leeb65:
[...]

PR #4369, opened 2018-09-14
Fix updating temperature while preheating bed or extruder

### LINKED ISSUE #3741: Cura 3.3.1 - Monitoring page temperature is not updating during heating phase
Application Version

3.3.1

Platform

Windows 10

Printer

I3 pro B 

Issue: 
During the pre-heating phase of each print, the temperatur for the extruder and the bed is not update , like it was in 3.2.1

fieldOfView:
This PR fixes updates to the print monitor temperatures while preheating.

While preheating the bed/extruder with M190 or M109, the firmware keeps outputting temperature lines, but these do not contain "ok" because no new command was acknowledged.

This PR is against master, but since it fixes a bug and the risk of regressions is low, it might be considered for cherry-picking to 3.5

Fixes #3741

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -313,6 +313,7 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
         while self._connection_state == ConnectionState.connected and self._serial is not None:
             try:
                 line = self._serial.readline()
+                print(line)
             except:
                 continue
 
@@ -326,8 +327,8 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
                 if self._firmware_name is None:
                     self.sendCommand("M115")
 
-            if (b"ok " in line and b"T:" in line) or line.startswith(b"T:") or b"ok B:" in line or line.startswith(b"B:"):  # Temperature message. 'T:' for extruder and 'B:' for bed
-                extruder_temperature_matches = re.findall(b"T(\d*): ?([\d\.]+) ?\/?([\d\.]+)?", line)
+            if re.search(b"[B|T\d*]: ?\d+\.?\d*", line):  # Temperature message. 'T:' for extruder and 'B:' for bed
+                extruder_temperature_matches = re.findall(b"T(\d*): ?(\d+\.?\d*) ?\/?(\d+\.?\d*)?", line)
                 # Update all temperature values
                 matched_extruder_nrs = []
                 for match in extruder_temperature_matches:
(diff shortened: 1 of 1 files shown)


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -313,6 +313,7 @@ def _update(self):
         while self._connection_state == ConnectionState.connected and self._serial is not None:
             try:
                 line = self._serial.readline()
+                print(line)
```
nallath: I guess that's a debug print that sneaked along with the commit.

nallath:
CURA-5727

plugins/USBPrinting/USBPrinterOutputDevice.py
fieldOfView: 'doh!

alekseisasin:
@fieldOfView , I wonder, will it solve this issue: https://github.com/Ultimaker/Cura/issues/3994
Because the user complain about for a long time about this and we could not reproduce it.

fieldOfView:
I have not been able to reproduce that issue either. If only there were a way to give them a build with this fix in it...

alekseisasin:
Ok, I will make a build and ask somebody to test it.

fieldOfView:
Any chance this will be merged and cherry-picked into 3.5? If it solves the #3994 issue, that would be pretty big.

alekseisasin:
Unfortunately, not for 3.5. It is already closed for adding new stuff.

fieldOfView:
This is not adding new stuff though. It is fixing a bug. What good is a beta period if bugs are not fixed?

(And you've been adding new stuff all through the beta cycle!)

alekseisasin:
It is closed because of the release already should be today.

joba-1:
@alekseisasin if you have not found someone to test, you could ask me.

I have the problem since today. I updated cura and marlin and selected a different first level temperature. I get that pause every time right before the print start.

Unfortunately I couldn't get the build of cura going on my linux box. I would require an app image, not just the patch or I would already have tested it.

TheThief:
@fieldOfView 
> Any chance this will be merged and cherry-picked into 3.5? If it solves the #3994 issue, that would be pretty big.

I've been suffering from that and I'm pretty sure I can see what the problem is - Cura is spamming M105s at the printer which it is ignoring because it's busy heating the bed, and when the printer's command buffer gets full it truncates one into an "M1" instead, which is a pause command.

If that's true then this PR won't fix it, because this doesn't reset `_last_temperature_request` when it receives a temp update that it hasn't asked for - so it will still continue to spam requests. The PR posted in the #3994 bug should fix it because it prevents it from sending temperature requests in the first place.

fieldOfView:
I know. I made the PR posted in #3994 (#4487). I was never convinced that this PR would fix #3994, but still would have liked this PR to have made it into 3.5.

PR #4383, opened 2018-09-18
Test incremental inclusion of setting visibility presets

Ghostkeeper:
This implementation tests whether each setting visibility preset includes more settings than the previous one.

Implements CURA-5734.


Ghostkeeper:
I think we have to write this as an actual automated test in our testing framework. Otherwise everyone will just forget about this script and it will never be run again.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+        # basic.cfg -> advanced.cfg -> expert.cfg. Like: if the setting 'layer_height' in 'basic.cfg' then the same setting
+        # also should be in 'advanced.cfg'
+
+        all_settings_categories = list(self.all_settings_keys.keys())
+
+        # visibility_type = basic, advanced, expert
+        for visibility_type in setting_visibility:
+            item = setting_visibility_items[visibility_type]
+
+            not_valid_setting_by_category[visibility_type] = [] # this list is for keeping invalid settings.
```
Ghostkeeper: Should be a set, not a list.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+            not_valid_categories[visibility_type] = []
```
Ghostkeeper: Should be a set, not a list.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # Check whether the settings are added in right order or not.
+        # The basic settings should be in advanced, and advanced in expert
+        for visibility_type in setting_visibility:
+
+            # Skip the basic because it cannot be compared to previous list
+            if visibility_type == 'basic':
+                continue
+
+            all_settings_in_this_type = []
```
Ghostkeeper: Should be a set, not a list.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+            not_valid_setting_by_order[visibility_type] = []
```
Ghostkeeper: Should be a set, not a list.

Ghostkeeper: If you set this test up as an automated test, it will be unnecessary to track all of this state. You can simply let it fail on the first setting it encounters is wrong. Correcting it and then re-running the test will reveal the next.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+
+            item = setting_visibility_items[visibility_type]
+            for category, category_settings in item.items():
+                all_settings_in_this_type.extend(category_settings)
+
+
+            for setting in visible_settings_order:
+                if setting not in all_settings_in_this_type:
+                    not_valid_setting_by_order[visibility_type].append(setting)
```
Ghostkeeper: This can be implemented easier with set arithmetic. Subtracting the set of advanced settings from the set of basic settings will result in the set of settings that are wrong in advanced.cfg.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+                print("  Visibility type : '%s'" % (type))
+                print("  Incorrect settings : '%s'" % (settings))
+
+        return has_invalid_settings
+
+
+if __name__ == "__main__":
+
+    all_setting_visibility_files = glob.glob(os.path.join(os.path.join(SCRIPT_DIR, "..", "resources", "setting_visibility"), '*.cfg'))
+    fdmprinter_def_path = os.path.join(SCRIPT_DIR, "..", "resources", "definitions", "fdmprinter.def.json")
```
Ghostkeeper: Use `Resources.getPath(Resources.DefinitionContainers)` to get the path to the definition files regardless of where the resource folders are.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+
+    inspector = SettingVisibilityInspection()
+    inspector.defineAllCuraSettings(fdmprinter_def_path)
+
+    setting_visibility_items = {}
+    for file_path in all_setting_visibility_files:
+        temp = inspector.getSettingsFromSettingVisibilityFile(file_path)
+
+        base_name = os.path.basename(file_path)
+        visibility_type = base_name.split('.')[0]
```
Ghostkeeper: Use double quotes `"` for strings.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+
+    all_setting_visibility_files = glob.glob(os.path.join(os.path.join(SCRIPT_DIR, "..", "resources", "setting_visibility"), '*.cfg'))
+    fdmprinter_def_path = os.path.join(SCRIPT_DIR, "..", "resources", "definitions", "fdmprinter.def.json")
+
+    inspector = SettingVisibilityInspection()
+    inspector.defineAllCuraSettings(fdmprinter_def_path)
+
+    setting_visibility_items = {}
+    for file_path in all_setting_visibility_files:
+        temp = inspector.getSettingsFromSettingVisibilityFile(file_path)
```
Ghostkeeper: `temp` is not a variable name.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+if __name__ == "__main__":
+
+    all_setting_visibility_files = glob.glob(os.path.join(os.path.join(SCRIPT_DIR, "..", "resources", "setting_visibility"), '*.cfg'))
+    fdmprinter_def_path = os.path.join(SCRIPT_DIR, "..", "resources", "definitions", "fdmprinter.def.json")
+
+    inspector = SettingVisibilityInspection()
+    inspector.defineAllCuraSettings(fdmprinter_def_path)
+
+    setting_visibility_items = {}
+    for file_path in all_setting_visibility_files:
```
Ghostkeeper: Except the first file. Use `all_setting_visibility_files[1:]` after you've sorted these by weight.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+        temp = inspector.getSettingsFromSettingVisibilityFile(file_path)
+
+        base_name = os.path.basename(file_path)
+        visibility_type = base_name.split('.')[0]
+
+        setting_visibility_items[visibility_type] = temp
+
+    has_invalid_settings = inspector.validateSettingsVisibility(setting_visibility_items)
+
+    sys.exit(0 if not has_invalid_settings else 1)
```
Ghostkeeper: Just use PyTest to fail the test then.

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+#
+from typing import Dict
+import os
+import sys
+import json
+import configparser
+import glob
+
+# Directory where this python file resides
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
```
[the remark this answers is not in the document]
alekseisasin: This script is attached to CI, it is triggered every time, @LipuFei made changes for this.
It was implemented outside the UM because we already had the issue in getting file using the path to a certain file and window build crashed. When we made "check_shortcut_keys.py", @LipuFei knows more details why it was like this.

scripts/check_setting_visibility.py
alekseisasin: Resources.getPath(Resources.DefinitionContainers) 
Returns:
'/home/asasin/.local/share/UM/unknown/definitions' And this folder is empty.

scripts/check_setting_visibility.py
alekseisasin: There was a reason why I followed this approach, in CI it will catch the output and show "fail" or "pass"

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+setting_visibility = ["basic", "advanced", "expert"]
+
+
+class SettingVisibilityInspection:
+
+
+    def __init__(self) -> None:
+        self.all_settings_keys = {}
+
+    def defineAllCuraSettings(self, fdmprinter_json_path: str) -> None:
```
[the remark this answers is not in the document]
alekseisasin: I tried to do it and it fails, 
n UM.Settings.Interfaces.py
'NoneType' object has no attribute 'updateFilesData' in 

in method.
result = VersionUpgradeManager.getInstance().updateFilesData

Probably I'm missing dependecies

scripts/check_setting_visibility.py
```diff
@@ -0,0 +1,180 @@
...
+
+# Directory where this python file resides
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+
+
+# The order of settings type. If the setting is in basic list then it also should be in expert
+setting_visibility = ["basic", "advanced", "expert"]
+
+
+class SettingVisibilityInspection:
```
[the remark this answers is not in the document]
alekseisasin: I added at the top

scripts/check_setting_visibility.py
alekseisasin: ok

scripts/check_setting_visibility.py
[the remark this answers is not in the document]
Ghostkeeper: Could you move the class documentation to the class then?

Ghostkeeper:
Most of the issues I raised in my review of this PR are not yet solved as far as I'm concerned.

nallath:
Agreed. It's a bit better now, but fundamentally it has the same issues.

LipuFei:
My main argument is that, in my opinion, checking resource files is not unit testing. A unit test is to test a single class/file/module, and it should have its own test cases/test data. The test cases are designed specifically to test if a module in a given scenario provides the expected results. The resource data however is for the application to use, so they are not suitable for the purpose unit testing. The scripts we lately added in the `scripts` folder are standalone sanity checks on the application's resource files, so they are not parts of the application code. Although whether to reuse the existing code for certain things like loading definition files is still for debate (I'd like to reuse code, too), one major problem with reusing those code is their huge dependencies, which require a full user environment (we can not mess with user environments on CI) and display to run. So, using the current code, we cannot run those sanity check scripts on CI.

nallath:
Sure, but now there is quite a bit of code duplication. We could also have used a setting definition Container, mock the version upgrader, load the fdmprinter and ask it for all the keys.

It also only needs to compare two sets (eg; subtract list, check if empty). There is no need to find out in what category it's going wrong (we really shouldn't care, its a sanity check).

ChrisTerBeke:
Agree with @Ghostkeeper that it has still a lot of issues, so at least a refactor should be planned. Also the build is only marked as unstable, not failed, so the CI status for this will never be looked at...

(these are suggestions, do with them whatever you want)

Ghostkeeper:
Checking resource files is indeed not unit testing. Then again, PyTest is not only for unit testing but also for acceptance testing, which is what this test does.

PR #4391, opened 2018-09-20
Cura 5095 collect user modified settings

LipuFei:


fieldOfView:
Does this mean Cura has to ask for permission to send this again? The data that is sent is not the data that I have agreed upon.

PR #4393, opened 2018-09-20
[3.5] CURA-5095 Reset send slice info and show privacy dialog

LipuFei:
Because the data Cura collects has been changed.


LipuFei:
@nallath 
 - Unit test added
 - Fixed that Preferences value check in SliceInfo

PR #4408, opened 2018-09-22
Fix showing info about SliceInfo in general preferences

fieldOfView:
This PR fixes a crash in the Cura general preferences, when the user wants to learn more about the information that is sent to Ultimaker when slicing.

See https://community.ultimaker.com/topic/24494-introducing-ultimaker-cura-35-beta/?page=2&tab=comments#comment-219744

Since this is a regression in Cura 3.5 and the patch is very small, this PR is against that branch. Let me know if you prefer it against master.


fieldOfView:
Nevermind, this had already been fixed

PR #4414, opened 2018-09-23
Complete Polish translation for version 3.5

drzejkopf:
Translation was incomplete, phrases related to functions implemented after version 3.3 were missing. I have translated all missing phrases in Polish translation of files cura.po, uranium.po and fdmprinter.def.json.po. For now, translation is complete. I tried running it on 3.5 beta and I can't encounter any issues.
I closed and re-opened pull request, because previous one had attached only one of the files, and I had no idea how to change it.


Ghostkeeper:
Looks good, initially. We'll check these translations together with the rest when the rest of the translations are in too. Thanks!

Developers, see issue CURA-5741 (the translation ticket for 3.5).

PR #4419, opened 2018-09-24
Update target version of the dictionary of doom

houmam:
Update target version of the dictionary of doom to be compliant with the instance container's version number.

### DIFF
--- a/plugins/LegacyProfileReader/DictionaryOfDoom.json
+++ b/plugins/LegacyProfileReader/DictionaryOfDoom.json
@@ -1,6 +1,6 @@
 {
     "source_version": "15.04",
-    "target_version": 3,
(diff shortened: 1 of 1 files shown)


diegopradogesto:
This `target_version` allow us to keep track of the settings that were in that `version` of the containers within Cura. What it does now is to translate the legacy profiles to this version and if it's not the latest version, then the VersionUpgradeManager will take care of it.

That means that just increasing the `target_version` is not correct because some settings has changed between version 3 and 4, so those has to be updated as well. You will need to also update the translations to match the latest in the version 4 of containers.

That being said, we think is better to leave it like currently is and let the VersionUpgradeManager do its job.

So for several reasons we decided to reject this PR. Thanks anyway for contributing.

PR #4423, opened 2018-09-24
Cura 5744 move oauth login

nallath:
Moving the oauth login from the drivePlugin to a more central place.

### DIFF
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
-from typing import Tuple, Optional, Dict
+from typing import Optional, Dict
 
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty
 
@@ -48,7 +48,7 @@ class Account(QObject):
         self._authorization_service.onAuthStateChanged.connect(self._onLoginStateChanged)
         self._authorization_service.onAuthenticationError.connect(self._onLoginStateChanged)
 
-        self._error_message = None
+        self._error_message = None  # type: Optional[Message]
         self._logged_in = False
 
     @pyqtProperty(bool, notify=loginStateChanged)
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,0 +1,88 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+from typing import Tuple, Optional, Dict
+
+from PyQt5.QtCore.QObject import QObject, pyqtSignal, pyqtSlot, pyqtProperty
+
+from UM.Message import Message
+from cura.OAuth2.AuthorizationService import AuthorizationService
+from cura.OAuth2.Models import OAuth2Settings
+from UM.Application import Application
+
+from UM.i18n import i18nCatalog
+i18n_catalog = i18nCatalog("cura")
+
+
+##  The account API provides a version-proof bridge to use Ultimaker Accounts
+#
+#   Usage:
+#       ``from cura.API import CuraAPI
+#       api = CuraAPI()
+#       api.account.login()
+#       api.account.logout()
+#       api.account.userProfile # Who is logged in``
+#
+class Account(QObject):
+    # Signal emitted when user logged in or out.
+    loginStateChanged = pyqtSignal()
+
+    def __init__(self, parent = None) -> None:
+        super().__init__(parent)
+        self._callback_port = 32118
+        self._oauth_root = "https://account.ultimaker.com"
+        self._cloud_api_root = "https://api.ultimaker.com"
+
+        self._oauth_settings = OAuth2Settings(
+            OAUTH_SERVER_URL= self._oauth_root,
+            CALLBACK_PORT=self._callback_port,
+            CALLBACK_URL="http://localhost:{}/callback".format(self._callback_port),
+            CLIENT_ID="um---------------ultimaker_cura_drive_plugin",
+            CLIENT_SCOPES="user.read drive.backups.read drive.backups.write",
+            AUTH_DATA_PREFERENCE_KEY="cura_drive/auth_data",
+            AUTH_SUCCESS_REDIRECT="{}/cura-drive/v1/auth-success".format(self._cloud_api_root),
+            AUTH_FAILED_REDIRECT="{}/cura-drive/v1/auth-error".format(self._cloud_api_root)
+        )
+
+        self._authorization_service = AuthorizationService(Application.getInstance().getPreferences(), self._oauth_settings)
+
+        self._authorization_service.onAuthStateChanged.connect(self._onLoginStateChanged)
+        self._authorization_service.onAuthenticationError.connect(self._onLoginStateChanged)
+
+        self._error_message = None
+        self._logged_in = False
+
+    @pyqtProperty(bool, notify=loginStateChanged)
+    def isLoggedIn(self) -> bool:
+        return self._logged_in
+
+    def _onLoginStateChanged(self, logged_in: bool = False, error_message: Optional[str] = None) -> None:
+        if error_message:
+            if self._error_message:
+                self._error_message.hide()
+            self._error_message = Message(error_message, title = i18n_catalog.i18nc("@info:title", "Login failed"))
+            self._error_message.show()
+
+        if self._logged_in != logged_in:
+            self._logged_in = logged_in
+            self.loginStateChanged.emit()
+
+    def login(self) -> None:
+        if self._logged_in:
+            # Nothing to do, user already logged in.
+            return
+        self._authorization_service.startAuthorizationFlow()
+
+    #   Get the profile of the logged in user
+    #   @returns None if no user is logged in, a dict containing user_id, username and profile_image_url
+    @pyqtProperty("QVariantMap", notify = loginStateChanged)
+    def userProfile(self) -> Optional[Dict[str, Optional[str]]]:
+        user_profile = self._authorization_service.getUserProfile()
+        if not user_profile:
+            return None
+        return user_profile.__dict__
+
+    def logout(self) -> None:
+        if not self._logged_in:
+            return  # Nothing to do, user isn't logged in.
+
+        self._authorization_service.deleteAuthData()
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -36,11 +36,11 @@ class Account(QObject):
             OAUTH_SERVER_URL= self._oauth_root,
             CALLBACK_PORT=self._callback_port,
             CALLBACK_URL="http://localhost:{}/callback".format(self._callback_port),
-            CLIENT_ID="um---------------ultimaker_cura_drive_plugin",
-            CLIENT_SCOPES="user.read drive.backups.read drive.backups.write",
-            AUTH_DATA_PREFERENCE_KEY="cura_drive/auth_data",
-            AUTH_SUCCESS_REDIRECT="{}/cura-drive/v1/auth-success".format(self._cloud_api_root),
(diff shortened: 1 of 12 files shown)


nallath:
I don't know what the typing issues are since my local mypy is okay with it...

@ChrisTerBeke: I'd like it if you could also have a (quick) look at it since you originally wrote the code.

cura/API/Account.py
```diff
@@ -0,0 +1,104 @@
...
+##  The account API provides a version-proof bridge to use Ultimaker Accounts
+#
+#   Usage:
+#       ``from cura.API import CuraAPI
+#       api = CuraAPI()
+#       api.account.login()
+#       api.account.logout()
+#       api.account.userProfile # Who is logged in``
+#
+class Account(QObject):
```
ChrisTerBeke: Still not a fan of all this logic here, but up to you to deal with it in the future :P

cura/OAuth2/AuthorizationHelpers.py
```diff
@@ -0,0 +1,124 @@
...
+from typing import Optional
+
+import requests
+
+from UM.Logger import Logger
+
+from cura.OAuth2.Models import AuthenticationResponse, UserProfile, OAuth2Settings
+
+
+class AuthorizationHelpers:
```
ChrisTerBeke: I guess now that this code is in the Cura main repo, we should switch to doxygen style comments?

tests/TestOAuth2.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Validate that there is no user profile or token
+    assert authorization_service.getUserProfile() is None
+    assert authorization_service.getAccessToken() is None
+
+
+def test_localAuthServer() -> None:
+    preferences = Preferences()
+    authorization_service = AuthorizationService(preferences, OAUTH_SETTINGS)
+    with patch.object(webbrowser, "open_new") as webrowser_open:
```
ChrisTerBeke: It would be easier to read to use the @patch decorator syntax to prevent this with waterfall.

ChrisTerBeke: Lol, some dude registered his username as patch

cura/API/Account.py
ChrisTerBeke: Note: it could be added to the AuthorizationService (as a service, it could easily be responsible for storing the auth state).

cura/API/Account.py
```diff
@@ -0,0 +1,104 @@
...
+        self._oauth_root = "https://account.ultimaker.com"
+        self._cloud_api_root = "https://api.ultimaker.com"
+
+        self._oauth_settings = OAuth2Settings(
+            OAUTH_SERVER_URL= self._oauth_root,
+            CALLBACK_PORT=self._callback_port,
+            CALLBACK_URL="http://localhost:{}/callback".format(self._callback_port),
+            CLIENT_ID="um---------------ultimaker_cura",
+            CLIENT_SCOPES="user.read drive.backups.read drive.backups.write client.package.download",
+            AUTH_DATA_PREFERENCE_KEY="general/ultimaker_auth_data",
```
ChrisTerBeke: Note: now that this key is changed to something more generic, people that will upgrade to Cura 3.6 (?) and the then updated backups plugin with it, will have to sign in again. Not per se a problem, just noting the implication here.

cura/OAuth2/AuthorizationService.py
```diff
@@ -0,0 +1,162 @@
...
+    def _onAuthStateChanged(self, auth_response: AuthenticationResponse) -> None:
+        """Callback method for an authentication flow."""
+        if auth_response.success:
+            self._storeAuthData(auth_response)
+            self.onAuthStateChanged.emit(logged_in=True)
+        else:
+            self.onAuthenticationError.emit(logged_in=False, error_message=auth_response.err_message)
+        self._server.stop()  # Stop the web server at all times.
+
+    def _loadAuthData(self) -> None:
```
LipuFei: I think this should be in the `Account` class. This service should just do things like launching a browser for login and handling login callbacks. It also makes testing easier.

cura/API/Account.py
ChrisTerBeke: I'm missing a property to get the access token (needed for some plugins to call our own APIs).

cura/API/Account.py
```diff
@@ -0,0 +1,104 @@
...
+#   Usage:
+#       ``from cura.API import CuraAPI
+#       api = CuraAPI()
+#       api.account.login()
+#       api.account.logout()
+#       api.account.userProfile # Who is logged in``
+#
+class Account(QObject):
+    # Signal emitted when user logged in or out.
+    loginStateChanged = pyqtSignal()
```
ChrisTerBeke: Does this signal have a callback argument that passes the new login state? The backups plugin relies on this to see if we need to fetch the backups after logging in.

ChrisTerBeke:
@nallath for reference, this is the usage of this API from the backups plugin: https://github.com/Ultimaker/CuraDrivePlugin/pull/38.

cura/API/__init__.py
```diff
@@ -1,17 +1,20 @@
...
 
 ##  The official Cura API that plug-ins can use to interact with Cura.
 #
 #   Python does not technically prevent talking to other classes as well, but
 #   this API provides a version-safe interface with proper deprecation warnings
 #   etc. Usage of any other methods than the ones provided in this API can cause
 #   plug-ins to be unstable.
-
-class CuraAPI:
+class CuraAPI(QObject):
```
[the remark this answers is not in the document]
nallath: It's either a pyqtProperty or a pyqtSlot. You want to access this in QML.

cura/API/__init__.py
```diff
@@ -21,3 +24,9 @@ class CuraAPI:
 
     # Interface API
     interface = Interface()
+
+    _account = Account()
```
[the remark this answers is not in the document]
nallath: Okay, so what do you propose? Moving it into the authenticationService won't actually solve that problem. Moving it to the oauth settings won't really decrease the problem, since then you suddenly need to add extra functionality to store the data somewhere.

cura/OAuth2/AuthorizationHelpers.py
```diff
@@ -0,0 +1,124 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+import json
+import random
+from _sha512 import sha512
```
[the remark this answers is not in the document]
nallath: This is something that was copied from the drive plugin. @ChrisTerBeke, was there a specific reason why hashlib.sha512 wasn't used in the first place?

cura/OAuth2/AuthorizationService.py
nallath: Do you mean make it public (and let the account class call the loading of the data or fully move to the account class?)

cura/API/Account.py
nallath: I can add it, but this is code I copied from the plugin. Account does have a isLoggedIn property though. 
So let me know if you need it, it's easy enough to add.

nallath: Uhg. i need to read better. You already made a pull request and that one uses it. Fixing it.

cura/API/Account.py
nallath: Done.

cura/CuraApplication.py
```diff
@@ -942,6 +949,9 @@ def registerObjects(self, engine):
         qmlRegisterSingletonType(ContainerManager, "Cura", 1, 0, "ContainerManager", ContainerManager.getInstance)
         qmlRegisterType(SidebarCustomMenuItemsModel, "Cura", 1, 0, "SidebarCustomMenuItemsModel")
 
+        from cura.API import CuraAPI
+        qmlRegisterSingletonType(CuraAPI, "Cura", 1, 1, "API", self.getCuraAPI)
```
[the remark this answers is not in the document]
nallath: We could, but then we'd need to copy the functionality since we'd also need the account to be accessible in both cases. (eg; we want it to be accessible from GUI and for plugins). 

Is there an advantage to doing this that i'm missing?

cura/API/Account.py
LipuFei: Sorry, I probably put this in the wrong place. I meant something like this for Preferences.

cura/OAuth2/AuthorizationHelpers.py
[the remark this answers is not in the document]
ChrisTerBeke: Nope. Probably some older examples on the www use _sha512

cura/API/Account.py
```diff
@@ -0,0 +1,108 @@
...
+        super().__init__(parent)
+        self._callback_port = 32118
+        self._oauth_root = "https://account.ultimaker.com"
+        self._cloud_api_root = "https://api.ultimaker.com"
+
+        self._oauth_settings = OAuth2Settings(
+            OAUTH_SERVER_URL= self._oauth_root,
+            CALLBACK_PORT=self._callback_port,
+            CALLBACK_URL="http://localhost:{}/callback".format(self._callback_port),
+            CLIENT_ID="um---------------ultimaker_cura",
```
ChrisTerBeke: I now see you have changed the client_id. This is not possible as it's the primary identifier of the OAuth2 server to know which application is requesting access. As it's not a value that users will ever see on screen I propose we keep the old value.

cura/API/__init__.py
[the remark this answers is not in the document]
nallath: I've changed it so that there is now a weak link between preferences and the account.

ChrisTerBeke:
The CI is still failing btw

cura/API/__init__.py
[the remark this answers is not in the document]
nallath: Ah, right. But Cura in itself has a dependency on Qt in the first place. I do agree that we need to try and keep them as seperate as possible (eg; we should have done this for the settings). In this case it would introduce a fair bit of code duplication.

nallath:
Yeah I know, but I can't see why it's failing...

cura/Machines/QualityManager.py
```diff
@@ -458,7 +460,7 @@ def duplicateQualityChanges(self, quality_changes_name: str, quality_model_item)
     #   stack and clear the user settings.
     @pyqtSlot(str)
     def createQualityChanges(self, base_name: str) -> None:
-        machine_manager = Application.getInstance().getMachineManager()
+        machine_manager = CuraApplication.getInstance().getMachineManager()
```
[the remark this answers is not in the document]
LipuFei: Fixed in 7ae6800

cura/OAuth2/AuthorizationHelpers.py
```diff
@@ -24,37 +24,39 @@ def __init__(self, settings: "OAuth2Settings") -> None:
...
-    def getAccessTokenUsingRefreshToken(self, refresh_token: str) -> AuthenticationResponse:
-        return self.parseTokenResponse(requests.post(self._token_url, data={
-            "client_id": self._settings.CLIENT_ID,
-            "redirect_uri": self._settings.CALLBACK_URL,
-            "grant_type": "refresh_token",
-            "refresh_token": refresh_token,
-            "scope": self._settings.CLIENT_SCOPES
-        }))  # type: ignore
+    def getAccessTokenUsingRefreshToken(self, refresh_token: str) -> "AuthenticationResponse":
+        data = self.getCommonRequestDataDict()
```
ChrisTerBeke: I personally find this less readable than before.

ChrisTerBeke:
I made this needed change to the backups plugin for testing. One call in the backups module was broken, I fixed in a commit on this PR.

ChrisTerBeke:
Tested again after last change, seems fully backwards compatible. Okay to merge IMO.

ChrisTerBeke:
When will this be merged? It's quite big and there's already conflicts coming up.

ChrisTerBeke:
Praise the Lord!

fieldOfView:
You rang?

PR #4441, opened 2018-09-28
Resolve circular imports for CuraAPI

LipuFei:
@nallath https://github.com/Ultimaker/Cura/pull/4423 Looks good. I made some changes to resolve circular imports here.

 - Resolve circular imports between oauth, CuraAPI, and CuraApplication which implies circular dependencies.
 - It should always be safe to call `import` at the beginning of the code.
 - `__init__()` should always be safe to call instead of keeping, for example, `self._cura_api` `None` until all its dependencies have been initialized and then create `CuraAPI`.

### DIFF
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,15 +1,18 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
-from typing import Optional, Dict
+from typing import Optional, Dict, TYPE_CHECKING
 
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty
 
+from UM.i18n import i18nCatalog
 from UM.Message import Message
+
 from cura.OAuth2.AuthorizationService import AuthorizationService
 from cura.OAuth2.Models import OAuth2Settings
-from UM.Application import Application
 
-from UM.i18n import i18nCatalog
+if TYPE_CHECKING:
+    from cura.CuraApplication import CuraApplication
+
 i18n_catalog = i18nCatalog("cura")
 
 
@@ -26,8 +29,9 @@ class Account(QObject):
     # Signal emitted when user logged in or out.
     loginStateChanged = pyqtSignal(bool)
 
-    def __init__(self, parent = None) -> None:
+    def __init__(self, application: "CuraApplication", parent = None) -> None:
         super().__init__(parent)
+        self._application = application
 
         self._error_message = None  # type: Optional[Message]
         self._logged_in = False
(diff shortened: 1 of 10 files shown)


cura/Backups/Backup.py
```diff
@@ -29,7 +29,8 @@ class Backup:
     # Re-use translation catalog.
     catalog = i18nCatalog("cura")
 
-    def __init__(self, zip_file: bytes = None, meta_data: Dict[str, str] = None) -> None:
+    def __init__(self, application: "CuraApplication", zip_file: bytes = None, meta_data: Dict[str, str] = None) -> None:
+        self.application = application
```
ChrisTerBeke: This should be private?

ChrisTerBeke:
Good changes IMO.

cura/Backups/Backup.py
LipuFei: I don't really have an opinion here because `zip_file` and `metadata` are already public.

ChrisTerBeke: Those are used by the backups manager, application is (should) not (at least not via Backup).

LipuFei: Done in c04c7654c

LipuFei:
@ChrisTerBeke Still need to discuss with @nallath

ChrisTerBeke:
Does @nallath agree?

ChrisTerBeke:
Jaime indicated that API is not a singleton anymore now. That can quite easily be fixed (see Preferences). Without it being a singleton a plugin developer could make another instance of API and have 2 account objects for instance.

cura/API/Account.py
```diff
@@ -1,15 +1,18 @@
...
+from UM.i18n import i18nCatalog
 from UM.Message import Message
+
 from cura.OAuth2.AuthorizationService import AuthorizationService
 from cura.OAuth2.Models import OAuth2Settings
-from UM.Application import Application
 
-from UM.i18n import i18nCatalog
+if TYPE_CHECKING:
+    from cura.CuraApplication import CuraApplication
```
sedwards2009: This is really a symptom that you are passing around far more than necessary. The `Account` class takes the whole application object when all it needed was some preference and settings data. That is a massive amount of unneeded data to share around. It also makes reasoning about what the `Account` class can and cannot do from the outside really hard. (It could do anything it likes, it has the whole application object to play with!). It is not clear from a higher level what other subsystems the `Account` class talks to.

Passing around a whole application object is not a good thing.

LipuFei: I agree, there's still too many dependencies here. But giving specific objects such as `Preferences` doesn't solve the problem either, because those dependencies are still there. IMO, this is a problem that should be solved in a Cura and/or Uranium refactor.

LipuFei:
For Jaime's concern that plugins will import can create other API instances, this can be fixed by giving the plugin a context of where they belong. For example, now in the plugin initialization takes the `Application` instance that created those plugins, but the fact is no plugin uses it. Instead, they keep doing the same `import Application; Application.getInstance()` everywhere. If they starting using the `app` and/or `api` instances they are given, there is no need to import or create their own copies.

cura/API/Account.py
sedwards2009: My point is that over sharing data and overcoupling is a bad idea. Making it clear what is and isn't shared is desirable, even if you have more parameters. (See 'dependency injection')

Defining a clear application interface is a good also.

nallath:
I didn't pay enough attention, so apologies for that. Lipu's solution is correct.

PR #4445, opened 2018-09-29
Reduce the maximum precision of floats in SettingTextField inputs

### LINKED ISSUE #4368: Enhancement Request; Layer Height Decimal Limit in UI
I would like to bring to the attention of developers that the current value for layer height is restricted to 3 digits past the decimal place in cumulative z-stage settings in the g-code file.  However, I need to bring to the developers attention that the field for custom printer settings allows for a 4th digit after the decimal point that remains as entered.

My request is to only allow the UI to accept 3 digits after the decimal point or issue an alert to the user if they enter a 4th digit after the decimal point.

I did confirm that the last digit is ignored and the apparent error is cumulative.  I have been set straight as to understand this is not an error or a bug.  Simply a UI oversight.  There are probably other fields that this Enhancement Request could/should/or would apply to.

**Application Version**
Ultimaker Cura 3.4.1

**Platform**
Any

**Printer**
Any

**Steps to Reproduce**
Enter a custom layer height of 0.1925 (or any 0.nnnn).
Prepare the model.  
Save g-code to file
Open file with favorite editor
Calculate values from one layer to the next at least 3 times (to rule out rounding).
Compare the average layer height with the value provided by the user (still reading 0.1925)

**Actual Results**
Each layer to layer z-value calculates to 0.192.  This becomes a perceived cumulative error.

**Expected results**
I say perceived above only in my own ignorance that layer heights are in whole microns.
I would prefer the UI issue an alert regarding layer height values with a 4 decimal places -or- the UI excepts and shows only what will be used in g-code generation.

**Additional Information**
I've done this test and I know it is little thing.  Yet, the UI is the face of all the black magic under the hood.  Sometime it is good to know what the brew is made up of -exactly- what we put into it.

Thank you for considering this enhancement request.

fieldOfView:
This PR reduces the maximum precision of floats in the sidebar to 3 digits. No settings in Cura require more than 3 digits of precision. Many settings are in mm, and CuraEngine does its calculations in while microns. Having more precision in the frontend than in the backend apparently thoroughly confuses some people. 

This PR fixes #4368

### DIFF
--- a/resources/qml/Settings/SettingTextField.qml
+++ b/resources/qml/Settings/SettingTextField.qml
@@ -152,7 +152,7 @@ SettingItem
             maximumLength: (definition.type == "str" || definition.type == "[int]") ? -1 : 10;
             clip: true; //Hide any text that exceeds the width of the text box.
 
-            validator: RegExpValidator { regExp: (definition.type == "[int]") ? /^\[?(\s*-?[0-9]{0,9}\s*,)*(\s*-?[0-9]{0,9})\s*\]?$/ : (definition.type == "int") ? /^-?[0-9]{0,10}$/ : (definition.type == "float") ? /^-?[0-9]{0,9}[.,]?[0-9]{0,10}$/ : /^.*$/ } // definition.type property from parent loader used to disallow fractional number entry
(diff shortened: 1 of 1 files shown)


CadDude:
Thank you.

Appesteijn:
CURA-5780

PR #4450, opened 2018-09-30
Add variants to Tizyx K25

ValentinPitre:


resources/variants/tizyx_k25_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.2 mm
+version = 4
+definition = tizyx_k25
+
+[metadata]
+setting_version = 4
```
LipuFei: `setting_version`s should be `5` now.

Appesteijn:
CURA-5779

Vandrasc:
I noticed that the variant selection is left empty when upgrading from a previous version w/o variants. It updates correctly when selecting any variant from the drop-down. 
empty

PR #4455, opened 2018-10-01
Fix toolbox download

diegopradogesto:
This fixes an issue that makes Cura crashing when clicking on the "Back" button while downloading a package. I did it this way to avoid risks since we are close to the 3.5 release.


Appesteijn:
CURA-5778

PR #4459, opened 2018-10-02
3.5

Patola:
Updated strings for Cura 3.5.


diegopradogesto:
Thanks for the translations. We were worried that they won't be in, but you finally did it!!

Issue #4434, opened 2018-09-27
Crash when changing WiFi


BagelOrb:
**Application Version**
master

**Platform**
Linux

**Printer**
UM3

**Steps to Reproduce**
- Be connected to an UM3
- change WiFi to a different WiFi a couple of times

**Actual Results**
Crash

**Expected results**
Any exception is caught and the zeroconf plugin is restarted (or whatever)

**Additional Information**
```
2018-09-27 10:25:22,181 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/resources/qml/Topbar.qml:22: TypeError: Cannot read property 'acceptsCommands' of undefined
2018-09-27 10:25:22,182 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/cura/../plugins/UM3NetworkPrinting/resources/qml/UM3InfoComponents.qml:109: TypeError: Cannot read property 'materialNames' of undefined
2018-09-27 10:25:22,182 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/cura/../plugins/UM3NetworkPrinting/resources/qml/UM3InfoComponents.qml:100: TypeError: Cannot read property 'hotendIds' of undefined
2018-09-27 10:25:22,182 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/cura/../plugins/UM3NetworkPrinting/resources/qml/UM3InfoComponents.qml:16: TypeError: Cannot read property 'acceptsCommands' of undefined
2018-09-27 10:25:22,182 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/cura/../plugins/UM3NetworkPrinting/resources/qml/UM3InfoComponents.qml:17: TypeError: Cannot read property 'authenticationState' of undefined
2018-09-27 10:25:22,182 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [352]: file:///home/t.kuipers/Development/CuraTK/cura/../resources/qml/SidebarHeader.qml:24: TypeError: Cannot read property 'connectedPrintersTypeCount' of undefined
[...]

PR #4463, opened 2018-10-03
CURA-5791 Use "singleton" OutputDevicesModel

LipuFei:
In association with https://github.com/Ultimaker/Uranium/pull/401 to fix https://github.com/Ultimaker/Cura/issues/4434


nallath:
I was able to fix this in a much simpler way by setting the objectOwnership.

PR #4485, opened 2018-10-05
Don't skip empty first layers in the layer view

### LINKED ISSUE #4172: Layer view does not account for floating objects
**Application Version**

4.3.1

**Platform**

Windows 10

**Printer**

[CCR10_Border.curaproject.3mf.zip](https://github.com/Ultimaker/Cura/files/2249090/CCR10_Border.curaproject.3mf.zip)

**Steps to Reproduce**

Load STL
Set Z height
Uncheck "Remove Empty Layers First"
Slice

**Actual Results**

This pic shows what happened.
If you look at the shadow at the back of the print, you can see the top of the domes that was not sliced.
screenshot 37

**Expected results**

Expected all of the STL to be sliced.

**Additional Information**

If you increase the Z to 5 ( because my "border STL" is only 2mm thick ), you can see in this pic, there is no Layer data to display. It appears that the +Z is deducted from the overall project height prior to slicing.
screenshot 38

It would be easier to treat the project as though the "Remove Empty Layers First" is checked.
Then, if the user unchecks "Remove Empty Layers First", run an algorithm on the gcode to add the +Z position, before displaying the "Layer View".
This should make the process easier to do that try to include the +Z into the gcode beforehand.

( I don't intend to step on anyone's toes, but I have a programming background, and this is how I would have approached this option in Cura  )

diegopradogesto:
CURA-5768

Fixes #4172

### DIFF
--- a/plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
+++ b/plugins/CuraEngineBackend/ProcessSlicedLayersJob.py
@@ -2,6 +2,7 @@
 #Cura is released under the terms of the LGPLv3 or higher.
 
 import gc
+import sys
 
 from UM.Job import Job
 from UM.Application import Application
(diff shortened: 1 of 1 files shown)


fieldOfView:
Could you explain what this does, and how it relates to the "remove_empty_first_layers" setting?

diegopradogesto:
This was a reported issue in GH, #4172 . This code intends to fix this problem.

PR #4493, opened 2018-10-05
Remove unused and incorrect "icon" metadata from definitions

fieldOfView:
This PR removes the unused (and incorrect) "icon" metadata from definitions. The "icon" metadata points to non-existing files for all printers (and most 3rd party contributors have copied the bad habbit of just pointing to an Ultimaker2 icon).


fieldOfView:
I'm closing this one-commit PR in favor of a more interesting one which contains this single commit.

PR #4498, opened 2018-10-05
Make the "Add Printer" dialog less spartan

fieldOfView:
This PR applies some TLC to one of the first dialogs a new user of Cura is presented with. The layout is loosely modeled on the layout of the UM3/UMS5 "Connect to Network" dialog, and adds some instructions/intention to the dialog.

image

### DIFF
--- a/resources/definitions/3dator.def.json
+++ b/resources/definitions/3dator.def.json
@@ -7,7 +7,6 @@
         "author": "3Dator GmbH",
         "manufacturer": "3Dator GmbH",
         "file_formats": "text/x-gcode",
-        "icon": "icon_ultimaker2",
         "supports_usb_connection": true,
         "platform": "3dator_platform.stl",
         "machine_extruder_trains":
--- a/resources/definitions/anycubic_i3_mega.def.json
+++ b/resources/definitions/anycubic_i3_mega.def.json
@@ -8,7 +8,6 @@
         "author": "TheTobby",
         "manufacturer": "Anycubic",
         "file_formats": "text/x-gcode",
-        "icon": "icon_ultimaker2",
         "platform": "anycubic_i3_mega_platform.stl",
         "has_materials": false,
         "has_machine_quality": true,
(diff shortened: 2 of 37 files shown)


Appesteijn:
👍

fieldOfView:
So, uhm, JIRA issue?

diegopradogesto:
I'm discussing some PRs right now, and creating tickets.

diegopradogesto:
Here you are! Devs.: CURA-5847

Vandrasc:
Hi @fieldOfView , great improvement, thanks! 

Quick question, is the text correct? In "...and match adjust the settings to match your printer in the next dialog" , does the first "match " belongs there?

fieldOfView:
'doh! Well spotted. If someone could fix that for me quickly, that would be great.

Vandrasc:
Maybe I can, else will call in the big guns :)

diegopradogesto:
Fix it in e884605cc3c1b272b0be501c948e763f66191656

Vandrasc:
Diego found an issue where sometimes the manufacturer and author don't update after switching between categories:

screenshot-1

1. Go to Add Printer and open 'Others' now go to 'Ultimaker'

fieldOfView:
Ah, I have found the problem (but still need to find the fix).

For most printers, only the metadata is loaded, not the whole definition (to speed up loading). However it turns out that for printers that have been fully loaded (ie: printers that already have an instance added to Cura), the structure of the definition in the DefinitionContainersModel is different; the "metadata" is a QVariant(PyQt_PyObject) instead of a normal javascript object.

fieldOfView:
The issue is fixed here: https://github.com/Ultimaker/Uranium/pull/414

Vandrasc:
Thank you, @fieldOfView! 

@diegopradogesto, could we use the same ticket for this? Many thanks

diegopradogesto:
Yes, I think it's a good idea since the internal ticket is still in QA.

fieldOfView:
> he internal ticket is still in QA.

In that case, here's a followup PR that cleans up the code a bit (because it was a bit spotty in places).
https://github.com/Ultimaker/Cura/pull/4709

Issue #4474, opened 2018-10-04
[3.5]machine extruder trains  not specified in Creality ender 3 definition file


ilporcupine:
Seems to default to some Ultimaker profile, as the filament size reverts to 2.85mm. May explain horrible underextrusion problem I had. On inspecting the .json files, the CR10 series all define the extruder, as do the other machines I inspected. Only the Ender seems to be missing an extruder definition.

SaskiFX:
Looks like the extruder is listed in the extruders folder, just didn't get mentioned in the definition for the printer.

stelgenhof:
This is because Cura 3.5 includes an incorrect version of the Ender-3 machine definition. The incorrect version was corrected and committed but apparently didn't make it in the 3.5 release.

You can see in the master branch (https://github.com/Ultimaker/Cura/tree/master/resources) that the correct profile and extruder train definition are present.

BagelOrb:
So this is fixed already for the next release - if I understand correctly.

I can reopen if need be.

stelgenhof:
@BagelOrb Thanks for the quick response!

PR #4501, opened 2018-10-06
Ender-3 machine definition update

stelgenhof:
This update for the Ender-3 machine definition contains a couple enhancements:

- Changed bed size to 235x235mm (including the STL Mesh). 
- Removed all M commands (M104, M140, M190, M109) from the custom start GCode, and removed all M commands (M104, M140, M106, M107) from the custom end GCode as Cura already generates these. 
- Decreased the height position of the extruder (from 3 to 2mm) upon purging. 
- Increased the height position of the extruder (from 10 to 20mm) after the print has finished.

To the release managers: Please make sure these commits altogether are included in the next release. (This version of the machine definition, the already committed extruder definition and this updated STL mesh are a bundle are the correct ones).


stelgenhof:
As said, including all the 3 files (creality_ender3_extruder_0.def, creality_ender3.def.json and creality_ender3_platform.stl) in the next release, will solve many of the issues that are reported (such as https://github.com/Ultimaker/Cura/issues/4474)

rburema:
Devs: CURA-5841.

resources/definitions/creality_ender3.def.json
```diff
@@ -57,7 +57,7 @@
         },
         "layer_height_0": {
             "default_value": 0.2
-        },
+        },
```
fieldOfView: remove trailing spaces
```suggestion
        },
```

fieldOfView:
In my opinion, that platform STL is _way_ to large for what it does.

resources/definitions/creality_ender3.def.json
stelgenhof: Thanks! Must have missed that.

stelgenhof:
@fieldOfView Do you mean in terms of file size too large or the dimensions?

fieldOfView:
File size.

stelgenhof:
@fieldOfView Let me see if I can reduce the size. I thought it was not too big as there are other meshes of similar or larger size.

Do you have any guidelines regarding the size?

fieldOfView:
For me it is the file size / effect ratio. Your stl is mostly just a plate, with an embossed logo. The Ultimaker Original platform mesh (ultimaker_platform.stl) also has a logo in the mesh, and it has much more detail showing parts of the printer, and it is only 171 kb (a factor 9 smaller). 

There are other printers that have similar platform mesh file sizes to yours, which I am also not agreeing with, but you should be able to create a mesh which has the same effect in much fewer polygons.

fieldOfView:
Will you even be seeing the embossed logo? Won't it be covered by the buildplate shader?

stelgenhof:
I quickly made it to see how it would look like, so indeed it can be much improved. Let me check the others and remodel it. Thanks!

rburema:
I'd like to add a friendly reminder that the next release may be sooner than you think, so please keep that in mind while planning when to do the remodel. :-)

fieldOfView:
reminder? That's the first time I hear about it.

rburema:
Hi @fieldOfView (& everyone else bumping into this issue), That comment was mostly to @stelgenhof, for more context, see the discussion in #4531.

stelgenhof:
@rburema What is soon?  A matter of days, or weeks, or... :)

stelgenhof:
@fieldOfView @rburema Reduced the size of the STL (about 3 times smaller) :)
Created a new PR for this: https://github.com/Ultimaker/Cura/pull/4645

Hope it will be included in the next release. If there is anything wrong/missing, just let me know.

Cheers! Sacha

stelgenhof:
@rburema Ok. Understood.

PR #4521, opened 2018-10-08
[3.5] CURA-5799 Fix sending extruder settings to curaengine

LipuFei:


ChrisTerBeke:
Idea to prevent this in the future: make engine string snapshots from this method and store them in the repo. Then if anything changes in a PR, you'll see immediately.

PR #4529, opened 2018-10-09
gMax 1.5+ Config Files

gordo3di:
Updated config files for the gMax 1.5+ single and dual setups for Cura 3.5.

Hopefully I did this correctly. We tested the files locally and they work well.


gordo3di:
We only want to merge "gMax 1.5+ Cura 3.5 Config files" so I'm not sure how to ignore our other commits.

diegopradogesto:
Devs. CURA-5848

diegopradogesto:
> We only want to merge "gMax 1.5+ Cura 3.5 Config files" so I'm not sure how to ignore our other commits.

I don't understand what do you mean with this. If you just want to add this commit: a9364a0b02babdcdf09becab03c9e11d4a2ed902 then create a PR only with this commit please. I'll close this PR for now. If you do it in this in the same PR, then you can reopen it again.

diegopradogesto:
Ok, thanks for the heads up!

PR #4531, opened 2018-10-10
[3.5] Use 0.1 layer height as normal for Ender 3

LipuFei:
hi @stelgenhof, because the default layer height of Ender 3 is 0.15mm which is the same as "Draft" quality, you can see double qualities in the dropdown list:
image

So, we propose to change the default layer height to 0.1mm. For Ender 3, you already have `"preferred_quality_type": "draft"`, so it will be draft by default which is 0.15mm

### DIFF
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -60,7 +60,7 @@
             "default_value": 20
         },
         "layer_height": {
-            "default_value": 0.15
+            "default_value": 0.10
         },
         "layer_height_0": {
             "default_value": 0.2


stelgenhof:
@LipuFei I am not sure what is happening here, but I already corrected this (including the missing extruder definition) in an earlier PR (https://github.com/Ultimaker/Cura/pull/4345). It says the PR was merged into the master branch, but those commits didn't make into the 3.5 release apparently.

Please have a look at that PR first before creating any other new PR's :)

Cheers! Sacha

stelgenhof:
@LipuFei Thank you for the update. So to summarize, for 3.5.1:
- PR #4345 will be included
- PR #4501 will ***not*** be included

Is my understanding correct?

LipuFei:
Yes.

stelgenhof:
@LipuFei Thanks! I updated my previous comment with more details.

drzejkopf:
I also ask to merge #4501 to 3.5.1, because it fixes one of major issues with Ender 3 definition - because of start gcode, temperatures set in Print Setup>Material will not be applied to gcode. I consider that behavior a bug, so that PR should be included in bugfix release :)

LipuFei:
@Ghostkeeper @nallath @Appesteijn forward message

stelgenhof:
@rburema This PR is actually not needed. The `layer_height` value was at first included in the definition but was removed in PR #4345 as it was wrong. PR #4345 was never merged with the 3.5 branch, hence the issues.

Please review the comments in this PR. Again, merging PR's #4345 and #4501 solve all the issues and no other PR's need to be created.

If anything is unclear, just let me know.

Cheers! Sacha

rburema:
Hi @stelgenhof.
This was discussed internally.
I'm really sorry, I know those changes seem small, but the 3.5.1 release will almost exclusively fix major crashes and issues of equal magnitude (as is the norm for a point-release). Arguably this one shouldn't even have been in.
Those other PR's will in all probability be merged for the upcoming 3.6, which we also hope to release this year.
cheers, Remco.

stelgenhof:
Hi @rburema,

Thank you for the update. Understand your decision from your release policy perspective.

Only one other question in that case:

As you may know, the start gcode in the definition (as it is right now) contains a command for setting the extruder temperature: `M104 S{material_print_temperature_layer_0}`. In 3.5, this causes an issue in that the extruder temperature set by the user is not observed. Cura will use its internal default temperature instead.

By only including PR #4345 in the 3.5.1 release, will this issue with the extruder temperature be resolved in 3.5.1 also?

Cheers! Sacha

Ghostkeeper:
> As you may know, the start gcode in the definition (as it is right now) contains a command for setting the extruder temperature: M104 S{material_print_temperature_layer_0}. In 3.5, this causes an issue in that the extruder temperature set by the user is not observed. Cura will use its internal default temperature instead.

Please use something like this for the printing temperature:

    M104 S{material_print_temperature_layer_0, 0} #Heat up to first extruder's printing temperature.

stelgenhof:
@Ghostkeeper Thanks for the tip. Actually this is not relevant anymore as these commands have been removed from the latest definition altogether :)

PR #4554, opened 2018-10-12
Prevent Cura from messing with serial ports of other devices

joba-1:
…by filtering ports with regular expressions passed in as environment variables.

## The problem
Cura scans all USB serial ports for printers. To do this it reads (steals) data from the ports, writes data and changes settings of those ports. 
Due to this you cannot use USB serial ports for other things while using Cura.

## Not a solution here
Cura could check for lock files and leave associated ports alone. Not all other software uses those lock files, so this would not be a universal solution.
On linux you can prevent unwanted access to the devices by authorization. E.g. make the device file owned by a group and make only the services or users that should access the device a member of that group. This does not help here, since the Cura user and the user of the other devices can easily be the same. Also it would need an administrator to set this up.

## Solution
This patch allows an enduser to set environment variables with regular expression patterns describing ports Cura is allowed to use, based on device name, device type or physical port.

## Example usage 
Call this to find a unique pattern for your 3d printer port:
```
python3 -m serial.tools.list_ports -v

/dev/ttyUSB0        
    desc: CP2102 USB to UART Bridge Controller
    hwid: USB VID:PID=10C4:EA60 SER=0001 LOCATION=2-1.1
/dev/ttyUSB1        
    desc: USB2.0-Serial
    hwid: USB VID:PID=1A86:7523 LOCATION=1-1.1
/dev/ttyUSB2        
    desc: Qualcomm Gobi 2000
    hwid: USB VID:PID=05C6:9204 LOCATION=2-1.4
/dev/ttyUSB3        
    desc: CP2102 USB to UART Bridge Controller
    hwid: USB VID:PID=10C4:EA60 SER=0001 LOCATION=1-1.2
4 ports found
```
Then set an environment variable according to your needs, e.g.:

 CURA_DEVICETYPES=CP2102 -> match a desc type of serial converter
 CURA_DEVICEINFOS=LOCATION=2-1.4 -> match a specific physical port
 CURA_DEVICEINFOS=VID:PID=10C4:EA60 -> match a vendor:product id
 CURA_DEVICENAMES=USB[1-9] -> not matching /dev/ttyUSB0. 

The last example is not recommended though, since device names are volatile outside of the users control.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -5,6 +5,8 @@ import threading
 import platform
 import time
 import serial.tools.list_ports
+from os import environ
+from re import search
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtProperty, pyqtSignal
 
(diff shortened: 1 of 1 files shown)


joba-1:
If someone using openSUSE linux wants to test this: I just have created a [wiki page](https://github.com/Ultimaker/Cura/wiki/Running-Cura-from-Source-on-openSUSE) that describes how to do it.

diegopradogesto:
Devs. CURA-5917

Ghostkeeper:
> Not all other software uses those lock files, so this would not be a universal solution.

At least if we can create those files ourselves, we'd be preventing one major case of print failures due to this bug: People starting a second instance of Cura while USB printing with the first instance.

fieldOfView:
Except that's an impossible scenario. If the first Cura is printing, it has opened the serial port. If the serial port is already open, it cannot be opened again.

The "print killing issue" is when the printer is printing off SD card, but connected to the USB port. Then opening the serial port will kill the ongoing print.

joba-1:
> Except that's an impossible scenario. If the first Cura is printing, it has opened the serial port. If the serial port is already open, it cannot be opened again.

I'm afraid this is only true for windows. Linux has no problem opening a file or device multiple times by different processes. That is why I cant use standard Cura: it kills my attached sensors, and if I start a second instance, it kills the communication of the first one. 

Let's face it: serial existed long before plug&play was invented. You have to configure it manually or loose.

Issue #3630, opened 2018-04-06
feature request: initial bottom layer thickness


rooiejoris:
**Feature request**

initial bottom layer thickness, make the bottom thickness 0 so you will start with printing the infill at the build plate.

Many times I don't need a solid bottom, sometimes I want to fill the object with plaster afterwards, or I like the view of the infill at the bottom, especially with the Cubic [or other '3D' infill] patterns.
Now I can set the bottom thickness to 0, but if you have some overhang half way, you will get holes.

**Additional Information**
Somebody helped me to get started, but a little bit more attention was needed... : )
See below the changes for the cura engine and the json files [please note at that time the name of the feature was "build_plate_bottom_layer_count" I think  "initial_bottom_layer_count" is better since you [at least I do] search on "initial" in Cura to control all the initial layer settings.
The code below works, but the infill doesn't start the 'corrected layers'.

dmprinter.def.json:

    +                                },
    +                                "build_plate_bottom_layer_count":
    +                                {
    +                                    "label": "Bottom Layers on Build Plate",
    +                                    "description": "The number of bottom layers directly touching the build plate.",
    +                                    "minimum_value": "0",
    +                                    "minimum_value_warning": "2",
    +                                    "default_value": 6,
    +                                    "type": "int",
    +                                    "value": "999999 if infill_sparse_density == 100 else math.ceil(round(bottom_thickness / resolveOrValue('layer_height'), 4))",
    +                                    "limit_to_extruder": "top_bottom_extruder_nr",
    +                                    "settable_per_mesh": true

CuraEngine:

    diff --git a/src/skin.cpp b/src/skin.cpp
    index 972dd26..5d4fe92 100644
    --- a/src/skin.cpp
    +++ b/src/skin.cpp
    @@ -63,6 +63,7 @@ SkinInfillAreaComputation::SkinInfillAreaComputation(int layer_nr, const SliceDa
     : layer_nr(layer_nr)
     , mesh(mesh)
     , bottom_layer_count(mesh.getSettingAsCount("bottom_layers"))
    +, build_plate_bottom_layer_count(mesh.getSettingAsCount("build_plate_bottom_layer_count"))
     , top_layer_count(mesh.getSettingAsCount("top_layers"))
     , wall_line_count(mesh.getSettingAsCount("wall_line_count"))
     , skin_line_width(getSkinLineWidth(storage, mesh, layer_nr))
[...]

PR #4557, opened 2018-10-12
feat: option to leave the very bottom of the print open

BagelOrb:
frontend setting for https://github.com/Ultimaker/CuraEngine/pull/872


diegopradogesto:
Devs. CURA-5918 (same as the PR in CuraEngine)

fieldOfView:
Does this fix https://github.com/Ultimaker/Cura/issues/3630? @rooiejoris

BagelOrb:
I would think so.

PR #4561, opened 2018-10-12
Feat: flow per print-feature

BagelOrb:
frontend settings for https://github.com/Ultimaker/CuraEngine/pull/874


BagelOrb:
I don't understand. The Jenkins link links to something which doesn't look like an error report.

What's the problem? What do you need me to do?

diegopradogesto:
This is the error:
image

What you need to do is to move the `prime_tower_flow` entry in the file `expert.cfg` from the `[dual]` section to the `[material]` section. Also if the new settings are considered expert or advanced or basic, they should be added to the right file.

BagelOrb:
Ah thank! I see where to find the error now.

Fixed it!
Made all flow settings on by default in expert settings.

(Minor remark: it's a bit weird that you have to specify the category in the default setting visibility file. I don;t really see the reason.)

cjp0008:
Any update on this?

BagelOrb:
+1

Ghostkeeper:
The Jira ticket says that it would be discussed with Ultimaker's Materials & Processing team. A few months ago I asked if this had happened yet but nobody seems to read that (and then I forgot about it too).

I don't know why the M&P team needs to get involved with this, really, other than by telling them that this is now an option, and they can use it if they like, or not if they don't like...

cjp0008:
> 
> 
> The Jira ticket says that it would be discussed with Ultimaker's Materials & Processing team. A few months ago I asked if this had happened yet but nobody seems to read that (and then I forgot about it too).
> 
> I don't know why the M&P team needs to get involved with this, really, other than by telling them that this is now an option, and they can use it if they like, or not if they don't like...

Ok, thanks for the info, yes it doesn't make much sense.... hopefully they will just add the functionality and then let the M&P team play around with it!

PR #4596, opened 2018-10-17
Feature: "Support Interface Offset"

vgribinchuk:
Feature: "Support Interface Offset"
This feature allows to apply offset for support interface polygons.
New Settings:
"Support Interface Offset" ("Support Roof Offset", "Support Floor Offset")
Default values for these settings are 0mm. 
Minimum allowable value is 0mm.
This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/879

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4336,6 +4336,50 @@
                         }
                     }
                 },
+                "support_interface_offset": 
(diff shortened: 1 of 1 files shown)


nallath:
For our reference; CURA-5880

nallath:
For our reference; CURA-5880

PR #4598, opened 2018-10-17
Feature: "Support Infill/Interface Line Directions"

vgribinchuk:
Feature: "Support Infill/Interface Line Directions"
CuraEngine allows to select infill that prints in user selected angles. This feature works the same as "Infill Line Directions", but it’s applied to the support / interface layers. 
Current setting "Support Infill Line Direction" was modified to be a list of integer line directions to use.
New Settings:
"Support Infill Line Directions"(modified version of the previous "Support Infill Line Direction");
"Support Interface Line Directions"("Support Roof Line Directions", "Support Floor Line Directions").
This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/881


nallath:
For our reference; CURA-5882

moyesboy:
When I use the feature that changes the support line distance as the support grows (**gradual support infill steps**), If f use zig zag or lines at the change of density it drops lines half way in between the existing lines so it is not successful.
Can the support line direction be changed 90 degrees when the support density doubles? This would make this feature much more useful as it only really works with grid type support structures currently.

forgive me if this is in the wrong place I'm new to github.

alekseisasin:
What is the case of using the support interface direction? Can you show an example?

vgribinchuk:
Support interface line direction provide more precise control of interface printing. Please, take a look at the pic below, where yellow tool path is interface lines.
interfacelinedirections

alekseisasin:
If you proposed to change one of the settings: "support_infill_angle" to "support_infill_angles". For his, the "version upgrade" needs to be adjusted. 
Needs to be adjusted: type, default value and removed old types(minimum, maximum value). It happens when Cura load old Cura version projects. 

The example of how to do it you can find in VersionUpgrade plugin. At this moment there is no new package. It needs to be created "VersionUpgrade40to41"

resources/definitions/fdmprinter.def.json
```diff
@@ -4336,6 +4333,39 @@
                         }
                     }
                 },
+                "support_interface_angles":
+                {
+                    "label": "Support Interface Line Directions",
+                    "description": "A list of integer line directions to use. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the default angles(alternates between 45 and 135 degrees if interfaces are quite thick or 90 degrees).",
```
Ghostkeeper: Typo here: Needs a space before the bracket. I'll make this change as it's merged.

resources/definitions/fdmprinter.def.json
```diff
@@ -4336,6 +4333,39 @@
...
+                    "default_value": "[ ]",
+                    "enabled": "support_interface_enable and support_interface_pattern != 'concentric'",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true,
+                    "children":
+                    {
+                        "support_roof_angles":
+                        {
+                            "label": "Support Roof Line Directions",
+                            "description": "A list of integer line directions to use. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the default angles(alternates between 45 and 135 degrees if interfaces are quite thick or 90 degrees).",
```
Ghostkeeper: Same typo.

resources/definitions/fdmprinter.def.json
```diff
@@ -4336,6 +4333,39 @@
...
+                            "type": "[int]",
+                            "default_value": "[ ]",
+                            "enabled": "support_roof_enable and support_roof_pattern != 'concentric'",
+                            "settable_per_mesh": false,
+                            "settable_per_extruder": true
+                        },
+                        "support_bottom_angles":
+                        {
+                            "label": "Support Floor Line Directions",
+                            "description": "A list of integer line directions to use. Elements from the list are used sequentially as the layers progress and when the end of the list is reached, it starts at the beginning again. The list items are separated by commas and the whole list is contained in square brackets. Default is an empty list which means use the default angles(alternates between 45 and 135 degrees if interfaces are quite thick or 90 degrees).",
```
Ghostkeeper: Same typo.

resources/definitions/fdmprinter.def.json
vgribinchuk: Added a space before the bracket

resources/definitions/fdmprinter.def.json
vgribinchuk: Fixed

resources/definitions/fdmprinter.def.json
vgribinchuk: Fixed

vgribinchuk:
@Ghostkeeper @alekseisasin  Hello, gentlemen! Sorry for the delay on this! Unfortunately, I have no experience to create/modify mentioned VersionUpgrade plugins, so could you please clarify some points, when your time permits? I'm afraid to make changes without understanding first how this should be done. And, please, forgive me if I didn't get something right. 

- I see there is an upgrade plugin[ 42to43 ](https://github.com/Ultimaker/Cura/blob/master/plugins/VersionUpgrade/VersionUpgrade42to43/VersionUpgrade42to43.py). Should I make changes to this existing plugin or create new? 
- This PR modifies existing setting, so, if I understood correctly, I should, basically, put the  `"support_infill_angle": "support_infill_angles"` into `_renamed_settings` map and then rename settings inside the `upgradeInstanceContainer()` like this:
```
if "values" in parser:
            for old_name, new_name in _renamed_settings.items():
                if old_name in parser["values"]:
                    parser["values"][new_name] = parser["values"][old_name]
                    del parser["values"][old_name]
```
- And for adjusting the type, default value, maximum/minimum values should do something following:
```
parser["values"]["support_infill_angles"]["type"] = "[int]"
parser["values"]["support_infill_angles"]["default_value"] = "[ ]"
del parser["values"]["support_infill_angles"]["minimum_value"]
del parser["values"]["support_infill_angles"]["maximum_value"]
```
If I mistake, let me please know how this should be done correctly.

nallath:
@vgribinchuk 
You should make those changes to the 42to43 versionupgrade

But yeah, as far as I can see, those changes should do the trick.

vgribinchuk:
VersionUpgrade was updated. Could you please review?

plugins/VersionUpgrade/VersionUpgrade42to43/VersionUpgrade42to43.py
```diff
@@ -62,10 +66,19 @@ def upgradeInstanceContainer(self, serialized: str, filename: str) -> Tuple[List
...
         if "values" in parser:
+            for old_name, new_name in _renamed_settings.items():
+                if old_name in parser["values"]:
+                    parser["values"][new_name] = parser["values"][old_name]
+                    del parser["values"][old_name]
             for key in _removed_settings:
                 if key in parser["values"]:
                     del parser["values"][key]
 
+        parser["values"]["support_infill_angles"]["type"] = "[int]"
```
LipuFei: This part is not entirely correct, parser["values"]["some_setting"] only stores the value of a setting. It doesn't store anything such as `type`, `default_value`, etc. So, here, the value needs to be updated to a list of integers.

vgribinchuk: So instead should I write just:
parser["values"]["support_infill_angles"] = "[ ]"
?

LipuFei: You can change this part to:
```
if "values" in parser:
    # Change the value from "float" to "[int]"
    if "support_infill_angles" in parser["values"]:
        old_value = float(parser["values"]["support_infill_angles"])
        new_value = int(round(old_value))
        parser["values"]["support_infill_angles"] = new_value
```

LipuFei: @vgribinchuk Don't worry. I'll fix this later.

LipuFei: Fixed in https://github.com/Ultimaker/Cura/pull/4598/commits/3a10345c0169a1424f02fbaa6d458f6d7a1415da

vgribinchuk: OK, thank you!

PR #4599, opened 2018-10-17
Feature: "Nozzle Switch Extra Prime Distance"

vgribinchuk:
Feature: "Nozzle Switch Extra Prime Distance"
This feature allows to set the extra material to prime after nozzle switching.
New Setting:
"Nozzle Switch Extra Prime Distance"
This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/882

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2466,6 +2466,19 @@
                             "settable_per_extruder": true
                         }
                     }
+                },
+                "switch_extruder_extra_prime_distance": 
+                {
+                    "label": "Nozzle Switch Extra Prime Distance",
+                    "description": "Extra material to prime after nozzle switching.",
+                    "type": "float",
+                    "unit": "mm",
+                    "default_value": 0,
+                    "minimum_value_warning": "0",
+                    "maximum_value_warning": "100",
+                    "enabled": "retraction_enable",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
                 }
             }
         },
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2467,12 +2467,12 @@
                         }
                     }
                 },
-                "switch_extruder_extra_prime_distance": 
+                "switch_extruder_extra_prime_amount": 
                 {
-                    "label": "Nozzle Switch Extra Prime Distance",
+                    "label": "Nozzle Switch Extra Prime Amount",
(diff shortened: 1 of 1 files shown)


nallath:
CURA-5883

alekseisasin:
Can you explain why do you need extra prime material, with example:  "Extra material to prime after nozzle switching".
Does it mean, if you use prime tower it is not enough extruded material?

vgribinchuk:
> Can you explain why do you need extra prime material, with example: "Extra material to prime after nozzle switching".
> Does it mean, if you use prime tower it is not enough extruded material?

The nozzle switch extra prime distance was created in order to compensate for the "ooze" that can happen in between printing the purge tower and the part itself. The ability to restart at a extrusion distance that is different than where the last extrusion ended can allow for a user to tune the extra restart distance to that full extrusion width is achieved as soon as the part is resumed. This may not be helpful for all materials, but is helpful for materials that ooze a lot during retraction and long moves.

alekseisasin:
If you want to compensate the "ooze" then I would expect an extra retraction amount.
For your solution, you probably would need to add extra Gcode at the beginning. Check Machine Settings "Start G-code", if you want to add the extra retraction for the first layer.
The second option to make a Post-processing plugin, which can perform extra manipulation with Gcode.
Did you try the setting "Retraction Extra Prime Amount"?
What type of material do you want to use, is it plastic?

vgribinchuk:
We have used extra prime amount, which is essentially what we have replicated, only with a separate value for tool changes. This was done because in our machine we have a long travel from the extruder resting position to where the part is. This problem is likely to be experienced by all people using an independent dual extruder type setup.

resources/definitions/fdmprinter.def.json
```diff
@@ -2466,6 +2466,19 @@
                             "settable_per_extruder": true
                         }
                     }
+                },
+                "switch_extruder_extra_prime_distance":
```
LipuFei: Hi @vgribinchuk, we'd suggest to use the name `switch_extruder_extra_prime_amount` and with unit `mm3` like `wipe_retraction_extra_prime_amount`. This makes those settings more consistent. https://github.com/Ultimaker/CuraEngine/pull/882 needs to be adjusted for this.

vgribinchuk: Hi @LipuFei ! Thank you, I will have a look.

vgribinchuk: @LipuFei , requested changes were submitted. Could you please review these pull requests again?

PR #4600, opened 2018-10-17
Update OAuth2 scopes

ChrisTerBeke:
Cloud team has updated the OAuth2 scopes format to be more uniform, while keeping compatibility with the old scopes) It is still advised though that all clients switch their requested scopes to the new format. This is currently deployed on staging as a test and will be deployed to production early next week, so this PR should be merged after that. Part of STAR-273.


ChrisTerBeke:
This PR can now be merged.

PR #4609, opened 2018-10-18
Cura 5808 machine action manager

alekseisasin:


Ghostkeeper:
For full transparency (since it's not mentioned in the title here), this Pull Request is meant to fix an issue where the network connectivity action did not pop up when adding a new Ultimaker 3 printer.

cura/MachineActionManager.py
```diff
@@ -38,11 +38,10 @@ def initialize(self):
         # Ensure that all containers that were registered before creation of this registry are also handled.
         # This should not have any effect, but it makes it safer if we ever refactor the order of things.
         for container in container_registry.findDefinitionContainers():
-            self._onContainerAdded(container)
+            self.addActionsForContainer(container)
 
-        container_registry.containerAdded.connect(self._onContainerAdded)
```
Ghostkeeper: Huh, I would expect this to... No longer register the machine actions upon adding a new printer during the runtime of Cura. I guess a test will point this out when I'm running the code...

tests/TestMachineAction.py
```diff
@@ -63,14 +63,15 @@ def test_addMachineAction(machine_action_manager):
 
     # Check if adding multiple (the same) actions to first start actions work.
     machine_action_manager.addFirstStartAction(test_machine, "test_action")
-    machine_action_manager.addFirstStartAction(test_machine, "test_action")
-    assert machine_action_manager.getFirstStartActions(test_machine) == [test_action, test_action]
+    machine_action_manager.addFirstStartAction(test_machine, "test_action_2")
+
+    assert machine_action_manager.getFirstStartActions(test_machine) == [test_action, test_action_2]
```
Ghostkeeper: These tests do not test any more what the documentation says they would test. Please test:
* Adding multiple of the same actions.
* Inserting an action in any other location than the last location.

cura/MachineActionManager.py
```diff
@@ -38,11 +38,10 @@ def initialize(self):
         # Ensure that all containers that were registered before creation of this registry are also handled.
         # This should not have any effect, but it makes it safer if we ever refactor the order of things.
         for container in container_registry.findDefinitionContainers():
```
Ghostkeeper: It's not your change, but could you change this into:
```suggestion
        for container in container_registry.findDefinitionContainersMetadata():
```

That could save a bit of boot time.

cura/MachineActionManager.py
Ghostkeeper: I figured out why it still works. In `MachineManager.addMachine`, it directly calls this function upon adding a new printer. I think that you removed the wrong line then of the two.

The solution with the `containerAdded` signal is better because it doesn't add any new dependencies. The solution with `MachineManager.addMachine` calling `addActionsForContainer` adds a dependency of `MachineManager` to `MachineActionManager` which is illogical and circular.

PR #4613, opened 2018-10-19
CURA-5808 Fix machine actions

LipuFei:
- Fix some typing issues
 - Make sure that a machine's default machine actions are added before it gets activated.


LipuFei:
Apparently I used the wrong ticket number...

LipuFei:
Apparently I used the wrong ticket number...

LipuFei:
@diegopradogesto Unit test added in 89cb670

PR #4614, opened 2018-10-19
CURA-5839 Add CC 0.6 for UM3 and S5

LipuFei:


fieldOfView:
Aww :-(
But I wanted it...

PR #4618, opened 2018-10-19
CURA-5834 Fix material model update

LipuFei:


plugins/MachineSettingsAction/MachineSettingsAction.qml
```diff
@@ -616,7 +621,14 @@ Cura.MachineAction
...
                             if (propertyProvider && text != propertyProvider.properties.value)
                             {
-                                propertyProvider.setPropertyValue("value", text);
+                                if (_setValueFunction !== undefined)
+                                {
+                                    _setValueFunction(text);
+                                }
+                                else
+                                {
+                                    propertyProvider.setPropertyValue("value", text);
```
Ghostkeeper: I don't really understand why this is necessary. Could you document why it's necessary to change the value of the materialDiameterField via this _setValueFunction? Shouldn't we just be listening to the value change on the other end? Update the field on the activeStack that way?

LipuFei: Documented in ea9aa2c77

plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -926,7 +926,7 @@ def _applyMaterials(self, global_stack, extruder_stack_dict):
             build_plate_id = global_stack.variant.getId()
 
             # get material diameter of this extruder
-            machine_material_diameter = extruder_stack.materialDiameter
+            machine_material_diameter = extruder_stack.comptabileMaterialDiameter
```
diegopradogesto: Use the getter function.
```suggestion
            machine_material_diameter = extruder_stack.getComptabileMaterialDiameter()
```

ChrisTerBeke:
It now takes much longer than before when I switch the selected extruder.

PR #4642, opened 2018-10-23
Feature: "Raft Top Line Directions"

vgribinchuk:
Feature: "Raft Top Line Directions"
CuraEngine allows to select infill that prints in user selected angles. This feature works the same as "Infill Line Directions", but it’s applied to the raft top layers. 
New Setting:
"Raft Top Line Directions"
This PR is frontend setting for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/886


nallath:
CURA-5891

alekseisasin:
Theoretically, this feature might add value and improve quality for certain materials, but this property is very specific and will be not used by other users. We should find a solution/proposal/workaround which could improve quality for a bigger auditorium. 

Adding a new setting is always maintenance. This is the reason we try to apply settings which can be used by a wide range of users.

PR #4645, opened 2018-10-24
Update Ender-3 STL Mesh Bed

stelgenhof:
- Reduced the number of polygons of the Ender-3 logo in the mesh bed in order to reduce file size of the STL. 
- Also made the Ender logo punched through so it is a bit more noticeable on the build shader. (Note: It will still be faint and is intentional: to avoid the the STL mesh background distracts too much from the loaded 3D models on the build plate).


nallath:
CURA-5890

PR #4649, opened 2018-10-25
TPU temperatures update

THeijmans:
Removal of all final and initial print temperature offsets, and huge increase in first layer print temperature. These changes should greatly improve first layer extrusion.


diegopradogesto:
Devs. CURA-5858

PR #4650, opened 2018-10-25
[3.6] [CURA-5857] Fix setting extruder material diameter in Machine Settings dialog

LipuFei:
It should update the value of the selected Extruder TAB, not the active extruder on the printer.


rburema:
To all: In the future, we probably shouldn't rely on the `settingsTabs.currentIndex` and the like to get the extruder-number, but the redesign that this'll involve is out of scope for what should be a simple bug-fix, so this'll pass for now.

PR #4652, opened 2018-10-25
CL-1119 Use shorter machine name for skeleton and longer for text (Fix for 3.6)

ianpaschal:
Contributes to CL-1119

Fixes this issue:

image

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/PrinterCard.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/PrinterCard.qml
@@ -118,7 +118,10 @@ Item {
                     Item {
                         id: machineNameLabel;
                         height: UM.Theme.getSize("monitor_text_line").height;
-                        width: Math.round(parent.width * 0.3);
+                        width: {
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/resources/qml/PrinterCard.qml
```diff
@@ -118,7 +118,10 @@ Item {
                     Item {
                         id: machineNameLabel;
                         height: UM.Theme.getSize("monitor_text_line").height;
-                        width: Math.round(parent.width * 0.3);
+                        width: {
```
ChrisTerBeke: Couldn't this just be `width: Math.round(parent.width * (printer ? 0.75 : 0.3)) // width depends on actual printer vs skeleton`?

ianpaschal: Yes, although I decided at the last second it was a little bit hard to read at a glance and figured "why not expand it slightly".

I'll change it if you think single line is better.

PR #4662, opened 2018-10-26
[3.6] [CURA-5767] Disable per-meshgroup settings in the per-mesh settings when in one-at-a-time mode.

rburema:
The backend can't handle per mesh-group settings (as opposed to per mesh settings)
when sequence was in one-at-a-time mode. The problem is that the extruder train is
in between the per-meshgroup-settings (child-of) and per-mesh-settings (parent-of).
For any per-mesh-group settings, the possibility exists that they'll be overridden
by settings we always set (and that need to be set) per extruder. Changing this
would require the engine-architecture to change, and _all_ settings would need to
be thouroughly retested. As this was a too extensive change, it was decied just to
disable the per-meshgroup settings when printing one-at-a-time (too).

The issue was originally reported as:
'"Printing Temperature Initial layer" setting per mode does not work'


cura/Settings/PerObjectContainerStack.py
```diff
@@ -57,7 +58,22 @@ def getProperty(self, key: str, property_name: str, context: Optional[PropertyEv
...
+        # that correctly. This check is also here is because if an old project file includes per-object settings that
+        # are settable-per-meshgroup, when it is loaded, we need to filter out those settings.
+        if not global_stack.getProperty(key, "settable_per_mesh"):
+            Logger.log("i", "Setting [%s] ignored as a per-object setting because it's not settable-per-mesh.", key)
+            return
+        super().setProperty(self, key, property_name, property_value,
+                            container = container, set_from_cache = set_from_cache)
+
+    @override(CuraContainerStack)
+    def setNextStack(self, stack: "CuraContainerStack") -> None:
```
Ghostkeeper: I don't think this new code is desirable, for three reasons:
* We should try to let the project files behave the same way as in previous releases as much as possible. The fact that the per-meshgroup settings didn't always have an effect is a reason to not allow people to set it again to prevent confusion for when it doesn't have an effect, but removing them from the stack causes it to actually change the project files even if it did have an effect.
* This new code adds complexity because it adds to the setProperty function, which should be as transparent as possible.
* It should've been implemented in a version upgrade. If it's implemented here, this code will live as legacy forever and affect us with every setting property that we set on these stacks.

Could you revert it? In the Cura & Cookies meeting we discussed that we would just prevent people from setting new meshgroup settings in the per-mesh settings without needing a version upgrade.

LipuFei: Done in 39158a643

PR #4664, opened 2018-10-26
Add JGAurora Z-603S

pinchies:
Add a preset for the JGAurora Z-603S to model list


nallath:
For our reference; CURA-5902

PR #4666, opened 2018-10-26
Add JGAurora A5

pinchies:
Add a preset for the JGAurora A5, with STL model as well.


nallath:
For our reference; CURA-5902

pinchies:
This profile should also be labelled for the JGAurora A5S too. Both printers have the same mechanical configuration.

PR #4667, opened 2018-10-26
Add JGAurora A1

pinchies:
Add a preset for the JGAurora A1 to model list


nallath:
For our reference; CURA-5902

PR #4670, opened 2018-10-26
Add Cocoon Create ModelMaker

pinchies:
Add a preset for the Cocoon Create ModelMaker (Mini 3D printer sold by Aldi in Australian stores) to model list


nallath:
For our refrence; CURA-5902

iTrooper5118:
Hey pinchies,

There's also the Cocoon Create Touch that ALDI Australia sells as well, and should also be added to the list. That one is a rebadged Wanhao i3 Plus.

PR #4678, opened 2018-10-28
Fix type error in NetworkedPrinterOutputDevice.post

fieldOfView:
This PR fixes a type error in NetworkedPrinterOutputDevice.post: NetworkedPrinterOutputDevice.post takes data as a Python string, and passes that to QNetworkManager.post(), which does not handle strings.

Strictly speaking the QNetworkManager.post() method takes a QByteArray instead of Python bytes, but according to the PyQt documentation, PyQt handles that conversion transparently:
http://pyqt.sourceforge.net/Docs/PyQt5/gotchas.html#python-strings-qt-strings-and-unicode

The type mismatch only came up recently when I added typing to OctoPrintPlugin:
https://github.com/fieldOfView/OctoPrintPlugin/issues/88#issuecomment-433621940
Only that plugin and the LegacyUM3OutputDevice use the NetworkedPrinterOutputDevice.post method. It would cause a crash when using the manual printer controls. I don't think LegacyUM3OutputDevice gets tested much, which is why this error was not spotted before.

From a product hygiene standpoint, I would argue that this needs to be fixed in Cura 3.6, though I can work around it for OctoPrintPlugin.

### DIFF
--- a/cura/PrinterOutput/NetworkedPrinterOutputDevice.py
+++ b/cura/PrinterOutput/NetworkedPrinterOutputDevice.py
@@ -213,7 +213,7 @@ class NetworkedPrinterOutputDevice(PrinterOutputDevice):
         request = self._createEmptyRequest(target)
         self._last_request_time = time()
         if self._manager is not None:
-            reply = self._manager.post(request, data)
+            reply = self._manager.post(request, data.encode())
             if on_progress is not None:
                 reply.uploadProgress.connect(on_progress)
             self._registerOnFinishedCallback(reply, on_finished)
--- a/plugins/UM3NetworkPrinting/src/LegacyUM3OutputDevice.py
+++ b/plugins/UM3NetworkPrinting/src/LegacyUM3OutputDevice.py
@@ -499,8 +499,8 @@ class LegacyUM3OutputDevice(NetworkedPrinterOutputDevice):
         self._authentication_id = None
 
         self.post("auth/request",
-                  json.dumps({"application":  "Cura-" + CuraApplication.getInstance().getVersion(),
-                                               "user": self._getUserName()}).encode(),
+                  json.dumps({"application": "Cura-" + CuraApplication.getInstance().getVersion(),
+                              "user": self._getUserName()}),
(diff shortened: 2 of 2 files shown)


nallath:
We should also put this in 3.6

Appesteijn:
CURA-5904

fieldOfView:
The PR so nice, it was approved twice

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -213,7 +213,7 @@ def post(self, target: str, data: str, on_finished: Optional[Callable[[QNetworkR
         request = self._createEmptyRequest(target)
         self._last_request_time = time()
         if self._manager is not None:
-            reply = self._manager.post(request, data)
+            reply = self._manager.post(request, data.encode())
```
alekseisasin: Did you try to call this function?
I have tried to call data.encode() and it returns "AttributeError: 'bytes' object has no attribute 'encode'"
Also while testing I discovered Cura does not support UM3 old firmware (I tested 3.7...).

fieldOfView: On https://github.com/Ultimaker/Cura/pull/4678/files/517428a308fcff91c68c54a226ad95ec63ac81fd#diff-0bb654171ac2bf659b787a51c1f10823L211 data is defined as a string. If the legacy UM3 output device supplies a bytes array, it is doing it wrong.

This PR can be implemented in a different way too, and that is to redefine the attribute types of the post method. You tell me, what should I do?

fieldOfView: Note: I don't have a UM3 (with or without legacy firmware), so I cannot test with that. I saw that the typing and the implementation do not match, so I fixed that.

LipuFei: Made a PR to fix this: https://github.com/fieldOfView/Cura/pull/4

LipuFei:
@fieldOfView This fix is also merged into 3.6

PR #4679, opened 2018-10-28
Use preferred material instead of generic PLA when creating a new material

fieldOfView:
This PR changes how the "Create" button on the Materials pane of the Preferences creates a new material. Since Cura 2.3, a new material is created by duplicating the Generic PLA material. This is done because *most of the time* creating a duplicate of PLA creates a material that is sure to be available for the current printer. However, for some printers, (generic) PLA may not be available or applicable (eg a printer that is specifically not compatible with the generic materials and only shows the materials that are offered/tested by a specific vendor).

Generic PLA is still used for printers that do not specify a preferred material.


nallath:
CURA-5889

PR #4690, opened 2018-10-30
Rename user visible entries of ToolBox to Marketplace

Appesteijn:
CURA-5874

### DIFF
--- a/resources/qml/Cura.qml
+++ b/resources/qml/Cura.qml
@@ -283,7 +283,7 @@ UM.MainWindow
             Menu
             {
                 id: plugin_menu
-                title: catalog.i18nc("@title:menu menubar:toplevel", "&Toolbox")
+                title: catalog.i18nc("@title:menu menubar:toplevel", "Marke&tplace")
 
                 MenuItem { action: Cura.Actions.browsePackages }
             }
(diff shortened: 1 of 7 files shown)


fieldOfView:
😱 Can there seriously not be more than two releases where the name for this doesn't change??

resources/qml/Cura.qml
```diff
@@ -283,7 +283,7 @@ UM.MainWindow
             Menu
             {
                 id: plugin_menu
-                title: catalog.i18nc("@title:menu menubar:toplevel", "&Toolbox")
+                title: catalog.i18nc("@title:menu menubar:toplevel", "Marke&tplace")
```
ChrisTerBeke: Typo?

alekseisasin: &t - it is a keyboard shortcut

ChrisTerBeke: Mm it's kinda weird to have a keyboard shortcut that is not the first letter of the word (like we have in all other places).

Appesteijn: I wanted to keep the shortcut the same so it wouldn't conflict with other shortcuts

fieldOfView: This type of shortcut is only used while the menu is open, and since there is only a single menu item in the menu, there is no conflict possible.

PR #4704, opened 2018-11-01
CURA-5842 Set CuraApplication display name

LipuFei:
Needs https://github.com/Ultimaker/Uranium/pull/413

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -161,6 +161,7 @@ class CuraApplication(QtApplication):
 
     def __init__(self, *args, **kwargs):
         super().__init__(name = "cura",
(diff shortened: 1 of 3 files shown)


cura/CuraApplication.py
```diff
@@ -161,6 +161,7 @@ class ResourceTypes:
 
     def __init__(self, *args, **kwargs):
         super().__init__(name = "cura",
+                         app_display_name = "Ultimaker Cura",
```
fieldOfView: Could this be retreived from CuraVersion.py(.in)?

LipuFei: Good point. I'll change that.

PR #4705, opened 2018-11-01
Fixes required for semantic versioning

nallath:
CURA-5840

See also; https://github.com/Ultimaker/Uranium/pull/416


ChrisTerBeke:
* Note that once we make the API changes, Cura can submit the full SemVer number for more specific filtering in the compatible plugins.
* The value in CuraVersion.py also needs to be SemVer, most devs will have an integer number in their file still I think?

alekseisasin:
CuraVersion.py, probably yes, do we want to update it in this ticket?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -209,11 +209,11 @@ def _getCloudAPIVersion(self) -> int:
     # Get the packages version depending on Cura version settings.
     def _getSDKVersion(self) -> Union[int, str]:
```
ChrisTerBeke: Note: Cura can send the full number here once the cloud changes are deployed, not just the getMajor() part. It'll work with both, but the API won't return compatible packages from earlier minor versions.

rburema: Look at the last line of this function (r.217 as of writing): It returns the 'full' SDK-version (whichever type that is).

(for the other lines): The 'getMajor()-part' is only called in case that can't be found (for whatever reason), on the API version, which gives us the thing we can rely on if the SDK version can't be found, which is the major version of the API.

ChrisTerBeke: It doesn't have to use the getMajor in any case anymore because the API supports full semver now.

rburema: I (briefly!) checked my thoughts about this with Ruben when I did the code review. I can ask Lipu (who changed those lines), when he gets back.

What I got from it is that during normal operation, the function _does_ return a full semver (see the last line). The other 'outs' only happen if someone deletes CuraDSKVersion (both the setting _and_ the default value thereof).

ChrisTerBeke: That's true, but I don't see why it wouldn't send the full number when not having CuraVersion.py available or correctly configured, it might lead to confusion as the result (package list) is suddenly different.

ChrisTerBeke: Great! The API will only take integers or SemVer, so you cannot just leave out the last number as that's not to spec.

PR #4715, opened 2018-11-02
Redesign of the Action Panel

diegopradogesto:
CURA-5786.


resources/qml/ActionPanel/OutputDevicesActionButton.qml
```diff
@@ -0,0 +1,103 @@
...
+
+            closePolicy: Popup.CloseOnEscape | Popup.CloseOnPressOutsideParent
+
+            contentItem: Column
+            {
+                Repeater
+                {
+                    model: devicesModel
+
+                    delegate: Cura.ActionButton
```
nallath: Why is this an action button? You could just use the (much simpler) button, right? Especially if you're just setting everyting to transparent  & not using a radius.

resources/qml/ActionPanel/OutputProcessWidget.qml
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2018 Ultimaker B.V.
+// Cura is released under the terms of the LGPLv3 or higher.
```
nallath: As before; What is this thing supposed to do?

resources/qml/ActionPanel/OutputProcessWidget.qml
```diff
@@ -0,0 +1,121 @@
...
+                right: printInformationPanel.left
+                rightMargin: UM.Theme.getSize("thin_margin").height
+            }
+
+            Cura.IconLabel
+            {
+                id: estimatedTime
+                width: parent.width
+
+                property var printDuration: PrintInformation.currentPrintTime
```
nallath: Why use a property here? You only use it once.

resources/qml/ActionPanel/OutputProcessWidget.qml
```diff
@@ -0,0 +1,121 @@
...
+
+            Cura.IconLabel
+            {
+                id: estimatedCosts
+                width: parent.width
+
+                property var printMaterialLengths: PrintInformation.materialLengths
+                property var printMaterialWeights: PrintInformation.materialWeights
+
+                function getText()
```
nallath: This needs a better name.

nallath: Also; since you're only using it once, you could just put the content in of the function in the text property;
text: { var totalLengths = 0 , etc }

resources/qml/ActionPanel/PrintJobInformation.qml
```diff
@@ -0,0 +1,166 @@
...
+                    if(!printTime[feature].isTotalDurationZero)
+                    {
+                        text += "<tr><td>" + feature + ":</td>" +
+                            "<td align=\"right\" valign=\"bottom\">&nbsp;&nbsp;%1</td>".arg(printTime[feature].getDisplayString(UM.DurationFormat.ISO8601).slice(0,-3)) +
+                            "<td align=\"right\" valign=\"bottom\">&nbsp;&nbsp;%1%</td>".arg(Math.round(100 * parseInt(printTime[feature].getDisplayString(UM.DurationFormat.Seconds)) / totalSeconds)) +
+                            "</tr>"
+                    }
+                }
+                text += "</table>"
+                print(text)
```
nallath: Unneeded print

resources/qml/ActionPanel/PrintJobInformation.qml
```diff
@@ -0,0 +1,166 @@
...
+                return text
+            }
+
+            text: getTimeSpecifications()
```
nallath: As before; the function is only used once, so it can just be removed.

nallath: Same as before; No need to put it in it's own function

resources/qml/Cura.qml
```diff
@@ -312,17 +312,16 @@ UM.MainWindow
...
                 anchors.bottomMargin: UM.Theme.getSize("thick_margin").height
-                onShowTooltip:
-                {
-                    base.showTooltip(item, location, text)
-                }
-                onHideTooltip:
-                {
-                    base.hideTooltip()
-                }
+//                onShowTooltip:
```
nallath: Commented out code should be removed.

resources/qml/ActionPanel/ActionPanelWidget.qml
```diff
@@ -0,0 +1,53 @@
...
+// Cura is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.7
+import QtQuick.Controls 2.1
+import QtQuick.Layouts 1.3
+
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+Rectangle
```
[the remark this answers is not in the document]
diegopradogesto: This is not intended to be a reusable component. It's only one component, as far as I know. That's why I keep it as a rectangle.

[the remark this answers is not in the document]
nallath: Fair enough.

resources/qml/ActionPanel/ActionPanelWidget.qml
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2018 Ultimaker B.V.
+// Cura is released under the terms of the LGPLv3 or higher.
+
```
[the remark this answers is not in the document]
diegopradogesto: Documentation added in dba31c3af435202ef93363859017927d4a043bd2

resources/qml/ActionPanel/OutputDevicesActionButton.qml
```diff
@@ -0,0 +1,103 @@
...
+        iconSource: popup.opened ? UM.Theme.getIcon("arrow_top") : UM.Theme.getIcon("arrow_bottom")
+        color: UM.Theme.getColor("action_panel_secondary")
+        visible: (devicesModel.deviceCount > 1)
+
+        onClicked: popup.opened ? popup.close() : popup.open()
+
+        Popup
+        {
+            id: popup
+            padding: 0
```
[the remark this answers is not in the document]
diegopradogesto: By default in a popup, it looks like the default padding is 1 unit. In this case I don't want to have any padding, that's why I set it to zero.

resources/qml/ActionPanel/OutputDevicesActionButton.qml
diegopradogesto: I'm using a onHover color and instead of defining another button, I used the ActionButton that has that property. I don't know what is best then, I've just tried to reuse the elements that we have.

I will keep it like this. But I remove one line that is not useful here: 059e681de59f073870d6858daac7b0cd85528be1

resources/qml/ActionPanel/PrintJobInformation.qml
diegopradogesto: Removed here: 059e681de59f073870d6858daac7b0cd85528be1

resources/qml/Cura.qml
diegopradogesto: Removed here: 059e681de59f073870d6858daac7b0cd85528be1

resources/qml/ActionPanel/OutputProcessWidget.qml
diegopradogesto: Removed in 2dfd354b25e3187f403cc1cffc028d61eb7bae2e

resources/qml/ActionPanel/OutputProcessWidget.qml
diegopradogesto: I wrote it in a different function for readability. I prefer to have a separate function that does something more than a line in a function, but if you don't like it, I don't have a problem in changing it 😃 

Done in 2dfd354b25e3187f403cc1cffc028d61eb7bae2e

resources/qml/ActionPanel/PrintJobInformation.qml
diegopradogesto: Done in 6f2e8d726dac8488f6547f39556844afbd3cf4db

diegopradogesto: Done in 6f2e8d726dac8488f6547f39556844afbd3cf4db

resources/qml/ActionPanel/OutputDevicesActionButton.qml
nallath: Right, my bad. I was confused with the controls 2 button which does have a hover.

diegopradogesto:
Yes, the code is approved and the component test was done, but we are still waiting for the other PRs to be merged.

PR #4717, opened 2018-11-02
Fix Wanhao profiles to use speed_travel macro

forkineye:
The Wanhao profiles are broken as they still use the old `travel_speed` macro. This just updates them to use `speed_travel`.

### DIFF
--- a/resources/definitions/wanhao_d4s.def.json
+++ b/resources/definitions/wanhao_d4s.def.json
@@ -39,10 +39,10 @@
       "default_value": "RepRap (Marlin/Sprinter)"
     },
     "machine_start_gcode": {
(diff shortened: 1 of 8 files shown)


nallath:
For our reference; CURA-5901

OutsourcedGuru:
Okay, this was changed for the D4S and the title indicates "Fix Wanhao profiles..." in plural. The bug still remains in the [D9 definition](https://github.com/Ultimaker/Cura/blob/490d933923d0255363a907f66e4d52bcf95afb6f/resources/definitions/wanhao_d9.def.json), btw. Just had to deal with this for an end-user.

Ghostkeeper:
Indeed, the D9 still had the problem. I fixed that up for you: 76d95037771ac1d8f6c23f4fb994204b382de82c

The string `travel_speed` now only occurs in Cura in the plug-in that reads profiles from legacy Cura versions.

OutsourcedGuru:
👍 Good to hear, thanks.

PR #4730, opened 2018-11-05
Make quality manager also include profiles with same GUID in fallback.

nallath:
This ensures that if a (duplicated) material does not have a generic variant, that
the material it was copied from will be added to the fallback list.

CURA-5898

### DIFF
--- a/cura/Machines/MaterialManager.py
+++ b/cura/Machines/MaterialManager.py
@@ -12,7 +12,6 @@ from UM.Application import Application
 from UM.ConfigurationErrorMessage import ConfigurationErrorMessage
 from UM.Logger import Logger
 from UM.Settings.ContainerRegistry import ContainerRegistry
-from UM.Settings.InstanceContainer import InstanceContainer
 from UM.Settings.SettingFunction import SettingFunction
 from UM.Util import parseBool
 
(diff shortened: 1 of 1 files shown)


cura/Machines/QualityManager.py
```diff
@@ -259,11 +259,13 @@ def getQualityGroups(self, machine: "GlobalStack") -> Dict[str, QualityGroup]:
...
-                material_type = extruder.material.getMetaDataEntry("material")
-                fallback_root_material_id = self._material_manager.getFallbackMaterialIdByMaterialType(material_type)
-                if fallback_root_material_id:
-                    root_material_id_list.append(fallback_root_material_id)
+                # Also try to get the fallback materials
+                fallback_ids = self._material_manager.getFallBackMaterialIdsByMaterial(extruder.material)
+
+                if fallback_ids:
+                    root_material_id_list.extend(fallback_ids)
+                root_material_id_list = list(set(root_material_id_list))  # Weed out duplicates
```
Ghostkeeper: This set transformation makes the resulting `root_material_id_list` have a nondeterministic ordering. The ordering is used down below in the inline `addNodesToCheck` function where the first material that is compatible in the current set-up is chosen, so which material is chosen is then random.

Here's a suggestion to save you some effort: https://stackoverflow.com/questions/480214/how-do-you-remove-duplicates-from-a-list-whilst-preserving-order

PR #4731, opened 2018-11-05
[3.6] [CURA-5887] Fix material shown in the print job tile

LipuFei:
CURA-5887

 - Use MaterialManager to get materials
 - For a GUID with mulitple matches, show read-only materials first if any. Otherwise, show non-read-only materials.


nallath:
Yeah, there was a bit of discussion if this should be in 3.6 at all (hence it not yet being merged). Since this has been an issue for quite some time, this is going to be in master (and not 3.6).

PR #4732, opened 2018-11-05
[3.6] [CURA-5896] Fix naming bugs and object shadow

LipuFei:
CURA-5896

### DIFF
--- a/cura/PrintInformation.py
+++ b/cura/PrintInformation.py
@@ -369,6 +369,11 @@ class PrintInformation(QObject):
             else:
                 self._base_name = ""
 
+            # Strip the old "curaproject" extension from the name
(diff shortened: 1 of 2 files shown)


cura/PrintInformation.py
```diff
@@ -369,6 +369,11 @@ def setBaseName(self, base_name: str, is_project_file: bool = False) -> None:
             else:
                 self._base_name = ""
 
+            # Strip the old "curaproject" extension from the name
+            OLD_CURA_PROJECT_EXT = ".curaproject"
```
Ghostkeeper: Code style: Use lower_case_underscores for variable names.

Ghostkeeper: Ah, actually, we can see it as a constant and then it's fine like this.

PR #4736, opened 2018-11-06
Updated pt_BR strings for 3.6

Patola:


Ghostkeeper:
I found some more mistakes, and corrected them here: https://github.com/Ultimaker/Cura/commit/9fa9c04517ff9a53b318ccece22bf22788275f22

I hope you don't mind, otherwise we can revert some of the changes. I think some of these things were done to make the texts more consistent (period at the end of every setting description) but for us it's easier if the translations are consistent with the source texts. The source texts should be corrected if they are inconsistent with each other so then it'll be corrected in every language.

PR #4748, opened 2018-11-07
[3.6] [CURA-5900] Rename "CC Red 0.6" to "CC 0.6"

LipuFei:
CURA-5900


fieldOfView:
Will this affect people who have run the 3.6 beta?

diegopradogesto:
Yes, it will but only for the people that have CC print cores. Since we are not providing those print core publicly yet, we are (sort of) safe. People that created project files with the previous CC Red 0.6 in 3.6 BETA, will get configuration errors.

We are aware of it, but it's the only thing we could do. If we don't change this back to CC 0.6, the firmware won't be prepared for that, so this is needed.

PR #4753, opened 2018-11-08
Feature: "Z Hop After Extruder Switch Height"

vgribinchuk:
Feature: "Z Hop After Extruder Switch Height"
This feature allows to set the height difference when performing a Z Hop after extruders switching. Currently, all z hops have the same height. This setting would allow a different z hop distance to be set for when the extruders switch. Z hop during extruders switching is turned on by original setting "retraction_hop_after_extruder_switch".
General behavior:
After previous extruder finishes, z should be moved by "Z Hop After Extruder Switch Height" amount. After next extruder starts, the extruder should move to the print, and then z should back.
New setting:
"Z Hop After Extruder Switch Height"

This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/903.


diegopradogesto:
Devs. CURA-5920 (same as for the PR in CuraEngine)

sozkani:
Z hope should work at travel speed, but working at printing speed or less, it leaves alot of material during the PETG printing.

resources/definitions/fdmprinter.def.json
```diff
@@ -3529,6 +3529,19 @@
                     "enabled": "retraction_hop_enabled and extruders_enabled_count > 1",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
+                },
+                "retraction_hop_after_extruder_switch_height":
```
rburema: Just like the last PR, it would be good to add this too to one of the cfg-files that regulate when to show the setting, like `expert.cfg`.

resources/definitions/fdmprinter.def.json
```diff
@@ -3529,6 +3529,19 @@
...
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
+                },
+                "retraction_hop_after_extruder_switch_height": 
+                {
+                    "label": "Z Hop After Extruder Switch Height",
+                    "description": "The height difference when performing a Z Hop after extruder switch.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 1,
```
rburema: I see you have copied the `default_value` from `retraction_hop`, to not get different behaviour if people leave the settings on default. This is fine, but consider that (the config files of) some types of printer (like the ultimaker3 or s5 for instance) can override the previous setting, and this one will not change. In that case, it will behave differently.

Luckily, this is easy to fix; in addition to `"default_value": 1,`, also add `"value": "retraction_hop",`. This will ensure that it copies the value of `retraction_hop` to here (even if overridden or changed by the user), unless this value (`retraction_hop_after_extruder_switch_height`) is changed manually by the user.

rburema:
@vgribinchuk Just a heads-up: We're going to start a new release branch soon(ish), so if you want this to be in there, you might not want to wait too long.

resources/definitions/fdmprinter.def.json
vgribinchuk: Thanks, retraction_hop_after_extruder_switch_height was added to expert.cfg as you suggested

resources/definitions/fdmprinter.def.json
vgribinchuk: Thank you, added `"value": "retraction_hop"`

vgribinchuk:
Sorry for the delay! Issues you mentioned have been fixed.

rburema:
No need to apologize, we would have had plenty of other things to work on :-)

I'll see if I can do the code review, component test (& then hopefully merge) later today.

vgribinchuk:
Thank you!

rburema:
@sozkani 

>Z hope should work at travel speed, but working at printing speed or less, it leaves alot of material during the PETG printing.

That seems like a separate thing from this PR. If you think that's a problem, please raise an issue for it.

PR #4756, opened 2018-11-08
While initial loading Machine Manager was triggered global stack chan…

alekseisasin:
…ged, but extruder stacks were not ready.

CURA-5893


cura/Settings/ExtruderManager.py
```diff
@@ -307,15 +305,15 @@ def getActiveExtruderStacks(self) -> List["ExtruderStack"]:
...
+    def globalContainerStackChanged(self) -> None:
         # If the global container changed, the machine changed and might have extruders that were not registered yet
         self._addCurrentMachineExtruders()
 
         self.resetSelectedObjectExtruders()
 
     ##  Adds the extruders of the currently active machine.
     def _addCurrentMachineExtruders(self) -> None:
-        global_stack = self._application.getGlobalContainerStack()
+        global_stack = self._application.getMachineManager()._global_container_stack
```
nallath: Access of a protected property

cura/Settings/MachineManager.py
```diff
@@ -375,10 +375,13 @@ def setActiveMachine(self, stack_id: str) -> None:
             # Mark global stack as invalid
             ConfigurationErrorMessage.getInstance().addFaultyContainers(global_stack.getId())
             return  # We're done here
-        ExtruderManager.getInstance().setActiveExtruderIndex(0)  # Switch to first extruder
+
         self._global_container_stack = global_stack
+
+        ExtruderManager.getInstance().globalContainerStackChanged()
```
nallath: Instead of doing this (and changing the _addCurrentMachineExtruders), i think its better to do the following:
addMachineExtruders(machine_id) <- This function gets the machine as provided by the ID and ensures that it's extruders have been initialised. That way it doesn't have to get a different kind of global stack.

cura/Settings/ExtruderManager.py
alekseisasin: ok

cura/Settings/MachineManager.py
alekseisasin: I think adding a new method addMachineExtruders(self, machine_id : str): will just duplicate part of the code:
        container_registry = CuraContainerRegistry.getInstance()
        containers = container_registry.findContainerStacks(id = stack_id)
        global_stack = containers[0]

I will just pass global_stack as an argument:
ExtruderManager.getInstance().globalContainerStackChanged(global_stack)

nallath: But the function would also work if another stack changed (eg; it's not just about global stack, right?) So in that case it would be better to make it more generic (naming wise)

diegopradogesto:
Is this PR still needed, otherwise it has to be closed.

nallath:
I guess not.

PR #4782, opened 2018-11-12
CPE and CPE+ Comb retractions

THeijmans:
In this commit I changed all CPE and CPE+ profiles to have a 50mm combing max distance without retraction. This should decrease blobs caused by combing long distances.
I also reverted the CPE infill back to triangles, as we now connect this infill type as well.

### DIFF
--- a/resources/quality/ultimaker_s5/um_s5_aa0.25_CPE_Normal_Quality.inst.cfg
+++ b/resources/quality/ultimaker_s5/um_s5_aa0.25_CPE_Normal_Quality.inst.cfg
@@ -12,6 +12,7 @@ material = generic_cpe
 variant = AA 0.25
 
 [values]
+retraction_combing_max_distance = 50
 retraction_extrusion_window = 0.5
 speed_infill = =math.ceil(speed_print * 40 / 55)
 speed_topbottom = =math.ceil(speed_print * 30 / 55)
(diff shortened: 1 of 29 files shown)


Ghostkeeper:
Something must be changed in our build system then to remove the double data. Either:
* we delete the files from Cura and fdm_materials, make cura-resources public and modify Cura's build structure to put cura-resources in the build, or
* we delete cura-resources and build the cloud-based packages from the current state of fdm_materials and Cura.

I'd argue for the second one, as to keep the file history intact and make it easier on Embedded and external contributors.

ChrisTerBeke:
@nallath and I didn't opt for the second option as that means we have to scrape all the data from multiple places every time a change is made. In my opinion the data should be separate from the logic, not in the same repository. Of course this is entirely up to Cura team, but I strongly suggest not having to go through the flow that @Appesteijn and I had to do to get the right data collected. In the end, it's up to you to keep the packages in the Marketplace up to date.

diegopradogesto:
Hi @THeijmans, the PR looks correct. However I was wondering why you only change the profiles for the CPE and CPE+ for UMS5 and the ABS profile for UM3. Shouldn't we keep the profiles as similar as possible?

THeijmans:
Hi Diego, the CPE+ and ABS profiles were already different between the two printers, to compensate for issues we see mostly on bigger prints. The ABS issues were therefore only seen on the UM3.
CPE+ on the UMS5 had combing turned off, this is a nicer solution.

PR #4788, opened 2018-11-13
Make Cura app name configurable in CuraVersion

fieldOfView:
This PR makes the "app name" configurable in CuraVersion, like the version and application display name already are. The app name is mainly used for configuration folder names. This PR is makes it less likely that forks use the same configuration folder as mainline Cura and mess up configuration folders with their configurations.


nallath:
CURA-5949

fieldOfView:
Is this being considered at all? I wouldn't mind updating the branch (there are conflicts now), but only if it stands any chance of being included.

nallath:
Yes it is, but we're currently lagging behind bigtime on the pull requests. The 4.0 release has been eating up most of our time due to that getting more priority. 

Right now we're in a spot where we have a bit more breathing room (hardly any work left to "finish" the 4.0 release) so I expect that we will start to pick up the work we left lying around for far too long.

fieldOfView:
I did not mean to be impatient, but I was really wondering if the PR was something that would be accepted.  On the one hand it promotes creating forks (which I don't think you would want), on the other hand it could alleviate support requests from people mixing forks on a single computer (which is preferable).

I'll fix the merge conflicts.

nallath:
I didn't see it as any form of impatience. It's just explaining why it took us so long to handle this. It's the least we can do (because if people put effort into something, I feel it should also "force" me to put in an effort as well, and we've been slow on our side of things)

PR #4810, opened 2018-11-16
Edit to fix an issue with Marlin Printers

N95JPL:
Marlin Printers saw the : after layer as a split command and issued "Unknown command: #" (# being what the layer should be)

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/DisplayFilenameAndLayerOnLCD.py
+++ b/plugins/PostProcessingPlugin/scripts/DisplayFilenameAndLayerOnLCD.py
@@ -36,10 +36,10 @@ class DisplayFilenameAndLayerOnLCD(Script):
             name = self.getSettingValueByKey("name")
         else:
             name = Application.getInstance().getPrintInformation().jobName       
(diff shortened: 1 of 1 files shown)


plugins/PostProcessingPlugin/scripts/DisplayFilenameAndLayerOnLCD.py
```diff
@@ -36,10 +36,10 @@ def execute(self, data):
             name = self.getSettingValueByKey("name")
         else:
             name = Application.getInstance().getPrintInformation().jobName       
-        lcd_text = "M117 " + name + " layer: "
+        lcd_text = "M117 " + name + " layer "
         i = 0
         for layer in data:
-            display_text = lcd_text + str(i)
+            display_text = lcd_text + str(i) + "!"
```
fieldOfView: Why are you shouting? (vestigial debug statement?)

N95JPL: Ignore this commit - I have since opened a fresh commit with enhancements.

PR #4816, opened 2018-11-16
Tests for um3networkplugin

marijnd:
In preparation for printing from the cloud I started adding tests to the UM3NetworkPrinting plugin.
I started adding test for SendMaterialJob and refactored that class to make it more testable.

This also fixes CURA-5947.


plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,87 @@
...
+    brand: None  # type: Optional[str]
+
+    version = None  # type: Optional[int]
+
+    color: None  # type: Optional[str]
+
+    density: None  # type: Optional[float]
+
+
+class LocalMaterialProperties(BaseModel):
```
ChrisTerBeke: Could use some comments.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
 
-import json #To understand the list of materials from the printer reply.
-import os #To walk over material files.
-import os.path #To filter on material files.
-from PyQt5.QtNetwork import QNetworkReply, QNetworkRequest #To listen to the reply from the printer.
-from typing import Any, Dict, Set, TYPE_CHECKING
-import urllib.parse #For getting material IDs from their file names.
-
-from UM.Job import Job #The interface we're implementing.
+import json  # To understand the list of materials from the printer reply.
```
ChrisTerBeke: These comments were already here, but they seem not needed as these are just used in the class.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
 #   This way it won't freeze up the interface while sending those materials.
 class SendMaterialJob(Job):
     def __init__(self, device: "ClusterUM3OutputDevice") -> None:
         super().__init__()
-        self.device = device #type: ClusterUM3OutputDevice
+        self.device = device  # type: ClusterUM3OutputDevice
 
     def run(self) -> None:
-        self.device.get("materials/", on_finished = self.sendMissingMaterials)
+        self.device.get("materials/", on_finished=self.sendMissingMaterials)
```
ChrisTerBeke: Cura uses spaces around assignments.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
-                materials_to_send.add(material["id"])
-                continue
-            try:
-                if int(material["version"]) > remote_materials_by_guid[guid]["version"]:
-                    materials_to_send.add(material["id"])
-                    continue
-            except KeyError:
-                Logger.log("e", "Current material storage on printer was an invalid reply (missing version).")
-                return
+        # Find out what materials are new or updated annd must be sent to the printer
```
ChrisTerBeke: typo: annd

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
+            material.id
+            for guid, material in local_materials_by_guid.items()
+            if guid not in remote_materials_by_guid or
+               material.version > remote_materials_by_guid[guid].version
+        }
 
+        # Send materials to the printer
+        self.sendMaterialsToPrinter(materials_to_send)
+
+    def sendMaterialsToPrinter(self, materials_to_send):
```
ChrisTerBeke: Missing comments and typing.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
         for file_path in Resources.getAllResourcesOfType(CuraApplication.ResourceTypes.MaterialInstanceContainer):
             try:
                 mime_type = MimeTypeDatabase.getMimeTypeForFile(file_path)
             except MimeTypeDatabase.MimeTypeNotFoundError:
-                continue #Not the sort of file we'd like to send then.
+                continue  # Not the sort of file we'd like to send then.
+
```
ChrisTerBeke: It might be nice to extract a method for the actual sending so there is not a giant code block in this for loop.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
+    @classmethod
+    def _getLocalMaterials(cls):
+        result = {}
+        for material in ContainerRegistry.getInstance().findContainersMetadata(type="material"):
+            try:
+                localMaterial = LocalMaterial(**material)
+
+                if localMaterial.GUID not in result or localMaterial.version > result.get(localMaterial.GUID).version:
+                    result[localMaterial.GUID] = localMaterial
+            except (ValueError):
```
ChrisTerBeke: Maybe log the actual value error here (except ValueError as e) as something else than the material ID might be incorrect?

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
```diff
@@ -0,0 +1,314 @@
...
+        "material": "PLA",
+        "brand": "Generic",
+        "version": 2,
+        "color": "Black",
+        "density": 1.00
+    }
+
+    def setUp(self):
+        # Make sure the we start with clean (log) slate
+        _logentries.clear()
```
ChrisTerBeke: new_log and _logentries seem only used in this test class, so it could just be a property on the class instead of a global one.

plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+from typing import Optional
```
DanielSchiavini: I believe the Ultimaker standard says you need to put every class in its own file

plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class BaseModel:
```
DanielSchiavini: We have a basic model system that provides more features in https://github.com/Ultimaker/stardust-websocket_protocol/blob/95a4e1ff/protocol/models/BaseModel.py
I prefer to reuse that than to create a new one

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
         try:
-            remote_materials_list = json.loads(remote_materials_list)
+            remote_materials_by_guid = self._parseReply(reply)
         except json.JSONDecodeError:
             Logger.log("e", "Request material storage on printer: I didn't understand the printer's answer.")
             return
-        try:
-            remote_materials_by_guid = {material["guid"]: material for material in remote_materials_list} #Index by GUID.
         except KeyError:
             Logger.log("e", "Request material storage on printer: Printer's answer was missing GUIDs.")
```
DanielSchiavini: please add the key to the output (`KeyError as key`)

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
marijnd: I tried to split it up into two parts, but it does not make it clearer. The file_name used in the extraction of the material_id is also used to send the contents of the file to the printer.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
marijnd: Someone must have found them useful

plugins/UM3NetworkPrinting/src/Models.py
ChrisTerBeke: I discussed this with Marijn and we decided it was not a good idea to bring that whole (part of the) code base into Cura at the moment (as no data modeling is done at all yet in this repo).

plugins/UM3NetworkPrinting/src/Models.py
nallath: Yup.

plugins/UM3NetworkPrinting/src/Models.py
nallath: I think it's actually better to use named tuples in this case; https://stackoverflow.com/questions/34269772/type-hints-in-namedtuple

plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__ if type(self) == type(other) else False
+
+
+##  Represents an item in the cluster API response for installed materials.
+class ClusterMaterial(BaseModel):
+    def __init__(self, **kwargs):
```
nallath: Missing typing -> None

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
+        remote_materials_list = json.loads(reply.readAll().data().decode("utf-8"))
+        return {material["guid"]: ClusterMaterial(**material) for material in remote_materials_list}
+
+    ##  Retrieves a list of local materials
+    #
+    #   Only the new newest version of the local materials is returned
+    #
+    #   \return a dictionary of LocalMaterial objects by GUID
+    @classmethod
+    def _getLocalMaterials(cls):
```
nallath: Missing typing

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
+        result = {}
+        for material in ContainerRegistry.getInstance().findContainersMetadata(type="material"):
+            try:
+                localMaterial = LocalMaterial(**material)
```
nallath: should be local_material

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
nallath: It also doesn't need to be encapsulated in a tuple.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,129 @@
...
+
+                if localMaterial.GUID not in result or localMaterial.version > result.get(localMaterial.GUID).version:
+                    result[localMaterial.GUID] = localMaterial
+            except (ValueError):
+                Logger.log("e", "Material {material_id} has invalid version number {number}.".format(
```
nallath: You can use Logger.logException("e") if you want to log the exception as well as log a message.

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def tearDown(self):
+        # If there are still log entries that were not checked something is wrong or we must add checks for them
+        self.assertEqual(len(_logentries), 0)
+
+    @patch("plugins.UM3NetworkPrinting.src.ClusterUM3OutputDevice")
+    def test_run(self, device_mock):
+        with mock.patch.object(Logger, 'log', new=new_log):
```
nallath: Use double quotes for strings

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
```diff
@@ -0,0 +1,314 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
```
nallath: Plugins need to contain their own tests (eg; see CuraProfileReader or OBJReader for instance)

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
nallath: Also, why patch the logger?

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
nallath: It would make it easier to mock though. Right now you need a ton of mocking & patching to fix this (which makes the test pretty hard to read)

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
ChrisTerBeke: It seems the logger is used for asserting later on that certain things have been logged, but I'm also not a fan of this as the logs are meant for developer debugging only and can change.

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
ChrisTerBeke: I think this was already moved but not pushed yet :(

Also, the tests in plugins are not ran by the CI under Linux (as they are not covered in pytest.ini), that was fixed on Marijns machine as well, including some things that fell over because of it (incorrect import paths in the version upgrader tests).

nallath:
A bit more general overview; I think there is way too much magic patching going on in this set of tests. I think it would be better to change the way that the sendMaterialsJob does its thing; Separate the finding of the profiles it needs to send with the sending of a single profile (and let the run simply do all of these in sequence; Find the profiles & send them one by one).

That way it's way easier to test since you can test each of these functions one at a time.

ChrisTerBeke:
@nallath I'm working on this a bit right now. My plan is to test the complete job in a few scenarios, not every single method (as that's prone to break with the slightest change in the code).

ChrisTerBeke:
@nallath I have now only 2 tests left just to be sure the setup works,and they run fine locally, but in the CI they continue to fail. What should we change in the code to make them pass?

ChrisTerBeke:
OMG they run :P

plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,32 @@
...
+    'guid',
+    'material',
+    'brand',
+    'version',
+    'color',
+    'density'
+])
+
+LocalMaterial = namedtuple('LocalMaterial', [
+    'GUID',
```
fieldOfView: why is GUID capitalised here, but not in ClusterMaterial?

ChrisTerBeke: Because the keys are different between the networked printer response and the local registry. Don’t ask why...

fieldOfView: But wh... oh, ok.

Perhaps that should be documented then, and here could be a good place to do so.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
sedwards2009: That's not really an assignment though. It's just parameter passing.

plugins/UM3NetworkPrinting/src/Models.py
ChrisTerBeke: Fair enough. This whole thing is not near it's final from though, still figuring out a lot of stuff to make it more testable.

sedwards2009:
Regarding NameTuples etc, I don't know exactly which Python version Cura is on these days but it may be the lucky first team to be able to use Python's new Data Classes which is probably the best way of doing small data structures now that also work with mypy.

ChrisTerBeke:
@sedwards2009 Cura's on 3.6 in the latest builds, so no `dataclass` yet as I see that's new in 3.7.

sedwards2009:
me sad

tests/plugins/UM3NetworkPrinting/TestSendMaterialJob.py
marijnd: Should the code change for that? I see that the other test use a different root for imports.

ChrisTerBeke: Nope, sort of resolved now by having a custom pycharm setup for the tests.

ChrisTerBeke:
@nallath the CI seems to have stopped running for all new commits.

sedwards2009:
For the material handling tests I don't like seeing all the mocking that is going on. It is very implementation specific and greatly reduces the amount of code being tested and thus its usefulness. The tests are fragile.

An improvement would be to not mock the Qt network code but to keep it and send the HTTP requests to a mock server. Moving the boundary to the rest API is a smarter move because it is a natural system boundary and it is stable. We can't easily change this Rest API. We have too many machines out in the wild using it already. This would make the test more realistic and less likely to need changes/updates in the future.

There are packages to help do this kind of testing. The first one I saw was https://github.com/Gr1N/pytest-mockservers . There could be better ones out there.

ChrisTerBeke:
It's very hard to test Cura anyways because of how it's set up. I don't really see a problem with mocking the reply this was versus having a whole mock server running which basically does the same thing. We can just assume that at least QNetworkManager does it's thing like it's supposed to, we don't have to test that. In the end it's up to you guys to review this, but at least there are tests now at all, previously this whole functionality was just broken because there were not tests and no one saw it breaking...

marijnd:
This is a Unit test for one class, not an integration test for the whole plugin. We certainly do not want to be testing the Qt framework.

If you feel that the data does not represent real replies from printers, let me know what you would like to add.

sedwards2009:
My objections are basically:

* assuming the use of Qt for the HTTP requests is an unnecessary coupling to the implementation which makes the test fragile to code changes.
* the tests just assume that the code correctly calls into QtNetwork. Why assume this when a better test with a mock server can cover it?

Testing against a mock server is likely to be a little more work but still worth it.

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
```diff
@@ -0,0 +1,210 @@
...
+    @patch("plugins.UM3NetworkPrinting.src.ClusterUM3OutputDevice")
+    def test_run(self, device_mock):
+        job = SendMaterialJob(device_mock)
+        job.run()
+
+        # We expect the materials endpoint to be called when the job runs.
+        device_mock.get.assert_called_with("materials/", on_finished = job._onGetRemoteMaterials)
+
+    @patch("plugins.UM3NetworkPrinting.src.ClusterUM3OutputDevice")
+    @patch("PyQt5.QtNetwork.QNetworkReply")
```
DanielSchiavini: the mocks needed in every method might as well be in the class level

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
```diff
@@ -0,0 +1,210 @@
...
+        reply_mock.attribute.return_value = 404
+        job = SendMaterialJob(device_mock)
+        job._onGetRemoteMaterials(reply_mock)
+
+        # We expect the error string to be retrieved and the device not to be called for any follow up.
+        self.assertEqual([call.attribute(0), call.errorString()], reply_mock.method_calls)
+        self.assertEqual(0, device_mock.createFormPart.call_count)
+
+    @patch("plugins.UM3NetworkPrinting.src.ClusterUM3OutputDevice")
+    @patch("PyQt5.QtNetwork.QNetworkReply")
```
DanielSchiavini: It doesn't make sense to mock `QNetworkReply` because that could still mean the request is being actually sent via HTTP.
What we need is to mock the class that results in this reply, i.e. `postFormWithParts.mock_calls[0][2]['on_finished'](reply_mock)`

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -147,6 +147,9 @@ def _createEmptyRequest(self, target: str, content_type: Optional[str] = "applic
         request.setHeader(QNetworkRequest.UserAgentHeader, self._user_agent)
         return request
 
+    def createFormPart(self, content_header: str, data: bytes, content_type: Optional[str] = None) -> QHttpPart:
```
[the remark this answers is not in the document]
ChrisTerBeke: That breaks existing code, as in some places it was called privately....
As this could also be done in plugins, we should never break an existing interface, even though it was supposed to be public in the first place.

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
marijnd: They are

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
marijnd: I'm not sure what you mean.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,191 @@
...
 
-        remote_materials_list = reply.readAll().data().decode("utf-8")
+        # Collect materials from the printer's reply and send the missing ones if needed.
         try:
-            remote_materials_list = json.loads(remote_materials_list)
+            remote_materials_by_guid = self._parseReply(reply)
+            self._sendMissingMaterials(remote_materials_by_guid)
         except json.JSONDecodeError:
-            Logger.log("e", "Request material storage on printer: I didn't understand the printer's answer.")
+            Logger.logException("w", "Error parsing materials from printer")
```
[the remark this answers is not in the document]
marijnd: Yes, previously this was not possible, but after refactoring it does.

Not sure what the Cura Team policy is on logging the response. It could be large en fill up the log file with garbage. I just refactored what was there.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -46,9 +46,8 @@ def _onGetRemoteMaterials(self, reply: QNetworkReply) -> None:
         # Collect materials from the printer's reply and send the missing ones if needed.
         try:
```
ChrisTerBeke: I think both exceptions can be in the _parseReply method then?

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
```diff
@@ -0,0 +1,197 @@
...
+        container_registry_mock.findContainersMetadata.return_value = [localMaterialWhiteWithInvalidVersion]
+
+        application_mock.getContainerRegistry.return_value = container_registry_mock
+
+        with mock.patch.object(CuraApplication, "getInstance", new = lambda: application_mock):
+            job = SendMaterialJob(device_mock)
+            job._onGetRemoteMaterials(reply_mock)
+
+        self.assertEqual([call.attribute(0), call.readAll()], reply_mock.method_calls)
+        self.assertEqual([call.getContainerRegistry()], application_mock.method_calls)
```
Ghostkeeper: You're testing an implementation detail with these assertions. It doesn't matter that the `Application.getContainerRegistry()` method is called; it's not a requirement of the `SendMaterialJob`. Instead test the results, so that if we choose to change the implementation, the test won't break.

This goes for most of the assertions against `method_calls` in these tests. The ones that test against `createFormPart` and such are the only ones that actually test the public behaviour.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,191 @@
...
 ##  Asynchronous job to send material profiles to the printer.
 #
 #   This way it won't freeze up the interface while sending those materials.
 class SendMaterialJob(Job):
+
     def __init__(self, device: "ClusterUM3OutputDevice") -> None:
         super().__init__()
-        self.device = device #type: ClusterUM3OutputDevice
+        self.device = device  # type: ClusterUM3OutputDevice
+        self._application = CuraApplication.getInstance()  # type: CuraApplication
```
Ghostkeeper: We only need to use `Application`, `CuraApplication` is unnecessary because you're not using one of its specialised functions.

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
```diff
@@ -0,0 +1,197 @@
...
+from unittest.mock import patch, call
+
+from PyQt5.QtCore import QByteArray
+
+from UM.MimeTypeDatabase import MimeType
+from cura.CuraApplication import CuraApplication
+from plugins.UM3NetworkPrinting.src.SendMaterialJob import SendMaterialJob
+
+
+@patch("builtins.open", lambda _, __: io.StringIO("<xml></xml>"))
```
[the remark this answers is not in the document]
marijnd: For this these tests the contents of the file are not relevant.

plugins/UM3NetworkPrinting/tests/TestSendMaterialJob.py
marijnd: I agree. But that is the only way we can verify the methods. This is the best we can do now.

I think this a result of retrofitting tests on a class that not developed test-first. No consideration is given to how to check for the correct working of the code when the code was written.

nallath: If it's the only way to verify the methods, then don't. The alternative is far worse; Checking behavior that doesn't matter if it's done in that way. 
It's about testing the result, not about testing the way it did something.

marijnd: Ok, I removed the asserts on the internals and only look at the output.

plugins/UM3NetworkPrinting/src/SendMaterialJob.py
```diff
@@ -1,99 +1,193 @@
...
+        material_containers = container_registry.findContainersMetadata(type = "material")
+
+        # Find the latest version of all material containers in the registry.
+        for material in material_containers:
+            try:
+                # material version must be an int
+                material["version"] = int(material["version"])
+
+                # Create a new local material
+                local_material = LocalMaterial(**material)
```
ChrisTerBeke: Populating the `namedtuple` with unknown keys will throw a `TypeError`, we should make this a bit more robust (as containers from the registry can contain all sorts of data).

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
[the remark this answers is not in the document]
ChrisTerBeke: protected/private, what's in the name. Anyways it was called from outside the class even though it has an underscore.

plugins/UM3NetworkPrinting/src/Models.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+
+## Base model that maps kwargs to instance attributes.
+class BaseModel:
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
ChrisTerBeke: Fixed, thanks.

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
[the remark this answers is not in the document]
ChrisTerBeke: Not anymore though because I fixed it in this PR, but we don't know if any plugins are doing it. Although we could argue that the next release has breaking changes anyways, we can break this as well?

plugins/UM3NetworkPrinting/src/Models.py
ChrisTerBeke: There's not real other way to fix it right now as we're stuck on Python 3.5. I don't like it 100% either, but this PR is already a HUGE improvement over the old situation.

sedwards2009: Explicitly write out the arguments and type them. I know it is not an ideal solution but it will allow mypy to check the call sites for the models.

ChrisTerBeke: I'll do that for the only attributes we use then so far. The rest can be added later.

marijnd:
I have not seen that before, but it means that you have a local material that has no GUID.

What should be done with such a material? Ignore it?

ChrisTerBeke:
It’s the empty material container. It was there before and is ignored in the result, but now we have a logging line there.

diegopradogesto:
OK, fair enough, but do we need to show the traceback there?

nallath:
I'd say no, since it's an expected event. We could just give the empty material an empty GUID?

PR #4817, opened 2018-11-16
Unittests FirmwareUpdateChecker

rburema:
Some unit-tests for the FirmwareUpdateChecker. Also some small refactors on the same, mostly to make testing easier (but the code is also very slightly improved thereby).


plugins/FirmwareUpdateChecker/tests/TestFirmwareUpdateChecker.py
```diff
@@ -0,0 +1,69 @@
...
+                "name": "olivia",
+                "check_urls": ["http://urlecho.appspot.com/echo?status=200&body=4.3.2.1", "https://this-url-clearly-doesnt-exist.net"],
+                "update_url": "https://ultimaker.com/en/resources/20500-upgrade-firmware",
+                "version_parser": "default"
+            },
+        "emmerson":
+            {
+                "id": 5,
+                "name": "emmerson",
+                "check_urls": ["http://urlecho.appspot.com/echo?status=200&body=0.2.2.2", "http://urlecho.appspot.com/echo?status=200&body=6.7.8.1"],
```
nallath: I'm not quite sure about this (although I don't quite have a better way of doing it), but this will fail the unit test if there is no internet connection.

rburema: You're right. I could have a look to see if it works with 'file://', since it seems that is supported (even for 2, despite the misleading name): https://stackoverflow.com/questions/20558587/opening-local-file-works-with-urllib-but-not-with-urllib2

rburema: Update: No dice; 'file://...' (like the other schemes) requires absolute paths, so it would have to be different for every builder.

plugins/FirmwareUpdateChecker/tests/TestFirmwareUpdateChecker.py
```diff
@@ -0,0 +1,69 @@
...
+    assert lookup.getRedirectUserUrl() is not None
+
+@pytest.mark.parametrize("name, version", [
+    ("ned"     , Version("5.1.2.3")),
+    ("olivia"  , Version("4.3.2.1")),
+    ("emmerson", Version("6.7.8.1")),
+])
+def test_FirmwareUpdateCheckerJob_getCurrentVersion(name, version):
+    machine_data = json_data.get(name)
+    job = FirmwareUpdateCheckerJob(False, name, machine_data, dummyCallback)
```
ChrisTerBeke: dummyCallback could be a MagicMock here, then you can check if the callback was (or was not) called depending on what you expect, e.g.:

```python
callback_mock = MagicMock()  # from unittest.mock import MagicMock
job = FirmwareUpdateCheckerJob(......, callback_mock)
assert callback_mock.call_count == 0  # or 1, or whatever it's expected to be
```

rburema: Thanks for the hints! I'm still learning the python way of doing unit-tests.

Unfortunately, in this specific case, the callback isn't ever reached, since that part of the code (where it decides to call it or not: `FirmwareUpdateCheckerJob.run()`) isn't called by the tests (the object still needs the callback in it's constructor though).

Maybe that method could be refactored into multiple smaller ones, but I already spent a lot of time on what was supposed to be a 'oh, also add the S5 to this check' task :-)

ChrisTerBeke: In that case you can still assert that the callback was indeed NOT called (not sure of the exact syntax for that). Then you know the method returned before the callback as expected.

PR #4832, opened 2018-11-19
CURA-5856 In case of using custom fdm printer use quality type by definition field

LipuFei:


fieldOfView:
It's a roller-coaster ride of emotions

PR #4834, opened 2018-11-19
Add BIBO2 printer (dual and single extruder settings)

mcmidget:
Please, add BIBO2 3D printer definition and extruders settings to supported resources.

### DIFF
--- a/resources/definitions/bibo2_dual.def.json
+++ b/resources/definitions/bibo2_dual.def.json
@@ -1,0 +1,98 @@
+{
+	"id": "BIBO2 dual",
+	"version": 2,
+	"name": "BIBO2 dual",
+	"inherits": "fdmprinter",
+	"metadata": {
+		"visible": true,
+		"author": "na",
+		"manufacturer": "BIBO",
+		"category": "Other",
+		"file_formats": "text/x-gcode",
+		"has_materials": true,
+		"machine_extruder_trains": {
+			"0": "bibo2_dual_extruder_0",
+			"1": "bibo2_dual_extruder_1"
+		},
+		"first_start_actions": [
+			"MachineSettingsAction"
+		]
+	},
+	"overrides": {
+		"machine_name": {
+			"default_value": "BIBO2 dual"
+		},
+		"machine_width": {
+			"default_value": 214
+		},
+		"machine_height": {
+			"default_value": 160
+		},
+		"machine_depth": {
+			"default_value": 186
+		},
+		"machine_center_is_zero": {
+			"default_value": true
+		},
+		"machine_heated_bed": {
+			"default_value": true
+		},
+		"machine_nozzle_size": {
+			"default_value": 0.4
+		},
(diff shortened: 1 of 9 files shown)


alekseisasin:
For our reference; CURA-5902

diegopradogesto:
This is a non-sense. There are 3 printers here, one is dual printer and two single extruder printers. The E1 and E2 (single printers) have 2 extruders. And indeed the three look almost exactly the same.

Could you please explain what did you want to do? Otherwise I would think that the best option is to remove the two single extruder printers.

mcmidget:
Hi and thanks for your comments,

The 'dual extruder definition' is the default configuration to use both extruders at the same time. I consider this is as default definition.

Both single extruders definitions are there to setup the printer as a 'single extruder printer', and configure **ONLY** one extruder for printing, either left (E2) or right (E1).

In real life, the 'dual extruder definition' performs great (even for single extruder printings), but manufacturer states the 3 definitions as part of provided configurations.

If you consider 'single extruders definitions' are superfluous, please feel free to remove them.

Anyway, It would be great to have at least the 'dual extruder definition' included.

Thanks in advance.

fieldOfView:
> Both single extruders definitions are there to setup the printer as a 'single extruder printer',

That's what disabling an extruder is for. Right click on an extruder tab to disable it, making the printer a single extrusion printer.

diegopradogesto:
Yes, of course we can keep the dual extruder printer definition. What I didn't understand is why to have a "single extruder" printer with two extruders. You could in this case create a printer definition and reuse the extruder definitions.

However the best option is to create a dual extrusion printer and let the user enable/disable the left or right extruder whenever they want by right clicking on the extruder tab.

mcmidget:
Thanks for clarifications. I guess 'single extruder defs' should be removed then.

As it seems is already merged in master, could you please remove single extruder def files for me?

 * resources/definitions/bibo2_single_extruder_0.def.json
 * resources/definitions/bibo2_single_extruder_1.def.json
 * resources/extruders/bibo2_single_extruder_0_0.def.json
 * resources/extruders/bibo2_single_extruder_0_1.def.json
 * resources/extruders/bibo2_single_extruder_1_0.def.json
 * resources/extruders/bibo2_single_extruder_1_1.def.json

Please, let me know if further input|actions are needed from my side, ok?

Thanks in advance.

diegopradogesto:
Yes, I've just done it here: 0794a2c8c9197c536ffcba090df72edd7d6e7fc6

diegopradogesto:
I also did another fix, that is moving the `material_diameter` and the `machine_nozzle_size` to the extruder definitions, otherwise we will have some issues with the materials. Please take a look if this is still correct: a6a364c337efb4f732247f30e60d14884aabdadc

Another quick question: the printer has the origin at center, is that correct?

mcmidget:
The fix looks perfect, and yes, the origin is at center.

Issue #4550, opened 2018-10-11
[3.5] Unable to add custom materials


TakuroFukamizu:
**Application Version**
3.5.0

**Platform**
Windows 10 Pro 1803

**Printer**
Custom Printer: Dobot Mooz-2

**Steps to Reproduce**
1. Change `Compatible material diameter` to `1.75 mm`  at Machine Settings dialog.
2. Open the `Materials` preference dialog
3. Duplicate the `Generic PLA` material and change `Diameter` to `1.75 mm`.

**Actual Results**
- Showing an information dialog that says "The new filament diameter is set to 1.75mm., which is not compatible with the current extruder. Do you wish to continue?".
- When I press the YES button, that custom material was disappeared.

**Expected results**
- Shouldn't show the above information dialog because filament diameter is compatible.
- Don't dissolve current material setting.

surfrock66:
This is happening on Cura 3.5.1 on Linux (the Appimage) as well.

surfrock66:
FYI this is NOT fixed in Cura 3.6; exact same thing is still happening

surfrock66:
Additionally, looking at Issue 4582, in Cura 3.6.0 the "Printer Settings" menu is not possible in the sidebar in the Linux AppImage.  Even when "All Settings" are checked, that sub-menu is not an option.  I don't see a place to configure a compatible filament diameter in the machine or extruder setting; this may be a Linux AppImage issue?

fieldOfView:
Could you test this prerelease version of th "Printer Settings" plugin?
http://files.fieldofview.com/cura/PrinterSettingsPlugin-v5.0.0-2018-11-16T14_28_49Z.curapackage
Just drop the file into the Cura window, and restart Cura.

surfrock66:
I tested that, changed the "Diameter" to 1.75, the setting stuck.  Went to settings, changed the material diameter to 1.75, it popped up the error and when I clicked continue the material profile disappeared.  The bug persists.  3.6.0 AppImage on Debian Buster.

fieldOfView:
In the materials dropdown in the sidebar, do you see Ultimaker materials or a whole list of brands?

surfrock66:
In settings, or in the actual print setup?  I don't see anything in the print setup sidebar, but here's what I have in the settings:

Materials Sidebar: https://imgur.com/ZCYZ81R
Print Settings Sidebar: https://imgur.com/wrwWmPm
[...]

PR #4836, opened 2018-11-19
Not activate materials if the printer has no materials.

diegopradogesto:
Do not allow the user to activate a material if the printer is not defined to have materials. In 3.0.4 it was like this, but it was changed in some refactor.

Contributes to #4550.

### DIFF
--- a/resources/qml/Preferences/Materials/MaterialsPage.qml
+++ b/resources/qml/Preferences/Materials/MaterialsPage.qml
@@ -83,7 +83,7 @@ Item
         {
             text: catalog.i18nc("@action:button", "Activate")
             iconName: "list-activate"
-            enabled: !isCurrentItemActivated && Cura.MachineManager.hasMaterials()
+            enabled: !isCurrentItemActivated && Cura.MachineManager.hasMaterials
             onClicked:
             {
                 forceActiveFocus()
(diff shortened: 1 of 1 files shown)


fieldOfView:
I would argue that the whole material list should be hidden, and a message should be shown that says that materials are disallowed for the current printer.

fieldOfView:
Creating, duplicating, importing and exporting should also be disabled when the current printer does not support materials. I think you can disable the whole row in one go.

diegopradogesto:
I sort of agree but you could still play around with the list of materials, by adding materials for other printers. It is like a complete manager. However it makes sense to disallow the whole entry in the left panel list (or all the buttons).

I would discuss it with UX and the rest of the team.

fieldOfView:
> I sort of agree but you could still play around with the list of materials, by adding materials for other printers.

But that's the whole UX problem! The list of materials on the Materials pane is closely related to the current printer; not only the approximate diameter, but also materials that are hidden for the current printer (eg PETG for Ultimaker). Playing around with the list of materials for other printers only confuses things.

diegopradogesto:
Agree, let's see if out UX team provides an even better solution or they agree with your proposal.

diegopradogesto:
Devs. CURA-6107

nallath:
I've made a follow-up issue for disabling the entire page.

PR #4837, opened 2018-11-19
[WIP] Cloud output device

ChrisTerBeke:
This is a WIP PR to add remote monitoring and queueing to Cura. Many iterations still to come, feedback on how to implement things is welcome. We'll try to keep the code neat.

* This PR is based on the `tests-for-um3networkplugin` branch but could also do without that.
* A `CloudOutputDeviceManager` is used to encapsulate all cloud related HTTP calls and logic. This could be merged into the main plugin class at a later stage, but the plugin class should first be refactored to be more readable before we add even more code to it.
* The `CloudOutputDevice` is currently inherited directly from `NetworkedPrinterOutputDevice` but could be inherited from the `UM3ClusterOutputDevice` later on to prevent code duplication. That class should then better support the differences in functionalities between local and remote via the controller (extension of feature flags needed in the controller).
* All networking related logic is refactored out of `NetworkedPrinterOutputDevice` to `NetworkClient`. This way we can easily re-use the nice functionality it offers of doing a network call with a specific callback (otherwise we have to parse the network reply to find the appropriate handling, like currently done in the `UM3OutputDevicePlugin`, which could be refactored as well to use the new base class). This reply parsing mess was originally the reason for me to no use QNetworkManager in for example the backups plugin but just Python requests which allows for neater synchronous code. I might also update that plugin once this makes it into a Cura release. @nallath any further suggestions on this?
* We still need a way to 'seamlessly' switch between local and remote cluster output devices depending on the network status. Not sure yet where this logic should be implemented.


cura/NetworkClient.py
```diff
@@ -0,0 +1,220 @@
...
+
+from UM.Logger import Logger
+from cura.CuraApplication import CuraApplication
+
+
+##  Abstraction of QNetworkAccessManager for easier networking in Cura.
+#   This was originally part of NetworkedPrinterOutputDevice but was moved out for re-use in other classes.
+class NetworkClient:
+    
+    def __init__(self, application: CuraApplication = None):
```
nallath: Missing return type. It's also not needed to pass the application in the constructor, it's a singleton.

cura/NetworkClient.py
```diff
@@ -0,0 +1,220 @@
...
+    ##  Removes all cached Multi-Part items.
+    def _clearCachedMultiPart(self, reply: QNetworkReply) -> None:
+        if reply in self._kept_alive_multiparts:
+            del self._kept_alive_multiparts[reply]
+
+    ##  Makes sure the network manager is created.
+    def _validateManager(self) -> None:
+        if self._manager is None:
+            self._createNetworkManager()
+        assert (self._manager is not None)
```
nallath: Superfluous ()

nallath: on older versions of Osx (not sure if it's still the case), you need to re-create the networkManager if the computer went into sleep and got out of it again (otherwise it wouldn't be able to send / receive messages at all)

ChrisTerBeke: Ah okay. Will keep it this way then.

DanielSchiavini:
Continues in #4926

ChrisTerBeke:
Closing in favour of https://github.com/Ultimaker/Cura/pull/4926.

PR #4841, opened 2018-11-20
CL-1148 Re-styling monitor queue tiles

ianpaschal:
Part 1 of ? for the Cura 4.0 monitor tab rework!

There's quite a lot of theming that needs to be done but that kind of brainless find and replace work is a lot easier to just do in one fell swoop at the end rather than in waves.


ianpaschal:
Note that no files were deleted, yet. This is because they contain lots of important "how did I solve that last time?" kind of information. That's another thing which I'll take care of at the end once the new design is fully implemented.

ChrisTerBeke:
Shouldn't this point at the 4.0 UI rework branch as base?

ianpaschal:
I was going to make a separate branch but when I asked this to Simon, he said to use master. 🤷‍♂️

The TODOs are because we've broken this into smaller pieces to avoid making one massive PR down the line. It's expected that the monitor tab will be WIP for the next two sprints, however I made sure that everything uses `screenScaleFactor` so that it's 100% looking correct from the front-end.

ianpaschal:
I mean, don't get me wrong, I'm happy to change the base if needed but Simon said it made more sense to put it on master and also get nightly builds automatically. Do you still feel that way @sedwards2009 or should I reconfigure the PR?

plugins/UM3NetworkPrinting/resources/qml/ExpandableCard.qml
```diff
@@ -0,0 +1,82 @@
...
+
+// TODO: Theme & documentation!
+// The expandable component has 3 major sub components:
+//      * The headerItem Always visible and should hold some info about what happens if the component is expanded
+//      * The popupItem The content that needs to be shown if the component is expanded.
+Item
+{
+    id: base
+
+    property var expanded: false
```
nallath: Should be property bool

plugins/UM3NetworkPrinting/resources/qml/ExpandableCard.qml
```diff
@@ -0,0 +1,82 @@
...
+    property var borderWidth: 1
+    property color borderColor: "#EAEAEC"
+    property color headerBackgroundColor: "white"
+    property color headerHoverColor: "#f5f5f5"
+    property color drawerBackgroundColor: "white"
+    property alias headerItem: header.children
```
nallath: Isn't it headerItems if you're setting the children (since it can be a list). Item implies that it can be only a single one.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
```diff
@@ -0,0 +1,179 @@
...
+                        if (printJob !== null) {
+                            if (printJob.assignedPrinter == null)
+                            {
+                                if (printJob.state == "error")
+                                {
+                                    return catalog.i18nc("@label", "Unavailable printer")
+                                }
+                                return catalog.i18nc("@label", "First available")
+                            }
+                            else
```
nallath: This else isn't really needed, because it already returns in all other cases

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterConfiguration.qml
```diff
@@ -0,0 +1,66 @@
...
+ */
+Item
+{
+    id: base
+
+    // Extracted buildplate configuration
+    property alias buildplate: buildplateConfig.buildplate
+
+    // Extracted extruder configuration for position 0
+    property var config0: null
```
nallath: Why not use an array and subsequently use a repeater to draw the components?

plugins/UM3NetworkPrinting/resources/qml/ExpandableCard.qml
ianpaschal: I thought it could be only one (since you'd write `headerItem: Row { ... }` or whatever), and `children` was just the QML nomenclature for that?

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterConfiguration.qml
ianpaschal: Smort.

ianpaschal: I didn't think too much about it since we only have dual extruder network enabled machines, but I'll add it in for future proofness.

plugins/UM3NetworkPrinting/resources/qml/ExpandableCard.qml
ianpaschal: Fixed in 2f84339

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
ianpaschal: Fixed in 1de21c1

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterConfiguration.qml
ianpaschal: Fixed in 55554c6

PR #4842, opened 2018-11-20
Improve the toolbar style

diegopradogesto:
... by modifying the rectangles and the behavior to get a rounded rectangle on the right-top and right-bottom. The hover and active colors also changed.

Contributes to CURA-5962.


nallath:
I've added an element in my branch (RoundedRectangle) that handles this case in a standardized way.

diegopradogesto:
But we will need to increase the functionality of this RoundedRectangle to only have a rounded corner in one of the corners, for the elements of the Toolbar.

PR #4857, opened 2018-11-22
CURA-5879 If Quality profile has 'Experimenat' word in name then quality name w…

LipuFei:
…ill be

used in quality dropdown menu

### DIFF
--- a/cura/Machines/QualityGroup.py
+++ b/cura/Machines/QualityGroup.py
@@ -34,6 +34,9 @@ class QualityGroup(QObject):
     def getName(self) -> str:
         return self.name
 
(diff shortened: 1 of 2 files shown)


ChrisTerBeke:
Wouldn't it be better to parse this from the name (or other fields) when deserializing the files and then store that in a new property on the profile? Then the reading is in the same place as all the other data.

LipuFei:
@ChrisTerBeke We will discuss solutions such as adding an extra field such as `is_experimental = True/False` to the quality profiles.

nallath:
Closing this in favor of https://github.com/Ultimaker/Cura/pull/4883

PR #4863, opened 2018-11-23
Feature: "Raft Material Selection"

vgribinchuk:
Feature: "Raft Material Selection"
This feature supplies the option to select different extruders for the raft top layers, raft middle and raft base. 
New Settings:
"Raft Extruder"("Raft Top Extruder", "Raft Middle Extruder", "Raft Base Extruder")

"Build Plate Adhesion Extruder" will be used for printing the skirt/brim. Separate "Raft Extruder" will be used when 'raft' type of adhesion is turned on.
This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/918

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4632,13 +4632,59 @@
                 "adhesion_extruder_nr":
                 {
                     "label": "Build Plate Adhesion Extruder",
-                    "description": "The extruder train to use for printing the skirt/brim/raft. This is used in multi-extrusion.",
+                    "description": "The extruder train to use for printing the skirt/brim. This is used in multi-extrusion.",
                     "type": "extruder",
                     "default_value": "0",
                     "value": "defaultExtruderPosition()",
-                    "enabled": "extruders_enabled_count > 1 and resolveOrValue('adhesion_type') != 'none'",
+                    "enabled": "extruders_enabled_count > 1 and resolveOrValue('adhesion_type') != 'none' and resolveOrValue('adhesion_type') != 'raft'",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
+                },
+                "raft_extruder_nr": 
+                {
+                    "label": "Raft Extruder",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
We're having some difficulty coming up with a use case for this feature. Why would you want the different layers of your raft to be printed with different extruders?

resources/definitions/fdmprinter.def.json
```diff
@@ -4632,14 +4632,60 @@
...
+                    "description": "The extruder train to use for printing the skirt/brim. This is used in multi-extrusion.",
                     "type": "extruder",
                     "default_value": "0",
                     "value": "defaultExtruderPosition()",
-                    "enabled": "extruders_enabled_count > 1 and resolveOrValue('adhesion_type') != 'none'",
+                    "enabled": "extruders_enabled_count > 1 and resolveOrValue('adhesion_type') != 'none' and resolveOrValue('adhesion_type') != 'raft'",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false
                 },
+                "raft_extruder_nr":
```
Ghostkeeper: You could re-use `adhesion_extruder_nr` for this. There's no reason why it should be a separate setting, I think.

resources/definitions/fdmprinter.def.json
```diff
@@ -4744,9 +4790,8 @@
                     "minimum_value_warning": "raft_interface_line_width",
                     "maximum_value_warning": "20",
                     "enabled": "resolveOrValue('adhesion_type') == 'raft'",
-                    "limit_to_extruder": "adhesion_extruder_nr",
                     "settable_per_mesh": false,
-                    "settable_per_extruder": true
+                    "settable_per_extruder": false
```
Ghostkeeper: These need to be settable per extruder and limited to the adhesion extruder, otherwise a quality, variant or material profile can't determine this line width any more.

nallath:
Closing this since the requested changes have not been made.

PR #4864, opened 2018-11-23
Cura 5943 ui usb printing

nallath:


plugins/MonitorStage/MonitorMainView.qml
```diff
@@ -1,45 +1,61 @@
...
+
+    Loader
+    {
+        id: monitorViewComponent
+
+        anchors.top: parent.top
+        anchors.bottom: parent.bottom
+        anchors.left: parent.left
+
+        width: parent.width * 0.7
```
nallath: Needs rounding.

nallath: Actually; This should be at full width if the sidebar is not set at all.

resources/qml/PrinterOutput/MonitorButtonStyle.qml
```diff
@@ -0,0 +1,88 @@
...
+import QtQuick.Controls.Styles 1.4
+import QtQuick.Layouts 1.3
+
+import UM 1.2 as UM
+import Cura 1.0 as Cura
+
+
+Component
+{
+    ButtonStyle
```
nallath: Styles should be defined in the theme. Also have a look if it's possible to ensure that we don't need a new unique style for these buttons.

plugins/MonitorStage/MonitorMainView.qml
LipuFei: Done in 845b320

resources/qml/PrinterOutput/MonitorButtonStyle.qml
LipuFei: Done in 6bb010e

PR #4884, opened 2018-11-26
Add border to rounded rectangle.

nallath:
It's designed so that it works in exactly the same way as Rectangle.


resources/qml/BorderGroup.qml
```diff
@@ -0,0 +1,7 @@
+import QtQuick 2.0
```
diegopradogesto: Add copyright header.

PR #4888, opened 2018-11-27
CURA-5982 Return empty material model for empty material

LipuFei:
CURA-5982


plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -593,6 +593,15 @@ def _createConfigurationChanges(self, data: List[Dict[str, Any]]) -> List[Config
     def _createMaterialOutputModel(self, material_data: Dict[str, Any]) -> "MaterialOutputModel":
         material_manager = CuraApplication.getInstance().getMaterialManager()
         material_group_list = material_manager.getMaterialGroupListByGUID(material_data["guid"])
+        # This can happen if the connected machine has no material in one or more extruders, so we should return an
+        # "empty" material model.
+        if material_group_list is None:
```
nallath: This will actually ensure that if a material is known by Cura (but is known and loaded on the machine) that it's labeled as "Empty", which is incorrect.

LipuFei: Done in 4ff5e43a2

PR #4895, opened 2018-11-27
Added definitions for Wanhao Duplicator 9

matshch:
We have Wanhao Duplicator 9 printer, and using Cura provided by Wanhao. However, it is quite outdated. I've ported definitions from their version to new Cura, and looks like it works perfectly for us.


diegopradogesto:
Devs. CURA-6105

PR #4905, opened 2018-11-29
Feature: "Build Interface If Support Is Present"

vgribinchuk:
Feature: "Build Interface If Support Is Present"
Some support could be pushed away by large support XY distance, but interface could be left on overhangs, which leads to situation when interface is generated without support. Please refer the pic below.
interfacewithoutsupport
If this feature is turned on, then interface will not be generated without support under it.
New setting:
"Build Interface If Support Is Present"
This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/923

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4480,16 +4480,6 @@
                         }
                     }
                 },
-                "build_interface_if_support_present": 
-                {
-                    "label": "Build Interface If Support Is Present",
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
I believe this is actually a bug in support generation and should be fixed regardless of having a new setting turned on or off.

diegopradogesto:
Devs. CURA-6106

diegopradogesto:
If this PR is not needed, you can close it.

PR #4925, opened 2018-12-03
CL-1153 printer states in monitor

ianpaschal:
This brings the progress bar and printer state in line with the Cura 4.0 redesign

### DIFF
--- a/cura/PrinterOutput/PrintJobOutputModel.py
+++ b/cura/PrinterOutput/PrintJobOutputModel.py
@@ -118,11 +118,11 @@ class PrintJobOutputModel(QObject):
             self.nameChanged.emit()
 
     @pyqtProperty(int, notify = timeTotalChanged)
-    def timeTotal(self):
+    def timeTotal(self) -> int:
         return self._time_total
 
     @pyqtProperty(int, notify = timeElapsedChanged)
-    def timeElapsed(self):
+    def timeElapsed(self) -> int:
         return self._time_elapsed
 
     @pyqtProperty(int, notify = timeElapsedChanged)
@@ -138,7 +138,7 @@ class PrintJobOutputModel(QObject):
         return result
 
     @pyqtProperty(str, notify=stateChanged)
-    def state(self):
+    def state(self) -> str:
         return self._state
 
     @pyqtProperty(bool, notify=stateChanged)
--- a/cura/PrinterOutput/PrintJobOutputModel.py
+++ b/cura/PrinterOutput/PrintJobOutputModel.py
@@ -133,9 +133,8 @@ class PrintJobOutputModel(QObject):
     @pyqtProperty(float, notify = timeElapsedChanged)
     def progress(self) -> float:
         result = self.timeElapsed / self.timeTotal
-        if result > 1.0:
-            result = 1.0
-        return result
+        # Never get a progress past 1.0
+        return min(result, 1.0)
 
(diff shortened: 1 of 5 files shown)


cura/PrinterOutput/PrintJobOutputModel.py
```diff
@@ -118,17 +118,41 @@ def updateName(self, name: str):
...
+    def timeRemaining(self) -> int:
+        # Never get a negative time remaining
+        return max(self.timeTotal - self.timeElapsed, 0)
+
+    @pyqtProperty(float, notify = timeElapsedChanged)
+    def progress(self) -> float:
+        result = self.timeElapsed / self.timeTotal
+        if result > 1.0:
+            result = 1.0
+        return result
```
sedwards2009: It is not a  big deal, but it interesting to see `max()` used above but `min()` isn't being used here.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobPreview.qml
```diff
@@ -47,11 +54,32 @@ Item
...
+                case "pausing":
+                    return "../svg/paused-icon.svg"
+                case "paused":
+                    return "../svg/paused-icon.svg"
+                case "resuming":
+                    return "../svg/paused-icon.svg"
+                default:
+                    return ""
+            }
+        }
```
sedwards2009: This function can return `undefined` when it should probably return something else like "".

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobProgressBar.qml
```diff
@@ -82,54 +29,85 @@ Item
...
+                case "pausing":
+                    return catalog.i18nc("@label:status", "Pausing...")
+                case "paused":
+                    return catalog.i18nc("@label:status", "Paused")
+                case "resuming":
+                    return catalog.i18nc("@label:status", "Resuming...")
+                case "queued":
+                    return catalog.i18nc("@label:status", "Action required")
+                default:
+                    return catalog.i18nc("@label:status", "Finishes ") + OutputDevice.getDateCompleted( printJob.timeRemaining ) + " at " + OutputDevice.getTimeCompleted( printJob.timeRemaining )
```
sedwards2009: This string building needs to be replaced with 1 translatable string/template.

cura/PrinterOutput/PrintJobOutputModel.py
ianpaschal: Good point. Don't know why I didn't bother with that. Updated.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobPreview.qml
ianpaschal: Ummmm can it? The switch is the first block and it has a default case at the end to ensure it always returns a string.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobProgressBar.qml
sedwards2009: Maybe it is now time to move some of stuff to the Python side? The Python code is actually full of i18n strings too.

sedwards2009:
The Windows CI server is borked and no one is around to fix the build slave thingy.

PR #4926, opened 2018-12-03
[4.0] STAR-322: Cloud connection

DanielSchiavini:
This PR add remote monitoring and printing via Cura Connect using the Ultimaker Cloud. By signing in to Cura with your Ultimaker account, and then doing the same in Cura Connect, you can remotely see your printers, print jobs and add new jobs to the queue. More functionality will be added in later releases.

Updating the monitor page QML will be done in a separate PR by @ianpaschal.


cura/NetworkClient.py
```diff
@@ -0,0 +1,219 @@
...
+        
+        if on_progress is not None:
+            reply.uploadProgress.connect(on_progress)
+            
+        self._registerOnFinishedCallback(reply, on_finished)
+        
+    ##  Does a POST request with form data to the given URL.
+    def postForm(self, url: str, header_data: str, body_data: bytes,
+                 on_finished: Optional[Callable[[QNetworkReply], None]],
+                 on_progress: Callable = None) -> None:
```
nallath:
```suggestion
                 on_progress: Optional[Callable[[int, int], None]] = None) -> None:
```

cura/NetworkClient.py
```diff
@@ -0,0 +1,219 @@
...
+        if not self._manager:
+            Logger.log("e", "No network manager was created to execute the GET call with.")
+            return
+        
+        reply = self._manager.get(request)
+        self._registerOnFinishedCallback(reply, on_finished)
+
+    ##  Does a POST request to the given URL.
+    def post(self, url: str, data: str, on_finished: Optional[Callable[[QNetworkReply], None]],
+             on_progress: Callable = None) -> None:
```
nallath:
```suggestion
             on_progress: Optional[Callable[[int, int], None]]) -> None:
```

cura/NetworkClient.py
```diff
@@ -0,0 +1,219 @@
...
+                 on_progress: Callable = None) -> None:
+        post_part = QHttpPart()
+        post_part.setHeader(QNetworkRequest.ContentDispositionHeader, header_data)
+        post_part.setBody(body_data)
+        self.postFormWithParts(url, [post_part], on_finished, on_progress)
+
+    ##  Does a POST request with form parts to the given URL.
+    def postFormWithParts(self, target: str, parts: List[QHttpPart],
+                          on_finished: Optional[Callable[[QNetworkReply], None]],
+                          on_progress: Callable = None) -> Optional[QNetworkReply]:
```
nallath:
```suggestion
                          on_progress: Optional[Callable[[int, int], None]] = None) -> Optional[QNetworkReply]:
```

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -181,13 +180,17 @@ def _validateManager(self) -> None:
             self._createNetworkManager()
         assert (self._manager is not None)
 
-    def put(self, target: str, data: str, on_finished: Optional[Callable[[QNetworkReply], None]]) -> None:
+    def put(self, target: str, data: Union[str, bytes], content_type: str = None,
+            on_finished: Optional[Callable[[QNetworkReply], None]] = None,
+            on_progress: Optional[Callable] = None) -> None:
```
nallath:
```suggestion
            on_progress: Optional[Callable[[int, int], None]] = None) -> None:
```

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -211,12 +214,13 @@ def get(self, target: str, on_finished: Optional[Callable[[QNetworkReply], None]
         else:
             Logger.log("e", "Could not find manager.")
 
-    def post(self, target: str, data: str, on_finished: Optional[Callable[[QNetworkReply], None]], on_progress: Callable = None) -> None:
+    def post(self, target: str, data: Union[str, bytes], on_finished: Optional[Callable[[QNetworkReply], None]],
+             on_progress: Callable = None) -> None:
```
nallath:
```suggestion
             on_progress: Optional[Callable[[int, int], None]] = None) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputController.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2018 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+from cura.PrinterOutput.PrinterOutputController import PrinterOutputController
+
+
+class CloudOutputController(PrinterOutputController):
+    def __init__(self, output_device):
```
nallath:
```suggestion
    def __init__(self, output_device: OutputDevice) -> None:
```
Don't forget to also add the import for the output device

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+from UM.Message import Message
+from UM.OutputDevice import OutputDeviceError
+from UM.Scene.SceneNode import SceneNode
+from UM.Version import Version
+from cura.CuraApplication import CuraApplication
+from cura.PrinterOutput.PrinterOutputController import PrinterOutputController
+from cura.PrinterOutput.MaterialOutputModel import MaterialOutputModel
+from cura.PrinterOutput.NetworkedPrinterOutputDevice import NetworkedPrinterOutputDevice, AuthState
+from cura.PrinterOutput.PrinterOutputModel import PrinterOutputModel
+from plugins.UM3NetworkPrinting.src.UM3PrintJobOutputModel import UM3PrintJobOutputModel
```
nallath: In case of plugins, import should be relative

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+    CLUSTER_API_ROOT = "{}/connect/v1/".format(ROOT_PATH)
+    CURA_API_ROOT = "{}/cura/v1/".format(ROOT_PATH)
+
+    # Signal triggered when the printers in the remote cluster were changed.
+    printersChanged = pyqtSignal()
+
+    # Signal triggered when the print jobs in the queue were changed.
+    printJobsChanged = pyqtSignal()
+
+    def __init__(self, device_id: str, parent: QObject = None):
```
nallath:
```suggestion
    def __init__(self, device_id: str, parent: QObject = None) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+        self._print_jobs = []  # type: List[UM3PrintJobOutputModel]
+        self._number_of_extruders = 2  # All networked printers are dual-extrusion Ultimaker machines.
+        
+        # We only allow a single upload at a time.
+        self._sending_job = False
+        self._progress_message = None  # type: Optional[Message]
+
+    @staticmethod
+    def _parseReply(reply: QNetworkReply) -> Tuple[int, Union[None, str, bytes]]:
+        """
```
nallath: Should be doxygen.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+            Logger.log("e", "Missing file or mesh writer!")
+            self._onUploadError(self.I18N_CATALOG.i18nc("@info:status", "Could not export print job."))
+            return
+
+        stream = io.StringIO() if file_format["mode"] == FileWriter.OutputMode.TextMode else io.BytesIO()
+        writer.write(stream, nodes)
+        self._sendPrintJob(file_name + "." + file_format["extension"], file_format["mime_type"], stream)
+
+    # TODO: This is yanked right out of ClusterUM3OutputDevice, great candidate for a utility or base class
+    def _determineFileFormat(self, file_handler) -> Optional[Dict[str, Union[str, int]]]:
```
nallath: This entire function should be removed. The cloud should always support UFP. The only reason this is in the cloud plugin is because there are older firmware versions that don't support this.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+        if len(file_formats) == 0:
+            Logger.log("e", "There are no file formats available to write with!")
+            raise OutputDeviceError.WriteRequestFailedError(
+                self.I18N_CATALOG.i18nc("@info:status", "There are no file formats available to write with!")
+            )
+        return file_formats[0]
+
+    # TODO: This is yanked right out of ClusterUM3OutputDevice, great candidate for a utility or base class
+    @staticmethod
+    def _determineWriter(file_handler, file_format) -> Optional[FileWriter]:
```
nallath: Same as above; Can just be removed

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+)
+
+
+##  The cloud output device is a network output device that works remotely but has limited functionality.
+#   Currently it only supports viewing the printer and print job status and adding a new job to the queue.
+#   As such, those methods have been implemented here.
+#   Note that this device represents a single remote cluster, not a list of multiple clusters.
+#
+#   TODO: figure our how the QML interface for the cluster networking should operate with this limited functionality.
+class CloudOutputDevice(NetworkedPrinterOutputDevice):
```
nallath: Why not make this inherit from the Cluster UM3 output and change the things you actually want different? Right now there is quite a bit of code duplication.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+        if status_code > 204 or not isinstance(response, dict) or "data" not in response:
+            Logger.log("w", "Got unexpected response while trying to request printing: %s, %s", status_code, response)
+            self._onUploadError(self.I18N_CATALOG.i18nc("@info:status", "Could not add print job to queue."))
+            return
+
+        print_response = PrintResponse(**response["data"])
+        Logger.log("i", "Print job requested successfully: %s", print_response.__dict__)
+        self._onUploadSuccess()
+
+    def _updateUploadProgress(self, progress: int):
```
nallath:
```suggestion
    def _updateUploadProgress(self, progress: int) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+                title = self.I18N_CATALOG.i18nc("@info:title", "Sending Data..."),
+                progress = -1,
+                lifetime = 0,
+                dismissable = False,
+                use_inactivity_timer = False
+            )
+        self._progress_message.setProgress(progress)
+        self._progress_message.show()
+
+    def _resetUploadProgress(self):
```
nallath:
```suggestion
    def _resetUploadProgress(self) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+        if self._progress_message:
+            self._progress_message.hide()
+            self._progress_message = None
+
+    def _onUploadError(self, message: str = None):
```
nallath:
```suggestion
    def _onUploadError(self, message: Optional[str] = None) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+                text = message,
+                title = self.I18N_CATALOG.i18nc("@info:title", "Error"),
+                lifetime = 10,
+                dismissable = True
+            )
+            message.show()
+        self._sending_job = False  # the upload has failed so we're not sending a job anymore
+        self.writeError.emit()
+
+    def _onUploadSuccess(self):
```
nallath:
```suggestion
    def _onUploadSuccess(self) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,483 @@
...
+        self._resetUploadProgress()
+        message = Message(
+            text = self.I18N_CATALOG.i18nc("@info:status", "Print job was successfully sent to the printer."),
+            title = self.I18N_CATALOG.i18nc("@info:title", "Data Sent"),
+            lifetime = 5,
+            dismissable = True,
```
nallath: That's the default value.

ChrisTerBeke:
@nallath thanks for the review, note that we're still going to refactor a lot of this after we have something working (mainly the networking stuff as well).

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -0,0 +1,142 @@
...
+#
+class CloudOutputDeviceManager(NetworkClient):
+    
+    # The cloud URL to use for remote clusters.
+    API_ROOT_PATH = "https://api-staging.ultimaker.com/connect/v1"
+
+    # The interval with which the remote clusters are checked
+    CHECK_CLUSTER_INTERVAL = 5  # seconds
+    
+    def __init__(self):
```
nallath:
```suggestion
    def __init__(self) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
ChrisTerBeke: We're still figuring out whether to do that or not depending on how much we need to override if we use cluster output device as base.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -0,0 +1,142 @@
...
+
+    ##  Gets all remote clusters from the API.
+    def _getRemoteClusters(self) -> None:
+        Logger.log("i", "Retrieving remote clusters")
+        if self._account.isLoggedIn:
+            self.get("/clusters", on_finished = self._onGetRemoteClustersFinished)
+
+        # Only start the polling thread after the user is authenticated
+        # The first call to _getRemoteClusters comes from self._account.loginStateChanged
+        timer = Timer(5.0, self._on_cluster_received.emit)
```
nallath: Don't use Python timers. They are horribly unrealiable and also have the nice "habit" of messing up threads / event queue from qt. It's better to use QTimers.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -0,0 +1,142 @@
...
+        request = super()._createEmptyRequest(self.API_ROOT_PATH + path, content_type = content_type)
+        if self._account.isLoggedIn:
+            # TODO: add correct scopes to OAuth2 client to use remote connect API.
+            # TODO: don't create the client when not signed in?
+            request.setRawHeader(b"Authorization", "Bearer {}".format(self._account.accessToken).encode())
+        return request
+
+    ##  Gets all remote clusters from the API.
+    def _getRemoteClusters(self) -> None:
+        Logger.log("i", "Retrieving remote clusters")
```
nallath: This is going spam the log (since it's logging this every 5 seconds)

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -0,0 +1,142 @@
...
+        except UnicodeDecodeError:
+            Logger.log("w", "Unable to read server response")
+        except json.decoder.JSONDecodeError:
+            Logger.logException("w", "Unable to decode JSON from reply.")
+        except ValueError:
+            Logger.logException("w", "Response was missing values.")
+        return {}
+
+    ##  Adds a CloudOutputDevice for each entry in the remote cluster list from the API.
+    def _addCloudOutputDevice(self, cluster: CloudCluster):
```
nallath:
```suggestion
    def _addCloudOutputDevice(self, cluster: CloudCluster) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -0,0 +1,142 @@
...
+        device = CloudOutputDevice(cluster.cluster_id)
+        self._output_device_manager.addOutputDevice(device)
+        self._remote_clusters[cluster.cluster_id] = device
+        device.connect()  # TODO: remove this
+        self._connectToActiveMachine()
+
+    ##  Remove a CloudOutputDevice
+    def _removeCloudOutputDevice(self, cluster: CloudCluster):
```
nallath:
```suggestion
    def _removeCloudOutputDevice(self, cluster: CloudCluster) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
ChrisTerBeke: Yup, but we need it for debugging right now.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
nallath: Fair enough. It's not that big of an issue, but i want to prevent having duplicated code as much as possible. We might want to add an extra level of abstraction though (eg; move the notion that there are multiple printers to a shared cloud / cura connect parent)

ChrisTerBeke: Yup, that's part actually of what we're refactoring right now :)

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -50,7 +48,7 @@ class ClusterUM3OutputDevice(BaseCuraConnectDevice):
 
     # This is a bit of a hack, as the notify can only use signals that are defined by the class that they are in.
     # Inheritance doesn't seem to work. Tying them together does work, but i'm open for better suggestions.
-    clusterPrintersChanged = pyqtSignal()
+    _clusterPrintersChanged = pyqtSignal()
```
ChrisTerBeke: Not sure if this should be private, since QML can listen to it as well.

DanielSchiavini: _clusterPrintersChanged gets notified when things change, but it's only called in
`self.printersChanged.connect(self._clusterPrintersChanged)`
so the signal to be used externally should be printersChanged which is in a base class

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -41,31 +41,34 @@ def __init__(self):
...
+        self._update_timer.setInterval(int(self.CHECK_CLUSTER_INTERVAL * 1000))
         self._update_timer.setSingleShot(False)
         self._update_timer.timeout.connect(self._getRemoteClusters)
 
+        # Make sure the timer is started in case we missed the loginChanged signal
+        self._onLoginStateChanged()
+
+    #  Called when the uses logs in or out
+    def _onLoginStateChanged(self) -> None:
+        if self._account.isLoggedIn and not self._update_timer.isActive():
```
DanielSchiavini: if the user is logged in and the timer is active we shouldn't remove all clusters

marijnd: That should not happen... but you are right.

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -0,0 +1,154 @@
...
+from .Models.CloudPrintResponse import CloudPrintResponse
+from .Models.CloudPrintJobResponse import CloudPrintJobResponse
+
+
+## The cloud API client is responsible for handling the requests and responses from the cloud.
+#  Each method should only handle models instead of exposing any HTTP details.
+class CloudApiClient(NetworkClient):
+
+    # The cloud URL to use for this remote cluster.
+    # TODO: Make sure that this URL goes to the live api before release
```
ChrisTerBeke: We can use the new CuraConstants file that @LipuFei is adding in another PR (this issue was also there for the backups plugin that was moved to the Cura repository).

plugins/CuraDrive/src/DriveApiService.py
```diff
@@ -0,0 +1,185 @@
...
+    PUT_BACKUP_URL = "{}/backups".format(Settings.DRIVE_API_URL)
+    DELETE_BACKUP_URL = "{}/backups".format(Settings.DRIVE_API_URL)
+
+    # Emit signal when restoring backup started or finished.
+    onRestoringStateChanged = Signal()
+
+    # Emit signal when creating backup started or finished.
+    onCreatingStateChanged = Signal()
+
+    def __init__(self, cura_api) -> None:
```
nallath: Missing typing

plugins/CuraDrive/src/DrivePluginExtension.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # Signal emitted when creating has started. Needed to prevent parallel creation of backups.
+    creatingStateChanged = pyqtSignal()
+
+    # Signal emitted when preferences changed (like auto-backup).
+    preferencesChanged = pyqtSignal()
+    
+    DATE_FORMAT = "%d/%m/%Y %H:%M:%S"
+
+    def __init__(self, application):
```
nallath: Missing typing

plugins/CuraDrive/src/models/BackupListModel.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2018 Ultimaker B.V.
+from typing import Any, List, Dict
+
+from UM.Qt.ListModel import ListModel
+
+from PyQt5.QtCore import Qt
+
+
+class BackupListModel(ListModel):
```
nallath: This should be added in a separate pull request.

plugins/CuraDrive/src/qml/components/ActionButton.qml
```diff
@@ -0,0 +1,67 @@
...
+import QtQuick.Controls 2.1
+import QtQuick.Layouts 1.3
+
+import UM 1.1 as UM
+
+Button
+{
+    id: button
+    property alias cursorShape: mouseArea.cursorShape
+    property var iconSource: ""
```
nallath: Missing types.

plugins/CuraDrive/src/DriveApiService.py
ChrisTerBeke: All the cura drive related stuff should be fixed in that other PR (and we'll merge it in here then). Reason it's here is explained in the PR description.

ChrisTerBeke:
The codestyle checks are failing, but in a file that wasn't touched by us. Weirdly enough it doesn't seem to be failing on the 4.0 branch?

cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -181,49 +192,85 @@ def _validateManager(self) -> None:
...
         self._last_request_time = time()
-        if self._manager is not None:
-            reply = self._manager.deleteResource(request)
-            self._registerOnFinishedCallback(reply, on_finished)
-        else:
-            Logger.log("e", "Could not find manager.")
 
-    def get(self, target: str, on_finished: Optional[Callable[[QNetworkReply], None]]) -> None:
+        if not self._manager:
+            return Logger.log("e", "No network manager was created to execute the DELETE call with.")
```
nallath:
```suggestion
            Logger.log("e", "No network manager was created to execute the DELETE call with.")
            return
```

cura/PrintersModel.py
```diff
@@ -53,9 +53,8 @@ def _update(self) -> None:
         container_stacks = ContainerRegistry.getInstance().findContainerStacks(type = "machine")
 
         for container_stack in container_stacks:
-            connection_type = container_stack.getMetaDataEntry("connection_type")
-            has_remote_connection = connection_type in [ConnectionType.NetworkConnection.value, ConnectionType.CloudConnection.value]
-
+            connection_type = container_stack.getMetaDataEntry("connection_type", 0)
```
nallath: This has been fixed in 4.0 branch already by defaulting to ConnectionType.NotConnected (which is also 0, but is a bit clearer)

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,431 @@
...
+from .Models.CloudPrintJobUploadRequest import CloudPrintJobUploadRequest
+from .Models.CloudPrintResponse import CloudPrintResponse
+from .Models.CloudPrintJobResponse import CloudPrintJobResponse
+from .Models.CloudClusterPrinterStatus import CloudClusterPrinterStatus
+from .Models.CloudClusterPrintJobStatus import CloudClusterPrintJobStatus
+from .Utils import findChanges, formatDateCompleted, formatTimeCompleted
+
+
+## Class that contains all the translations for this module.
+class T:
```
nallath: Please rename this to something else. T is often used for abstract classes within python (and it's also extremely not verbose as to what it is). It should also be in its own file.

ChrisTerBeke:
I'll merge 4.0 into this next week.

resources/qml/PrinterSelector/MachineSelector.qml
```diff
@@ -29,15 +29,13 @@ Cura.ExpandablePopup
         text: isNetworkPrinter ? Cura.MachineManager.activeMachineNetworkGroupName : Cura.MachineManager.activeMachineName
         source:
         {
-            if (isNetworkPrinter)
-            {
-                if (machineSelector.outputDevice != null && machineSelector.outputDevice.clusterSize > 1)
-                {
-                    return UM.Theme.getIcon("printer_group")
-                }
+            if (isGroup) {
```
nallath:
```suggestion
            if (isGroup) 
            {
```

resources/qml/PrinterSelector/MachineSelector.qml
```diff
@@ -52,12 +50,22 @@ Cura.ExpandablePopup
                 leftMargin: UM.Theme.getSize("thick_margin").width
             }
 
-            source: UM.Theme.getIcon("printer_connected")
+            source:
+            {
+                if (isNetworkPrinter) {
```
nallath: Codestyle; { go on the next line.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,424 @@
...
+        self._progress.hide()
+        Message(
+            text = I18N_CATALOG.i18nc("@info:status", "Print job was successfully sent to the printer."),
+            title = I18N_CATALOG.i18nc("@info:title", "Data Sent"),
+            lifetime = 5
+        ).show()
+        self.writeFinished.emit()
+
+    ##  Gets the remote printers.
+    @pyqtProperty("QVariantList", notify=_clusterPrintersChanged)
```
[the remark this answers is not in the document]
ChrisTerBeke: Not sure why Git is showing this, it's not in my local branch anymore :P

plugins/UM3NetworkPrinting/tests/Cloud/NetworkManagerMock.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Dict, Tuple, Union, Optional, Any
+from unittest.mock import MagicMock
+
+from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkRequest
+
+from UM.Logger import Logger
+from UM.Signal import Signal
+
+
+class FakeSignal:
```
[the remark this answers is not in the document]
ChrisTerBeke: I'll wait for Daniel to come back to discuss all the testing related stuff, as I'm not sure what some of his intentions were.

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # The cloud URL to use for this remote cluster.
+    ROOT_PATH = UltimakerCloudAuthentication.CuraCloudAPIRoot
+    CLUSTER_API_ROOT = "{}/connect/v1".format(ROOT_PATH)
+    CURA_API_ROOT = "{}/cura/v1".format(ROOT_PATH)
+
+    ## Initializes a new cloud API client.
+    #  \param account: The user's account object
+    #  \param on_error: The callback to be called whenever we receive errors from the server.
+    def __init__(self, account: Account, on_error: Callable[[List[CloudError]], None]) -> None:
```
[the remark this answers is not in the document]
ChrisTerBeke: Probably to have the API client be an isolated part of the code (e.g. makes it easier to inject a fake account object during testing).

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -0,0 +1,168 @@
...
+    def requestUpload(self, request: CloudPrintJobUploadRequest, on_finished: Callable[[CloudPrintJobResponse], Any]
+                      ) -> None:
+        url = "{}/jobs/upload".format(self.CURA_API_ROOT)
+        body = json.dumps({"data": request.toDict()})
+        reply = self._manager.put(self._createEmptyRequest(url), body.encode())
+        self._addCallback(reply, on_finished, CloudPrintJobResponse)
+
+    ## Uploads a print job mesh to the cloud.
+    #  \param print_job: The object received after requesting an upload with `self.requestUpload`.
+    #  \param mesh: The mesh data to be uploaded.
```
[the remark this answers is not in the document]
ChrisTerBeke: It's actually G-code even :P I'll fix the comment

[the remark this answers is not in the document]
ChrisTerBeke: And the method name (uploadToolPath).

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -0,0 +1,168 @@
...
+        url = "{}/clusters".format(self.CLUSTER_API_ROOT)
+        reply = self._manager.get(self._createEmptyRequest(url))
+        self._addCallback(reply, on_finished, CloudClusterResponse)
+
+    ## Retrieves the status of the given cluster.
+    #  \param cluster_id: The ID of the cluster.
+    #  \param on_finished: The function to be called after the result is parsed.
+    def getClusterStatus(self, cluster_id: str, on_finished: Callable[[CloudClusterStatus], Any]) -> None:
+        url = "{}/clusters/{}/status".format(self.CLUSTER_API_ROOT, cluster_id)
+        reply = self._manager.get(self._createEmptyRequest(url))
```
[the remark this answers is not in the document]
ChrisTerBeke: Because networkedPrinterOutput device does more stuff that we don't want to happen when talking to the cloud (e.g. http headers are different).

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -0,0 +1,168 @@
...
+    #  \param on_finished: The function to be called after the result is parsed.
+    def getClusters(self, on_finished: Callable[[List[CloudClusterResponse]], Any]) -> None:
+        url = "{}/clusters".format(self.CLUSTER_API_ROOT)
+        reply = self._manager.get(self._createEmptyRequest(url))
+        self._addCallback(reply, on_finished, CloudClusterResponse)
+
+    ## Retrieves the status of the given cluster.
+    #  \param cluster_id: The ID of the cluster.
+    #  \param on_finished: The function to be called after the result is parsed.
+    def getClusterStatus(self, cluster_id: str, on_finished: Callable[[CloudClusterStatus], Any]) -> None:
```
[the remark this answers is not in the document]
ChrisTerBeke: To keep the coding pattern the same for all network requests.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,424 @@
...
+    #  \param job: The print job received.
+    def _addPrintJob(self, job: CloudClusterPrintJobStatus) -> None:
+        model = job.createOutputModel(CloudOutputController(self))
+        model.stateChanged.connect(self._onPrintJobStateChanged)
+        if job.printer_uuid:
+            self._updateAssignedPrinter(model, job.printer_uuid)
+        self._print_jobs.append(model)
+
+    ## Handles the event of a change in a print job state
+    def _onPrintJobStateChanged(self) -> None:
```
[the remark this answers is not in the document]
ChrisTerBeke: I'm not aware of that, so probably needs input from CC team?

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrintCoreConfiguration.py
```diff
@@ -0,0 +1,52 @@
...
+##  Class representing a cloud cluster printer configuration
+#  Spec: https://api-staging.ultimaker.com/connect/v1/spec
+class CloudClusterPrintCoreConfiguration(BaseCloudModel):
+    ## Creates a new cloud cluster printer configuration object
+    #  \param extruder_index: The position of the extruder on the machine as list index. Numbered from left to right.
+    #  \param material: The material of a configuration object in a cluster printer. May be in a dict or an object.
+    #  \param nozzle_diameter: The diameter of the print core at this position in millimeters, e.g. '0.4'.
+    #  \param print_core_id: The type of print core inserted at this position, e.g. 'AA 0.4'.
+    def __init__(self, extruder_index: int,
+                 material: Union[None, Dict[str, Any], CloudClusterPrinterConfigurationMaterial],
```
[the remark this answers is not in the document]
ChrisTerBeke: Well, we're not sure what values are actually in those, so we made it an Any.

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -449,8 +389,9 @@ def _printJobStateChanged(self) -> None:
 
     ##  Called when the connection to the cluster changes.
     def connect(self) -> None:
-        super().connect()
-        self.sendMaterialProfiles()
+        # super().connect()
```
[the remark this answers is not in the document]
ChrisTerBeke: For testing, needs to be reverted before merging.

ChrisTerBeke:
@nallath I have fixed all the code style, typo's and nitpicks, as well as some easier logic changes. Let's see next week how we can fix the code duplication by making a Cluster base class for both local and remote networking.

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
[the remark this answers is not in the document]
nallath: As far as I can see, it's about as easy to inject it though.

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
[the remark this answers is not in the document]
nallath: In that case you can just override the _createEmptyRequest instead of re-implementing parts. This just contributes to code duplication.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,424 @@
...
+            self._active_printer = printer
+            self.activePrinterChanged.emit()
+
+    @pyqtProperty(int, notify = _clusterPrintersChanged)
+    def clusterSize(self) -> int:
+        return len(self._printers)
+
+    ##  Get remote print jobs.
+    @pyqtProperty("QVariantList", notify = printJobsChanged)
+    def printJobs(self) -> List[UM3PrintJobOutputModel]:
```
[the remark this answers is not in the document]
nallath: Well, since the beta is delayed by a bit, I guess we do now.

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrintCoreConfiguration.py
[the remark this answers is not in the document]
nallath: That's why I'm saying if it's the metadata of the material, it's str ;)

plugins/UM3NetworkPrinting/src/Cloud/MeshUploader.py
```diff
@@ -0,0 +1,148 @@
...
+        if status_code > 308:
+            self._errorCallback()
+            return
+
+        Logger.log("d", "status_code: %s, Headers: %s, body: %s", status_code,
+                   [bytes(header).decode() for header in reply.rawHeaderList()], bytes(reply.readAll()).decode())
+        self._chunkUploaded()
+
+    ## Handles a chunk of data being uploaded, starting the next chunk if needed.
+    def _chunkUploaded(self) -> None:
```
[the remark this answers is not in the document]
nallath: Large being? Because these issues didn't occur with the cluster. This implementation can still cause issues with the GIL. 
Did you also check if the network manager from QT doesn't already support this? 
Also, if this is the reason; please document this. Because if I would have seen this at a later stage, I would have removed this in a re-factor.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -0,0 +1,424 @@
...
+        # A set of the user's job IDs that have finished
+        self._finished_jobs = set()  # type: Set[str]
+
+        # Reference to the uploaded print job / mesh
+        self._tool_path = None  # type: Optional[bytes]
+        self._uploaded_print_job = None  # type: Optional[CloudPrintJobResponse]
+
+    ## Connects this device.
+    def connect(self) -> None:
+        if self.isConnected():
```
nallath: Isn't this something that should live in the base class? This isn't just specific to the CloudOutputDevice.

ChrisTerBeke: Well, that would change behaviour for other output devices and I'm not sure of the impact of that.

ChrisTerBeke:
@DanielSchiavini most of the tests now fail, but in Cura it actually works. Not sure what we messed up, could you take a look if you find the time?

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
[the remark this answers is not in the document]
ChrisTerBeke: I asked them to have a look at this.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputController.py
```diff
@@ -1,11 +1,11 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
+from UM.OutputDevice.OutputDevice import OutputDevice
 from cura.PrinterOutput.PrinterOutputController import PrinterOutputController
-from .CloudOutputDevice import CloudOutputDevice
 
 
 class CloudOutputController(PrinterOutputController):
-    def __init__(self, output_device: CloudOutputDevice) -> None:
+    def __init__(self, output_device: "OutputDevice") -> None:
```
nallath: A much better way of fixing this is:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .CloudOutputDevice import CloudOutputDevice

def __init__(self, output_device: "CloudOutputDevice") -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudProgressMessage.py
```diff
@@ -19,6 +19,11 @@ def __init__(self):
             use_inactivity_timer = False
         )
 
+    ## Returns a boolean indicating whether this message is currently visible
+    @property
```
ChrisTerBeke: This was moved to the base class, I'll remove it again here.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -101,7 +101,7 @@ def _connectToActiveMachine(self) -> None:
         if not active_machine:
             return
 
-        # Remove all output devices that we have registered.
+        # Remove all output devices that we have registered. TODO: Why??
```
ChrisTerBeke: I'll update the text why this is needed.

Appesteijn:
Tested, works correctly

ianpaschal:
🍾

PR #4959, opened 2018-12-10
CL-1152 config change monitor tab

ianpaschal:


plugins/UM3NetworkPrinting/resources/qml/MonitorConfigOverrideDialog.qml
```diff
@@ -0,0 +1,131 @@
...
+import UM 1.3 as UM
+
+UM.Dialog
+{
+    id: overrideConfirmationDialog
+
+    property var printer: null
+
+    minimumWidth: screenScaleFactor * 640;
+    minimumHeight: screenScaleFactor * 320;
```
sedwards2009: It is hard to kick the semi-colon habit, I know.

plugins/UM3NetworkPrinting/resources/qml/MonitorConfigOverrideDialog.qml
```diff
@@ -0,0 +1,133 @@
...
+                translationText = catalog.i18nc("@label", "Glass");
+                break;
+            case "aluminum":
+                translationText = catalog.i18nc("@label", "Aluminum");
+                break;
+            default:
+                translationText = null;
+        }
+        return translationText;
+    }
```
sedwards2009: Should we also change the codestyle for these JS functions above?

plugins/UM3NetworkPrinting/resources/qml/MonitorConfigOverrideDialog.qml
ianpaschal: Actually this was copy pasta'd from somewhere else because... you know... we totes use ; all over Cura 🙄

PR #4965, opened 2018-12-11
agile_extruder_0.def.json

robintiger:


LipuFei:
There's already definition files for `ZYYX Agile` in master. See https://github.com/Ultimaker/Cura/blob/master/resources/extruders/zyyx_agile_extruder_0.def.json

I don't think this PR is still needed.

PR #4971, opened 2018-12-11
CL-1173 Use short printer family tags

ianpaschal:
> _Ian, why are you putting this kind of hack into our code base?_

Well I'll tell you: Because Cura Connect uses the printer name/type all over the place so until we have a new design for CC to match Cura 4.0, this was _actually_ a case where the quick solution was also the cleaner one. We can maybe think about adding a "tag" to Cura Connect but then it has to be treated differently than all other printers which use the `NetworkedPrinterOutputModel`.

Contributes to CL-1173.

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterPill.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterPill.qml
@@ -12,7 +12,19 @@ import UM 1.2 as UM
 Item
 {
     // The printer name
-    property alias text: printerNameLabel.text;
+    property var text: ""
+    property var tagText: {
+        switch(text) {
+            case "Ultimaker 3":
+                return "UM 3"
+            case "Ultimaker 3 Extended":
+                return "UM 3 EXT"
+            case "Ultimaker S5":
+                return "UM S5"
+            default:
+                return ""
+        }
+    }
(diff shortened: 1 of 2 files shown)


plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterPill.qml
```diff
@@ -12,7 +12,19 @@ import UM 1.2 as UM
...
+    property var tagText: {
+        switch(text) {
+            case "Ultimaker 3":
+                return "UM 3"
+            case "Ultimaker 3 Extended":
+                return "UM 3 EXT"
+            case "Ultimaker S5":
+                return "UM S5"
+            default:
+                return ""
```
sedwards2009: Shouldn't it just return `text` if it is unknown?

ianpaschal: Mmmmm I guess it could. I didn't bother since we have only those 3 printers being showed in the UM3NetworkingPlugin but I guess might as well be ready for the next printer in the Ultimaker line up 😛

ianpaschal: Fixed

PR #4975, opened 2018-12-12
Every print ouput device should define its connection type(usb, netwo…

alekseisasin:
…rk, cluster and etc)

CURA-6011


resources/qml/PrinterSelector/MachineSelectorList.qml
```diff
@@ -32,7 +32,7 @@ Column
             id: networkedPrintersModel
             filter:
             {
-                "type": "machine", "um_network_key": "*", "hidden": "False"
+                "type": "machine", "um_network_key": "*", "hidden": "False", "um_connection_type": "[2,3,4]"
```
ChrisTerBeke: All references of um_network_key should now be removed from the UI? (There's more places in the UI where it's currently used).

nallath: That's the idea yeah.

cura/PrinterOutputDevice.py
```diff
@@ -34,6 +34,12 @@ class ConnectionState(IntEnum):
     busy = 3
     error = 4
 
+class ConnectionType(IntEnum):
+    none = 0
```
nallath: In most cases of Enums that could find we use UpperCamelCase

cura/PrinterOutputDevice.py
```diff
@@ -117,6 +124,16 @@ def setConnectionState(self, connection_state: ConnectionState) -> None:
             self._connection_state = connection_state
             self.connectionStateChanged.emit(self._id)
 
+    def checkConnectionType(self, connection_type: ConnectionType) -> bool:
```
nallath: uh? Why is this a function? This is essentially the same the getConnection...

cura/PrinterOutputDevice.py
```diff
@@ -34,6 +34,12 @@ class ConnectionState(IntEnum):
     busy = 3
     error = 4
 
+class ConnectionType(IntEnum):
+    none = 0
+    usbConnection = 1
+    networkConnection = 2
+    clusterConnection = 3
```
nallath: WHy also make a difference between a cluster connection? Shouldn't this be more generic (now this is just re-implementing all the output devices that we have).
I think it's better if you have the following types: None, Direct, Local, Cloud

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -116,22 +117,28 @@ def setGroupName(self, group_name: str) -> None:
             # Ensure that the connection states are refreshed.
             self._network_plugin.reCheckConnections()
 
-    @pyqtSlot(str)
-    def setKey(self, key: str) -> None:
-        Logger.log("d", "Attempting to set the network key of the active machine to %s", key)
+    @pyqtSlot(QObject)
+    def setKey(self, printer_device: Optional[PrinterOutputDevice]) -> None:
```
nallath: This shouldn't be called setKey if it doesn't accept a key...

resources/qml/PrinterSelector/MachineSelectorList.qml
nallath: Also; Put documentation here regarding what the filtering is actually doing. From this code, I have no idea what 2,3,4 means.

LipuFei:
I'll do a rebase for this branch. It's not based on `4.0`

cura/PrinterOutputDevice.py
LipuFei: Fixed in 6bf39a32a941dfc4ef96a3f250527c04d9060541

cura/PrinterOutputDevice.py
LipuFei: Removed in 0aa49270ebb53521b3364ce5ab795b51d3e3adba This function is not used.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
LipuFei: Renamed it to `associateActiveMachineWithPrinterDevice()` in aba3b6dd8

cura/PrinterOutputDevice.py
LipuFei: I think the names `Direct` and `Local` are not so clear when I first read it. `USB` and `Network` sounds more clear. The distinction between `network` (legacy) and `cluster` (new) can indeed be confusing, but I wonder if there is a good reason to separate them (because of some differences in implementation details maybe?)

LipuFei:
made `ConnectionType` enum type accessible via QML in 4252b956039f4875e63735944ee8d1f36d5c1cb8

cura/PrinterOutputDevice.py
nallath: Well, we don't show a different icon as far as I know. The main difference is mostly due to there being a different number of machines, but that doesn't change the connection. 
So in that regard, the cluster connection type should be removed (Because the connection isn't different, its the amount of machines per output type that's different)

ChrisTerBeke: There will be a different icon for local networking vs cloud (see designs).

resources/qml/PrinterSelector/MachineSelectorList.qml
```diff
@@ -7,78 +7,40 @@ import QtQuick.Controls 2.3
...
 
-    Label
+    section.delegate: Label
     {
-        text: catalog.i18nc("@label", "Connected printers")
-        visible: networkedPrintersModel.items.length > 0
+        text: section == "true" ? catalog.i18nc("@label", "Connected printers") : catalog.i18nc("@label", "Preset printers")
+        width: parent.width
         leftPadding: UM.Theme.getSize("default_margin").width
-        height: visible ? contentHeight + 2 * UM.Theme.getSize("default_margin").height : 0
```
diegopradogesto: Add this back to match the designs.

fieldOfView:
Is this still up for 4.0? It causes quite a couple of API breaks for plugins that implement an OutputDevice: NetworkPrinterOutputDevice constructor, ConnectionState enum changes, and probably others.

fieldOfView:
Thanks for the quick reply. Good news that there are going to be multiple betas. However, I want to raise a concern about this. If I publish an OctoPrintPlugin compatible with the first beta, it will probably cause a crashing bug in the second beta with the changes of this PR in. 

What would be the best course of action? Delay updating the (most-downloaded) plugin until the second beta?

ChrisTerBeke:
My advice would be to wait. The beta period is really to update plugins for the final release if needed. If someone really wants the new goodies and want to use the plugin they can either run both versions of Cura or manually install a temporary modified version of the plugin or something. We also advice production users to not use any Beta releases.

fieldOfView:
Ok, nect question: Will the (semver) API be updated between the 1st and 2nd beta to reflect the changes? Otherwise there will still be a problem when people install the plugin that is intended for beta 2 on their beta 1.

fieldOfView:
Ok, just as long as UM knows there will probably be a crashing bug because of changes and decisions like this...

PS: What is the whole semantic SDK versioning for then?

cura/PrintersModel.py
```diff
@@ -0,0 +1,69 @@
...
+        container_stacks = ContainerRegistry.getInstance().findContainerStacks(type = "machine")
+
+        for container_stack in container_stacks:
+            connection_type = container_stack.getMetaDataEntry("connection_type")
+            has_remote_connection = connection_type in [str(ConnectionType.NetworkConnection), str(ConnectionType.CloudConnection)]
+
+            if container_stack.getMetaDataEntry("hidden", False) in ["True", True]:
+                continue
+
+            # TODO: Remove reference to connect group name.
```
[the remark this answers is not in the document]
nallath: Yeah, but it's a bit much for this rework.

cura/Settings/MachineManager.py
```diff
@@ -751,7 +757,7 @@ def removeMachine(self, machine_id: str) -> None:
                 self.setActiveMachine(other_machine_stacks[0]["id"])
 
         metadata = CuraContainerRegistry.getInstance().findContainerStacksMetadata(id = machine_id)[0]
-        network_key = metadata["um_network_key"] if "um_network_key" in metadata else None
+        network_key = metadata.get("um_network_key", None)
```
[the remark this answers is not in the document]
nallath: Nope. It's still a UM network key :( I've tried to remove most instances of it, but I didn't dare change all of them just yet.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -116,22 +119,30 @@ def setGroupName(self, group_name: str) -> None:
...
+                global_container_stack.setMetaDataEntry("um_network_key", printer_device.key)
                 # Delete old authentication data.
-                Logger.log("d", "Removing old authentication id %s for device %s", global_container_stack.getMetaDataEntry("network_authentication_id", None), key)
+                Logger.log("d", "Removing old authentication id %s for device %s", global_container_stack.getMetaDataEntry("network_authentication_id", None), printer_device.key)
                 global_container_stack.removeMetaDataEntry("network_authentication_id")
                 global_container_stack.removeMetaDataEntry("network_authentication_key")
-                CuraApplication.getInstance().getMachineManager().replaceContainersMetadata(key = "um_network_key", value = previous_network_key, new_value = key)
+                CuraApplication.getInstance().getMachineManager().replaceContainersMetadata(key = "um_network_key", value = previous_network_key, new_value = printer_device.key)
+
+                if "um_connection_type" in meta_data:
```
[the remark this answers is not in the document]
nallath: Yes. That's still old code.

nallath:
Well, as for the Octorprint, as long as it doesn't touch the connection state (which is set to NetworkConnection by default), it should go well. 

If this is merged, after the first beta, we should just increase the semver. That's indeed the entire point. It's also not up to Chris to say something about this ;)

ChrisTerBeke:
Okay fine. I just commented with what I’ve seen so far, which is that we did not change the sdk version between beta and stable even if it had breaking changes... but if this is the new versioning strategy that’s good. Maybe we should document this behavior somewhere in the wiki though.

cura/PrinterOutputDevice.py
ianpaschal: @nallath PascalCase! How dare you rob me of my case name!

decading:
Hello all. Is there published and available an octoprint plugin for the first 4.0 beta version? If yes, how could I et a link for it ? or a git url

fieldOfView:
There is not. You will have to wait until the second beta.

PR #4992, opened 2018-12-14
Added definition for Creatable D3

jbkim:
Creatable D3 https://creatablelabs.com/

### DIFF
--- a/resources/definitions/creatable_d3.def.json
+++ b/resources/definitions/creatable_d3.def.json
@@ -9,7 +9,6 @@
         "file_formats": "text/x-gcode",
         "icon": "icon_ultimaker.png",
         "platform": "rostock_platform.stl",
-        "has_materials": false,
         "machine_extruder_trains":
         {
             "0": "creatable_d3_extruder_0"
--- a/resources/definitions/creatable_d3.def.json
+++ b/resources/definitions/creatable_d3.def.json
@@ -12,7 +12,7 @@
         "has_materials": false,
         "machine_extruder_trains":
         {
-            "0": "creatable_d3_extruder"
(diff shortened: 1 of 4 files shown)


nallath:
For our reference; CURA-6084

jbkim:
@nallath What's the meaning of 6064?

nallath:
It's our ticket number in Jira (our internal issue tracker). It basically means that we've added it to our sprint so that someone will have a look at it at some point.

nallath:
The profiles don't actually work. There is a json error with them and when you try to load them into Cura, it fails to add the machine.

jbkim:
@nallath I just fixed the json error. :)

nallath:
Did you actually test if this works? If you try to add the machine it fails. As far as I can see you didn't set the ID of the machine correctly in the extruder definition.

jbkim:
@nallath I added the machine and it works but I can't see the Material menu under
setting > Extruder 1. Could you tell me how I can fix this?

fieldOfView:
Remove the line ```"has_materials": false,``` from the machine definition file

jbkim:
@nallath Thanks. I fix it. ;)

jbkim:
@nallath Hi, I expected this update applied to 4.0-beta but it didn't. Is there any problem?

Ghostkeeper:
Sorry. It was merged to master, which is to be released in 4.1.

We've split off the 4.0 beta from the master code base in December, before the first beta. We do that to allow the 4.0 release to become more stable while we work on new things for 4.1. So as a rule, while the beta is going on, we won't add new things to keep the release stable. That includes new printers.

The 4.0 beta has been exceptionally long so it looks a bit strange, I understand.

PR #4994, opened 2018-12-14
STAR-322: Cloud connection resumable upload

DanielSchiavini:


cura/NetworkClient.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2018 Ultimaker B.V.
```
ChrisTerBeke: This class is not used anymore now that we bind directly to callbacks on the reply, so we can remove it.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -3,21 +3,25 @@
...
 from UM.FileHandler.FileHandler import FileHandler
 from UM.Logger import Logger
 from UM.Message import Message
 from UM.Qt.Duration import Duration, DurationFormat
 from UM.Scene.SceneNode import SceneNode
+from cura.CuraApplication import CuraApplication
 from cura.PrinterOutput.NetworkedPrinterOutputDevice import AuthState, NetworkedPrinterOutputDevice
 from cura.PrinterOutput.PrinterOutputModel import PrinterOutputModel
 from plugins.UM3NetworkPrinting.src.Cloud.CloudOutputController import CloudOutputController
+from src.Cloud.Models.CloudClusterResponse import CloudClusterResponse
```
ChrisTerBeke: This import should be relative.

nallath:
Could you guys adhere to the commit style guide as defined at; https://github.com/Ultimaker/Meta/blob/master/general/writing_proper_commit_messages.md

Jira tags, for instance, should be in the body of the commit message.

ChrisTerBeke:
okay sure.

PR #4997, opened 2018-12-14
Quick fix to prevent monitor stage from overlapping header

ianpaschal:


ChrisTerBeke:
Shouldn't this go into the 4.0 branch (and then into master from there)?

ianpaschal:
@nallath Can't anchor to `headerBackground` or `backgroundItem` because they're not parent or sibling of the `main` `Loader`.

I think the solution here is pretty robust instead.

PR #5006, opened 2018-12-17
Fix UM3InfoComponents warnings

ianpaschal:
Also improve readability of super-duper-ultra-long-lines


ianpaschal:
Logging into Jenkins shows build was successful 🤷‍♂️

PR #5017, opened 2018-12-18
Add time prime tower into Cura.proto

LipuFei:
Sync with https://github.com/Ultimaker/CuraEngine/pull/846


Ghostkeeper:
Github seems to be doing something a bit wrong with the list of commits in here. At least the diff is now correct.

Ghostkeeper:
In layer view with the Line Type colour scheme, the prime tower is now displayed white because these lines don't have a recognised type. Please add a thing to the colour scheme to fix it.

PR #5020, opened 2018-12-19
Cura 6004 missing tooltips

nallath:


resources/qml/ActionPanel/SliceProcessWidget.qml
```diff
@@ -109,6 +109,7 @@ Column
             fixedWidthMode: true
             anchors.fill: parent
             text: catalog.i18nc("@button", "Slice")
+            tooltip: "Start slicing process"
```
nallath: This is not translated.

resources/qml/ToolTip.qml
```diff
@@ -0,0 +1,66 @@
...
+        if (contentAlignment == Cura.ToolTip.ContentAlignment.AlignLeft)
+        {
+            return (label.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2) + padding * 2) * -1
+        }
+        return parent.width + Math.round(UM.Theme.getSize("default_arrow").width * 1.2 + padding)
+    }
+
+    y: Math.round(parent.height / 2 - label.height / 2 ) - padding
+
+    padding: 2
```
nallath: Hardcoded padding. I guess this should be based off something in the theme.

resources/qml/ActionButton.qml
```diff
@@ -110,11 +112,9 @@ Button
...
     }
 
-    ToolTip
+    Cura.ToolTip
     {
         id: tooltip
-        text: ""
-        delay: 500
-        visible: text != "" && button.hovered
+        show: button.hovered
```
nallath: The current wording of "show" feels more like it's a function than a variable. I think this needs to be refactored a bit. 

I actually like the previous version better ( `visible: text != "" && button.hovered`)

PR #5026, opened 2018-12-20
[4.0] CURA-6011 Fix config syncing and network group

LipuFei:


ChrisTerBeke:
FYI: This will cause many merge conflicts with the cloud printing PR.

ChrisTerBeke:
@LipuFei please check our PR (https://github.com/Ultimaker/Cura/pull/4926), we've also worked on some of these files and issues as we came across them. Let's investigate if it's worth re-basing these changes or if we'll fix the merge conflicts when they appear after this is merged.

cura/Settings/MachineManager.py
```diff
@@ -1197,10 +1197,17 @@ def _setQualityChangesGroup(self, quality_changes_group: "QualityChangesGroup")
...
+    #
+    def _setVariantNode(self, position: str, container_node: Optional["ContainerNode"]) -> None:
+        if self._global_container_stack is None:
             return
-        self._global_container_stack.extruders[position].variant = container_node.getContainer()
+        container = empty_variant_container
+        if container_node is not None:
+            container = container_node.getContainer()
+        self._global_container_stack.extruders[position].variant = container
         self.activeVariantChanged.emit()
```
Ghostkeeper: Maybe we should only emit if the variant actually changed? Previously it would not emit in more cases due to the return statement.

Checking before emitting tends to help preventing signal storms.

cura/Settings/MachineManager.py
```diff
@@ -1368,15 +1368,12 @@ def applyRemoteConfiguration(self, configuration: ConfigurationModel) -> None:
                                                                                        configuration.buildplateConfiguration,
                                                                                        extruder_configuration.material.guid)
 
-                if variant_container_node:
+                # If the remote machine has an empty variant, do nothing
+                if variant_container_node is not None:
                     self._setVariantNode(position, variant_container_node)
-                else:
-                    self._global_container_stack.extruders[position].variant = empty_variant_container
```
Ghostkeeper: Uh, but if `variant_container_node` was None, the variant is retained from whatever it was previously?

plugins/UM3NetworkPrinting/src/util.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Any, Dict, Optional
+
+from UM.Logger import Logger
+
+
+# Finds all machines (GlobalStacks) with the given network_key or group_name and update the metadata of those
+# machines with the key/values in the given metadata dict.
+# Note that the old authentication data "network_authentication_id" and "network_authentication_key" will be
+# deleted if found.
+def updateMetadataForNetworkGroupMachines(network_key: Optional[str], group_name: Optional[str],
```
Ghostkeeper: I think it makes more sense to put this in the DiscoverUM3Action.py file because it's a function about updating the metadata of printers with data discovered through zeroconf.

plugins/UM3NetworkPrinting/src/util.py
```diff
@@ -0,0 +1,37 @@
...
+    if group_name is not None:
+        filter_dict["connect_group_name"] = group_name
+
+    from cura.CuraApplication import CuraApplication
+
+    active_global_stack = CuraApplication.getInstance().getMachineManager().activeMachine
+    container_registry = CuraApplication.getInstance().getContainerRegistry()
+    global_stack_list = container_registry.findContainerStacks(type = "machine", **filter_dict)
+    if active_global_stack is not None and active_global_stack not in global_stack_list:
+        global_stack_list.append(active_global_stack)
```
Ghostkeeper: You need to document (at least for me) to understand why the active machine is always added to this list.

plugins/UM3NetworkPrinting/src/util.py
```diff
@@ -0,0 +1,37 @@
...
+def updateMetadataForNetworkGroupMachines(network_key: Optional[str], group_name: Optional[str],
+                                          new_metadata: Dict[str, Any]) -> None:
+    # Either match with network_key or group_name
+    filter_dict = {}
+    if network_key is not None:
+        filter_dict["um_network_key"] = network_key
+    if group_name is not None:
+        filter_dict["connect_group_name"] = group_name
+
+    from cura.CuraApplication import CuraApplication
```
Ghostkeeper: Why is this import not at the top of the file? I would hope that there is not a circular dependency on this because this is in a plug-in, and CuraApplication should not import plug-ins.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -70,6 +64,18 @@ def __init__(self):
         self._service_changed_request_queue = Queue()
         self._service_changed_request_event = Event()
         self._service_changed_request_thread = Thread(target=self._handleOnServiceChangedRequests, daemon=True)
+
+        from cura.CuraApplication import CuraApplication
```
Ghostkeeper: Please put this import at the top of the file.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -70,6 +64,18 @@ def __init__(self):
...
+        CuraApplication.getInstance().initializationFinished.connect(self._initialize)
+
+    def _initialize(self) -> None:
```
Ghostkeeper: Please document this. It makes no sense for the reader of this code that this has to be delayed to after `initializationFinished`. It might also be more useful to call this function `_onInitializationFinished` instead of a meaningless `_initialize`.

nallath:
I'm not quite sure why we need this many changes to actually fix the issues. I've accidentally missed this pull request when fixing other review remarks, so we might want to close this.

PR #5037, opened 2018-12-21
Cura 6033 unknown material

nallath:


resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
```diff
@@ -14,6 +14,23 @@ Button
     property var configuration: null
     hoverEnabled: true
 
+    property bool isValidMaterial:
+    {
+        var extruderConfigurations = configuration.extruderConfigurations
+
+        for (var index = 0; index < extruderConfigurations.length; index++)
```
nallath:
```suggestion
        for (index in extruderConfigurations)
```
You can also use for each syntax with javascript.

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
```diff
@@ -14,6 +14,23 @@ Button
...
+        {
+            var name = extruderConfigurations[index].material ? extruderConfigurations[index].material.name : ""
+
+            if (name == "" || name == "Unknown")
+            {
+                hoverEnabled = false
```
nallath: The hoverEnabled property shouldn't be set like that. It's better to just use `hoverEnabled: isValidMaterial `

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
```diff
@@ -40,19 +57,99 @@ Button
...
             {
                 id: repeater
-                model: configuration.extruderConfigurations
+                model:
+                {
+                    if (configurationItem.isValidMaterial)
+                    {
+                        return configuration.extruderConfigurations
+                    }
+                    return []
```
nallath: Why change the model instead of changing the delegate? 

It's not that there is no configuration, the point is that we need to show a different call to action if the material is not valid.

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
```diff
@@ -40,19 +57,99 @@ Button
...
+                }
+
+                Label
+                {
+                    id: unknownMaterialMessage
+                    text:
+                    {
+                        var extruderConfigurations = configuration.extruderConfigurations
+                        var unknownMaterials = []
+                        for (var index = 0; index < extruderConfigurations.length; index++)
```
nallath:
```suggestion
                        for (var index in extruderConfigurations)
```

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
```diff
@@ -40,19 +57,99 @@ Button
...
+                    verticalAlignment: Text.AlignVCenter
+                    linkColor: UM.Theme.getColor("text_link")
+
+                    onLinkActivated:
+                    {
+                        Cura.Actions.browsePackages.trigger()
+                    }
+                }
+
+                MouseArea {
```
nallath:
```suggestion
                MouseArea 
                {
```

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
alekseisasin: it is set to true (Line: 15,  hoverEnabled: true), check above code.

Ghostkeeper: Yes, but the problem is that this breaks the binding. If it becomes a valid material during runtime, then it would still not have a hover.

Currently that is no problem because we can't get materials that are already on the printer during runtime (you have to restart Cura after you install a package), but if we get that working better then this would be a bug.

resources/qml/Menus/ConfigurationMenu/ConfigurationItem.qml
Ghostkeeper: Fixed: 5e9854454185fee9490b0d9aa996d36f65b7e2fd

PR #5058, opened 2018-12-24
Make settings panel resizable

Ghostkeeper:
This makes it possible to drag the bottom side of the panel, stretching the panel to make more settings visible.
screenshot from 2018-12-24 16-19-42

Some caveats: You can drag it to below the action panel (it's very hard to fix since the action panel is in a different context) but this is recoverable. You can't drag it below the window height, but you can then make the window height smaller so that it's now below the window height. This is also difficult to fix but hopefully the user understands that when he resizes the window.

Implements CURA-6054.

### DIFF
--- a/resources/qml/Preferences/ProfilesPage.qml
+++ b/resources/qml/Preferences/ProfilesPage.qml
@@ -408,7 +408,7 @@ Item
                     {
                         anchors.left: parent.left
                         anchors.leftMargin: UM.Theme.getSize("default_lining").width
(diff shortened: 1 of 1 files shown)


fieldOfView:
Cool! Perhaps add a gripper icon to the bottom of the floater as an affordance? Like https://fontawesome.com/icons/grip-lines?style=solid

Ghostkeeper:
I like the gripper icon idea. We'll ask UX.

PR #5090, opened 2018-12-31
Unify font types

Ghostkeeper:
A couple of changes with regarding to fonts (and some small changes underway):

1. The number of fonts have been reduced in order to make the interface more consistent and easier to maintain.
2. We went through the design and saw where the font size was incorrect.
3. The default font size has been adjusted because it was larger than in the designs, and the difference with "medium" font size was too small to notice.
4. Here and there some text colours or font weights have been adjusted. Not just the size.
5. Added font faces where they were missing, instead of using default system fonts (except in the preferences screens and other pop-ups).
6. Fixed "Slicing..." text when not autoslicing.

Note that this breaks our automated GUI tests. We can consider reverting the 3rd item if this is considered to be more important than differentiating between font sizes.

This implements CURA-6025. Merge together with https://github.com/Ultimaker/Uranium/pull/440


fieldOfView:
Woah, type has become a LOT smaller on Windows.

PR #5093, opened 2019-01-01
fix start gcode on alfawise u20

pinchies:
small tweak to fix start gcode in order to avoid default position of bed clamps


nallath:
For our reference; CURA-6083

PR #5104, opened 2019-01-03
Add variant to the configuration menu headerItem

fieldOfView:
This PR adds the variant name (print-core / nozzle size) to the headerItem of the configuration menu:
image

This makes the header a better "summary" of the extruder configuration. Without this, the print core/nozzle size is only shown when expanding the configuration menu for the extruders.


diegopradogesto:
Devs. CURA-6100

PR #5107, opened 2019-01-04
Update DisplayFilenameAndLayerOnLCD.py

adecastilho:
updated to display layer count before filename (layer number was getting cut off in event of long filename).
updated to include option to start layer count at either 0 or 1.
removed ':' in the display string as it is a gcode command that was splitting the line into two different commands.


nallath:
For our reference; CURA-6078

nallath:
Sorry that this took so freaking long to merge. Thanks for taking the time to improve Cura!

PR #5113, opened 2019-01-04
Fix z-fighting on bottom layer of x-ray view

fieldOfView:
This PR fixes z-fighting in the X-Ray view, by not rendering convex hulls. It sort-of makes sense not to render a shadow in X-Ray view.

Example with a fairly flat model where the z-fighting is particularly distracting:
**Before**
image

**After**
image


nallath:
CURA-6082

fieldOfView:
Note that the test that fails is not in the code I touched.

PR #5126, opened 2019-01-07
Please add the new Anycubic Chiron Printer (New request from Fork)

BluefireXD:
Hi, I have the Anycubic Chiron printer and created the definition files.
Tested with Cura 3.6.0 and 4.0 Beta
This is my second request. This time from a fork^^. 

Bests,
Patrick 

anycubic chiron definitions cura 3 6 0

anycubic chiron definitions cura 4 0


nallath:
For our reference; CURA-6080

resources/definitions/anycubic_chiron.def.json
```diff
@@ -0,0 +1,80 @@
...
+        "has_materials": true,
+		"preferred_material": "generic_pla",
+        "has_machine_quality": true,
+		"quality_definition": "anycubic_chiron",
+		"preferred_quality_type": "normal",
+        "machine_extruder_trains":
+        {
+            "0": "anycubic_chiron_extruder_0"
+        },
+		"firmware_file": "CHIRON.hex"
```
nallath: Your pull request doesn't include this file, so there is no point in defining this.

resources/definitions/anycubic_chiron.def.json
```diff
@@ -0,0 +1,80 @@
...
+        "manufacturer": "Anycubic",
+		"category": "Other",
+        "file_formats": "text/x-gcode",
+        "icon": "icon_ultimaker2",
+        "platform": "anycubic_chiron_platform.obj",
+		"platform_texture": "anycubic-chiron.png",
+        "has_materials": true,
+		"preferred_material": "generic_pla",
+        "has_machine_quality": true,
+		"quality_definition": "anycubic_chiron",
```
nallath: Bit of nitpicking, but your indentation is a bit off.

BluefireXD: Y thx
Its not the same like it looks in notepad++. In notepad++ it was fine.
I will change this

resources/definitions/anycubic_chiron.def.json
BluefireXD: Thx, y in my local cura i had this firmware file. In the fork the firmware folder was missed. 
Should i add the folder including the firmware file or is there a reason why the firmware folder is missed in fork?

fieldOfView: Firmware files go in a separate repo:
https://github.com/Ultimaker/cura-binary-data/tree/master/cura/resources/firmware

MarlinChiron.hex would probably be a better filename.

BluefireXD:
OK, i will rebase the pullrequets in the evening

fieldOfView:
I think you can just use the github interface to change the target branch (using the "Edit" button next to the title of the pull request)

BluefireXD:
Ah y thx, didnt see that dropdown. Now it's pointing to your master. 
Because i added only new files it's ok to use my 4.0 branch from fork as source (can't change the source :/)?

resources/definitions/anycubic_chiron.def.json
BluefireXD: Thx a lot. I added a pull request on https://github.com/Ultimaker/cura-binary-data/pulls and updated the definition file.
I also saw that somehow my indentation is made with spaces instead of tabs. Maybe it’s because I used a anycubic i3 Mega or similar definition file as template for the chiron definition file. 
Me, I would prefer tabulator. I never use spaces. What do you prefer here?

BluefireXD:
"continuous-integration/jenkins/pr-merge — This commit has test failures"
Where can i see what failure it throws? If I open Details it just shows me "This Page Cannot Be Displayed..."

fieldOfView:
The link works for me (and I am also an external contributor). This is the most relevant output of the page:
```
The following tests FAILED:

	  2 - pytest-FirmwareUpdateChecker (Failed)
```

You changed nothing in the FirmwareUpdateChecker. Cura was already broken when you pushed your changes. In other words: nothing you have to worry about for now.

nallath:
Our CI system is a fickle beast and is quite prone to breaking (even though there is nothing wrong). We're trying to move to a different system so that it doesn't kill itself every so often, but that is going to take quite some time.

diegopradogesto:
I have taken the liberty to make some changes to align more with the other Anycubic printers: 0286c701fdfef3754ae0a5840baef2fdec95526d

I have corrected the indentation and made the quality profiles global to avoid confusion to the users with the layer height (it was showing 0.06 when it was actually 0.1 for instance)

BluefireXD:
Hi , is the printer already added to cura 4.0? 
I downloaded the offical 4.0 version but cant find the printer :S

Ghostkeeper:
No, I'm sorry. It'll be in 4.1. Like Nallath said earlier here:

> We try not to add any new machines once we have a release branch, as we want to decrease the risk for the actual release.

We don't want to add new features once we're stabilising a version in preparation for release. As this was added to master instead of 4.0, it was meant to be released with 4.1 rather than with 4.0. The beta track for 4.0 was unusually long so it was a bit weird, I'm afraid.

PR #5128, opened 2019-01-07
Add version upgrade for 4.0

nallath:
Contributes to CURA-6011 and CURA-5995


ChrisTerBeke:
Quality files changed, so a rebuild + upload of the Ultimaker resources packages is needed as well (in the other repo).

nallath:
We don't need to per-se, since the files on remote won't be downloadable anyway. But it's probably a bit cleaner to do it.

ChrisTerBeke:
Oh yeah sure, doesn't need to be right now, just a heads up as it's a new flow and we might forget about it.

PR #5150, opened 2019-01-11
Fix resizable Settings-Panel for change in window size. [CURA-6054]

rburema:


resources/qml/PrintSetupSelector/PrintSetupSelectorContents.qml
```diff
@@ -77,7 +77,12 @@ Item
                 target: UM.Preferences
                 onPreferenceChanged:
                 {
-                    customPrintSetup.height = UM.Preferences.getValue("view/settings_list_height");
```
fieldOfView: This is done each time ANY preference is changed. Seems a bit... overzealous. Is there no way to do this more efficiently?

rburema: It _does_ seem slightly too much. I've asked around here, and while there isn't a(n easy) way to see where the `onPreferenceChanged` came from, we _could_ try to cache the height (depending on wether QML is going to change any UI thing if the height stays the same _anyway_).

fieldOfView: CuraApplication could emit its own signal.

rburema: @nallath hinted that it's possible to add an argument to the signal emitted. I looked into this further, and it turns out this already happens. It _should_ be possible to access this parameter in QML as well I think. I'll try to get this to work.

rburema: Created PR #5175 .

PR #5158, opened 2019-01-13
Remove loading gif

diegopradogesto:
Instead of an external gif, use the BusyIndicator that exists in the ActionButton.


diegopradogesto:
Devs. CURA-6108

PR #5167, opened 2019-01-14
Ensure that the toolbox correctly handles multiple supported SDK vers…

nallath:
…ions

CURA-6087

### DIFF
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -235,12 +235,17 @@ class Toolbox(QObject, Extension):
 
     def _convertPluginMetadata(self, plugin_data: Dict[str, Any]) -> Optional[Dict[str, Any]]:
         try:
+            highest_sdk_version_supported = Version(0)
+            for supported_version in plugin_data["plugin"]["supported_sdk_versions"]:
+                if supported_version > highest_sdk_version_supported:
+                    highest_sdk_version_supported = supported_version
+
             formatted = {
                 "package_id": plugin_data["id"],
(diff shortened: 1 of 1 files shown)


plugins/Toolbox/src/Toolbox.py
```diff
@@ -235,12 +235,17 @@ def _createDialog(self, qml_name: str) -> Optional[QObject]:
...
+                if supported_version > highest_sdk_version_supported:
+                    highest_sdk_version_supported = supported_version
+
             formatted = {
                 "package_id": plugin_data["id"],
                 "package_type": "plugin",
                 "display_name": plugin_data["plugin"]["name"],
                 "package_version": plugin_data["plugin"]["version"],
-                "sdk_version": plugin_data["plugin"]["api"],
+                "sdk_version": highest_sdk_version_supported,
```
LipuFei: I don't think this is entirely correct. For example, if we release 4.0 (SDK v6) with this code, when we get to 5.0 (SDK v8), and a plugin is published to support `[6,7,8]`, and a user with 4.0 (SDK v6) picks up that plugin, it won't work because the highest SDK v8 != SDK v6. I think in this part, it should return the current SDK version if the plugin supposes that. If not, then return the highest SDK version.

LipuFei: If possible, I would add a unit test for this function as well. Those version comparison code makes me nervous...

diegopradogesto: Yes, indeed.

And +1 to add a new unit test.

PR #5172, opened 2019-01-15
Change the style of the resize bar

diegopradogesto:
Make it wider and more visible to the user.
Change icon.
Keep margins according to the other panels.

This is how it would look like with this change:
image


ChrisTerBeke:
There's more padding above than below the button now. My design OCD is acting up :P

diegopradogesto:
Done here: 86b0ac3

PR #5179, opened 2019-01-16
integration of printer profiles for ALYA 3DP and ALYA NX 3DP

nbennett1978:
See our website with information on our 3D printer line ALYA:
https://alya3dp.com/ 

We'd kindly like to request our printer profiles into Cura sources. I tested the proposed profiles locally.


diegopradogesto:
Thanks for contribute. I created the internal ticket CURA-6121. We will review it as soon as we can. Considering that we still have a lot of PR to review, this can take a while until we add it to Cura. Please be patient.

nallath:
Your material isn't that different from generic PLA. If you want specific settings for your printer for generic PLA, they should be added to the generic PLA_175 profile (like IMADE3D and Cartesio have done).

As for adding a completely new material; It's better to just use the marketplace for that. That way it's also clearer for you as to how many people actually use it (since it will also track downloads and enable you to update the material if that's ever required)

Could you also see if the mesh you provided can be reduced? We try to keep all the buildplate models below 1 MB.

nbennett1978:
Ok, no problem. I will do the changes as you proposed and also decrease the mesh file sizes. 
Just on question: should i create a new thread with a new pull request, or is there some way to continue this pull request? What is the meaning of "internal ticket CURA-6121" for me specifically?

diegopradogesto:
@nbennett1978 you can continue with this PR and make the changes here. The PR in materials can be closed if it's not useful anymore.

Regarding the internal ticket, don't worry, it's just for internal reference in our JIRA board. It's is just for developers to easy follow up this topic in our system.

nbennett1978:
Hi,
i just wanted to ask when i could get some feedback on my commit. I know you said you have lots to do for the upcoming release. I am not urging or anything, i just wondered if i have done everything right, that is, i wonder wether i am not getting any feedback because i didn't commit the right way, or just because you didn't have the time yet :). Would you mind giving me some feedback/confirmation regarding my proposed commits?
@diegopradogesto @nallath

Ghostkeeper:
Do you know why the definition excludes this large list of materials? Is it for some reason incompatible with the extruder train?

resources/definitions/alya3dp.def.json
```diff
@@ -1,50 +1,50 @@
...
+        "machine_height":              {   "default_value": 133             },
+        "machine_depth":               {   "default_value": 100             },
+        "machine_center_is_zero":      {   "default_value": false           },
+        "gantry_height":               {   "default_value": 55             },
+        "retraction_amount":           {   "default_value": 1.5 },
+        "support_enable":              {   "default_value": true},
+        "machine_head_with_fans_polygon": {
+            "default_value": [[75, 18],[18, 18],[18, 35],[75, 35]]
+        },
+        "adhesion_type": {"options": {"raft": "Raft" ,"none": "None", "brim": "Brim"},  "default_value": "raft"},
```
Ghostkeeper: Skirt is not allowed, and neither are any future adhesion types? Why are these options adjusted?

resources/definitions/alya3dp.def.json
```diff
@@ -1,50 +1,50 @@
...
...ltimaker_abs_white", "ultimaker_abs_yellow", "ultimaker_bam", "ultimaker_cpe_black", "ultimaker_cpe_blue", "ultimaker_cpe_dark-grey", "ultimaker_cpe_green", "ultimaker_cpe_light-grey", "ultimaker_cpe_plus_black", "ultimaker_cpe_plus_transparent", "ultimaker_cpe_plus_white", "ultimaker_cpe_red", "ultimaker_cpe_transparent", "ultimaker_cpe_white", "ultimaker_cpe_yellow", "ultimaker_nylon_black", "ultimaker_nylon_transparent", "ultimaker_pc_black", "ultimaker_pc_transparent", "ultimaker_pc_white", "ultimaker_pla_black", "ultimaker_pla_blue", "ultimaker_pla_green", "ultimaker_pla_magenta", "ultimaker_pla_orange", "ultimaker_pla_pearl-white", "ultimaker_pla_red", "ultimaker_pla_silver-metallic", "ultimaker_pla_transparent", "ultimaker_pla_white", "ultimaker_pla_yellow", "ultimaker_pp_transparent", "ultimaker_pva", "ultimaker_tough_pla_black", "ultimaker_tough_pla_green", "ultimaker_tough_pla_red", "ultimaker_tough_pla_white", "ultimaker_tpu_black", "ultimaker_tpu_blue", "ultimaker_tpu_red", "ultimaker_tpu_white", "verbatim_bvoh_175", "Vertex_Delta_ABS", "Vertex_Delta_PET", "Vertex_Delta_PLA", "Vertex_Delta_TPU", "zyyx_pro_flex", "zyyx_pro_pla","tizyx_pla","tizyx_abs","tizyx_pla_bois" ],
```
Ghostkeeper: Why do you exclude all of these materials? There's some materials of 2.85mm filament and some of 1.75mm filament in this list, so some of these wouldn't be shown anyway.

resources/quality/katihal/alya3dp_normal_flex.inst.cfg
```diff
@@ -0,0 +1,34 @@
...
+
+[metadata]
+setting_version = 5
+type = quality
+quality_type = alya_normal
+weight = 2
+material = alya_flex
+
+[values]
+material_diameter = 1.75
```
Ghostkeeper: This seems like a typical setting to put in the material profile...

resources/definitions/alya3dp.def.json
nbennett1978: Well, we don't have a heated bed, and we have a flexible print surface with holes. It allows easy printing using a raft. But without raft, prints come out quite ugly. If you don't mind, we'd like to leave it that way. If you insist, i could of course just use a default value as raft-

resources/quality/katihal/alya3dp_normal_flex.inst.cfg
nbennett1978: I thought it might be a good idea to add this just to be sure, as material settings are overriden by the print profile. Leaving this on would make material changes etc safer, in my humble opinion. Our extruder wouldn't support 3mm anyways. Does it have any other negative consequences levaing this line here?

resources/definitions/alya3dp.def.json
nbennett1978: I actually copied this line from a different printer already included in the sources of cura. Some printer specific settings are included in the material profiles, retraction speed and amount, as well as print temperature, of course. I find these mainly material dependent. In order to find optimum settings for the other materials would require testing and including these settings in all the material profiles. As all these materials were not tested, i excluded them, also for making it easier for users to pick the right material. I am open to suggestions, though-

nbennett1978: Well, most of the materials are not feasible due to missing bed heating, by the way.

resources/quality/katihal/alya3dp_normal_flex.inst.cfg
nbennett1978: There was a profile for flex materials, which i accidently left there, but which has no corresponding material definition in th ematerial database yet.

nbennett1978: I deleted alya_flex printing profiles

PR #5181, opened 2019-01-16
Let "retraction_combing" be overriden with "infill" as default value

drzejkopf:
"infill" option was not added to "resolve" after adding additional combing setting, so that wasn't possible to add it as override value in printer definition


LipuFei:
JIRA ticket: CURA-6129

PR #5185, opened 2019-01-17
STAR-322: Cloud connection improvements

DanielSchiavini:


nallath:
As I assume we also need this for 4.0, i've also merged it there.

PR #5187, opened 2019-01-17
CL-1165 missing cloud info

ianpaschal:
Shows a pop up blurb of info when hovering cloud devices' camera button or context menu.

Also includes shiny new reusable pop up component which fixed lots of warnings with the context menu.


plugins/UM3NetworkPrinting/resources/qml/GenericPopUp.qml
```diff
@@ -0,0 +1,227 @@
...
+        Item
+        {
+            id: pointedRectangle
+            anchors
+            {
+                horizontalCenter: parent.horizontalCenter
+                verticalCenter: parent.verticalCenter
+            }
+            height: parent.height - 10 * screenScaleFactor // Because of the shadow
+            width: parent.width - 10 * screenScaleFactor // Because of the shadow
```
sedwards2009: Isn't there a variable or constant somewhere which defines the size of the shadow?

ianpaschal: One would wish so, but no. I Googled far and wide and it's only possible if you have a reference to the actual Qt.Window which we do not.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
```diff
@@ -156,32 +158,61 @@ Item
...
                 topMargin: 12 * screenScaleFactor // TODO: Theme!
             }
-            printJob: printer ? printer.activePrintJob : null
             width: 36 * screenScaleFactor // TODO: Theme!
             height: 36 * screenScaleFactor // TODO: Theme!
-            enabled: base.enabled
-            visible: printer
+            enabled: !cloudConnection
+            
+            onClicked: enabled ? contextMenu.switchPopupState() : {}
```
sedwards2009: This should just be a normal if().

PR #5194, opened 2019-01-18
CURA-6094 fix crash on huge brim-widths

rburema:


rburema:
Closing this PR, because since it is a bugfix, not a feature, it should be for 4.0 instead of master.

PR #5203, opened 2019-01-21
Cura 6094 brim width crash 4 0

rburema:
Made a new PR for 4.0, instead of master, since it was a bugfix, not a feature.


rburema:
New 'version' of PR #5194

PR #5212, opened 2019-01-23
Chop longer material names in smaller pieces in print-job-info.

rburema:
[CURA-6141]

### DIFF
--- a/resources/qml/ActionPanel/PrintJobInformation.qml
+++ b/resources/qml/ActionPanel/PrintJobInformation.qml
@@ -90,6 +90,23 @@ Column
             property var printMaterialCosts: PrintInformation.materialCosts
             property var printMaterialNames: PrintInformation.materialNames
 
+            function putBreaksIfLong(item)
+            {
+                var result = ""
+
+                const max_length = 13  // No longer than 13 -> 'WWWWWWWWWWWWWW' doesn't fit.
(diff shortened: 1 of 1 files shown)


resources/qml/ActionPanel/PrintJobInformation.qml
```diff
@@ -90,14 +90,31 @@ Column
             property var printMaterialCosts: PrintInformation.materialCosts
             property var printMaterialNames: PrintInformation.materialNames
 
+            function putBreaksIfLong(item)
+            {
+                var result = ""
+
+                const max_length = 13  // No longer than 13 -> 'WWWWWWWWWWWWWW' doesn't fit.
```
fieldOfView: Would it be a good idea to better document the magic number, perhaps make it a constant of sorts defined at the top of the file?

nallath:
I guess an even better way is to do away with the HTML altogether and put it in actual QML. I don't really see a reason why it's in HTML table code in the first place.

resources/qml/ActionPanel/PrintJobInformation.qml
rburema: Probably, if we decide that it should remain HTML (see below).

rburema:
I've rewritten this to use QML-elements instead of HTML-tables, see #5224

PR #5224, opened 2019-01-26
Cura 6141 info long material name line split

rburema:


resources/qml/ActionPanel/PrintJobInformation.qml
```diff
@@ -36,30 +36,62 @@ Column
...
+            {
+                Repeater
+                {
+                    model: byLineType.getMaterialTable()
+                    Row
+                    {
+                        Repeater
+                        {
+                            model: modelData
+                            Text
```
fieldOfView: Use ```Label``` instead of ```Text``` (https://github.com/Ultimaker/Cura/wiki/QML-Best-Practices). Also since this is using QtQuickControls 2, you need to add ```renderType: Text.NativeRendering```

LipuFei: Fixed

rburema: Thanks.
I keep discovering new Best Practices documents in all kinds of different places :-)

PR #5229, opened 2019-01-28
[WIP] Improve property names to be more accurate

ianpaschal:
@nallath We can use this branch for more connection type improvements if we want, or just merge it in, up you.


ianpaschal:
#5235 makes this unnecessary.

PR #5255, opened 2019-01-31
CL-1154 Monitor empty states

ianpaschal:
- When you have a UM network-enabled printer, configured for network printing, connected, you should see the monitor tab, and if there are no jobs, see a link to Cura Connect.
- When you have a UM network-enabled printer, configured for network printing, not connected, you should see a warning to check that the printer is on and has connection.
- When you have a UM network-enabled printer, not configured for network printing, you should see a message telling you to select a network-enabled printer to monitor, and a link to the UM website with printer manuals.
- When you have any non-network enabled printer, you should see only the message telling you to select a network-enabled printer to monitor.

### DIFF
--- a/plugins/MonitorStage/MonitorMain.qml
+++ b/plugins/MonitorStage/MonitorMain.qml
@@ -2,7 +2,7 @@
 // Cura is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.10
-import QtQuick.Controls 1.4
+import QtQuick.Controls 2.0
 
 import UM 1.3 as UM
 import Cura 1.0 as Cura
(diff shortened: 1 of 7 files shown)


plugins/MonitorStage/MonitorMain.qml
```diff
@@ -2,20 +2,43 @@
...
+        // Check if configured connection types includes either 2 or 3 (LAN or cloud)
+        for (var i = 0; i < types.length; i++)
+        {
+            if (connectedTypes.indexOf(types[i]) >= 0)
+            {
+                return true
+            }
+        }
+        return false
+    }
```
sedwards2009: Isn't this code just a union of two sets with a test for non-zero?

sedwards2009:
@ian is there meant to be a Reconnect button on the page when the printer can't be reached?

plugins/MonitorStage/MonitorMain.qml
ianpaschal: Nope. 1 = USB printing.

PR #5259, opened 2019-02-01
Add {travel_speed} as an alias to {speed_travel}

fieldOfView:
This PR adds {travel_speed} as an alias to {speed_travel} for replacement patterns in start- and end-gcode snippets for all those people who still copy paste legacy start&end gcode snippets into Cura. It also fixes one current definition which uses the deprecated {travel_speed} pattern, to set a better example.


Ghostkeeper:
Developers, see issue CURA-6170.

Ghostkeeper:
The Windows CI seems to be confused about a Git commit that didn't exist. Maybe something weird happened with a rebase in this branch or something. The Linux CI did work though and this is not something that is likely to be platform-dependent.

PR #5260, opened 2019-02-01
CURA-6144 Fix combox style for support extruder in recommended

LipuFei:


diegopradogesto:
Why not to reuse part of the `SettingExtruder` component instead of duplicating the code?

LipuFei:
Tried that first, but `SettingItem` requires too many things to work, such as a bunch of definitions, and they are not easy to get here.

diegopradogesto:
No, I mean, `SettingExtruder.qml` uses a ComboBox as `contents`. I thought that maybe creating a component out of this ComboBox, then we can reuse it, right? That way we don't need to maintain two files that are exactly the same.

resources/qml/PrintSetupSelector/Recommended/RecommendedSupportSelector.qml
```diff
@@ -87,82 +88,219 @@ Item
...
+                    supportExtruderNr.setPropertyValue("value", model.getItem(index).index);
+                } else
+                {
+                    currentIndex = supportExtruderNr.properties.value;  // keep the old value
+                }
+            }
+
+            currentIndex: supportExtruderNr.properties.value
+
+            property string color: "#fff"
```
rburema: Even if we override it later, the initial value here is only half of a (non-alpha) colour.

resources/qml/PrintSetupSelector/Recommended/RecommendedSupportSelector.qml
```diff
@@ -87,82 +88,219 @@ Item
...
-                return (current_extruder.color) ? current_extruder.color : ""
+                for (var item_index in model.items)
+                {
+                    var item = model.getItem(item_index)
+                    if (item.index == position)
+                    {
+                        return item_index
+                    }
+                }
+                return -1
```
rburema: This seems to be a 'fail fast' measure, which is good, but should maybe be commented (like 'this shouldn't happen').

LipuFei: Changed in c237222d3

diegopradogesto:
What about reusing components...?

rburema:
@diegopradogesto He tried that, see above: The ComboBox/contents you're referring to takes up the entire class of `SettingExtruder.qml`, wrapped in a `SettingItem`.

PR #5270, opened 2019-02-04
Let each stage create its own action panel

Ghostkeeper:
The positioning and configuration of the action panel greatly differs per stage. In this pull request we make the stage create its own action panel widget. This way it can choose for itself whether or not it even wants an action panel, where the action panel is located and whether or not the preview button is visible on that panel.

Contributes to issue CURA-6086.

@fieldOfView You might want to check if this has any influence on your sidebar plug-in. It should behave the same, but you never know.


fieldOfView:
Thanks for the heads-up.

PR #5271, opened 2019-02-04
CL-1222 Cloud flow pop up

ianpaschal:
Making a PR already to start looking at the diff and discussing the implementation.


cura/Settings/MachineManager.py
```diff
@@ -517,6 +517,12 @@ def activeMachineId(self) -> str:
             return self._global_container_stack.getId()
         return ""
 
+    @pyqtProperty(str, notify = globalContainerChanged)
+    def activeMachineFirmwareVersion(self) -> str:
+        if not self._printer_output_devices[0]:
+            return ""
+        return self._printer_output_devices[0].firmwareVersion
+
```
ianpaschal: `active_machine = CuraApplication.getInstance().global_container_stack` and `active_machine = CuraApplication.getInstance().getMachineManager().activeMachine` are both used all over our codebase. Unsure which one is correct. I opted for the machine manager approach. However I'm also unclear about if we should be using the global stack for firmware. Checking if the machine has a network connection uses `self._printer_output_devices` but for almost everything else we use `self._global_container_stack.getWhatever()`. So I could write some some methods for the global stack to handle firmware but it feels a bit pointless right now because there's no prevailing way to handle the "current machine".

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -292,6 +307,7 @@ def _onAddDevice(self, name, address, properties):
         if global_container_stack and device.getId() == global_container_stack.getMetaDataEntry("um_network_key"):
             # Ensure that the configured connection type is set.
             global_container_stack.addConfiguredConnectionType(device.connectionType.value)
+            # global_container_stack.setFirmwareVersion(device.firmwareVersion)
```
ianpaschal: In connection to the comment above... we could set the firmware in the global stack so that the machine manager can access it, but why? Either way we still have spaghetti.

cura/Settings/MachineManager.py
nallath: There is no `active_machine = CuraApplication.getInstance().global_container_stack` There is a `_global_container_stack`, but that is protected (and should not be used outside of `Application` and it's children.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
nallath: The main reason we set the configuredConnection type is because we want that available if there is no connection. If we don't care about the firmware version if there is no connection, there is no point in storing it.

cura/Settings/MachineManager.py
ianpaschal: Sorry. Typo on my part. It's unclear which is the right way to get the machine:

- `active_machine = CuraApplication.getInstance().getGlobalContainerStack`, which returns the global container stack, or...

-  `active_machine = CuraApplication.getInstance().getMachineManager().activeMachine`, which also returns the global container stack.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
ianpaschal: Makes sense, except that there often is a connection and then the firmware is a bitch to get...

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -370,3 +400,124 @@ def _onServiceChanged(self, zero_conf, service_type, name, state_change):
...
+
+    def _onMachineSwitched(self):
+        if self._start_cloud_flow_message is not None:
+            self._start_cloud_flow_message.hide()
+            self._start_cloud_flow_message = None
+        if self._cloud_flow_complete_message is not None:
+            self._cloud_flow_complete_message.hide()
+            self._cloud_flow_complete_message = None
+
+        self.checkCloudFlowIsPossible()
```
sedwards2009: This whole file has a deficiency in Python type hints, mostly in the method defs.

ianpaschal: I'll add them in.

PR #5273, opened 2019-02-04
Fixing profile values for the NWA3D A5 Printer

DragonJe:
Please add this fix to the quality profile names in the NWA3D A5, per @fieldOfView's suggestion. I believe I changed everything to the correct nomenclature.

The over specifications are intentional, it prints best with those settings.

Thanks for all the help!


nallath:
For our reference; CURA-6186

resources/quality/nwa3d_a5/nwa3d_a5_best.inst.cfg
```diff
@@ -12,118 +12,120 @@ global_quality = True
 
 [values]
 layer_height = 0.08
-initial_layer_height = 0.25
+layer_height_0 = 0.24
 line_width = 0.4
-initial_layer_line_width = 100
+wall_line_width_0 = 100
```
diegopradogesto: This `wall_line_width_0` is not the same as the `initial_layer_line_width`. `wall_line_width_0` is the width of the outer wall. If you want to set the Initial layer line width you have to use `initial_layer_line_width_factor` instead.

resources/quality/nwa3d_a5/nwa3d_a5_best.inst.cfg
```diff
@@ -12,118 +12,120 @@ global_quality = True
...
+wall_0_inset = 0
 optimize_wall_printing_order = false
-outer_before_inner_walls = false
-alternate_extra_wall = false
-compensate_wall_overlaps = true
-compensate_outer_wall_overlaps = true
-compensate_inner_wall_overlaps = true
-minimum_wall_flow = 0
-fill_gaps_between_walls = everywhere
+outer_inset_first = false
```
diegopradogesto: Please replace all the occurrences of `false` with `False`

resources/quality/nwa3d_a5/nwa3d_a5_best.inst.cfg
```diff
@@ -12,118 +12,120 @@ global_quality = True
...
+alternate_extra_perimeter = false
+travel_compensate_overlapping_walls_enabled = true
```
diegopradogesto: Please replace all the occurrences of `true` with `True`

resources/quality/nwa3d_a5/nwa3d_a5_best.inst.cfg
DragonJe: Just did, thanks!

resources/quality/nwa3d_a5/nwa3d_a5_best.inst.cfg
DragonJe: Just did, thanks!

PR #5287, opened 2019-02-08
Fix token refresh

ChrisTerBeke:
This PR adds proper refreshing of access tokens before they are used in an API call that. By storing the creation date (and using the already stored expiry time) we calculate the expected expiry date. If the access token is requested from the account object after this time, we do a token refresh before returning the (new) access token. This prevents API calls to result in a 401.

### DIFF
--- a/cura/OAuth2/AuthorizationService.py
+++ b/cura/OAuth2/AuthorizationService.py
@@ -2,6 +2,7 @@
 # Cura is released under the terms of the LGPLv3 or higher.
 import json
 import webbrowser
+from datetime import timedelta, datetime
 from typing import Optional, TYPE_CHECKING
 from urllib.parse import urlencode
 import requests.exceptions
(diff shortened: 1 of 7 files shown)


cura/OAuth2/AuthorizationService.py
```diff
@@ -87,19 +87,20 @@ def _parseJWT(self) -> Optional["UserProfile"]:
...
-            return None
-
         if self._auth_data is None:
             Logger.log("d", "No auth data to retrieve the access_token from")
             return None
 
-        return self._auth_data.access_token
+        # Check if the current access token is expired and refresh it if that is the case.
+        received_at = datetime.strptime(self._auth_data.received_at, TOKEN_TIMESTAMP_FORMAT) \
+            if self._auth_data.received_at else datetime(2000, 1, 1)
```
DanielSchiavini: Shouldn't we store the expiration date instead? And handle it nicely if the token expires..

DanielSchiavini: We must handle 401's correctly anyway, as the token may be unvalidated by the server for other reasons

DanielSchiavini:
This PR was not ready @nallath

nallath:
According to @ChrisTerBeke it was. Also, if a pull request is not ready, don't make it. How else would anyone know it?

ChrisTerBeke:
IMO it was ready. It fixes the problem and it passed the code review.

For further improvements we can ensure that actual 401 API calls are handled more properly, but that's code in different places (cloud printing plugin, marketplace, backups plugin). Maybe a new unified API comm. class should handle that.

DanielSchiavini:
My fault, I should've requested changes. This helps with the issue in 90% of the cases

PR #5291, opened 2019-02-11
CL-1222 Ensure pop-up is not shown if cloud connection is already configured

ianpaschal:
Take 3... action!

### DIFF
--- a/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
+++ b/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
@@ -455,7 +455,7 @@ class UM3OutputDevicePlugin(OutputDevicePlugin):
         if not self._start_cloud_flow_message:
             self._start_cloud_flow_message = Message(
                 text = i18n_catalog.i18nc("@info:status", "Send and monitor print jobs from anywhere using your Ultimaker account."),
-                image_source = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../resources/svg/cloud-flow-start.svg",
+                image_source = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../resources/svg/cloud-flow-start.svg"),
                 image_caption = i18n_catalog.i18nc("@info:status", "Connect to Ultimaker Cloud"),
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -412,10 +412,15 @@ def checkCloudFlowIsPossible(self) -> None:
...
         if active_machine:
             
-            # Check 1: Printer isn't already configured for cloud
+            # Check 1A: Printer isn't already configured for cloud
             if ConnectionType.CloudConnection.value in active_machine.configuredConnectionTypes:
                 Logger.log("d", "Active machine was already configured for cloud.")
                 return
+            
+            # Check 1B: Printer isn't already configured for cloud
+            if active_machine.getMetaDataEntry("cloud_flow_complete", "value") is True:
```
diegopradogesto: Suggestion: Use `parseBool` instead of `is True`

nallath:
```suggestion
            if parseBool(active_machine.getMetaDataEntry("cloud_flow_complete", False)):
```
Importing it should be done by
`from UM.Util import parseBool`

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -412,10 +412,15 @@ def checkCloudFlowIsPossible(self) -> None:
...
+                Logger.log("d", "Active machine was already configured for cloud.")
+                return
 
             # Check 2: User did not already say "Don't ask me again"
             if active_machine.getMetaDataEntry("show_cloud_message", "value") is False:
```
nallath: Same as above. The default of the metadata entry is False, right (and not "value")

ianpaschal:
I've implemented your feedback, guys. Thank you.

ianpaschal:
I reverted the use of parseBool because it broke everything. I'm not sure why but honestly importing a separate function to parse booleans out of booleans is ugly anyway.

PR #5293, opened 2019-02-12
CL-1165 Improve cloud connection or not checking

ianpaschal:


ChrisTerBeke:
Seems fine to me, maybe @nallath wants to double check?

cura/Settings/MachineManager.py
```diff
@@ -541,14 +541,18 @@ def activeMachineIsGroup(self) -> bool:
         return bool(self._printer_output_devices) and len(self._printer_output_devices[0].printers) > 1
 
     @pyqtProperty(bool, notify = printerConnectedStatusChanged)
-    def activeMachineHasActiveNetworkConnection(self) -> bool:
```
diegopradogesto: This line in the plugin still uses this signature: https://github.com/Ultimaker/Cura/blob/face8d47995b76a49e760af8892a8ec0d138225f/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py#L431

How was this tested? It crashes immediately

sedwards2009: Well, it didn't crash immediately here. Obviously that line wasn't reached during the component test.

Ghostkeeper: How wasn't this line reached during the component test then? It's being called upon `_onDeviceConnectionChanged`.

ianpaschal: 🤷‍♂️ It happens. Let's not point any fingers while running a project with an unstable master branch. I can think of plenty of times I pulled and then couldn't launch Cura anymore. 😛

It's fixed now. 👌Thanks for letting me know Diego.

PR #5298, opened 2019-02-12
Arrange child-nodes (like blocker) with parents. [CURA-6120]

rburema:
See also https://github.com/Ultimaker/Uranium/pull/467


cura/Arranging/ArrangeObjectsJob.py
```diff
@@ -39,10 +39,17 @@ def run(self):
 
         arranger = Arrange.create(x = machine_width, y = machine_depth, fixed_nodes = self._fixed_nodes, min_offset = self._min_offset)
 
+        # Build set to exclude children (those get arranged together with the parents).
+        included_as_child = set([])
```
nallath: Why not just use set() (i don't really see the point of constructing a set with an empty list.)

rburema: Leftover from before I decided that this should be a set probably. I'll change it.

PR #5300, opened 2019-02-13
Add a where to buy link underneath the install button for material

nallath:
CURA-6148

### DIFF
--- a/plugins/Toolbox/resources/qml/ToolboxDetailTileActions.qml
+++ b/plugins/Toolbox/resources/qml/ToolboxDetailTileActions.qml
@@ -12,6 +12,7 @@ Column
     property bool installed: toolbox.isInstalled(model.id)
     property bool canUpdate: toolbox.canUpdate(model.id)
     property bool loginRequired: model.login_required && !Cura.API.account.isLoggedIn
+    property var packageData
 
     width: UM.Theme.getSize("toolbox_action_button").width
     spacing: UM.Theme.getSize("narrow_margin").height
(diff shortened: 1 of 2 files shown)


plugins/Toolbox/resources/qml/ToolboxDetailTileActions.qml
```diff
@@ -66,6 +67,27 @@ Column
         }
     }
 
+    Label
+    {
+        property var whereToBuyUrl:
+        {
+            var pg_name = "whereToBuy"
+            return (pg_name in packageData.links) ? packageData.links[pg_name] : undefined
```
ChrisTerBeke: This doesn't yet use the new 'type' field that was added to the API spec?

ChrisTerBeke: image

LipuFei: Hey @ChrisTerBeke , do you know what the `type`s will be for the following items (some existing `title`s in production):
 - technicalDataSheet
 - safetyDataSheet
 - printingGuidelines

I guess that they will be all lowercases with underscores?

ChrisTerBeke: image

Excerpt from https://api-staging.ultimaker.com/cura-packages/v1/spec.

LipuFei: I'll keep it for now to use `title` because the change is not in production yet. I'll put a reminder for changing to use `type`.

PR #5301, opened 2019-02-13
Explicitly check for Raft on prime-tower-brim. [CURA-5864]

rburema:
Disabling the setting is not enough, since it can still be true. Therefore, also check explicitly if Raft is the adhesion type everywhere when 'prime_tower_brim_enable' is checked for.

(Also make the setting visible for advanced users, instead of only showing up in search.)


cura/BuildVolume.py
```diff
@@ -746,7 +746,8 @@ def _updateDisallowedAreas(self):
                                 break
                         if prime_tower_collision: #Already found a collision.
                             break
-                        if ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable"):
+                        if (ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable") and
```
nallath:
```suggestion
                        if ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable") and
```
Nitpicking; The () around a if statement are not needed.

cura/BuildVolume.py
```diff
@@ -746,7 +746,8 @@ def _updateDisallowedAreas(self):
...
+                            ExtruderManager.getInstance().getResolveOrValue("adhesion_type") != 'raft'):
```
nallath:
```suggestion
                            ExtruderManager.getInstance().getResolveOrValue("adhesion_type") != "raft":
```
More nitpciking: Use " instead of '

cura/BuildVolume.py
```diff
@@ -788,7 +789,8 @@ def _computeDisallowedAreasPrinted(self, used_extruders):
                 prime_tower_x = prime_tower_x - machine_width / 2 #Offset by half machine_width and _depth to put the origin in the front-left.
                 prime_tower_y = prime_tower_y + machine_depth / 2
 
-            if ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable"):
+            if (ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable") and
```
nallath:
```suggestion
            if ExtruderManager.getInstance().getResolveOrValue("prime_tower_brim_enable") and
```

cura/BuildVolume.py
```diff
@@ -788,7 +789,8 @@ def _computeDisallowedAreasPrinted(self, used_extruders):
...
+                ExtruderManager.getInstance().getResolveOrValue("adhesion_type") != 'raft'):
```
nallath:
```suggestion
                ExtruderManager.getInstance().getResolveOrValue("adhesion_type") != "raft":
```

cura/BuildVolume.py
```diff
@@ -1036,7 +1038,9 @@ def getEdgeDisallowedSize(self):
             # We don't create an additional line for the extruder we're printing the skirt with.
             bed_adhesion_size -= skirt_brim_line_width * initial_layer_line_width_factor / 100.0
 
-        elif adhesion_type == "brim" or self._global_container_stack.getProperty("prime_tower_brim_enable", "value"):
+        elif (adhesion_type == "brim" or
+                (self._global_container_stack.getProperty("prime_tower_brim_enable", "value") and
+                    self._global_container_stack.getProperty("adhesion_type", "value") != 'raft')):
```
nallath:
```suggestion
                    self._global_container_stack.getProperty("adhesion_type", "value") != "raft")):
```

cura/Settings/ExtruderManager.py
```diff
@@ -265,7 +265,8 @@ def getUsedExtruderStacks(self) -> List["ContainerStack"]:
 
         # The platform adhesion extruder. Not used if using none.
         if (global_stack.getProperty("adhesion_type", "value") != "none" or
-            global_stack.getProperty("prime_tower_brim_enable", "value")):
+            global_stack.getProperty("prime_tower_brim_enable", "value") and
```
nallath: Shouldn't it be adhesion_type != None or (prime_tower_brim_enable && adhesion_type != raft) ?

rburema: Yes.

cura/BuildVolume.py
rburema: Turns out I did those brackets because line-breaks at that position are otherwise not allowed in python without a `\`, which I think is uglier than the brackets. Had to revert these changes, except for the 'raft' => "raft" ones.

PR #5305, opened 2019-02-14
CL-1250 set cloud build plate

ianpaschal:
This makes sure cloud printers set their build plate so Cura doesn't try to handle an empty build plate.

This includes a fix where the `buildplate` (e.g. "glass") was misidentified as `buildplate_name` (e.g. "Glass").

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
@@ -55,7 +55,6 @@ class CloudClusterPrinterStatus(BaseCloudModel):
     #  \param controller - The controller of the model.
     def createOutputModel(self, controller: PrinterOutputController) -> PrinterOutputModel:
         model = PrinterOutputModel(controller, len(self.configuration), firmware_version = self.firmware_version)
-        model.updateBuildplate(self.build_plate.type)
         self.updateOutputModel(model)
         return model
 
@@ -66,6 +65,7 @@ class CloudClusterPrinterStatus(BaseCloudModel):
         model.updateName(self.friendly_name)
         model.updateType(self.machine_variant)
         model.updateState(self.status if self.enabled else "disabled")
+        model.updateBuildplate(self.build_plate.type)
 
         for configuration, extruder_output, extruder_config in \
                 zip(self.configuration, model.extruders, model.printerConfiguration.extruderConfigurations):
--- a/plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
@@ -48,7 +48,7 @@ class CloudClusterPrinterStatus(BaseCloudModel):
         self.maintenance_required = maintenance_required
         self.firmware_update_status = firmware_update_status
         self.latest_available_firmware = latest_available_firmware
-        self.build_plate = self.parseModel(CloudClusterBuildPlate, build_plate)
+        self.build_plate = self.parseModel(CloudClusterBuildPlate, build_plate) if build_plate else None
         super().__init__(**kwargs)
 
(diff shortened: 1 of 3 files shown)


ianpaschal:
Jenkins killed itself and failed to clone the repo I guess.

Ghostkeeper:
That happens from time to time. Luckily you can also run the tests locally :)

ianpaschal:
@Ghostkeeper I know. But that's why I make a note that this red X is not actually because of failing tests. 😉

ChrisTerBeke:
You can do a force push without any commit changes to re-trigger Jenkins.

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, enabled: bool, firmware_version: str, friendly_name: str, ip_
     #  \param controller - The controller of the model.
     def createOutputModel(self, controller: PrinterOutputController) -> PrinterOutputModel:
         model = PrinterOutputModel(controller, len(self.configuration), firmware_version = self.firmware_version)
+        model.updateBuildplate(self.build_plate.type)
```
ChrisTerBeke: Ah, so this was the missing key :)

diegopradogesto: I would do this better in the `updateOutputModel()` method, where the output model is actually updated

ianpaschal: That's true. I've changed it.

ianpaschal: After 2-3 days of trying to track that bug down my eyes went blind and I didn't even notice the other stuff being set a few lines further down the file

ChrisTerBeke: we like our classes and methods coherent :)

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
```diff
@@ -65,7 +65,13 @@ def updateOutputModel(self, model: PrinterOutputModel) -> None:
         model.updateName(self.friendly_name)
         model.updateType(self.machine_variant)
         model.updateState(self.status if self.enabled else "disabled")
-        model.updateBuildplate(self.build_plate.type)
+
+        # Make sure to set the build plate even though we don't use it. Since it's optional, use
+        # glass as a default
+        if self.build_plate:
```
ChrisTerBeke: The cloud always returns a build place from the API.

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
```diff
@@ -65,13 +65,7 @@ def updateOutputModel(self, model: PrinterOutputModel) -> None:
...
         model.updateType(self.machine_variant)
         model.updateState(self.status if self.enabled else "disabled")
-
-        # Make sure to set the build plate even though we don't use it. Since it's optional, use
-        # glass as a default
-        if self.build_plate:
-            model.updateBuildplate(self.build_plate.type)
-        else:
-            model.updateBuildplate("glass")
+        model.updateBuildplate(self.build_plate.type or "glass")
```
ChrisTerBeke: mm like this it will fail if self.build_plate is None. Should be `self.build_plate.type if self.build_plate else "glass"`.

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterPrinterStatus.py
diegopradogesto: Teamwork!

PR #5306, opened 2019-02-14
Don't explode the log with errors when there's no cloud response

ianpaschal:


plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -159,6 +159,9 @@ def _addCallback(self,
                      model: Type[CloudApiClientModel],
                      ) -> None:
         def parse() -> None:
+            # Don't try to parse the reply if we didn't get one
+            if reply.attribute(QNetworkRequest.HttpStatusCodeAttribute) is None:
+                return
             status_code, response = self._parseReply(reply)
             self._anti_gc_callbacks.remove(parse)
             return self._parseModels(response, on_finished, model)
```
nallath: Nitpicking now that I see it; self._parseModels never returns anything but None, so there is no real need to return the value again.

ianpaschal: In Python does that mean it should `return None` or is a `return` enough? Not sure how that equates to returning `null` vs. returning `undefined` in JS for example.

nallath: `return None` is the same as `return` in python (it's even the same as not returning anything at all)

ianpaschal: Ok I'll update it. Thanks for the tip!

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -159,9 +159,13 @@ def _addCallback(self,
                      model: Type[CloudApiClientModel],
                      ) -> None:
         def parse() -> None:
+            # Don't try to parse the reply if we didn't get one
+            if reply.attribute(QNetworkRequest.HttpStatusCodeAttribute) is None:
```
ChrisTerBeke: We might wanna log at least something at the debug level that we didn't get a reply? Because this is never expected behaviour.

Ghostkeeper:
Developers, see issue CURA-6624.

PR #5312, opened 2019-02-15
Add few things to 'About' dialog

diegopradogesto:
Add native rendering to the labels

The 'close' button was also changed to use the common style.

### DIFF
--- a/resources/qml/Dialogs/AboutDialog.qml
+++ b/resources/qml/Dialogs/AboutDialog.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Ultimaker B.V.
+// Copyright (c) 2019 Ultimaker B.V.
 // Cura is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.2
@@ -6,6 +6,7 @@ import QtQuick.Controls 1.1
 import QtQuick.Window 2.1
 
 import UM 1.1 as UM
+import Cura 1.1 as Cura
 
 UM.Dialog
 {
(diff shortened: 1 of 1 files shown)


fieldOfView:
AFAIK the native rendering is already implied by Label in QTQuick Controls 1. But it does not hurt to have it twice.

diegopradogesto:
Ah, OK, I didn't know it. I saw a lot of `Label`s without it and I always tend to add it when I see it as a boy scouting process.

fieldOfView:
Note that this PR places a Cura-themed close button on a system-themed background. This is not necessarily "correct". I would say you either make the whole dialog Cura-themed, or make all of it (except maybe the logo, because that is defined in the theme) system-themed. Hisotrically, any dialog that was opened from Cura would be system-themed, but that rule has been abandoned a little while ago (with marketplace being Cura-themed).

Ghostkeeper:
We're slowly working towards making all dialogues Cura-themed. Next up is the preferences window.

That being said, I think we should change the whole dialogue at once rather than only one button of it, so that it would look good for release even in between. So I'm inclined to reject this PR. It only adds an unnecessary font rendering property and a button that would look out of place.

Ghostkeeper:
Nallath agreed. We're closing this.

Ghostkeeper:
I'll leave it up to @diegopradogesto to delete the branch unless he wants to argue this decision.

diegopradogesto:
OK, I agree! It's better to do it all at the same time.

PR #5314, opened 2019-02-15
Support gyroid support.

smartavionics:
Haven't actually tried printing gyroid support but at least now it can be evaluated.

As discussed in https://github.com/Ultimaker/CuraEngine/issues/991.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -3811,7 +3811,8 @@
                         "triangles": "Triangles",
                         "concentric": "Concentric",
                         "zigzag": "Zig Zag",
-                        "cross": "Cross"
+                        "cross": "Cross",
+                        "gyroid": "Gyroid"
                     },
                     "default_value": "zigzag",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Yeah I thought I might do this in between somewhere, but you were ahead of me! Thanks.

It slices all right. I'm making a test print because I'm curious but that's a print quality / tuning issue if the test print goes wrong.

smartavionics:
It should be OK, I think it needs to have the lines connected because otherwise you tend to get very ragged ends to the waves.

Ghostkeeper:
It's a bit like grid or triangle infill. Kind of hard to remove, kind of rigid. But similar properties. For a fairly thin strip of support it works kind of nicely to squeeze the support and remove it that way (similar to zigzag).

smartavionics:
Because it is so strong and stable maybe one can get away with reducing the density compared to the other support infills?

Ghostkeeper:
Or maybe even the connected lines.

PR #5327, opened 2019-02-19
Adjust description of Skin Overlap (Percentage). [CURA-6140]

rburema:


resources/definitions/fdmprinter.def.json
```diff
@@ -1792,7 +1792,7 @@
                 "skin_overlap":
                 {
                     "label": "Skin Overlap Percentage",
-                    "description": "The amount of overlap between the skin and the walls as a percentage of the skin line width. A slight overlap allows the walls to connect firmly to the skin. This is a percentage of the average line widths of the skin lines and the innermost wall.",
+                    "description": "Adjust the position where skin is extruded relative to the walls, as a percentage of the line widths of the skin lines and the innermost wall. A slight overlap allows the walls to connect firmly to the skin. Note that, given an equal skin and wall line-width, any percentage over 50% may already cause any skin to go past the wall, because at that point the position of the nozzle of the skin-extruder may already reach past the middle of the wall.",
```
Ghostkeeper: I don't really like the wording of `Adjust the position where the skin is extruded relative to the walls`. It sounds a bit like this could make your skin shift 2mm to the left or something.

Maybe something more specific would be to "Adjust the amount of overlap between the walls and the endpoints of the centerlines of the skin" but that is a fairly technical description. Any ideas?

rburema: How about 'Adjust the amount of overlap between the walls and (the endpoints of) the skin-centerlines.'?

PR #5328, opened 2019-02-19
CL-1245 Don't throw exceptions for non-severe stuff

ianpaschal:
Contributes to CL-1245

### DIFF
--- a/cura/OAuth2/AuthorizationHelpers.py
+++ b/cura/OAuth2/AuthorizationHelpers.py
@@ -92,7 +92,7 @@ class AuthorizationHelpers:
             })
         except requests.exceptions.ConnectionError:
             # Connection was suddenly dropped. Nothing we can do about that.
-            Logger.logException("e", "Something failed while attempting to parse the JWT token")
+            Logger.log("w", "Something failed while attempting to parse the JWT token")
             return None
         if token_request.status_code not in (200, 201):
             Logger.log("w", "Could not retrieve token data from auth server: %s", token_request.text)
--- a/cura/OAuth2/AuthorizationService.py
+++ b/cura/OAuth2/AuthorizationService.py
@@ -53,11 +53,7 @@ class AuthorizationService:
     #   \sa _parseJWT
     def getUserProfile(self) -> Optional["UserProfile"]:
         if not self._user_profile:
-            try:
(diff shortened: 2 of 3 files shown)


ianpaschal:
@ChrisTerBeke Really want to hear your opinions on this. I'm not understanding why we throw for something as normal as a connectivity issue. Makes more sense to me to simply return `None` and show the user as not-logged-in.

Ghostkeeper:
It's considered more semantic to throw at unexpected behaviour. Because you're then communicating through your API what went wrong, and the user of your functions can catch the exception explicitly. If you can have errors for different reasons (e.g. server not responding vs. user not having internet) then you can handle them in different ways. And you won't need to document why your return type could be `None`.

Depending on the return value though, it may as well be semantic to return `None`. For instance, the `getActiveGlobalStack` function could return `None` if there is no active global stack because it makes semantic sense, there is none so you return None.

ChrisTerBeke:
I guess it's fine not to throw this particular issue, but there should be a retry strategy or a user visible message telling them to sing in again, otherwise you're just signed out and nothing happens.

As @Ghostkeeper indicated, with throwing you can explicitly state which cases to handle and which ones should be handled at a higher level. In this case, the JWT parsing can go wrong because of different reasons (e.g. the user not being signed in in the first place, or an actual API error response because something was invalid). You wanna handle these in quite different ways later in your code.

ianpaschal:
@Ghostkeeper

>It's considered more semantic to throw at unexpected behaviour.

I get that. In this case though, being connected or not doesn't feel like unexpected behavior. If the JWT is actually... an instance of `GlobalStack` that would be unexpected. But it's very likely users fire up Cura without an internet connection.

@ChrisTerBeke 

>I guess it's fine not to throw this particular issue, but there should be a retry strategy or a user visible message telling them to sing in again, otherwise you're just signed out and nothing happens.

Disagree here. If you're actually trying to work offline that would get really annoying. By simply returning `None` instead of throwing, no weird behavior happens: the user is simply no longer logged in and the log in panel in the corner works as expected so they can try again. If Cura was really a real-time web-connected app like a chat app, I'd agree this should behave differently, but I think given how Cura is used this is a cleaner solution.

ianpaschal:
Closing in favor of the smaller fix in #5336

PR #5331, opened 2019-02-19
Fix the path to the pop up images in the monitor tab.

sedwards2009:
This may fix the images appearing in the messages for Cura Connect when running on Windows.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
+++ b/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
@@ -455,7 +455,7 @@ class UM3OutputDevicePlugin(OutputDevicePlugin):
             self._start_cloud_flow_message = Message(
                 text = i18n_catalog.i18nc("@info:status", "Send and monitor print jobs from anywhere using your Ultimaker account."),
                 lifetime = 0,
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -455,7 +455,7 @@ def _onCloudFlowPossible(self) -> None:
             self._start_cloud_flow_message = Message(
                 text = i18n_catalog.i18nc("@info:status", "Send and monitor print jobs from anywhere using your Ultimaker account."),
                 lifetime = 0,
-                image_source = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "resources", "svg",
+                image_source = "file:///" + os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "resources", "svg",
```
fieldOfView: Wouldn't QUrl().toLocalFile() be a better way to make a URL? Mixing the os.path path (which is backward-slash-based on windows) with a file:/// url is bound to result in issues.

sedwards2009: `toLocalFile()` goes the wrong direction, but getting QUrl to do it is indeed a good idea. `QUrl.fromLocalFile()` looks like the one we want.

PR #5333, opened 2019-02-19
Update mendel90.def.json

wgaonar:
The main modification is to set absolute extrusion because the mendel90 firmware is setting by default with relative extrusion mode. The other change is about to set up the machine center as true


resources/definitions/mendel90.def.json
```diff
@@ -23,10 +23,10 @@
     "overrides": {
         "machine_name": { "default_value": "Mendel90" },
         "machine_start_gcode": {
-            "default_value": "G21 ;metric values\nG90 ;absolute positioning\nG92 E0 ;zero the extruded length\nM107 ;start with the fan off\nG1 X90 Y200 F6000 ;go to the middle of the front\nG1 Z0.05 ;close to the bed\nG1 Z0.3 ;lift Z\n"
+            "default_value": "G21 ;metric values\nG90 ;absolute positioning\nM82 ;absolute extrusion\nM107 ;start with the fan off\nG28 ;home\nG92 E0 ;zero the extruded length\n;M109 S175; preheat the extruder temp & wait\n;G1 E120 F50; extrude filament for insert it\n;G92 E0 ;zero the extruded length\nM140 S{material_bed_temperature_layer_0} ; set the bed temperature and continue on\nG1 X-50 Y98 F9000 ;go to the left of the top\nG1 Z0.05 ; close to the bed\nM104 S{material_print_temperature_layer_0}; pre-heat the extruder continue on\nM190 S{material_bed_temperature_layer_0} ;set the bed temp & wait\nM109 S{material_print_temperature_layer_0};set the extruder temp for layer 0 & wait\nG92 E0 ;zero the extruded length\nG1 X50 E10 F300 ; make a thick line to prime extruder\nG92 E0 ; reset extruder\nG1 E-4 F1800\nG1 Z0.3 ;lift Z\n"
```
Ghostkeeper: For readability, this produces the following g-code:

```
M82 ;absolute extrusion mode
G21 ;metric values
G90 ;absolute positioning
M82 ;absolute extrusion
M107 ;start with the fan off
G28 ;home
G92 E0 ;zero the extruded length
;M109 S175; preheat the extruder temp & wait
;G1 E120 F50; extrude filament for insert it
;G92 E0 ;zero the extruded length
M140 S60 ; set the bed temperature and continue on
G1 X-50 Y98 F9000 ;go to the left of the top
G1 Z0.05 ; close to the bed
M104 S200; pre-heat the extruder continue on
M190 S60 ;set the bed temp & wait
M109 S200;set the extruder temp for layer 0 & wait
G92 E0 ;zero the extruded length
G1 X50 E10 F300 ; make a thick line to prime extruder
G92 E0 ; reset extruder
G1 E-4 F1800
G1 Z0.3 ;lift Z
```

I'm wondering, there's 3 lines there that are commented out. Why keep them in? I think they are unnecessary. The pre-heating is set to a fixed temperature which seems strange, since you could use `{material_print_temperature_layer_0}` but that is done correctly later. The extrusion and zeroing is done later, too.

wgaonar: Hello. 

- The 3 lines aren't necessary. You are right. 

- The suggestion that you give about using prefixed words is great. I want to keep these in order to prime the extruder.  I am going to modify the code according to it.

wgaonar: The 3 lines were deleted. I committed the changes. I hope it has been the correct way to do that

Ghostkeeper: Yeah it was. Looks good now.

PR #5336, opened 2019-02-20
CL-1245 Change exception logging when no connection

ianpaschal:
Contributes to CL-1245


ianpaschal:
Fixes by Cura team prevented the actual crash the but UM3 Networking plugin still spits out lots of `[ERROR]` logs despite handling the error and logging the user out if the JWT can't be parsed. This prevents the logs from filling up with errors when it's not severe.

ianpaschal:
Tried to copy my changes from #5328 but they now crash (as that branch does so forget them). The changes here fix the problem.

PR #5339, opened 2019-02-20
Use new property names for cloud connection detection

ianpaschal:


plugins/MonitorStage/MonitorMain.qml
```diff
@@ -11,7 +11,7 @@ Rectangle
 {
     id: viewportOverlay
 
-    property bool isConnected: Cura.MachineManager.activeMachineHasActiveNetworkConnection || Cura.MachineManager.activeMachineHasActiveCloudConnection
+    property bool isConnected: Cura.MachineManager.activeMachineIsUsingCloudConnection
```
ChrisTerBeke: This should look at any network connection, not just cloud.

ianpaschal: Ah yes. True. Good catch.

PR #5354, opened 2019-02-24
Adds a printer definition for the FLSUN QQ

radfahrer:
This is my first attempt to add a new printer to Cura. I think I've followed all the instructions, but Cura tells me that the profile is corrupt. Perhaps someone could help me.

### DIFF
--- a/resources/definitions/flsun_qq.def.json
+++ b/resources/definitions/flsun_qq.def.json
@@ -7,8 +7,6 @@
 		"manufacturer": "FLSUN",
 		"author": "Daniel Green",
 		"file_formats": "text/x-gcode",
-		"icon": "icon_ultimaker2",
-		"platform_offset": [0, 0, 0],
 		"machine_extruder_trains": {
 			"0": "flsun_qq_extruder"
 		}
--- a/resources/definitions/flsun_qq.def.json
+++ b/resources/definitions/flsun_qq.def.json
@@ -1,0 +1,48 @@
+{
+	"version": 1,
+	"name": "FLSUN QQ",
+	"inherits": "fdmprinter",
+	"metadata": {
+		"visible": true,
+		"manufacturer": "FLSUN",
+		"author": "Daniel Green",
+		"file_formats": "text/x-gcode",
+		"icon": "icon_ultimaker2",
+		"platform_offset": [0, 0, 0],
+		"machine_extruder_trains": {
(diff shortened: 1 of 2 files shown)


fieldOfView:
In both files, `version ` has to be `2`.

radfahrer:
Why 2?

fieldOfView:
It refers to the version of the format, not the version of the data.

resources/definitions/flsun_qq.def.json
```diff
@@ -0,0 +1,48 @@
...
+		},
+		"machine_width": {
+			"default_value": 240
+		},
+		"machine_height": {
+			"default_value": 285
+		},
+		"machine_depth": {
+			"default_value": 240
+		},
```
Ghostkeeper: [This page](https://all3dp.com/1/flsun-qq-3d-printer-review/) claims that the printer has a printing area of 260x260x370mm. You might want to review this. Or did you experimentally find that the usable print area is smaller?

resources/definitions/flsun_qq.def.json
```diff
@@ -0,0 +1,48 @@
...
+	"version": 2,
+	"name": "FLSUN QQ",
+	"inherits": "fdmprinter",
+	"metadata": {
+		"visible": true,
+		"manufacturer": "FLSUN",
+		"author": "Daniel Green",
+		"file_formats": "text/x-gcode",
+		"icon": "icon_ultimaker2",
+		"platform_offset": [0, 0, 0],
```
Ghostkeeper: This offset is useless if there is no platform mesh set. It's also set to the default value. It doesn't hurt to have it in here, but you might as well remove it.

Ghostkeeper:
Developers, see issue CURA-6250.

resources/definitions/flsun_qq.def.json
fieldOfView: icon is useless as well; it does not exist and it is not used

resources/definitions/flsun_qq.def.json
radfahrer: These are the settings recommended by the owners manual.

radfahrer:
I've removed the two fields that were pointed out as not necessary.

Ghostkeeper:
Thanks. Seems to work fine, then.

PR #5356, opened 2019-02-25
Feature: "Wipe Script"

vgribinchuk:
Feature: "Wipe Script"
As material is extruded and the head runs across the current layer being printed, material accumulates on the nozzle tip. This accumulated material effects the quality of the printed part. To address this, the nozzle tip must be cleaned by wiping it across a brush to the sides of the build platform. Following feature allows to insert such wipe moves for nozzles between layer switching.

Wipe always occur in between layers. It only occur once specified "Material Volume Between Wipes" were extruded (starting a new layer doesn't restart the counter of extruded material, only extruder switching restarts the counter, assuming some native wiping performs during extruder switching).

Wipe script starts with "WIPE_SCRIPT_BEGIN" comment tag and ends with "WIPE_SCRIPT_END" comment tag.
The overall script steps:
1. Perform retraction (if enabled)
2. Perform z hop (if enabled)
3. Go to Wipe Brush X Position
4. Make wipe moves (repeating "Wipe Repeat Count" times)
5. Return back to position before wipe
6. Return z hop (if enabled)
7. Perform unretraction (if enabled)
8. Write optional delay

Wipe script is configured with set of parameters, like retraction/unrectraction speed and amount, z hop, etc.
New settings:
- "Wipe Nozzle Between Layers"
- "Material Volume Between Wipes"
- "Wipe Retraction Enable"
- "Wipe Retraction Distance"
- "Wipe Retraction Extra Prime Amount"
- "Wipe Retraction Speed"
- "Wipe Pause"
- "Wipe Z Hop When Retracted"
- "Wipe Z Hop Height"
- "Wipe Hop Speed"
- "Wipe Brush X Position"
- "Wipe Repeat Count"
- "Wipe Move Distance"

This PR is frontend settings for backend implementation: https://github.com/Ultimaker/CuraEngine/pull/1000


Ghostkeeper:
Developers, see issue CURA-6249.

Ghostkeeper:
Note for testers: We should be careful how this works together with the "Retract at Layer Change" setting.

resources/definitions/fdmprinter.def.json
```diff
@@ -6954,6 +6954,203 @@
                     "type": "float",
                     "enabled": "bridge_settings_enabled and bridge_enable_more_layers",
                     "settable_per_mesh": true
+                },
+                "clean_between_layers":
```
rburema: In the engine (see my review there) you have made it so that `clean_between_layers` is mutually exclusive with `retract_at_layer_change`. This is fine, but should be handled here; that is, the values both `"enabled"`-fields of those options should be set to false if the other is true.

rburema: Another thing: If you want these to show up not just when specifically searching for it, but also browsing the settings, you need to put them in one of the `.cfg` files in `resources/setting_visibility`. I'd recommend putting all of them in 'expert' though.

maukcc:
Would it be possible to do this every x amount of layers, or even better : the layer change after x amount of time after the previous wipe.

vgribinchuk:
> Would it be possible to do this every x amount of layers, or even better : the layer change after x amount of time after the previous wipe.
> 
> Also some multi extruder printers, already have the wiping in the extruder start/end gcode. Is it possible to disable it for a layer that has a toolchange?

Hello, @maukcc !
Thank you for the review, I understood your points. Please, see my comments below:
- Some layers could contain a few g-code paths, some could contain a large amount of g-code paths, which results in different amount of extruded material per each layer. Doing wiping every x amount of layers could lead to unnecessary wiping, if previous x layers were really small. On the other hand, if those layers were large, a lot of material would accumulate on the nozzle tip before wiping occur. This is why there is implemented the following logic, which provides flexibility: perform wiping only if at least specified "Material Volume Between Wipes" were extruded.

- The fact that some multi extruder printers already have the wiping in the extruder start/end gcode is considered by Wipe Script algorithm. Toolchange will restart the counter of "Material Volume Between Wipes", that will disable wiping for a layer that has a toolchange.

Regards,
Vlad

maukcc:
Perfect ! Vlad
Thanks

best regards
Jos

resources/definitions/fdmprinter.def.json
vgribinchuk: Logic in the engine was left as it is. But description for this setting was clarified to make it more obvious for user how this will interact with `retract_at_layer_change` behavior.

vgribinchuk: Thank you! All of them were put into exert.cfg.

PR #5357, opened 2019-02-25
Bug: on Tevo Tornado nozzle and bed are turned off on start.

h0x91b:
Changed "[extruder0_temperature]" to "{material_print_temperature}", because "[extruder0_temperature]" is replaced by an empty string in actual g-code generated by Cura 3.6.
Generated g-code is turn off heating at all on mine Tornado right before cleaning nozzle which is harmful and unacceptable...

Bad g-code example:

M104 S
M109 S
M109 R


h0x91b:
`[extruder0_temperature]` is variable in simplify3D. Probably g-code was copy-pasted.

Ghostkeeper:
Nice, thanks.

I'm going to change that to `material_print_temperature_layer_0` if you don't mind, so that it's ready for the first layer if that temperature is different than for the rest of the layers.

Ghostkeeper:
My change is here: https://github.com/Ultimaker/Cura/commit/b723beb4353dbf2e4c3e16f2b469b86a4e3bfb22

h0x91b:
Nice, I am was copying this variable from another random definition JSON, didn't know about `
material_print_temperature_layer_0`, thanks.

Ghostkeeper:
If you copied it from another JSON, maybe we can also change it there :stuck_out_tongue:

PR #5366, opened 2019-02-27
Adding Discov3ry Complete Paste Printer to Cura.

afinkle:
Definition, extruder, default material, and nozzle variants.


Ghostkeeper:
Interesting printer definition. Definitely needed some tuning, I imagine! Thanks for the profiles.

But the material profile seems to be missing still, some file called `structur3d_dap100silicone.xml.fdm_material`?

afinkle:
I added the material profile to fdm_materials. I overlooked that it wasn't inside the Cura repository. My request can be found here: https://github.com/Ultimaker/fdm_materials/pull/73

resources/variants/structur3d_discov3ry1_complete_um2plus_0.20.inst.cfg
```diff
@@ -0,0 +1,14 @@
+[general]
+name = 0.20mm (Clear)
+version = 4
+definition = structur3d_discov3ry1_complete_um2plus
+
+[metadata]
+setting_version = 5
```
Ghostkeeper: Currently the Cura master version (set for 4.1 release) requires `setting_version = 7`. This field is used to detect whether the profile needs to be upgraded. And in your case, it needs to be upgraded. However it appears that your profiles don't use any of the settings whose name or functionality has changed in between, so we can just increment this number. I'll do that for you.

PR #5374, opened 2019-02-28
Add TiZYX Evy Dual Extruder 3D printer

ValentinPitre:
Hello,
Thanks for adding this 3D printer.

It has the same platform as tizyx_k25.

For optimal running, the "Classic Extruder" variant should be associated with extruder 1 and "Direct Drive" with extruder 2.

Yours faithfully


resources/definitions/tizyx_evy_dual.def.json
```diff
@@ -0,0 +1,60 @@
...
...red", "ultimaker_abs_silver-metallic", "ultimaker_abs_white", "ultimaker_abs_yellow", "ultimaker_bam", "ultimaker_cpe_black", "ultimaker_cpe_blue", "ultimaker_cpe_dark-grey", "ultimaker_cpe_green", "ultimaker_cpe_light-grey", "ultimaker_cpe_plus_black", "ultimaker_cpe_plus_transparent", "ultimaker_cpe_plus_white", "ultimaker_cpe_red", "ultimaker_cpe_transparent", "ultimaker_cpe_white", "ultimaker_cpe_yellow", "ultimaker_nylon_black", "ultimaker_nylon_transparent", "ultimaker_pc_black", "ultimaker_pc_transparent", "ultimaker_pc_white", "ultimaker_pla_black", "ultimaker_pla_blue", "ultimaker_pla_green", "ultimaker_pla_magenta", "ultimaker_pla_orange", "ultimaker_pla_pearl-white", "ultimaker_pla_red", "ultimaker_pla_silver-metallic", "ultimaker_pla_transparent", "ultimaker_pla_white", "ultimaker_pla_yellow", "ultimaker_pp_transparent", "ultimaker_pva", "ultimaker_tough_pla_black", "ultimaker_tough_pla_green", "ultimaker_tough_pla_red", "ultimaker_tough_pla_white", "ultimaker_tpu_black", "ultimaker_tpu_blue", "ultimaker_tpu_red", "ultimaker_tpu_white", "verbatim_bvoh_175", "Vertex_Delta_ABS", "Vertex_Delta_PET", "Vertex_Delta_PLA", "Vertex_Delta_TPU", "zyyx_pro_flex", "zyyx_pro_pla" ],
```
Ghostkeeper: You're probably not incompatible with all of these materials, are you?

Many of these are 2.85mm materials anyway so they wouldn't be shown anyway.

resources/definitions/tizyx_evy_dual.def.json
```diff
@@ -0,0 +1,60 @@
...
+        "machine_start_gcode":
+        {
+            "default_value": "M82\nG90\nG28 X\nG28 Y\nG28 Z\nG29\nG91\nG1 Z0\nG90\nM82\nG92 E0\nG1 X125 Y245 F3000\nG1 Z0"
+        },
+        "machine_end_gcode":
+        {
+            "default_value": "M104 S0\nM140 S0\nG91\nG1 E-5 F300\nG1 Z+3 F3000\nG1 Y245 F3000\nM84"
+        },
+        "layer_height": { "maximum_value": "(0.8 * min(extruderValues('machine_nozzle_size')))" },
+        "layer_height_0": { "maximum_value": "(0.8 * min(extruderValues('machine_nozzle_size')))" }
```
Ghostkeeper: Typically Cura doesn't actually prevent slicing if the result would just be a bad quality print, but only if it's physically impossible (like speed < 0). So please change these to `maximum_value_warning` instead, so that the user gets a warning that the quality might not be good but is allowed to slice.

resources/definitions/tizyx_evy_dual.def.json
ValentinPitre: No but It's a TiZYX’s wish to exclude these materials in order to simplify the user interface

Ghostkeeper: Then at least we need to remove the 2.85mm materials from this list, because that's unnecessary.

PR #5375, opened 2019-03-01
CL-1274 Don't use __file__ in the monitor tab

ianpaschal:
^What it says on the box.

I also sorted the Uranium imports alphabetically... just some minor Boy Scouts of America type stuff.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
@@ -85,8 +85,8 @@ class CloudOutputDevice(NetworkedPrinterOutputDevice):
 
         # We use the Cura Connect monitor tab to get most functionality right away.
         self._monitor_view_qml_path = os.path.join(
-            PluginRegistry.getPluginDirectory(),
-            "UM3NetworkPrinting", "resources", "qml", "MonitorStage.qml"
+            PluginRegistry.getInstance().getPluginPath("UM3NetworkPrinting"),
+            "resources", "qml", "MonitorStage.qml"
         )
 
         # Trigger the printersChanged signal when the private signal is triggered.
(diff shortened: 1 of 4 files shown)


fieldOfView:
Out of interest, what is wrong with using __file__ here? 

Your alternative hardcodes the name of the plugin(folder) in several places throughout the plugin, which could introduce a bug if the folder of the plugin gets renamed.

nallath:
The` __file__` gets hardcoded by the windows installer. This only has an impact for non plugin code (in which case it uses the x:/whateverweuse instead of the actual location of the file)

I'm not quite sure if this solves the issue, because as fieldOfView correctly mentiones, it does hardcode the plugin name (and yes, I'm aware that we've done it before, but that also doesn't make it right :P)

Ghostkeeper:
We have several problems with `__file__`. One is that if it's in the Cura base (or gets moved there in the future) then these files get packaged in the Cura.exe file on Windows, and don't have a file path any more. Our build system then changes the value of `__file__` to whatever it was on the build server, which is on the `X:` drive which probably doesn't exist on the user's computer (but at least is no longer the drive on which some people have their floppy disks).

That doesn't hold for plug-ins though since they are never packed (so far) in an executable. But another is that it's unknown whether the files relative to the .py file have been loaded yet. So for plug-ins it's better to use `UM.PluginRegistry.PluginRegistry.getPluginPath(self.getPluginId())` in order to get the root directory of the plug-in. That function handles those cases.

ianpaschal:
Hmm OK. Good to know. I will use `UM.PluginRegistry.PluginRegistry.getPluginPath(self.getPluginId()`)

Ghostkeeper:
Mind that `self.getPluginId()` refers to a function inside the `PluginObject`. So you may need something like `output_device = UM3OutputDevicePlugin()`, then `output_device.getPluginId()`.

ianpaschal:
@Ghostkeeper I'm running into problems with a circular dependency trying to create another instance of the plugin inside some of the other classes. The irritations/ugliness of this approach are significantly heavier IMO than simply using the string `"UM3NetworkPrinting"` so we're going to go with that.

Ghostkeeper:
Yeah I didn't intend you to create a whole new instance of the plug-in. I just wanted to indicate that `output_device` contains that class in it since Python doesn't have typing to its variables. This is fine.

PR #5388, opened 2019-03-04
[WIP] CL-1266 Cloud association for manual IPs

ianpaschal:
I'm not a big fan of using the manual IP address as network key, but handling it separately requires a _lot_ of changes, which I hear might be irrelevant in 4.1 with some reworking of how the printer is added.


plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -140,9 +140,17 @@ def clusterData(self, value: CloudClusterResponse) -> None:
...
-        # A network key looks like "ultimakersystem-aabbccdd0011._ultimaker._tcp.local."
+        # Typically, a network key looks like "ultimakersystem-aabbccdd0011._ultimaker._tcp.local."
         # the host name should then be "ultimakersystem-aabbccdd0011"
-        return network_key.startswith(self.clusterData.host_name)
+        if network_key.startswith(self.clusterData.host_name):
+            return True
+        
+        # However, for manually added printers, the local IP address is used in lieu of a proper
+        # network key, so check for that as well
+        if network_key == self.clusterData.host_internal_ip:
```
ChrisTerBeke: Maybe something like network_key.find(self.clusterData.host_internal_ip) to be more robust for cases like when the IP suddenly contains http:// in front of it or something.

ianpaschal: Fair point 👍

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterResponse.py
```diff
@@ -16,13 +16,14 @@ class CloudClusterResponse(BaseCloudModel):
     #  \param status: The status of the cluster authentication (active or inactive).
     #  \param host_version: The firmware version of the cluster host. This is where the Stardust client is running on.
     def __init__(self, cluster_id: str, host_guid: str, host_name: str, is_online: bool, status: str,
-                 host_version: Optional[str] = None, **kwargs) -> None:
+                 host_internal_ip: str, host_version: Optional[str] = None, **kwargs) -> None:
```
ChrisTerBeke: This should also be an Optional[str] as the field is not available in clusters running 5.2 (Beta) firmware.

nallath:
I doubt this will change in 4.1 though. The whole point of having the manual IP is that you don't know the host name.

Ghostkeeper:
No, indeed, the new UI for this in 4.1 has a menu to add a printer by IP. This is important, because for me at home the printer doesn't announce itself 50% of the time.

plugins/UM3NetworkPrinting/src/Cloud/Models/CloudClusterResponse.py
```diff
@@ -16,13 +16,14 @@ class CloudClusterResponse(BaseCloudModel):
     #  \param status: The status of the cluster authentication (active or inactive).
     #  \param host_version: The firmware version of the cluster host. This is where the Stardust client is running on.
     def __init__(self, cluster_id: str, host_guid: str, host_name: str, is_online: bool, status: str,
-                 host_version: Optional[str] = None, **kwargs) -> None:
+                 host_internal_ip: Optional[str], host_version: Optional[str] = None, **kwargs) -> None:
```
sedwards2009: The `Optional` here is for Mypy and that is fine. It means you can pass in `None` as a value. But if you want people to be able to call this constructor without passing any value for `host_internal_ip`, then you need to set a default value.

I think the unit tests may be failing because of this.

plugins/UM3NetworkPrinting/tests/Cloud/TestCloudOutputDeviceManager.py
```diff
@@ -55,7 +56,9 @@ def _beforeTearDown(self):
         devices = self.device_manager.getOutputDevices()
         # TODO: Check active device
 
-        response_clusters = self.clusters_response.get("data", [])
+        response_clusters = []
```
ChrisTerBeke: This could be simplified to 

```python
response_clusters = [CloudClusterResponse(**cluster).toDict() for cluster in self.clusters_response.get("data", [])]
```

ianpaschal: That's not really simpler is it? Same number of variables and function calls... It's just putting it all on one line which makes the line harder to read at a glance.

ChrisTerBeke: Whatever you prefer. Your code, your party.

ianpaschal: I appreciate the feedback but too often I find myself reading a whole for each loop in a single line and going, "Wait... heh?... ah... ok."

ChrisTerBeke: Sure thing. List comprehensions give better performance in Python, but probably in this case wouldn't make a noticeable difference.

PR #5393, opened 2019-03-05
Add an object list/selector

diegopradogesto:
This PR wants to add a list of the objects that are loaded in the buildplate. Currently in Cura we have an experimental feature (multi build plate) that does this workaround, but that feature is not maintained anymore and the intention is to remove it soon. However some people find useful to have a list of the objects that are currently loaded.

I intended to do it "not intrusive" to the user if they don't want to see the list. That means that it's a bit hidden, but still something they can show or hide whenever they want. There is a preference setting that stores when the user decided to keep it open and it will remember that state in the next start up.

This is how it looks:

**Closed state**
screenshot 2019-03-05 at 10 39 56

**Open state**
screenshot 2019-03-05 at 10 40 40

**Open state in the dark theme**
screenshot 2019-03-05 at 10 41 56


diegopradogesto:
FYI @mennoberg @Yian-Lai

ChrisTerBeke:
Nice! It could use a tad more padding between the object selector collapse button and the actual panel IMO, but that's just details :)

diegopradogesto:
Done in 7305643
screenshot 2019-03-06 at 10 37 52

Ghostkeeper:
Developers, see issue CURA-6283.

fieldOfView:
Would have been nice if this were implemented more as a plugin. That would have required some work for plugins to insert themself into the viewport area, but that would have been welcome anyway.

PR #5396, opened 2019-03-06
Networked printer architecture

Ghostkeeper:
This changes the position where we store the list of networked printers, so that we can use it outside of the UM networking plug-in.

Contributes to issue CURA-6179.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
+++ b/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
@@ -213,6 +213,11 @@ class UM3OutputDevicePlugin(OutputDevicePlugin):
 
         self._checkManualDevice(address)
 
+    def _createMachineFromDiscoveredPrinter(self, key: str) -> None:
+        # TODO: This needs to be implemented. It's supposed to create a machine given a unique key as already discovered
+        # by this plugin.
+        pass
+
     def _checkManualDevice(self, address):
         # Check if a UM3 family device exists at this address.
         # If a printer responds, it will replace the preliminary printer created above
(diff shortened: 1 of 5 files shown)


plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -213,6 +213,11 @@ def addManualDevice(self, address):
 
         self._checkManualDevice(address)
 
+    def _createMachineFromDiscoveredPrinter(self, key: str) -> None:
+        # TODO: This needs to be implemented. It's supposed to create a machine given a unique key as already discovered
+        # by this plugin.
+        pass
```
Ghostkeeper: We need to see who's going to implement this then...?

ChrisTerBeke: This would also help in creating machines for printers that were found over the cloud API, and remove the need for having to be on the same local network to make that connection possible.

ianpaschal:
If some of this logic is leaving the UM3 Network Printing plugin, I'd like to see this in the Cura API as a way of properly insulating between plugins and core Cura code.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -318,7 +323,7 @@ def _onAddDevice(self, name, address, properties):
             device = ClusterUM3OutputDevice.ClusterUM3OutputDevice(name, address, properties)
         else:
             device = LegacyUM3OutputDevice.LegacyUM3OutputDevice(name, address, properties)
-
+        self._application.getMachineManager().addDiscoveredPrinter(device.getId(), name, self._createMachineFromDiscoveredPrinter, properties[b"printer_type"].decode("utf-8"))
```
ChrisTerBeke: The CloudOutputDeviceManager also needs implementation for the add and remove discovered printer functions

nallath:
We could let the API pass it through to the actual managers, but other than that, I don't see a good way to make it more abstract than it already is.

PR #5402, opened 2019-03-07
monitor does not show "Build Plate"

lucik:
bed_temperature_matches : error in temperature search expression

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -265,7 +265,7 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
                     if match[2]:
                         extruder.updateTargetHotendTemperature(float(match[2]))
 
-                bed_temperature_matches = re.findall(b"B: ?(\d+\.?\d*)  ?\/?(\d+\.?\d*) ?", line)
+                bed_temperature_matches = re.findall(b"B: ?(\d+\.?\d*) ?\/?(\d+\.?\d*)?", line)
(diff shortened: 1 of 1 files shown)


plugins/USBPrinting/USBPrinterOutputDevice.py
```diff
@@ -265,7 +265,7 @@ def _update(self):
                     if match[2]:
                         extruder.updateTargetHotendTemperature(float(match[2]))
 
-                bed_temperature_matches = re.findall(b"B: ?(\d+\.?\d*)  ?\/?(\d+\.?\d*) ?", line)
+                bed_temperature_matches = re.findall(b"B: ?(\d+\.?\d*) ?\/?(\d+\.?\d*)?", line)
```
Ghostkeeper: To make it work with both formats, maybe we can use something more like this:

    r"B: ?(\d+\.?\d*)\s+?\/?(\d+\.?\d*)?\s*"

I've changed the single space into `\s+` and still allow the space at the end.

lucik: Then you need to regexp in "extruder_temperature_matches" lead to this type

lucik: \s+? should be \s*. 

The last "\s*" I think is not needed.

bed_temperature_matches = re.findall(b"B: ?(\d+\.?\d*)\s*\/?(\d+\.?\d*)?", line)

extruder_temperature_matches = re.findall(b"T(\d*): ?(\d+\.?\d*)\s*\/?(\d+\.?\d*)?", line)

PR #5408, opened 2019-03-08
New bundled plugin: UFPReader

diegopradogesto:
This PR adds the possibility to open and visualise the Ultimaker Format Package files. It was an internal request. 

The UFP file is used in the Ultimaker S5 as a 3mf format files that stores the gcode and other files that are needed for the right operation in that printer. The file includes a screenshot and information about the material that is used (among other files)

With this plugin, the user will be able to open the GCode that is included in that file.

### DIFF
--- a/plugins/UFPWriter/UFPWriter.py
+++ b/plugins/UFPWriter/UFPWriter.py
@@ -23,7 +23,7 @@ catalog = i18nCatalog("cura")
 
 class UFPWriter(MeshWriter):
     def __init__(self):
-        super().__init__()
+        super().__init__(add_to_recent_files = False)
 
         MimeTypeDatabase.addMimeType(
             MimeType(
--- a/plugins/UFPWriter/UFPWriter.py
+++ b/plugins/UFPWriter/UFPWriter.py
@@ -23,7 +23,7 @@ catalog = i18nCatalog("cura")
 
 class UFPWriter(MeshWriter):
     def __init__(self):
-        super().__init__(add_to_recent_files = False)
+        super().__init__()
 
         MimeTypeDatabase.addMimeType(
             MimeType(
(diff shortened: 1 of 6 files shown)


plugins/UFPWriter/UFPWriter.py
```diff
@@ -23,7 +23,7 @@
 
 class UFPWriter(MeshWriter):
     def __init__(self):
-        super().__init__(add_to_recent_files = False)
+        super().__init__()
```
fieldOfView: GCodeWriter and GCodeGzWriter both have this, why should UFPWriter be different?

resources/bundled_packages/cura.json
```diff
@@ -492,6 +492,23 @@
...
         }
     },
+    "UFPReader": {
+        "package_info": {
+            "package_id": "UFPReader",
+            "package_type": "plugin",
+            "display_name": "UFP Reader",
+            "description": "Provides support for reading Ultimaker Format Packages.",
+            "package_version": "1.0.0",
+            "sdk_version": "6.0",
```
ChrisTerBeke: 6.0 is not a valid semver number, it should be in the x.x.x format.

diegopradogesto: This format is like this in the entire file. It's out of the scope of this PR to change the `sdk_version` for all the packages.

plugins/UFPWriter/UFPWriter.py
diegopradogesto: Oh, yes, the add to recent should be only in the reader. Reverted on 66a2425b863c8e2a0cd8dee78b30afba0aead164

resources/bundled_packages/cura.json
ChrisTerBeke: Still, boyscout rule should result in at least this one being correct....

diegopradogesto: 53b59220f8824156c49e4f95a70ac2345dea8652 😜

PR #5429, opened 2019-03-14
Add AMF reader plugin

### LINKED ISSUE #5414: 3.6.0 Is there any way to use .amf files?
Thanks for Cura!

**Application Version**
3.6.0

**Platform**
Mint 19.1/Cinnamon

**Printer**
Any

**Steps to Reproduce**
None needed

**Actual Results**
I cannot see the file

**Expected results**
I can see and use the file

**Additional Information**
Is there any way to use .amf files to slice? Slic3r-PE seems to be happy with them and it would save me some duplication as I often have to save files in both formats. It is handy to be able to use Slic3r-PE to repair some outside (Thingi etc) files and then export them back out to the same .amf format and then open them in Cura as I get much better prints with Cura than Slic3r.

Thanks

fieldOfView:
This PR adds support for reading AMF files. It is based on the implementation of legacy Cura, and is similarly incomplete.

https://github.com/daid/LegacyCura/blob/ad7641e059048c7dcb25da1f47c0a7e95e7f4f7c/Cura/util/meshLoaders/amf.py

Multiple meshes inside a single file are grouped together so they are not pushed apart.

Fixes #5414


fieldOfView:
One difference between legacy Cura and this implementation for modern Cura is that legacy Cura automatically assigned different extruders to different parts (interestingly even when the current printer is a single extrusion printer). My initial implementation does not do that (because I think it is quite arbitrary), which means that you don't really see what parts there are in a multimaterial AMF file unless you blindly control-click to pick a sub-part.

nallath:
CURA-6390

VjyedSC0UFMc8z:
I think Microsoft borrowed a ton from AMF to make the 3MF. Is that true? AMF is so much better for multipart or multi material files.

fieldOfView:
No, on both counts. AMF might be superior to STL (which doesn't even define a unit), but 3MF is superior to AMF.

fieldOfView:
Since there are now 4.1 branches, I am guessing Cura has gone into feature freeze and this PR will not make it in time for Cura 4.1. 

I originally submitted this as a PR so I would not have to maintain it as a plugin (see discussion here: https://github.com/Ultimaker/Cura/issues/5414#issuecomment-472747963). I could still add it as a plugin for Cura 3.5 and newer if it takes a review issue off your JIRA board.

Ghostkeeper:
We do also review plug-ins anyway, or should, at least.

nallath:
I had to add some minor things to get it completely there (0bd516da4c9f2a5e0b81a2ebc1b6419b0a932983, 0bd516da4c9f2a5e0b81a2ebc1b6419b0a932983 and 0bd516da4c9f2a5e0b81a2ebc1b6419b0a932983)

PR #5430, opened 2019-03-15
Add profile for JGAurora JGMaker Magic

pinchies:
This is another printer I would like to add, similar to the other JGAurora printers, but with a smaller bed area and a PVC bed. The thermistor calibration is also slightly off, so I lowered the nozzle temp accordingly as well.


Ghostkeeper:
Is the thermistor calibration off for your instance of this printer, or for all printers of this model, you think? :smile:

Ghostkeeper:
Developers, see issue CURA-6391.

pinchies:
I find that specific models tend to be consistent. This model uses a new style of hotend for this manufacturer (as well as a new motherboard) so I think that may be a contributing factor - perhaps the model of the thermistor model has changed.

https://github.com/Ultimaker/Cura/issues/6391 doesn't seem to work for me?

Ghostkeeper:
> https://github.com/Ultimaker/Cura/issues/6391 doesn't seem to work for me?

Ah, no that CURA-6391 is a reference for the Cura developers to the Ultimaker-internal ticket tracker. The link would be https://jira.ultimaker.com:8443/browse/CURA-6391 but you're unable to see that.

Ghostkeeper:
Oh, QA found an issue with this profile. There were two profiles that set the layer height to 0.15mm. Corrected that here: https://github.com/Ultimaker/Cura/commit/74bfd52c364122b5258a83fe141b34dfcc9204fb

pinchies:
@Ghostkeeper - thanks for that fix. This one is a mistake, but it would be great to get some documentation from Cura about profile creation - in particular, which settings should be included as machine specific, and which settings are quality profile specific, or if there is a way to set which category a particular setting falls into.

Ghostkeeper:
It's our intent to put the process in the contribution portal at some point. I expect we'll be making some tools for it then.

PR #5436, opened 2019-03-16
Fix Alfawise U30 to get more realistic print times

NicolasNSSM:
I can see a lot of complains about the print time being really different from what Cura says from the reality so here is a PR for the U30.  
I don't own an Alfawise **U20** printer but as they are almost the same, it may be duplicated to its definition too, (i won't do it)

### DIFF
--- a/resources/definitions/alfawise_u30.def.json
+++ b/resources/definitions/alfawise_u30.def.json
@@ -14,80 +14,52 @@
         }
     },
     "overrides": {
-        "machine_name": {
-            "default_value": "Alfawise U30"
-        },
+        "machine_name": { "default_value": "Alfawise U30" },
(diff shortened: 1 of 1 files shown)


nallath:
For our reference; CURA-6402

philippnj:
Hey Nicolas,
I cant get it running. created json file took your code, the printer is shown in custom printers-menu, but when clicking "add printer" cura (4 I am using) is crashing - willing to send a crash report. Any ideas?!  

BR

NicolasNSSM:
Hey @philippnj ,
Did you created and copy the `resources/extruders/alfawise_u30_extruder_0.def.json` file too?

philippnj:
That was it!!!  Thank you so much!

PR #5439, opened 2019-03-16
Add display messages to your gcode

eagle00789:
This will add LCD display messages to your gcode of the current printing layer and total number of layers.
For example: Printing layer 24 of 86
This PostProcessing script will be updated at a later date to also include the ability to use a percentage (or both) in the same gcode to see how far your printing really is


Ghostkeeper:
We already have this, or something very similar: https://github.com/Ultimaker/Cura/blob/master/plugins/PostProcessingPlugin/scripts/DisplayFilenameAndLayerOnLCD.py

PR #5456, opened 2019-03-19
Update creality_ender3.def.json

drzejkopf:
Changed override of "default_value" to "value" of acceleration_travel and jerk_travel to make slicer keep those values, as they were overrided by conditional in fdmprinter.def.json, and also removed layer height as it does nothing because it's overriden by preffered quality (draft) settings

### DIFF
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -51,16 +51,13 @@
             "default_value": 500
         },
         "acceleration_travel": {
-            "default_value": 500
+            "value": 500
         },
         "jerk_enabled": {
             "default_value": true
         },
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
For "value" it needs to be a string, right? The value is then a Python expression.

drzejkopf:
@Ghostkeeper I have checked it and it works well on my computer, but on second thoughts I think that it would be better to change it to:
```
"acceleration_travel": {
   "value": "acceleration_print"
}
```
```
"jerk_travel": {
   "value": "jerk_print"
}

```
but I have no idea how to edit file in pull request

Ghostkeeper:
Thanks. Looks good now.

PR #5460, opened 2019-03-20
'WIP-onboarding-by-ip' into 'parent' branch

rburema:
[CURA-6294]


cura/CuraApplication.py
```diff
@@ -455,7 +453,7 @@ def startSplashWindowPhase(self) -> None:
             # Misc.:
             "ConsoleLogger", #You want to be able to read the log if something goes wrong.
             "CuraEngineBackend", #Cura is useless without this one since you can't slice.
-            "UserAgreement", #Our lawyers want every user to see this at least once.
+            # NOTE: User-Agreement is part of the 'onboarding flow' now (See Welcome Pages).
```
Ghostkeeper: This message will be useless after a while because it doesn't matter that this used to be a plug-in at some point.

cura/UI/WelcomePagesModel.py
```diff
@@ -7,6 +7,7 @@
 
 from UM.Qt.ListModel import ListModel
 from UM.Resources import Resources
+from logging import Logger
```
Ghostkeeper: Seems like an unused import from the diff in this PR :laughing: 
```suggestion
```

resources/qml/Cura.qml
```diff
@@ -45,7 +45,7 @@ UM.MainWindow
     WelcomeDialog
     {
         id: welcomeDialog
-        visible: false
+        visible: true  // True, so if somehow no preferences are found/loaded, it's shown anyway.
```
Ghostkeeper: Instead of setting both `visible` and `visibility` (via `show()` and `hide()`) I think it'll be easier to only use `visibility`. It needs to be set to `Hidden` when `needToShowUserAgreement` is false, or `Windowed` when it is true. Then you won't need to have the Component.onCompleted link any more down at line 77. So something like this:

`visibility: CuraApplication.needToShowUserAgreement ? Window.Windowed : Window.Hidden`

resources/qml/Cura.qml
```diff
@@ -73,6 +73,18 @@ UM.MainWindow
         // This has been fixed for QtQuick Controls 2 since the Shortcut item has a context property.
         Cura.Actions.parent = backgroundItem
         CuraApplication.purgeWindows()
+
+        if (CuraApplication.needToShowUserAgreement)
+        {
+            welcomeDialog.visible = true;
+            welcomeDialog.currentStep = 0;
```
Ghostkeeper: Should be the default, right?

resources/qml/WelcomePages/AddPrinterByIpContent.qml
```diff
@@ -116,14 +74,12 @@ Item
...
-                    anchors.margins: 20
+                    anchors.margins: UM.Theme.getSize("default_margin").width
                     font: UM.Theme.getFont("default")
 
-                    text: ""
-
                     validator: RegExpValidator
                     {
-                        regExp: /[a-zA-Z0-9\.\-\_]*/
+                        regExp: /[0-9\.\-\_]*/
```
Ghostkeeper: Should it also accept hostnames here? And maybe IPv6 addresses? If it's only IP addresses, we can remove the dash and the underscore.

cura/CuraApplication.py
rburema: Removed.

cura/UI/WelcomePagesModel.py
rburema: Removed.

resources/qml/Cura.qml
rburema: After discussion: Decided to favour 'visible' instead of 'visibility/show/hide', becasue that didn't seem to work.

resources/qml/Cura.qml
rburema: Yes, but can't be careful enough with these legal things like the user-agreement, what if someone changes the default for whatever reason?

resources/qml/WelcomePages/AddPrinterByIpContent.qml
rburema: Removed dash, underscore, added `:` and _a through f_ for ipv6 addresses.

PR #5475, opened 2019-03-22
Lithophane defaults

BagelOrb:
The defaults for image import were rather arbitrary.

I think more people use this option for lithophanes than for height fields.

I've also updated the tooltip to explain *when* to use which option, rather than only being descriptive.


Ghostkeeper:
Developers, see issue CURA-6415.

PR #5478, opened 2019-03-22
CL-1267 Don't hide Cloud flow message for wrong printer

ianpaschal:
The goal here is to prevent a cloud cluster which is connected but not tied to the active machine setting metadata on the active machine.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
+++ b/plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
@@ -458,7 +458,7 @@ class UM3OutputDevicePlugin(OutputDevicePlugin):
         if self._start_cloud_flow_message and not self._start_cloud_flow_message.visible:
             self._start_cloud_flow_message.show()        
 
-    def _onCloudPrintingConfigured(self) -> None:
+    def _onCloudPrintingConfigured(self, device) -> None:
         # Hide the cloud flow start message if it was hanging around already
         # For example: if the user already had the browser openen and made the association themselves
         if self._start_cloud_flow_message and self._start_cloud_flow_message.visible:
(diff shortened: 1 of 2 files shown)


plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -473,7 +473,16 @@ def _onCloudPrintingConfigured(self) -> None:
         # Set the machine's cloud flow as complete so we don't ask the user again and again for cloud connected printers
         active_machine = self._application.getMachineManager().activeMachine
         if active_machine:
-            active_machine.setMetaDataEntry("do_not_show_cloud_message", True)
+
+            # The active machine _might_ not be the machine that was in the added cloud cluster and
+            # then this will hide the cloud message for the wrong machine. So we only set it if the
+            # host names match between the active machine and the newly added cluster
+            saved_host_name = active_machine.getMetaDataEntry("um_network_key", "").split('.')[0]
+            added_host_name = device.toDict()["host_name"]
```
sedwards2009: Does this also have to match printers by IP address too?

ianpaschal: Shit, I guess it should... since manually added printers use the IP as the host name.

In this case I'm not really sure what to do. In theory every printer should just have a UUID tied to it, but that means a lot of infrastructural changes...

ianpaschal:
@ChrisTerBeke @nallath Do you guys have some thoughts about how we should handle this association? I'm thinking that it doesn't really matter what the host name string is (actual host name, IP, etc.) so long as it's a unique string. But following our discussion last Wednesday maybe there's a better way to make this association...? One which is better equipped to handle changes we want to make in the future?

ChrisTerBeke:
Not sure as I don't know all the incoming changes. Maybe it makes sense to revisit this after the new device adding structure has been merged?

Ghostkeeper:
It needs to associate with the actual device, filtering out duplicates if the actual device gets discovered in two different ways, right? So then you need something unique about the actual device: The MAC address can be considered unique in this context.

ianpaschal:
My opinion is that this needs a massive overhaul. Generally the `um_network_key` is used as the unique identifier everywhere and in trying to use the host name, since that's what comes from the cloud, the code is getting messier and messier the more I work on it.

Whether it's the MAC address or a UUID, we need a single unique identifier because using the `um_network_key`, the host name,  the IP address _as_ a host name, and finally (maybe) the MAC address simultaneously to achieve the same goals (knowing what is what) is... yeah... a mess.

Since the `um_network_key` is used pretty universally in the plugin right now and is easily compared to the host name which is available when a cloud cluster is available, I'm going to leave it at this for now to fix the bug. It will probably need to be revisited either way.

PR #5481, opened 2019-03-22
PrintSetupSelector: "G code" -> "G-code"

thopiekar:
Noticed that "g-code" is written at some places with and without a hyphen.
Since it think it should be written with, I'm correcting it at this point.


thopiekar:
Moreover, it is already written with a hyphen in a message, which pops up at the same time:
image

PR #5497, opened 2019-03-23
add Mark2 definition files

TinkerGnome:
This are the definitions / resource files for the Mark2 / Magnetic Toolchanger add-on


Ghostkeeper:
We're having our reservations about the post-processing script in this PR. It seems to try to fix bugs in CuraEngine without fixing those bugs in CuraEngine. So here's some questions for you:

* Did you find a travel move between G10 or M104 and T that was apparently implemented as a hack (as in `remove_hack()`)?
* Did you find a case where there are unnecessary travel moves after a tool change (as in `remove_superfluous()`)?

I'm suspecting that the `remove_superfluous()` function can be prevented by setting the Nozzle Switch location to be 0,0 and relative, or else by disabling the combing setting. I also think that these things should be fixed in CuraEngine if they are indeed bugs or hacks.

TinkerGnome:
Well, i didn't wrote the script and it was some time ago..., but I'll try:

Regarding `remove_hack`:
There is a good chance that `remove_hack` is not needed at all. I'll check it.

Regarding `remove_superfluous`:
Usually CuraEngine generates a move to the previous print position after every tool change. This is not a bug (AFAIK), but probably necessary to take extruder offsets into account. It's just not suitable for the magnetic tool changer where we want to move straight to the starting point of the new selected nozzle (like the "Griffin"-flavor does AFAIK). That's what `remove_superfluous` is intended for.
Is this behavior related to combing? And if:... combing can not be disabled for tool changes only, it's the same setting for the whole print, isn't it?

Can you explain what is meant by "Nozzle Switch location"? I found `machine_extruder_end_pos_*` / `machine_extruder_start_pos_*`, both seem to be relative and 0/0 by default.
Or is there more?

Ghostkeeper:
> Usually CuraEngine generates a move to the previous print position after every tool change. This is not a bug (AFAIK), but probably necessary to take extruder offsets into account.

IIRC, there is a move that you could see in layer view that is not in the g-code but purely caused by the tool change command suddenly changing the position of the print head without changing the position of the nozzle. Maybe that's the one you're thinking of?

> Can you explain what is meant by "Nozzle Switch location"? I found `machine_extruder_end_pos_*` / `machine_extruder_start_pos_*`, both seem to be relative and 0/0 by default.
Or is there more?

No, those are the ones that I meant.

TinkerGnome:
@Ghostkeeper - It turned out that Cura 4 does not generate the superfluous moves anymore,
(with only one exception which is addressed here: [issue #5545 ](https://github.com/Ultimaker/Cura/issues/5545))
I removed the script from the pull request, it is not longer necessary.

resources/extruders/Mark2_extruder2.def.json
```diff
@@ -0,0 +1,19 @@
...
+        "machine": "Mark2_for_Ultimaker2",
+        "position": "1"
+    },
+
+    "overrides": {
+        "extruder_nr": {
+            "default_value": 1,
+            "maximum_value": "1"
+        },
+        "machine_nozzle_offset_x": { "default_value": 0.0 },
```
LipuFei: Please check if your Extruder 2 and Extruder 1 are at the same location. They have the same offsets (0, 0) right now.

resources/definitions/Mark2_for_Ultimaker2.def.json
```diff
@@ -0,0 +1,241 @@
...
+        "jerk_layer_0": { "value": "jerk_topbottom" },
+        "jerk_prime_tower": { "value": "10 if jerk_print < 16 else math.ceil(jerk_print * 15 / 25)" },
+        "jerk_support": { "value": "10 if jerk_print < 16 else math.ceil(jerk_print * 15 / 25)" },
+        "jerk_support_interface": { "value": "jerk_topbottom" },
+        "jerk_topbottom": { "value": "10 if jerk_print < 25 else math.ceil(jerk_print * 10 / 25)" },
+        "jerk_wall": { "value": "10 if jerk_print < 16 else math.ceil(jerk_print * 15 / 25)" },
+        "jerk_wall_0": { "value": "10 if jerk_wall < 16 else math.ceil(jerk_wall * 6 / 10)" },
+        "jerk_travel_layer_0": { "value": "math.ceil(jerk_layer_0 * jerk_travel / jerk_print)" },
+        "extruder_prime_pos_abs": { "default_value": false },
+        "machine_extruder_start_pos_abs": { "default_value": false },
```
LipuFei: Those fields should not be in the machine definition files. CI is failing due to this problem. Please remove the following fields from this file:
```
machine_extruder_start_pos_abs
machine_extruder_start_pos_x
machine_extruder_start_pos_y
machine_extruder_end_pos_abs
machine_extruder_end_pos_x
machine_extruder_end_pos_y
```

resources/extruders/Mark2_extruder2.def.json
TinkerGnome: It's intentionally defined this way (it's a tool changer after all). Any offsets are (exclusively) processed by the firmware and must not depend on slicer settings (or on a specific slicer).

resources/definitions/Mark2_for_Ultimaker2.def.json
TinkerGnome: Done. It seems like checks were successful this time 😃

Ghostkeeper:
Thanks for your investigation, TinkerGnome :)

TinkerGnome:
In this case I'm only the messenger, but nevertheless - graag gedaan 😃

PR #5499, opened 2019-03-24
Fix for to display the text in the dark theme (Menu - Preferences - Materials).

bliser3d:
This fix and improvement а for to display the text in "Menu - Preferences - Materials - (element ScrollView)" in the dark theme.

Before corrections:
image

After corrections:
image

Note: 
1. To run the application using the command:
 "QT_QUICK_CONTROLS_STYLE=Desktop /home/User/Cura/AppRun"
2. I did not change line 141 in the file "MaterialsTypeSection.qml". The reason for its change is not clear to me.

### DIFF
--- a/resources/qml/Preferences/Materials/MaterialsTypeSection.qml
+++ b/resources/qml/Preferences/Materials/MaterialsTypeSection.qml
@@ -29,7 +29,7 @@ Item
             }
             else
             {
-                return "transparent"
+                return UM.Theme.getColor("favorites_subheader_hover")
             }
         }
         width: parent.width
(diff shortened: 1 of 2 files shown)


fieldOfView:
The colors in this dialog should not be using the (Cura) theme in the first place. It is perfectly possible to have a dark theme in Cura, but a bright system-theme, eg in Windows or OS X where a dark system theme either does not really exist or at least it does not carry over into the system-themed dialogs in Cura.

bliser3d:
You're right. If my correction is contrary to the general style, it should not be used.
How to do it right, I do not know, but in my case, the fix helps me.

PR #5511, opened 2019-03-26
Fix for weird home directories on Windows

### LINKED ISSUE #5471: cx_freeze app data not found?
3.6/4.0
Win10 64bit
Ender3
Install Cura
Does not startup with and w/o administrative rights
Should start
Screenshot
https://imgur.com/a/QJ09UYA

Ghostkeeper:
The expanduser function will expand to a combination of HOMEDRIVE and HOMEPATH which might not be the same as APPDATA if your configuration is very weird (e.g. you've changed your APPDATA location).

Contributes to CURA-6418.
Hopefully fixes #5471.

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -23,7 +23,10 @@ known_args = vars(parser.parse_known_args()[0])
 if not known_args["debug"]:
     def get_cura_dir_path():
         if Platform.isWindows():
-            return os.path.expanduser("~/AppData/Roaming/" + CuraAppName)
+            appdata_path = os.getenv("APPDATA")
+            if not appdata_path: #Defensive against the environment variable missing (should never happen).
+                appdata_path = "."
+            return os.path.join(appdata_path, CuraAppName)
         elif Platform.isLinux():
             return os.path.expanduser("~/.local/share/" + CuraAppName)
         elif Platform.isOSX():


thopiekar:
Personally, I dislike it. Using only "." Would lead to creations of "cura" directories, if you run the command from other locations.
Additionally, your implementation doesn't follow the unix style of creating config directories. Applications, like Atom, Origin, Eclipse and nuget, use "." as a prefix.
So I would suggest the following solution:
Checking whether HOMEDRIVE and HOMEPATH are present as environment variables and then trying to get the home path via `os.path.join(os.getenv("HOMEDRIVE"), os.getenv("HOMEPATH"))` first.
Then looking for "AppData" and use it if present.
If not, then ".{}".format(CuraAppName) and use that one in our home path.
Lastly, if HOMEDRIVE and HOMEPATH are not present, then use "." as you did.
But on the other hand I'm still devided about using ".", because if an installation misses even these then it can't be our fault and doesn't deserve any further efforts. 🤔

Ghostkeeper:
> Personally, I dislike it. Using only "." Would lead to creations of "cura" directories, if you run the command from other locations.

This is taken from the Resources.py file which would already also do that. It's meant to be as a fallback for if your system is broken. Since Cura is run from the home directory if you launch it via the start menu, it'll create a folder there which is better than nothing. I didn't dare to actually assume this though, so I didn't use `.\AppData\Roaming` because if it's not run from the home directory this would create an additional AppData folder elsewhere which would be confusing.

> So I would suggest the following solution:
Checking whether HOMEDRIVE and HOMEPATH are present as environment variables and then trying to get the home path via `os.path.join(os.getenv("HOMEDRIVE"), os.getenv("HOMEPATH"))` first.

We don't want to store anything in the HOMEDRIVE or HOMEPATH. Never. We want to store things in the APPDATA, regardless of whether that is inside the home directory or not.

fieldOfView:
To elaborate: if %APPDATA% is not set, the Windows installation is severely broken. 

%APPDATA% may not be inside the user's profile folder; eg on many "corporate" networks the %APPDATA% folder is on a network drive.

thopiekar:
Ok, didn't know about this behavior that `.` will be expanded to APPDATA. By the way, If it is really the case that an installation must be severally broken if APPDATA does not exist, then we should exactly notify about this if APPDATA is not present. Guess that would need to be done in UM.Resources. Guess that this one you are referencing to, @Ghostkeeper , right?

Ghostkeeper:
> then we should exactly notify about this if APPDATA is not present.

I think it's out of scope for Cura to notify the user of a broken Windows installation.

Point is, we were trying to find the AppData folder in the home directory. But it's not always there, so we should just ask the env variables for the AppData folder like we do in UM.Resources.

PR #5520, opened 2019-03-27
Add HMS434 machine

maukcc:


resources/definitions/hms434.def.json
```diff
@@ -0,0 +1,165 @@
...
+            "0": "hms434_tool_1",
+            "1": "hms434_tool_2",
+            "2": "hms434_tool_3",
+            "3": "hms434_tool_4",
+            "4": "hms434_tool_5",
+            "5": "hms434_tool_6",
+            "6": "hms434_tool_7",
+            "7": "hms434_tool_8"
+        },
+        "platform":            "hms_platform.stl",
```
Ghostkeeper: This platform file seems to be missing.

maukcc: Thanks for noticing !

Ghostkeeper:
Looks a bit unconventional with that texture, but it works fine.

PR #5556, opened 2019-04-02
CURA-6440 Exclude support blockers in used extruders

LipuFei:


Ghostkeeper:
This is also an issue in CuraEngine itself.

Ghostkeeper:
Developers, see issue CURA-6440.

cura/Settings/ExtruderManager.py
```diff
@@ -224,7 +224,16 @@ def getUsedExtruderStacks(self) -> List["ContainerStack"]:
 
         # Get the extruders of all printable meshes in the scene
         meshes = [node for node in DepthFirstIterator(scene_root) if isinstance(node, SceneNode) and node.isSelectable()] #type: ignore #Ignore type error because iter() should get called automatically by Python syntax.
+
+        # Exclude anti-overhang meshes
+        mesh_list = []
         for mesh in meshes:
+            stack = mesh.callDecoration("getStack")
+            if stack is not None and stack.getProperty("anti_overhang_mesh", "value"):
```
Ghostkeeper: I'm adding an additional check here to do the same for `support_mesh`. The support mesh also doesn't get printed with its own extruder but with the extruder that support gets printed with.

PR #5561, opened 2019-04-03
CURA-6188 Use GitLab CI

LipuFei:


plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -66,10 +66,11 @@ def __init__(self, device_id, address, properties, parent = None) -> None:
         self._received_print_jobs = False # type: bool
 
         if PluginRegistry.getInstance() is not None:
-            self._monitor_view_qml_path = os.path.join(
-                PluginRegistry.getInstance().getPluginPath("UM3NetworkPrinting"),
-                "resources", "qml", "MonitorStage.qml"
-            )
+            plugin_path = PluginRegistry.getInstance().getPluginPath("UM3NetworkPrinting")
```
sedwards2009: Where is this stuff coming from?

LipuFei: `PluginRegistry.getPluginPath()` returns `Optional[str]`, and I got a complain from mypy about this because it's used in `os.path.join()`.

Issue #3946, opened 2018-06-18
[3.3.1] Flow Setting in Material Profile


fatalelieberi:
**Application Version**
3.3.1

**Platform**
Windows 10

**Printer**
Tevo Tornado

**Expected results**
Flow setting should be put in Material Profile. The main reason is that flow/extrusion rate is more dependent on the material being used. Putting it in regular print profile would cause user to create several profiles or modify current profile to match the flow setting.

ChrisTerBeke:
All settings are in the place that we think they should be and work best for the profiles we create. We're probably not going  to change this anytime soon.

Ghostkeeper:
You can now define the Flow setting in a material profile but only when modifying the XML files by hand, not via the interface.

fatalelieberi:
Can you tell me where I can find the XML file? I've been digging through the configuration folder but cannot find it.

Ghostkeeper:
The built-in XML files will be in your installation. For instance: `C:/Program Files/Cura 3.5/resources/materials`. The custom XML files will be in your configuration folder (go in Cura to Help -> Show Configuration Folder...) under the folder `materials`.

knoopx:
I've managed to patch my current installation, here are the instructions https://gist.github.com/knoopx/0f600cfdd14863f45f4e90842688e888. However it would be nice to have this on by default.

filipgoc:
BTW If anyone finds this via search and cannot make it work: 

To add `material_flow` or other _extended let's say_ settings, you need to follow the settings are described in this PR
https://github.com/Ultimaker/Cura/pull/2333#ref-issue-333852978 it it won't work for you!

In short, there's one more namespace declaration
```
<fdmmaterial xmlns="http://www.ultimaker.com/material" xmlns:cura="http://www.ultimaker.com/cura" version="1.3">
```
And corresponding 
```<cura:setting key="material_flow">66</cura:setting>```

filipgoc:
> However it would be nice to have this on by default.

Just want to say I 100% agree with you.
I know the Ultimaker team/ Chris has said that this is undesirable... I guess I cannot but completely disagree (alternatively I'm more than happy to hear their reasoning.)

In fact, I would argue that changing the flow for the material is the _one thing_ that a user may want to do at home as a part of a basic calibration procedure. Every manufacturer makes slightly different effective diameter... and this is the most natural place to calibrate it (even though it can be done in many other places.)

Ghostkeeper:
[...]

PR #5570, opened 2019-04-04
Remove the link to the compatibility chart

diegopradogesto:
Actually it only made sense for Ultimaker materials. Now that we can include instruction link in every materials, a specific link for every material can be used instead. This is done in https://github.com/Ultimaker/fdm_materials/pull/78

Contributes to CURA-6423.

### DIFF
--- a/resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
+++ b/resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
@@ -175,59 +175,6 @@ Cura.ExpandablePopup
             }
         }
 
-        Item
-        {
-            height: visible ? childrenRect.height: 0
-            anchors.right: parent.right
-            anchors.rightMargin: UM.Theme.getSize("default_margin").width
-            width: childrenRect.width + UM.Theme.getSize("default_margin").width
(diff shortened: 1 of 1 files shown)


filipgoc:
Awesome. Previously, I actually had to use a patch to hide the hardlink as too many people were confused.

nallath:
Do you have any other hacky patches that you need to do? We're open to changing them if that makes it easier for everyone.

fieldOfView:
Sidebar GUI plugin is a rather hacky patch ;-)

nallath:
Well played.

filipgoc:
Well, since you asked :-) @nallath, the state of material settings and the related UI is behind. Inspired by your prompt here, I made a new write up on the old post about including Flow in the material settings. https://github.com/Ultimaker/Cura/issues/3946

I am currently using a qml patch to add flow to the material settings ui.  (NOT in the sidebar). 
And it works great.  Adding the Flow would be the MVP there. A discussion about what should be a part of the Material ui in general (to evolve from the olden days where only a very limited subset of settings was allowed) would be the gravy - but that is not fast and easy (unlike flow).

filipgoc:
I just learnt @fieldOfView may be working on a plugin for this.  Nice. https://github.com/Ultimaker/Cura/issues/3946#issuecomment-482290831

Issue #4758, opened 2018-11-08
Disabled extruder selected


BagelOrb:
Here's a small UI issue for you:

When you have the left extruder selected and are looking at the advanced settings and then load a project file which has the left extruder disabled, the UI doesnt automatically switch to the right extruder.

You are then able to edit settings of an extruder which isn't even enabled.

Not a big issue, but perhaps this is fixable in 5 minutes.

diegopradogesto:
Good catch!

Vandrasc:
devs CURA-5930

PR #5577, opened 2019-04-05
CURA-5930 Switch to the first enabled extruder after setActiveMachine

LipuFei:


nallath:
This should fix https://github.com/Ultimaker/Cura/issues/4758

PR #5595, opened 2019-04-08
Add skin edge support settings

smartavionics:
This PR adds the Skin Edge Support Thickness and Skin Edge Support Layers settings that control the generation of infill walls immediately below the edge of skin regions above.

Here's an example, you can see that some extra infill walls have been created...

Screenshot_2019-04-08_11-13-13

And here's the next layer with the skin that stops above the new infill wall...

Screenshot_2019-04-08_11-13-26

The infill walls are added to more layers depending on how sparse the infill is, this ensures that with really sparse infill, the walls do a good job. It is now possible to have skin regions finishing above very sparse infill with no real problems with the skin lines not getting terminated well.

I have been using this feature for a while and get very good results.


smartavionics:
Engine component is https://github.com/Ultimaker/CuraEngine/pull/1030.

rburema:
Devs: We've made an internal ticket to review/test/merge this PR as: CURA-6453

smartavionics:
BTW, here's a photo that shows how well this feature works. That's a single skin over sparse infill, note how there's no gaps in the skins at the base of the walls as you would normally expect when the infill density is low.

IMG_20190311_083229812

BagelOrb:
How does this feature deal with slanted top surfaces?
Does it generate skin edge supports on each layer?
What is the skin edge support area is too thin to be printed; will it still remove part of the infill below?
In that case there would be some angles of top surfaces which are missing infill right below and which would then be printed a lot worse.

smartavionics:
Hello Tim.

>How does this feature deal with slanted top surfaces?
Does it generate skin edge supports on each layer?

Yes, if it is configured for more than 1 layer of skin edge support it will generate infill walls beneath the sloping top surface. Here's an example which is configured to have 2 layers of skin edge support.

Screenshot_2019-06-08_11-06-24

Screenshot_2019-06-08_11-06-42

> What is the skin edge support area is too thin to be printed; will it still remove part of the infill below?
In that case there would be some angles of top surfaces which are missing infill right below and which would then be printed a lot worse.

No, it shouldn't do that. If there's no skin to be printed, the infill won't be changed.

BagelOrb:
Ok sounds good!

I wonder what happens at a 45* angled top/side surface tho.
Do the skin edge lines then not get printed?

smartavionics:
> I wonder what happens at a 45* angled top/side surface tho.
Do the skin edge lines then not get printed?

A 45 deg slope will not normally require skin underneath so those skin edge lines don't appear.

BagelOrb:
Seems good to me.

@Ghostkeeper I don't really understand how an issue can be 'in progress' for 2 months. Is the label outdated?

Ghostkeeper:
> @Ghostkeeper I don't really understand how an issue can be 'in progress' for 2 months. Is the label outdated?

It's literally been in our sprints for 2 months.

rburema:
This was merged via command line, but github doesn't seem to pick up on it (possibly because of the last merge with master here).

I'll wait a bit to see if github comes to its senses, otherwise I'll close this tomorrow.

PR #5596, opened 2019-04-08
CURA-6057 Fix onboarding bugs

LipuFei:


plugins/MachineSettingsAction/MachineSettingsExtruderTab.qml
```diff
@@ -20,12 +20,12 @@ Item
     anchors.right: parent.right
     anchors.top: parent.top
 
-    property int labelWidth: 180
-    property int controlWidth: UM.Theme.getSize("setting_control").width * 3 / 4
+    property int labelWidth: 210 * screenScaleFactor
```
nallath: This was incorrect in the first version, because it hardcodes the label width. It's better now, but i guess we shouldn't even have those in the first place, right?

LipuFei: I agree, but it's hard to theme those I suppose because they are only used in the Machine Settings dialog. Those are the numbers I got after a lot of tweaking...

PR #5606, opened 2019-04-10
CL-1325 Pass output device (or not) to cloud check method

ianpaschal:
Contributes to CL-1325


plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -488,7 +488,7 @@ def _onServiceChanged(self, zero_conf, service_type, name, state_change):
         return True
 
     ## Check if the prerequsites are in place to start the cloud flow
-    def checkCloudFlowIsPossible(self) -> None:
+    def checkCloudFlowIsPossible(self, cluster: Optional["CloudOutputDevice"]) -> None:
```
sedwards2009: You'll need to import `CloudOutputDevice` at the top of the file.

ianpaschal: I don't think so if it's in quotes.

PR #5611, opened 2019-04-11
CURA-6449 Disable printers that are not host of a group

LipuFei:
- Group printers into hosts and non-hosts and of unknown type.
 - Show available/connectable printers first in the list.
 - Show sections for connectable and non-connectable printers.

### DIFF
--- a/cura/Machines/Models/DiscoveredPrintersModel.py
+++ b/cura/Machines/Models/DiscoveredPrintersModel.py
@@ -62,12 +62,14 @@ class DiscoveredPrinter(QObject):
         from cura.CuraApplication import CuraApplication
         readable_type = CuraApplication.getInstance().getMachineManager().getMachineTypeNameFromId(self._machine_type)
         if not readable_type:
-            readable_type = "unknown"
+            readable_type = catalog.i18nc("@label", "Unknown")
         return readable_type
 
     @pyqtProperty(bool, notify = machineTypeChanged)
     def isUnknownMachineType(self) -> bool:
-        return self.readableMachineType == "unknown"
+        from cura.CuraApplication import CuraApplication
+        readable_type = CuraApplication.getInstance().getMachineManager().getMachineTypeNameFromId(self._machine_type)
+        return not readable_type
 
     @pyqtProperty(QObject, constant = True)
     def device(self) -> "NetworkedPrinterOutputDevice":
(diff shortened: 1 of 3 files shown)


cura/Machines/Models/DiscoveredPrintersModel.py
```diff
@@ -66,12 +66,23 @@ def readableMachineType(self) -> str:
 
     @pyqtProperty(bool, notify = machineTypeChanged)
     def isUnknownMachineType(self) -> bool:
-        return self.readableMachineType.lower() == "unknown"
+        return self.readableMachineType == catalog.i18nc("@label", "Unknown")
```
nallath: We really shouldn't do this. Comparing with translations is usually a big no-no, because it's rather prone to breaking.

nallath: Also; We could just ignore this situation, as it only happens in the UM office with UM3 prototypes.

LipuFei: I removed the translation of "Unknown" here in 26205fbdc. I still think we should keep the "isUnknown" flag though, at least it will be more clear when we are testing in the UM office. Your call.

Ghostkeeper: I'm fine with keeping it in. It should be more defensive if a printer firmware has a bug, or if we're dealing with different discovery plug-ins such as OctoPrint or USB.

cura/Machines/Models/DiscoveredPrintersModel.py
```diff
@@ -62,17 +62,28 @@ def readableMachineType(self) -> str:
         from cura.CuraApplication import CuraApplication
         readable_type = CuraApplication.getInstance().getMachineManager().getMachineTypeNameFromId(self._machine_type)
         if not readable_type:
-            readable_type = catalog.i18nc("@label", "Unknown")
+            readable_type = "unknown"
```
Ghostkeeper: Ah, but this should be human readable, and contrary to popular belief the French are also humans. I think what you intended to do is to have this properly translated, but not check against the translated texts later on. So I'd like you to:
* Revert this change.
* in `isUnknownMachineType`, check against `getMachineTypeNameFromId` to see whether it returns `None` instead of checking against the human-readable text.

LipuFei: Done in ce5682aa7

PR #5620, opened 2019-04-12
Add maximum deviation setting

Ghostkeeper:
This setting allows tweaking the deviation from the norm separately from the length of line segments.

Implements CURA-6458.


Ghostkeeper:
See also https://github.com/Ultimaker/CuraEngine/pull/1037 for the implementation of the setting.

resources/definitions/fdmprinter.def.json
```diff
@@ -5855,7 +5855,7 @@
                     "description": "The minimum size of a line segment after slicing. If you increase this, the mesh will have a lower resolution. This may allow the printer to keep up with the speed it has to process g-code and will increase slice speed by removing details of the mesh that it can't process anyway.",
                     "type": "float",
                     "unit": "mm",
-                    "default_value": 0.01,
+                    "default_value": 0.25,
                     "minimum_value": "0.001",
```
diegopradogesto: We will need to modify also the `minimum_value`, `minimum_value_warning` and `maximum_value_warning` accordingly to the new value.

resources/definitions/fdmprinter.def.json
```diff
@@ -5867,7 +5867,7 @@
                     "description": "The minimum size of a travel line segment after slicing. If you increase this, the travel moves will have less smooth corners. This may allow the printer to keep up with the speed it has to process g-code, but it may cause model avoidance to become less accurate.",
                     "type": "float",
                     "unit": "mm",
-                    "default_value": 0.02,
+                    "default_value": 0.5,
                     "value": "meshfix_maximum_resolution * speed_travel / speed_print",
                     "minimum_value": "0.001",
```
diegopradogesto: Same as before.

resources/definitions/fdmprinter.def.json
Ghostkeeper: Yeah, I saw this also when Tom was using it.

PR #5634, opened 2019-04-17
Add minimal support for discovering cloud printers outside of LAN

ChrisTerBeke:
This is a very MVP version of cloud cluster discovery. It shows the cloud printers in the 'add printer' discovery list when signed in and creates a stack with the correct printer type when clicking 'add'. There is no further UX changes yet (like telling the user in the UI that cloud printers will appear in the list when signed in). The code and UX are not optimal, but since we're going to re-work the plugin anyways we might as well have a temp solution?

ping @ianpaschal @sedwards2009 @nallath


Ghostkeeper:
Was there a ticket attached to this?

Also the UM3NetworkPrinting tests fail on this branch.

ChrisTerBeke:
Still tweaking some things. There is no ticket yet but I've alerted Jaime and the Connect team about this PR so it can be picked up if wanted.

ChrisTerBeke:
The tests work now :)

ianpaschal:
Since there's agreement from our team/PO that Simon and I can still do a refactor in the next 2 sprints, this is OK by me. I read through it, seems logical enough. 👌

PR #5636, opened 2019-04-18
New values for the max resolution and max deviation settings

diegopradogesto:
According to the experts


othermod:
Hey guys, 
@Ghostkeeper 
@diegopradogesto 

I've been running a couple of Ultimaker 3 Extended printers for the last year, and recently started running a lab with a couple dozen S5s. I'm getting much lower print quality on S5s when compared to the 3s, and after lots of testing and research I've discovered that its due to this commit that was made to modify resolution and deviation settings.

The issue I'm experiencing is that sharp curves have very low detail and appear faceted. I can avoid the issue by simply changing the resolution from the calculated value of (speed_wall_0 + speed_wall_x) / 60, back to the original 0.04 (or 0.05 that the UM3 currently has), or by greatly dropping print speed.

My question though, is why was this changed to a calculated value?  Was this done because the change in stepper drivers caused the mainboard to become bottlenecked with stepper pulses, and the resolution was lowered to avoid stuttering?

I attached two images. One showing the stock settings in Cura for the UM5, and one showing 0.05 resolution.
Modified
Standard



Thanks

Ghostkeeper:
> My question though, is why was this changed to a calculated value? Was this done because a hardware change between the UM3 and UMS5 caused the mainboard to become bottlenecked, and the resolution was lowered to avoid stuttering?

Yes, the bottleneck in the processing power lies in the Python code on the printer's Linux board that sends the g-code off to the microcontroller that runs Marlin. Since the UMS5 has to draw a touch screen, it has (in the worst case) a harder time to keep up to speed with the rate at which g-code commands are consumed.

The formula is intended to keep the maximum number of segments per second constant regardless of the speed at which you're printing. Printing at greater speed will then still have the same number of segments per second. AFAIK the latest firmware of the S5 will handle about 60seg/s in the worst case, but the tuning of the maximum resolution here also takes quality into account as slowing down is not always bad if it's very minute.

The values for UM3E and UMS5 have been carefully tuned by the process engineers of Ultimaker to work well for most cases, but there will always be exceptions. I'm not at liberty to change them (just to commit them since those process engineers have less experience with Git).

othermod:
Thank you for the response. 

The engineers seem to have erred on the side of caution when choosing the calculated value for the resolution. They went a little too far in my opinion, and the quality loss on small cylinders is excessive. A better solution would be to scale print speed of an area when the number of segments is high (as far as I know, this feature doesn't exist). Maybe another option would be to change how often the screen is drawn or how much of the screen is drawn during a print to reduce CPU usage.

The 60 segment/s info at least gives me a limit to watch for when I'm adjusting the resolution manually. I haven't seen any stuttering yet, but I definitely haven't pushed resolution very far. Even 2x the current number makes a pretty big difference visually.

I'll try to contact the engineers at Ultimaker to continue this discussion. I really appreciate the info.

PR #5657, opened 2019-04-23
CL-1055 Show "sliced for" pills on print jobs

ianpaschal:
Formerly the pills were rendered based on the print job's `compatible_machine_families` property.

This property is always the intersection of printers in the group and printers the job can use.* In the case of an S5 print job in a group with no S5 machines, the print job has no compatible machine families to show a pill for, and hence there's no pill and this bug exists. Although logical, this is not ideal because knowing what the print job was sliced for is exactly the case where it's very important to see the pill next to the blocked print job.

For this reason I have switched to using the `configuration.machine_variant` instead of `compatible_machine_families`.

*This is sort of a lie. In the Cura Connect code, the print job's side of the matching never uses more than a single printer type: whatever is set as the `machine_variant` property.

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
@@ -158,14 +158,9 @@ Item
                     spacing: 6 // TODO: Theme!
                     visible: printJob
 
-                    Repeater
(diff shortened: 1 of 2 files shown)


ChrisTerBeke:
Shouldn't this go into the 4.1 banch?

ianpaschal:
Yes, although when I made the PR there was no 4.1 branch...

PR #5663, opened 2019-04-25
CURA-6483 Move add by ip device discovery into DiscoveredPrintersModel

LipuFei:
Together with https://github.com/Ultimaker/Uranium/pull/484


Ghostkeeper:
Please fix the tests. When running this locally (using correct branches) I get failing tests as well though.

PR #5665, opened 2019-04-25
Add Stereotech STE320 and START definitions

frylock34:
Hello,
I'd like to add machine definitions for Stereotech 3D printers:
STE320 and START


Ghostkeeper:
I can't find any information about that START printer and everything about that STE320 printer is in Russian :smile:

Ghostkeeper:
Could you reduce the resolution of the `stereotech_ste320_platform.obj` file a bit? Our guideline is to not go above 1MB per platform file. It really adds up to the download and space requirements of Cura otherwise, the model files being about half of the current download of Cura.

frylock34:
> I can't find any information about that START printer and everything about that STE320 printer is in Russian 😄 But seeing as how this comes from the Stereotech repository I guess it's reliable.

This printers are not ready yet, we are just preparing) The more info about us could be found at [stereotech.org](https://stereotech.org/)

PR #5673, opened 2019-04-29
CURA-6483 Fix Add by IP Issues

LipuFei:


cura/Utils/networking.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+# Checks if the given string is a valid IPv6 address.
+def isIPv6(address: str) -> bool:
+    return _REGEX_IPV6.fullmatch(address) is not None
+
+
+# Checks if the given string is a valid IPv4 or IPv6 address.
+def isValidIP(address: str) -> bool:
+    return isIPv4(address) or isIPv6(address)
```
Ghostkeeper: Try this:
```
try:
    socket.inet_aton(address)
    return True
except socket.error:
    return False

File: resources/qml/WelcomePages/AddNetworkOrLocalPrinterContent.qml [hunk 2]
```diff
@@ -104,6 +104,20 @@ Item
         }
     }
 
+    // This "Back" button only shows in the "Add Machine" dialog, which has "back_button_text" set to "Cancel"
```
Ghostkeeper: This comment will get outdated FAST :grin: Better to describe when it should be shown and when not.

cura/Utils/networking.py
LipuFei: Done in 8c42ceb8e603151d1132bff24a07e02cf70f1f7d

resources/qml/WelcomePages/AddNetworkOrLocalPrinterContent.qml
LipuFei: Fixed in bb1442a04e1239693892cbea012bd6b9b7018727

LipuFei:
Merged them into a single `NetworkingUtil` in ed8127777

cura/CuraApplication.py
```diff
@@ -1028,6 +1030,8 @@ def registerObjects(self, engine):
         qmlRegisterSingletonType(SimpleModeSettingsManager, "Cura", 1, 0, "SimpleModeSettingsManager", self.getSimpleModeSettingsManager)
         qmlRegisterSingletonType(MachineActionManager.MachineActionManager, "Cura", 1, 0, "MachineActionManager", self.getMachineActionManager)
 
+        qmlRegisterType(NetworkingUtil, "Cura", 1, 0, "NetworkingUtil")
```
nallath:
```suggestion
        qmlRegisterType(NetworkingUtil, "Cura", 1, 5, "NetworkingUtil")
```
This number is supposed to indicate an API of sorts. As it was added later, we should let it have a higher number.

cura/Settings/GlobalStack.py
```diff
@@ -33,6 +35,7 @@ def __init__(self, container_id: str) -> None:
         super().__init__(container_id)
 
         self.setMetaDataEntry("type", "machine")  # For backward compatibility
+        self.setMetaDataEntry("group_id", str(uuid.uuid4()))  # Assign a new GlobalStack to a unique group by default
```
nallath: We now still know what group means and what it does, but lets make sure that we still know it months / years from now. We should have some documentation here as to what the point of the group_id is and what it means.

cura/CuraApplication.py
LipuFei: Done in 40d26bbb6ecff89540bcc2ccbe2797440339df4e

cura/Settings/GlobalStack.py
LipuFei: Done in 0cc00efa7983f18de6aa472be2f24b9c802a2630

PR #5674, opened 2019-04-29
Restyle send data dialog (CURA-6434)

diegopradogesto:
This PR changes the welcome screen that shows which data we collect.

Here below are a couple of screenshots:

image
image


plugins/SliceInfoPlugin/example_data.html
```diff
@@ -0,0 +1,40 @@
+Cura Version: 4.0<br/>
```
nallath: There is one set of data that is not in the example, we also send some "per object" data (eg; transformation meshes, mode hashes and axis aligned bounding boxes)

LipuFei: Added them and improved the HTML file in c26dbfd

Ghostkeeper:
A bug: I'm unable to scroll through the data. It does scroll if I select the text and drag the cursor down, but there are no scroll bars and the scroll wheel doesn't react either.

PR #5677, opened 2019-04-30
Fix rendering issues in Monitor Page

diegopradogesto:
I saw this many times before in macOS when connecting to a secondary screen. I added some missed renderType properties in the Labels in QML.

**Before**:
Screenshot 2019-04-30 at 11 35 12

**After**:
Screenshot 2019-04-30 at 11 32 54

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/MonitorBuildplateConfiguration.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/MonitorBuildplateConfiguration.qml
@@ -69,6 +69,7 @@ Item
             // FIXED-LINE-HEIGHT:
             height: 18 * screenScaleFactor // TODO: Theme!
             verticalAlignment: Text.AlignVCenter
+            renderType: Text.NativeRendering
         }
     }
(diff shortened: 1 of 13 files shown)


ChrisTerBeke:
Can't we make a "UltimakerLabel" component that has all the right settings applied to it by default? That would prevent a lot of code duplication.

diegopradogesto:
That would be a good idea.

nallath:
I'd prefer having something like UM.Label {}, but yeah that makes sense. We've already been working on getting more of those default things into a re-usable component.

fieldOfView:
It would be great if that UM.Label could automatically align itself to a whole pixel offset so all of that rounding fractional offsets in all QML won't be as critical anymore.

ianpaschal:
Thanks for the assistance @diegopradogesto. I agree a component would be dandy and I will use them in the monitor tab if you guys build it, but this seems like a fine fix to me for the time being.

ChrisTerBeke:
It's def out of scope for this PR

PR #5679, opened 2019-04-30
Fix error when libCharon is not available

fieldOfView:
This PR fixes an ugly error when libCharon is not available on the system. The same fix is also in place in UFPWriter, and similar fixes for a missing libSavitar are in 3MFReader and 3MFWriter. Functionally the PR has no impact, but it cleans up logs.


nallath:
CURA-6493

PR #5707, opened 2019-05-07
Added ids to materials for squish tests

SatishBiharie93:
-Added a text property to MaterialsBrandSection
-Added ids to materials menu buttons


resources/qml/Preferences/Materials/MaterialsBrandSection.qml
```diff
@@ -13,7 +13,8 @@ import Cura 1.0 as Cura
 Item
 {
     id: brand_section
-    
+
+    property var text: ""
```
Ghostkeeper: We don't need this property. We can get it just as well from the `sectionName` custom property.

PR #5718, opened 2019-05-09
Add option to prefer hiding seams in a smarter way

Ghostkeeper:
This option prefers hiding, but if no good hiding spot is available it'll expose the seam on the sharpest outer corner.

Contributes to issue CURA-6489.


Ghostkeeper:
See also https://github.com/Ultimaker/CuraEngine/pull/1052 for the implementation.

PR #5723, opened 2019-05-09
Adjust max Z-axis jerk on Ender 3

GunZi200:
I suggest a change to the value of max  Z-axis Jerk to be the same as on Marlin 1.1.9 firmware used by Ender 3. Its a small change, reducing 0,4 to 0,3. Having these settings match the firmware's configuration makes a print's estimated time more accurate.

### DIFF
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -60,7 +60,7 @@
             "value": "jerk_print"
         },
         "machine_max_jerk_z": {
-            "default_value": 0.3,
+            "default_value": 0.3
         },
         "layer_height_0": {
             "default_value": 0.2
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -59,6 +59,9 @@
         "jerk_travel": {
             "value": "jerk_print"
         },
+        "machine_max_jerk_z": {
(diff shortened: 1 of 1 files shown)


resources/definitions/creality_ender3.def.json
```diff
@@ -59,6 +59,9 @@
         "jerk_travel": {
             "value": "jerk_print"
         },
+        "machine_max_jerk_z": {
+            "default_value": 0.3,
```
Ghostkeeper: This doesn't parse as JSON because of the extra comma at the end.

Did you test this in Cura?

GunZi200: No I haven't tested it, I tried to edit my local file but couldn't find a way to do so. I'm very new at this.

Ghostkeeper: If you're on Windows you could edit the file at `C:\Program Files\Ultimaker Cura 4.0\resources\definitions\creality_ender3.def.json`, but you need administrator rights. If you're on Linux the AppImage is an ISO image that you could edit. If you're on MacOS the distribution is also an ISO image but since it is signed you'd need to disable the signature checking for it to run after editing it.

Since the impact of this setting change is minor and it has a good basis in theory I think I'll just accept the change like this without you testing it.

Issue #5701, opened 2019-05-06
Consider updating Creality Ender-3, CR-10 and CR-10 Mini profiles


probonopd:
**Application Version**
Any

**Platform**
Any

**Printer**
Creality Ender-3, CR-10 and CR-10 Mini

**Steps to Reproduce**
Some people say the [Creality Ender-3, CR-10 and CR-10 Mini profiles](https://www.chepclub.com/cura-profiles.html) by Chuck Hellebuyck (from Filament Friday) are superior to the ones that come with Cura. He has given permission to include them in Cura: https://twitter.com/ChuckHellebuyck/status/1125160431541211136

**Additional Information**
I did not run my own comparison tests yet and I would appreciate if someone really experienced in printer profiles could review them and maybe run some tests so that the project can decide, after a discussion, whether including those profiles (or using parts of them) does indeed increase the overall performance of the Creality Ender-3, CR-10 and CR-10 Mini printers.

nallath:
We're fine with it either way. If someone makes a pull request for machine profiles, we only check if it's not "sabotage" and if they don't break anything (as that's all we can do without actually having any of those printers)

Some work would need to be done as these would need to be added as "quality" types (and not "quality_changes").

Ghostkeeper:
Keep in mind that we've had a minor update to Ender 3 in the meanwhile from someone else as well: https://github.com/Ultimaker/Cura/pull/5456

probonopd:
It appears that the most popular Creality profiles right now are in [CreawsomeMod](
https://github.com/trouch/CreawsomeMod/), and it looks like the author is in discussion with the Cura team to have those integrated.

https://github.com/trouch/CreawsomeMod/issues/27

stelgenhof:
Perhaps this issue is not the perfect place to discuss, however would like to share a concern I have. 

The community has been active in producing a series of Creality profiles for Cura that all are good in their own ways. And some members (myself included) have submitted PR's to embed some of these settings into Cura. So, this raises some questions, at least for me:

- How do we govern this? 
- There were some initial definitions for a number of Creality printers, then some more improvements, now the Creawsome Mod, and what about tomorrow? 
- The beauty of Open Source is of course that anybody can make/request changes, however how do we know which are the right ones when there are many out there? 
- Who decides which changes are deemed correct and who makes sure these are fine so the end users don't experience a plethora of alterations with each Cura release?

Not expecting a definitive answer here, but just some questions that came to my mind lately :)

[...]

PR #5724, opened 2019-05-09
Adjust retraction distance and retraction speed on Ender 3

GunZi200:
#5701 Adjusted retraction distance and retraction speed for Ender 3. 

These settings work well for me. I can't say for certain if these settings really improve the print over the previous settings. 

Chuck talks about these numbers specifically in his video: https://www.youtube.com/watch?time_continue=6&v=GXSqZ68UdsE


Ghostkeeper:
This works. Thanks!

PR #5725, opened 2019-05-10
Minor fixes to the Marketplace UI

diegopradogesto:
For some reason, using the ScrollView with Qt.Controls 1 created blurry texts when connecting a secondary screen. Changing the ScrollView to Controls 2 did the trick. However I also did some other minor changes.

In a new PR (coming soon) I will align all the styles for all places where we use scroll views.


plugins/Toolbox/resources/qml/ToolboxInstalledPage.qml
```diff
@@ -1,50 +1,40 @@
...
-            left: parent.left
-            margins: UM.Theme.getSize("default_margin").width
-            top: parent.top
-        }
+        height: childrenRect.height + 2 * UM.Theme.getSize("wide_margin").height
 
         Label
         {
-            width: page.width
+            width: parent.width - 2 * parent.padding
```
nallath: It's better to anchor it (according to the QML best practices, it's faster that way).
As it's parent's width is already set, it would be something like
anchors.left: parent.left
anchors.right: parent.right
anchors.margins: parent.padding

plugins/Toolbox/resources/qml/ToolboxInstalledPage.qml
```diff
@@ -1,50 +1,40 @@
...
             font: UM.Theme.getFont("large")
             renderType: Text.NativeRendering
         }
+
         Rectangle
         {
             id: installedPlugins
             color: "transparent"
-            width: parent.width
+            width: parent.width - 2 * parent.padding
```
nallath: Same as above.

plugins/Toolbox/resources/qml/ToolboxInstalledPage.qml
```diff
@@ -65,8 +55,10 @@ ScrollView
                 }
             }
         }
+
         Label
         {
+            width: parent.width - 2 * parent.padding
```
nallath: Same as above

plugins/Toolbox/resources/qml/ToolboxInstalledPage.qml
```diff
@@ -77,7 +69,7 @@ ScrollView
         {
             id: installedMaterials
             color: "transparent"
-            width: parent.width
+            width: parent.width - 2 * parent.padding
```
nallath: Same as above

plugins/Toolbox/resources/qml/ToolboxProgressButton.qml
```diff
@@ -25,43 +25,36 @@ Item
 
     width: UM.Theme.getSize("toolbox_action_button").width
     height: UM.Theme.getSize("toolbox_action_button").height
-
-    Cura.PrimaryButton
+    fixedWidthMode: true
+    text:
```
nallath: Might want to have a look at states for this (https://doc.qt.io/qt-5/qml-qtquick-state.html) I find that a bit cleaner than very long if / else if statements. It also nicely groups the behavior on state change.

nallath:
Well, that's the problem of leaving it hang for to long, it now prevents the title items in the marketplace from being selected (it complains about` Cura/plugins/Toolbox/resources/qml/ToolboxDownloadsGridTile.qml:27: Error: Cannot assign to non-existent property "selection"`)

plugins/Toolbox/resources/qml/ToolboxProgressButton.qml
diegopradogesto: I didn't do this change. Maybe next time. But seems like a good idea.

PR #5726, opened 2019-05-10
CL-1331 Extending the Cura API for the Network Printing Plugin

ianpaschal:
> Heavily WIP! Mainly a conversation starter for now.

This PR is intended to make the network printing plugin less knowledgable about the internal structure of Cura in the hope that it will make it easier to maintain.

Specifically this means:

1. No more touching of container stacks. That's Cura's internal way to represent things and should not be our concern. We only want to handle "printers" and their respective connection types.
2. Output devices, container stacks, and metadata are all often different pieces of data related to a single physical machine (or cluster) which we want to operate on. Since Cura doesn't have a machine class (yet), we bundle that data into a more usable object via the API.
3. Reduced use of signals because they are icky and you never know who is triggering them to emit and who is listening on the other end of the line and what other pieces of the code you might be triggering to run. I know they're handy sometime but maintainability and readability is improved if their use is moderated.
4. General reduction of the scope of the plugin so that it serves as a minimalist-as-possible glue layer between Zeroconf, Ultimaker Cloud, and the rest of Cura.

### DIFF
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -7,7 +7,7 @@ from UM.i18n import i18nCatalog
 from UM.Logger import Logger
 if TYPE_CHECKING:
     from cura.CuraApplication import CuraApplication
-    from cura.PrinterOutput.PrinterOutputDevice import PrinterOutputDevice
+    from cura.PrinterOutput.PrinterOutputDevice import PrinterOutputDevice, ConnectionType
 
 i18n_catalog = i18nCatalog("cura")
 
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -7,6 +7,7 @@ from UM.i18n import i18nCatalog
 from UM.Logger import Logger
 if TYPE_CHECKING:
     from cura.CuraApplication import CuraApplication
+    from cura.PrinterOutput.PrinterOutputDevice import PrinterOutputDevice
 
 i18n_catalog = i18nCatalog("cura")
 
@@ -37,7 +38,7 @@ class Machines(QObject):
         self._application = application
 
     @pyqtSlot(result="QVariantMap")
-    def getCurrentMachine(self) -> "QVariantMap":
+    def getCurrentMachine(self) -> Machine:
         fake_machine = Machine() # type: Machine
         global_stack = self._application.getGlobalContainerStack()
         if global_stack:
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -68,6 +68,7 @@ class Machines(QObject):
     ##  Set the current machine's configuration from an (optional) output device.
     #   If no output device is given, the first one available on the machine will be used.
     #   NOTE: Group and machine are used interchangeably.
+    #   NOTE: This doesn't seem to be used anywhere. Maybe delete?
     @pyqtSlot(QObject)
     def updateCurrentMachineConfiguration(self, output_device: Optional["PrinterOutputDevice"]) -> None:
 
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2019 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
-from typing import Optional, Dict, TYPE_CHECKING
+from typing import Optional, Dict,  List, TYPE_CHECKING
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty
 from UM.i18n import i18nCatalog
 from UM.Logger import Logger
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -27,20 +27,26 @@ class Machines(QObject):
 
     @pyqtSlot(result=dict)
     def getCurrentMachine(self) -> dict:
+        # Since Cura doesn't have a machine class, we're going to make a fake one to make our
+        # lives a little bit easier.
+        fake_machine = {
+            "hostname": "",
+            "group_id": "",
+            "group_name": "",
+            "um_network_key": "",
+            "configuration": {}
+        }
         global_stack = self._application.getGlobalContainerStack()
         if global_stack:
             metadata = global_stack.getMetaData()
-        
-            # Since Cura doesn't have a machine class, we're going to make a fake one to make our
-            # lives a little bit easier.
-            fake_machine = {
-                "hostname": "",
-                "group_id": global_stack.getMetaDataEntry("group_id") if "group_id" in metadata else "",
-                "group_name": global_stack.getMetaDataEntry("group_name") if "group_name" in metadata else "",
-                "um_network_key": global_stack.getMetaDataEntry("um_network_key") if "um_network_key" in metadata else "",
-                "configuration": {}
-            }
-            return fake_machine
+            if "group_id" in metadata:
+                fake_machine["group_id"] = global_stack.getMetaDataEntry("group_id")
+            if "group_name" in metadata:
+                fake_machine["group_name"] = global_stack.getMetaDataEntry("group_name")
+            if "um_network_key" in metadata:
+                fake_machine["um_network_key"] = global_stack.getMetaDataEntry("um_network_key")
+            
+        return fake_machine
 
     ##  Set the current machine's friendy name.
     #   This is the same as "group name" since we use "group" and "current machine" interchangeably.
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2019 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
-from typing import Optional, Dict,  List, TYPE_CHECKING
+from typing import Optional, Dict,  List, TYPE_CHECKING, Any
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty
 from UM.i18n import i18nCatalog
 from UM.Logger import Logger
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -25,6 +25,58 @@ class Machines(QObject):
         super().__init__(parent)
         self._application = application
 
+    @pyqtSlot(result=dict)
+    def getCurrentMachine(self) -> dict:
+        global_stack = self._application.getGlobalContainerStack()
+        if global_stack:
+            metadata = global_stack.getMetaData()
+        
+            # Since Cura doesn't have a machine class, we're going to make a fake one to make our
+            # lives a little bit easier.
+            fake_machine = {
+                "hostname": "",
+                "group_id": global_stack.getMetaDataEntry("group_id") if "group_id" in metadata else "",
+                "group_name": global_stack.getMetaDataEntry("group_name") if "group_name" in metadata else "",
+                "um_network_key": global_stack.getMetaDataEntry("um_network_key") if "um_network_key" in metadata else "",
+                "configuration": {}
+            }
+            return fake_machine
+
+    ##  Set the current machine's friendy name.
+    #   This is the same as "group name" since we use "group" and "current machine" interchangeably.
+    #   TODO: Maybe make this "friendly name" to distinguish from "hostname"?
+    @pyqtSlot(str)
+    def setCurrentMachineGroupName(self, group_name: str):
+        Logger.log("d", "Attempting to set the group name of the active machine to %s", group_name)
+        global_stack = self._application.getGlobalContainerStack()
+        if global_stack:
+            # Update a GlobalStacks in the same group with the new group name.
+            group_id = global_stack.getMetaDataEntry("group_id")
+            machine_manager = self._application.getMachineManager()
+            for machine in machine_manager.getMachinesInGroup(group_id):
+                machine.setMetaDataEntry("group_name", group_name)
+
+            # Set the default value for "hidden", which is used when you have a group with multiple types of printers
+            global_stack.setMetaDataEntry("hidden", False)
+
+    ##  Set the current machine's configuration from an (optional) output device.
+    #   If no output device is given, the first one available on the machine will be used.
+    #   NOTE: Group and machine are used interchangeably.
+    @pyqtSlot(QObject)
+    def updateCurrentMachineConfiguration(self, output_device: Optional["PrinterOutputDevice"]) -> None:
+
+        if output_device is None:
+            machine_manager = self._application.getMachineManager()
+            output_device = machine_manager.printerOutputDevices[0]
+        
+        hotend_ids = output_device.hotendIds
+        for index in range(len(hotend_ids)):
+            output_device.hotendIdChanged.emit(index, hotend_ids[index])
+        
+        material_ids = output_device.materialIds
+        for index in range(len(material_ids)):
+            output_device.materialIdChanged.emit(index, material_ids[index])
+
     ##  Add an output device to the current machine.
     #   In practice, this means:
     #   - Setting the output device's network key in the current machine's metadata
@@ -32,33 +84,31 @@ class Machines(QObject):
     #     types.
     #   TODO: CHANGE TO HOSTNAME
     @pyqtSlot(QObject)
-    def addOutputDeviceToCurrentMachine(self, output_device):
+    def addOutputDeviceToCurrentMachine(self, output_device: "PrinterOutputDevice") -> None:
         if not output_device:
             return
-
         Logger.log("d",
             "Attempting to set the network key of the active machine to %s",
             output_device.key)
-
-        global_container_stack = self._application.getGlobalContainerStack()
-        if not global_container_stack:
+        global_stack = self._application.getGlobalContainerStack()
+        if not global_stack:
             return
+        metadata = global_stack.getMetaData()
 
-        metadata = global_container_stack.getMetaData()
-
-        if "um_network_key" in metadata:  # Global stack already had a connection, but it's changed.
+        # Global stack already had a connection, but it's changed.
+        if "um_network_key" in metadata: 
             old_network_key = metadata["um_network_key"]
+
             # Since we might have a bunch of hidden stacks, we also need to change it there.
             metadata_filter = {"um_network_key": old_network_key}
             containers = self._application.getContainerRegistry().findContainerStacks(
                 type = "machine", **metadata_filter)
-
             for container in containers:
                 container.setMetaDataEntry("um_network_key", output_device.key)
 
                 # Delete old authentication data.
                 Logger.log("d", "Removing old authentication id %s for device %s",
-                    global_container_stack.getMetaDataEntry("network_authentication_id", None),
+                    global_stack.getMetaDataEntry("network_authentication_id", None),
                     output_device.key)
 
                 container.removeMetaDataEntry("network_authentication_id")
@@ -69,8 +119,8 @@ class Machines(QObject):
                 container.addConfiguredConnectionType(output_device.connectionType.value)
 
         else:  # Global stack didn't have a connection yet, configure it.
-            global_container_stack.setMetaDataEntry("um_network_key", output_device.key)
-            global_container_stack.addConfiguredConnectionType(output_device.connectionType.value)
+            global_stack.setMetaDataEntry("um_network_key", output_device.key)
+            global_stack.addConfiguredConnectionType(output_device.connectionType.value)
 
         return None
     
--- a/cura/API/Machines.py
+++ b/cura/API/Machines.py
@@ -25,8 +25,8 @@ class Machines(QObject):
         super().__init__(parent)
         self._application = application
 
-    @pyqtSlot(result=dict)
-    def getCurrentMachine(self) -> dict:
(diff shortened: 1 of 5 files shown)


ianpaschal:
Specifically I'm curious if there is a better way to organize these new API points.

ChrisTerBeke:
I think this makes sense from a level of abstraction point of view. Cura team can probably give more in-depth comments on the API design.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+##  Since Cura doesn't have a machine class, we're going to make a fake one to make our lives a
+#   little bit easier.
+class Machine():
+    def __init__(self) -> None:
+        self.hostname = "" # type: str
+        self.group_id = "" # type: str
+        self.group_name = "" # type: str
+        self.um_network_key = "" # type: str
+        self.configuration = {} # type: Dict
```
sedwards2009: If at all possible the `Dict` needs to be specialised, i.e. `Dict[str,str]`.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+        self.connection_types = [] # type: List
```
sedwards2009: If at all possible the List needs to be specialised, i.e. `List[str]`.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+            fake_machine.connection_types = global_stack.configuredConnectionTypes
+            
+        return fake_machine
+
+    ##  Set the current machine's friendy name.
+    #   This is the same as "group name" since we use "group" and "current machine" interchangeably.
+    #   TODO: Maybe make this "friendly name" to distinguish from "hostname"?
+    @pyqtSlot(str)
+    def setCurrentMachineGroupName(self, group_name: str):
+        Logger.log("d", "Attempting to set the group name of the active machine to %s", group_name)
```
sedwards2009: Debug logging.

sedwards2009: Can this return `-> None`.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Set the default value for "hidden", which is used when you have a group with multiple types of printers
+            global_stack.setMetaDataEntry("hidden", False)
+
+    ##  Set the current machine's configuration from an (optional) output device.
+    #   If no output device is given, the first one available on the machine will be used.
+    #   NOTE: Group and machine are used interchangeably.
+    #   NOTE: This doesn't seem to be used anywhere. Maybe delete?
+    @pyqtSlot(QObject)
+    def updateCurrentMachineConfiguration(self, output_device: Optional["PrinterOutputDevice"]) -> None:
```
sedwards2009: Is this really an `update`? Or is it semantically the same as making the given output device the active one?

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if output_device is None:
```
sedwards2009: Thank you for writing this as `... is None:`

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Since we might have a bunch of hidden stacks, we also need to change it there.
+            metadata_filter = {"um_network_key": old_network_key}
+            containers = self._application.getContainerRegistry().findContainerStacks(
+                type = "machine", **metadata_filter)
+            for container in containers:
+                container.setMetaDataEntry("um_network_key", output_device.key)
+
+                # Delete old authentication data.
+                Logger.log("d", "Removing old authentication id %s for device %s",
```
sedwards2009: `"i"`?

plugins/UM3NetworkPrinting/resources/qml/DiscoverUM3Action.qml
```diff
@@ -27,13 +27,14 @@ Cura.MachineAction
         {
             var printerKey = base.selectedDevice.key
             var printerName = base.selectedDevice.name  // TODO To change when the groups have a name
-            if (manager.getStoredKey() != printerKey)
+            if (Cura.API.machines.getCurrentMachine()["um_network_key"] != printerKey)
```
sedwards2009: Use the machine class/instance, -> `.getCurrentMachine().um_network_key`

plugins/UM3NetworkPrinting/resources/qml/DiscoverUM3Action.qml
```diff
@@ -156,7 +157,7 @@ Cura.MachineAction
                             var selectedKey = manager.getLastManualEntryKey()
                             // If there is no last manual entry key, then we select the stored key (if any)
                             if (selectedKey == "")
-                                selectedKey = manager.getStoredKey()
+                                selectedKey = Cura.API.machines.getCurrentMachine()["um_network_key"]
```
sedwards2009: Same here. This needs to be updated.

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -272,46 +273,22 @@ def _createMachineFromDiscoveredPrinter(self, key: str) -> None:
...
-
-            # Ensure that these containers do know that they are configured for network connection
-            machine.addConfiguredConnectionType(printer_device.connectionType.value)
+        self._api.machines.addOutputDeviceToCurrentMachine(discovered_device)
 
+        # ensure that the connection states are refreshed.
         self.refreshConnections()
 
-    def _checkManualDevice(self, address: str) -> "QNetworkReply":
+    def _checkManualDevice(self, address: str):
```
sedwards2009: Needs a return type.

cura/API/Machines.py
ianpaschal: Debug logging was in existing code already. Don't see any reason to remove it; it's useful to have in there.

cura/API/Machines.py
ianpaschal: Honestly, I'm not sure. It looks like a weird hack where we force an update by emitting signals but it's not actually used anywhere as far as I can tell so it's tough to know what precisely the goal is.

I'd like to leave it in for now but will probably remove it in the near future.

cura/API/Machines.py
ianpaschal: 👌

cura/API/Machines.py
ianpaschal: `"d"` fits with the other debug statements. It's the sort of thing you only really care about when debugging.

cura/API/Machines.py
sedwards2009: Do you want to add a TODO or maybe a warning here?

cura/API/Machines.py
sedwards2009: cool

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+#   Usage:
+#       ```
+#       from cura.API import CuraAPI
+#       api = CuraAPI()
+#       api.machines.addOutputDeviceToCurrentMachine()
+#       ```
+
+##  Since Cura doesn't have a machine class, we're going to make a fake one to make our lives a
+#   little bit easier.
+class Machine():
```
nallath: This is not a public class, it's a private one. By making it public, it becomes a new interface (which is clearly not what we want)

ianpaschal: Could you please clarify what you mean by plugin specific logic and also what do you propose as an alternative? In our discussion you said Cura needs a Machine class, my hope was that this could eventually be depreciated in favor of that, but for now is just a stop-gap.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+                # Ensure that these containers do know that they are configured for the given
+                # connection type (can be more than one type; e.g. LAN & Cloud)
+                container.addConfiguredConnectionType(output_device.connectionType.value)
+
+        else:  # Global stack didn't have a connection yet, configure it.
+            global_stack.setMetaDataEntry("um_network_key", output_device.key)
+            global_stack.addConfiguredConnectionType(output_device.connectionType.value)
+
+        return None
```
Ghostkeeper: This statement does nothing.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -34,7 +34,10 @@ def __init__(self) -> None:
 
         self.__additional_components_view = None #type: Optional[QObject]
 
-        CuraApplication.getInstance().engineCreatedSignal.connect(self._createAdditionalComponentsView)
+        self._application = CuraApplication.getInstance()
```
Ghostkeeper: Don't cache singletons (see above).

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -34,7 +34,10 @@ def __init__(self) -> None:
...
+        self._api = self._application.getCuraAPI()
```
Ghostkeeper: Don't cache singletons (see above).

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -105,63 +108,27 @@ def foundDevices(self):
...
-        global_container_stack = CuraApplication.getInstance().getGlobalContainerStack()
-        if global_container_stack:
-            meta_data = global_container_stack.getMetaData()
-            if "um_network_key" in meta_data:
-                return global_container_stack.getMetaDataEntry("um_network_key")
-
-        return ""
-
+    # TODO: Improve naming
+    # TODO: CHANGE TO HOSTNAME
```
Ghostkeeper: Do these?

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
```diff
@@ -105,63 +108,27 @@ def foundDevices(self):
...
     @pyqtSlot(result = str)
     def getLastManualEntryKey(self) -> str:
         if self._network_plugin:
             return self._network_plugin.getLastManualDevice()
         return ""
 
+    # TODO: Better naming needed. Exists where? On the current machine? On all machines?
+    # TODO: CHANGE TO HOSTNAME
```
Ghostkeeper: Do these?

plugins/UM3NetworkPrinting/src/UM3OutputDevicePlugin.py
```diff
@@ -67,11 +67,11 @@ class UM3OutputDevicePlugin(OutputDevicePlugin):
 
     def __init__(self):
         super().__init__()
-        
         self._zero_conf = None
         self._zero_conf_browser = None
 
         self._application = CuraApplication.getInstance()
+        self._api = self._application.getCuraAPI()
```
Ghostkeeper: Don't cache singletons (see above).

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # Set the default value for "hidden", which is used when you have a group with multiple types of printers
+            global_stack.setMetaDataEntry("hidden", False)
+
+    ##  Set the current machine's configuration from an (optional) output device.
+    #   If no output device is given, the first one available on the machine will be used.
+    #   NOTE: Group and machine are used interchangeably.
+    #   NOTE: This doesn't seem to be used anywhere. Maybe delete?
+    @pyqtSlot(QObject)
+    def updateCurrentMachineConfiguration(self, output_device: Optional["PrinterOutputDevice"]) -> None:
```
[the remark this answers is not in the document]
ianpaschal: Already known, see the comment at the top. Unsure why though or what it was used for since there was never any documentation with it. Wasn't going to delete it until I knew why.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+class Machine():
+    def __init__(self) -> None:
+        self.hostname = "" # type: str
+        self.group_id = "" # type: str
+        self.group_name = "" # type: str
+        self.um_network_key = "" # type: str
+        self.configuration = {} # type: Dict[str, Any]
+        self.connection_types = [] # type: List[int]
+
+class Machines(QObject):
```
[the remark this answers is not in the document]
ianpaschal: What would you call this then?

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+#
+#   Usage:
+#       ```
+#       from cura.API import CuraAPI
+#       api = CuraAPI()
+#       api.machines.addOutputDeviceToCurrentMachine()
+#       ```
+
+##  Since Cura doesn't have a machine class, we're going to make a fake one to make our lives a
+#   little bit easier.
```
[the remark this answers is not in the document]
ianpaschal: Fair enough. It used to just be a dict in a REST style but Simon preferred it as a class. Should have updated it.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
ianpaschal: No. That's the next step.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
ianpaschal: No. That's the next step.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+        fake_machine = Machine() # type: Machine
+        global_stack = self._application.getGlobalContainerStack()
+        if global_stack:
+            metadata = global_stack.getMetaData()
+            if "group_id" in metadata:
+                fake_machine.group_id = global_stack.getMetaDataEntry("group_id")
+            if "group_name" in metadata:
+                fake_machine.group_name = global_stack.getMetaDataEntry("group_name")
+            if "um_network_key" in metadata:
+                fake_machine.um_network_key = global_stack.getMetaDataEntry("um_network_key")
```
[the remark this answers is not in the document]
ianpaschal: Proposal to change that? We're not going to interact directly with global stack.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+#       ```
+
+##  Since Cura doesn't have a machine class, we're going to make a fake one to make our lives a
+#   little bit easier.
+class Machine():
+    def __init__(self) -> None:
+        self.hostname = "" # type: str
+        self.group_id = "" # type: str
+        self.group_name = "" # type: str
+        self.um_network_key = "" # type: str
```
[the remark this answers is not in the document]
ianpaschal: In what way is it actually UM specific? It's just the `NetworkedPrinterOutputDevice.key` property used everywhere else.

Either way, part of the rationale behind all this is to stop using keys and start using host names to keep track of what is what.

cura/API/Machines.py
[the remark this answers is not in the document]
ianpaschal: Unfortunately having a single "machine" and not a bunch of disparate metadata, settings, configurations, output devices, connection types, etc. is one of the requirements for us to continue maintaining this plugin.

If you have another idea how to do it, let me know, but Jaime also said to me that Cura is overdue for having an actual Machine class. ;)

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: It's the `um_network_key`, which is used to identify the connection with the UM3NetworkPrinting plug-in. It's only used by that plug-in and other plug-ins are not allowed and not able to use it. For instance, USB printing would need COM port names and OctoPrint has their own keys. Those two examples are not necessarily things that need to be implemented now, but enforcing this is a way to enforce separation of concern.

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: When did your requirements of maintaining this plug-in change, then?

We can help you think of architectures to store the data you need to store, if you need help.

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: That depends on what the class needs to do. Currently it seems to track remote printers to output towards, but is restricting itself in the API by getting only one of them, the UM3NetworkConnection. And it allows changing some metadata of the client-server connection. Looking at that, I'd go with a name like `PrinterOutputManager` or something of the like. However that collides with the `cura.PrinterOutput.PrinterOutputController` class. That indicates that we might be duplicating some logic with this class in general.

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: I think exposing the global stack is a very good idea to extend our API towards being able to change profiles, but you'd want a facade to direct API users towards the active global stack only.

Alternatively, we could expose the currently available connections through the API by just requesting it every time you make the API call, rather than having to update it every time it changes. Just remove the cache, really. It's not going to save a lot of processing time, because the network connection metadata shouldn't change often. Just once or twice per printer, probably.

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
Ghostkeeper: It's best to not merge it until the feature is done, lest it be decided that the other work gets cancelled. Except if there is a good reason why it's not done yet (as in, it gets blocked by something).

plugins/UM3NetworkPrinting/src/DiscoverUM3Action.py
Ghostkeeper: It's best to not merge it until the feature is done, lest it be decided that the other work gets cancelled. Except if there is a good reason why it's not done yet (as in, it gets blocked by something). You're creating technical debt otherwise.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+        self.configuration = {} # type: Dict[str, Any]
+        self.connection_types = [] # type: List[int]
+
+class Machines(QObject):
+
+    def __init__(self, application: "CuraApplication", parent = None) -> None:
+        super().__init__(parent)
+        self._application = application
```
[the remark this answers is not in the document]
ianpaschal: This is the only way I know to avoid a circular dependency, and was done already by Cura team.

cura/API/Machines.py
```diff
@@ -0,0 +1,139 @@
...
+            output_device.materialIdChanged.emit(index, material_ids[index])
+
+    ##  Add an output device to the current machine.
+    #   In practice, this means:
+    #   - Setting the output device's network key in the current machine's metadata
+    #   - Adding the output device's connection type to the current machine's configured connection
+    #     types.
+    #   TODO: CHANGE TO HOSTNAME
+    @pyqtSlot(QObject)
+    def addOutputDeviceToCurrentMachine(self, output_device: "PrinterOutputDevice") -> None:
```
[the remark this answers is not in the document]
ianpaschal: It wont, since no one else uses this function. we know its only serving the purpose of one plugin right now because we only moved code around as part of the process of figuring out what exactly we need and want to build, with the assumption of more prs incoming soon as we make this usable in more than a single circumstance.

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: Yes, whenever I see this being done by someone in the Cura team or some external contributor, I'm also asking them to change it.

cura/API/Machines.py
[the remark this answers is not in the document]
Ghostkeeper: The intent is to have a single entry point, right? So other features will need to change to use this function as well. Nobody is using it since it got merged a few days ago. I understand that it's not your responsibility to make sure  that Cura works for other connection methods and future circumstances, but it is mine.

It was intended that this functionality would be implemented by the individual plug-ins anyway. That was the architecture design we've discussed a few months ago with you. The plug-in gets a callback to store the output device in a newly added printer so that it can implement its plug-in specific behaviour.

cura/API/Machines.py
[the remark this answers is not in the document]
ianpaschal: Ok but then you guys can figure out how to avoid that circular dependency and in the mean time i'm going to copy the established solution that is in the other API endpoints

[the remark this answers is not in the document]
Ghostkeeper: You can work around this circular dependency by calling `#import cura.CuraApplication` instead of `from cura.CuraApplication import CuraApplication`. In general it's strange that you'd have a circular dependency with CuraApplication from a plug-in though, because that would mean that importing CuraApplication somehow imports your plug-in?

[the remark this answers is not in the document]
ianpaschal: No, it's a circular dependency with Cura importing the API module (presumably for the `.getCuraAPI()` method). Lipu made a commit caching the app instance with the description "Fix circular dependency" or something along those lines. I copied that pattern when extending it.

Ghostkeeper:
We decided now to revert this merge, as discussed with @ianpaschal. We can come up with a better architecture then without deadline later. To reiterate, our two main points of concern are:
1. The naming is too wide. This is a terminology mismatch of what the Cura developers see as a "machine" (i.e. everything related to a printer, including both profiles and network connection) and what the Connect developers see as a "machine" (i.e. the physical device).
2. The functionality is too specific. It's developed to work only with Ultimaker connections. We expect this API endpoint to work with any type of remote connection.

PR #5735, opened 2019-05-13
HMS434 updates

maukcc:
Small updates for the HMS434 machine


resources/definitions/hms434.def.json
```diff
@@ -144,6 +144,8 @@
         "skirt_gap":                 {"value": 1},
         "skirt_brim_minimal_length": {"value": 50},
 
+        "prime_tower_enable":        {"value": false},
```
Ghostkeeper: This key is also present earlier in this dictionary. It's now duplicate.

PR #5736, opened 2019-05-13
Added id to print setup menu to identify profile used in automated test

SatishBiharie93:


resources/qml/PrintSetupSelector/PrintSetupSelectorHeader.qml
```diff
@@ -15,6 +15,7 @@ RowLayout
 
     Cura.IconWithText
     {
+        id: printSetupSelectorProfileText
```
Ghostkeeper: It doesn't indicate the profile but the layer height here. I suggest changing it to `printSetupSelectorLayerHeightText`, but you'd have to change that in your test as well.

diegopradogesto:
Not needed.

PR #5738, opened 2019-05-14
Amazing fixes for cloud discovery

ChrisTerBeke:
This PR fixes some small bugs in the cloud discovery flow and make it more robust:

* Allow printer_type field from cloud cluster list endpoint (will make a cloud PR that adds this field to the /clusters endpoint so we know the printer type without having to do a 2nd call to /status). This is more robust than the current solution that tries to map the machine variant to printer type.
* Use a default value for printer type (discovered printer list does not always fully refresh once actual data comes in).
* Using `device.key` instead of `cluster.cluster_id` to be more consistent between the add, update and remove function calls for discovered devices.
* Added some missing documentation.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
@@ -58,14 +58,6 @@ class CloudOutputDevice(NetworkedPrinterOutputDevice):
     # Therefore we create a private signal used to trigger the printersChanged signal.
     _clusterPrintersChanged = pyqtSignal()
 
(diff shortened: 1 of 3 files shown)


Ghostkeeper:
Was there a Jira ticket for this?

ChrisTerBeke:
There is not yet, I made these fixes right after testing the Beta build. I think it's up to the connect team to make a ticket if they want to?

ianpaschal:
This seems OK, but we'll have to update it again anyway to use the host name instead of the um network key.

ChrisTerBeke:
This is for the 4.1 branch, I guess that other change should be in master at some point right?

PR #5750, opened 2019-05-17
Updating NWA3D A5 for 4.1 Release

DragonJe:
If possible, please merge these updates to the nwa3d a5 for the 4.1 release.

Thank you for your help!


Ghostkeeper:
Works. Indeed, 100mm line widths is outrageous. I don't know how we didn't see that before.

I think I'm going to remove the entire line from the profiles though, looking more closely at it. Currently, the outer wall line width is 0.4mm no matter what. If the user adjusts their line width setting, the outer wall line width remains 0.4mm. If we remove the override from the profile, the outer wall line width will become `=wall_line_width` so that automatically adjusts along.

I'll make this change on master, so that the 4.1 release is safe in case you disagree.

DragonJe:
I know! I must have thought it was 100% instead of 0.4.

Thank you for all of your help. Have a great rest of your day!

PR #5761, opened 2019-05-20
Rename UM3NetworkPrinting plugin to UltimakerNetworkPrinting

ianpaschal:
This simply changes the name of the plugin to more accurately reflect the fact that it doesn't only apply to Ultimaker 3 machines.


ianpaschal:
Two questions:

1. The idea is that 4.1 is the last release to bundle this plugin, from there on out we'd update via the Marketplace. What impact exactly does this have on the fact that the plugin is included in the bundled plugins JSON? Should that be removed now? After 4.1, but before 4.2? Please advise.
2. This shouldn't have an effect on translated strings, but maybe on `.po`/`.pot` files. Please advise as well.

nallath:
I think we should wait with this rename for the 4.2 release. We already launched the beta, so we should only commit bugfixes at that point.

ChrisTerBeke:
I would still bundle the plugin in the builds, but then just the latest version that is available at that moment (because we want UM customers to use network printing, so they shouldn't have to install a plugin to do so). Then we can distribute updates more frequently via Marketplace while still having the functionality out of the box.

nallath:
Yeah, that is a hard requirement for any of this to work. But the rename will actually interfere with that. Because if we don't rename it now, we need to put the not renamed (um3networkplugin) on the server. If at some point we do rename it, the bundled ID will be the new name. If we don't watch out, someone might be able to still download the um3networkplugin and have both of them run side by side.

Ghostkeeper:
> What impact exactly does this have on the fact that the plugin is included in the bundled plugins JSON?

If we choose to unbundle this, it's not going to be installed any more in the new version. The user will need to find it in the Marketplace and install it. Indeed as Nallath pointed out, after unbundling it, it's going to be harder to change the name because then you need to change it locally on every user's computer. Not impossible, but harder.

> This shouldn't have an effect on translated strings, but maybe on `.po`/`.pot` files. Please advise as well.

This change doesn't give trouble to the .po and .pot files. It lists where the string came from, but as a comment. If the string has moved to a different file, then there will be a new comment in the .pot file but it won't have any real impact.

Moving the plug-in out of the Cura repository does have an impact. If it's to be bundled but in a separate repository (my preferred solution) then our build system has to pull in those translation items. If it's not going to be bundled any more then it'll need to build its own translation catalogue. This hasn't been done for plug-ins before, but it might be due time to take a look at how that's going to function.

ianpaschal:
> After unbundling it, it's going to be harder to change the name because then you need to change it locally on every user's computer. Not impossible, but harder.

Yeah that was my understanding; hence wanting to do this now and not wait until 4.2 to unbundle.

The Connect team feels very strongly that we do want to "unbundle" but only insofar as to allow us to update before/after/independently of Cura.

It sounds like having it on the Marketplace but bundling whatever was the last released version with Cura would be the way to go?

nallath:
We also have to consider the notion that changing the release cycle (eg; Constantly updating a certain plugin) is also going to cause some annoyance with our users. The vast majority of the users doesn't use the network plugin, so it's going to get old really fast if they constantly get update notifications for it.

Ghostkeeper:
I don't think that this strategy decision can be made through the comments of a pull request here. Shall we discuss this in person?

ChrisTerBeke:
Shall we close this? I have the feeling this name change will cause more problems than it solves.

ianpaschal:
> Shall we close this? I have the feeling this name change will cause more problems than it solves.

Meh, fair enough, because it wasn't really solving a problem than incorrect labelling. I just thought it was a logical thing considering it's not UM3 specific. I'm not that invested in it though and it's now very out of date.

PR #5765, opened 2019-05-20
Logarithmic lithophane thickness

BagelOrb:
Introduction of an option to compute the thicknesses logarithmically.

The transparency model is based on the following reasoning:
- multiplying the thickness by 2 means the light reduction factor is multiplied by itself (e.g. if 2mm lets through only 50% of light then 2x 2mm lets through 50% *50% = 25% of light)
- an image brightness of zero would require an infinite thickness, so we need to scale the image brightnesses such that they fit within the user-specified range of thicknesses.

A user-setting is provided to determine the translucency property of the material used.


BagelOrb:
Will add photos as soon as the prints are finished.

BagelOrb:
input:
mr-t_2

early version based on slightly faulty / uncalibrated formula, compared with normal linear model:

BagelOrb:
Here's the new non-linear luminance-to-thickness function:
image

When the user sets the peak high higher it allows for darker colors and the input image values are scaled to span the new range:
image

Ghostkeeper:
Looks nice. Safe change, too.

Developers, look for issue CURA-6540.

BagelOrb:
I still haven't done a print with the final algorithm and final material parameter.

For pearl white PLA, the transmittance should be approx 50%.

It could be interesting to somewhere publish how to estimate the material parameter, which is as follows:

* You print a print with varying thickness: e.g. a ramp from 0.4mm to 1.4mm.
* You take a picture holding the print up to a light source
* Import the picture in your image editing software
* Make into black-and-white / desaturate
* Check the color at two places which are 1mm thickness apart.
* Calculate their luminance: (pixel value / 255)^2.2
* The transmittance is then given by: thick_luminance / thin_luminance

BagelOrb:
Here's what happens when you turn the max height up to something like 5mm.
Clearly the log model has a better gradient than the linear model.

plugins/ImageReader/ImageReader.py
```diff
@@ -99,12 +101,14 @@ def _generateSceneNode(self, file_name, xz_size, peak_height, base_height, blur_
...
-                height_data[y, x] = avg
+                if use_transparency_model:
+                    height_data[y, x] = (0.299 * math.pow(qRed(qrgb) / 255.0, 2.2) + 0.587 * math.pow(qGreen(qrgb) / 255.0, 2.2) + 0.114 * math.pow(qBlue(qrgb) / 255.0, 2.2))
+                else:
+                    height_data[y, x] = (0.212655 * qRed(qrgb) + 0.715158 * qGreen(qrgb) + 0.072187 * qBlue(qrgb)) / 255 # fast computation ignoring gamma and degamma
 
         Job.yieldThread()
 
-        if not lighter_is_higher:
+        if lighter_is_higher is use_transparency_model:
```
ninovanhooff: use == for equality, `is` for identity checks

plugins/ImageReader/ConfigUI.qml
```diff
@@ -143,6 +143,51 @@ UM.Dialog
...
+                    id: color_model
+                    objectName: "ColorModel"
+                    model: [ catalog.i18nc("@item:inlistbox","Translucency"), catalog.i18nc("@item:inlistbox","Linear") ]
+                    width: 180 * screenScaleFactor
+                    onCurrentIndexChanged: { manager.onColorModelChanged(currentIndex) }
+                }
+            }
+        }
+
+        UM.TooltipArea {
```
ninovanhooff: The transmittance textfield is not used for the linear model right? Then `visible` for this item should depend on the `color_model`

plugins/ImageReader/ImageReader.py
```diff
@@ -124,8 +128,15 @@ def _generateSceneNode(self, file_name, xz_size, peak_height, base_height, blur_
...
             Job.yieldThread()
 
-        height_data *= scale_vector.y
-        height_data += base_height
+        if use_transparency_model:
+            p = 1.0 / math.log(transmittance_1mm / 100.0, 2)
+            min_luminance = 2.0 ** ((peak_height - base_height) / p)
+            for (y, x) in numpy.ndindex(height_data.shape):
+                mapped_luminance = min_luminance + (1.0 - min_luminance) * height_data[y, x]
+                height_data[y, x] = base_height + p * math.log(mapped_luminance, 2)
```
ninovanhooff: Should the log base not be `2.2` here?

plugins/ImageReader/ImageReader.py
BagelOrb: ok fixed

plugins/ImageReader/ImageReader.py
BagelOrb:
```
math.log(mapped_luminance, 2) * p
=
math.log(mapped_luminance, 2) / math.log(transmittance_1mm / 100.0, 2)
=
math.log(mapped_luminance) / log(2) / (math.log(transmittance_1mm / 100.0) / log(2) )
=
math.log(mapped_luminance) / math.log(transmittance_1mm / 100.0)
```
So the base of the log doesn't matter as long as it's the same in `p`.

I added a comment.

plugins/ImageReader/ConfigUI.qml
BagelOrb: I can't get it to work. I'm not even sure it's possible to change the layout of the dialog box depending on some variable automatically. 

Or do you know how it's done?

I don't think it's really important, though.

plugins/ImageReader/ImageReader.py
```diff
@@ -124,8 +128,15 @@ def _generateSceneNode(self, file_name, xz_size, peak_height, base_height, blur_
...
             Job.yieldThread()
 
-        height_data *= scale_vector.y
-        height_data += base_height
+        if use_transparency_model:
+            p = 1.0 / math.log(transmittance_1mm / 100.0, 2) # base doesn't matter here. use base 2 for fast computation
+            min_luminance = 2.0 ** ((peak_height - base_height) / p)
+            for (y, x) in numpy.ndindex(height_data.shape):
+                mapped_luminance = min_luminance + (1.0 - min_luminance) * height_data[y, x]
+                height_data[y, x] = base_height + p * math.log(mapped_luminance, 2) # use same base as a couple lines above this
```
ninovanhooff: Thanks for clarifying that the base should be the same. In that case, you can just remove the second parameter so the default will be used, which is a more powerful way of saying "I don't care what base you use, as long as you are consistent."

Also, please don't use single letter variables. Because code style, but also because I actually don't know what it stands for. Sorry for not spotting it before.

BagelOrb:
sidenote: I don't like merge commits. Please pull with `git pull --rebase`. And update a repo with master using `git rebase master`. The conflicts should be resolved by the PR commits, not by a merge-commit which affects the commits in master.

BagelOrb:
> I agree that it would be nice to explain how to obtain the transmittance value somehow. Maybe a weblink to a markdown file? which is shown/hidden together with the transmittance field?

It already has an explanation in the tooltip:
"The percentage of light penetrating a print with a thickness of 1 millimeter. Lowering this value increases the contrast in dark regions and decreases the contrast in light regions of the image."

plugins/ImageReader/ConfigUI.qml
BagelOrb: Thanks for resolving it.

BagelOrb:
> I agree that it would be nice to explain how to obtain the transmittance value somehow. Maybe a weblink to a markdown file? which is shown/hidden together with the transmittance field?

I'm not sure how to do this and I am not really up for it right now. :/

ninovanhooff:
DSC_0412
Printing the same picture you used, I don't see a lot of difference in the actual result. Will try another picture somewhere this week

BagelOrb:
See my comments on 20 May. The difference between the old model and the new model is quite small if the thickness of the lithophane is small, but the quality of the lithophane improved with thicker ones under my model but the quality reduces under the old model.

i.e. don't try a different picture, try a higher thickness.

ninovanhooff:
After discussion with the team I changed the default to linear because it is easier to understand for users. We do value the benefits for thicker litho's tho, thanks for your contribution!

BagelOrb:
The default should be based on what gives the best results, not what is the easiest to understand.

BagelOrb:
If ease of understanding is the criterion we should for example disable overlap compensation by default and change all of the support setting defaults.

Ghostkeeper:
It's a balance, surely. Ease of use vs. magic behaviour. We've had this problem with overlap compensation too indeed.

If someone is printing a height map it's harder to explain why the height is not what they'd expect. With overlap compensation this is not necessarily a problem because it's not the final print result that becomes weird, just the g-code.

PR #5766, opened 2019-05-21
Fixed USB auto baudrate detection

goldfingyr:
Now done using state machinery.

The original code uses serial.readline which will hang during baudrate mismatch as no newline will be sent by printer.
This fix uses state machinery which will not hang


goldfingyr:
Also changed minimum timeout from 3 to 4 as the Creality 10S-Pro takes 3.56 secs to init

plugins/USBPrinting/AutoDetectBaudJob.py
```diff
@@ -55,32 +55,57 @@ def run(self) -> None:
...
+                        mychar = serial.read()
+                        if mychar == b"o":
+                            mychar = serial.read()
+                            if mychar == b"k":
+                                mychar = serial.read()
+                                if mychar == b" ":
+                                    mychar = serial.read()
+                                    if mychar == b"T":
+                                        mychar = serial.read()
+                                        if mychar == b":":
```
Ghostkeeper: These nested `if`s are just bad style. Could you try this with a loop? Something like:

```
expected = "\nok T:"
for expected_char in expected:
    mychar = serial.read()
    if mychar != expected_char:
        break
else:
    successful_responses += 1
    ...
```

(This code was written from the top of my head without testing. Please test first.)

Ghostkeeper:
Developers, see issue CURA-6538.

plugins/USBPrinting/AutoDetectBaudJob.py
goldfingyr: Not "bad style" at all - standard readable optimized interpreter code. Not so readable but even more optimized would be using &&.
Anyway had some time to clean up a lot of mess and *really* bad code aka zombie code or dead-beef.
Now the serial parameters are actually set, buffer is cleared before detection and the boot log from printer included in the log. (one of mine was missing an initial filament parameter)
Cheers
Karsten

goldfingyr:
Made a lot of changes due to the quite messy code in this file.
Its tested and ok.
One soft spot though:
I think its possible (but I haven't seen it) that the ASCII an/decode may throw an exception if a weird byte from a bad baud-rate is detected (illegal unicode) ?!?
In such case we might need a try-catch construction. Lets see what gives

nallath:
Sorry it took so long for us to check it, but unfortunately, this change doesn't actually work.

When testing the orginal code for an UM2, the baud rate gets correctly detected. When testing it with the proposed changes, no baud rate is detected, because the responses it got could not be decoded. 
I got the following error: 
2019-11-27 13:06:12,902 - ERROR - [Thread-3] UM.Logger.logException [84]: Exception: Job <USBPrinting.AutoDetectBaudJob.AutoDetectBaudJob object at 0x7f8304be3eb8> caused an exception
2019-11-27 13:06:12,903 - ERROR - [Thread-3] UM.Logger.logException [88]: Traceback (most recent call last):
2019-11-27 13:06:12,903 - ERROR - [Thread-3] UM.Logger.logException [88]:   File "/home/jaime/Development/Uranium/UM/JobQueue.py", line 121, in run
2019-11-27 13:06:12,904 - ERROR - [Thread-3] UM.Logger.logException [88]:     job.run()
2019-11-27 13:06:12,904 - ERROR - [Thread-3] UM.Logger.logException [88]:   File "/home/jaime/Development/Cura/cura/../plugins/USBPrinting/../USBPrinting/AutoDetectBaudJob.py", line 73, in run
2019-11-27 13:06:12,904 - ERROR - [Thread-3] UM.Logger.logException [88]:     readBack = serial.read(10000).decode("ASCII")
2019-11-27 13:06:12,904 - ERROR - [Thread-3] UM.Logger.logException [88]: UnicodeDecodeError: 'ascii' codec can't decode byte 0xe1 in position 0: ordinal not in range(128)

PR #5775, opened 2019-05-23
HMS434 update

maukcc:
HMS434 multi tool update


Ghostkeeper:
Can you take a look at the merge conflict?

maukcc:
I am guessing it is complaining about value should be default_value.
Is there a reason to use default_value?
machine_gcode_flavor, machine_start_gcode and machine_end_gcode will not stick if you use value, but the others work just fine.
on the other hand gantry_height will not stick if you use default_value.
Maybe we could use some uniformity here.

maukcc:
Also: the buildplate raster seems to have disappeared over the weekend. For all machines.

Ghostkeeper:
> Is there a reason to use default_value?

Not really. It's really only there now to provide a default for CuraEngine which is not able to read the Python expressions. There is no reason for you to override `default_value` in a printer definition rather than to override `value`.

PR #5776, opened 2019-05-23
CL-1348 Enable cloud actions

ianpaschal:
Contributes to CL-1348


ChrisTerBeke:
Not sure if this is done yet, but simply enabling the UI elements won't make it work. The functions also need to be implemented in the CloudOutputDevice.

ianpaschal:
Are you sure? I didn't commit this without checking; I was able to delete and move to top as expected.

ChrisTerBeke:
I guess you were still connected locally then behind the scenes. When you left the print job action endpoints weren't even deployed in the cloud yet :)

diegopradogesto:
@ianpaschal , can we close this PR? The internal ticket says "Unresolved" but maybe it was changed to another internal reference.

ChrisTerBeke:
Working on a new PR, closing this one.

PR #5809, opened 2019-05-29
fix: import creawsome definitions

trouch:
This PR imports last CreawsomeMod definitions.
Machines have different ID than originally bundled ones to allow old users settings to be imported with old machine definition that doesn't use variants and materials.

I'll understand if you want me to change the category, but I highly recommend opening that up.
Ultimaker machines remain at the top, while providing extra and clearer categories instead of an endless `Other` one.

Edit: I also need to submit a PR onto fdm_materials repo


zebons:
Waouh That's great recognition for the work done !

stelgenhof:
Thanks for this! Good progress and the right steps :)

Ghostkeeper:
This pull request is... impressive! A lot of work must've gone into this.

With the redesign of the add printer wizard in Cura 4.1 we're now considering properly grouping 3rd-party printers under their respective categories as well. Previously this was blocked by Ultimaker's marketing department though as a way of promoting their own printers over others, with them being the sponsor. We'll see what comes out of that discussion...

I think instead of the different printer IDs we can create a version upgrade for existing Creality users, so that their preferred settings get transferred over to the new system. I can't work on that during office hours though, but maybe I can help out in my free time a bit.

Creality has a significant portion of Cura's user base so it'd be awesome (or "awsome" if you will) if these people could be using these improved quality profiles.

Ghostkeeper:
Developers, see issue CURA-6554.

stelgenhof:
@Ghostkeeper It's indeed impressive. All possible combinations of all Creality's printer models, nozzle size, material types and quality definitions are in there :)

trouch:
> This pull request is... impressive! A lot of work must've gone into this.
Thanks ! ~ 70 quality/variant/material combination multiply 15 prints = 1000+ files
But I'm cheating with a single set of combination that get copied for all machines.
It would be nice if machines could inherit quality/variant/material from another one.

> With the redesign of the add printer wizard in Cura 4.1 we're now considering properly grouping 3rd-party printers under their respective categories as well. Previously this was blocked by Ultimaker's marketing department though as a way of promoting their own printers over others, with them being the sponsor. We'll see what comes out of that discussion...
Can't wait to know more about it ! I fairly understand Ultimaker politic.

> I think instead of the different printer IDs we can create a version upgrade for existing Creality users, so that their preferred settings get transferred over to the new system. I can't work on that during office hours though, but maybe I can help out in my free time a bit.
I'm concerned by this. Importing a user profile from old machine definition into a creawsome one could bring weird behavior as some fdmprinter rules are overridden.
 
> Creality has a significant portion of Cura's user base so it'd be awesome (or "awsome" if you will) if these people could be using these improved quality profiles.
That will be be definitely helpful for new users to Cura !

Ghostkeeper:
> It would be nice if machines could inherit quality/variant/material from another one.

They can, actually! You need to put `"quality_definition": "some_other_definition_id",` in the metadata of the .def.json file. Here is an example:

https://github.com/Ultimaker/Cura/blob/55c7f546b22633e3c288a2868d70aa634f2f6d4e/resources/definitions/ultimaker3_extended.def.json#L8

It'd be very useful if this cuts the number of files down. I'm not too worried about it since Cura lazily reads profiles, but other engineers have expressed their concern about start-up time with these new profiles. We haven't tested yet.

> I'm concerned by this. Importing a user profile from old machine definition into a creawsome one could bring weird behavior as some fdmprinter rules are overridden.

I think though that in the long term it would be better if there is only one printer to choose from per printer model. Introducing multiple will cause legacy to build up, so this is (as far as I'm concerned) a requirement from the Cura developers. I'm willing to help though. I couldn't find time for it this weekend like I said previously, but I'm planning to make a pull request on your branch so that you can review whether you like that solution.
The transition won't be perfect since people will have made changes that don't work together well in combination with the new settings. But in our experience with previous printer changes such collisions are pretty rare and the people who have them tend to accept that they made the change and revert to the default.

A second requirement from Ultimaker is that the model files for the platform have to be reduced to 1MB each. That's our policy to keep the download size of Cura somewhat more manageable.

trouch:
> > It would be nice if machines could inherit quality/variant/material from another one.
> 
> They can, actually! You need to put `"quality_definition": "some_other_definition_id",` in the metadata of the .def.json file. Here is an example:
> 
> https://github.com/Ultimaker/Cura/blob/55c7f546b22633e3c288a2868d70aa634f2f6d4e/resources/definitions/ultimaker3_extended.def.json#L8

Seems to be working and I confirm startup time is shorter then. 

> A second requirement from Ultimaker is that the model files for the platform have to be reduced to 1MB each. That's our policy to keep the download size of Cura somewhat more manageable and to keep it performing somewhat better for people with less powerful computers. Some of your models have all sorts of panels and Z screws in them which might be removed at no significant cost (or just changed into a cylinder instead of a screw/spring). Sometimes I'm using Blender's decimate modifier on them but in your case I don't think that would be very effective.

That was expected. I need to check with the actual platform modeler who went crazy with with Ender and CR-10S.

trouch:
Additional testing - variants seems to be required per machine.

Regarding user profile import, I wonder if we could add some sanity check in https://github.com/Ultimaker/Cura/blob/master/plugins/CuraProfileReader/CuraProfileReader.py and add default quality/variant/material if any exist in the machine def but missing in the user profile ?

Not sure about the migration from `creawsome_` to `creality_` IDs within user settings then.

nallath:
Variants are still needed, but this at least cuts down on the amount of quality profiles needed (We also use it for the UM3 and the UM3E

trouch:
Should be better now, removed ~ 900 files as well as heavy platforms files for now.
There are some tuning still in progress on the base definition.

Ghostkeeper:
We're implementing a version upgrade for a different change now, which will share some code so it makes this easier.

trouch:
Hi @diegopradogesto @Ghostkeeper 
Any ETA for upcoming Cura release ?
This is just to know how much time I have to adjust the settings 😅
I'll probably release another mod version in the meantime.

nallath:
The current planning is week 29 for the beta and week 31 for the final release.

Ghostkeeper:
> The current planning is week 29 for the beta and week 31 for the final release.

Translation for people who don't tend to count weeks: Week 29 is July 15th to 21st. Week 31 is July 29th to August 4th. So in about a month.

PR #5816, opened 2019-05-30
fdmprinter.def.json: draft_shield_enable -> draft_shield_enabled

BarrensZeppelin:
This seems to have been typos.


Ghostkeeper:
Indeed! I wonder why Python hasn't complained to us about that. Good catch.

Issue #5770, opened 2019-05-21
Dual extruder, single heater, mixing nozzle gcode order causes failure.


andymozza:
**Application Version**
4.0

**Platform**
Windows 10 and others

**Printer**
M3D Promega 
Dual extruder stepper, single mixing nozzle, single heater

**Steps to Reproduce**
When slicing for dual tools, Cura turns un-used tools or switched tools to 0 or the standby temperature, however it does this 'after' the temperature for the new tool is set.
Because this is the same shared heater, then the tool starts heading towards 0.... filament gets stuck and doesn't extrude etc.
M104 S200
G1 F600 E-2
T2
M109 S210
M104 T0 S0  <-- Set T0 to o degrees which affects the T2 which has already been switched to

**Actual Results**
Filament will no longer extrude because it's been cooled too much as the tool is switched off.

**Expected results**
M104 to turn the tool off should be sent 'before' the temperature for the new tool is set.
- very simple fix, just switch the order of the temp gcode.

M104 T0 S0 <-- Switch off un-used tool first, then work on the used tool
M104 S200
G1 F600 E-2
T2
M109 S210

**Additional Information**
I realise there are other issues related to this which have been rejected as you don't have these tools or heads etc, but I think this simple change can be visually validated and will not affect others, but be of benefit to these mixing nozzles with single shared heater.
Selection of gcode from switch between brim printed with extruder 0 and then main object printed with extruder 2 as an example.

andymozza:
You could actually put it ‘just’ before the tool switch ( T2 ) and then new temp is set after - to make sure any retraction isn’t adversely affected... however pretty sure the timing is so close it makes little difference.

smartavionics:
This problem with dual extruder shared heaters has been discussed before but the Cura developers have no interest in solving the issue as Ultimaker printers have a heater per nozzle. I am happy to work on a solution but I do not have any way of testing my code as I do not have such a printer. If you are willing to test using my Cura builds then I can help.

andymozza:
Thanks, For this change I believe it can be done simply without check boxes or UI changes ( although this is of course an option).
I still have multiple 'tools' configured to do different colour ratios, so we still need to tool switches, just need to make sure that the old tool is set to 0 or standby temp 'before' the temp on the new tool is set, so that the new temperature overrides the old temperature ( as a quick fix ).
Obviously there are more detailed changes you could do as well if you are so inclined :-)

Yes, I am willing to test yes builds.

[...]

PR #5840, opened 2019-06-03
Add support for printers that have multiple extruders that share a single heater

smartavionics:
This PR adds the machine_extruders_share_heater setting and an associated checkbox in the printer settings dialog. Defaults to false.

See #5770.

### DIFF
--- a/plugins/MachineSettingsAction/MachineSettingsPrinterTab.qml
+++ b/plugins/MachineSettingsAction/MachineSettingsPrinterTab.qml
@@ -314,6 +314,18 @@ Item
                     onGlobalContainerChanged: extruderCountModel.update()
                 }
             }
+
+            Cura.SimpleCheckBox  // "Shared Heater"
+            {
+                id: sharedHeaterCheckBox
+                containerStackId: machineStackId
+                settingKey: "machine_extruders_share_heater"
+                settingStoreIndex: propertyStoreIndex
+                labelText: catalog.i18nc("@label", "Shared Heater")
+                labelFont: base.labelFont
+                labelWidth: base.labelWidth
+                forceUpdateOnChangeFunction: forceUpdateFunction
+            }
         }
     }
 
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -375,6 +375,16 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_extruders_share_heater":
+                {
+                    "label": "Extruders Share Heater",
+                    "description": "Whether the extruders share a single heater rather than each extruder having its own heater.",
+                    "type": "bool",
+                    "default_value": false,
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": false,
+                    "settable_per_meshgroup": false
+                },
                 "machine_disallowed_areas":
                 {
                     "label": "Disallowed Areas",
@@ -2113,7 +2123,7 @@
                     "minimum_value": "-273.15",
                     "minimum_value_warning": "material_standby_temperature",
                     "maximum_value_warning": "material_print_temperature",
-                    "enabled": "machine_nozzle_temp_enabled",
+                    "enabled": "machine_nozzle_temp_enabled and not machine_extruders_share_heater",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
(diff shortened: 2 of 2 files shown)


smartavionics:
Associated engine PR is https://github.com/Ultimaker/CuraEngine/pull/1063.

fieldOfView:
I think the single checkbox is a bit inflexible. I can already see the requests/complaints about people who have three extruder motors, two of which share a single heater/nozzle and the other is used for a different material with its own hotend. Could this somehow be made into an extruder-option (grouping extruders) instead of a printer-option?

smartavionics:
Hi Aldo. The functionality provided is what was requested. I don't use a printer with a shared heater myself. Until the people with those printers put a case for making the  functionality more complicated, I am not going to invent stuff that may never actually get used. If people use this simple implementation and decide that it needs to be made more capable then we can have that discussion in the future.

Ghostkeeper:
Do we know of any printer that has 3 extruders but 2 heaters?

nallath:
Not that I'm aware of. But I do remember that there is a 3 extruder 1 heater printer out there.

andymozza:
This fulfils requests so far for all shared heater problems I have read through over the years on the this git. This change has been used so far in a test build with GEETECH A10M, M3D dual extruder ( 3 mix ratio tools defined) and CR10s ( the latter having only single extruder but used to check it didn't break things).

Ghostkeeper:
Developers, see issue CURA-6579.

LisaErl:
Sorry I'm so late describing my concerns and I apologize in advance if I do anything wrong, I'm brand new to Github. I just signed up for this comment.

In the closed Issue  #5770 smartavionics explained that a possibly necessary temperature change should take place shortly before the toolchange, so that at the time of the toolchanges the target temperature is nearly reached.

In my opinion, this should be reconsidered for the following reason: Immediately after the toolchange, the new extruder starts to work, but the hotend still contains the filament fed by the old extruder. This filament is deposited in the PurgeBucket or on the PrimeTower during the Purge phase and actually still needs the "old" temperature. Particularly when changing to a lower temperature, too early cooling of the hotend to the new temperature could cause problems in the form of clogging.

Personally, I am convinced that 2:1 Hotends are only suitable for pure MultiColor without temperature differences, but there are users who actually want to print materials with different temperatures.

Apart from that, as the owner of an A10M with 2:1 Hotend, I'm really looking forward to the modification. Many, many thanks for that.

smartavionics:
Hi @LisaErl , many thanks for your contribution to this issue. As I do not use a printer that has a shared heater I can only base my implementation on what people tell me. So far, the testers have been happy. See https://github.com/smartavionics/Cura/issues/8 for more discussion on this.

 If you wish to try one of my builds that has this feature (Linux or Windows only), please go to https://www.dropbox.com/sh/s43vqzmi4d2bqe2/AAADdYdSu9iwcKa0Knqgurm4a?dl=0.

LisaErl:
Hi @smartavionics ,
If it works well, then forget what I wrote. I just sat yesterday at a Cura postprocessing script with which I wanted to remove the obsolete temperature commands and therefore I have such things right in my head.

P.S. Especially because you don't have such a printer, I think it's really great of you to make such a change for us. Once again a big thank you. I know a lot of people from the groups around the A10M/A20M who will be very grateful to you.

smartavionics:
If the early heat up / cool down does cause a problem when people use materials with greatly different print temps on a shared heater machine (sounds crazy to me) then I will consider adding a setting to disable the early temp change. However, it may then cause problems because when moving from a high temp material to a low temp material, the extruder has to cool and I am not sure that M108 R? is supported by all firmwares so waiting for the extruder to cool to the right temperature may not be possible.

LisaErl:
Yes, I think it's just a theoretical problem. Whoever tries to process materials with a significant temperature difference with a 2:1 Hotend will always get problems, no matter what the G-Code looks like.

andymozza:
Indeed - we thought it through and this solution makes sense for most scenarios already reported to Cura dev. Massive temp differences even if you think you have purged the nozzle fully always come back to haunt you with partial clogs etc. I just don’t think it makes sense to use this setup if your filament is so different. The ramp up and down works perfectly for those small temp differences to optimise printing for filaments in roughly the same range.

Bomberski:
Im a a10m user, and wish you could support 2 in 1 feature.

PR #5845, opened 2019-06-04
Post processing plugin for percentage flow for the support structures

jonnygraham:
Post processing plugin that adjusts the percentage flow for the support parts of the print.

This would ideally be a feature of the engine itself, but this post-processing plugin adjusts the flow rate after the gcode is already generated.

I saw somebody else also wanted a similar feature here:
https://community.ultimaker.com/topic/24618-feature-request-1-flow-rate-control-on-support-structures/

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/AdjustSupportFlow.py
+++ b/plugins/PostProcessingPlugin/scripts/AdjustSupportFlow.py
@@ -1,0 +1,96 @@
+# Created by Jonny Graham
+
+from ..Script import Script
+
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This PR sort of gets overridden by this PR: https://github.com/Ultimaker/Cura/pull/4561

I'm not sure if we really want this temporary solution.

jonnygraham:
That's fine by me. It's definitely hacky to provide this as a
post-processing step.

PR #5855, opened 2019-06-05
First_emotionTech_Strateo3D_pull

KOUBeMT:
Hello,
Adding the all new eMotionTech's professional 3D printer Strateo3D
Thanks


resources/definitions/strateo3d.def.json
```diff
@@ -0,0 +1,136 @@
+{
+	"version": 2,
```
nallath: Your indentation is a bit off at some locations. It's not a big issue, but it helps to keep it readable.

KOUBeMT: Weird, i have no prob with indentation on my pc. I use sublime-text for editing and my indentation settings are set to tabs. should i set it to space maybe?

nallath: Yeah, we use 4 spaces for indentation.

KOUBeMT: just fixed it... thanx

Ghostkeeper:
Developers, see issue CURA-6570.

pkuiper-ultimaker:
Hi,

I checked your code and I have a few change requests.

- Can you upgrade:  Cura\plugins\PostProcessingPlugin\scripts\PauseAtHeight.py script instead of supplying a new post processing script Pause_emt.py? The function seem very similar (except for the multiple layer options) and we like to keep the number of scripts small.
- Can you upgrade the the setting_version numbers of the quality and variant types to 8 (setting_version = 8).
Kind regards,

         Paul Kuiper

plugins/PostProcessingPlugin/scripts/Pause_emt.py
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }"""
+
+    def execute(self, data: list):
+
+        """data is a list. Each index contains a layer"""
+        layer_nums = self.getSettingValueByKey("layer_number")
+
+        pause = "M600"
```
diegopradogesto: Looking at this code, it looks to me that we have a post-processing plugin that does exactly that (look at https://github.com/Ultimaker/Cura/blob/934248e218c65bc8c4135a465f1c8db37fbd985a/plugins/PostProcessingPlugin/scripts/FilamentChange.py#L76)

Could you explain why do you need this extra post-processing plugin?

KOUBeMT: Hello Diego.
One of your colleague suggested that we upgrade the existing PauseAtHeight plugin script, one of our engineers is working on it... in a couple of hours i'll commit it. 
thanks

KOUBeMT: Hello Diego,
We've been discussing the matter internally and indeed there is no much difference between filament_change and pause_emt. 
So we decided to remove / delete our script (see commit d444b40) and use the existing one...
thanks for your help

KOUBeMT:
Hello there,
I was expecting to see our printer in the new release of Cura 4.2, but it's not the case....
Can you give me an idea of when we can expect that to happen...
Thanks , have a good day!

KOUBeMT:
@Ghostkeeper 
Just installed 4.2.1, and S3D is in it... thanks guys

Ghostkeeper:
Yes, I'm sorry. You made the change to our master branch. The 4.2 branch was already branched off at that point, so it wasn't included in 4.2.

The master branch will be used to create 4.3 from, so the new printer will appear in Cura 4.3, released in around 2 months.

PR #5863, opened 2019-06-06
Don't prime-blob when print-sequence is one-at-a-time. [CURA-6414]

rburema:


resources/definitions/fdmprinter.def.json
```diff
@@ -4622,11 +4622,12 @@
...
                     "description": "Whether to prime the filament with a blob before printing. Turning this setting on will ensure that the extruder will have material ready at the nozzle before printing. Printing Brim or Skirt can act like priming too, in which case turning this setting off saves some time.",
                     "type": "bool",
-                    "resolve": "any(extruderValues('prime_blob_enable'))",
+                    "resolve": "any(extruderValues('prime_blob_enable')) and resolveOrValue('print_sequence') != 'one_at_a_time'",
                     "default_value": false,
                     "settable_per_mesh": false,
                     "settable_per_extruder": true,
-                    "enabled": false
+                    "enabled": false,
+                    "warning_value": "True if resolveOrValue('print_sequence') == 'one_at_a_time' else None"
```
Ghostkeeper: Is there a reason why this should evaluate to `None` when there needs to be no warning, rather than to `False`? If it can evaluate to `False` we can simplify this to:

    "warning_value": "resolveOrValue('print_sequence') == 'one_at_a_time'"

Ghostkeeper: I tried with the simplified formula and the behaviour seemed to be the same. I think we can simplify this.

Ghostkeeper:
The reset button currently always causes the setting to go into warning state. Clicking the checkbox works fine and also repairs the warning state if you reset the setting previously, but the reset arrow of the Enable Prime Blob setting makes it look like this:

Screenshot from 2019-06-12 11-42-12

resources/definitions/fdmprinter.def.json
rburema: I think it's _meant_ to represent a three-state value though: a. warn on true/enabled, b. warn on false/disabled and c. don't warn. If the behaviour doesn't change to reflect that, it proably explains the 'other' :-) bug that you found.

Ghostkeeper: Ah, yes that makes sense.

PR #5864, opened 2019-06-06
Mesh processing rank

BagelOrb:
Changed description of Infill Mesh order to Mesh Processing Order because it is now also used for normal meshes.

See https://github.com/Ultimaker/CuraEngine/pull/1069


Ghostkeeper:
Seems we forgot this when we merged the sister-PR in CuraEngine. It makes sense.

PR #5885, opened 2019-06-11
Add deprime settings

Ghostkeeper:
This adds a number of settings required for the deprime procedure in order to make better long retractions.

These settings are currently not used by Cura but only by the Ultimaker S5 firmware. It's here because people need to be able to change these settings in the material profiles (that get sent to the firmware).

Contributes to issue CURA-6329.


resources/definitions/fdmprinter.def.json
```diff
@@ -2234,6 +2234,107 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "material_crystallinity":
+                {
+                    "label": "Crystalline Material",
+                    "description": "Is this material the type that breaks off cleanly when heated, or is it the type that produces long intertwined polymer chains?",
```
rburema: In my opinion: I don't think this question makes it clear (enough) which one of the options is enabled/true and which one disabled/false. I'd assume the first is enabled/true, but I'd be uncertain ... though, on the other hand, of course everyone that's actually supposed to use this should already know what crystalinity is.

PR #5904, opened 2019-06-16
Create creality_ender5.def.json

PDBeal:
Every release I copy this file into my resource folder for the Ender-5.  Figured I should add it to the repo for future updates.

### DIFF
--- a/resources/definitions/creality_ender5.def.json
+++ b/resources/definitions/creality_ender5.def.json
@@ -1,0 +1,94 @@
+{
+    "name": "Creality Ender-5",
+    "version": 2,
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Michael Wildermuth",
+        "manufacturer": "Creality3D",
+        "file_formats": "text/x-gcode",
+        "preferred_quality_type": "draft",
+        "machine_extruder_trains":
+        {
+            "0": "creality_cr10_extruder_0"
+        }
+    },
+    "overrides": {
+        "machine_width": {
+            "default_value": 220
+        },
+        "machine_height": {
+            "default_value": 300
+        },
+        "machine_depth": {
+            "default_value": 220        },
+        "machine_head_polygon": {
+            "default_value": [
+                [-30, 34],
+                [-30, -32],
+                [30, -32],
+                [30, 34]
+            ]
+        },
+        "layer_height_0": {
+            "default_value": 0.2
+        },
+        "top_bottom_thickness": {
+            "default_value": 0.6
+        },
+        "top_bottom_pattern": {
+            "default_value": "concentric"
+        },
+        "infill_pattern": {
+            "value": "'triangles'"
+        },
+        "retraction_enable": {
+            "default_value": true
+        },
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is about to collide with the Creawsome merge: https://github.com/Ultimaker/Cura/pull/5809

Ghostkeeper:
Perhaps you can see what you can contribute to that pull request over at https://github.com/Ultimaker/Cura/pull/5809?

PDBeal:
This won’t collide with any of his changes. Besides that pull request looks like his own setup for some of these machines, not the true Creality setup that matches what is already included in Cura.

If it’s not wanted, that’s fine, moving forward I’ll continue to add it to my Cura installations. I’d rather use the Creality provided definition that someone’s custom definitions.

I didn’t write this definition, it was downloaded from Creality as part of the software package for the Ender-5.

Ghostkeeper:
The reason of how I know it's more battle-tested is because at the release of version 4.1 we were getting thousands of crash reports from that mod because it caused Cura to crash upon start-up.

If it doesn't collide with those changes, maybe you could rebase it on his branch?

trouch:
@PDBeal existing Creality machines definition has been community provided AFAIK
However I can't find original source except `Michael Wildermuth` as the original CR-10 definition author as well as your Ender-5 one. I have no clue who is he, but he doesn't seem to be a Creality employee.

Having said that, I can only just suggest to give a try to my Creawsome machine definition. And let me know how it goes. I'm willing to improve and adjust for Ender-5 eventually !

https://github.com/trouch/CreawsomeMod

PDBeal:
The file was part of the download package from Creality for the Ender-5.  They have links for the SD card images that is at https://www.creality3d.cn/download/sd-card-file_c0005.  Within the Ender-5 download link you'll find the Cura definition 
https://www.dropbox.com/s/1d0ygffu7i13yg1/Ender-5.rar?dl=0&file_subpath=%2FEnder-5+%E9%9A%8F%E6%9C%BA%E8%AF%B4%E6%98%8E%EF%BC%88%E8%8B%B1%E6%96%87%EF%BC%89%E5%BA%95%E7%A8%BF20181007%2F2.Software+and+USB+Driver%2FCURA%2FCura3.4.1

Granted the definition was from Cura 3.4.1, but thats the original source for the json.  Perhaps he does not work for Creality, or perhaps he does.  All I know is the definition was part of the Ender-5 SD image from Creality.

I personally do not use user created mods because I don't trust them.  If it's not supplier provided, I'll create my own.  There are too many well documented cases of usermods containing malicious code.  in this case it may not be true, but as a general rule of thumb, I don't have any inclination to even trial a usermod.

trouch:
I confirm CreawsomeMod has malicious configuration that makes prints better, but nothing else 😅

nallath:
We do actually check what is merged into Cura. It's by no means a guarantee that there is no malicious code, but i think the quality of the files as provided by @trouch is orders of magnitude better than the old profiles.

The file you provided is also not a full definition. If you want the original profile in, that can be discussed, but we need all the required files in order to do so.

PR #5905, opened 2019-06-16
Update imade3d jellyboxes' printer settings

filipgoc:
This pull request updates support for IMADE3D JellyBOX printers. It adds JellyBOX 2 printer and revises settings of JellyBOX Original.  This request corresponds to a similarly named one in the fdm_materials repository.


Ghostkeeper:
(Developers, see ticket CURA-6586.)

diegopradogesto:
There are conflicts, probably because the setting_version has changed to value 8 instead of 7. Could you please fix the conflicts?

filipgoc:
@diegopradogesto I appreciate your help. I have pushed a commit changing the settings_version to 8, but the conflicts are still here. Is there a way for me to inspect the problem?

When I look at the diff (https://patch-diff.githubusercontent.com/raw/Ultimaker/Cura/pull/5905.diff), it looks fine.

If I had the rights, I would simply overwrite the existing quality files - they are all previously submitted by me. Accept all incoming changes, leave nothing old in place. I'm not sure how to resolve the conflicts otherwise. I'm open to all suggestions. Thank you.

filipgoc:
@diegopradogesto Never mind. I went ahead and just merged it locally with the latest code (I pulled today as `master-2019-30-17`). It may be a bit of an overkill, but hopefully it will work now! Please let me know if there are any issues.

Vandrasc:
The PR is now merged to master, I do get some config errors when upgrading from 4.2 with the old definitions for the original type: 

config

nallath:
Is the Imade3D_2 just the same thing as the variant with the fans? Because if that's the case, we probably need to revert this.

filipgoc:
@nallath @Vandrasc 

It's not the same. 
JellyBOX 2 has only one variant as of now.

JellyBOX Original used to have two variants
- 1 fan
- 2 fans

I simplified the profiles, which meant that the 1 and 2 fan variant of JellyBOX Original were identical, and thus unnecessarily complex. Only one was necessary, so didn't want to have two for no reason. 

However, maybe it cannot be done without breaking old profiles? Is the only way keeping it in even if it's superfluous? Maybe the simplification is not worth the breaking of old user profiles.

Pffff. That will be a bit of a mess to redo. If this is the way you feel is better (I trust your Cura knowledge), then let's revert this one and I'll have to make a new PR?

nallath:
I've fixed a part of the issue. Basicly, if you rename the file name of profiles, they won't survive an upgrade. So for the 1-fan variant, I've renamed the profiles back again so that at least those work.
I guess it would be less trouble to just put the 2-fan one back again (and just leave the profiles as they are).

nallath:
See https://github.com/Ultimaker/Cura/pull/6046

filipgoc:
> See #6046

Thanks so much. This is a great start. I'll dig into it this week and try to make it consistent, with the upgrade procedure in mind.

nallath:
The version upgrade should solve everything. The only issue we couldn't think of a solution for is for the variant to another definition upgrade.

PR #5912, opened 2019-06-19
CS-134: Print job actions via cloud

ChrisTerBeke:
This adds support for print job actions over the cloud. Note that some actions can take a while to reflect in the UI because the data stream has to travel all the way to the printer and back. We can improve the UX for this in the future by adding loading states or something.


ChrisTerBeke:
@ianpaschal FYI

plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -96,6 +96,21 @@ def requestPrint(self, cluster_id: str, job_id: str, on_finished: Callable[[Clou
         reply = self._manager.post(self._createEmptyRequest(url), b"")
         self._addCallback(reply, on_finished, CloudPrintResponse)
 
+    ##  Send a print job action to the cluster for the given print job.
+    #  \param cluster_id: The ID of the cluster.
+    #  \param cluster_job_id: The ID of the print job within the cluster.
+    #  \param action: The name of the action to execute.
+    def doPrintJobAction(self, cluster_id: str, cluster_job_id: str, action: str, data: Dict[str, Any] = None) -> None:
```
nallath:
```suggestion
    def doPrintJobAction(self, cluster_id: str, cluster_job_id: str, action: str, data: Optional[Dict[str, Any]] = None) -> None:
```

plugins/UM3NetworkPrinting/src/ClusterUM3OutputDevice.py
```diff
@@ -140,6 +140,11 @@ def _spawnPrinterSelectionDialog(self):
         if self._printer_selection_dialog is not None:
             self._printer_selection_dialog.show()
 
+    ##  Whether the printer that this output device represents supports print job actions via the local network.
+    @pyqtProperty(bool, constant=True)
+    def supportsPrintJobActions(self) -> bool:
```
nallath: Shouldn't this be moved to a parent class and set to true in case of this specific one?

ChrisTerBeke: It's only needed in the network plugin so far, and within that plugin the output devices do not share a common class. If Cura team things this should be added to some base output device class in Cura, go ahead.

nallath: Oh, I assumed that they did share a common ancestor. Nevermind then.

PR #5928, opened 2019-06-23
Add Colorblind Accessibility Themes

nubnubbud:
One of my friends (who happens to be colorblind) and I shared a word about Cura's color choices for layer view. Nothing else presented much of an issue, but most of the line type colors ended up becoming very similar to people with colorblindness, most notably the yellow/green being close.

Ideally, this would be changed with some kind of program-wide post-processing, but because most issues were relegated to the line type view, that's where most of the changes are, and post-processing can increase complexity in a way that's unnecessary. 

I used a simulator to simulate very strong colorblindness, and this theme should make it easier for someone with any kind of colorblindness to use the line type view. (except the rarest kind, tritanopia, which works fine with both the normal colors and decently with the colorblind accessibility theme.)

I also changed the overhang color to a bright magenta in these themes, which will cause it to appear the opposite of the default material color, making it more usable with deuteranopia and even moreso for protanopia.

there is both a light and dark colorblind assist, and they both inherit from ultimaker light, though that can be changed easily. It's a quick change, but because cura is for both new printers and experienced engineers, it's good to have basic accessibility options actually packaged with the releases.

Untitledredoverhangs
dark
---------------------------------------------------------------------------------------------------------------
Untitledassistoverhangs
assistdark


Ghostkeeper:
This has been on our wish list for quite a while. Seems like the correct solution to the problem too. I'm discussing it with a teammate but it looks good so far.

nubnubbud:
> This has been on our wish list for quite a while. Seems like the correct solution to the problem too. I'm discussing it with a teammate but it looks good so far.

thanks for taking a look at it! I'm open to making changes to more than just the layer view, but I just didn't notice anything else that needed it. if you see something where yellow and green share a space or are used in conjunction, that needs to be changed for a colorblind mode, as both common types of colorblindness will see it as the same rough color, as seen in that second image.

Ghostkeeper:
You think the red/green colours of the X/Y components of the axis labels are clear enough? Like the red/green/blue lines in the origin of the printer volume?

Ghostkeeper:
Developers, see issue CURA-6622.

nubnubbud:
> You think the red/green colours of the X/Y components of the axis labels are clear enough? Like the red/green/blue lines in the origin of the printer volume?

They're definitely workable, with one being blue, one being light, and one being dark in each kind of colorblindness, though it could be better. I'll make a quick change to lower the brightness of the red a little further, and for the rotation and translate tools too.

pattonwebz:
This would be a hugely welcomed feature. Thank you @nubnubbud for working in this :)

nallath:
Sorry this took so long! I've also taken the liberty to greatly simplify the theme. Since we use inheritance, you only need to re-define the values that you actually want to have different from the original (this makes maintaining it a whole lot easier!). You can see my changes at 450302071f54aff5b3fe31a86fb8b6f5e1f9ed76

PR #5935, opened 2019-06-24
feat: "Z-Seam Position" presets to ease XY setting

trouch:
Add a dropdown with position presets to quickly set Z-Seam X/Y when using `z_seam_type='back'`, eg. **User Specified** alignement.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1399,6 +1399,27 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "z_seam_position":
+                {
+                    "label": "Z Seam Position",
+                    "description": "The position near where to start printing each part in a layer.",
+                    "type": "enum",
+                    "options":
+                    {
+                        "backleft": "Back Left",
+                        "back": "Back",
+                        "backright": "Back Right",
+                        "right": "Right",
+                        "frontright": "Front Right",
+                        "front": "Front",
+                        "frontleft": "Front Left",
+                        "left": "Left"
+                    },
+                    "enabled": "z_seam_type == 'back'",
+                    "default_value": "back",
+                    "limit_to_extruder": "wall_0_extruder_nr",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Very good idea. Nice small ease of use thing.

resources/definitions/fdmprinter.def.json
```diff
@@ -1399,14 +1399,35 @@
...
+                        "front": "Front",
+                        "frontleft": "Front Left",
+                        "left": "Left"
+                    },
+                    "enabled": "z_seam_type == 'back'",
+                    "default_value": "back",
+                    "limit_to_extruder": "wall_0_extruder_nr",
+                    "settable_per_mesh": true
+                },
                 "z_seam_x":
```
Ghostkeeper: We have the convention that CuraEngine slices with only the leaf settings. In this case, Z Seam X and Z Seam Y are the ones that CuraEngine slices with, so the X and Y settings should be child settings of the new Z Seam Position.

This is consistent with other such settings such as Wall Line Count vs. Wall Thickness.

resources/definitions/fdmprinter.def.json
```diff
@@ -1399,14 +1399,35 @@
...
+                        "backleft": "Back Left",
+                        "back": "Back",
+                        "backright": "Back Right",
+                        "right": "Right",
+                        "frontright": "Front Right",
+                        "front": "Front",
+                        "frontleft": "Front Left",
+                        "left": "Left"
+                    },
+                    "enabled": "z_seam_type == 'back'",
```
Ghostkeeper: Shouldn't this say something like `z_seam_type == 'user'`?

Ghostkeeper:
Developers, see issue CURA-6621.

resources/definitions/fdmprinter.def.json
trouch: Updated with https://github.com/Ultimaker/Cura/pull/5935/commits/a9540c1ff04bd98e1a8e74cefab9f0f0bddf41b0

PR #5953, opened 2019-06-30
Save "light mode"

lorduncan:
Each time i do a print i screenshot and save a "picture" of the settings so time later i know what i used for each print.  I tried to use "save" command but it generate so huge files. Would be great a way to save settings used in a print even as a txt file. Sounds useful right? Thanks for perfect Cura <3


fieldOfView:
@lorduncan: This should not be a pullrequest. You open a pullrequest is when you have created or changed code to be added to Cura. This seems to be a suggestion. Please open an Issue instead.

Ghostkeeper:
I'm closing this PR as to not accidentally merge it. That would be disastrous...

Ghostkeeper:
You can save the current set-up with the "Save..." command. But this saves the model as well, which is why you get such large files. So first clear the build plate (Ctrl+D) then use "Save...", and the result should be a ~70kB file.

Ghostkeeper:
**DEVELOPERS, DO NOT DELETE THE BRANCH `feature_intent`, LIKE GITHUB SUGGESTS BELOW!**

lorduncan:
> 
> 
> You can save the current set-up with the "Save..." command. But this saves the model as well, which is why you get such large files. So first clear the build plate (Ctrl+D) then use "Save...", and the result should be a ~70kB file.

That´s a good option , but i would like to know also other things, like if the location or rotation of the model was changed (like a Z -10 ) or resize it to 80%  or if there is any "pause at layer" commands or what printer i used with what noozle size,... things like that.
Other improve could be the posibility that you could upload that file to places like thingiverse so anyone could check all the settings you used to print that "i make one".
Much appreciated your help. Thanks. And one more time sorry, is my first time at gitHub.
I share an example what i use to screen shot.
V5 con pausa al 1mm para cambio a blanco

PR #5961, opened 2019-07-02
Import CreawsomeMod profiles and implement version upgrade

Ghostkeeper:
This is an extension of this PR: https://github.com/Ultimaker/Cura/pull/5809 . Merging this PR will also merge that one. Github allows me to make a PR to Trouch's branch, but since my Master was much farther ahead than Trouch's, this PR would show hundreds of commits that he didn't have when the branch split off. Theoretically I could update the remote by pushing to Trouch's repository but Github was giving me trouble on that with access rights to `git@github.com:trouch/Cura:fix/creawsomemod` for some reason. Maybe I was using it wrong... So here is a separate PR.

This PR does the following things:
* Remove the old definitions for Creality printers except Creality CR-X (which was not present in CreawsomeMod).
* Replace them with the CreawsomeMod definitions, variants, quality profiles.
* CreawsomeMod profiles have been renamed to just "Creality". There can only be one definition for each printer model.
* Version upgrade is implemented so that users from vanilla Cura 4.1 would upgrade seamlessly to Cura 4.2. Their settings and custom profiles are retained. The nozzle is filled in as 0.4. The quality type snaps to the nearest possible quality type in CreawsomeMod.
* Version upgrade is implemented so that users from CreawsomeMod would upgrade seamlessly to Cura 4.2.

@trouch, could you take a look if you find this acceptable? To me this is fine and ready to be merged.

The category is still kept to "other". I'll bring up the topic of categorising by manufacturer again, tomorrow.

Implements CURA-6554.


trouch:
Thanks a lot @Ghostkeeper I'll have a look later today and update the base definition from your branch at the same time.

trouch:
Overall looks good 👍 
I wanted to push an update of the base def, but same for me got an ACL error 😅 
Could you pull https://github.com/trouch/Cura/tree/trouch-fix/creawsomemod and push it back here ?

Also renamed "Creality Base Machine" to "Creawsome Custom Machine" and made it visible to ease non Creality owner to use my settings. Seem to get pretty good result with other brands of printers !

I might need to test the update script eventually.

Ghostkeeper:
> Also renamed "Creality Base Machine" to "Creawsome Custom Machine" and made it visible to ease non Creality owner to use my settings. Seem to get pretty good result with other brands of printers !

We don't want to publish everyone's favourite settings as separate printers. The printers are meant to be printers, not profiles. So I don't really like having that printer visible. If we want other printers to use it, we should update the defaults in fdmprinter.

trouch:
Thanks a lot @Ghostkeeper, @LipuFei, @nallath 

With regards to the "Custom Creawsome Machine" visibility, I didn't want to change fdmprinter straight ahead which could impact ALL other printers. But I could submit another PR after 4.2 if you'd like.

Also maybe I missed the ability to import a full definition from a curaprofile ZIP file, and not only quality ?
I really need math formulas such as below, this is so convenient: 
```
"support_angle": {
     "value": "math.floor(math.degrees(math.atan(line_width/2.0/layer_height)))"
},
```

Having the ability to import definition/variants/quality would be great. Maybe that comes down to the _Machine/Material Marketplace_ ?

Ghostkeeper:
No indeed, it is currently not possible to import definition files. All the rest can be imported through project files (File -> Save...), but not definition files.

We have vague plans to extend the marketplace to be able to do this. Not gonna be for a while though.

PR #5963, opened 2019-07-03
Categorise printers by manufacturer in Add Printer wizard

Ghostkeeper:
This should make it easier to find your printer among the long list of printers that we have now.

Implements CURA-6635.


fieldOfView:
Wohoo!

PR #5966, opened 2019-07-03
Listen to 'material-changed' since it doesn't update immediately.

rburema:
part of CURA-6590


resources/qml/Preferences/Materials/MaterialsView.qml
```diff
@@ -117,8 +117,8 @@ TabView
 
                     onNo:
                     {
-                        properties.diameter = old_diameter_value;
-                        diameterSpinBox.value = properties.diameter;
+                        base.properties.diameter = old_diameter_value;
+                        diameterSpinBox.value = Qt.binding(function() { return base.properties.diameter })
```
Ghostkeeper: I figured that Qt would have some way to re-bind properties. Good find!

rburema: Ha, what a coincidence, somebody just asked this in the course as well :-)
( ... and it's one of those 'forbidden knowledge' things that 'you should probably not use' / I'm OK with using it here though, since we're going to refactor this anyway )

LipuFei: I found it in Qt documentation: https://doc.qt.io/qt-5/qtqml-syntax-propertybinding.html#creating-property-bindings-from-javascript

PR #5992, opened 2019-07-09
Hide 'move to top' option from actions

ChrisTerBeke:
Due to an issue with Qt list models the order of the queued print jobs is not the actual order of them in the print queue. This can be solved by adding a custom sort to the list model, but we currently have no data to base this sorting on as the cluster API does not return the order of a print job in the queue (and fields like UUID or creation date are not appropriate for sorting either). For now we have decided to just hide the 'move to top' option that is somewhat useless at the moment (you will still be able to use it via the cloud or connect web interfaces).

### DIFF
--- a/plugins/MonitorStage/MonitorMain.qml
+++ b/plugins/MonitorStage/MonitorMain.qml
@@ -97,7 +97,7 @@ Rectangle
                 horizontalCenter: parent.horizontalCenter
             }
             visible: isNetworkConfigured && !isConnected
-            text: catalog.i18nc("@info", "Please make sure your printer has a connection:\n- Check if the printer is turned on.\n- Check if the printer is connected to the network.")
+            text: catalog.i18nc("@info", "Please make sure your printer has a connection:\n- Check if the printer is turned on.\n- Check if the printer is connected to the network.\n- Check if you are signed in to discover cloud-connected printers.")
             font: UM.Theme.getFont("medium")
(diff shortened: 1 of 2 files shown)


sedwards2009:
Cura Connect does report the print queue to the cloud in the correct order. I don't know where it is getting mixed. (Cura/Qt itself?)

ChrisTerBeke:
It's the Qt list models. They sort by themselves, unless you provide a specific sorting algorithm (which we, as explained, cannot create based upon relevant information at the moment).

Ghostkeeper:
Qt's ListModels don't sort by themselves. They [don't have a sorting key](https://doc.qt.io/qt-5/qml-qtqml-models-listmodel-members.html). They basically just work as an array: You can insert objects and remove them.

Uranium provides an extension on that class that has a `sort` convenience function, to which you'd need to provide a function pointer that is the sorting key. You don't have to call that though.

The way that Cura usually works with ListModels is to extend UM.ListModel and provide a custom `_update` override that fills the list. Here's a simple example: [NozzleModel.py](https://github.com/Ultimaker/Cura/blob/master/cura/Machines/Models/NozzleModel.py). In your case, you'd want that `_update` override to put the items in the list in the same order as how you get them from the printer's API.

ChrisTerBeke:
Thanks for the tip, I'll investigate.

PR #5993, opened 2019-07-09
Create an intent model per intent category

Ghostkeeper:
This gives the intent model a property for which intent category it should list the intents. And then it lists those intents.

Implements CURA-6597.


fieldOfView:
Is this PR is intense intentionally?

Ghostkeeper:
No, the intention is to intenerate the intensive interface PR that will follow.

PR #5998, opened 2019-07-10
Added Vertex Nano K8600 profile

velbn:


Ghostkeeper:
Developers, see issue CURA-6660.

resources/definitions/vertex_nano_k8600.def.json
```diff
@@ -0,0 +1,83 @@
+{
+    "version": 1,
```
Ghostkeeper: To be compatible with Cura 2.3 and higher, this version number needs to be `2`. Did you test this in Cura 4.1? The entire printer doesn't get added at all.

resources/extruders/vertex_nano_k8600_extruder_0.def.json
```diff
@@ -0,0 +1,16 @@
+{
+    "id": "vertex_nano_k8600_extruder_0",
```
Ghostkeeper: There is no "id" field. Please remove it.

resources/extruders/vertex_nano_k8600_extruder_0.def.json
```diff
@@ -0,0 +1,16 @@
...
+    "version": 1,
```
Ghostkeeper: This version number needs to be 2 as well in order to get the profile to load in Cura 2.3 or higher.

resources/extruders/vertex_nano_k8600_extruder_0.def.json
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "vertex_nano_k8600",
+        "position": "0"
+    },
+
+    "overrides": {
+        "extruder_nr": { "default_value": 0 },
+        "machine_nozzle_size": { "default_value": 0.35 },
```
Ghostkeeper: Is the nozzle size actually 0.35mm or did you intend to have a line width of 0.35mm? Not a problem here really but I'm asking in case you misunderstood this.

resources/definitions/vertex_nano_k8600.def.json
velbn: Thanks for the comments,
Is there a reference manual to be found so I won't make the same mistake twice?

resources/extruders/vertex_nano_k8600_extruder_0.def.json
velbn: Nozzle is actually 0.35

velbn:
I've pushed a new commit

Vandrasc:
The PR is now merged to master, just want to mention that I noticed that the PLA and PETG are doubled displayed in the material list. The duplicates seem to be unique though.

PR #5999, opened 2019-07-11
CURA-6659 Use VersionUpgradeManager to update the imported quality

LipuFei:
CURA-6659

### DIFF
--- a/plugins/CuraProfileReader/CuraProfileReader.py
+++ b/plugins/CuraProfileReader/CuraProfileReader.py
@@ -67,9 +67,10 @@ class CuraProfileReader(ProfileReader):
             return []
 
         version = int(parser["general"]["version"])
+        setting_version = int(parser["metadata"].get("setting_version", 0))
         if InstanceContainer.Version != version:
             name = parser["general"]["name"]
-            return self._upgradeProfileVersion(serialized, name, version)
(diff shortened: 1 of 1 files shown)


plugins/CuraProfileReader/CuraProfileReader.py
```diff
@@ -98,19 +99,23 @@ def _loadProfile(self, serialized: str, profile_id: str) -> Optional[InstanceCon
...
 
-        if not profile_convert_funcs:
-            Logger.log("w", "Unable to find an upgrade path for the profile [%s]", profile_id)
+        parser = configparser.ConfigParser(interpolation = None)
+        parser.read_string(serialized)
+        if "general" not in parser:
+            Logger.log("w", "Missing required section 'general'.")
+            return []
+        if "version" not in parser["general"]:
+            Logger.log("w", "Missing required 'version' property")
```
Ghostkeeper: Instead of re-parsing the entire file, `updateFilesData` also just returns the version number you're trying to read. You can check it with:
```suggestion
        new_source_version = results.version
        if int(new_source_version / 1000000) != InstanceContainer.Version or new_source_version % 1000000 != CuraApplication.SettingVersion:
            Logger.log("e", "Failed to upgrade profile [%s]", profile_id)
```

(Note that this code is not tested. Please test first.)

PR #6004, opened 2019-07-12
Add camera_view as stats

LipuFei:
@diegopradogesto

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -126,16 +126,7 @@ class SliceInfo(QObject, Extension):
             else:
                 data["active_mode"] = "custom"
 
-            camera_view = application.getPreferences().getValue("general/camera_perspective_mode")
-            from UM.Scene.Camera import Camera
-            if camera_view == Camera.PerspectiveMode.PERSPECTIVE.value:
-                camera_view_string = "perspective"
-            elif camera_view == Camera.PerspectiveMode.ORTHOGONAL.value:
-                camera_view_string = "orthographic"
-            else:
-                Logger.log("w", "Unexpected perspective mode [%s], use 'perspective' as the default value.", camera_view)
-                camera_view_string = "perspective"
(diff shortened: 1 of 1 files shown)


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -126,6 +126,17 @@ def _onWriteStarted(self, output_device):
             else:
                 data["active_mode"] = "custom"
 
+            camera_view = application.getPreferences().getValue("general/camera_perspective_mode")
+            from UM.Scene.Camera import Camera
```
Ghostkeeper: Please import these at the top of the file.

plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -126,6 +126,17 @@ def _onWriteStarted(self, output_device):
...
+            if camera_view == Camera.PerspectiveMode.PERSPECTIVE.value:
+                camera_view_string = "perspective"
+            elif camera_view == Camera.PerspectiveMode.ORTHOGONAL.value:
+                camera_view_string = "orthographic"
+            else:
+                Logger.log("w", "Unexpected perspective mode [%s], use 'perspective' as the default value.", camera_view)
+                camera_view_string = "perspective"
+            data["camera_view"] = camera_view_string
```
Ghostkeeper: You can simplify all this by simply taking the value from the preferences, which is already exactly the string you're saving:

```suggestion
            data["camera_view"] = application.getPreferences().getValue("general/camera_perspective_mode")
```

I don't suggest checking for "expected" values, because that would mean that a modification to the set of valid values needs to happen in two places, and we'll certainly forget to update the one in here. If you disagree, you could at least change it to check for presence in a set, like so:

```suggestion
            if camera_view not in {"perspective", "orthographic"}:
                Logger.log("w", "Unexpected perspective mode {mode}, sending 'perspective' as the default value.".format(mode = camera_view))
                camera_view = "perspective"
            data["camera_view"] = camera_view
```

PR #6011, opened 2019-07-15
Add data slot for peripherals

Ghostkeeper:
We're not showing peripherals yet in the interface right now, but this might be available in the future.

Implements CURA-6651.


cura/PrinterOutput/Peripheral.py
```diff
@@ -0,0 +1,17 @@
...
+#   peripherals. This set is then possibly shown in the interface of the monitor
+#   stage.
+class Peripheral:
+    ##  Constructs the peripheral.
+    #   \param id A unique ID for the peripheral object, like a MAC address or
+    #   some hardware ID.
+    #   \param type A unique ID for the type of peripheral.
+    #   \param name A human-readable name for the peripheral.
+    def __init__(self, type: str, name: str):
+        self.type = type
```
nallath: Shadows built-in type(), so I guess it's better to rename this.

PR #6018, opened 2019-07-16
Ender-2: Increase build plate to match physical size

probonopd:
Tested on my Ender-2


nallath:
Are you sure about this change? The official stats list a build volume of 150x150x200

Ghostkeeper:
For reference: https://www.creality3d.cn/ender-2-p00093p1.html

Keep in mind that there is always some variability in the assembly. I'd rather be on the safe side.

probonopd:
Positive, there is 1 cm more, 16x16.
I‘d like to see another owner to confirm.

trouch:
Honestly, this kind of issue is always 50/50.
Going up to 160x160, then the other 50% people will complain about not using standard size.
It is safe to change it yourself in the machine settings if you'd like.

JohnEdwa:
It's the same thing with all Creality printers, officially the Ender 2 is 150x150 and the Ender 3 is 220x220 , but the actual build plates are physically larger at 165x165 and 235x235.
[Here's the size sheet from Whambam for reference](https://cdn.shopify.com/s/files/1/0016/6179/6450/products/14_sizes_of_FBS_029415f0-da56-45fa-a80f-3245858fcc1f_1024x1024@2x.png?v=1561959371) - Ender 2 is at the bottom right.


However, AFAIK the stock firmware has them set up as 150 and 220 and they probably won't let you print over those sizes, so unless someone can confirm they can, Cura shouldn't change the default sizes.

probonopd:
I definitely print with 160x160 all the time (but I am using a stick Marlin firmware rather than what came with the printer). Would be good if someone with stock Ender-2 firmware could confirm that 160x160 prints work fine indeed.

diegopradogesto:
The Creality Ender-2 specifications says it's 150x150x200, so we are keeping it like this. It's the safer side. We also think it should work with the by-default firmware version, that probably also indicates the same dimensions.

Of course you could also extend a little bit more the buildplate of a Ultimaker, but still the specifications are what the manufacturer indicates and states in the firmware.

I'm closing this PR, but thanks anyway for the contribution.

PR #6028, opened 2019-07-19
Preserve extruder-only moves in post stretch script

sgtnoodle:
The post stretch script currently treats extruder-only moves (retraction and priming) as normal extrusion moves. As the script stretches x-y coordinates around, this can result in slight x-y motions while retracting and priming.  Although seemingly benign, these slight x-y motions confuse certain printer firmwares to the point of ruining the print job.

This change detects extruder-only moves during gcode parsing and preserves the extruder-only motion in the output.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/Stretch.py
+++ b/plugins/PostProcessingPlugin/scripts/Stretch.py
@@ -128,9 +128,26 @@ class Stretcher():
                     onestep = GCodeStep(0, in_relative_movement)
                     onestep.copyPosFrom(current)
                 elif _getValue(line, "G") == 1:
+                    last_x = current.step_x
+                    last_y = current.step_y
+                    last_z = current.step_z
+                    last_e = current.step_e
                     current.readStep(line)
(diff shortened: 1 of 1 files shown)


plugins/PostProcessingPlugin/scripts/Stretch.py
```diff
@@ -128,9 +128,26 @@ def execute(self, data):
...
-                    onestep = GCodeStep(1, in_relative_movement)
-                    onestep.copyPosFrom(current)
+                    if (current.step_x == last_x and current.step_y == last_y and
+                        current.step_z == last_z and current.step_e != last_e
+                    ):
+                        # It's an extruder only move. Preserve it rather than process it as an
+                        # extruded move. Otherwise, the stretched output might contain slight
+                        # motion in X and Y in addition to E. This can cause problems with
+                        # firmwares that implement pressure advance.
+                        onestep = GCodeStep(-1, in_relative_movement)
```
Ghostkeeper: Accidentally snuck in a minus there.

nallath:
Thanks for your contribution!

PR #6042, opened 2019-07-22
Hide incorrect tooltip

ChrisTerBeke:
The preferred solution was to change the copywriting, but since we are after string freeze we cannot do this. This is the next best solution to prevent user confusion.


ChrisTerBeke:
The CI fails on a codestyle issue that is not related to this PR. Somehow that snuck into the main code base :(

nallath:
The mypy of the CI was updated, so now it found a bunch of extra issues. You can ignore them for the moment being.

Issue #5207, opened 2019-01-22
Add configuration option to specify serial port connection.


uchobby:
**Application Version**
3.6.0

**Platform**
Win 10

**Printer**
Creality CR10

**Steps to Reproduce**
Run Cura then try to open any serial port COMx

**Actual Results**
Access error indicating that the COMx port is already open

**Expected results**
Cura should not open and thereby kill all serial devices on a PC

**Additional Information**
CURA holds serial ports so they can not be accessed by other applications
The previous ticket is similar https://github.com/Ultimaker/Cura/issues/4039.  Is unfathomable that that ticket was closed without resolution.  Just WOW!

USB connection handling proposal #1049
https://github.com/Ultimaker/Cura/issues/1049
A good solution to this significant bug in Cura.  If you can not see how to do this, you should start Cura with a big red bordered splash, "When this software runs, all serial ports will fail to open. We could fix this with a few minutes of coding work but can't be bothered to make things work well for other users."

nallath:
I understand this is frustrating, but we have limited time to fix things. As we're being paid by Ultimaker (and despite us giving this software way for free and open source, we're still a company), we focus on Ultimaker issues first and foremost. 

So if you want a fix, convince the Creatily people to start contributing back to Cura. They are very much welcome to provide a pull request to fix this. But I've made this appeal many times before to them, but I guess them not providing any software is the reason their machines are so cheap. You do get what you pay for I'm afraid. 

That being said, did leaving snarky comments ever help you in getting your issue resolved? It's not doing wonders with regards to motivating me to fix it on my own time (as that would be the only alternative at this point).

uchobby:
Jaime,

Thank you for the feedback.  Sorry for being snarky.  I was, as you say, frustrated and seeing that the issue had been reported twice with what seemed like a “who cares” response as rejection. It just rubbed me wrong.

I had lost a great deal of personal development time with another project due to the lockout on serial.  This was not a minor thing to me.  I was trying to troubleshoot several issues at the same time, and the serial port thing greatly aggravated the process.  It only dawned on me that Cura was the cause because I had made the same mistake long ago, with trying to make automatic serial port selection work in an application.  Cura was not connecting to my printer when it was available, due to some interaction with other software I had using serial ports, so I went on the hunt to find where I could force the selection, and it dawned that you must be doing automatic and did not provide a config option.

[...]

PR #6048, opened 2019-07-23
USB -COM open

3dmecanico:
When you open Cura it´s try to connect to printer If it´s succesefully Cura get port. This action have to problems:
First , some printer reset when connect.
Second if you need to use this port (ex. Pronterface) is busy.
It`s possibly to turn this auto connect to manually opción.
This option is necessary a long time ago.
Please sorry for mi bad english


Ghostkeeper:
I think you made a mistake with this pull request. It merges the 4.2 branch into Master (something that should be done any time we push to 4.2 anyway).

Did you intend to make an issue? That would be here: https://github.com/Ultimaker/Cura/issues/new/choose

However your issue is a duplicate of https://github.com/Ultimaker/Cura/issues/5207. Please continue discussion there.

PR #6054, opened 2019-07-24
Fix dimensions of print job information popup when using Qt 5.13.

smartavionics:
Remains compatible with Qt 5.10.

Tested using Qt 5.13 on Linux and Qt 5.10 on Windows.


Ghostkeeper:
Fixes it for Qt 5.12, where I was seeing the issue as well. Thanks!

PR #6070, opened 2019-07-28
Qt 5.13 compatibility - fix size of ExpandablePopup background.

smartavionics:
Without this, the view selector dropdown has no background.

Tested with Qt 5.13 on Linux and Qt 5.10 on Windows.


Ghostkeeper:
Yup, still works fine for older Qt versions.

PR #6074, opened 2019-07-29
Calculated Speed or Temperature for simple speed boost

UltimateSports:
Please move this to wherever it is supposed to be, 

we use a couple of printers for R&D and small batch production, when it comes to quick turn around trials we get a little backlogged, 
therefore I have been experimenting with speed/temp increases, but this can be a little hit and miss at times, 
would it be possible to have a fx that would meant that the temp is controled by speed changes or visa-versa?


nallath:
I don't think you intended to do this, but you created a pull request (instead of a feature request). Could you re-create the issue?

UltimateSports:
how does one do that? I can't see a tab for feature request either here, or on cura.

nallath:
Check https://github.com/Ultimaker/Cura/issues/new/choose

mbartlett21:
@nallath 
#6075 is the feature request by @UltimateSports

PR #6101, opened 2019-07-31
Create DisplayRemainingTimeOnLCD.py

iLyngklip:
Added a post-processing script that prints out the estimated time remaining generated by Cura.

It works by taking the estimated total time and the estimated print time from each layer and calculating the remaining time from that. It is as accurate as the estimated Total Time from Cura.


iLyngklip:
Maybe I should say, I have been using this script on all of my prints with my Ender-3 for the past month and it works flawlessly.

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+                    "label": "Enable",
+                    "description": "When enabled, It will write Time Left HHMMSS on the display",
+                    "type": "bool",
+                    "default_value": true
+                }
+            }
+        }"""
+
+    def execute(self, data):
+        if self.getSettingValueByKey("TurnOn"):
```
nallath: bit of nitpicking;

You can make this a bit easier to read by using the inverse. Eg; 
```python3
if not self.getSettingValueByKey("TurnOn"):
    return data
```
That way you don't have to ident the entire block that's below it.

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+            TotalTime = 0           # Var for total time
```
nallath: According to our codestyle, this should be total_time (as it's a variable, it should use _ in between words and be in lower cases). 
It's good that you put comments in your code, but this one is just repeating what's already there (when in doubt, try to write comments that explain why you do something. You should only add comments about the "what" if it is really difficult (or it's a trick or something).

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+            TotalTimeString = ""    # Var for the string we insert
```
nallath: Same as above; Should be total_time_string

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+            for layer in data:
+                layer_index = data.index(layer)
+                lines = layer.split("\n")
+                for line in lines:
+                    if line.startswith(";TIME:"):
+                        # At this point, we have found a line in the GCODE with ";TIME:"
```
nallath: +1 bonus points for adding comments like this.

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+                        # which is the indication of TotalTime. Looks like: ";TIME:1337", where
+                        # 1337 is the total print time in seconds.
+                        line_index = lines.index(line)      # We take a hold of that line
+                        minString = re.split(":", line)     # Then we split it, so we can get the number
```
nallath: Should be min_string. 

Also; you can do this a bit more robustly (and easier) by using:
```python3
total_time_string = line.partition(";Time:")[2]
total_time = int(total_time_string) 
```

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
```diff
@@ -0,0 +1,86 @@
...
+
+                        StringMedTal = "{}".format(minString[1])    # Here we insert that number from the
+                                                                    # list into a string.
+
+                        TotalTime = int(StringMedTal)               # Only to contert it to a int.
+
+                        m, s = divmod(TotalTime, 60)    # Math to calculate
```
nallath: We already have a nice convenience class that handles this for you (Duration).

You can use it like:
```python3
duration = Duration(total_time)
total_time_string = duration.getDisplayString()
```

nallath:
I've added some comments as to how you can improve the code (mostly some tips & tricks on how to also ensure it's easier to maintain)

iLyngklip:
Thank you so much for your comments! 

I have updated variable names according to your codestyle.

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
iLyngklip: I have tried this, but I could not get it to work. Since what is in the code now works, and should continue to work I think that's fine.

plugins/PostProcessingPlugin/scripts/DisplayRemainingTimeOnLCD.py
iLyngklip: Thanks for pointing that out, is was not aware of that. The names should be fixed now.

Ghostkeeper:
Developers, please refer to ticket CURA-6705.

iLyngklip:
I am unable to find the ticket with the number 6705, can you give a link? Thanks!

Ghostkeeper:
That's a reference to Ultimaker's internal ticketing system. I could give you a [link](https://jira.ultimaker.com/browse/CURA-6705), but you need to log in there because it also contains tickets for future Ultimaker products. It's just for developers to know where it is. We log public tickets in that system too to have an overview of the work that was completed in total during a sprint.

iLyngklip:
Ah okay, thanks :-)

nallath:
Sorry it took so long to merge this!

Issue #4114, opened 2018-07-23
Cura: Change Filament Script does not extract Filament


Sislar:
**Application Version**
3.4.0

**Platform**
Windows 10

**Printer**
Ender 3

**Steps to Reproduce**
Create any valid print,  do go to postprocessing and add "Change filament at layer"
Keep the defaults
   initial retraction  30mm
   retraction            300mm
save as "gcode"

**Actual Results**
The printer pauses for filament change at the correct spot but the filment moves forward (well
not really as it can't go forward that much) instead of backward

Gcode is inserted as follows
M600 E30.00 L300.00 ; Generated by FilamentChange plugin

**Expected results**
Filament should be retracted I edited the gcode and changed
M600 E30.00 L300.00 ; Generated by FilamentChange plugin
to
M600 E-30.00 L-300.00 ; Generated by FilamentChange plugin

With this change the filament retracted correctly.

diegopradogesto:
You're right, in Marlin firmware this is how it should be: http://marlinfw.org/docs/gcode/M600.html

Issue #4249, opened 2018-08-17
3.4.1 Filament Change post processing has backwards g-code for extrusion/retraction


boomercat:
**3.4.1**

**Ubuntu 16.04**

**MP Mini Select**

**Steps to Reproduce**

**Post Processing > Plugins > Filament Change (I've tried this for changing at different levels and it is always the same result.)**

**When the print gets to the point where it should pause and retract the filament: it actually pauses then extrudes filament. After the filament change, when I hit continue, it then retracts the filament. This causes me to have to manually feed the filament or my print will create a big hole. Also, the Z axis setting that it returns to is much too high because subsequent layers after the filament change have no "smush" and simply sit on top of the previous layers (with a noticable gap). This causes a massive split to form in the print.**

**I could take a video or something if somebody needed it.**

putridp:
I've also noticed this problem of the retract/extrusions being reversed - major clicking of the extruder then a retration as it starts the next layer.

Also, is it correct that the action takes place at the beginning of the specified payer? In contrast, the 'Pause at Height' post processing script, when specifying the layer, carries out the pause at the end of the layer. It would be good for the behaviours to be consistent, or at least specified in the settings

Ghostkeeper:
Cura's post-processing script doesn't actually determine the order in which these things happen. That's in your firmware. All we output is something like this:

    M600 E-30 L-300

That indicates that there should be a short retraction of 30mm before the switch, and a long retraction of 300mm to remove the material from the printer. Note that we do make the distance negative. Apparently Cura interprets the E and L parameters as a delta position for the extruder wheel, so a negative value would retract it.

It looks like Marlin, by default, doesn't care about what direction the retraction is going (notice the `ABS()` call):

https://github.com/MarlinFirmware/Marlin/blob/5cf6c44f47737bdf93df6ae7c835fe2d54d0675d/Marlin/Marlin_main.cpp#L10993

I don't know what heritage of firmware MP Mini is using or Putridp's printer, so maybe it's wrong for that firmware. Also the printers at Cura's development team are all Ultimaker printers that don't even support this command and completely ignore it, so we can't test for it.

I'll believe you that the distance should not be inverted. That also seems more logical to me. I'll remove the negative signs.

Ghostkeeper:
The Z position definitely is a bug in your printer's firmware. Cura doesn't specify anything with regards to the Z position.

PR #6139, opened 2019-08-03
Added minus sign to have negative retractation when pausing for change

prochat:
The generated gcode is faulty as described here:
https://github.com/Ultimaker/Cura/issues/4114

My proposal seems to solve this problem.


Ghostkeeper:
Ah, but that then contradicts the change that was made here: https://github.com/Ultimaker/Cura/issues/4249 based on the findings of @boomercat and @putridp. We've had this value change back and forth a couple of times now!

diegopradogesto:
I guess keeping it as positive would allow the user to enter both positive or negative numbers in the interface. So I would keep it as it is now.

multiholle:
If you add a negative number, the generated gcode is missing the E parameter. This should really be fixed.

Also is the description of the parameter misleading.

Ghostkeeper:
We're going to reject this. An explanation of why is written here: https://github.com/Ultimaker/Cura/issues/7502#issuecomment-617601243

PR #6153, opened 2019-08-05
Modify Cubicon device profile

HUNIBEST-HyVISION:
Modify Cubicon devices profile (Style Plus-A15, Dual Pro-A30)
- machine_nozzle_offset_x, y
- platform_offset


resources/definitions/cubicon_dual_pro_a30.def.json
```diff
@@ -0,0 +1,45 @@
+{
+    "id": "Dual Pro-A30",
```
Ghostkeeper: The "id" field doesn't exist. Please remove it.

resources/definitions/cubicon_style_plus_a15.def.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id": "Style Plus-A15",
```
Ghostkeeper: The "id" field doesn't exist. Please remove it.

HUNIBEST-HyVISION:
Remove 'id' field

PR #6155, opened 2019-08-05
CS-234: Network plugin improvements

ChrisTerBeke:
* Fix URLs for calling some endpoints (makes 'add by IP' more robust across platforms).
* Fix setting the device display name (shows the pretty name in the 'add by IP' dialog).
* Centralize code for showing upload messages (prevent duplication in translations).


ChrisTerBeke:
@Ellecross This should fix that issue you had today with the 'add by IP' feature (turns out it was somewhat platform specific because of an extra '/' in the URL).

PR #6165, opened 2019-08-06
Update_materialView_qml

Vandrasc:
This will allows us to run the automation test for ExportingMaterial. Contributes to Cura-6366.


resources/qml/Preferences/Materials/MaterialsView.qml
```diff
@@ -394,7 +394,7 @@ TabView
                 {
                     width: childrenRect.width
                     height: childrenRect.height
-                    text: model.description
+                    text: model.label
```
Ghostkeeper: This is not a good fix. Instead of the setting description, it now puts the setting name in the tooltip when you hover over a setting. This makes the tooltip useless since the setting name is already displayed in the tooltip's hover area. So when hovering over the "Retraction Distance" setting you get a tooltip that says "Retraction Distance" instead of "The length of material retracted during a retraction move."

A more proper fix would be to search the children of this element for Label elements with the text in it. That would not require any changes to the code under test for the purpose of the test, which is good practice. Or if that fails, we could add a property new property to this TooltipArea.

Vandrasc: It didn't make it useless, in my opinion, as we discovered the parent which contains two children, the label and spinbox, necessary for the getChildren method, so next step will be to the add a new property if .label is not accepted.  I will ask @SatishBiharie93 for his help and see what we can do. Thank you for your reply.

Ghostkeeper: It makes it useful for the GUI test yes, but useless to the user that wants to read the description in the final product.

PR #6240, opened 2019-08-22
Cura 6732 available configurations

nallath:
CURA-6732

### DIFF
--- a/cura/PrinterOutput/Models/PrinterOutputModel.py
+++ b/cura/PrinterOutput/Models/PrinterOutputModel.py
@@ -321,8 +321,9 @@ class PrinterOutputModel(QObject):
         return self._available_printer_configurations
 
     def addAvailableConfiguration(self, new_configuration: PrinterConfigurationModel) -> None:
-        self._available_printer_configurations.append(new_configuration)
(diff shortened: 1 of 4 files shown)


cura/PrinterOutput/Models/PrinterOutputModel.py
```diff
@@ -290,7 +293,7 @@ def canUpdateFirmware(self) -> bool:
     def _onControllerCanUpdateFirmwareChanged(self) -> None:
         self.canUpdateFirmwareChanged.emit()
 
-    # Returns the configuration (material, variant and buildplate) of the current printer
+    # Returns the active configuration (material, variant and buildplate) of the current printer
     @pyqtProperty(QObject, notify = configurationChanged)
     def printerConfiguration(self) -> Optional[PrinterConfigurationModel]:
         if self._printer_configuration.isValid():
```
LipuFei: Changing this property name will break the API, but I think we can change `self._printer_configuration` to `self._active_printer_configuration` so internally the naming will be more clear.

nallath: Fair enough.

Issue #6245, opened 2019-08-23
[master] New crash


smartavionics:
**Application version**

master

**Platform**

Linux using QT 5.13

**Printer**

Sorry, can't remember

**Reproduction steps**

Sorry, can't remember, will try to update if it occurs again.

**Actual results**

Cura crashed

**Expected results**

No crash!

**Additional information**

stderr and the log get a large number of the following error reports...

```
2019-08-22 21:56:47,130 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [61]: An uncaught error has occurred!
2019-08-22 21:56:47,133 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]: Traceback (most recent call last):
2019-08-22 21:56:47,134 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:   File "/opt/cura/lib/python3/dist-packages/cura/CuraApplication.py", line 990, in event
2019-08-22 21:56:47,134 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:     return super().event(event)
2019-08-22 21:56:47,134 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:   File "/opt/cura/lib/python3/dist-packages/UM/Qt/QtApplication.py", line 428, in event
2019-08-22 21:56:47,135 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:     event._function_event.call()
2019-08-22 21:56:47,135 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:   File "/opt/cura/lib/python3/dist-packages/UM/Event.py", line 189, in call
2019-08-22 21:56:47,135 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:     self._function(*self._args, **self._kwargs)
2019-08-22 21:56:47,136 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:   File "/opt/cura/lib/python3/dist-packages/UM/Signal.py", line 315, in __performEmit
2019-08-22 21:56:47,136 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:     func(dest, *args, **kwargs)
2019-08-22 21:56:47,136 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [64]:   File "/opt/cura/lib/python3/dist-packages/UM/Backend/Backend.py", line 191, in _onMessageReceived
[...]

PR #6246, opened 2019-08-23
Set _slicing before assigning to gcode_dict to avoid potential race.

smartavionics:
Asynchronously, _the entry in gcode_dict could be cleared if _slicing is false.

Maybe fixes https://github.com/Ultimaker/Cura/issues/6245.


Ghostkeeper:
This doesn't fix a potential race condition. It could still be executing in the `_onErrorCheckFinished` past the check for `if not self._slicing`, then switch threads to the `slice()` function and set both `self._slicing` to True and clear the build plate, then switch back to the `_onErrorCheckFinished` thread and call `needsSlicing` which clears the g-code from the scene and causes the next g-code layer message to crash the application.

smartavionics:
Understood. I'll close this PR then and leave it to you guys to come up with a fix.

PR #6248, opened 2019-08-24
Parallelize Stretch post processing script

sgtnoodle:
I'm still testing it for correctness, but initial results show that the Stretch script scales very nicely with CPU cores. I went from 62 seconds to 14 seconds on my 4 core desktop with my test model, so that's pretty neat.

I'd appreciate a quick review pass before I spend too much more time on this, in case folk disagree with the idea of parallelizing post processing scripts, or have a better idea how to approach it.


sgtnoodle:
The link to that failed jenkins run doesn't work for me. I successfully printed a benchy using the modified Stretch script though.

Ghostkeeper:
I don't know why the link doesn't work for you. Maybe this link works better?

    http://hal9000.ultimaker.com:29130/job/Ultimaker/job/Cura/job/PR-6248/1/display/redirect

It's failing on the UM3NetworkConnection plug-in. That must be something happening on the master branch when you branched off.

Ghostkeeper:
Developers, see issue CURA-6929.

plugins/PostProcessingPlugin/Script.py
```diff
@@ -22,6 +22,24 @@
...
+# This module is importable, so it's possible to provide a workaround. The trick is to set up functions to call
+# before multiprocessing forks the main process. That way, each forked process has a copy of the processing script.
+# For example:
+#         pool = multiprocessing.Pool(processes=multiprocessing.cpu_count(), initializer=multiprocessing_init,
+#                                     initargs=({"process_layer": self.process_layer},))
+#        result = pool.apply_async(multiprocessing_call, ("process_layer", (layer_steps,))))
+#        return result.get()
+def multiprocessing_init(functions_by_name):
+    """ An inititializer for multiprocessing pools. The passed in keyword arguments are callable using the
+        multiprocessing_call function below """
```
Ghostkeeper: Please put function and class documentation in Doxygen-style comments above the function as well. For now at least.

You can use `\param` and such to denote parameters.

plugins/PostProcessingPlugin/Script.py
```diff
@@ -22,6 +22,24 @@
...
+# Some post-processing tasks are easy to parallelize. The easiest way to parallelize python code is via multiprocessing.
+# Post-processing scripts aren't directly importable, though,  making them incompatible with multiprocessing.
+# This module is importable, so it's possible to provide a workaround. The trick is to set up functions to call
+# before multiprocessing forks the main process. That way, each forked process has a copy of the processing script.
+# For example:
+#         pool = multiprocessing.Pool(processes=multiprocessing.cpu_count(), initializer=multiprocessing_init,
+#                                     initargs=({"process_layer": self.process_layer},))
+#        result = pool.apply_async(multiprocessing_call, ("process_layer", (layer_steps,))))
+#        return result.get()
+def multiprocessing_init(functions_by_name):
```
Ghostkeeper: Is this a `List[str]`? It'd help if you could add some typing here.

plugins/PostProcessingPlugin/scripts/Stretch.py
```diff
@@ -83,6 +86,24 @@ def copyPosFrom(self, step):
     def setInRelativeMovement(self, value: bool) -> None:
         self.in_relative_movement = value
 
+class GCodeStepSlot:
+    """
+    GCodeStep was originally used both during input parsing and during processing and output writing. In order to
+    work with multiprocessing pools, the step objects need to be manually pickled (classes in plugins aren't
+    directly pickleable). Rather than reconstruct GCodeSteps again, this class is used during processing. The hope
+    is that, by decoupling the two, future improvements will be easier.
+    """
```
Ghostkeeper: Please put the documentation in Doxygen-style comments above the class, for now.

Ghostkeeper: This documentation is about the change. It's irrelevant for future readers.

plugins/PostProcessingPlugin/scripts/Stretch.py
```diff
@@ -83,6 +86,24 @@ def copyPosFrom(self, step):
     def setInRelativeMovement(self, value: bool) -> None:
         self.in_relative_movement = value
 
+class GCodeStepSlot:
```
Ghostkeeper: Perhaps we can just use a `namedtuple` for this instead?

rburema:
> I'm still testing it for correctness,

Have you concluded anything from your tests? An other words: Is the PR (minus the requested changes) correct according to you at least?

sgtnoodle:
Yep, it appears to be correct. I've printed a variety of parts
successfully. I may be able to find some time this week to address the
feedback.

plugins/PostProcessingPlugin/Script.py
sgtnoodle: It's a dictionary of callable functions. The keys are nominally strings, but I suppose any hash-able type would work. I'm not all that familiar with typing in python 3, but I tried to make the doxygen comment clear.

plugins/PostProcessingPlugin/scripts/Stretch.py
sgtnoodle: I'll tweak it to hopefully be more invariant.

plugins/PostProcessingPlugin/scripts/Stretch.py
sgtnoodle: Good idea!

sgtnoodle:
I just updated my branch to address the feedback. I haven't tested it yet, though.

rburema:
I'm not sure what is going on exactly, but on the execution of the script (on 'save to file'), it starts up a second instance of Cura for some reason.

Also, the gcode produced is different when compared to master, at least that's what I saw. (But not to different runs of the same model on the same branch.)

sgtnoodle:
@rburema I'm guessing you tested it on windows? I've been running it in linux. I suspect that the multiprocessing pool is forking the parent process on linux, but re-importing all the modules in windows (and one of the modules creates the GUI on import). Unfortunately my approach to get around the inability to directly import the plugin relies on forking, so it might not work at all on windows. Would you mind trying it again and sending me the unprocessed gcode and the processed gcode?

rburema:
Hi @sgtnoodle 

> I'm guessing you tested it on windows?

Good guess :-)

>  ... so it might not work at all on windows.

That's unfortunate as a very large percentage of our users are on Windows...

> Would you mind trying it again ...

Sorry, it still spawns a second Cura instance.

> ... and sending me the unprocessed gcode and the processed gcode?

Here's the processed gcode and unprocessed gcode.
(Though the gcode at least is not much different in this case from the 'old' behaviour ... this is either the latest changes or because I'm not using the exact reproduction route.)

[UM3_spikes.zip](https://github.com/Ultimaker/Cura/files/4008587/UM3_spikes.zip)

ninovanhooff:
Can this PR be closed? Seems that the lack of Windows support is not easily fixed.

PR #6254, opened 2019-08-25
add CR-10 Max and Ender-5 Plus definitions

trouch:
CR-10 Max uses same print head as CR-10S Pro, but has bigger build volume.
Ender-5 Plus uses same print head as Ender-5, with bigger volume too.

Nothing fancy here, just inherit base Creality definitions.

### DIFF
--- a/resources/definitions/creality_cr10max.def.json
+++ b/resources/definitions/creality_cr10max.def.json
@@ -1,0 +1,32 @@
+{
+    "name": "Creality CR-10 Max",
+    "version": 2,
+    "inherits": "creality_base",
+    "overrides": {
+        "machine_name": { "default_value": "Creality CR-10 Max" },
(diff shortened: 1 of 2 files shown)


resources/definitions/creality_cr10max.def.json
```diff
@@ -0,0 +1,32 @@
...
+                [38, -34],
+                [38, 34]
+            ]
+        },
+
+        "gantry_height": { "value": 30 }
+
+    },
+    "metadata": {
+        "quality_definition": "creality_base"
```
LipuFei: This printer is not visible. You need to add another field here: `"visible": true`

resources/definitions/creality_ender5plus.def.json
```diff
@@ -0,0 +1,34 @@
...
+            ]
+        },
+
+        "gantry_height": { "value": 25 },
+
+        "speed_print": { "value": 80.0 }
+
+    },
+    "metadata": {
+        "quality_definition": "creality_base"
```
LipuFei: The same visibility issue. This printer is not visible. You need to add another field here: `"visible": true`

trouch:
Thanks @LipuFei for the feedback.
Should be better now.

PR #6263, opened 2019-08-27
CURA-6537 Add UI feedback on invalid firmware update

LipuFei:


Ghostkeeper:
Looks like this belongs to CURA-6537.

PR #6274, opened 2019-08-28
Fix materials API call

LipuFei:
@ChrisTerBeke


ChrisTerBeke:
Actually this didn't fix it. Made a follow-up PR: https://github.com/Ultimaker/Cura/pull/6276.

PR #6276, opened 2019-08-28
Network plugin material and configuration sync improvements

ChrisTerBeke:
* Fixes showing available configurations when one of the print cores has no materials loaded to it.
* Fixes sending the material profiles to the Connect API.
* Shows a new message when syncing materials so the user knows what is going on.

Note: I have removed the material syncing tests because a) they were hard to update, and b) they were testing many internal components.


Ghostkeeper:
Was this reviewed by anyone? It's introduced a few mistakes with material syncing. In particular, it now seems to sync multiple times per connection and gives a warning message when the printer is unable to change its material file (even though that is an expected outcome).

ChrisTerBeke:
In what lines did the intended behavior change?

Ghostkeeper:
The intended behaviour never changed as far as I know. The actual behaviour did.

As far as lines of code, it's hard to say which lines of code caused the materials to be synced multiple times per connection. As for the warnings for expected failures to sync materials, that'd be the `Logger.log("w", "Error while syncing material: %s", reply.errorString())` warnings that were added in [this commit](https://github.com/Ultimaker/Cura/pull/6276/commits/5ee6cd7e8fcc282bf41404926a609c7d892e9081) by the looks of it.

The issue should be raised by @Ellecross at the moment towards CS. They found it to be "infinitely sending materials" but it was later not found to be infinite, just a few minutes. I was just looking through notifications here and wondered if this was ever code reviewed then.

ChrisTerBeke:
We're looking into it right now.

PR #6278, opened 2019-08-29
Add plug-in to read PLY files

Ghostkeeper:
Cura can natively read PLY files now.

There is a change that needs to be made to cura-build as well in order to allow the Windows installer to open these files as well. I'll make that change promptly.

Contributes to issue CURA-6739.

### DIFF
--- a/cura.desktop.in
+++ b/cura.desktop.in
@@ -13,6 +13,6 @@ TryExec=@CMAKE_INSTALL_FULL_BINDIR@/cura
 Icon=cura-icon
 Terminal=false
 Type=Application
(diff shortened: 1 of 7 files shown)


fieldOfView:
Yay for using Trimesh to do this!

At some point it would be nice to move _toMeshData into something that can be used by multiple plugins.

Ghostkeeper:
Yeah, true. I'm working now on some of the other file formats as well and each time it needs this _toMeshData function...

fieldOfView:
Perhaps it could be worth it to rename this plugin to TrimeshReader.py and also let it handle other relevant formats such as Collada (.dae).

Ghostkeeper:
Yeah, I think I'll do that.

PR #6281, opened 2019-08-29
Avoid crash due to race in CuraEngineBackend.

smartavionics:
Similar fix to those in https://github.com/Ultimaker/Cura/commit/23f4aa6e4f6d02b0342fb727cbed2991f3ea16d1


Ghostkeeper:
Good catch, thank you! I'm still unable to reproduce your crash, but in theory this should fix it and I tested that it doesn't break anything.

PR #6291, opened 2019-09-02
Update printrbot_simple_extended.def.json

samirabaza:
manufacturer name changed from "PrintrBot" to "Printrbot" to allow the profile under the "Add a printer" menu
not appearing in version 4.2.1


Ghostkeeper:
Ooh that's pretty bad. Merged into 4.3 because the change was simple. I'll make sure that we look into other printers having the same problem. Maybe we can make a more general fix.

PR #6292, opened 2019-09-02
Update printrbot_simple.def.json

samirabaza:
manufacturer name changed from "PrintrBot" to "Printrbot" to allow the profile under the "Add a printer" menu
not appearing in version 4.2.1


diegopradogesto:
I'm adding this to 4.3 since it's a small change that could affect people not finding their printers under the "Add printer" dialog.

BTW, good catch @samirabaza

PR #6297, opened 2019-09-03
Fixed resume code so height returns to starting level

bkp23:
PauseAtHeight.py was resuming 1mm higher than it paused at (line 312).

I modified code to resume at the same level it paused at. Also added additional comments to generated code.


Ghostkeeper:
I see you retracted the PR. Didn't it work?

Ghostkeeper:
Ah nevermind. I see that it got replaced by https://github.com/Ultimaker/Cura/pull/6298.

Issue #5924, opened 2019-06-21
[4.0.0] Pause at Height plugin restarts too high


bluegizmo83:
**Application version**
4.0.0

**Platform**
Windows 10 X64

**Printer**
Ender 3 Pro

**Reproduction steps**
Enable the Pause at Height plugin, select Layer to pause at (12 in my case), leave redo layers at 0, slice and print.

**Actual results**
The printer pauses at the layer as expected (at the end of layer 12 / start of layer 13), print head raises and moves to the set position, and upon restarting the Z height is up too high when it starts printing the first layer after pause. It looks to be about one or two layers too high. successive layers are ok after that though, just that first layer after the pause is too high, almost like it skips a layer or two when it goes back to printing.

**Expected results**
Should start printing at the correct layer height when resumed.

**Additional information**

pkuiper-ultimaker:
I tested the LayerAtHeight script on Cura 4.1 with an Ultimaker 2 (has Marlin firmware like your machine)

It produced this G code ( added some comments in code to explain):

;TYPE:CUSTOM
;added code by post processing
;script: PauseAtHeight.py
;current layer: 12
M83
G1 F300 Z2.92                                   # The head move 1mm above its current z position
G1 F9000 X190 Y190                         # Head moves to the side
G1 F300 Z15                                      # Below 15mm the head is moved even higer to 15mm
M0;Do the actual pause                   # Pause the machine
G1 F300 Z2.92                                  # Continue after pause, move 1mm above last layer position
G1 F9000 X107.164 Y107.164          # Move back to old XY position
G1 F9000
M82
G92 E272.4471
;LAYER:12                                        # Next layer starts
;MESH:___AA-Cubet.stl
G0 X107.164 Y107.164 Z2.07          # Head moves to correct z height 
                                                        # (1,92mm old z pos. + 0,15mm  layer height = 2,07 new z pos.)

The script seem to work OK. Are you sure that the printer is printing a FULL layer at the incorrect height or is you printer oozing material at the return position (which is 1mm above the last z position) before it starts printing layer 13 (which bring it back to the correct z height)?

Please check the G code and see if the error is in the G code or in the printer (Oozing).

Ghostkeeper:
The bug reproduces for me. Here's a snippet of g-code that I get:

```
G0 F600 X145.882 Y154 Z2.6 ;<<<<<last height: 2.6mm
G0 F7200 X145.759 Y145.759
;TIME_ELAPSED:117.972096
[...]

PR #6298, opened 2019-09-03
Fixed PauseAtHeight.py to resume at starting height

bkp23:
See issue #5924 

PauseAtHeight.py was resuming 1mm higher than it paused at (line 312). If the original script doesn't happen to move to the correct Z height before resuming the layer, the entire layer will be printed 1mm too high.

I modified code to resume at the same level it paused at _after_ the X-Y jog completes (to avoid hitting any prints). Also added additional comments to generated code.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
+++ b/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
@@ -265,7 +265,7 @@ class PauseAtHeight(Script):
 
                 if not is_griffin:
                     # Retraction
-                    prepend_gcode += self.putValue(M = 83) + "\n"
+                    prepend_gcode += self.putValue(M = 83) + " ; switch to relative E values for any needed retraction\n"
                     if retraction_amount != 0:
                         if firmware_retract: #Can't set the distance directly to what the user wants. We have to choose ourselves.
                             retraction_count = 1 if control_temperatures else 3 #Retract more if we don't control the temperature.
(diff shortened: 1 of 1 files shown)


tlhintoq:
Terrific news!  Thank you.
So... I'm not well versed on python scripting.  But am I correct in thinking that there is a script file for this installed on the client machine (our PC's) - and that we should be able to replace it with your latest version and be good-to-go?  IE: No need to wait for a new version of Cura-We just need to replace the script?

bkp23:
That's correct.

I went about this all backwards. I spent the weekend scratching my head about why pauses wouldn't work for me, then found the python scripts directory and located the bug. Then submitted the fix here and only afterwards noticed that there was an open issue related to it.

Hopefully, my fix gets accepted. If you download it and still encounter problems, let me know right away and I'll take a look. It's working great for me now.

tlhintoq:
I must be blind... I can see the change comparison but not someplace to download just an updated version of the file.

Is it possible for you to just attach the file to the conversation so it can be tested by others without a ton of GitHub rigamarole or waiting for acceptance by the gitGods?  add a .jpg extension so the site thinks its an image?

bkp23:
There's probably a simpler way, but this should work:
1. Click the "Files changed" tab at the top
2. Click the three dots to the right of the filename to open the menu
3. Select "View file" from the menu
4. Click "Raw"

That results in the following URL, which may or may not change 5 minutes after I post this:
https://raw.githubusercontent.com/Ultimaker/Cura/6ed57838a9c35bb063b68bd609c6a5a452898aa9/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py

tlhintoq:
That works.  And... That works.
I did a quick test. .2 layer height and a pause at layer 5.  Makes the match easy.
```
;added code by post processing
;script: PauseAtHeight.py
;current layer: 5
M83 ; switch to relative E values for any needed retraction
G1 F300 Z2.2 ; move up a millimeter to get out of the way
G1 F9000 X190 Y190
G1 F300 Z15 ; too close to bed--move to at least 15mm
M104 S0 ; standby temperature
M0 ; Do the actual pause
M109 S235 ; resume temperature
G1 F300 Z2.2
G1 F9000 X224.782 Y256.097
G1 F300 Z1.2 ; move back down to resume height
```
Sure enough, you can see the Zup to 2.2, Zdown to 2.2, then Zdown 1mm to 1.2

Then checking layer 6 it is correctly at 1.4
```
;MESH:NONMESH
G0 F600 X253.715 Y257.036 Z1.4
G0 F7200 X253.737 Y257.123
G0 X251.776 Y257.425
G0 X250.177 Y257.53
G0 X224.894 Y256.182
;TIME_ELAPSED:265.888060
;LAYER:6
```

Looks like a winner to me!  Very much appreciated!  Thank you.

Ghostkeeper:
Yeah we saw that too. We need to make a fix there.

PR #6302, opened 2019-09-04
Print job uploading

ChrisTerBeke:
This PR is an investigation into the print job uploading issues we have experienced recently. Several small changes combined hopefully reduce the chance that a print job upload fails.

* Reduce logging for material syncing (which was found to be annoying).
* Disable material syncing when uploading print job to reduce network activity to the printer. This changes the current behavior but I think it is fine as the material profiles are still synced after installing a new material profile (which forces a Cura restart and thus device reconnect).
* Remove the re-creation of the network manager (I have found that job uploading got stuck while the network manager was re-created every 30 seconds). I'm aware this could cause new issues on MacOS (as described in the removed lines), but I think this has lower impact than users on all platforms having issues with job uploading.

Before these changes I could reproduce job upload being stuck around 1 in 10 times. After these changes I have yet to experience it (using Ubuntu 18.04 in combination with Connect 5.3.0).

### DIFF
--- a/cura/PrinterOutput/NetworkedPrinterOutputDevice.py
+++ b/cura/PrinterOutput/NetworkedPrinterOutputDevice.py
@@ -35,6 +35,8 @@ class NetworkedPrinterOutputDevice(PrinterOutputDevice):
     def __init__(self, device_id, address: str, properties: Dict[bytes, bytes], connection_type: ConnectionType = ConnectionType.NetworkConnection, parent: QObject = None) -> None:
         super().__init__(device_id = device_id, connection_type = connection_type, parent = parent)
         self._manager = None    # type: Optional[QNetworkAccessManager]
+        self._last_manager_create_time = None       # type: Optional[float]
+        self._recreate_network_manager_time = 30
         self._timeout_time = 10  # After how many seconds of no response should a timeout occur?
 
         self._last_response_time = None     # type: Optional[float]
@@ -131,6 +133,12 @@ class NetworkedPrinterOutputDevice(PrinterOutputDevice):
 
             self.setConnectionState(ConnectionState.Closed)
 
+            # We need to check if the manager needs to be re-created. If we don't, we get some issues when OSX goes to
+            # sleep.
+            if time_since_last_response > self._recreate_network_manager_time:
+                if self._last_manager_create_time is None or time() - self._last_manager_create_time > self._recreate_network_manager_time:
+                    self._createNetworkManager()
+                assert(self._manager is not None)
         elif self._connection_state == ConnectionState.Closed:
             # Go out of timeout.
             if self._connection_state_before_timeout is not None:   # sanity check, but it should never be None here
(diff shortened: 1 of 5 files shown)


cura/PrinterOutput/NetworkedPrinterOutputDevice.py
```diff
@@ -133,12 +131,6 @@ def _update(self) -> None:
 
             self.setConnectionState(ConnectionState.Closed)
 
-            # We need to check if the manager needs to be re-created. If we don't, we get some issues when OSX goes to
-            # sleep.
-            if time_since_last_response > self._recreate_network_manager_time:
```
nallath: This will break osx going to standby and then going out of sleep mode again.

ChrisTerBeke: I know, see my comments.

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
```diff
@@ -122,9 +122,6 @@ def requestWrite(self, nodes: List[SceneNode], file_name: Optional[str] = None,
 
         self.writeStarted.emit(self)
 
-        # Make sure the printer is aware of all new materials as the new print job might contain one.
-        self.sendMaterialProfiles()
-
```
sedwards2009: oooh, yeah. That could have been a problem. It would certainly give our little printer more to do at the same time.

If we still need something like this, we could trigger it on the completion of the print job transmission.

nallath: I'm mostly wondering why this would suddenly be a problem now. I'm pretty sure it has been doing this for a while now.

sedwards2009: It seems to be tempting fate here.

It is possible that with the new MISP firmware on the printer we've hit some kind of limit regarding CPU/memory. We'll have to do some more investigation on the printer too to cover those bases.

ChrisTerBeke: To me it seams that recent changes in logging have made the problem more visible, but it's always been there (especially for large files).

ChrisTerBeke: @sedwards2009 interesting thought, maybe indeed the printers are now slower in processing those requests.

sedwards2009: We should check that out anyway, but it sounds like the QNetworkManager gets the real blame here.

ChrisTerBeke:
@nallath I have just pushed this patch on top of this PR. We could revert some of the other commits if wanted, but we'll discuss in our team first as well.

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
nallath: I don't think it was the QNetworkManager. I'm running it with only the "only send materials on first connect" change and the decreased logging and I've not had the issue.

ChrisTerBeke: We can revert this particular commit then to not introduce other issues

ChrisTerBeke:
After reverting that change to the network manager recreating, I had the issue again sometimes. I still believe this is the cause of some occurrences of this problem.

sedwards2009:
At this stage of the dev cycle I'm 100% for a cautious approach which reduces potential problems, and then testing the crap out of it so that we know we have something stable.

PR #6308, opened 2019-09-05
Remove gl_PrimitiveID from shader when it's not available.

smartavionics:


Ghostkeeper:
We're still messing around with this a bit. We might add the necessary information to the vertices instead of this solution, so that the drop to build plate still works. Not sure yet.

rburema:
@smartavionics Just a heads-up: I/we somehow overlooked this and also forgot to double check if all gl_PrimitiveID's where handled properly now until just after the beta-release, unfortunately :-/ (It'll be in the final though.),

PR #6313, opened 2019-09-06
CS-171 Add R2 support

ianpaschal:


plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDeviceManager.py
```diff
@@ -135,10 +135,13 @@ def _getPrinterTypeIdentifiers() -> Dict[str, str]:
...
         found_machine_type_identifiers = {}  # type: Dict[str, str]
         for machine in ultimaker_machines:
-            machine_bom_number = machine.get("firmware_update_info", {}).get("id", None)
             machine_type = machine.get("id", None)
-            if machine_bom_number and machine_type:
-                found_machine_type_identifiers[str(machine_bom_number)] = machine_type
+            machine_bom_numbers = machine.get("bom_numbers", [])
+            if machine_type and machine_bom_numbers:
+                for bom_number in machine_bom_numbers:
+                    # This produces a n:1 mapping of bom numberss to machine types
```
ChrisTerBeke: spelling mistake in numbers

ianpaschal:
@ChrisTerBeke Yes, this is why I felt that was a "small no-no" which was that duplicate (or in some cases inconsistent) data, but let's go with this for now and Cura can update the firmware checker to handle multiple boms in the future. My only concern though is that if you want to link firmware urls to boms, it would have been better to make a list of bom+url objects or something.

Ghostkeeper:
This merged master into 4.3. I've discussed this with ChrisTerBeke and we're going to undo the merge and then cherry-pick the two commits from ianpaschal to 4.3.

PR #6325, opened 2019-09-09
Switch the scroll bars to controls2

### LINKED ISSUE #6299: [4.2.1] Invisible dragging bars in the machine settings
**Application version**
4.2.1

**Platform**
Win10 x64

**Printer**
Any

**Reproduction steps**
Try to move the bars on the gcode fields

**Actual results**
Bars are not visible before you drag them, and barely visible while selected.

**Expected results**
Bars visible

**Additional information**
https://youtu.be/MMqV_Un639I

diegopradogesto:
Use also the same component across Cura, to always have the same style.

Fixes #6299


diegopradogesto:
Devs., look at internal ticket CURA-6763

ChrisTerBeke:
For some reason this breaks clicking on a marketplace tile to see the details (probably not passing a property along anymore?).

diegopradogesto:
Right, good catch. It should work now.

DonExo:
I've pulled latest master and tried out the PR.

This PR does not fixes the mentioned issue, but rather just "blacks out" the scroll bar. The reporter of the issue requests that the scroll bar in the Machine Settings (Gcode area) are always visible, and not only "when needed" i.e. when hovered.

Do we want to have them shown by default, or only when hovered ?

diegopradogesto:
That's a good one. I think I didn't choose to show them by default, but if you think it's better to show them, then we just need to change the `scrollAlwaysVisible` to true.

PR #6329, opened 2019-09-10
HMS434 change

maukcc:
getting rid of materials
setting quality as material based
added build volume
changed travel speeds


Ghostkeeper:
This will require a version upgrade. The material needs to be removed from the stack and the quality profile needs to be swapped.

maukcc:
Do you require anything from our side to make this happen?

diegopradogesto:
Yes, you should add the changes in the Version Upgrade to remove the material from the stack and handle the new quality profiles.

maukcc:
> add the changes in the Version Upgrade

How do we do that ?

Ghostkeeper:
I think it's a bit much to ask if there is no existing version upgrade plug-in yet. Our version upgrade system is a bit verbose. I'll see what I can do for you.

maukcc:
@Ghostkeeper  thanks, we owe you (another) one :)

Ghostkeeper:
Cookies are welcome ;) Ultimaker B.V. at Watermolenweg 2, Geldermalsen!

maukcc:
There are 2 major problems with the current material setup:
1 materials are vendor based, creating a way too long list of materials
There are 20 (and counting) different PLA and the only thing they differ from, is the vendor name. We used an exclude materials setting in our machine definition file, but this is (from Dutch) mopping with the fosset open. Better to have an include materials.
2 (and this is the important one) Material files do not have the necessary settings in them to produce good quality prints. 
Temp settings and retracts are not enough. At this point we have about 10 other settings that are highly material dependent.
 
For these reasons we had to think up this different route.

possible solutions:
1 only generic materials are standard. All other vendors must be placed in marketplace. (is marketplace in github somewhere?)
2 freely configurable material settings

nallath:
But that's also what the current setup supports. Ultimaker doesn't provide specific quality profiles for Ultimaker Blue PLA or Yellow PLA. The material profile is different, but all of them fall back to the same quality profiles (those of generic PLA)

Ghostkeeper:
Yes but the problem is then that many of those 3rd party materials specifically want their material to not be of the material type "PLA" because that shares the quality profiles with generic PLA. They want their material to be of the material type "Igidur PLA" or somesuch so that they define their own quality profiles for the UMS5 but the rest of the printers say "not supported".

maukcc:
This new commit should make material settings OK.
Still we would like to see all of the vendor specific materials in the marketplace.

How can one add to the marketplace ?

Ghostkeeper:
You can add materials to the marketplace by submitting a material package to https://contribute.ultimaker.com/

The material package is a .zip file that should contain one material profile (like the files in this repo) with optionally multiple quality profiles.

resources/quality/hms434/abs-fr/hms434_0.4_absfr_high.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = High
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = high
+weight = 0
```
diegopradogesto: This weight is used for sorting. I would use `1` here

resources/quality/hms434/abs-fr/hms434_0.8_absfr_coarse.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = Coarse
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = coarse
+weight = 0
```
diegopradogesto: This weight is used for sorting. I would use `-3` here

resources/quality/hms434/asa/hms434_0.4_asa_high.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = High
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = high
+weight = 0
```
diegopradogesto: This weight is used for sorting. I would use `1` here

resources/quality/hms434/asa/hms434_0.8_asa_coarse.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = Coarse
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = coarse
+weight = 0
```
diegopradogesto: This weight is used for sorting. I would use `-3` here

resources/quality/hms434/pa12/hms434_0.4_pa12_high.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = High
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = high
+weight = 0
```
diegopradogesto: I would set a `1` here

resources/quality/hms434/pa12/hms434_0.8_pa12_coarse.inst.cfg
```diff
@@ -0,0 +1,37 @@
+[general]
+version = 4
+name = Coarse
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = coarse
+weight = 0
```
diegopradogesto: I would set a `-3`

resources/quality/hms434/pla/hms434_0.8_pla_coarse.inst.cfg
```diff
@@ -0,0 +1,38 @@
+[general]
+version = 4
+name = Coarse
+definition = hms434
+
+[metadata]
+setting_version = 8
+type = quality
+quality_type = coarse
+
```
diegopradogesto: Code style: remove empty space

resources/quality/hms434/pla/hms434_0.8_pla_coarse.inst.cfg
```diff
@@ -0,0 +1,38 @@
...
+weight = 0
```
diegopradogesto: Because is coarse, I would use `-3`

resources/quality/hms434/pla/hms434_0.8_pla_coarse.inst.cfg
Ghostkeeper: I don't think we really have any rule for this, do we? If it's more readable... Oh well, either way is fine with me.

resources/quality/hms434/abs-fr/hms434_0.4_absfr_high.inst.cfg
```diff
@@ -0,0 +1,37 @@
...
+
+wall_0_wipe_dist = 0.0
+ironing_enabled = true
+
+infill_overlap_mm = -0.1
+infill_wipe_dist = 0
+
+default_material_print_temperature = 255
+material_standby_temperature = 160
+default_material_bed_temperature = 115
```
Ghostkeeper: Seems a bit hot for ABS! Fine with me of course if you think this prints well, but I wanted to point this out in case you made a mistake.

maukcc: This is a special ABS that prints very well at 255

Ghostkeeper:
These profiles refer to ABS-FR, ASA and PA-12 materials, but I can't find them anywhere. Do they still need to be added to fdm_materials?

maukcc:
Our clients will get these material settings from us as an import.

Ghostkeeper:
Your clients will have to mod Cura? Why not supply it by default then?

maukcc:
Company policy to do it this way.

fieldOfView:
Then why have these profiles in Cura? Having half of a dataset in Cura and half of it in a "private" import/mod does not make sense to me.

maukcc:
Our clients need to import just 1 thing : the material.
Not per material also another 4 (for now) profiles, resulting in at least 5 actions for our clients.
We will be looking in to adding these materials into marketplace and then deleting them here. But until that time we will have to do it like this.

fieldOfView:
A .curapackage file can contain both material (xml) files and quality profiles. A .curapackage file can be installed by simply dropping it onto the Cura application window, which would be easier than moving files to specific directories. The .curapackage file would not even have to be hosted on the Ultimaker Marketplace.

fieldOfView:
Here's a .curapackage from the Marketplace as an example:
https://api.ultimaker.com/cura-packages/v1/cura/v6.0.0/packages/3DU_Nylon_CF/download

maukcc:
That is fantastic, thanks. We will delete the profiles and make a new commit.

At this point we make a very long exclude material list in our machine definition file. And this list needs to be updated every time someone else adds yet another PLA or whatever.
Is there a way to exclude all materials that come bundled?

Ghostkeeper:
It sort of goes against the grain of an open filament system to decline 3rd party filaments by default. We haven't built that, no.

We did plan on removing the entire exclude list at some point and encode that in the material profiles, but we didn't get around to doing that yet. It's a bit difficult with which materials get displayed by Ultimaker's printer firmware. But even then it'll have the same problem for your perspective.

resources/quality/hms434/abs-fr/hms434_0.4_absfr_high.inst.cfg
ninovanhooff: In a later commit, -1 is used for coarse, 0 for normal and 1 for high. These have been used consistently so this comment can be closed now

resources/quality/hms434/hms434_global_Super_Coarse_Quality.inst.cfg
```diff
@@ -1,16 +0,0 @@
-[general]
```
LipuFei: If you remove or rename profiles, you also need to add version upgrade to make sure that the machines using those removed/renamed profiles in the old version can be upgraded correctly to use the correct profiles in the new version. I will merge this and then make the version upgrade.

LipuFei: Version upgrade made in 1977a865f1b0d7ebf301603a4c2e1323e1e5b695

maukcc:
@fieldOfView  is there a specific compressing being done on the .curapackage files?

fieldOfView:
It is just a "deflated" zip package as far as I know, but I always have the contribute.ultimaker.com site build my packages.

Ghostkeeper:
Yeah as far as I understood those profiles should've been accepted.

For an example of a .curapackage file, you can get these online: https://github.com/Ghostkeeper/SVGToolpathReader/releases/download/v1.1.2/SVGToolpathReader6.0.0.curapackage
However that is a plug-in. But the only real difference is the contents of the `/files` folder. In your case the `/files` folder should contain something more like [my material package](https://github.com/Ghostkeeper/3D4MakersFacilianC8). I don't have an example of a material package at the moment though. I can only get that when I'm back home due to a bug in the Contribution Portal.

PR #6334, opened 2019-09-11
CS-158 Fix monitor tab print status

ianpaschal:
This is two things in one to hopefully avoid having this bug from ever returning from the grave for a 3rd time:

1. Reduce text size from 'medium' to 'default'. It looks a little small, but there simply isn't more room in the design as it was created and wrapping to the next line looks silly.
2. Reduce the progress bar length a bit so that the percent label's left side as at the "column edge" (aligned with the 2nd extruder info).

Screen Shot 2019-09-11 at 11 27 57


resources/themes/cura-light/theme.json
```diff
@@ -619,6 +619,8 @@
         "monitor_empty_state_offset": [5.6, 5.6],
         "monitor_empty_state_size": [35.0, 25.0],
         "monitor_external_link_icon": [1.16, 1.16],
-        "monitor_column": [18.0, 1.0]
+        "monitor_column": [18.0, 1.0],
+        "monitor_progress_bar": [16.5, 1.0],
+        "monitor_margin": [1.5, 1.5]
```
Ghostkeeper: Is there a reason for the progress bar and margins in the monitor to look different from the progress bar and the margins in the rest of Cura? Wouldn't that look inconsistent?

ianpaschal: Since this bug has been solved once before but then reappeared, I'm trying to lock it down very strongly so that the spacing and sizing here doesn't accidentally change somehow from external factors in the future.

This is the problem I always had with the Cura theming system. It doesn't cascade or inherit with overrides like CSS would. So we end up in a situation where it should probably look similar, but that means changing it in one place will likely break the UI in other places. That's what I'm trying to avoid here.

ianpaschal: Also it's worth noting that this is just setting the length of the progress bar which should change depending on what part of the UI it's in...

ianpaschal:
Doh, closing because of the branching

PR #6335, opened 2019-09-11
CS-158 Fix monitor tab print status (4.3 cherry pick)

ianpaschal:
> **Note:** Cherry picked some commits to a new branch from 4.3

This is two things in one to hopefully avoid having this bug from ever returning from the grave for a 3rd time:

1. Reduce text size from 'medium' to 'default'. It looks a little small, but there simply isn't more room in the design as it was created and wrapping to the next line looks silly.
2. Reduce the progress bar length a bit so that the percent label's left side as at the "column edge" (aligned with the 2nd extruder info).

Screen Shot 2019-09-11 at 11 27 57


ChrisTerBeke:
That printer name cracks me up every time 😆

ianpaschal:
@ChrisTerBeke 
US_Army_Detroit_Tank_Plant

PR #6347, opened 2019-09-13
Store in preferences if the user wants to hide the message

ChrisTerBeke:
Users might not want to see the message that new cloud printers were found every time. This PR adds a checkbox to allow them to disable the message in the future.

Also fixed loading another popup's image on Windows.


nallath:
As a token of my gratitude, i would like to present this picture;
kittenImage

PR #6353, opened 2019-09-14
Create Original_Prusa_i3_MK3S_MK3_extruder_0.def

samirabaza:
This is the latest extruder definition for Prusa i3 MK3/MK3s .
reference :
https://manual.prusa3d.com/Guide/How+to+import+profiles+to+Cura+4.x+(Windows+&+macOS)/1421#_ga=2.78043415.205833688.1568467545-859345073.1567270611

### DIFF
--- a/resources/extruders/Original_Prusa_i3_MK3S_MK3_extruder_0.def.json
+++ b/resources/extruders/Original_Prusa_i3_MK3S_MK3_extruder_0.def.json
@@ -1,0 +1,16 @@
+{
+    "id": "Original_Prusa_i3_MK3S_MK3_extruder_0",
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "Original_Prusa_i3_MK3S_MK3",
+        "position": "0"
+    },
+
+    "overrides": {
+        "extruder_nr": { "default_value": 0 },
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
To mirror Nallath's comment on the other PRs, quoth:

> The 4.3 release is already closed for new machines, so you will need to make a pull request for master. Could you also put all the files in a single pull request?

We'd prefer to have these changes in a single PR since one cannot be merged without the other. For instance, this PR adds an extruder definition that isn't being used by anything.

samirabaza:
> 
> 
> To mirror Nallath's comment on the other PRs, quoth:
> 
> > The 4.3 release is already closed for new machines, so you will need to make a pull request for master. Could you also put all the files in a single pull request?
> 
> We'd prefer to have these changes in a single PR since one cannot be merged without the other. For instance, this PR adds an extruder definition that isn't being used by anything.

@Ghostkeeper and @nallath 
I may not be the right contributor to understanding what is the procedure to add the printer?
if possible, explain it to me like for a newbie.

Ghostkeeper:
Well you were mostly correct, except:

* Please always submit PRs to master. Only for bug fixes can we make an exception to add it to different branches.
* It's not required, but much easier for us to have a pull request that works on its own, so not several pull requests that if one gets merged but not the other we have a broken application. It doesn't need to be split off per file or per commit or anything.

PR #6355, opened 2019-09-14
Fix spelling mistakes in blurb for small feature settings.

smartavionics:
Sorry, didn't notice this untll today.


Ghostkeeper:
Normally I'd merge these sort of things in the release branch but since that is string frozen right now it'll have to be for the next release. Thanks for the correction!

PR #6359, opened 2019-09-15
Add gyroid infill density variants

smartavionics:
Adds low and medium resolution gyroid infill variants.

See https://github.com/Ultimaker/CuraEngine/pull/1135.


fieldOfView:
Wouldn't it make sense to have one Gyroid option, and a second setting (Gyroid resolution) that is only visible when Gyroid is selected as infill pattern?

smartavionics:
> Wouldn't it make sense to have one Gyroid option, and a second setting (Gyroid resolution) that is only visible when Gyroid is selected as infill pattern?

That would require more more code in the engine and another setting in the UI. I try hard not to add settings as the UI is already dog slow.

Personally, I don't care that there's now extra entries in the pattern menu.

Ghostkeeper:
I agree though. Having an option "Gyroid (medium)" doesn't really tell me what's medium about that Gyroid.

smartavionics:
> I agree though. Having an option "Gyroid (medium)" doesn't really tell me what's medium about that Gyroid.

Agree with what? Having the extra menu entries or having a new gyroid resolution setting? Actually, that would be two new settings one for the infill and one for the support.

It's true, "medium" isn't very descriptive. Adding "resolution" would be informative but would make the menu wider.

smartavionics:
Closing this for now, may open again in the future or replace with another PR.

PR #6360, opened 2019-09-15
Add Infill Vertical Scaling setting

smartavionics:
Adds setting that scales (stretches/shrinks) infill pattern vertically. Currently only implemented by gyroid pattern.

See https://github.com/Ultimaker/CuraEngine/pull/1135.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1790,13 +1790,13 @@
                 "infill_scaling_z":
                 {
                     "label": "Infill Vertical Scaling",
-                    "description": "Scales the vertical infill pattern. Values larger than 100% stretch the pattern, values less than 100% compress the pattern.",
+                    "description": "Scales the vertical infill pattern. Values larger than 100% stretch the pattern, values less than 100% compress the pattern. Note that some infill patterns do not work well with scaling values that differ greatly from 100%.",
                     "unit": "%",
                     "minimum_value": "1",
                     "maximum_value": "1000000",
                     "default_value": 100,
                     "type": "float",
-                    "enabled": "infill_sparse_density > 0 and infill_pattern.startswith('gyroid')",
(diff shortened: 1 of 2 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1787,6 +1787,18 @@
                     "limit_to_extruder": "infill_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "infill_scaling_z":
+                {
+                    "label": "Infill Vertical Scaling",
+                    "description": "Scales the vertical infill pattern. Values larger than 100% stretch the pattern, values less than 100% compress the pattern.",
```
BagelOrb: "Scales the infill pattern **vertically**"
(There is no such thing as the 'vertical infill pattern')

Please explain why a user would use this. E.g. "Values larger than 100% make the infill more stiff in the vertical direction."

BagelOrb:
If you just multiply the `coord_t z` in infill.h by the `scaling_z` ?
That way this setting holds for all infill patterns instead of only the gyroid one!

(Except that it only does something for the 3D infill patterns: cross 3D, octet, etc)

resources/definitions/fdmprinter.def.json
smartavionics: Somebody asked me to implement it because they thought it would be useful. I did so but can't offer any justification for it other than it can make interesting variations in the gyroid infill.

Screenshot_2019-09-14_09-25-52

smartavionics:
> If you just multiply the coord_t z in infill.h by the scaling_z ?
That way this setting holds for all infill patterns instead of only the gyroid one!

Yes, I have tried that and it does scale the other 3d patterns so I will mod my PRs. Thanks for the idea.

smartavionics:
Closing this for now, may open again in the future or replace with another PR.

Issue #5651, opened 2019-04-22
cura keeps crashing 


EduardoOliveira:
**Application Version**
3.6 and 4.0

**Platform**
Windows 10

**Printer**
Anycubic i3 mega

**Steps to Reproduce**
Normal utilization, random actions

**Actual Results**
Stops working, even wen restarted.
It seems to corrupt some config file, only works when completely uninstalled and reinstalled.

**Expected results**
to not stop working

**Additional Information**
Log of a crashing boot
```2019-04-22 22:34:25,042 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin ConsoleLogger
2019-04-22 22:34:25,057 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin CuraDrive
2019-04-22 22:34:25,082 - INFO - [MainThread] CuraEngineBackend.CuraEngineBackend.__init__ [81]: Found CuraEngine at: C:\Program Files\Ultimaker Cura 4.0\CuraEngine.exe
2019-04-22 22:34:25,085 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin CuraEngineBackend
2019-04-22 22:34:25,095 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin CuraProfileReader
2019-04-22 22:34:25,104 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin CuraProfileWriter
2019-04-22 22:34:25,114 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin OBJReader
2019-04-22 22:34:25,126 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin OBJWriter
2019-04-22 22:34:25,142 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin STLReader
2019-04-22 22:34:25,153 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin STLWriter
2019-04-22 22:34:25,162 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin FileLogger
2019-04-22 22:34:25,175 - INFO - [MainThread] UM.PluginRegistry.loadPlugin [436]: Loaded plugin FirmwareUpdateChecker
[...]

PR #6385, opened 2019-09-20
Add preference to (not) restore the window geometry on startup

fieldOfView:
This PR adds a preference around restoring the previous window position/size to the last used position/size on start up. This would be a workaround for those setups where starting Cura on a secondary screen will prevent it from working. 

Contributes to https://github.com/Ultimaker/Cura/issues/5651#issuecomment-488048392 and the rest of these issues: https://github.com/ultimaker/cura/issues?utf8=%E2%9C%93&q=is%3Aissue+window_left+is%3Aopen+

Requires https://github.com/Ultimaker/Uranium/pull/523

See https://github.com/Ultimaker/Uranium/pull/523

### DIFF
--- a/resources/qml/Preferences/GeneralPage.qml
+++ b/resources/qml/Preferences/GeneralPage.qml
@@ -94,6 +94,8 @@ UM.PreferencesPage
         zoomToMouseCheckbox.checked = boolCheck(UM.Preferences.getValue("view/zoom_to_mouse"))
         UM.Preferences.resetPreference("view/top_layer_count");
         topLayerCountCheckbox.checked = boolCheck(UM.Preferences.getValue("view/top_layer_count"))
+        UM.Preferences.resetPreference("general/restore_window_geometry")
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
For user experience, if someone is having this crash, it's going to be hard to reset this preference though. Even if the user is sometimes able to start Cura, he's not going to link that crash to this checkbox in the preferences screen.

fieldOfView:
The UX is always going to be bad; it provides a workaround for a crash/non-starting application. I can rename the checkbox to "check this if Cura crashes on startup after you have moved it to a secondary display".  Instead I chose a more positive description of a feature that at least will give a workaround for some users being unable to launch Cura if they did not remember to move the window back to the primary screen before closing it.

nallath:
And it will prevent the issue from happening after the first time (because then the user set the preference)

nallath:
CURA-6901

Issue #6375, opened 2019-09-18
"Pause at height" resumes with a wrong speed


VijitCoder:
**Application version**
3.6.0

**Platform**
Kernel: 4.15.0-54-generic 
x86_64 bits: 64 
Desktop: Cinnamon 4.0.10
Distro: Linux Mint 19.1 Tessa

**Printer**
Creality Ender 3

**Reproduction steps**

1. Take [this model](https://www.thingiverse.com/thing:2203039/files).

2. Add Post Processing Plugin > Pause at height. Config:

```
Pause at: layer 12
Park Print Head X: 190
Park Print Head Y: 190
Retraction: 4
Retraction Speed: 40
Extrude Amount: 0
Extrude Speed: 3.3333
Redo Layers: 1
Standby Temperature: 245
```

3. In the slicer these settings are important:
```
Infill/Wall speeds: 30/15 mm/s
Special Modes > Spiralize Outer Contour = true
Smooth Spiralized Contours = true 
```

4. Generate the gcode file.

**Actual results**

A few layers after pause will be printed on high speed, then it will be reduced to an expected value.

**Expected results**

The same printing speed as before so and after "Pause at height".

**Additional information**

The bug in the speed setting after removing pause. Here is an example of the generated code:

```
;TYPE:CUSTOM
;added code by post processing
;script: PauseAtHeight.py
;current layer: 12
M83
G1 F2400 E-4
G1 F300 Z4.2
G1 F9000 X190 Y190
G1 F300 Z15
M104 S255; standby temperature
M0;Do the actual pause
M109 S245; resume temperature
G1 F2400 E4
G1 F2400 E-4
G1 F300 Z4.2
G1 F9000 X90.765 Y95.63
G1 F2400 E4
G1 F9000         ; <----------------- Here is the wrong instruction
M82
G92 E1740.80036
;LAYER:11
;TYPE:WALL-OUTER
;MESH:cloth_vase.stl
G1 X90.765 Y95.63 E1740.80036
G1 X90.765 Y95.63 E1740.80036
G1 X91.494 Y94.841 Z2.952 E1740.88969
G1 X92.26 Y94.038 Z2.953 E1740.98196
...
```

[...]

PR #6400, opened 2019-09-23
Cura 6799 pause at height resume speed

ninovanhooff:
Curretly the Feedrate is set to 9000 at the end of the Pause at height script. @ChrisTerBeke, please let me know if there was a valid reason for that.
After a Pause at a certain height, the previous feed rate should be restored. This PR implements it for the general Pause At height script and the Repetier script.

The Duet and BQ, scripts should be unaffected by this issue and thus not need a change


ninovanhooff:
Solves https://github.com/Ultimaker/Cura/issues/6375

PR #6428, opened 2019-09-27
Set BIBO print speed to 40mm/s

photodude:
BIBO and members from the BIBO group on Facebook say this printer should be run at 40-45mm/s and not the FDMprinter default of 60mm/s. BIBO provided test gcode files also show that they are setting 40mm/s in the slicer to create their test files.

This change sets the print speed override so the printer profile does not start at the FDMPrinter print speed default of 60mm/s but at the reccommended value of 40mm/s


Ghostkeeper:
Thank you for the tweak! Let's home that prints come out more pretty and more reliable now for them :smiley:

photodude:
You are welcome @Ghostkeeper

Issue #4011, opened 2018-07-02
[3.4.0] Arranging models for one at a time printing fails


upatamby:
**Application Version**
Ultimaker Cura 3.4.0

**Platform**
Windows 10 Pro

**Printer**
Ultimaker 3 Extended

**Steps to Reproduce**
Laod body.stl in Cura. 
Select Extruder 2 for this model.
Rotate the model 90 degrees anti clockwise using the green circle (this is the X/Z-plane).
Multiply the model 8 times.
In my profile I used PETG (based upon a CPE-profile), but I don't think the material is important here.
Don't use any support and no raft or brim.
Choose Special Modes, Print Sequence = One at a Time.
**Actual Results**
The models are ordered in 3 rows and 3 columns.
Seen from above:
![image](https://user-images.githubusercontent.com/35468193/42172562-a84e2816-7e1c-11e8-8961-9964402138da.png)

Start printing.

The print starts with the model in the left column, front row.
1.
x x x
x x x
X x x
2.
x x x
X x x
X x x
3.
X x x
X x x
X x x
So far so good.
But then comes:
4.
X x x
X x x
X X x
While printing the 4. model the printhead touches the already printed model left from it.
It made the whole build plate waggling, so I aborted the printers menu.
And then - this is a severe issue!!! - the printhead moved straight back to its home position (the back corner at the left) without first lowering the build plate, so it jammed to the printed models and got stuck before reaching the corner because a model got in the way and then the printhead started to make a very bad noise. I switched off the printer and hope nothing got damaged.

**Expected results**
First of all: the arrangement of the models should be correct so the models and printhead should never collide. Second (but very important): if the print ends or aborts, the buildplate should lower itself before the printhead starts moving.
[issue.curaproject.3mf.zip](https://github.com/Ultimaker/Cura/files/2155867/issue.curaproject.3mf.zip)

Vandrasc:
@jackha , is this related to the other arrange ticket(Cura-5468) or a different issue? Thank you

jackha:
1) Aborting does not work correctly, which is a known issue in the firmware team. Unfortunately it hasn't been picked up yet EM-2284
2) Ah! I think it's a bug of Extruder 2 + one at a time. I'll check.

jackha:
One at a time bug confirmed. Currently testing if it only affects Extruder 2, or that the big convex hull doesn't take skirt and brim into account. Or a combination of those.
Devs: CURA-5538

[...]

PR #6434, opened 2019-09-27
Cura 6785 one at a time ordering

ninovanhooff:
In juli 2018, the OneAtATimeIterator - which determines the order in which objects are printed - was rewritten. Based on bug reports, the old implementation performed better than the new one. It was decided that the old implementation was a better starting point for this fix.

I verified that some obvious improvements of 06-2018 are kept in place. The Shapely library is still used for higher precision in intersection calculation.

As another improvement and hopefully fix to the problem we see now, the nozzle offset is taken into account. 

Given that...
A: An earlier ticket, https://jira.ultimaker.com/browse/CURA-5538
( source: https://github.com/Ultimaker/Cura/issues/4011 ) explicitly states that the issue only occurs with extruder 2 and
B: The nozzle offset relative to the printhead for extruder 1 is usually 0 and non-zero for other extruders (eg: Ultimaker 3 Extended, mentioned in the github issue)

...It is likely that not taking nozzle offset into account is the root cause.

Note that this fix was made in ConvexHullDecorator and should benefit other code which deals with the printhead geometry as well.

### DIFF
--- a/cura/OneAtATimeIterator.py
+++ b/cura/OneAtATimeIterator.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2019 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
+from typing import List
+
 from UM.Scene.Iterator import Iterator
 from UM.Scene.SceneNode import SceneNode
 from functools import cmp_to_key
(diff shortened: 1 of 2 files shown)


cura/OneAtATimeIterator.py
```diff
@@ -1,149 +1,112 @@
...
+            for node in current.todo:
+                # Check if the object can be placed with what we have and still allows for a solution in the future
+                if not self._checkHitMultiple(node, current.order) and not self._checkBlockMultiple(node, current.todo):
+                    # We found a possible result. Create new todo & order list.
+                    new_todo_list = current.todo[:]
+                    new_todo_list.remove(node)
+                    new_order = current.order[:] + [node]
+                    if len(new_todo_list) == 0:
+                        # We have no more nodes to check, so quit looking.
+                        todo_node_list = None
```
Ghostkeeper: This value is unused. It always returns after setting this to `None`.

Ghostkeeper: Removed here: https://github.com/Ultimaker/Cura/pull/6434/commits/7bf2fa3b43d1ffe30af1f6161d37ceb8ed09d905

PR #6450, opened 2019-09-29
Changed the JG Aurora z603s definition to send the print head to the …

oorail:
…rear of the printer. The z603S is a COREXY style printer, unlike the other JG Aurora printers where moving the Y-axis shifts the bed to the front for easy print removal. On the z603s, it shoves the print head into the frame of the printer, the X-axis bar and the print head are then in the way, making it difficult to remove the print. The z603S should have the print head homed on X and Y axis to tuck it out of the way in the rear right of the printer at the end of a print.

### DIFF
--- a/resources/definitions/jgaurora_z_603s.def.json
+++ b/resources/definitions/jgaurora_z_603s.def.json
@@ -21,7 +21,7 @@
             "default_value": "; -- START GCODE --\nG21                     ;set units to millimetres\nG90                     ;set to absolute positioning\nM106 S0                 ;set fan speed to zero (turned off)\nG28                     ;home all axis\nM420 S1                 ;turn on mesh bed levelling if enabled in firmware\nG92 E0                  ;zero the extruded length\nG1 Z1 F1000             ;move up slightly\nG1 X60.0 Z0 E9.0 F1000.0;intro line\nG1 X100.0 E21.5 F1000.0 ;continue line\nG92 E0                  ;zero the extruded length again\n; -- end of START GCODE --"
         },
         "machine_end_gcode": {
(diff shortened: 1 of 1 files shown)


pinchies:
Thanks @oorail , you are correct, that is my mistake with the original profile.

Ghostkeeper:
Since Github doesn't handle such long lines for clarity, the actual difference is that

    G0 Y280 F600

got changed into:

    G28 Y0

resources/definitions/jgaurora_z_603s.def.json
```diff
@@ -21,7 +21,7 @@
...
         },
         "machine_end_gcode": {
-            "default_value": "; -- END GCODE --\nM104 S0                 ;turn off nozzle heater\nM140 S0                 ;turn off bed heater\nG91                     ;set to relative positioning\nG1 E-10 F300            ;retract the filament slightly\nG90                     ;set to absolute positioning\nG28 X0                  ;move to the X-axis origin (Home)\nG0 Y280 F600            ;bring the bed to the front for easy print removal\nM84                     ;turn off stepper motors\n; -- end of END GCODE --"
+            "default_value": "; -- END GCODE --\nM104 S0                 ;turn off nozzle heater\nM140 S0                 ;turn off bed heater\nG91                     ;set to relative positioning\nG1 E-10 F300            ;retract the filament slightly\nG90                     ;set to absolute positioning\nG28 X0                  ;move to the X-axis origin (Home)\nG28 Y0            ;move to the Y-axis origin (Home) for easy print removal\nM84                     ;turn off stepper motors\n; -- end of END GCODE --"
```
Ghostkeeper: Shouldn't it also specify the `F` again? Otherwise it's going to move at `F300`, the speed that was used for the retraction.

oorail: That is a good point. I went ahead and updated the commit to move it at F600 which the original command was doing. I also optimized the command into a single command instead of two, since we didn't really need to do G28 X0 and then G28 Y0, plus the X0 command would have also run at F300.

oorail: I also tested the last change G28 X0 Y0 F600 in Cura 4.3.0 to make sure it worked as expected.

Ghostkeeper:
Looks like it produces the desired result now! Thank you.

I did realign the comment.

PR #6456, opened 2019-10-01
Consider user preference when opening project files from cli.

ninovanhooff:
The logic using just a "skip_project_file_check" boolean was too obscure
imho, so used a string value with more explicit values to improve readability

CURA-6824

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1600,8 +1600,11 @@ class CuraApplication(QtApplication):
 
     openProjectFile = pyqtSignal(QUrl, arguments = ["project_file"])  # Emitted when a project file is about to open.
 
-    @pyqtSlot(QUrl, bool)
-    def readLocalFile(self, file, skip_project_file_check = False):
+    @pyqtSlot(QUrl, str)
+    ## Open a local file
+    # \param force_project_mode \type{str} Either "open_as_model" or "open_as_project". This parameter is only
+    # considered if the file is a project file.
+    def readLocalFile(self, file, force_project_mode = None):
         if not file.isValid():
             return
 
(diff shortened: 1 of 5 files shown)


cura/CuraApplication.py
```diff
@@ -1600,8 +1600,11 @@ def log(self, msg):
 
     openProjectFile = pyqtSignal(QUrl, arguments = ["project_file"])  # Emitted when a project file is about to open.
 
-    @pyqtSlot(QUrl, bool)
-    def readLocalFile(self, file, skip_project_file_check = False):
+    @pyqtSlot(QUrl, str)
+    ## Open a local file
+    # \param force_project_mode \type{str} Either "open_as_model" or "open_as_project". This parameter is only
+    # considered if the file is a project file.
+    def readLocalFile(self, file, force_project_mode = None):
```
Ghostkeeper: Please provide typing in the parameter rather than in the documentation. If you're looking at other places where it was in the documentation, that comes from legacy from before we were using Python 3.5. If it's in the parameter, MyPy can read it.

When providing the typing, MyPy will also start complaining that it should be `Optional[str]` rather than just `str`.

cura/CuraApplication.py
```diff
@@ -1612,10 +1615,25 @@ def readLocalFile(self, file, skip_project_file_check = False):
                 self.deleteAll()
                 break
 
-        if not skip_project_file_check and self.checkIsValidProjectFile(file):
+        is_project_file = self.checkIsValidProjectFile(file)
+        project_mode = force_project_mode
```
Ghostkeeper: Useless alias. You can use `force_project_mode` for the rest of the code or rename the parameter to `project_mode`. I'd prefer `force_project mode` (your new name) since it's a bit more explicit.

ninovanhooff:
> I'm wondering why this was only going wrong in the CLI, if this is the fix. Is `readLocalFile` only used by the CLI?

Opening as project was only done inside the dialog qml.
The boolean that existed was used to open as model

cura/CuraApplication.py
ninovanhooff: Okay, I'm used to languages where mutating parameters is considered bad style.
I'm going with `project_mode` though

cura/CuraApplication.py
```diff
@@ -1600,8 +1600,11 @@ def log(self, msg):
 
     openProjectFile = pyqtSignal(QUrl, arguments = ["project_file"])  # Emitted when a project file is about to open.
 
-    @pyqtSlot(QUrl, bool)
-    def readLocalFile(self, file, skip_project_file_check = False):
+    @pyqtSlot(QUrl, str)
+    ## Open a local file
+    # \param project_mode How to handle project files. Either None(default): Follow user preference, "open_as_model" or
+    # "open_as_project". This parameter is only considered if the file is a project file.
+    def readLocalFile(self, file: QUrl, project_mode: Optional[str] = None):
```
nallath: This function is used in other places (such as recentFilesMenu, line 66). There it's still giving a bool instead of a string.

PR #6473, opened 2019-10-03
CURA-6632 Add intent into stats

LipuFei:


cura/Settings/GlobalStack.py
```diff
@@ -131,6 +131,14 @@ def getConfigurationTypeFromSerialized(cls, serialized: str) -> Optional[str]:
             return "machine_stack"
         return configuration_type
 
+    def getIntentCategory(self) -> str:
+        intent_category = "default"
+        for extruder in self.extruderList:
+            category = extruder.intent.getMetaDataEntry("intent_category", "default")
+            if category != "default" and category != intent_category:
+                intent_category = category
+        return intent_category
```
Ghostkeeper: This code duplicates `MachineManager.activeIntentCategory`. We need to remove either one.

LipuFei: `MachineManager.activeIntentCategory` calls this function to get the intent cateogry. It also checks if the active global stack is present and always returns a string.

cura/Settings/MachineManager.py
```diff
@@ -622,12 +622,7 @@ def activeIntentCategory(self) -> str:
 
         if not global_container_stack:
             return ""
-        intent_category = "default"
```
nallath: This change is breaking one of the tests. So either this change is wrong, or the test is wrong.

LipuFei: Fixed by rebasing on the latest `feature_intent` branch.

ninovanhooff: Confirmed: no more failing tests

PR #6480, opened 2019-10-04
CURA-6706 Make intents work on quality management page

LipuFei:


cura/Machines/Models/IntentModel.py
```diff
@@ -74,21 +74,7 @@ def _update(self) -> None:
                     new_items.append(intent)
                     layer_heights_added.append(intent["layer_height"])
 
-        # Now that we added all intents that we found something for, ensure that we set add ticks (and layer_heights)
```
nallath: This will break the recommended view (which is also mentioned in the comment that it's needed for that reason).

cura/Machines/Models/QualityManagementModel.py
```diff
@@ -233,6 +247,31 @@ def _qualityChangesListChanged(self, container: "ContainerInterface") -> None:
...
+        intent_category = quality_model_item["intent_category"]
+
+        # Not a custom quality
+        if is_read_only:
+            return display_name
+
+        # A custom quality
+        if intent_category != "default":
+            from cura.Machines.Models.IntentCategoryModel import IntentCategoryModel
+            intent_display_name = IntentCategoryModel.name_translation.get(intent_category,
```
nallath: We don't need to do it like this, right? The catalog will already have the labels, so you should be able to just directly ask it (which prevents you from having to import the category model and ask that what the translation is)

cura/Machines/Models/QualityManagementModel.py
```diff
@@ -253,27 +292,53 @@ def _update(self):
...
+        result = []
+        for intent_category, quality_type in available_intent_list:
+            result.append({
+                "name": quality_group_dict[quality_type].name,  # Use the quality name as the display name
+                "is_read_only": True,
+                "quality_group": quality_group_dict[quality_type],
+                "quality_type": quality_type,
+                "quality_changes_group": None,
+                "intent_category": intent_category,
+                "section_name": IntentCategoryModel.name_translation.get(intent_category, catalog.i18nc("@label", "Unknown")),
```
nallath: Same as above. You can just directly ask the catalog!

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -43,14 +43,22 @@ Item
...
+    property var currentItemDisplayName: hasCurrentItem ? base.qualityManagementModel.getQualityItemDisplayName(base.currentItem) : ""
 
     property var isCurrentItemActivated:
     {
         if (!base.currentItem)
         {
             return false;
         }
-        return base.currentItem.name == Cura.MachineManager.activeQualityOrQualityChangesName;
+        if (base.currentItem.is_read_only)
```
nallath: Why do you need a different behavior if the item is read_only?

cura/Machines/Models/QualityManagementModel.py
LipuFei: Fixed in 2033f24a1

cura/Machines/Models/QualityManagementModel.py
LipuFei: Fixed in 2033f24a1

resources/qml/Preferences/ProfilesPage.qml
LipuFei: Because the functions (or APIs) for selecting a custom profile and a built-in quality/intent profile are different.

cura/Machines/Models/IntentModel.py
LipuFei: Fixed in d99386cf2

PR #6499, opened 2019-10-07
Cura 6853 layer slider behind action panel

nallath:


plugins/SimulationView/LayerSlider.qml
```diff
@@ -333,7 +338,7 @@ Item
         // set the slider position based on the lower value
         function setValue(value)
         {
-
+            print("lower handle set value: " + value)
```
nallath:
```suggestion
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -209,6 +216,7 @@ Item
         // Make sure the slider handlers show the correct value after switching views
         Component.onCompleted:
         {
+            print("paneltop", panelTop, "screenscaleFactor")
```
nallath:
```suggestion
```

nallath:
I'm not really a fan of this solution, but apart from changing the actual location of those sliders, I don't really see any other way to fix this. 

Changing the order of the loader and the actionPanelWidget in previewMain at least ensures that both of them are usable, but then it simply lets the sliders overlap the actionPanelWidget.

ninovanhooff:
I removed the dependency on external code. Kindly reconsider.

Issue #6106, opened 2019-07-31
Discussion and suggestions about changes for the default Creawesome Creality printer profiles.


JohnEdwa:
The new Creawesome profiles have some odd choices, some missed Cura features and in my opinion, could be better. After [a very small Reddit discussion](https://www.reddit.com/r/ender3/comments/ck9agf/help_making_the_new_cura_42_creawesome_profiles/), I compiled my proposed changes and would like to get some feedback and as a possible result, have updated and better default printer profiles included into Cura.

For reference, these are all changes for the "[creality/base/base_global_standard.inst.cfg](https://github.com/Ultimaker/Cura/blob/master/resources/quality/creality/base/base_global_standard.inst.cfg)" using the printer "[creality_ender3.def](https://github.com/Ultimaker/Cura/blob/master/resources/definitions/creality_ender3.def.json)" that itself is based on the "[creality_base.def.json](https://github.com/Ultimaker/Cura/blob/master/resources/definitions/creality_base.def.json)" definition.

The comments are mostly mine, and might (will) have factual errors and misunderstood functionality as I'm not a Cura expert.
~~Cura Profile: [creawesome_ender3_change_suggestions_v1.zip](https://github.com/Ultimaker/Cura/files/3453974/creawesome_ender3_change_suggestions_v1.zip)~~

---

#Suggested in the discussion but I don't know about them
* Print Thin Walls should be enabled.
* Minimum Extrusion Distance Window: 10 is to large of a value according to /u/Liger_Phoenix
* Line Width should be equal to Nozzle size, according to /u/Liger_Phoenix

---

creality_base.def.json

#Shell    

* Outer Wall Wipe Distance: 0 -> 0.2mm // this is supposed to reduce the Z-seam, so why not use it?
* Filter Out Tiny Gaps: False -> True // the tooltip just says "Filter out tiny gaps to reduce blobs on the ouside of the model". Sound fine to me
* Seam Corner Preference: None -> Smart Hiding // Hiding works rather well and looks good.
* Z Seam X: 117.5 -> 0 // Using the Relative Z seam instead
* Z Seam Relative: False -> True // see above
* Optimize wall printing order = true # helps reducing unnecessary travel (Pimentoso)

#Material
* Retraction Extra Prime Amount: 0 -> 0.064 // see Enable Coasting at the bottom - should be a nozzle-based function and not a hard value.

#Infill
[...]

PR #6501, opened 2019-10-07
Creality Legacy Printer Definitions

kazooless:
The major changes to the Creality printer definitions has caused quite a stir in the miniature hobbyist forums and Facebook groups. Importing old profiles does not give the end-user the expected results due to the previous template profiles no longer available to the new Creality printer definitions. This is my first step into helping that community get their previous qualities and definitions back.

Note, apparently there are enough differences between the legacy Creality definitions and the plain custom fdmprinter definitions to make a significant difference. We have found creating a custom fdmprinter and importing the old quality profiles does not yeild great prints, but using the legacy definitions I've included here, our prints are right back to their great quality.

I've given these files to a couple of Facebook groups and they are receiving a lot of accolades.

My thought is if the powers that be are okay with approving this, we can work to integrate them into the same manufacturer for the next release, instead of creating a second "Creality Legacy 3D" manufacturer.

PS Daniel Herrero says he discussed this with @nallath , which is why I'm submitting this for a pull. This is the first time I've ever used GitHub to contribute as I'm not really a programmer. Hopefully I'm doing this right.


resources/variants/creality_legacy_base_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.2mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_0.4.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.4mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_0.5.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.5mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_0.6.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.6mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_0.8.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.8mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_1.0.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 1.0mm Nozzle
+version = 4
+definition = creality_legacy_base
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 9
```
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.2mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.3.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.3mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.4.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.4mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.5.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.5mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.6.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.6mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_0.8.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.8mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_cr10_1.0.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 1.0mm Nozzle
+version = 4
+definition = creality_legacy_cr10
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_ender3_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.2mm Nozzle
+version = 4
+definition = creality_legacy_ender3
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_ender3_0.3.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.3mm Nozzle
+version = 4
+definition = creality_legacy_ender3
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_ender3_0.4.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.4mm Nozzle
+version = 4
+definition = creality_legacy_ender3
+
+[metadata]
+setting_version = 9
```
nallath:
```suggestion
setting_version = 10
```

resources/variants/creality_legacy_base_0.2.inst.cfg
nallath: Note; this should be for all profiles

nallath:
The profiles you provided are only slightly different from the fdmprinter. Are you sure that these are correct?

kazooless:
Hello Nallath,

I am sure they are working and producing significant results for the people who have tried both the fdmprinter & the legacy files. I can't tell you exactly why or which settings, but as soon as you apply the legacy profiles I created here, the expected results showed up. By trying to do the same but only with the default fdmprinter definition the print results are worse.

I've updated the 28 profiles and committed them to my copy of the profile. So should this pull request be denied and I create new pull request with the latest commit? 

Kazooless

nallath:
> I can't tell you exactly why or which settings, but as soon as you apply the legacy profiles I created here, the expected results showed up. By trying to do the same but only with the default fdmprinter definition the print results are worse.

Shouldn't that be the thing to find out then? Because if it is one setting (or a combination of a small number of them) causing it, the already existing profiles could be improved.

kazooless:
I understand the hesitation. Honestly I was hesitant about contributing because I've never done so before and I am not a coder. I have no formal training in code, etc. Plus I didn't think it would even be considered based on the discussion a while ago about profile changes in version 4.2.0 (which I see are integrated in 4.3 now, thank you for that). But Danny showed me the Twitter discussion he had with you so I promised him I'd give it a try. He posted his video about it, btw: https://www.youtube.com/watch?v=05ElvxnYZFA

To be clear, I'm not sure why the legacy definitions I created produce different results from the _fdmprinter definition_; not the creawesome definitions.  I do know why the creawesome integration has caused problems.  I've tried comparing the changes in the legacy printer definitions with the fdmprinter definitions and I'm not sure if any of those new settings are unavailable in the GUI profile panel. Can you take a look and confirm?

Regarding the addition of the "Creality3D_legacy" manufacturer, definitions & variants, it seems to me the word "legacy" should make it pretty clear what they are.

As I see it, there are two main problems we're trying to overcome. Maybe we can come up with a different solution you'd be more willing to accept and compromise on.

1.  Importing old profiles for the Creality printers is now broken 
2. The new printer definitions drastically change the behavior of the printers such that it seems impossible with them to get the previous behavior back

Regarding #1, this is because the default profile names changed to a new scheme and the old profiles are referencing names which no longer exist for the Creality definitions now. (Please correct me on terminology so I know better). It seems to me this should be fixed somehow regardless of other decisions. Users should never get an error message when importing a profile from a previous version, at least not the same major version numver (4.xxx).

Possible solutions for #1: 
     a. Add the older profile names back in, maybe side by side with the new profiles. So in addition to profiles named like "Dynamic Quality," there would also be "Draft," etc. I would do this work if I knew this would be approved. Is there a way maybe to add them back in but make them invisible to the profile manager?

     b. Introduce logic in the code which would look outside of the printer names embedded in the profile if it can't match to a variant linked to the printer definition named. This is way out of my league.

Regarding #2, I realize work has been done on the existing creawesome definitions from v4.2 to v4.3 and they're not going away. Would you consider a compromise? What if we added the legacy profiles back in but put them back in the single manufacturer, Creality3D? If the names aren't to your liking, maybe we can come up with alternatives you do like. Maybe something like "Ender-3 earlier settings" or "Ender-3 original settings" or something like that? Or maybe we go the other way and rename the creawesome printers to include "CreawesomeMod" and then put the older definitions in with the standard name? Again, I'd be willing to put this work into the project as well.












You can try to import one of these profiles to see the errant behavior: http://www.fatdragongames.com/fdgfiles/?p=4934

nallath:
I hope people are never hesitant to consider a pull request to Cura. Even if something doesn't get accepted, having the discussion will improve our and your understanding.

If you want to have a bit more insight into how / why certain settings are the way they are, I'd recommend installing the godmode plugin(https://github.com/sedwards2009/cura-god-mode-plugin). It's a real godsend in that regard.  

Importing the older profiles should work in 4.3 again, we made some changes to facilitate it.

Ghostkeeper:
Importing the older profiles works for some profiles, but not if the quality type is different. The imported profile needs to match on one of the quality types and we still give an error right now when the original quality type is not present in the new printer.

mpmel:
Is there not a way make it so that if a profile is imported based on the old definitions, it is changed to be based on the closest matching new definition, but overrides are applied to make it so that the settings are identical to what they would have been before the CreawsomeMod? This way, there is only 1 base set of definitions, but the older profiles still generate the same results.

mpmel:
Is there not a way make it so that if a profile is imported based on the old definitions, it is changed to be based on the closest matching new definition, but overrides are applied to make it so that the settings are identical to what they would have been before the CreawsomeMod? This way, there is only 1 base set of definitions, but the older profiles still generate the same results.

This has the added benefit of now showing an indicator where every setting differs from what is now the default behavior.

This could also be done as part of the profile migration on upgrade. This would help people to not be so afraid to upgrade as all of the countless hours getting their profiles right would not be lost.

I work on a large piece of software at my job and this is how we would handle a situation like this. Large functional changes always ensure the current configuration for any existing users remains the same regardless of how different the 2 versions are for a new user.

fieldOfView:
Though what you propose is an elegant solution, it would require quite a bit of rearchitecting in Cura to implement it. Since this issue currently only really affects Creality printers, I don't think Ultimaker is going to put development time into this beyond evaluating and testing pull requests.

mpmel:
I'm not sure of the details of the code of Cura completely, but while browsing a few commits I thought I already saw methods for profile migration being applied, and I thought the profiles already handled the fact that only overrides from the base are set, etc.

I could definitely be missing something, but I wouldn't think this is something to be completely discounted purely because it is "only" the Ender 3 users. That's a pretty large base of users and people testing the software for them, which also helps their direct 3D printer customers.

I am new to 3D printing, but I am a member of several groups on Facebook and Reddit with tens of thousands of users using Ender 3 and most every group is complaining about Cura, even if it is not here. Most may not even know what Github is or the fact that they can get in touch with the company this way.

You are right that nothing may happen, and Ultimaker may completely ignore this community, but in the end the exact same situation could arise in one of their own printers where the base settings should change and this will affect their user profiles as well. 

Sounds like a worthwhile improvement to me. If the users don't trust the software, they will stop using it. If Ultimaker only wanted their own printers to be supported and ignore the Ender community completely, the creawsome mod would have never made it in to begin with, and we wouldn't be talking about this.

fieldOfView:
> You are right that nothing may happen, and Ultimaker may completely ignore this community,

Please do not twist my words and suggest that Ultimaker is ignoring anyone. The Creawesome mod profiles were added to Cura because a pull-request was made. At the time Ultimaker had data that a lot (thousands) of crashes were happening from people trying to install these profiles themselves but doing so incorrectly, which made merging that PR sound like a good idea. Ultimaker engineers worked with community members to get these profiles included,, but they cannot *test* these profiles because they do not have the printers.  Ultimaker has to rely on community members for that.

I am not an Ultimaker employee. But like Ultimaker employees I have no vested interest in supporting the Creality community other than the pride I have in people using and enjoying my work. If I do this to the best of my ability, and people are still whining that they are being ignored, that does not motivate me to further do that work.

mpmel:
First, I did not twist your words. 

> I don't think Ultimaker is going to put development time into this 

I simply said you could be right. The same data exists now that a lot (thousands) of people can not upgrade to the latest versions of Cura because the profiles they have been using for a long time are no longer working properly due to a change that was made in the software. 

If you see those complaints of the software they have been trusting becoming unusable as whining and unmotivating then I can't help with that.

I personally am not affected by the changes since I am new to the software. I have no profiles that were already tuned from the old world and am able to start with the new stuff and make the necessary changes based on the things I have learned.

I was simply offering a solution to a problem that pull request was aiming to fix after someone who I believe is one of the main contributor offered up that he didn't like the current idea.

If you are not an Ultimaker employee, then you should not speak to whether Ultimaker has a vested interest in supporting the Creality community. They may not, but I can't imagine that they also would just not care that it is not working for so many people. All professional software developers I know have more pride in the products they produce than that.

I realize that Ultimaker developers worked with the community to get those PR added in and am fully aware that they rely on users to test. This is exactly why I suspect that they will want to do what they can to help the currently affected users after the change.

Regardless of whether and Ultimaker employee takes it on, or someone else does the work and submits a pull request, I was only offering up a possible solution to be discussed

nallath:
It's a lot more nuanced than caring about it working. It's not just about if the software developers care about if it works (or not), there are other factors at play here, such as providing reasons why we spend time on doing things that are not Ultimaker related. It's obviously not such that every time that we spend time on other printers that we get an angry manager telling us that we can only spend time on our own products, but at some point there will be people asking questions about where our priorities lie (and hey, Ultimaker is paying my salary, so they do get to decide where and what that time is spent on). 

Now, I do agree that these people test our software and that it has some value, but it's not as much as it could be. The only moment we started getting feedback from the people that are now mad is when we "broke" their workflow (and subsequently getting a lot of claims that no-one would ever use Cura again, they would never test anything ever again (even though they didn't in the first place), etc.). As far as I can see from the stats, only a few people are lagging behind with upgrading. A bit more than usual, but not any mind shockingly large numbers. 

The change required to fully import the old profiles in exactly the same way would, as fieldofview mentioned, required large architectural changes, which would cost a significant amount of work / resources (Which, as you might imagine, makes it super hard to "sell" it, especially since we don't get that much feedback from those users in the first place!).

mpmel:
I'm aware there's more to it than that. I have been a software developer professionally for 15 years.

My comment was only that I believe that you cared enough that you would not just ignore the fact that there are issues and were willing to at least work towards making it better, which I understand may have been done a little between 4.2.1 and 4.30. Again, I don't know, because I wasn't affected by it.

If it's too much work and not worth it to you or anyone else, then that's fine. It was only a suggestion. I only offered it up because ghostkeeper, who I assumed was heavily involved with the project and maybe an employee, suggested that the current suggestion should not be accepted and suggested that Cura be forked instead.

I can't speak to those users and what they did previously or what they claim to do in the future. I know a lot of users of open source software are very demanding and treat the developers like they are owed something, etc. I get it. I've followed some large OS developments for years like Blender, GIMP, etc and I see it everywhere. But for every one of those there are usually a few people who are willing to help as much as they can.

Personally, If I wanted to use the old profiles with the new version (which I have done) I just downloaded Cura 4.0 side-by-side and had the two versions open on each monitor and went down the entire list and made sure every setting is identical between the two. I don't know if that gets me 100% there though. I don't quite understand the full profile system yet apparently.

I think you should take the lack of feedback and even the outrage as testament that you have a good piece of software that works well. You usually won't hear from most people using open source software until something breaks.

mpmel:
I actually didn't know there was an /r/Cura....but I'll check it out :wink:

I only own a single Ender 3 Pro, so I am mostly lurking in Ender 3 specific places like /r/Ender3, etc

kazooless:
@mpmel 

>> I don't know if that gets me 100% there though.
> Since we also make changes to CuraEngine, you'll not always be able to get exactly the same slice even if you have all of the settings exactly the same.

It's also from the printer definitions. As best as I understand it, not everything in the printer definition is configurable through the GUI. That is actually the reason for this pull request.

> But for every one of those there are usually a few people who are willing to help as much as they can.

I am one of those few, but since I've been able to understand this type of code and since I benefit from this software, I am willing to put work back into it. I have some unanswered questions above though and I'm hoping to get them answered before I put some of that effort into it. I'd rather have the discussion and the tentative approval before I put in the effort. This request is a case in point; effort made but request denied.

For reference the questions I asked are in the last comment I posted here.

resources/variants/creality_legacy_base_0.2.inst.cfg
kazooless: @nallath I have several people in one of the FB groups where I shared these files telling me they're getting errors when they use the files with "setting_version = 10" in them. If they put the older files in with "setting_version = 9" back, then the errors go away. They tell me this is with Cura 4.3.0.

The error is that the nozzle size is not available in the materials drop down. It just says empty. One person emptied the cache and said it now shows nozzle .4mm but doesn't give a drop down to choose anything else. It seems most of these people are Windows 10 and one is Linux. Mac users haven't reported this problem, and I am working on a Mac myself.

Can you tell me what to look into to troubleshoot this problem please? Also, how should I have learned about the versioning difference so in the future I can get that right before submitting any pull requests.

Ghostkeeper: > @nallath I have several people in one of the FB groups where I shared these files telling me they're getting errors when they use the files with "setting_version = 10" in them. If they put the older files in with "setting_version = 9" back, then the errors go away. They tell me this is with Cura 4.3.0.

That's correct. In Cura 4.3 the setting version was 9. In Cura 4.4 it'll be 10. If it's v9, Cura 4.4 will not load it, only 4.3 will. If it's v10, only Cura 4.4 will read it. These profiles are submitted for Master which is next going to be published as Cura 4.4, so it should be v10.

The difference between v9 and v10 is that stacks must now have intent profiles, several profiles have been renamed and that the Alternate Skin Rotation setting can no longer be overridden in profiles. For a complete reference of the renamed profiles, see [this file](https://github.com/Ultimaker/Cura/blob/master/plugins/VersionUpgrade/VersionUpgrade43to44/VersionUpgrade43to44.py). As long as you're not using Alternate Skin Rotations, just updating the `setting_version` to 10 should be fine for you.

> Also, how should I have learned about the versioning difference so in the future I can get that right before submitting any pull requests.

Normally, with pull requests to Master, programmers tend to expect that you've tested your code in Master as well. You would've seen that it didn't load immediately then, and why. We understand that profile designers don't always have the expertise to run Cura from source and then run from the release of 2 months ago, but we haven't really found a way to deal with that yet. Usually it's not really a problem... :confused:

Note that internally the Cura developers haven't agreed yet whether or not this should be merged in. On one hand it would make it easier for profile designers who created custom profiles before May 2019 to get them to use those again. On the other hand we don't want to confuse new users with which definition to use, prevent subjective opinions on which is better, and to open up the possibility of having different definitions for the same printer depending on a person's preferred workflow.

kazooless: @nallath 

> These profiles are submitted for Master which is next going to be published as Cura 4.4, so it should be v10... Normally, with pull requests to Master, programmers tend to expect that you've tested your code in Master as well.

Okay, that makes a lot of sense. Thank you so much.

> Note that internally the Cura developers haven't agreed yet whether or not this should be merged in. 

I completely understand. I want to help where I can and I can appreciate the two sides of the issue here. I think the heart of the matter is that there isn't any way to change all of the settings from within Cura itself so the average user is stuck with a choice of using the old version and missing out on the new features or using the new version and missing out on the print quality they were once able to get. If all of the settings found in the definitions files were available in the GUI, then it would be easy enough for those more familiar with the program to make tutorials and such. Since that isn't a possibility right now, I'm interested in finding a solution for that typical user so they won't need to make that choice. For me, I've figured it out for myself and don't "need" anything further.

So for now, I'll wait on you and the other developers to come to a consensus and if there is something I can do afterwards to contribute, then I'll do so. You might have other ideas I haven't thought of and I'm all ears if you do. Maybe you'd be okay with adding just one extra single printer under the existing manufacturer with a name that makes it clear it is for a generic Creality printer with the old settings and for advanced users. The bed size would have to be edited but that is easy enough in the GUI. Maybe a readme about it can be put in as comments in the start gcode. 

Thanks again for your help and explanations. 

PS Is there a list that tells you which settings for the definitions file are only available in the files and not in the GUI?

Ghostkeeper: Yeah a list is [here](https://github.com/Ultimaker/Cura/blob/66005f28f31c343aff697d22ecdbc1d06a6e86c2/resources/definitions/fdmprinter.def.json#L27). We explicitly hide the `machine_settings` category of settings and the `command_line_settings` category. The latter is not very interesting if you have a GUI. The former can be made visible by installing the "Printer Settings" plug-in from fieldOfView.

FatDragonGames:
I own Fat Dragon Games, a company that sells .stl models (www.fatdragongames.com). We've been recommending kazooless's 'legacy' Creality printer definitions for our customers using Cura 4.3 (over 29k subscribers on our Youtube channel and 9k on our FB group) and have gotten only positive feedback on it. The Creawesome settings do not produce good results on tiny models such as the ones we sell (fantasy gaming miniatures) but the legacy files have given all of us back the great quality Cura had always delivered up thru 4.0. Please consider adding these legacy files as an option in future versions of Cura.

fieldOfView:
Perhaps the new "intent" feature could be used to make a single unified Creawesome definition, where the user can choose between Creawesome and Custom (legacy).

Ghostkeeper:
The "intent" is meant for the user to choose what he'd like to use his print for. It can technically be used for other things, but it's a bit like defining "material" profiles that allow the user to choose between different types of build plates. Can be done, but it could be confusing. The intent categories available are also pre-defined since they need to be translated.

In my opinion ideally we'd identify the changes between the new and the old profiles that cause the print quality to be worse and fix those. Reverting to the old profiles is a judgement call that causes a big division among our users, vocal ones and non-vocal ones. We should move forward only. There is a discussion about this here: https://github.com/Ultimaker/Cura/issues/6106
In that topic we've applied all the changes that at least 2 people agreed on. That was a pleasant and productive discussion as far as I'm concerned! But more testing is always needed to bring the profiles further.

As for this though, we've decided to not add any secondary set of profiles for the same printers. Sorry! We're trying to reduce the divide here.

rcpoison:
@kazooless thank you so much for this.
I've replaced my extruder, belts, rollers, tubing, nozzle and tried calibrating everything countless times because I couldn't figure out why my prints went from absolutely perfect to mostly failed and utter garbage.
I was just about to ditch my Ender 3 and buy a different printer.
Turns out it wasn't my hardware, it was the profile. 
This fixed everything.

PR #6503, opened 2019-10-08
BeamUp S definition, extruder, mesh and quality files added

beamup3d:


konskarm:
Tests are failing because there is an ID field in the beamup_s_extruder_0.def.json. Please have a look in the test results and remove this field.

beamup3d:
> Tests are failing because there is an ID field in the beamup_s_extruder_0.def.json. Please have a look in the test results and remove this field.

Thanks. The "id" field removed.

nallath:
The build volume is going right through the build plate model. Is that intentional? 

Also, the name of the printer that it's in the model is almost impossible to read. So you either want to make it a bit deeper (so it can actually be read) or remove it altogether (to decrease the size of the model) 
Buildplate

Ghostkeeper:
Look into the `platform_offset` metadata entry to adjust the position of the build plate.

beamup3d:
Platform model and "platform_offset" updated.

Ghostkeeper:
That new platform model is stored very inefficiently and exceeds our target of 1MB per mesh. It has 69730 faces but most of them are completely flat. There are actually only 313 planes in the mesh (616 triangles). I've compressed the model losslessly with Blender now and that reduces it to 100kB!

The printer loads well, and the build plate is all nicely aligned now. Thanks for your contribution!

beamup3d:
I'll keep in mind this detail. Thank you for explanations. We are excited to be a part of Cura :)

PR #6533, opened 2019-10-14
CURA-6863 duplicated material oddness

LipuFei:


rburema:
> Why is ViewOrientationControls relevant to "duplicate material oddness"?
> @bremco this my only comment, looks good otherwise

Hi @ninovanhooff,

That's because master was merged into this branch :-)
See @LipuFei 's first commit in the list of commits above.

P.S. @bremco is my personal account. Since I usually don't have my personal mail open during work hours, it's best to use @rburema instead!

cheers,
Remco.

LipuFei:
Fixed merge conflicts

PR #6537, opened 2019-10-15
CURA-6862_truncate_profile_names

ninovanhooff:
Fix some long texts extending their bounds

### DIFF
--- a/resources/qml/PrintSetupSelector/Custom/CustomPrintSetup.qml
+++ b/resources/qml/PrintSetupSelector/Custom/CustomPrintSetup.qml
@@ -4,6 +4,7 @@
 import QtQuick 2.10
 import QtQuick.Controls 2.3
 import QtQuick.Controls 1.4 as OldControls
+import QtQuick.Layouts 1.3
 
 import UM 1.3 as UM
 import Cura 1.6 as Cura
@@ -66,7 +67,6 @@ Item
         {
             id: intentSelection
             onClicked: menu.opened ? menu.close() : menu.open()
-            text: generateActiveQualityText()
 
             anchors.right: parent.right
             width: UM.Theme.getSize("print_setup_big_item").width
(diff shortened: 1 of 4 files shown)


resources/qml/PrintSetupSelector/Custom/CustomPrintSetup.qml
```diff
@@ -75,62 +75,94 @@ Item
...
-                var resultMain = resultMap["main"]
-                var resultSuffix = resultMap["suffix"]
-                var result = ""
 
-                if (Cura.MachineManager.isActiveQualityExperimental)
+                Label
                 {
-                    resultSuffix += " (Experimental)"
+                    id: textLabel
+                    text: qualityName()
```
rburema: Keep in mind that this isn't a binding anymore!

resources/qml/PrintSetupSelector/Custom/CustomPrintSetup.qml
```diff
@@ -75,62 +75,94 @@ Item
...
+                    Layout.maximumWidth: parent.width * 0.7
+                    height: contentHeight
+                    verticalAlignment: Text.AlignVCenter
+                    renderType: Text.NativeRendering
+                    elide: Text.ElideRight
+
+                    function qualityName() {
+                        var resultMap = Cura.MachineManager.activeQualityDisplayNameMap
+                        return resultMap["main"]
+                    }
```
rburema: You could also write this as `Cura.MachineManager.activeQualityDisplayNameMap["main"]`, then it can be back to be a binding.

Also, if the function is this small, and only used in one place, you can also put the brackets directly where the value was. So instead of
~~~
thing: blah()
function blah() { ... bloeh... }
~~~
you just get:
~~~
thing: { ... bloeh ... }
~~~

resources/qml/PrintSetupSelector/Custom/CustomPrintSetup.qml
```diff
@@ -75,62 +75,94 @@ Item
...
+                Label
                 {
-                    resultSuffix += " (Experimental)"
+                    id: textLabel
+                    text: qualityName()
+                    font: UM.Theme.getFont("default")
+                    color: UM.Theme.getColor("text")
+                    anchors.verticalCenter: intentSelection.verticalCenter
+                    Layout.margins: 0
+                    Layout.maximumWidth: parent.width * 0.7
```
rburema: Probably best to document where the `0.7` comes from.

fieldOfView: and add a Math.floor() around it.

PR #6541, opened 2019-10-16
turn off Union overlapping volumes when Surface mode is Surface only

BagelOrb:
See https://github.com/Ultimaker/CuraEngine/pull/1150

When using surface mode to directly determine the print paths you generally don;t want to union overlapping volumes.

At least that is not how I would want to use it.

I'm not 100% sure about this (super small) PR.


rburema:
Devs: See internal ticket CURA-7080.

PR #6550, opened 2019-10-17
Return early from _checkStackForErrors() if top container is empty.

smartavionics:
Checking the whole stack can be time consuming and we don't need to do it again if there are no per-object settings that could possibly cause errors.

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -107,7 +107,7 @@ class StartSliceJob(Job):
             return False
 
         # if there are no per-object settings we don't need to check the other settings here
-        if stack.getTop() == None or len(stack.getTop().getAllKeys()) == 0:
+        if stack.getTop() is None or not stack.getTop().getAllKeys():
             return False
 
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -106,6 +106,10 @@ def _checkStackForErrors(self, stack: ContainerStack) -> bool:
         if stack is None:
             return False
 
+        # if there are no per-object settings we don't need to check the other settings here
+        if stack.getTop() == None or len(stack.getTop().getAllKeys()) == 0:
```
nallath:
```suggestion
        if stack.getTop() is None or not stack.getTop().getAllKeys():
```
Comparison with none should be done with is, and an an empty list also evaluates to false.

smartavionics:
Thanks for the suggested change.

Ghostkeeper:
Developers, this is ticket CURA-6902.

Ghostkeeper:
Yup, error checking still seems to be correct but it shaves off about half a second per model to start a slice (according to the log entry about sending slice message).

PR #6565, opened 2019-10-21
Marlin meshes marking post-processing script

LinFor:
This is a simple script to mark different objects (meshes) with new-introduced Marlin rules. This adds ability to cancel objects during print job (for example, if it fails to print or something bad happened).

Here is feature PR in Marlin with details of realization: https://github.com/MarlinFirmware/Marlin/pull/15590


Ghostkeeper:
This is something I'd rather see implemented in CuraEngine, actually. Since Marlin defines a standard and it's the first standard I know of to do this, we should try to follow it. Opinions?

LinFor:
It would be nice.

But if you look at the problem a little wider, it would be ideal to separate directly the machine’s action planning from translating the resulting plan directly into g-code.
In this case, the slicer could note the fact of the beginning / end of the object (as it happens now), and when translating to the g-code, a direct decision would be made on how to reflect this.
Of course, there is FffGcodeWriter, but its implementation is limited to fixed behavioral "models" of machines, while within the same model specific capabilities may or may not be present. As a result, both models (machine_gcode_flavor, Volumetric) and additional parameters (machine_firmware_retract, relative_extrusion) multiply, although even existing parameters do not allow you to get ready-made code taking into account the actual firmware capabilities.
I recently ran into this while adapting machine_firmware_retract. Modern Marlin can do Z-hop with a firmware retract, but Cura still inserts separate G0 Z commands for this.
Perhaps we could give the opportunity to implement our custom FffGcodeWriter in python in relation to the machine definition?

Ghostkeeper:
You're touching there on the question of whether the g-code should be produced on the printer, which has more information on its own inner workings, or in the slicer, which has more information on what the user intends to use the print for. To resolve this conundrum, so far we've been moving as much information about the printer into the slicer as we can and then let the slicer make the decision based on the combination of what we know about the printer (printer definitions) and what the user wants (e.g. smooth surfaces free of blobs vs. faster printing: Z hop or not).

LinFor:
> To resolve this conundrum, so far we've been moving as much information about the printer into the slicer as we can and then let the slicer make the decision based on the combination of what we know about the printer (printer definitions) and what the user wants (e.g. smooth surfaces free of blobs vs. faster printing: Z hop or not).

A bit wrong, I mean a little different:
I mean, the final behavior of the machine will not change, no matter how we translate our plan into g-code. If we want to do retract with z-hop, we can:
1. Don't use fwretract at all: G1 E-n + G1 Z-m
2. Use fwretract (current state): M207 Sn Z0 ...... G10 + G1 Z-m
3. Fully use [fwretract](http://marlinfw.org/docs/gcode/M207.html): M207 Sn Zm ..... G10

In all these cases there will be a retract with a z-hop. I prefer to maximize the use of firmware features such as a fwretract, as it is possible to tune the values on the fly using the LCD display.
At the moment, the last option is not available, because Cura forcibly inserts G1 Z-m after G10.

But, probably, it's a new feature and I can made PR into CuraEngine?

Ghostkeeper:
We're going to close this for the aforementioned reasons. The default responsibility of generating g-code lies with CuraEngine.

PR #6566, opened 2019-10-21
Change SDK version to 7.0.0

ninovanhooff:
CURA-6858

Companion PRs

https://github.com/Ultimaker/Uranium/pull/534
https://github.com/Ultimaker/CuraPrintProfileCreator/pull/36

### DIFF
--- a/plugins/AMFReader/plugin.json
+++ b/plugins/AMFReader/plugin.json
@@ -3,5 +3,5 @@
     "author": "fieldOfView",
     "version": "1.0.0",
     "description": "Provides support for reading AMF files.",
-    "api": "6.0.0"
+    "api": "7.0.0"
 }
--- a/plugins/TrimeshReader/plugin.json
+++ b/plugins/TrimeshReader/plugin.json
@@ -3,5 +3,5 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.0",
     "description": "Provides support for reading model files.",
-    "api": "6.0.0"
+    "api": "7.0.0"
 }
--- a/plugins/UFPReader/plugin.json
+++ b/plugins/UFPReader/plugin.json
@@ -3,6 +3,6 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.0",
     "description": "Provides support for reading Ultimaker Format Packages.",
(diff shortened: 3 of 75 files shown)


plugins/AMFReader/plugin.json
```diff
@@ -3,5 +3,5 @@
     "author": "fieldOfView",
     "version": "1.0.0",
     "description": "Provides support for reading AMF files.",
-    "api": "6.0.0"
+    "api": "7.0.0"
```
LipuFei: To keep it consistent with the others, I'd change this to `7.0`

plugins/TrimeshReader/plugin.json
```diff
@@ -3,5 +3,5 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.0",
     "description": "Provides support for reading model files.",
-    "api": "6.0.0"
+    "api": "7.0.0"
```
LipuFei: To keep it consistent with the others, I'd change this to `7.0`

plugins/UFPReader/plugin.json
```diff
@@ -3,6 +3,6 @@
     "author": "Ultimaker B.V.",
     "version": "1.0.0",
     "description": "Provides support for reading Ultimaker Format Packages.",
-    "supported_sdk_versions": ["6.0.0"],
+    "supported_sdk_versions": ["7.0.0"],
```
LipuFei: To keep it consistent with the others, I'd change this to `7.0`

plugins/AMFReader/plugin.json
nallath: It must be a semver, so it's better to have it like 7.0.0

plugins/TrimeshReader/plugin.json
nallath: Same as above, 7.0.0 is correct semver, 7.0 (even though we accept it) is technically not.

plugins/UFPReader/plugin.json
nallath: Same as above, 7.0.0 is correct semver, 7.0 (even though we accept it) is technically not.

PR #6569, opened 2019-10-21
Xray in solid view

BagelOrb:
EDIT:
Please credit @fieldOfView @smartavionics and me (@BagelOrb ) when this PR is incorporated into Cura. This PR is a nice collab :)

------------------

X-ray view shows problems in your model which can lead to nasty print results which can be difficult to spot.

Currently there is little indication that there is something wrong and it is very hard for users to find out why their prints are coming out ugly.
This is exemplified by the countless number of issues on reddit which have to do with non-solid models. https://www.reddit.com/r/3Dprinting/search?q=(non-manifold)%20OR%20(mesh-repair)%20OR%20(non-solid)%20OR%20(non-euclidean)%20OR%20manifold&restrict_sr=1&sort=new

Please don't say "There is no demand for this feature." because the whole problem is a problem of discoverability; users currently don't know where the problem comes from because we never show them the problem.

I've removed the X-ray view and incorporated it into the normal SolidView.
I've changed the problem color to black so that it doesn't collide with the overhang color. (I've chosen black because it really stands out; also in the dark themes it stands out more to black filament compared to white on white filament)
I've fixed a bug in the X-ray view which caused some faces not to be colored with the problematic color.

image

Here's a nasty model you can test on: 
[Debailey_x10.stl.tar.gz](https://github.com/Ultimaker/Cura/files/3751941/Debailey_x10.stl.tar.gz)

I'm not 100% sure whether the naming still makes sense because now we have an xray_composite shader in the SolidView, rather than it being called "solid_view_composite.shader".


BagelOrb:
(Not sure why Nino's commits show up in here. Only the last 3 commits are actually in the PR.)

BagelOrb:
See also CURA-3230 in JIRA

BagelOrb:
Perhaps a static noise is more alarming:
image

fieldOfView:
I think the noise looks like a rendering error.

Every so often I see the question "why is my model showing red faces?" and I have to explain that that is overhang. The red looks like an error. I would suggest using red for actual model errors (ie: this X-Ray feature) and choosing a different color for overhang. A legend like the layer view has it would further clarify what is going on in the view.

BagelOrb:
But then the color doesn't have a lot of contrast with the model color.

That's also why I explored pure black; it's in stark contrast to any plastic color (even black which is just dark grey)

Now I decide on the color of the error based on the lightness of the underlying pixel.
What do you think?
image

Ghostkeeper:
After some discussion with the team, we've concluded that the latest solutions would also be very confusing for an unsuspecting user to see. You're loading in a model and instead of the usual yellow you see some darker shades as well. What do you do? This could be some sort of overhang detection (which has a similar problem), or other problems with the model, but you're not told what it is. Without knowing what this is, they'd probably print it anyway.

So we came up with the idea to use a solid (instead of the varying shades technique, more similar to your first solution) but to display an image instead as if it shines through. One that shows something that informs the user that the model is not manifold. Would that be a nice idea to implement instead?

BagelOrb:
I wished I would have received this feedback earlier - when I was actually working on this. Now it's hard for me to get back into the code. Alas.

Let's see. I should have time in a month or so.

BagelOrb:
image
Muy bueno, no?

BagelOrb:
image
There was too much detail in the image.

BagelOrb:
image

I found it difficult to decide on an image. First of all it needs to be international, because it's way too much trouble to generate an image for each language. So it needs to be an icon. Then it also needs to be a small icon because it should be visible for small geometry. Then we also need contrast because it needs to be very visible against any printable color.

I also put a cross in the background to reduce the chance of a small error region totally being covered by the background of the icon, which could cause it to be totally invisible on a red print.

Take all of those considerations and you get what you see above.

rburema:
Hi @BagelOrb 
Just discussed this with @Ghostkeeper this morning (as by coincidence, I picked up this PR at that time). The new proposal looks excellent! Reviewing & testing it now.

BagelOrb:
Here's a less extreme example, showing the problems of visualizing error areas which are very small:
image

[ultirobot_low_poly.stl.zip](https://github.com/Ultimaker/Cura/files/4018973/ultirobot_low_poly.stl.zip)

rburema:
Thanks.

Would you be OK if I undelete the XRay view plugin? (I know it has some duplicate code in that case, that's why I'm asking.) Ruben says it has some advantages not related to model-errors that it can be used for. Also now there is only one possible view in 'preview' that still has its own drop down menu. (Thirdly users expect it there, but that's less important IMHO.)

BagelOrb:
Let me list all of the reasons I can think of to remove the plugin.

I think it's a classical programmers mistake to keep all options open. There is no real need for the xray view plugin, so let's keep the interface clean!

If people really want to use the xray view for some reason then they can use an old version of Cura.

That dropdown was the weirdest place for the xray view to begin with. When we redesigned the interface to separate [PREPARE] and [PREVIEW] the xray view suddenly had no real place to go. We can finally solve that! :D

Also if you keep the plugin then there will be code duplication or some complexity to prevent the code duplication. You wil need to make sure that the fixes I did to the xray view are also ported to the plugin and likewise for any future improvements.

In Cura 15.04 there was also a transparent view which could be used to identify problems in the mesh, but nobody ever asked for it and nobody seems to need it. The same can hold for the xray veiw now that it is incorporated. Cura 15.04 screenshot showing a superfluous face:
image
[extra_face.stl.zip](https://github.com/Ultimaker/Cura/files/4019028/extra_face.stl.zip)

For these reasons I think it is a good idea to remove the old plugin.

Do you want me to figure out how to remove the whole dropdown?

rburema:
You make some good points.

> Do you want me to figure out how to remove the whole dropdown?

It's probably easier to 'not have a dropdown if there's only one item, otherwise do'. Just in case a plugin uses it ;-) Also the structure is re-used from the other views so you cant really remove all drop-down functionality (at least that's what I think I know from the top of my head).

For now I'm going to test just a bit more (it's still visible in PLA/red at least for example -- just a bit less), maybe merge it, then discuss it with Yi-An when she's back either way.

BagelOrb:
The xray view *does* offer functionality for looking inside the model which is made impossible by the new xray incorporated solid view. 

xray:
image

solid:
image

unsliced preview gives me something better:
image


I could see my internal geometry using the xray view, but it's not the best in doing that:
image

Neither is the unsliced preview:
image

plugins/SolidView/xray_composite.shader
```diff
@@ -110,6 +122,13 @@ fragment41core =
 
     float kernel[9];
 
+    float hash12(vec2 p)
```
rburema: Waar gebruik je deze functie(s)?

plugins/SolidView/xray_composite.shader
```diff
@@ -39,6 +40,13 @@ fragment =
...
```
rburema: Waar gebruik je deze functie(s)?

BagelOrb: In een oude commit. Ik heb ze verwijderd.

rburema:
Hi @BagelOrb 
We just had a discussion about this with the team, and we now think it's better to go the route of the MeshFixer plugin: Have a popup when the mesh isn't manifold, and have to option to fix them (which is avialable in the trimesh library, which we already include anyway).

The problems with this approach are:
 - the UX:
   . clearness for the user: a.k.a. "what is actually going on": both in the sense that there is no label to what the 'x /!\ x' image means (yet), and the combined interpretation of the image-and-label which needs to also convey to the user _what_ exactly is wrong (as opposed to only _where_). Keep in mind that not everyone is as comfortable as we are in figuring out what's going on from a projection (we're so used to it that we often don't even notice where users struggle).
   . further UX (re)design required), if we where to go this route, we'd probably need Yi-An to have a deep look at this, both the changes to the prepare mode, but also the suddenly missing X-Ray view and what needs to happen to that menu.
- the ability of the user to _directly_ fix the problem there-and-then is absent in this approach 'as written'
- difficulties with avoiding code-duplication since we _do_ want to keep the old x-ray view available as well

As such, I propose we close this (with a bit of sadness because it looked cool and you put a lot of work into it).

fieldOfView:
:cry:

I'll probably steal the concept for a future version of Sidebar GUI plugin, which already brings the XRay view to the Prepare tab.

BagelOrb:
Your arguments not to use my xray overlay hold even more so for the current xray view. It's a bit weird to say that the xray overlay is unclear, but to want to keep the xray view. What does the xray view have that the xray overlay does not have?

As a user you don't know when you should open the xray view, because you have no idea whether your model is problematic or not. The user should be informed as early as possible and not when the user decides to go to the xray view. 

Sure it would be nice if there would also be a pop-up informing the user that there are geometry problems (like Aldo's plugin does), and to offer the user to try to fix the model automatically, but those are future PRs and shouldn't have to be included in this one.

BagelOrb:
Note that the mesh analysis and MexhFixer might not be perfect. As a user I would like to see what Cura thinks is wrong with my model and either ignore it if the problem areas are very small or all perfectly horizontal. Even still as a designer I would want to know what is wrong with my model instead of having it automatically fixed. I would want to make my future models good so that they will never need a MeshFixer.

Ghostkeeper:
Just my opinion here since there seem to be some disagreements.

The red `x /!\ x` things are pretty clear to indicate that something is wrong. Just not what is wrong in particular, but it's better than the current situation where nobody bothers to go into X-ray mode unless they are told to by the developers after reporting a bug.

Being able to see the insides of the model in X-ray mode is useful though, so to keep that behaviour we need to keep the X-ray view itself. And this makes the issue about code duplication a valid one.

BagelOrb:
@fieldOfView Is it possible to incorporate this PR into your mesh tools plugin?

Bit of a long shot...

BagelOrb:
Xray veiw erros are now a bit transparent:
image

A warning is shown:
image

There's a preference to show errors / messages:
image

BagelOrb:
All code duplication has been solved.

plugins/SolidView/SolidView.py
```diff
@@ -98,6 +139,50 @@ def beginRendering(self):
...
+
+                self.getRenderer().addRenderPass(self._xray_pass)
+
+                if not self._composite_pass:
+                    self._composite_pass = self.getRenderer().getRenderPass("composite")
+
+                self._old_layer_bindings = self._composite_pass.getLayerBindings()
+                self._composite_pass.setLayerBindings(["default", "selection", "xray"])
+                self._old_composite_shader = self._composite_pass.getCompositeShader()
+                self._composite_pass.setCompositeShader(self._xray_composite_shader)
```
Ghostkeeper: I think it's better to do all this in `self._checkSetup()`, right?

resources/shaders/xray_composite.shader
```diff
@@ -50,10 +51,12 @@ fragment =
 
         result = layer0 * layer0.a + result * (1.0 - layer0.a);
 
-        float intersection_count = (texture2D(u_layer2, v_uvs).r * 255.0) / 5.0;
-        if(mod(intersection_count, 2.0) == 1.0)
+        float intersection_count = texture2D(u_layer2, v_uvs).r * 51; // (1 / .02) + 1 (+1 magically fixes issues with high intersection count models)
```
Ghostkeeper: Why the 0.02 then?

Ghostkeeper:
I'm not seeing any big hurdles here. Maybe I'll discuss these changes tomorrow with a colleague to see if he agrees and we'll just apply these changes ourselves. Except the "0.02", which is just an undocumented magic number. I don't really know what to do with that one. Seems a bit weird. It's hiding the effect when the number of intersections is too high.

plugins/SolidView/SolidView.py
```diff
@@ -175,4 +260,49 @@ def beginRendering(self):
...
+            self._next_xray_checking_time = time.time() + self._xray_checking_update_time
+
+            xray_img = self._xray_pass.getOutput()
+            xray_img = xray_img.convertToFormat(QImage.Format.Format_RGB888)
+            ptr = xray_img.bits()
+            ptr.setsize(xray_img.byteCount())
+            reds = np.array(ptr).reshape(xray_img.height(), xray_img.width(), 3)[:,:,0]  # Copies the data
+            bad_pixel_count = np.sum(np.mod(reds, 2)) # check number of pixels with an odd intersection count
+
+            if bad_pixel_count > 10: # allow for 10 pixels to be erroneously marked as problematic
```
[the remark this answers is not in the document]
BagelOrb: I thought it might be possible that there are ever so tiny creases or perhaps there are some rounding errors in the GPU which might cause false negatives.

I'm okay with this being set to a different hard coded value...

fieldOfView:
Does it have to be this ugly though?

Ghostkeeper:
Would you like to redesign the icon?

plugins/SolidView/SolidView.py
```diff
@@ -27,12 +49,31 @@ def __init__(self):
...
         self._global_stack = None
 
+        self._old_composite_shader = None
+        self._old_layer_bindings = None
+
+        self._next_xray_checking_time = time.time()
+        self._xray_checking_update_time = 1.0 # seconds
+        self._xray_warning_cooldown = 60 * 10 # reshow Model error message every 10 minutes
+        self._xray_warning_message = Message(catalog.i18nc("@info:status", "Your model is not manifold. The highlighted areas indicate either missing or extraneous surfaces.")
+                                             , lifetime = 60 * 5 # leave message for 5 minutes
```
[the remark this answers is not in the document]
BagelOrb: The only reason I set a finite timer to it was so that it reshows itself after five minutes when you move the camera. That reshowing can draw more attention to it.

I think the user really needs to se this message. He will continually be reminded of it until he expressly sayd "Do not show again".

Ghostkeeper:
I don't think it really needs to be pretty. It's only shown if the model is broken after all. But better is better.

plugins/SolidView/SolidView.py
BagelOrb: You might be right there. I don't know about that function. I haven't considered it while writing the code. It didn't figure in the code for the xray view, so I haven;t looked at it.

fieldOfView:
I am not convinced an icons is needed/wanted. Has a UX designer had a look at this?

I would much prefer a more holistic revamp of the look of meshes in the viewport, including the look of the overhang simulation.

fieldOfView:
Wouldn't it be more efficient (less code to maintain) to use TriMesh node.is_watertight instead of relying on the view to have the proper orientation to see the defect and doing readbacks from the GPU?

resources/shaders/xray_composite.shader
BagelOrb: 0.02 is the base color which is added onto a pixel for each face behind that pixel.
It's defined in xray.shader

BagelOrb: About the magic formula:
I have no idea, but 255/5=51
I don't understand why the xray works the way it does, but at least now it says it's magic in a comment, rather than being undocumented.

BagelOrb:
Doesn't that function have problems with geometry such as this:
image

Ghostkeeper:
As I was going to start work on something else tonight I forgot I had this branch checked out. But it crashed upon start-up:

```
  File "/home/trin/Gedeeld/Projects/Uranium/UM/Qt/Bindings/MainWindow.py", line 240, in _render
    view.endRendering()
  File "/home/trin/Gedeeld/Projects/Cura/plugins/SolidView/SolidView.py", line 272, in endRendering
    reds = np.array(ptr).reshape(xray_img.height(), xray_img.width(), 3)[:,:,0]  # Copies the data
ValueError: cannot reshape array of size 10424576 into shape (1376,2525,3)
```

I tried merging master into the branch (in case Uranium had gotten ahead or something) but it still crashes the same.

Ghostkeeper:
> I am not convinced an icons is needed/wanted. Has a UX designer had a look at this?

She was there when we came up with the idea, yes. She agreed it was the best intuitive thing we could come up with and that it was better than a flat black colour or the noise that was shown here earlier.

> Wouldn't it be more efficient (less code to maintain) to use TriMesh node.is_watertight instead of relying on the view to have the proper orientation to see the defect and doing readbacks from the GPU?

It would probably be more efficient. I want to see if the impact here can be felt at all though. I'm not fussed about the load on the GPU but more on Numpy. But I've dealt with much bigger problems in Numpy before that then manage to run in a dozen milliseconds, so maybe it'll work.

On reliability, I don't know. I've seen Trimesh give many false positives, which is also pretty annoying. In particular it struggles with coincident edges (like BagelOrb's example above but also just with 2 cubes diagonally adjacent on an edge). CuraEngine by design stitches up holes smaller than 5 micrometres too. Most models are going to have such cases. I like the MeshTools plug-in a lot (especially for its break-apart-into-pieces function) but it's complaining in about 30% of the models I create with OpenSCAD.

Of course I don't like that it depends on the camera position and the resolution of your monitor either. In the end I see this as a "best effort" from Cura's side to warn the user that they did something wrong before they entered Cura's stage. Certainly an improvement.

plugins/SolidView/SolidView.py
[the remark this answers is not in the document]
Ghostkeeper: Leaving this as-is for now until testing.

plugins/SolidView/SolidView.py
Ghostkeeper: I've moved it into that function: a4fbf3c613bfe8f91b1bb63b472e2f9ff22b69bb

plugins/SolidView/SolidView.py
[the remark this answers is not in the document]
Ghostkeeper: I think that you originally had to set a limit but not due to rounding errors. You were taking data directly from the pixel buffer of the image, but his pixel buffer has padding bytes to align each column of pixels to word sizes. These padding bytes are not initialised, so they contain random noise. These could be recognised as bad pixels. When I fixed that, the noise was completely gone. Normally the rounding errors in triangle coordinates don't cause gaps to appear between the triangles; otherwise you'd see gap pixels in every 3D model.

resources/shaders/xray_composite.shader
```diff
@@ -50,10 +51,12 @@ fragment =
...
+        float rest = mod(intersection_count + .01, 2.0);
+        if (rest > 1.0 && rest < 1.5 && intersection_count < 49)
         {
-            result = u_error_color;
+            vec2 scaling = textureSize(u_layer0, 0) / textureSize(u_xray_error, 0);
```
rburema: Note that `textureSize` was added to GLSL 1.30, which is part of OpenGL 3.0 ... thus it'll probably not work on OpenGL 2.1 systems (which is what the not-core-4.1 shaders use)!

BagelOrb: That's a lot of different version numbers!

What exactly needs to be done?

fieldOfView: An alternative to using textureSize() needs to be found, such as eg passing the texturesize as a uniform.

BagelOrb: Ok. I don't have time for that atm.

fieldOfView: It would probably be more performant anyway to pass this scaling factor as a uniform once per frame instead of recalculating it for every pixel.

fieldOfView: @BagelOrb, I have made a fix on top of this branch, here: https://github.com/Ultimaker/Cura/pull/7363

fieldOfView:
I think that having the "Don't show this message again" checkbox leading to not showing model errors at all anymore is counter-intuitive, and will lead to many people turning off the feature getting us back to where we started. Perhaps there should be two preferences; one for visualizing the model errors, and one for showing a message if model errors are detected. I can do a PR on top of this PR if people agree.

resources/shaders/xray_composite.shader
fieldOfView: @smartavionics could you test this branch (and my PR) with your Raspberry Pi build?

BagelOrb:
We wouldn't want the preferences to explode. Two preferences for such a small feature is too much.

Perhaps we can just have one preference like it is now, but change the message so that it has a checkbox "Do not warn me again this session." Which resets after Cura is restarted.

resources/shaders/xray_composite.shader
BagelOrb: That commit got merged into here, so no need to test your branch separately from this PR.

Let's hope the Cura team can test this.

BagelOrb:
@fieldOfView 
This PR has been open for about half a year. The Cura team is now working on merging it. I think the time for proposing integral changes is past.

fieldOfView:
> This PR has been open for about half a year.

My comments are about a feature you added a month ago, when it was very unclear if this PR was going anywhere. Now that it looks like the PR is going to get merged, I would like it to be the best it can be. But let me know if my contributions are not wanted.

BagelOrb:
I like discussing the ideas, but I am afraid that the discussion means the feature will be put on hold again for another half year.

smartavionics:
The Pi 4 build is actually using GLES 3.1. I had to make some changes to the shaders. It looks the same as what I see on a x86_64 Linux box. This image is from the Pi 4.

Screenshot_2020-03-28_10-58-05

The changes I made are...

[x-ray-shaders-diff.txt](https://github.com/Ultimaker/Cura/files/4396983/x-ray-shaders-diff.txt)

BagelOrb:
Err. I'm sorry, why were these changes needed?

smartavionics:
Because otherwise the shaders are not valid on the Pi 4.

smartavionics:
GLES 3.1 doesn't allow implicit conversion between integer and floats and also you copy/pasted texture rather than using the required texture2D.

smartavionics:
BTW, when I checked the do not display this message again checkbox in the dialog, it made the new shading go away. I wanted it to not show the dialog but I would still like to see the error regions on the model highlighted.

fieldOfView:
> but I am afraid that the discussion means the feature will be put on hold again for another half year.

Understood. My concern is what happened with @smartavionics; from a UX standpoint I am afraid many users will turn off the entire feature because they don't want to see the message. Two simpler solutions than what I initially suggested:
1. Have the preference only act on detecting errors to show the message. This leaves the user unable to remove the xray visualisation, which I think is preferable to them turning it off for the wrong reason, or
2. Remove the option to turn off the feature from the messagebox, and only have it in the Preferences 

On a related note, I will adapt MeshTools so there it does not show a message that is already shown by Cura itself.

fieldOfView:
>> why were these changes needed?
>
> Because otherwise the shaders are not valid on the Pi 4.

The changes (especially those in xray_composite.shader) are also necessary for normal legacy renderers (ie: not just for the raspberry pi 4). Sorry I missed those in my PR.

You can test the legacy shaders by adding this to the `[view]` section of cura.cfg:

```opengl_version_detect = force_legacy``` 

Without Mark's changes, the viewport stays black because the shader does not compile.

File: resources/qml/Preferences/GeneralPage.qml [hunk 8]
```diff
@@ -86,6 +86,8 @@ UM.PreferencesPage
         prefixJobNameCheckbox.checked = boolCheck(UM.Preferences.getValue("cura/jobname_prefix"))
         UM.Preferences.resetPreference("view/show_overhang");
         showOverhangCheckbox.checked = boolCheck(UM.Preferences.getValue("view/show_overhang"))
+        UM.Preferences.resetPreference("view/show_xray_error");
+        showXrayErrorCheckbox.checked = boolCheck(UM.Preferences.getValue("view/show_xray_error"))
```
fieldOfView: Should be `view/show_xray_warning` instead of `view/show_xray_error` in this line and the one above.

Ghostkeeper:
> Two simpler solutions than what I initially suggested:
> 
>     1. Have the preference only act on detecting errors to show the message. This leaves the user unable to remove the xray visualisation, which I think is preferable to them turning it off for the wrong reason, or
> 
>     2. Remove the option to turn off the feature from the messagebox, and only have it in the Preferences

I would love to see the first option. That's what I thought it would do anyway, going by the text "do not show this message again" I'd expect it to just not show the message again but still render the X-ray.

nallath:
So, basically what needs to happen is :
- Disabling the option in the message should hide the message _only_
- Add the patches from @smartavionics so it works on PI's. 

Is that correct?

rburema:
@nallath 

> Add the patches from @smartavionics so it works on PI's.

Not _just_ on the PI's ... The legacy OpenGL fallback possibility is even more important. Hopefully the fixes work in both instances.

Also UX still needs to agree to it.

smartavionics:
Aldo says above that those changes are good for the legacy OpenGL but I don't know if they have actually been tested.

Yian-Lai:
Hi all, 
Regarding message I think it should not have the option to "Don't show again". Since it is an error message, we want to always inform the user about it. And only visualizing the error areas without a message we might run the risk of " user seeing there's something wrong but still don't know why or how to fix it".
So I would propose not to put any user preferences for this. 

(If this really happens a lot and starts to annoy people then we can put a preference to disable the detecting and messaging.)

fieldOfView:
For your convenience, I have made a PR on top of this PR including the changes mentioned by @nallath and two more fixes.

rburema:
@Yian-Lai Are you OK with the visuals? There was some discussion about that.

fieldOfView:
I updated #7373 to align with @Yian-Lai's comments.

Yian-Lai:
@rburema I suggested @nallath to try the stripes(like when the model is out of buildplate you mentioned last week). Use red stripes with opacity in the error areas, because it has the similar meaning that "you can not print this model". And I assume stripes will work better in following the phases of the model, compared to the icon always have its own perspective.

BagelOrb:
> And only visualizing the error areas without a message we might run the risk of " user seeing there's something wrong but still don't know why or how to fix it".

I don't agree. If a user has seen the message he will understand the link between the message and the colored patches.

If you - as a user - don't know exactly what the message says you shouldn't hit the 'Don't say this again' checkbox. That holds true for *any* popup which has a 'Don't say this again' checkbox.

BagelOrb:
>  I assume stripes will work better in following the phases of the model, compared to the icon always have its own perspective.

The striped shader for models which are outside the build plate also don't follow the model; they also have their own perspective. I think it's good that they have their own perspective, because it makes those areas stand out more. When you move around the model you can immediately see one face is changing all the time, while if a face only has a slightly different color you might not notice it if it is small.

BagelOrb:
That being said. Its currently just an image, so you can easily change it for an image with stripes.

PR #6570, opened 2019-10-23
CURA-6665 Add @api decorator

LipuFei:


cura/Utils/Decorators.py
```diff
@@ -0,0 +1,19 @@
...
+import functools
+
+
+##  Decorator for functions that belong to a set of APIs. For now, this should only be used for officially supported
+#   APIs, meaning that those APIs should be versioned and maintained.
+#
+#   \param since The earliest version since when this API becomes supported. This means that since this version, this
+#                API function is supposed to behave the same. This parameter is not used. It's just a documentation.
+#
+def api(since = "Unknown"):
```
nallath: I think we should force a correct API version to be used here (especially since we want to use this in the future to auto generate documentation!)

ninovanhooff: From the requirements:
> The decorator should just be simple, without any arguments, even lacking 'from' and 'to' versions:

So I'd say that the parameter should be removed alltogether

LipuFei: I added a semantic version check in 08b1e0e.

PR #6574, opened 2019-10-24
disable shell for infill meshes by default

BagelOrb:
When modifying the infill we generally don't want the shell.

Up till now we've been telling people to disable them by hand.

Adding an extra shell on the inside should be the special case and removing the shell should be default.

This is a no-brainer.

**EDIT:**
The initial implementation using inheritance formulae was flawed (see comments below).

I've redone the whole implementation of this PR. Now it changes the per-object settings stack instead of using obscure inheritance formulae.


ianpaschal:
There you go, @BagelOrb 😉

rburema:
What if the user manually specifies the wall line count? That'd overwrite the formula.

I discussed it and this solution may still be better than the previous behaviour, but it does introduce an inconsistency.

BagelOrb:
What do you mean? The whole idea of those formulas is that settings are automatically changed unless the user specifies an absolute value which overrides the formula.

BagelOrb:
No the problem is when the user specifies a wall thickness, not when he specifies a wall line count. You are right. The formula should be in the parent setting.

rburema:
Exactly, so the user sets up a scenario with 2 meshes, where one is an infill (modifier) mesh. Then, on the entire combined model, the user wants thicker walls (like 5 for example). So the users sees '3' in that field (which was the result of that formula), types in '5', which is a plain value, and thus overrides it, also for the infill-mesh, which again results in walls where you'd expect infill (like in the original scenario we're trying to prevent).

BagelOrb:
That's a highly unlikely scenario. People will generally edit the wall thickness of the total thing in the global settings, not in the per-object settings of the group.

My scenario is more likely, where a users wants more internal walls and therefore changes teh wall thickness of the infill mesh, which then doesn't do anything.

BagelOrb:
> This is a no-brainer.

... he said overconfidently.

rburema:
> That's a highly unlikely scenario. People will generally edit the wall thickness of the total thing in the global settings, not in the per-object settings of the group.

> My scenario is more likely, where a users wants more internal walls and therefore changes teh wall thickness of the infill mesh, which then doesn't do anything.

And we agree with that, but it _does_ bear mentioning! So we all remember next time the highly unlikely scenario occurs and don't go looking for a bug that doesn't exist once this is merged :-)

BagelOrb:
You misunderstood. My scenario which is presumably more likely is also a bug scenario.

BagelOrb:
I was thinking of disabling the wall thickness setting for infill meshes because of this.

rburema:
> You misunderstood. My scenario which is presumably more likely is also a bug scenario.

Ok, shall I put this back then until you've had some time to think about it?

BagelOrb:
I don't need thinking time -- I'd like to _discuss_ this with you.

rburema:
> I was thinking of disabling the wall thickness setting for infill meshes because of this.

That'd probably a good idea I guess? There's not a lot of uses for an infill mesh with walls, at least not that I can think of.

BagelOrb:
Note that currently the default is actually a single wall.

There are definitely uses. We know the walls give a part the majority of its stiffness, so adding walls in certain places makes those regions more stiff.

BagelOrb:
> Note that currently the default is actually a single wall.

That was incorrect of me.

BagelOrb:
Disabling the setting doesn't work for the per-object settings :(

BagelOrb:
Hmmmm.

I think it's better to implement the following:
when you set the mesh type to infill mesh, the wall thickness and top bottom thickness get automatically added to the object setting stack and the value is automatically set to 0. When you set it to a different mesh type those two settings are automatically removed from the stack if they are still zero (i.e. not changed by user).

I can see if I can get that working today, but if not I won't spend time on it, so then you will have to decide whether we want to merge this PR or not - given all of the uglinesses it gives.

BagelOrb:
The current code works, but there is a UX update bug which causes the settings visibility change not to be represented in the view. Just toggle to another tool and back to the per object settings tool to see the settings with changed values.

Surely this needs to be fixed, but me and @rburema spent about an hour on it and we couldn't figure it out yet.

Ghostkeeper:
Found the bug! This seems to work fine then. Thanks for the contribution!

BagelOrb:
:tada:

PR #6577, opened 2019-10-24
make lithophane alpha channel control total final width multiplier.

BagelOrb:
That way the picture is not always square. We can use the alpha channel to encode the actual shape

### DIFF
--- a/plugins/ImageReader/ImageReader.py
+++ b/plugins/ImageReader/ImageReader.py
@@ -5,7 +5,7 @@ import numpy
 
 import math
 
-from PyQt5.QtGui import QImage, qRed, qGreen, qBlue
+from PyQt5.QtGui import QImage, qRed, qGreen, qBlue, qAlpha
 from PyQt5.QtCore import Qt
 
 from UM.Mesh.MeshReader import MeshReader
@@ -137,6 +137,11 @@ class ImageReader(MeshReader):
         else:
             height_data *= scale_vector.y
             height_data += base_height
+
+        if img.hasAlphaChannel():
+            for x in range(0, width):
+                for y in range(0, height):
+                    height_data[y, x] *= qAlpha(img.pixel(x, y)) / 255.0
 
         heightmap_face_count = 2 * height_minus_one * width_minus_one
         total_face_count = heightmap_face_count + (width_minus_one * 2) * (height_minus_one * 2) + 2


BagelOrb:
Requested here: https://www.reddit.com/r/3Dprinting/comments/dmdbvn/lithophane_png/

BagelOrb:
This test image has an alpha channel:
convex_hulk_txt

BagelOrb:
It actually works!
(Who would have thought!) :P
image

BagelOrb:
Note that the zero alpha pixels still show up as a zero-width surface.

I know it's not ideal, but let's not put too much effort into this, okay?
The solution to that would be super complicated!

fieldOfView:
Cool! How does it look in X-Ray View?

BagelOrb:
I'm probably thinking about it too easily again, but do we really need to code check this and QA this?
Lithophanes are super uncritical.

fieldOfView:
One could argue that the ImageReader should be a Marketplace plugin. I think it would even be *more* discoverable that way. That way only @diegopradogesto has to test it ;-)

BagelOrb:
Not a bad idea. Maybe I will do that in the future

Ghostkeeper:
> One could argue that the ImageReader should be a Marketplace plugin. I think it would even be more discoverable that way. That way only @diegopradogesto has to test it ;-)

Since Diego left our team, the development team has been spreading the work, actually.

fieldOfView:
Ah, that explains why my plugin has been "submitted" without review for >2 weeks now. Bye Diego 👋

Ghostkeeper:
Yeah we've only recently been spreading the work because Diego's successor had not been doing it for a while. Should be better now though!

ninovanhooff:
Cool feature, also love the simplicity of the implementation.
Might be nice for height maps as well, or maybe sprites where you can just create a lot of board game / terrain tiles 
Devs: https://jira.ultimaker.com/browse/CURA-7056

ninovanhooff:
Force pushed a rebase since @BagelOrb previously indicated he prefers that.

BagelOrb:
I do prefer rebasing over merging, but I don't require you to also force push it. My opinion was about keeping a clean history in the master branch without those ugly merge commits. Merge commits sometimes hide the introduction of bugs which are then very difficult to got bisect.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Issue #6433, opened 2019-09-27
Ultimaker Cura 4.3.0 Printer Selection not available


tronicgr:
I have two printers, an Ultimaker3 and Ultimaker3 extended. They are setup as group with one of them being the host using the Cura Connect. But I cannot select which one will print over network anymore, the pop up window is missing.

**Application version**

**Platform**
 Windows 10 Enterprise 

**Printer**
 Ultimaker3 and Ultimaker3 extented 

**Reproduction steps**
 In previous Cura version, when I pressed the button to "Print over network" it would popup a printer selection menu to choose which one to send the print to. 

**Actual results**
 Now after installing the Cura 4.3.0 this window does not pop up anymore and the print is send to the first available. This is inconvenient as I have different color material on each and I'd like to chose which one can print...  

**Expected results**
Can you bring back that printer selection menu? Is there a setting somewhere to enable it back?

tronicgr:
I have two printers, an Ultimaker3 and Ultimaker3 extended. They are setup as group with one of them being the host using the Cura Connect.
 
In previous Cura version, when I pressed the button to "Print over network" it would popup a printer selection menu to choose which one to send the print to. Now after installing the Cura 4.3.0 this window does not pop up anymore and the print is send to the first available. This is inconvenient as I have different color material on each and I'd like to chose which one can print...

I have Windows 10 Enterprise.
 
Can you bring back that printer selection menu? Is there a setting somewhere to enable it back?

Ghostkeeper:
I'm told it was intentionally removed. It was no longer possible due to a refactor that was made during this release. I'll ask the developers of that plug-in on whether it needs to be put back.

tlum2:
Not being able to select from our 4 printers is a huge feature deletion, basically it is now taking us back to putting everything on thumb drives so we can select the correct printer. Please consider reinstating this feature.

ChrisTerBeke:
Addressed by https://github.com/Ultimaker/Cura/pull/6580.

PR #6580, opened 2019-10-25
Bring back the printer selection dialog for networked printers

ChrisTerBeke:
We removed this feature in 4.3, but market feedback let us know that this feature was actually used quite a bit. This PR brings it back! It also polishes the previous code for this feature a bit (no more 4 yields).

(see https://github.com/Ultimaker/Cura/issues/6433 and https://community.ultimaker.com/topic/29672-cant-select-specific-printer-43/)


Quasardian:
Changing the printer selection in the Cura Cloud could also be useful in my opinion.
(as well as a material override setting)

PR #6581, opened 2019-10-25
CURA-6557 Add CMake options to exclude plugins for installation

LipuFei:


cmake/CuraPluginInstall.cmake
```diff
@@ -0,0 +1,98 @@
...
+    get_filename_component(_plugin_dir ${_plugin_json_path} DIRECTORY)
+    file(RELATIVE_PATH _rel_plugin_dir ${CMAKE_CURRENT_SOURCE_DIR} ${_plugin_dir})
+    get_filename_component(_plugin_dir_name ${_plugin_dir} NAME)
+
+    # Make plugin name comparison case-insensitive
+    string(TOLOWER "${_plugin_dir_name}" _plugin_dir_name_lowercase)
+
+    # Check if this plugin needs to be skipped for installation
+    set(_add_plugin ON)
+    set(_is_no_install_plugin OFF)
```
ninovanhooff: Why do you need 2 booleans here which are opposites?

cmake/mod_bundled_packages_json.py
```diff
@@ -0,0 +1,71 @@
...
+
+import argparse
+import collections
+import json
+import os
+import sys
+
+
+def find_json_files(work_dir: str) -> list:
+    """
```
ninovanhooff: You can use docstrings rather than doxygen as long as you are writing non- core cura script?
I am obviously a big proponent of docstrings, just wanted to make sure what is allowed by the documentation police

cmake/CuraPluginInstall.cmake
LipuFei: Added more docs in fc9ee6e85. This is more for debugging/information, that I want to know if a plugin is not added, is it because it's specified in the NO_INSTALL_PLUGINS list or not.

cmake/mod_bundled_packages_json.py
LipuFei: I don't know. I thought we still want to move to docstrings because we can't get arguments and types from doxygen.

nallath: Officially we're on Doxygen, so that's what we need to use (because ya know, that's what the codestyle says)

LipuFei: Converted to doxygen style in 7b9ababc1

PR #6591, opened 2019-10-28
Hopefully Adding the Makeblock 3d Printer files to the latest version

pkz0313:
I created definition, extruder, and quality for the Makeblock 3D Printer. I hope that it could be in the latest release of Ultimaker Cura!

Thanks a lot!

### DIFF
--- a/resources/definitions/makeblock.def.json
+++ b/resources/definitions/makeblock.def.json
@@ -1,0 +1,68 @@
+{
+    "version": 2,
+    "name": "Makeblock",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "author": "Makeblock",
+        "manufacturer": "Makeblock",
+        "visible": true,
+        "file_formats": "application/gzip;text/x-gcode",
+        "has_machine_quality": true,
+        "preferred_quality_type": "normal",
+        "machine_extruder_trains": {
+            "0": "makeblock_extruder_0"
+        }
+    },
+    "overrides": {
+        "machine_name": {
+            "default_value": "makeblock"
+        },
+        "machine_width": {
+            "default_value": 225
+        },
+        "machine_depth": {
+            "default_value": 225
+        },
+        "machine_height": {
+            "default_value": 300
+        },
+        "machine_heated_bed": {
+            "default_value": true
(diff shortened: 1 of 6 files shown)


Ghostkeeper:
I tried to find this 3D printer, but on the [Makeblock website](https://www.makeblock.com/) it doesn't list them. After searching for it with a search engine I did find [this one](https://www.makeblock.com/project/mgiraffe-3d-printer-kit) which seems to be a community project of sorts. However that printer is called the "mGiraffe". And it's a Delta printer, which this definition is not.

What is this printer and where can I find information about it?

pkz0313:
@Ghostkeeper Hi, sorry for bringing you inconvenience. This 3d printer is our new product and it can't be found on the Makeblock Website yet.

[](url)
[Makeblock mCreate Product Introduction--EN.pdf](https://github.com/Ultimaker/Cura/files/3791962/Makeblock.mCreate.Product.Introduction--EN.pdf)

Ghostkeeper:
Ah, right. Then maybe can we call this printer "mCreate" or "Makeblock mCreate" then, instead of just generic Makeblock? The manufacturer can still be Makeblock. But that would allow for other printers by Makeblock to be added in the future as well and would reduce confusion of people that have those other printers at home.

And if you do, please also rename the files.

pkz0313:
ok. The files had been renamed.

Ghostkeeper:
Developers, please refer to issue CURA-6952.

Ghostkeeper:
The `gantry_height` override has no effect, since you are only overriding the `default_value` but this setting has a `value`, which gets precedence over the default_value.

A gantry height of 0 will effectively prevent one-at-a-time mode as well. Are you sure that's what you intended? You could also just deactivate the `print_sequence` setting.

pkz0313:
The gantry_height had been updated.

fieldOfView:
The `gantry_height` override still has no effect, because you are still setting a default value instead of a value.

nallath:
I've already fixed that in master (since our CI was also complaining about that)

pkz0313:
Thank you! You guys are so nice.

PR #6592, opened 2019-10-28
CURA-6874_path_slider_layout

ninovanhooff:
Dynamically resize the simulation path slider width based on 
the window width and the preferred width from the theme.
Same approach as used for the layer slider

### DIFF
--- a/plugins/PreviewStage/PreviewMain.qml
+++ b/plugins/PreviewStage/PreviewMain.qml
@@ -24,7 +24,6 @@ Item
         y: safeArea.y - parent.y
         width: actionPanelWidget.x - x
         height: actionPanelWidget.y - y
-        visible: false
     }
 
     Loader
--- a/plugins/PreviewStage/PreviewMain.qml
+++ b/plugins/PreviewStage/PreviewMain.qml
@@ -17,21 +17,14 @@ Item
     property var safeArea: parent
 
     // Subtract the actionPanel from the safe area. This way the view won't draw interface elements under/over it
-    Rectangle
+    Item
     {
         id: childSafeArea
         x: safeArea.x - parent.x
         y: safeArea.y - parent.y
         width: actionPanelWidget.x - x
         height: actionPanelWidget.y - y
-        visible: false // true for debug only
-        color:"#800000FF"
-
-        Component.onCompleted: {
-            print("parent", parent.x, parent.y)
-            print("parent safe", safeArea.x, safeArea.y)
-            print("previewmain safe", childSafeArea.x, childSafeArea.y, childSafeArea.width, childSafeArea.height)
-        }
+        visible: false
     }
 
     Loader
--- a/plugins/PreviewStage/PreviewMain.qml
+++ b/plugins/PreviewStage/PreviewMain.qml
@@ -24,7 +24,7 @@ Item
         y: safeArea.y - parent.y
         width: actionPanelWidget.x - x
         height: actionPanelWidget.y - y
-        visible: true // true for debug only
+        visible: false // true for debug only
         color:"#800000FF"
 
         Component.onCompleted: {
--- a/plugins/PreviewStage/PreviewMain.qml
+++ b/plugins/PreviewStage/PreviewMain.qml
@@ -12,14 +12,26 @@ import Cura 1.0 as Cura
 Item
 {
 
+    // An Item whose bounds are guaranteed to be safe for overlays to be placed.
+    // Defaults to parent, ie. the entire available area
+    property var safeArea: parent
+
     // Subtract the actionPanel from the safe area. This way the view won't draw interface elements under/over it
-    Item {
-        id: safeArea
-        visible: false
-        anchors.left: parent.left
-        anchors.right: actionPanelWidget.left
-        anchors.top: parent.top
-        anchors.bottom: actionPanelWidget.top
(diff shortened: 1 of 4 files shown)


plugins/PreviewStage/PreviewMain.qml
```diff
@@ -12,14 +12,26 @@ import Cura 1.0 as Cura
...
+        width: actionPanelWidget.x - x
+        height: actionPanelWidget.y - y
+        visible: false // true for debug only
+        color:"#800000FF"
+
+        Component.onCompleted: {
+            print("parent", parent.x, parent.y)
+            print("parent safe", safeArea.x, safeArea.y)
+            print("previewmain safe", childSafeArea.x, childSafeArea.y, childSafeArea.width, childSafeArea.height)
+        }
```
nallath: This Rectangle needs to be removed.

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
 
 
     property bool isSimulationPlaying: false
+    readonly property var layerSliderSafeYMin: safeArea.y
```
nallath:
```suggestion
    readonly property real layerSliderSafeYMin: safeArea.y
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
...
     readonly property var layerSliderSafeYMax: safeArea.y + safeArea.height
+    readonly property var pathSliderSafeXMin: safeArea.x + playButton.width //todo playbutton margin or group button + slider in an item?
```
nallath:
```suggestion
    readonly property real pathSliderSafeXMin: safeArea.x + playButton.width //todo playbutton margin or group button + slider in an item?
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
...
+    readonly property var pathSliderSafeXMax: safeArea.x + safeArea.width
```
nallath:
```suggestion
    readonly property real pathSliderSafeXMax: safeArea.x + safeArea.width
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
...
 
     visible: UM.SimulationView.layerActivity && CuraApplication.platformActivity
 
     // A slider which lets users trace a single layer (XY movements)
     PathSlider
     {
         id: pathSlider
+        
+        readonly property var preferredWidth: UM.Theme.getSize("slider_layerview_size").height // not a typo, should be as long as layerview slider
```
nallath:
```suggestion
        readonly property real preferredWidth: UM.Theme.getSize("slider_layerview_size").height // not a typo, should be as long as layerview slider
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
...
+        readonly property var margin: UM.Theme.getSize("default_margin").width
```
nallath:
```suggestion
        readonly property real margin: UM.Theme.getSize("default_margin").width
```

plugins/SimulationView/SimulationViewMainComponent.qml
```diff
@@ -18,21 +18,32 @@ Item
...
+        readonly property var pathSliderSafeWidth: pathSliderSafeXMax - pathSliderSafeXMin
```
nallath:
```suggestion
        readonly property real pathSliderSafeWidth: pathSliderSafeXMax - pathSliderSafeXMin
```

resources/qml/Cura.qml
```diff
@@ -301,6 +301,17 @@ UM.MainWindow
...
+            // A hint for the loaded content view. Overlay items / controls can safely be placed in this area
+            Rectangle {
+                id: mainSafeArea
+                anchors.left: viewOrientationControls.right
+                anchors.right: main.right
+                anchors.top: main.top
+                anchors.bottom: main.bottom
+                visible: false // set to true for debugging only
+                color:"#8000FF00"
+            }
```
nallath: Debug code can be removed (There is no need for it to be created in "live" code, but it will cost a bit of resources to do so)

plugins/PreviewStage/PreviewMain.qml
ninovanhooff: It is used in line 47. If you object to a Rect it can be an Item, but I'd say that an invisible Rectangle is not more expensive than an invisible Item in production code

ninovanhooff: I should remove the print statements tho

nallath: An invisible rectangle is slightly more expensive than an Item (Although it's not a whole lot)

ninovanhooff: Allright, I'll replace them

resources/qml/Cura.qml
ninovanhooff: Replaced by Item

PR #6594, opened 2019-10-29
Make jobname less sensitive to being touched

fieldOfView:
This PR makes the jobname less sensitive to getting "frozen" by merely clicking in the jobname area.

When the jobname is edited using the jobname field in the UI, it is "frozen"; it will no longer update eg when changing to a different machine. The current implementation is a bit overly sensitive in that even when just clicking inside the jobname textfield and then outside it without changing the text, the jobname is considered to be edited. With this PR, the jobname is only frozen when the jobname text is actually changed.


LipuFei:
JIRA ticket: CURA-7051

PR #6595, opened 2019-10-29
Fix Z hops when retracting

Ghostkeeper:
These are profile changes needed to ensure that the behaviour is the same as what it used to be.

See https://github.com/Ultimaker/CuraEngine/pull/1155 for a full explanation.

Implements CURA-6860.


Ghostkeeper:
Need to hold those horses a little bit as this ticket is still in progress and we're investigating the purpose and expected behaviour of the Only Hop Over Printed Parts setting. We're doing our best to keep everything both working consistently and working the same as in previous versions, though those may be in contradiction here. Still in progress!

smartavionics:
Ok, so here's a look at some gcode created today and you can see it z-hops just fine.

Screenshot_2019-10-30_09-08-05

smartavionics:
Travel settings are:

Screenshot_2019-10-30_09-11-00

Ignore the max comb extra distance addition as that is set to 0 and not having any effect.

Ghostkeeper:
Superseded by https://github.com/Ultimaker/CuraEngine/pull/1185.

PR #6605, opened 2019-10-31
Adds visual intents

THeijmans:
and create subdirectories per printer. As discussed 31-10-2019.


nallath:
Ugh. this is why I hate rebase. I'll just cherypick it.

PR #6608, opened 2019-10-31
Leapfrog Bolt Pro machine profile

VincentRiemens:
definition
extruder  
meshes   
quality
variants


resources/extruders/leapfrog_bolt_pro_extruder_left.def.json
```diff
@@ -0,0 +1,23 @@
+{
+    "id": "leapfrog_bolt_pro_extruder_left",
+    "version": 2,
+    "name": "Left extruder",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "leapfrog_bolt_pro",
+        "position": "1"
```
nallath: This is going to place the left extruder on the right (and the right extruder on the left)

resources/definitions/leapfrog_bolt_pro.def.json
```diff
@@ -0,0 +1,131 @@
...
+        "supports_network_connection": false,
+        "has_materials": true,
+        "preferred_material": "leapfrog",
+        "has_machine_quality": true,
+        "has_variants": true,
+        "preferred_variant_name": "Brass 0.4",
+        "preferred_material": "leapfrog_epla_natural",
+        "variants_name": "Hot end",
+        "exclude_materials": [
+            "generic_pla",
```
nallath: You don't need to mention all the materials that you couldn't use in the first place (eg; Your machine is 1.75 mm, so you don't need to exclude generic_pla, since that is a 2.85mm material and will be filtered away because of that)

resources/definitions/leapfrog_bolt_pro.def.json
```diff
@@ -0,0 +1,131 @@
...
+    },
+    "overrides": {
+        "machine_name": {"default_value": "Leapfrog Bolt Pro" },
+        "machine_extruder_count": {"default_value": 2},
+		"machine_center_is_zero": {"default_value": false},
+        "machine_width": {"default_value": 302},
+        "machine_height": {"default_value": 205},
+        "machine_depth": {"default_value": 322},
+        "machine_heated_bed": {"default_value": true},
+        "machine_head_with_fans_polygon": {"default_value": [[-1, 1 ], [1, 1], [1, -1 ], [-1, -1]]},
```
nallath: Note that this might cause damage to printer or prints if a user prints with one at a time.

resources/extruders/leapfrog_bolt_pro_extruder_left.def.json
VincentRiemens: Dear Nallath, Thank you for your feedback.

With the Leapfrog Bolt PRO, T0 is considered right extruder , T1 is considered left extruder. 
We also placed a "1" and "2" above the extruders in the mesh file (leapfrog_bolt_pro_platform.stl) to clarify on this. 

We believe that the naming "Left extruder" or "Right extruder" (for example in slice preview) matches this also.

resources/definitions/leapfrog_bolt_pro.def.json
VincentRiemens: Thank you for your feedback. The extruder heads are ''homed'' outside of the build volume. When only one of the extruders used, this one will be printing within the build volume while the other tool will be inactive on its homing location outside of the build volume.

resources/definitions/leapfrog_bolt_pro.def.json
VincentRiemens: Thank you for your feedback. We have removed all materials that are being filtered away because of a different diameter.

Ghostkeeper:
Developers, please refer to issue CURA-6951.

resources/definitions/leapfrog_bolt_pro.def.json
nallath: Printing one at the time has little to do with homing, but with knocking a print over because they are not printed one layer at a time. So if you have 2 objects on the buildplate (called A & B) and all of them are printed with the same extruder, it will first completely print A and then print B.

This means that some avoidance needs to be done in order to prevent colisisons, which is calculated by this property. It's fine by me to leave it, but i just wanted to make sure that this is intentional.

VincentRiemens: Dear Nallath, Thank you for clarifying this . We have adjusted the values (also gantry height).

LipuFei:
I will fixed the problems separately.

PR #6609, opened 2019-10-31
Removes visual intents for 0.2mm layer height profiles

THeijmans:


nallath:
You based this branch of master (and not 4.4). So we can't merge this.

PR #6614, opened 2019-11-01
CURA-6522_one_at_a_time_overlapping_build_area

ninovanhooff:
This PR contains the following
- remove a 0.1 mm edge border which caused the bug reported in 6522
- merge machine_head_polygon and machine_head_with_fans_polygon because 
  I don't see a usecase for both to exist. Defenition maintainers seem 
  confused and overide either, both, or none (see Remco's comment in github)
- Version upgrade 44 to 45
- Merging of settings in the version upgrade + unit test

Please make sure you agree with merging of those settings, 
can't hurt to convince yourself that they server the same purpose.

Also, this is the first version upgrade, please make sure I didn't
forget anything there


nallath:
We should also actually change the setting_version of the profiles (otherwise they get constantly upgraded in memory)

PR #6615, opened 2019-11-01
Packages API speed improvements

ChrisTerBeke:
Cura hits the packages API on every startup, causing the API to have to handle many requests. Cura uses these API calls for checking which packages have updates available. This PR changes that call to another (new) endpoint that is way quicker. This reduces load on the API, as well as the time it takes for Cura to display the update notification in the menu bar.

It would be great to have this in 4.4, but since we're in feature freeze I can understand why you'd want this on master instead.

Proof that it works:

image

_Note: I did the best I could in the current 'toolbox' code base. It definitely needs refactoring at some point._


ChrisTerBeke:
The failing tests are not caused by these changes it seems.

fieldOfView:
A privacy-minded person would argue that this comes at the cost of loss of privacy, but speedups are welcome.

ChrisTerBeke:
> A privacy-minded person would argue that this comes at the cost of loss of privacy, but speedups are welcome.

I see your point, as Cura now sends a list of installed plugins to the server. However this endpoint can be used when logged out (so anonymous), so I think the impact is minimal. As far as I know, there's no plans to enforce signing in in order to use the Marketplace for the community.

fieldOfView:
My remark was a bit nitpicky. But come to think of it, does the Marketplace update query adhere to the privacy setting on the General page of the preferences? It probably should.

nallath:
Why not have an endpoint that just gives the latest version of all plugins? That way the server doesn't have to do any of the checking (and that way the server doesn't have to know anything at all)

ChrisTerBeke:
> Why not have an endpoint that just gives the latest version of all plugins? That way the server doesn't have to do any of the checking (and that way the server doesn't have to know anything at all)

That would still require the same amount of DB lookups on the server side not giving the desired speed increase.

fieldOfView:
> That would still require the same amount of DB lookups 

But you could cache the results, right? The cache would only need to be invalidated when a new package is added.

ChrisTerBeke:
> > That would still require the same amount of DB lookups
> 
> But you could cache the results, right? The cache would only need to be invalidated when a new package is added.

Caching would introduce a whole slew of other problems (not predictable when data is up-to-date) that we don't want to get into at the moment.

fieldOfView:
> As far as I know, there's no plans to enforce signing in in order to use the Marketplace for the community.

Well, that didn't take long: https://github.com/Ultimaker/Cura/pull/6624

ChrisTerBeke:
> > As far as I know, there's no plans to enforce signing in in order to use the Marketplace for the community.
> 
> Well, that didn't take long: #6624

I know that that is, and it's not enforcing login for normal users. Please ask before making such a statement.

rburema:
Failing check not related to PR, will fix later.

PR #6623, opened 2019-11-04
Organize Toolbox (marketplace) qml files according to type

ninovanhooff:
CURA-6569

Organize a gazillion qml files


cura/ApplicationMetadata.py
```diff
@@ -7,7 +7,7 @@
 DEFAULT_CURA_APP_NAME = "cura"
 DEFAULT_CURA_DISPLAY_NAME = "Ultimaker Cura"
 DEFAULT_CURA_VERSION = "master"
-DEFAULT_CURA_BUILD_TYPE = ""
+DEFAULT_CURA_BUILD_TYPE = "essentials"
```
Ghostkeeper: Probably don't want this on by default.

LipuFei: I will fix this.

PR #6616, opened 2019-11-01
CURA-6856 signed plugins and packages

rburema:
Smaller Cura counterpart of the main work in Uranium, see [here](https://github.com/Ultimaker/Uranium/pull/541).

PR #6624, opened 2019-11-04
WIP: CURA-6569_mandatory_marketplace_login

ninovanhooff:
Wip: Needs marketplace icon


fieldOfView:
Why?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -126,6 +125,16 @@ def __init__(self, application: CuraApplication) -> None:
     showLicenseDialog = pyqtSignal()
     uninstallVariablesChanged = pyqtSignal()
 
+    ##  Go back to the start state (welcome screen or loading if no login required)
+    def _restart(self):
+        self._updateRequestHeader()
+        # For an Essentials build, login is mandatory
+        if not self._application.getCuraAPI().account.isLoggedIn and ApplicationMetadata.CuraBuildType == "essentials":
```
ChrisTerBeke: I think there's already an isEnterprise flag in some config file somewhere right? That would be cleaner than hardcoding 'essentials' here again.

rburema: That has yet to be merged, but it would be nice to use the same code eventually.

rburema:
@fieldOfView The title is a bit unclear about it, but it won't be 'mandatory' for home users.

plugins/Toolbox/src/Toolbox.py
Ghostkeeper: There is no such flag. It's stored in CuraVersion.py as either `debug`, `release` or `essentials`.

A flag has the disadvantage of requiring additional flags for each of our types, of which multiple could be true. If something symbolic would be desired we can go for an enum. But since we're working in Python and there is no pre-compilation going on there is hardly any advantage to that over strings, so people tend to call it "Pythonic" to use strings instead of enums.

ChrisTerBeke: https://github.com/Ultimaker/Cura/pull/6616/files#diff-1bce51b6e513b01618a843adcc615850R48

Ghostkeeper:
> Why?

The normal Cura build will not require log-in for Marketplace. But we're building a stripped-down version, aimed at high-security companies, that's not allowing installation of plug-ins unless that plug-in has had a security review, and where an IT administrator can select plug-ins for their users to install. For this, the Marketplace requires a log-in if you're using the stripped-down build.

People are not agreeing on the name yet but "Cura Essentials" was going around. Which is why everyone is calling it "shampoo version".

plugins/Toolbox/src/Toolbox.py
rburema: @Ghostkeeper I did make an 'IsSecured' flag (in the other branch), which currently is on if it is an 'essentials' build, on the groudns that we might keep the functionality separate. On second thought, that wouldn't work here then ... so nevermind :-)

ChrisTerBeke: Mmm I agree it's maybe better to separate those indeed. Nvm then, was just a thought.

LipuFei: Added an `IsEnterpriseVersion` flag in 689562e61

fieldOfView:
Thanks for the clarification.

PR #6635, opened 2019-11-07
Fix window resize bug in LayerSlider concerning handle position

ninovanhooff:
Upper handle position would not be updated on resize

CURA-6854


ninovanhooff:
Immidetely accepted this after I found out this branch was merged and deleted before.

PR #6636, opened 2019-11-07
CURA-6917_cannot_unlink_duplicated_materials

ninovanhooff:
I'm not very knowledgeable about this area of the code. 
I'd say fixing the casing of the dict index is okay, 
but not sure why it worked originally (was the default for ignore_case changed?)

The getLinkedMaterials call returns 20+ materials after aforementioned fix. 
Not sure what amount to expect there. Depending on the answer it might 
be a good idea to only show the first result.


ninovanhooff:
Wrong branching, closing in favor of https://github.com/Ultimaker/Cura/pull/6637

PR #6640, opened 2019-11-08
update HMS434

maukcc:


nallath:
There are a number of conflicts with the master branch. Did you use an outdated version to make your changes?

maukcc:
Probably, we are at 4.2.

PR #6644, opened 2019-11-10
Added Ender 3 Prime Line to Start G-code

TheOneOgre:
Added the popular prime line for the Ender 3 as the default starting g-code and added a modification that works around the Cura 4.3.0 update that moves the print head to Z0 before starting the print, which moves the printhead down into the filament.

### DIFF
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -26,9 +26,7 @@
                 [32, 34]
             ]
         },
-         "machine_start_gcode":
-        {
-            "default_value": "
+         "machine_start_gcode": { "default_value": "
             ; Ender 3 Custom Start G-code
             G92 E0 ; Reset Extruder
             G28 ; Home all axes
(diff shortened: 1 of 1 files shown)


nallath:
Would this work for all creality printers? If so, this could be something that can be put in the creality_base definition.

baddogg1231:
> Would this work for all creality printers? If so, this could be something that can be put in the creality_base definition.

This definitely could. I am not familiar with all the Creality machines as I've only had experience with my Ender 3 as well as a CR-10. I would assume they should all function exactly the same considering this is some simple g-code. If anyone could test it out to confirm, then it definitely could be useful to put into base definitions.

Ghostkeeper:
Shall we give this another 3 days for people to optionally test it for a bigger size, and otherwise just merge it for just the Ender 3 only?

It seems safer for me to just merge it for the printer it's been tested with. The Creality printers have some weirdnesses going on with their coordinate system sometimes, where zero is not always reachable and such.

Ghostkeeper:
Needed to make a minor modification for the JSON formatting. But thanks for the new start g-code!

PR #6654, opened 2019-11-13
Added ChituMods.py

Spanni26:
Added ChituMods.py which adds a preview (M4010), time (M2100) and time elapsed (M2101) for chitu boards.


plugins/PostProcessingPlugin/scripts/ChituMods.py
```diff
@@ -0,0 +1,189 @@
...
+        self.out_data = in_data
+        if self.getSettingValueByKey("insert_preview_image"):
+            self.img_data=[]
+            self._createSnapshot()
+            self.img_data.append('\n'.join(self.generate_image_code(self._snapshot))) # create one long string and add it as item to a list
+            self.img_data[0] += ('\n') # add an additional newline, looks better
+            self.img_data.extend(in_data)
+            self.out_data=self.img_data      
+        else:
+            self.out_data=in_data
```
fieldOfView: This else statement is superflous

plugins/PostProcessingPlugin/scripts/ChituMods.py
```diff
@@ -0,0 +1,189 @@
...
+                    "description": "Selecting this the calculated time and the elapsed time will be included in chitu style",
+                    "type": "bool",
+                    "default_value": false
+                }
+            }
+        }"""
+
+    def execute(self, in_data):
+        # we get a list, each list item is the command set for a complete layer
+        self.out_data = in_data
```
fieldOfView: There is no need to keep the out_data beyond the scope of this execute method. Instead of `self.out_data` you can just use `out_data`.

plugins/PostProcessingPlugin/scripts/ChituMods.py
```diff
@@ -0,0 +1,189 @@
...
+        return return_data        
+        
+
+    def _createSnapshot(self, *args):
+        Logger.log("d", "Creating tronxy thumbnail image ...")
+        try:
+            self._snapshot = Snapshot.snapshot(width = 300, height = 300)
+        except Exception:
+            Logger.logException("w", "Failed to create snapshot image")
+            self._snapshot = None
```
fieldOfView: instead of storing the _snapshot in the class instance, it would be nicer to return it from this method.

plugins/PostProcessingPlugin/scripts/ChituMods.py
Spanni26: Removed the last else statement. Thanks

plugins/PostProcessingPlugin/scripts/ChituMods.py
Spanni26: You are right. Thanks.

Done

plugins/PostProcessingPlugin/scripts/ChituMods.py
Spanni26: Done.

ninovanhooff:
Hi @Spanni26 , thanks for your contribution and sorry for the long wait. I discussed this with @Ghostkeeper and unfortunately we don't see enough reason to merge this in Cura itself. It currently is not possible to create postprocessing script plugins, but you are free to create a `mesh_writer` plugin. 

1. In your machine definition, register your machine with your own custom mime-type, like `text/chitu-g-code`. See for example `ultimaker_s5.def.json` for a way to indicate that your machine supports multiple `file_formats`
2. Create a `mesh_writer` plugin. See `GCodeWriter` as an example
3. Test your plugin with your machine
4. Submit your plugin to https://contribute.ultimaker.com/

PR #6661, opened 2019-11-15
Added LotMaxx sc10 and sc20 profile

sm3dp:
We want to add our printer to Cura profile.  Please check.  thansk.


resources/extruders/lotmaxx_base_extruder_0.def.json
```diff
@@ -0,0 +1,15 @@
+{
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "lotmaxx_base",
```
LipuFei: There's no machine `lotmaxx_base`.

PR #6666, opened 2019-11-16
Add VORON2 printer profile

Fulg:
I would like to add native support for the VORON2 printer in Cura. Hopefully this can make it into 4.4. Thanks!


Maglin:
I've tested this and love how it's working.  I'm hoping it can get merged into the next release.

resources/definitions/voron2_base.def.json
```diff
@@ -0,0 +1,155 @@
...
+        "machine_head_with_fans_polygon":               { "default_value": [ [-35, 65], [-35, -50], [35, -50], [35, 65] ] },
+        "machine_max_feedrate_z":                       { "default_value": 40 },
+        "machine_max_feedrate_e":                       { "default_value": 120 },
+        "machine_gcode_flavor":                         { "default_value": "RepRap (RepRap)" },
+        "machine_start_gcode":                          { "default_value": "print_start" },
+        "machine_end_gcode":                            { "default_value": "print_end" },
+        "adhesion_type":                                { "default_value": "skirt" },
+        "skirt_brim_minimal_length":                    { "default_value": 550 },
+        "retraction_speed":                             { "default_value": 80, "maximum_value_warning": 130 },
+        "retraction_retract_speed":                     { "default_value": "retraction_speed", "maximum_value_warning": 130 },
```
LipuFei: Please take a look at the check result.
```
E       AssertionError: Unnecessary default_values for ['retraction_retract_speed'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/voron2_base.def.json
```

resources/extruders/voron2_extruder_0.def.json
```diff
@@ -0,0 +1,17 @@
+{
+    "id": "voron2_extruder_0",
```
LipuFei: The tests didn't pass.

```
>       assert "id" not in doc, "Definitions should not have an ID field."
E       AssertionError: Definitions should not have an ID field.
E       assert 'id' not in {'id': 'voron2_extruder_0', 'inherits': 'fdmextruder', 'metadata': {'machine': 'voron2_base', 'position': '0'}, 'name': 'Toolhead', ...}
```

resources/definitions/voron2_base.def.json
ninovanhooff: Seems like this override was done to change the warning value. @Fulg, please confirm

Fulg: Yes, the intent was to raise the warning, since the default speed is already above the allowed maximum. I tested without it and the updated default value worked fine.

resources/extruders/voron2_extruder_0.def.json
Fulg: Nice, missed this. I guess this is new from 4.4. Fixed.

Fulg:
Let me know if you want me to change anything else.

Fulg:
@LipuFei, this is what I did originally but since Cura 4.4 it no longer seems to work (it did in Cura 4.3 and earlier). If you don't have a `material` property in your quality profile, then _no materials are supported at all:_

image

I am happy to fix this another way if you have a suggestion. Perhaps this behavior is not normal... I can submit the (broken) fix if you want to test it out yourself without the hassle of changing all the quality files, I can always revert later.

Fulg:
Adding relevant debug log entries:
```
2019-12-13 17:41:02,832 - DEBUG - [MainThread] cura.Settings.MachineManager._updateQualityWithMaterial [1122]: Updating quality/quality_changes due to material change
2019-12-13 17:41:02,837 - DEBUG - [MainThread] cura.Settings.MachineManager._updateQualityWithMaterial [1127]: Current quality type = [not_supported]
2019-12-13 17:41:02,841 - INFO - [MainThread] cura.Settings.MachineManager._updateQualityWithMaterial [1136]: No available quality types found, setting all qualities to empty (Not Supported).
```

ninovanhooff:
You might just have found a bug here, since I haven't heard of anyone knowingly changing this behavior. We would like to have a look at the optimized non-working profiles

Ghostkeeper:
We did change something here in 4.4, indeed: If you have machine-specific quality profiles, then it will load the quality profiles and they will also be material-specific and variant-specific. In 4.3 and before it was reading a metadata field `has_material_quality` and if it was `True` it would make quality profiles material-specific and variant-specific. However we found that not a single printer was using that so we didn't re-implement this when we refactored that for 4.4.

However the Voron printer doesn't override that metadata so I'd expect it to behave the same.

Now your profiles seem to specify the material and variant anyway though, so is it not working correctly?

Fulg:
@Ghostkeeper it was suggested that the profile could be simplified a whole lot by removing the material-specific properties, since all of them are identical. I replied that this worked before but not anymore in 4.4+.

My only concern with the new method is that this effectively blocks out materials that have no quality profiles "forever" (or at least until the VORON profile is updated in Cura). For example a user could not make his own TPU profile without poking in Cura's resource directory manually. I guess you could always lie to Cura and say it was ABS...

Perhaps the correct answer is to disable the `has_material_quality` flag, and disable the duplication in my files. I will test this out when I get back, thanks for the suggestion.

Fulg:
As requested. Note that commit f71b234 makes the profile no longer work as-is with the current Cura version.

Ghostkeeper:
> My only concern with the new method is that this effectively blocks out materials that have no quality profiles "forever" (or at least until the VORON profile is updated in Cura). For example a user could not make his own TPU profile without poking in Cura's resource directory manually. I guess you could always lie to Cura and say it was ABS...

If there are no quality or quality-changes profiles for a material, Cura will say that it's unsupported. If the user created their own material profile with material type "BLA", then there will not be profiles for it so it'll be marked as unsupported. But if the user then creates a custom profile for it as well (e.g. by changing a setting and then saving the changes to a profile) then it will no longer be marked as unsupported.

LipuFei:
FYI, I tested the changes here in 4.3 and master. If the quality profiles only specify machine and variant, it works in 4.3 but not in master. In 4.3, it finds the correct Voron2 quality profiles. In master, it doesn't find any available quality profile.

ninovanhooff:
As discussed yesterday with the team, the support for these 'simplified' profiles was dropped in 4.4 because it wasn't used at the time. It was also suggested that for optimal results you will want to tweak the individual profiles anyways (this is not required for this PR to be accepted). Shall we merge https://github.com/Ultimaker/Cura/pull/6666/commits/9b583be6ef3ac7364693d739ced52ee61566eaad then, @LipuFei ?

LipuFei:
@ninovanhooff We can merge this PR but without the commit that removes the extra quality profiles.

LipuFei:
Hi @Fulg , sorry for the extra hassle. Could you revert commit f71b234cf7ed27a4873de989c9750961c591c816? For the newer Cura versions we still need those profiles to make it work.

ninovanhooff:
@Fulg @LipuFei : No need, I just merged 9b583be. This PR can be closed. No worries, @Fulg is still attributed for his work, see for example commit https://github.com/Ultimaker/Cura/commit/dd477fe1367f36152e79f92e935b73a139e3fc09

Fulg:
Nice, thank you very much!

PR #6668, opened 2019-11-16
Updated pt_BR strings for Cura 4.4.

Patola:
Updated pt_BR strings for Cura 4.4.


rburema:
Great! Thanks for your work. :-)
There where a few updates to Uranium as well I thought, though?

PR #6718, opened 2019-11-26
Fixed a typo in the title

iLyngklip:
I noticed a typo in the title of the plugin, so that is fixed. I also
added more info to the description.


Ghostkeeper:
Honestly we should probably remove the "TurnOn" setting. But yeah aside from that, good find!

PR #6722, opened 2019-11-27
CURA-6978/Added the project name in Window Title

DonExo:
Added the project name in Window Title.

### DIFF
--- a/resources/qml/Cura.qml
+++ b/resources/qml/Cura.qml
@@ -21,7 +21,8 @@ UM.MainWindow
     id: base
 
     // Cura application window title
-    title: catalog.i18nc("@title:window", "Ultimaker Cura")
+    title: PrintInformation.jobName + " - " + catalog.i18nc("@title:window", "Ultimaker Cura")
+
     backgroundColor: UM.Theme.getColor("viewport_background")
 
     UM.I18nCatalog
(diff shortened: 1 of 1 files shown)


BagelOrb:
Idea is conform almost any other program, format corresponds with how other programs do it as well.

Code looks good; simple oneliner without any risk.

I vote to merge this right away.

DonExo:
FYI @fieldOfView , since you are the creator of https://github.com/fieldOfView/Cura-VersionInTitlebarPlugin

fieldOfView:
"there's a plugin for that"

It would be nice if this could use the application display name (Application.getApplicationDisplayName()) instead of hardcoding Ultimaker Cura.

BagelOrb:
> It would be nice if this could use the application display name (Application.getApplicationDisplayName()) instead of hardcoding Ultimaker Cura.

I don't see why that makes this PR better. It only introduces a small extra risk in order for some benefit which is not clearly described here.

rburema:
> ... some benefit ...

@BagelOrb If we ever decide to grey-label Cura.

BagelOrb:
What I'm saying is that we should make changes for "grey-labeling' when we decide to implement or merge 'grey-labeling' - not in this PR.

PS: please don't interpret my comments as required work; I'm just making a principled argument here.

rburema:
@BagelOrb Hardcoding stuff comes with it's own risks, and is generally considered a code smell.

rburema:
@DonExo For future reference: If we've got other PR's related to this one, we usually put it here.

PR #6724, opened 2019-11-27
Fixed typo

translucentfocus:
Only correcting in master; may want to consider correcting in other branches depending your opinion


nallath:
Thanks for the fix :)

PR #6747, opened 2019-12-02
Spanish i18n fix fillament weight

juanluisvaladas:
"Anchura" means width not weight. Weight is translated as "peso"


Ghostkeeper:
Nice. I like to see more pesos.

PR #6751, opened 2019-12-03
CURA-6994/Prevent deleting comments from Start/End G-codes

### LINKED ISSUE #6689: [4.4] Comments removed from Start/End G-codes after update
**Application version**
4.4 after updating from 4.3

**Platform**
Win10

**Printer**
Custom built

**Reproduction steps**
1. Put some G-code descriptions as new line comments in Start/End G-codes in 4.3
2. Update from 4.3 to 4.4 (uninstalled 4.3 beforehand, kept configuration files)
3. Open machine settings window again

**Actual results**
MOST comment only lines (lines starting with ";") were removed. Comments following actual G-codes ("G28 ;home" for example) remain unchanged.
Commented first line didn't get removed.

No more change between restarts, only once after update.

**Expected results**
Whole G-code scripts remain unchanged.

**Additional information**
https://github.com/Ultimaker/Cura/issues/1237 fixed this issue before, only difference is this doesn't happen between restarts.

My End G-code is as follows:
```
;<END>
G0 Z305 F1500 ; lower bed
M140 S0
M104 S0 ; M104 T0 S0
M124 S0 ; M104 T1 S0
M134 S0 ; M104 T2 S0
M107
;<DOCK>
T0
G28 X0 
G28 Y0
M84
;</END>
```

Only `;<DOCK>` and `;</END>` got removed. `;<END>` at the top remained.

`;<DOCK>` and other brackets in comments are used by my external post processing script as fix for #5861 and other patches for Smoothieware.

DonExo:
Prevent deleting comments from Start/End G-codes by changing the `comment_prefixes` in ConfigParser" -m "CURA/6994".

Note: We should add this in every future VersionUpgrade.

Fixes #6689

### DIFF
--- a/plugins/VersionUpgrade/VersionUpgrade41to42/VersionUpgrade41to42.py
+++ b/plugins/VersionUpgrade/VersionUpgrade41to42/VersionUpgrade41to42.py
@@ -239,7 +239,7 @@ class VersionUpgrade41to42(VersionUpgrade):
     #
     #   This renames the renamed settings in the containers.
     def upgradeInstanceContainer(self, serialized: str, filename: str) -> Tuple[List[str], List[str]]:
-        parser = configparser.ConfigParser(interpolation = None)
+        parser = configparser.ConfigParser(interpolation = None, comment_prefixes=())
         parser.read_string(serialized)
 
         # Update version number.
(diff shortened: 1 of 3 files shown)


ninovanhooff:
@Ghostkeeper The fix will prevent any text after a ; in a `ini` file for instance containers to be removed when loaded. I guess this is okay but can we have your second-opinion about that?

Ghostkeeper:
@ninovanhooff Yes that's a good fix. We had this problem originally as well in our normal profile loading code and had to disable all of the comments and interpolation features of configparser in order to load them properly. Seems we just missed a spot, here in the version upgrade.

Issue #6646, opened 2019-11-10
Change at Z; Allow printer speed to have a specified value in mm/s.


photodude:
**Is your feature request related to a problem? Please describe.**
In the post-processing change at Z 5.1.1 settings you can set the speed and printer speed as a percentage. It would be very helpful for speed tuning tests to be able to specify a specific value and not a percentage. Think temperature tower but with print speeds as specified values.

see image showing the current percent values
![Capture](https://user-images.githubusercontent.com/10253980/68538541-6de28a80-0333-11ea-8d3f-4e5b2967b9ff.PNG)

**Describe the solution you'd like**
give an option for specified speed settings 

End result should allow very clear progression of exactly specified values something like this for the set of postprocessing gcode from change at z. Like a temperature tower but with print speeds as the specified values i.e. a speed tower.
initial print speed 40mm/s (0mm on model)
- increase by 5mm
  - new print speed 50mm/s
- increase by 5mm (now at 10mm on model)
  - new print speed 60mm/s
- increase by 5mm (now at 15mm on model)
  - new print speed 70mm/s
- increase by 5mm (now at 20mm on model)
  - new print speed 80mm/s
- increase by 5mm (now at 25mm on model)
  - new print speed 90mm/s
- increase by 5mm (now at 30mm on model)
  - new print speed 100mm/s

**Describe alternatives you've considered**
1.  replace existing percent speed with exact values
2.  show both options and calculate the other option
  - i.e. calculate speed from entered percentage, or if speed is entered calculate the percentage

**Affected users and/or printers**
users who like to tune their slicer settings to get the most out of their printer
manufacturers developing printer profiles for their printers

**Additional context**
a random Thingiverse model for this speed tower concept
https://www.thingiverse.com/thing:3249901
we just lack a very precise way of specifying the speeds for the changes.

rburema:
Hi @photodude ,
Sadly, we've chosen to defer this for now, though pull requests are of course welcome.
cheers, Remco.

photodude:
@rburema point me in the right direction and I'll make an attempt to implement the change. I need to know where to look, what to look for, and what Unit test changes would need to be made/implemented. I consider my python skills very introductory as I've only used it for numerical methods with FEA in jupyter notebook.

rburema:
The relevant file is: https://github.com/Ultimaker/Cura/blob/master/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py

[...]

PR #6778, opened 2019-12-09
ChangeAtZ 5.2.0

novamxd:
**Primary Objectives**

- Added support for changing Retract Length and Speed. 
- Fixed issue of cumulative ChangeZ mods so they can now properly be stacked on top of each other. 

**Other Changes**

- Added comments.
- Applied code refactoring to clean up various coding styles. 
- Broke up functions for clarity. 
- Removed layer spread option in favor for fixing ChangeZ stacking
- Split up class so it can be debugged outside of Cura.


plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
```diff
@@ -4,496 +4,954 @@
...
+					"label": "Change Retract Length",
+					"description": "Changes the retraction length during print (M207)",
+					"type": "bool",
+					"default_value": false
+				},
+				"caz_retractlength": {
+					"label": "Retract Length",
+					"description": "New Retract Length (units)",
+					"unit": "units",
+					"type": "int",
```
willt: f

willt: type should be float not int

novamxd: What about the feedrate? Float as well?

novamxd: I found a few more as well.

novamxd: Done. Have a look and let me know if you need anything else!

photodude:
@novamxd This is some great work. Hopefully, this makes working on [the feature request I'm looking for](https://github.com/Ultimaker/Cura/issues/6646) easier. So far trying to dig into the code for ChangeAtZ 5.1.1 has been a bear to read and decipher to plan out changes to make the feature improvement. I look forward to improvements like this one.

novamxd:
@photodude Glad you like it. I think your feature request may be difficult to accomplish if Cura handles it anything like retract length, which instead of using M207 (set retract length) and G10/G11, it using normal linear move commands without extrusions. This probably makes it more compatible with printers, but harder to modify in post processing.

Technically you could do something similar with what you want with 5.2.0 now that cumulative effects are disabled. If you set your base speed in your print profile, you can use ChangeZ 5.2.0 to incrementally increase it by a percent.

So if your initial speed was 40mm/s, your first ChangeZ increase would be 125% to 50mm/s. Your next would be 150% for 60mm/s. The challenge, however, is that both print speed adjustments increase all manner of print speed. This includes feed rates, retraction rates, and jerk technically. So I don't know if the effect will be what you're looking for.

photodude:
@novamxd sounds it would be great if the ChangeAtZ could be applied at slicing instead of post-processing the gcode. But that's probably a very complex technical hurdle.

It sounds like changing the print speed adjustments is a royal mess if it's also affecting the feed rates, retraction rates, and jerk. If that's the case, it basically makes attempting to tune a printer with a speed test changing speed with ChangeAtZ an exercise in futility; as you would be modifying too many things to make the test useful. (might also explain the extreme under extrusion I was getting in my tests).

I'll be following the changes here in any case.

nallath:
> @novamxd sounds it would be great if the ChangeAtZ could be applied at slicing instead of post-processing the gcode. But that's probably a very complex technical hurdle.

Yup, that would be pretty hard to do.

nallath:
It seems that you set up your python IDE incorrectly (at least according to our specs ;)) since we use 4 spaces instead of tabs.  

Other than that, it's a big step up from the mess that it was beforehand! Good job :)

photodude:
@nallath do you not have code style checks in the continuous integration for PRs implemented? I worked a bunch on improving the code style checks for Joomla. CI code style checks on PRs went a long way to improving the overall codebase. Of course, there is a lot of code in the codebase which doesn't comply with the code style checks.  What doesn't comply are listed as exceptions to the checks because it takes a lot of review and work to clean everything up from when code style was enforced only by reviewers. It also requires taking lots of small bites at cleaning things up to keep from breaking all the open PRs (Joomla has hundreds of open PRs).

I hope Cura would consider implementing some type of code style checker in the CI.

nallath:
We do have code style checks, but for some reason, we never had to check for tabs vs spaces (since usually python just flat out refuses to do anything if you mix them up).

photodude:
@nallath ya nothing is worse than files with mixed tabs and spaces.

Ghostkeeper:
The code style checks are not being performed as part of the CI since we had too big of a legacy code base and a mixture of languages. It's 2.5 million lines of code in total and we'd need a big refactor and break of the history to fix them.

It only checks the typing as part of the CI system.

novamxd:
@Ghostkeeper @nallath 

I'll switch it to use spaces. As a personal preference I don't, so I have my IDEs configured to use tabs.

novamxd:
Code style updated. Switched to spaces.

rburema:
Hi!

So I've dived a bit further into review and I found a few issues:

Big concerns:
 - Since the scripts are saved to projects, any older projects opened with a Cura with this version of the script will have to be upgraded some way, since the 'single layer' option doesn't exist any more.
 - Since you can't input formula's as script values, 'single layer'-like changes will have to be updated _manually_ each time the user changes the settings, which happens automatically right now.
 - Changing M140 to M190 isn't always a good idea, M190 waits until the build-plate has heated up, M140 continues while doing other stuff.
   (In general, there are a couple of other G/M codes that are inserted differently, which should also be checked to work for the multiple gcode flavours.)

Smaller concerns:
 - It removes the M117 'print to display' feature that was added by another pull request sometime last year.
 - (We _may_ be able to cut this out, since this _may_ not have worked correctly to begin with, but:) Without a single layer change, the M605/M606 commands that where apparently necessary for the Ultimaker 2 can't be inserted.
 - While the feature is gone, as described in the PR, the drop-down of whether to 'keep value' or 'single layer'.
 - 'Unit' or 'unit/second' as unit of measurement.
 - M207 in description, even though the printer might not actually do retraction (parameters) that way.
 - ;[CAZD Isn't as clear as the original description in the gcode comments.
 - This is in the original code as well, so this isn't needed for acceptance, but: occasionally, there's a check whether a line contains "G1", which also might detect any other G-codes _starting_ on G1.
 - (This could be on our side, I didn't go too deep into this:) The mesh comment now just says 'Object 1' instead of '<filename>.stl'.

novamxd:
Hello!

Thanks for the feedback, my **replies** are below.

Since the scripts are saved to projects, any older projects opened with a Cura with this version of the script will have to be upgraded some way, since the 'single layer' option doesn't exist any more.
**WH**: Like the whole .py file or the config? The config variables are left the same, so the only thing that won't really work is the layer spread option for those who use it. I'd probably suggest a warning, in the UI maybe? Do you have details around the usage of the layer spread?

Since you can't input formula's as script values, 'single layer'-like changes will have to be updated manually each time the user changes the settings, which happens automatically right now.
**WH**: Can you clarify as to the usage here? As far as I can tell, the config values before and after are simple primitive types (floats/ints), so I think I must be missing something.

Changing M140 to M190 isn't always a good idea, M190 waits until the build-plate has heated up, M140 continues while doing other stuff.
(In general, there are a couple of other G/M codes that are inserted differently, which should also be checked to work for the multiple gcode flavours.)
**WH**: Swapping it out for the non-waiting variants is easy enough. Comparing with the RepRap wiki, M140 and M104 seems to widely adopted. The only concern I can see is support for individual extruder temps, which you can only do so much for.

Smaller concerns:

It removes the M117 'print to display' feature that was added by another pull request sometime last year.
**WH**: Easy enough to add back.

(We may be able to cut this out, since this may not have worked correctly to begin with, but:) Without a single layer change, the M605/M606 commands that where apparently necessary for the Ultimaker 2 can't be inserted.
**WH**: While on the surface this would appear useful, it's a bit misleading because it doesn't backup/restore all the settings, and it only benefits UM2. I can add it back, but it sounds like it's not probably not worth it.

While the feature is gone, as described in the PR, the drop-down of whether to 'keep value' or 'single layer'.
**WH**: Layer spread is gone. Doing a single layer or not is still there.

'Unit' or 'unit/second' as unit of measurement.
**WH**: Will set to mm or mm/s.

M207 in description, even though the printer might not actually do retraction (parameters) that way.
**WH**: Will change. I initially planned on using M207 but in my experience Cura uses G-codes for retraction instead. The 5.2.0 looks for G-codes.

;[CAZD Isn't as clear as the original description in the gcode comments.
**WH**: It's not supposed to be. It's designed to be added when changed and removed if reverted by a later ChangeZ.

This is in the original code as well, so this isn't needed for acceptance, but: occasionally, there's a check whether a line contains "G1", which also might detect any other G-codes starting on G1.
**WH**: Good point. I'll update the checks to look for the specific commands G0 and G1.

(This could be on our side, I didn't go too deep into this:) The mesh comment now just says 'Object 1' instead of '.stl'.
**WH**: Let me know if you find anything.

Ghostkeeper:
> Like the whole .py file or the config?

Just the config. While it fails gracefully, it does produce errors in the log if the setting is not removed. It can be removed though, via a version upgrade step.

> Can you clarify as to the usage here? As far as I can tell, the config values before and after are simple primitive types (floats/ints), so I think I must be missing something.

What he meant is that e.g. you want the print temperature to be 190 degrees for just a single layer because it has a big bridge or something that's best printed a bit colder. Previously you could do that by setting "Behaviour" to "Single Layer". With this change you can still do that by setting the temperature to 190 degrees in one layer and back to 200 degrees for the next layer, but then if you change the normal printing temperature setting (like from 200 to 202) you have to adjust the temperature in the script along with it, or it'll go back to 200 instead of 202.

> While on the surface this would appear useful, it's a bit misleading because it doesn't backup/restore all the settings, and it only benefits UM2. I can add it back, but it sounds like it's not probably not worth it.

Doesn't hurt either though. Most people won't even see it or know that it is supposed to restore settings. And the people that do see it by inspecting the g-code will understand that the g-code is meant to function for all sorts of printer models.

> Layer spread is gone. Doing a single layer or not is still there.

Oh, good. This was a misunderstanding on our part then. The single layer is i.m.o. a much more useful feature than the layer spread. That makes our previous issue with the setting formula thing moot.

> Will change. I initially planned on using M207 but in my experience Cura uses G-codes for retraction instead. The 5.2.0 looks for G-codes.

Keep in mind that Cura can either output `G10` and `G11` for retractions (if `machine_firmware_retract` is enabled), or G1 moves with negative delta E's.

> Good point. I'll update the checks to look for the specific commands G0 and G1.

The `Script` class has a built-in helper function called `getValue` that will parse this for you and is safe against this problem (`G1` vs. `G10` and such). Try that!

jcharaoui:
Just for the record, I was having really weird issues with the current version of the plugin (5.1.1), trying to change print speed at a single layer. With this version the problems are gone and the GCODE modifications are as expected. As a bonus this PR seems to correct an n-1 error when specifying the Layer no. So :+1: from me, FWIW

novamxd:
Hey Guys,

I will hopefully be addressing these items this weekend and should have an updated release for the weekend as well.

Also that's great to hear @jcharaoui. =D

Cheers!

rburema:
@novamxd Good to hear!

In general, I want to mention to people not to expect this in 4.5, since we're _way_ past that window. If this is going to be merged (which seems likely if the proposed changes are in), it'll probably be for 4.6 instead.

novamxd:
@rburema When do I need to have this ready for 4.6? I got a bunch done over the long weekend but I now have a bunch of testing and tuning to do.

fieldOfView:
Cura releases are generally 2 months apart. That means that in two months we will already have passed the window for inclusion in Cura 4.6. Features need to be merged the latest a week before the release of a beta. But you also have to account for the review process taking time.

Realistically, you have about a short month from now.

Ghostkeeper:
We started a feature freeze the beginning of this week, so this would have to be for the next release then.

novamxd:
Hey Guys,

I'm not dead! Went to Japan and then was swamped at work. I finished my updates to the plugin. Key updates I've made:

- Added additional option for disabling all retract options
- Added an enabled flag, allowing users to enable/disable ChangeAtZ layers at will without removing them
- Added support for firmware based retractions.
- Added support for outputting changes to LCD (untested).
- Created GCodeCommand class for better detection of G1 vs G10 or G11 commands, and accessing arguments.
- Fixed issue of properly restoring previous values in single layer option. Basically for anything other than print speed adjustments the previous values wouldn't be restored correctly.
- Improved wording of Single Layer vs Keep Layer to better reflect what was happening.

**Important Notes**

> It removes the M117 'print to display' feature that was added by another pull request sometime last year.
> WH: Easy enough to add back.

After some investigation it appears previous the code would output to the LCD screen only when doing layer spread, which was removed. In an effort to add this back, I instead changed the plugin to output the changes to the LCD screen when the target layer or Z index was hit. While my printer does have an LCD screen, it seems to ignore the M117 command, so I haven't been able to validate that it works as expected and I couldn't find any documentation around line breaks or character limits.

> > While on the surface this would appear useful, it's a bit misleading because it doesn't backup/restore all the settings, and it only benefits UM2. I can add it back, but it sounds like it's not probably not worth it.
> 
> Doesn't hurt either though. Most people won't even see it or know that it is supposed to restore settings. And the people that do see it by inspecting the g-code will understand that the g-code is meant to function for all sorts of printer models.

So in the end I didn't put back the UM2 specific commands and instead just built a universal value tracking/restore system instead, which is what the UM2 specific commands was for.

bondus:
Nice to see this old hack get some love, it looks good.
                if "Z10.8" in line:
                    derp = True
The feature to spread a change over a number of layers, or to get a stepped change is gone. Those feature only worked sporadically anyway.

novamxd:
@bondus 
> if "Z10.8" in line:
> derp = True

:neutral_face: *ahem* I'll just remove that...

> The feature to spread a change over a number of layers, or to get a stepped change is gone. Those feature only worked sporadically anyway.

Yea the idea now is you just layer the ChangeZ mods. So you create one for every layer/Z you want to target and set whether you want it to apply to that specific layer or subsequent layers. Being able to enable/disable the mods is also nice.

ninovanhooff:
Hi there, I think this code looks great; great job of cleaning up those terribly nested conditions. I see the debug function contains your local path. That path should not be there of course and the entire debug function is unused. The file itself is quite big and contains a class that would ideally be in its own file. Unfortunately we do not support multiple files or unit tests for a postprocessing script. 
I think I will just accept this and remove the debug code myself. We'll test it a bit internally first. Thanks for your effort!

Hirsute:
Yay! I'm excited for this. Thank you novamxd!

Skuzee:
What is the reason behind removing layer spread? It was one of the only features I used changeatz for. :( I usually use it to calibrate my printer/filaments to test a range of speeds/temps/etc. I understand if it was causing some type of bug, but I would really like to avoid manually entering a bunch of individual heights and values. Thanks!

pikolo84:
I want to use layer spread to lower bed temp from 90ºC to 40ºC by 1ºC per layer. How can i do it without enter 50 llines of changeatZ?

PR #6785, opened 2019-12-10
Create AddGcodeAtHeight

owendare:
Post processing script to add G or M codes at specific height.
Intended for use with Duet or other RepRap Firmware based printers that use G code for configuration and tuning.


ninovanhooff:
One serious issue with this PR is that the filename should end in .py :)

Other than that, I think a nice addition would be to allow users to enter height as a layer number, because layer numbers are what is displayed in the Preview stage. 
See https://github.com/Ultimaker/Cura/blob/ef4ee5d6b6d83e6800451b925bad57e897879bdc/plugins/PostProcessingPlugin/scripts/ColorMix.py
For an example on how to do that.

ninovanhooff:
Developers: I created ticket CURA-7212 to track this PR

ninovanhooff:
Considering no response received after 20 days, we considered this PR as-is. The units-option can be left to those who feel like implementing it.

ninovanhooff:
Discussed this with another team member. We feel this is too much of a catch-all solution which will make it hard to pinpoint weird print results when used or accidentally left enabled. Consider an addition to the changeAtZ script.

fieldOfView:
And "Search and Replace" is not a catch-all?

PR #6803, opened 2019-12-13
Sentry crash integration

nallath:
Change the crash logging to sentry.


plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,43 @@
...
+    def log(self, log_type: str, message: str) -> None:
+        level = self._translateLogType(log_type)
+        if level is None:
+            if message not in self._show_once:
+                level = self._translateLogType(log_type[0])
+                if level is not None:
+                    self._show_once.add(message)
+                    add_breadcrumb(level=level, message=message)
+        else:
+            add_breadcrumb(level=level, message=message)
```
rburema: This'll only actually _send_ anything when we do a crash report, right?

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,43 @@
...
+        elif log_type == "i":
+            level = "info"
+        elif log_type == "c":
+            level = "fatal"
+        elif log_type == "e":
+            level = "error"
+        elif log_type == "d":
+            level = "debug"
+
+        return level
```
rburema: EOF, add newline plz.

plugins/SentryLogger/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2019 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+from . import SentryLogger
+
+
+def getMetaData():
```
rburema: typing?

plugins/SentryLogger/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    return {}
+
+
+def register(app):
```
rburema: typing?

cura/CrashHandler.py
```diff
@@ -9,11 +9,11 @@
...
 import json
-import ssl
-import urllib.request
-import urllib.error
+from typing import cast
 
-import certifi
+from sentry_sdk.hub import Hub
+from sentry_sdk.utils import event_from_exception
+from sentry_sdk import configure_scope
```
rburema: we should make sure Lipu's PR (add `sentry_sdk`) ends up in the build env. :-)

cura_app.py
```diff
@@ -9,16 +9,24 @@
...
 parser = argparse.ArgumentParser(prog = "cura",
                                  add_help = False)
 parser.add_argument("--debug",
                     action="store_true",
                     default = False,
                     help = "Turn on the debug mode by setting this option."
                     )
+
 known_args = vars(parser.parse_known_args()[0])
+sentry_sdk.init("https://5034bf0054fb4b889f82896326e79b13@sentry.io/1821564",
```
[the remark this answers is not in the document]
nallath: They know that by grace of those keys. So yes, this should remain the same in the forseable future.

cura_app.py
```diff
@@ -9,16 +9,24 @@
...
+                release = "cura%s" % ApplicationMetadata.CuraVersion,
```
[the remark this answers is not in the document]
nallath: Well, releasing something without version numbers in lockstep is going to be confusing regardless, which is also why I think we should not do that in the first place. So as far as I'm concerned, the versions will be released in lockstep.

cura_app.py
```diff
@@ -9,16 +9,24 @@
...
+                default_integrations = False,
+                max_breadcrumbs = 200)
```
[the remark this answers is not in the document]
nallath: The max breadcrumbs? Yeah, I had to put something there. This basicly limits the amount of log entries that we send to 200. The default was lower though.

plugins/SentryLogger/SentryLogger.py
nallath: Yeap. This is just to collect the data so it's there if we want to send something.

cura/CrashHandler.py
nallath: He didn't do it with a PR, so it's already in the build env.

cura_app.py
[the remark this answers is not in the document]
rburema: I mean, can someone just mess with us through this?

cura/CrashHandler.py
rburema: Nice.

cura_app.py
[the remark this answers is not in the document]
rburema: Fair enough.

cura_app.py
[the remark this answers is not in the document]
nallath: Yeah, but that was already the case with the current setup. All of the code is in python, so that's always a risk.

Ghostkeeper:
See issue CURA-7050.

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,52 @@
...
+    #   \param message String containing message to be logged
+    def log(self, log_type: str, message: str) -> None:
+        level = self._translateLogType(log_type)
+        message = self._pruneSensitiveData(message)
+        if level is None:
+            if message not in self._show_once:
+                level = self._translateLogType(log_type[0])
+                if level is not None:
+                    self._show_once.add(message)
+                    add_breadcrumb(level=level, message=message)
```
Ghostkeeper: Code style: Spaces around binary operators.

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,52 @@
...
+
+from UM.Logger import LogOutput
+from typing import Set
+from sentry_sdk import add_breadcrumb
+from typing import Optional
+import os
+
+home_dir = os.path.expanduser("~")
+
+class SentryLogger(LogOutput):
```
Ghostkeeper: Would be good to add some documentation here as to what this class (or module) does.

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,52 @@
...
+        level = self._translateLogType(log_type)
+        message = self._pruneSensitiveData(message)
+        if level is None:
+            if message not in self._show_once:
+                level = self._translateLogType(log_type[0])
+                if level is not None:
+                    self._show_once.add(message)
+                    add_breadcrumb(level=level, message=message)
+        else:
+            add_breadcrumb(level=level, message=message)
```
Ghostkeeper: Code style: Spaces around binary operators.

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @staticmethod
+    def _pruneSensitiveData(message):
+        if home_dir in message:
+            message = message.replace(home_dir, "<censored_path>")
```
Ghostkeeper: Maybe better to say what the path was, such as `<user_home>`?

plugins/SentryLogger/SentryLogger.py
```diff
@@ -0,0 +1,52 @@
...
+        if log_type == "w":
+            level = "warning"
+        elif log_type == "i":
+            level = "info"
+        elif log_type == "c":
+            level = "fatal"
+        elif log_type == "e":
+            level = "error"
+        elif log_type == "d":
+            level = "debug"
```
Ghostkeeper: A bit of a style thing, but I think it looks nicer with a dictionary look-up:

```suggestion
            levels = {
                "w": "warning",
                "i": "info",
                "c": "fatal",
                "e": "error",
                "d": "debug
            }
            return levels.get(log_type)
```

cura/CrashHandler.py
```diff
@@ -194,6 +196,17 @@ def _informationWidget(self):
         self.data["os"] = {"type": platform.system(), "version": platform.version()}
         self.data["qt_version"] = QT_VERSION_STR
         self.data["pyqt_version"] = PYQT_VERSION_STR
+        self.data["locale_os"] = locale.getlocale(locale.LC_MESSAGES)[0] if hasattr(locale, 'LC_MESSAGES') else locale.getdefaultlocale()[0]
```
Ghostkeeper: Code style: Double quotes for strings.

cura/CrashHandler.py
```diff
@@ -211,6 +224,30 @@ def _getOpenGLInfo(self):
...
+                active_machine_manufacterer = global_stack.definition.getMetaDataEntry("manufacturer", "unknown")
+        except:
+            pass
+
+        with configure_scope() as scope:
+            scope.set_tag("opengl_version", opengl_instance.getOpenGLVersion())
+            scope.set_tag("gpu_vendor", opengl_instance.getGPUVendorName())
+            scope.set_tag("gpu_type", opengl_instance.getGPUType())
+            scope.set_tag("active_machine", active_machine_definition_id)
+            scope.set_tag("active_machine_manufacterer", active_machine_manufacterer)
```
Ghostkeeper: Spelling: `manufacterer` -> `manufacturer`

cura/CrashHandler.py
```diff
@@ -211,6 +224,30 @@ def _getOpenGLInfo(self):
 
         self.data["opengl"] = {"version": opengl_instance.getOpenGLVersion(), "vendor": opengl_instance.getGPUVendorName(), "type": opengl_instance.getGPUType()}
 
+        active_machine_definition_id = "unknown"
+        active_machine_manufacterer = "unknown"
```
Ghostkeeper: Spelling: `manufacterer` -> `manufacturer`

cura/CrashHandler.py
```diff
@@ -194,6 +196,17 @@ def _informationWidget(self):
...
+        self.data["locale_cura"] = self.cura_locale
+
+        with configure_scope() as scope:
+            scope.set_tag("qt_version", QT_VERSION_STR)
+            scope.set_tag("pyqt_version", PYQT_VERSION_STR)
+            scope.set_tag("os", platform.system())
+            scope.set_tag("os_version", platform.version())
+            scope.set_tag("locale_os", self.data["locale_os"])
+            scope.set_tag("locale_cura", self.cura_locale)
+            scope.set_tag("is_enterprise", ApplicationMetadata.IsEnterpriseVersion)
```
Ghostkeeper: Currently in our crash report screen we display the following information only:

* Cura version
* Cura build type
* Platform / OS
* Qt Version
* PyQt Version
* OpenGL version

Should we add the OS locale, Cura locale and isEnterprise to this? Perhaps enterprise or not can be displayed together with the build type?

cura_app.py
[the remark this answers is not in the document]
DonExo: I share the same concern as Remco. I think we need to find a way to hide this value (or at least mask it) somehow, in a .env file  Let's discuss it tomorrow on C&C.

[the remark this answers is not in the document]
nallath: Hiding it won't really make a difference. It's a simple matter of just writing a plugin that prints them again. 
If someone is already going through the trouble of writing a semi-automated script to spam us with crashes, hiding it in an .env file is also not going to stop them :P

cura/CrashHandler.py
ninovanhooff: I'm also kinda missing buildtype still, but it's okay to add that when we start using it from something other than enterprise. development builds

cura_app.py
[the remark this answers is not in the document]
ninovanhooff: Agree that build type should not be part of the version here. They are based on the same commit, which to me is what defines different versions

PR #6813, opened 2019-12-16
Cura-6979/Show changes in packages to sync

DonExo:
This PR adds logic for checking discrepancies in Cloud subscribed packages and Cura installed packages.
It does this check in two cases:
1. On Cura start-up, if the user is logged-in.
2. If Cura is already running and the users logs in.

This is still WIP, I'm still fixing some tiny bits and making it more clear. 

Depends on[ # 552 from Uranium repo](https://github.com/Ultimaker/Uranium/pull/552).


plugins/Toolbox/src/Toolbox.py
```diff
@@ -540,9 +557,7 @@ def getTotalNumberOfMaterialPackagesByAuthor(self, author_id: str) -> int:
 
     @pyqtSlot(str, result = bool)
     def isEnabled(self, package_id: str) -> bool:
-        if package_id in self._plugin_registry.getActivePlugins():
-            return True
-        return False
+        return package_id in self._plugin_registry.getActivePlugins()
```
DonExo: This is just refactor, doesn't relate to the ticket.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -61,13 +62,15 @@ def __init__(self, application: CuraApplication) -> None:
...
             "updates":             [],
+            "installed_packages":   [],
         }  # type: Dict[str, List[Any]]
 
         # Models:
         self._models = {
             "authors":             AuthorsModel(self),
             "packages":            PackagesModel(self),
             "updates":             PackagesModel(self),
+            "installed_packages":  PackagesModel(self),
```
nallath: I think it's better to rename this to prevent confusion. This doesn't hold the installed packages, but the packages the user is subscribed to.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -237,6 +249,11 @@ def _fetchPackageData(self):
         # Gather installed packages:
         self._updateInstalledModels()
 
+    def _fetchUserInstalledPlugins(self):
+        self._prepareNetworkManager()
```
nallath: Nitpicking: The network manager only needs to be prepared if the user is logged in.

plugins/Toolbox/src/Toolbox.py
nallath: It's still a valid change. No idea why it was like this in the first place.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -646,8 +662,10 @@ def _onRequestFinished(self, reply: QNetworkReply) -> None:
                                 Logger.log("e", "Could not find the %s model.", response_type)
                                 break
 
-                            self._server_response_data[response_type] = json_data["data"]
-                            self._models[response_type].setMetadata(self._server_response_data[response_type])
+                            # Workaround: Do not add Metadata for "installed_packages" check JUST YET
```
nallath: Why not? In cases of work-arrounds, it's best to also explain why it's there (eg; I can see what you're doing, but in 3 months from now, no-one will remember the why)

plugins/Toolbox/src/Toolbox.py
```diff
@@ -661,6 +679,27 @@ def _onRequestFinished(self, reply: QNetworkReply) -> None:
                                 # Tell the package manager that there's a new set of updates available.
                                 packages = set([pkg["package_id"] for pkg in self._server_response_data[response_type]])
                                 self._package_manager.setPackagesWithUpdate(packages)
+                            elif response_type == "installed_packages":
+                                user_subscribed = [(plugin['package_id'], plugin['package_version']) for plugin in json_data['data']]
+                                Logger.log("i", "- User is subscribed to {} package(s).".format(len(user_subscribed)))
```
nallath: Why add the "- " in front of the message (and also, should it be info? it feels more like a debug message)

plugins/Toolbox/src/Toolbox.py
DonExo: I'm very much open to name-suggestions!
I've went for the "installed_packages" because locally (in Cura) they are shown as "Installed", but on the Cloud as "Subscribed" to. Anyway, I don't mind renaming them to something more specific, just give me a name.

1. "subscribed_packages" ?
2. "subscribed" ?
3. ...

nallath: I think subscribed_packages would fit the bill best.

plugins/Toolbox/src/Toolbox.py
DonExo: My bad, sorry, should have been more verbose.
Because I wanted to skip the creation of the designated model, because at this point CURA-6979 asks just for the check of discrepancies between Cura and Cloud in installed packages.
I'm probably gonna come back to revise it later when there is the need for showing the packages that are avaialable/ not available to be synced.

plugins/Toolbox/src/Toolbox.py
DonExo: Good catch, must have slipped among some print statements from previous.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -197,6 +200,11 @@ def _onAppInitialized(self) -> None:
             cloud_api_version = self._cloud_api_version,
             sdk_version = self._sdk_version
         )
+        # https://api.ultimaker.com/cura-packages/v1/user/packages
+        self._api_url_user_packages = "{cloud_api_root}/cura-packages/v{cloud_api_version}/user/packages".format(
+            cloud_api_root=self._cloud_api_root,
+            cloud_api_version=self._cloud_api_version,
```
Ghostkeeper: Code style: Spaces around binary operators.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -661,6 +678,27 @@ def _onRequestFinished(self, reply: QNetworkReply) -> None:
                                 # Tell the package manager that there's a new set of updates available.
                                 packages = set([pkg["package_id"] for pkg in self._server_response_data[response_type]])
                                 self._package_manager.setPackagesWithUpdate(packages)
+                            elif response_type == "subscribed_packages":
+                                user_subscribed = [(plugin['package_id'], plugin['package_version']) for plugin in json_data['data']]
```
Ghostkeeper: Code style: Double quotes for strings.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -661,6 +678,27 @@ def _onRequestFinished(self, reply: QNetworkReply) -> None:
...
+                                    sync_message = Message(i18n_catalog.i18nc(
+                                        "@info:generic",
+                                        "\nDo you want to sync material and software packages with your account?"),
+                                        lifetime=0,
+                                        title=i18n_catalog.i18nc("@info:title", "Changes detected from your Ultimaker account", ))
+                                    sync_message.addAction("sync",
+                                                           name=i18n_catalog.i18nc("@action:button", "Sync"),
+                                                           icon="",
+                                                           description="Sync your Cloud subscribed packages to your local environment.",
+                                                           button_align=Message.ActionButtonAlignment.ALIGN_RIGHT)
```
Ghostkeeper: Code style: Space around binary operators.

DonExo:
This PR can be closed as it is superseded by #6845

PR #6814, opened 2019-12-16
Fix PPA crashing on single-extruder printers

konskarm:
The machine manager was leading to a crash when trying to enable
the second extruder in single-extrusion printers, because the check
for the second extruder was not correctly implemented. This commit
fixes that issue by checking if the global stack has the specified
extruder. If it does not, then the function returns while logging the
issue.

CURA-7048

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -894,8 +894,8 @@ class MachineManager(QObject):
 
     @pyqtSlot(int, bool)
     def setExtruderEnabled(self, position: int, enabled: bool) -> None:
-        if self._global_container_stack is None:
-            Logger.log("w", "Could not find extruder on position %s", position)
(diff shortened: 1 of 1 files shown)


cura/Settings/MachineManager.py
```diff
@@ -894,8 +894,8 @@ def forceUpdateAllSettings(self) -> None:
 
     @pyqtSlot(int, bool)
     def setExtruderEnabled(self, position: int, enabled: bool) -> None:
-        if self._global_container_stack is None:
-            Logger.log("w", "Could not find extruder on position %s", position)
+        if self._global_container_stack is None or position not in self._global_container_stack.extruders:
```
nallath: I don't think this will work. Position is an integer and self._global_container_stack.extruders is a dict with a string as the key

PR #6817, opened 2019-12-17
Add lotmaxx printers

sm3dp:


resources/definitions/lotmaxx_sc10.def.json
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "0": "lotmaxx_sc10_extruder_0"
+        }
+    },
+    "overrides": {
+        "machine_name": { "default_value": "Lotmaxx SC-10" },
+        "machine_width": { "default_value": 235 },
+        "machine_depth": { "default_value": 235 },
+        "machine_height": { "default_value": 320 },
+        "machine_head_polygon": { "default_value": [[-32,11],[-32,-32],[28,-32],[28,11]] },
```
LipuFei: The test failed because of this setting. There's no longer `machine_head_polygon` as a setting. You can use `machine_head_with_fans_polygon` instead.

Log from the tests:
```
./tests/Settings/TestDefinitionContainer.py:59: AssertionError
----------------------------- Captured stdout call -----------------------------
[MainThread] UM.Settings.DefinitionContainer._preprocessParsedJson [270]: Unable to override setting machine_head_polygon
=============================== warnings summary ===============================
```

resources/definitions/lotmaxx_sc20.def.json
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "0": "lotmaxx_sc20_extruder_0"
+        }
+    },
+    "overrides": {
+        "machine_name": { "default_value": "Lotmaxx SC-20" },
+        "machine_width": { "default_value": 235 },
+        "machine_depth": { "default_value": 235 },
+        "machine_height": { "default_value": 320 },
+        "machine_head_polygon": { "default_value": [[-32,11],[-32,-32],[28,-32],[28,11]] },
```
LipuFei: Same as above, you can use `machine_head_with_fans_polygon`.

sm3dp: Thanks for your reply.

Ghostkeeper:
Developers, see issue CURA-7075.

nallath:
You didn't use the right extension for the extruders, but i fixed that for you in 19e659735da1e3dfbfc209cbd78d61c621e55d34
I also converted the platform stl to binary (instead of ascii) to decrease the size in 580b544642ffd60f7245180640b08848edff4c82

sm3dp:
Thank you very much!

PR #6825, opened 2019-12-19
CURA-6387 Refactor HTTP request management

LipuFei:
Needs https://github.com/Ultimaker/Uranium/pull/553


ChrisTerBeke:
Cewl. Would be nice to eventually replace the networking calls to account and cloud as well.

LipuFei:
I fixed the merge conflicts.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -576,52 +556,85 @@ def isLoadingComplete(self) -> bool:
     # Make API Calls
     # --------------------------------------------------------------------------
     def _makeRequestByType(self, request_type: str) -> None:
-        Logger.log("d", "Requesting '%s' metadata from server.", request_type)
-        request = QNetworkRequest(self._request_urls[request_type])
-        for header_name, header_value in self._request_headers:
-            request.setRawHeader(header_name, header_value)
+        Logger.log("d", "Requesting [%s] metadata from server.", request_type)
         self._updateRequestHeader()
-        if self._network_manager:
```
DonExo: You are refactoring the `_makeRequestByType` and removing the `_network_manager` here (therefore removing the `_onRequestFinished` slot), replacing it with a callback `_onAuthorsDataRequestFinished`, which is pretty much the same function as the `_onRequestFinished`, BUT is missing few parts, one of which the check if the request_type is "subscribed_packages" (something I've been working on for the last sprint).
Is this left out intentionally or am I missing something?

https://github.com/Ultimaker/Cura/pull/6845/files#diff-a2d4773eeefc7f207c5a6a0d637a2c4fR680
Here's what I'm talking about.

I realize that you might have wanted to use this only for Authors (i.e. Materials), hence naming it `_onAuthorsDataRequestFinished`, but in fact this same callback is called on any call, including those for packages and updates.

plugins/Toolbox/src/Toolbox.py
```diff
@@ -226,38 +218,27 @@ def _onAppInitialized(self) -> None:
         # On boot we check which packages have updates.
         if CuraApplication.getInstance().getPreferences().getValue("info/automatic_update_check") and len(installed_package_ids_with_versions) > 0:
             # Request the latest and greatest!
-            self._fetchPackageUpdates()
+            self._makeRequestByType("updates")
         self._fetchUserSubscribedPackages()
 
-    def _prepareNetworkManager(self):
```
DonExo: If we are deleting this from here, shouldn't we also delete the called functions `_onRequestFinished` and `_onNetworkAccessibleChanged`, since they are not used and just can confuse people around?
They are already in the PackageManager in Uranium.

ninovanhooff: Yeah, it seems these are not used, I agree

plugins/Toolbox/src/Toolbox.py
ninovanhooff: Also a good point. Confused me as well

LipuFei:
@DonExo The long delay issue is fixed in https://github.com/Ultimaker/Uranium/pull/562

PR #6831, opened 2019-12-20
Hopefully Adding the Anet 3d Printer files to the latest version

springtiger:
Hopefully Adding the Anet 3d Printer files to the latest version


konskarm:
Checks are failing because of the Anet ET5 s. It seems that the setting "initial_extruder_nr" defined in the anet3d_et5 s cannot be overridden.

springtiger:
Thank you for pointing it out. Let me check it.

PR #6851, opened 2019-12-24
CURA-6255_incorrect_usage_of__file__

ninovanhooff:
I'm undecided as to whether the creating of a new class is really worth it. Please advise, reviewer.

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDevice.py
+++ b/plugins/USBPrinting/USBPrinterOutputDevice.py
@@ -88,9 +88,7 @@ class USBPrinterOutputDevice(PrinterOutputDevice):
         self._firmware_name_requested = False
         self._firmware_updater = AvrFirmwareUpdater(self)
 
-        from .USBPrinterOutputDeviceManager import USBPrinterOutputDeviceManager
-        plugin_id = USBPrinterOutputDeviceManager.getInstance().getPluginId()
-        plugin_path = PluginRegistry.getInstance().getPluginPath(plugin_id)
+        plugin_path = PluginRegistry.getInstance().getPluginPath("USBPrinting")
         self._monitor_view_qml_path = os.path.join(plugin_path, "MonitorItem.qml")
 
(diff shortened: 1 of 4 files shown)


ninovanhooff:
Will fix the mypy error once we established that we want to merge this change

fieldOfView:
Out of interest, because I use __file__ this way in many of my plugins, why is this incorrect?

nallath:
The description from our internal ticket: 
_"In some places in live code we're using __file__. This is not allowed, because packing the Python source files on Windows gets rid of the file itself and freezes this variable to the path that it had on the build server, causing potential issues about the X: drive not existing. There's 2 places where using __file__ is allowed:

    In code that only runs when running from source (so the source is not frozen).
    In automated tests.

In plug-ins, using __file__ won't cause an immediate problem but it is more defensive to use PluginRegistry.getPluginDirectory instead to get at relative paths. In main code, we should use the Resources class."_

fieldOfView:
From a defensive programming point of view I get this. But functionally, it makes little sense: In both of the cases in this PR, the `__file__` construct is used to create a path relative to the .py file. But if the .py file is packaged in a zip (frozen Windows situation), that relative path is not going to make sense even if you get it from the pluginregistry.

fieldOfView:
`XmlMaterialProfile/PluginInfo.py` is now unused and should be removed.

I would argue that hardcoding the plugin id is a bigger evil than using __file__ here, but that's just my opinion.

nallath:
It's slightly smaller since it solves the issue for people that have an X: disk that's on the network (which slows down a bunch of things).
Since it's only for the built-in plugins that we're doing this (and ID's of plugins should not change without an upgrade script anyway due to the enabled list needing to be updated) it should be okay to do this.

fieldOfView:
> it solves the issue for people that have an X: disk that's on the network 

Does it though? When used in plugins, the __file__ does not get "baked" to the X: drive (or else things would have stopped working ages ago). The X: drive being an issue is not related to using __file__ in plugins, but due to how the frozen files are handled (ie: looking for the file in X: before looking in the python35.zip archive).

But lets not waste more time on discussing this change. There are more important things to worry about.

PR #6878, opened 2019-12-29
Hungarian translation for 4.3 Cura

greatnam:


Ghostkeeper:
We'd like to add these, but we won't show it in the interface unless the translation is completely finished. Since we won't re-release Cura 4.3 any more, to make the translation show up in the interface we'd need you to update the translations for 4.5.

The translations also need to be kept up to date. Every 2 months we get some 50 to 100 new lines to be translated. Can I contact you somewhere to notify when a translation update needs to happen? It's completely voluntary, but we'll remove the translation from the interface if it's no longer complete (the translations still remain in the code though so that someone can complete them later).

Ghostkeeper:
We've gone into a string freeze for 4.5 now. If you can update the template from 4.5, we can merge it in for this release.

Ghostkeeper:
I'll add this partial translation in so that someone else might complete it some day. It won't be seen in the interface though.

PR #6883, opened 2019-12-30
UX enhancements for Prusa i3 MK3S/MK3 and Prusa Mini integration

Krypta1:
Add UX enhancements to Cura for Prusa i3 MK3S/MK3

1. Nozzle size selector (0.25 / 0.4 / 0.6mm)
2. Specific profiles for each nozzle based on the ones available in PrusaSlicer


fieldOfView:
I think it is great to have better profiles for more printers, but this *will* lead to people complaining that their well crafted custom profiles are changed, their saved projects print differently and that they cannot set the nozzle size anymore in the Machine Settings, much like the backlash after merging the creawesome profiles for creality machines.

Krypta1:
These are just premade profiles but you can definitely continue to use custom profiles since Cura is using protected/standard profiles (the ones I'm proposing) and custom profiles which are some kind of unprotected/public profiles and accessible for any printer.

So I think that's not conflicting with custom made configurations except for nozzle size... but you can definitely copy one of the nozzle files and make your custom one !

fieldOfView:
> So I think that's not conflicting with custom made configurations 

You think wrong... :) Cura has layered profiles. Custom profiles are in a layer named "quality changes", the profiles you wrote are in a layer named "quality". With your changes, someone loading their old custom profile will get a mix of your profiles (in "quality") and theirs (in "quality changes"). That is not something you did wrong, but it is the way profiles work in Cura. 

I am not saying this should not be merged, but I am saying this could lead to people complaining that Cura broke their profiles.

nallath:
I recently made a wiki page to explain this (the profile system) a bit better: https://github.com/Ultimaker/Cura/wiki/Profiles-&-Settings

Ghostkeeper:
In order to merge this, we need a couple of changes:
* All old quality types need to be captured in the new profiles, or where old quality types are no longer available, a version upgrade needs to adjust the quality types of users' custom profiles.
* A version upgrade needs to be put in place to move them from their old quality profiles to the new ones (since the old ones are no longer available, this would create corrupt configs).
* Some of the STLs are above our limit of 1MB per build platform. Please decimate them so that they are faster to load, render and download.

PR #6886, opened 2019-12-30
Update nwa3d_a31.def.json

DragonJe:
Updated to remove material setting and allow the profiles to be supported again using Generic PLA.

Thanks!

### DIFF
--- a/resources/definitions/nwa3d_a31.def.json
+++ b/resources/definitions/nwa3d_a31.def.json
@@ -8,7 +8,6 @@
         "manufacturer": "NWA 3D LLC",
         "file_formats": "text/x-gcode",
         "platform_offset": [0, 0, 0],
-        "has_materials": true,
         "has_variants": true,
         "variants_name": "Nozzle Size",
         "preferred_variant_name": "Standard 0.4mm",
--- a/resources/quality/fabtotum/fabtotum_abs_fast.inst.cfg
+++ b/resources/quality/fabtotum/fabtotum_abs_fast.inst.cfg
@@ -9,6 +9,7 @@ type = quality
 quality_type = fast
 weight = -1
 material = generic_abs
+global_quality = True
 
 [values]
 adhesion_type = raft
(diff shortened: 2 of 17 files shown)


DragonJe:
I removed material setting from the A31 and made the qualities global for Fabricator. 

I believe this should fix the issue with the qualities displaying as "not supported." 

The Deltacomb, that I also mentioned in my issue, has global qualities already, so I'm not sure why its profiles are still not supported.

DragonJe:
Updated material settings for the A31 to align with variants

Ghostkeeper:
Did you also want to submit a change to fdm_materials?

DragonJe:
@Ghostkeeper no, because I thought the values in Fabtotum's fdm_materials will override the quality settings. So when their material is changed, it will change to the correct temps, etc. Am I correct to think that?

Ghostkeeper:
No, the quality profiles override the material profiles. Here is some documentation on the precedence order: https://github.com/Ultimaker/Cura/blob/master/docs/Profile%20Stack.png

Note that your changes to me look fine (didn't test them yet though). Just something you might want to take into account since we can't test the actual print quality.

DragonJe:
@Ghostkeeper will these changes be added to version 4.5.0?

Ghostkeeper:
Ah, sorry. I think this one slipped through the cracks. I don't think we'll make it for v4.5 :disappointed: We have a feature freeze for that now.

It's good that you mention it again though. Sorry again.

DragonJe:
Ok, thanks for the quick reply. I'll keep an eye out for it!

nallath:
@DragonJe Could you resolve the issues mentioned by Ghostkeeper? If we don't get a response, we will close this pull request.

DragonJe:
> @DragonJe Could you resolve the issues mentioned by Ghostkeeper? If we don't get a response, we will close this pull request.

Hi, yes! When COVID, this completely escaped my mind. I will work on it next week.

nallath:
@DragonJe Another friendly reminder :)

PR #6892, opened 2019-12-31
CURA-6959_collapse_categories

ninovanhooff:
Adds a menu item in the Custom Print Settings list to collapse all categories

Does not work when a search is active. See Accompanying Uranium PR.

### DIFF
--- a/resources/qml/Settings/SettingCategory.qml
+++ b/resources/qml/Settings/SettingCategory.qml
@@ -30,11 +30,11 @@ Button
             {
                 return UM.Theme.getColor("setting_category_disabled")
             }
-            else if (base.hovered && base.checkable && base.checked)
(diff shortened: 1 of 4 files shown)


resources/qml/Settings/SettingCategory.qml
```diff
@@ -30,11 +30,11 @@ Button
             {
                 return UM.Theme.getColor("setting_category_disabled")
             }
-            else if (base.hovered && base.checkable && base.checked)
+            else if (base.hovered && base.expanded)
             {
                 return UM.Theme.getColor("setting_category_active_hover")
             }
-            else if (base.pressed || (base.checkable && base.checked))
+            else if (base.pressed || (base.expanded))
```
rburema: Those brackets aren't necessary any more `(base.expanded)`.

PR #6901, opened 2020-01-02
Speed up layerview

nallath:
Since python is bad at looping, only re-calculate the start/end index when it's changed (instead of doing this for each render)


rburema:
I could get into a situation where the head is trailing the lines its supposed to lay:
 - slice
 - set to a range of layers (I used a single layer)
 - leave the simulation in the middle of the layer
 - (change something about the object so you can) reslice
 - in the new view, the head trails the lines being laid down

PR #6906, opened 2020-01-02
SentrySDK: Turn on deep integration on demand

thopiekar:
Only whenever the sentry_sdk module is there functions of this module will be used.
The only changes, which were needed to be made, are done on cura_app.py and cura.CrashHandler.
Whenever the module is not available, it's functions will be omitted.

The if-clauses could happen earlier, but this at least the bare minimum, and, to be honest, on Ultimaker's distribution it won't speed up anything.
I expect the if-clause to take the same amount of runtime sooner or later. The check is the same and it should be on Ultimaker's distribution always be "True".

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -12,9 +12,13 @@ import json
 import locale
 from typing import cast
 
-from sentry_sdk.hub import Hub
-from sentry_sdk.utils import event_from_exception
-from sentry_sdk import configure_scope
+try:
+    from sentry_sdk.hub import Hub
+    from sentry_sdk.utils import event_from_exception
+    from sentry_sdk import configure_scope
+    with_sentry_sdk = True
+except ImportError:
+    with_sentry_sdk = False
 
 from PyQt5.QtCore import QT_VERSION_STR, PYQT_VERSION_STR, QUrl
 from PyQt5.QtWidgets import QDialog, QDialogButtonBox, QVBoxLayout, QLabel, QTextEdit, QGroupBox, QCheckBox, QPushButton
@@ -66,8 +70,9 @@ class CrashHandler:
         if has_started and exception_type in skip_exception_types:
             return
 
-        with configure_scope() as scope:
-            scope.set_tag("during_startup", not has_started)
+        if with_sentry_sdk:
+            with configure_scope() as scope:
+                scope.set_tag("during_startup", not has_started)
 
         if not has_started:
             self._send_report_checkbox = None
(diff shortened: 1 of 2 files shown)


nallath:
But why not use the sentry SDK? It's a rather big source of bug information for us (since a lot of people seem to be completely uninterested in writing bug reports, which is understandable but rather frustrating)

thopiekar:
Well, first aspect would be privacy. The PR here is quick and dirty, but prevents sending data not only to Ultimaker (what I don't care much about), but sending it to a third-party.
Theoretically the data might be encrypted, but practically they could monitor the data as man-in-the-middle.
Secondly, don't need to maintain another new package in the PPA.

Technically seen.. Is the init in cura_app really needed at this point? Maybe we could move all the code to the crash handler first and then add an commandline option. :thinking:

fieldOfView:
> But why not use the sentry SDK?

Do you really want all my crashes when I am developing my plugins?

thopiekar:
Also in my case, I also doubt you want to see all the crashes and logs from PPA users, since you as Ultimaker want to focus on your distributed software.

nallath:
I don't mind the crash data from the plugins, sentry is very easy to filter (hell, we already got some crash reports from the PPA that we could fix :)). 

We filter out any identifying data, so neither us or the people from sentry cant track any privacy-sensitive data.

fieldOfView:
> We filter out any identifying data

They don't get an IP address?

Anyway, this Sentry integration is not critical to running Cura, so I think it should be optional when running from source. Even loading libCharon and libSavitar is surrounded by similar try/except statements.

nallath:
> > We filter out any identifying data
> 
> They don't get an IP address?

Nope! The old system didn't store those either and I want to keep it like that. I'm a big proponent of strong privacy protection. 

> Anyway, this Sentry integration is not critical to running Cura, so I think it should be optional when running from source. Even loading libCharon and libSavitar is surrounded by similar try/except statements.

Yeah, fair enough.

fieldOfView:
Off-topic: I fail to see how they don't get an IP address of the sender of the report. But if you say so...

nallath:
Well, yeah they get one. But they aren't legally allowed to store it (in exactly the same as as that *every* website has your IP and can only use it to send a response back).

thopiekar:
Thank you for merging the code!

@nallath Is there any intentions to share the error lines of plugins with the maintainers? I guess it won't happen soon, right?

nallath:
Right now we still do this manually. It is something that we want to do at some point in time.

PR #6924, opened 2020-01-05
CURA-6577/Implemented the logic for syncing packages from Cura MP to Cloud MP.

DonExo:
Implemented the logic for syncing packages from Cura MP to Cloud MP when a user installs a package.

ToDo: Check the materials sync


DonExo:
Superseded by #6937

PR #6845, opened 2019-12-23
CURA-7038/Show list of packages to be synced

DonExo:
This PR introduces a new dialog called Compatibility Dialog that will show packages that the user has already "Subscribed" to from the Cloud Marketplace, but haven't yet installed them in Cura Marketplace.

https://github.com/Ultimaker/Uranium/pull/555

PR #6932, opened 2020-01-07
CURA-7038/Remove the label(s) from the compatibility dialog if there are no pac…

DonExo:
_Addition to #6845 , but since it got merged and its branch deleted - thought it'll be wiser to create a new branch for this addition._

Added checks in place, if there are no (in)compatible packages to be shown in the compatibility Dialog - the label won't be shown at all, as suggested by @evtrados in CURA-7038.

### DIFF
--- a/plugins/Toolbox/src/SubscribedPackagesModel.py
+++ b/plugins/Toolbox/src/SubscribedPackagesModel.py
@@ -35,9 +35,9 @@ class SubscribedPackagesModel(ListModel):
                 continue
             package = {"name": item["display_name"], "sdk_versions": item["sdk_versions"]}
             if self._sdk_version not in item["sdk_versions"]:
-                package.update({"is_compatible": "False"})
+                package.update({"is_compatible": False})
             else:
-                package.update({"is_compatible": "True"})
+                package.update({"is_compatible": True})
             try:
                 package.update({"icon_url": item["icon_url"]})
             except KeyError:  # There is no 'icon_url" in the response payload for this package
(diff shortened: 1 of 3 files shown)


plugins/Toolbox/src/SubscribedPackagesModel.py
```diff
@@ -27,20 +28,34 @@ def addValue(self, discrepancy):
...
             except KeyError:  # There is no 'icon_url" in the response payload for this package
                 package.update({"icon_url": ""})
 
-            items.append(package)
-        self.setItems(items)
\ No newline at end of file
+            self._items.append(package)
+        self.setItems(self._items)
+
+    def has_compatible_packages(self):
```
nallath: It's not a property, so it should be hasCompatiblePackages (functions are lower camel case)

plugins/Toolbox/src/SubscribedPackagesModel.py
```diff
@@ -27,20 +28,34 @@ def addValue(self, discrepancy):
...
+        has_compatible_items  = False
+        for item in self._items:
+            if item['is_compatible'] == True:
+                has_compatible_items = True
+        return has_compatible_items
+
+    def has_incompatible_packages(self):
```
nallath: Same as above (hasIncompatiblePackages)

plugins/Toolbox/src/SubscribedPackagesModel.py
```diff
@@ -27,20 +28,34 @@ def addValue(self, discrepancy):
...
             except KeyError:  # There is no 'icon_url" in the response payload for this package
                 package.update({"icon_url": ""})
 
-            items.append(package)
-        self.setItems(items)
\ No newline at end of file
+            self._items.append(package)
+        self.setItems(self._items)
+
+    def hasCompatiblePackages(self):
```
ChrisTerBeke: Missing some typing here ;)

PR #6942, opened 2020-01-09
Add GEEETECH printer profile for A10 & A20 Series

amitlobo:
I would like to add support for the GEEETECH A10 & A20 printer in Cura. Hopefully this can make it coming release. Thanks!


Ghostkeeper:
You might consider letting these profiles inherit from each other, if they are mostly the same. Would make it a bit easier to maintain. Not required, but just a consideration if you ever want to update it.

The code looks good for the rest. I'm giving it a good test then.

resources/definitions/GEEETECH_A10.def.json
```diff
@@ -0,0 +1,57 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "GEEETECH A10" },
+        "machine_width": {
+            "default_value": 220
+        },
+        "machine_height": {
+            "default_value": 220
+        },
+        "machine_depth": {
+            "default_value": 220
```
Ghostkeeper: In the [specifications](https://www.geeetech.com/geeetech-a10-open-source-big-building-volume-3d-printer-p-1033.html) they say that it's 220x220x2*6*0mm build volume. Should this be changed or was that intentional?

resources/definitions/GEEETECH_A10M.def.json
```diff
@@ -0,0 +1,58 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "GEEETECH A10M" },
+        "machine_width": {
+            "default_value": 220
+        },
+        "machine_height": {
+            "default_value": 220
+        },
+        "machine_depth": {
+            "default_value": 220
```
Ghostkeeper: [This publication](http://www.geeetech.com/wiki/index.php/Geeetech_A10M_3D_printer) claims that the size is 220x220x260. Did you make it smaller intentionally or was that a mistake?

resources/definitions/GEEETECH_A10T.def.json
```diff
@@ -0,0 +1,59 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "GEEETECH A10T" },
+        "machine_width": {
+            "default_value": 220
+        },
+        "machine_height": {
+            "default_value": 220
+        },
+        "machine_depth": {
+            "default_value": 220
```
Ghostkeeper: In the [specifications on their website](https://www.geeetech.com/geeetech-a10t-3-in-1-out-mixcolor-auto-leveling-3d-printer-p-1107.html) they claim a build volume of 220x220x250. Did you make it smaller intentionally or was that a typo?

resources/definitions/GEEETECH_A20T.def.json
```diff
@@ -0,0 +1,59 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "GEEETECH A20T" },
+        "machine_width": {
+            "default_value": 255
+        },
+        "machine_height": {
+            "default_value": 255
+        },
+        "machine_depth": {
+            "default_value": 255
```
Ghostkeeper: In the [specifications of this printer](https://www.geeetech.com/geeetech-a20t-3in1out-mix-color-3d-printer-p-1108.html) it claims that the size is only 250x250x250 for the T edition. Did you make this bigger intentionally? It could damage the printer if it goes out of bounds.

nallath:
These are still not quite correct. You renamed the files, but you should also rename the ID's inside of the files (eg: resources/extruders/geeetech_A10M_1.def.json still refers to "GEEETECH_A10M", which doesn't exist anymore)

Ghostkeeper:
Yay, green check mark now. Let's see if it actually runs then.

Ghostkeeper:
Thanks for the profiles! We squashed your commits, and then Github doesn't properly detect that they were merged, but they are still here: https://github.com/Ultimaker/Cura/commit/004ddb3e2c28be30c24158837fe1a14971db29b2

PR #6944, opened 2020-01-09
Added Only Wall Gaps and Wall And Skin Gaps options to Fill Gaps Between Walls Setting.

smartavionics:
Added a couple of further options to the wall gap filling setting:

Only Wall Gaps - fills gaps between walls but doesn't fill gaps in the skin or infill.

Wall And Skin Gaps - fills gaps between walls and also gaps where skin would be but doesn't fill gaps in infill.


smartavionics:
Not working correctly, maybe replace with a new PR sometime.

PR #6950, opened 2020-01-10
Cura 6983 sync packages

ninovanhooff:
Adds the CloudSync flow. The SyncOrchestrator is addded as an independent extension within the toolbox package plugin. 

Uses the HttpRequestManager for all Api requests

Refactors some code out of Toolbox.py into the CloudSync folder. Toolbox should show completely unchanged behavior.

Note this is the happy flow only, see Jira ticket for scope

See Uranium for accompanying PR


plugins/Toolbox/src/CloudApiModel.py
```diff
@@ -0,0 +1,20 @@
+from typing import Union
```
DonExo: Why is this file not in the CloudSync folder?
Is it somehow different than the others? Serves a different purpose?

DonExo:
I do fear that this PR has gotten way off its initial request in the ticket with the refactoring.
It might have been better if we split it and use a separate ticket for refactoring, and another for the actual implementation of the ticket requirements.

There are 715 lines to review them and 10 or more NEW models/classes that honestly, I find confusing and it's hard to follow which model is used for what purpose.

0. CloudApiModel
1. CloudPackageManager
2. CloudPackageChecker
3. DiscrepanciesPresenter
4. DownloadPresenter
5. LicensePresenter
6. RestartApplicationPresenter ??
7. LicenseModel
8. UltimakerCloudScope
9. SyncOrchestrator
10. HttpRequestScope
11. DefaultUserAgentScope

Even though the model names are self-descriptive, I feel like this is somewhat of over-engineering things. Like, do we actually need a RestartApplicationPresenter ?

plugins/Toolbox/src/CloudSync/SyncOrchestrator.py
```diff
@@ -0,0 +1,99 @@
...
+    def _onPackageMutations(self, mutations: SubscribedPackagesModel):
+        self._download_Presenter = self._download_Presenter.resetCopy()
+        self._download_Presenter.done.connect(self._onDownloadFinished)
+        self._download_Presenter.download(mutations)
+
+    ## Called when a set of packages have finished downloading
+    # \param success_items: Dict[package_id, file_path]
+    # \param error_items: List[package_id]
+    def _onDownloadFinished(self, success_items: Dict[str, str], error_items: List[str]):
+        # todo handle error items
```
rburema: If it is _not_ part of this ticket, then we're missing a ticket. We should either handle it in this one or create a new one.

As far as I can see now, all errors are written to the logs and/or appended to a list of erroneous packages. I'd like to see if things go wrong in the front-end. We should also have a think about what we should do with error'd packages in general... retry to download each time? Does it depend on what part of the download or installation went wrong?

plugins/Toolbox/src/Toolbox.py
```diff
@@ -61,17 +59,15 @@ def __init__(self, application: CuraApplication) -> None:
         self._server_response_data = {
             "authors":              [],
             "packages":             [],
-            "updates":              [],
-            "subscribed_packages":  [],
+            "updates":              []
```
rburema: I think it's better to leave 'end of list' comma's in place, so git-blame works (more smoothly). It's not a big deal though, so we can leave it.

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _onFinished(self, package_id: str, reply: QNetworkReply):
+        self._progress[package_id]["received"] = self._progress[package_id]["total"]
+
+        try:
+            with tempfile.NamedTemporaryFile(mode ="wb+", suffix =".curapackage", delete = False) as temp_file:
+                bytes_read = reply.read(256 * 1024)
+                while bytes_read:
+                    temp_file.write(bytes_read)
+                    bytes_read = reply.read(256 * 1024)
```
rburema: This hardcoded value occurs twice. Please consider using a const-var.

plugins/Toolbox/src/CloudSync/SyncOrchestrator.py
```diff
@@ -0,0 +1,99 @@
...
+        self._package_manager = app.getPackageManager()
+        self._cloud_package_manager = CloudPackageManager(app)
+
+        self._checker = CloudPackageChecker(app)  # type: CloudPackageChecker
+        self._checker.discrepancies.connect(self._onDiscrepancies)
+
+        self._discrepancies_presenter = DiscrepanciesPresenter(app)  # type: DiscrepanciesPresenter
+        self._discrepancies_presenter.packageMutations.connect(self._onPackageMutations)
+
+        self._download_Presenter = DownloadPresenter(app)  # type: DownloadPresenter
```
rburema: Capital 'P' in the middle of a non-camel-cased variable.

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
```diff
@@ -0,0 +1,127 @@
...
+            lifetime = 0,
+            use_inactivity_timer=False,
+            progress = 0.0,
+            title = i18n_catalog.i18nc("@info:title", "Changes detected from your Ultimaker account", ))
+
+    def _onFinished(self, package_id: str, reply: QNetworkReply):
+        self._progress[package_id]["received"] = self._progress[package_id]["total"]
+
+        try:
+            with tempfile.NamedTemporaryFile(mode ="wb+", suffix =".curapackage", delete = False) as temp_file:
```
rburema: If `... deleted = False) ...`, where _does_ it get deleted?

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
```diff
@@ -0,0 +1,127 @@
...
+                scope = self._scope)
+
+            self._progress[package_id] = {
+                "received": 0,
+                "total": 1,  # make sure this is not considered done yet. Also divByZero-safe
+                "file_written": None,
+                "request_data": request_data
+            }
+
+        self._started = True
```
rburema: I'm not sure if this method will ever be called in such a way that it matters, but as written it's not thread-safe. (Or at least the defensive programming 'only run this once' isn't.)

plugins/Toolbox/src/Toolbox.py
```diff
@@ -735,8 +674,10 @@ def _onDownloadComplete(self, file_path: str) -> None:
             self.openLicenseDialog(package_info["package_id"], license_content, file_path)
             return
 
-        self.install(file_path)
-        self.subscribe(package_info["package_id"])
+        package_id = self.install(file_path)
+        if package_id != package_info["package_id"]:
+            Logger.error("Installed package {} does not match {}".format(package_id, package_info["package_id"]))
```
rburema: ... and then what? :-)

plugins/Toolbox/src/CloudSync/SyncOrchestrator.py
```diff
@@ -0,0 +1,99 @@
...
+    def _onLicenseAnswers(self, answers: List[Dict[str, Any]]):
+        Logger.debug("Got license answers: {}", answers)
+
+        has_changes = False  # True when at least one package is installed
+
+        for item in answers:
+            if item["accepted"]:
+                # install and subscribe packages
+                if not self._package_manager.installPackage(item["package_path"]):
+                    Logger.error("could not install {}".format(item["package_id"]))
```
rburema: ... and then what? :-)

plugins/Toolbox/src/CloudApiModel.py
ninovanhooff: Because it is used by Toolbox.py too, ie other scope than only Sync.

plugins/Toolbox/src/CloudSync/SyncOrchestrator.py
ninovanhooff: Added to the CCB

plugins/Toolbox/src/Toolbox.py
ninovanhooff: Agree, will do that in the future

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
ninovanhooff: SyncOrchestrator._onLicencseAnswers

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
ninovanhooff: because there is no lock on executing this function? It was not designed to be thread-safe. You suggest something like this on the download function, I guess?

https://stackoverflow.com/a/39147428

plugins/Toolbox/src/Toolbox.py
ninovanhooff: I suggest ignoring this as a fluke, but i'll add it to the CCB. I think this check was added by me, so an error in the log is an improvement over silent failure

plugins/Toolbox/src/CloudSync/SyncOrchestrator.py
ninovanhooff: Added to CCB

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
rburema: > It was not designed to be thread-safe.

I was thinking more about future adaptations, but it's not a big issue if it's not threaded right now.

(I wish there was a good way to indicate the 'seriousness' of a review comment, from 'needs fixing immediately' to 'just making sure you're aware of this' the latter of which are the majority of my comments in this.)

rburema: Also yes, the lock is a way to solve it.

plugins/Toolbox/src/Toolbox.py
rburema: > an error in the log is an improvement over silent failure

Agreed!

plugins/Toolbox/src/CloudSync/DownloadPresenter.py
ninovanhooff: I think I'll leave it the way it is. I expect a singular SyncOrchestrator with a singular DownloadPresenter

PR #6974, opened 2020-01-14
Fix typing in __init__ methods to appease MYPY

fieldOfView:
This PR adds a number of typing fixes in __init__ statements, because MYPY is complaining about them.


Ghostkeeper:
Thanks for the contribution again :)
Wasn't complaining about it for me though.

PR #6983, opened 2020-01-15
Add MAKEiT Pro-MX Profile

MakeIt3D:


nallath:
Is this a different machine than the one added in https://github.com/Ultimaker/Cura/pull/1549?

MakeIt3D:
@nallath yes, this version has an extended Z-axis height

Vandrasc:
@MakeIt3D, the extruders, and material definition are not there, also the Fine and Draft both have 0.2 mm LH, is this intended? 
double
mat

PR #6986, opened 2020-01-16
Cura 7027 add more materials from marketplace button

konskarm:
Button was fixed to open the in-Cura marketplace.

### DIFF
--- a/resources/qml/Actions.qml
+++ b/resources/qml/Actions.qml
@@ -192,7 +192,6 @@ Item
     Action
     {
         id: marketplaceMaterialsAction
-        onTriggered: Qt.openUrlExternally("https://marketplace.ultimaker.com/app/cura/materials")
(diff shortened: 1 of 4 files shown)


resources/qml/Actions.qml
```diff
@@ -188,6 +189,13 @@ Item
         shortcut: "Ctrl+K"
     }
 
+    Action
+    {
+        id: marketplaceMaterialsAction
+        iconName: "configure"
```
ninovanhooff: Why this iconName? Does it doe anything? Don't think it is mandatory, when looking at other usages

ninovanhooff:
```suggestion
```

Ghostkeeper: These icons are visible on some operating systems. It displays a little icon then at the context menu option.

PR #6987, opened 2020-01-16
Strateo3D

KOUBeMT:


Ghostkeeper:
Merge together with https://github.com/Ultimaker/fdm_materials/pull/122.

nallath:
Merging this would require a version upgrade script, since the names of the files are changed.

konskarm:
Aside from the changes in Strateo3D printers, this PR reports that it changes many unrelated files, such as the ContainerRegistry or resources/intent/ultimaker_s3/um_s3_aa0.4_CPEP_Fast_Print_Accurate.inst.cfg, which is quite strange.

Ghostkeeper:
Yeah it looks like all of the newlines in those files were changed. That makes our history very confusing. If we're merging this, we'd need to have that changed back and then squash and merge to prevent it ending up in our history.

KOUBeMT:
Did i do anything the wrong way? If i can fix it please let me know... thanks.

Ghostkeeper:
Yeah, look at the diff here on Github. It's changing the whitespace in 200 files. Most of those are not related to the change you intended to make at all. Please undo that!
If we'd ever look for why a change was made on a certain line, we'll only find that it was changed by KOUBeMT to update the Strateo3D profiles, which defeats the purpose of `git blame`.

PR #6988, opened 2020-01-16
CURA-7071/Added option to redirect to Web MP from Cura MP

DonExo:
Added a small icon in the right corner of the header of the Toolbox.
Clicking on it redirects directly to Web Marketplace.


cura/ApplicationMetadata.py
```diff
@@ -56,3 +56,5 @@
 
 except ImportError:
     CuraAppDisplayName = DEFAULT_CURA_DISPLAY_NAME
+
+WEB_MARKETPLACE_URL = "https://marketplace.ultimaker.com/app/cura/materials"
```
ChrisTerBeke: This should be dynamic depending on environment (prod vs staging).

DonExo: 1. What is the dynamic url? "marketplace-staging...." ?
2. Why would a normal user be interested in the staging environment?

ChrisTerBeke: 1. yup! 
2. Not for end users, but when making internal staging builds for testing it's good to have all the URLs correctly in place to prevent confusion.

plugins/Toolbox/resources/qml/components/ToolboxHeader.qml
```diff
@@ -51,39 +51,61 @@ Item
...
+            rightMargin: UM.Theme.getSize("default_margin").width
+            verticalCenter: parent.verticalCenter
+        }
+        Image
+        {
+            id: cloudMarketplaceButton
+            source: "../../images/marketplace.png"
+            height: 45
+            width: height
+            //width: UM.Theme.getSize("toolbox_header_tab").width
```
ninovanhooff: Remove leftover code

cura/ApplicationMetadata.py
Ghostkeeper: I've made the URL a build parameter. See also this PR: https://github.com/Ultimaker/cura-build/pull/231

plugins/Toolbox/resources/qml/components/ToolboxHeader.qml
Ghostkeeper: Done!

Ghostkeeper:
Please also review the following PR, which is a dependency of this one now: https://github.com/Ultimaker/cura-build/pull/231

cura/ApplicationMetadata.py
ninovanhooff: Unused, remove

plugins/Toolbox/src/Toolbox.py
```diff
@@ -17,6 +16,7 @@
 from UM.Version import Version
 
 from cura import ApplicationMetadata
+from cura.CuraVersion import CuraMarketplaceRoot
```
ninovanhooff: This import should be surrounded with try.. catch. See other implementations

PR #6989, opened 2020-01-16
HMS434 update

maukcc:
HMS434 update
exclude materials
speed settings
cooling settings
extruder gcode settings


maukcc:
Is there something from our side that is needed to make the changes ??

Ghostkeeper:
No, I think we've been confusing the changes you've made and removed the internal ticket thinking that the changes were already applied. We'll review it again.

PR #6994, opened 2020-01-17
Prepare PostProcessingPlugin for security.

rburema:


nallath:
We should only check the scripts in the user folder for signatures (and then of course only if you are in the Enterprise mode). RIght now the check is a bit easy to defeat by simply not having a signature at all :-P

Ghostkeeper:
This PR has no Jira ticket attached. Did this get forgotten? Was it made outside of a Jira ticket?

nallath:
I think there never was a ticket for this. Anwyay, this is something that should be in 4.6, so i've created it (CURA-7319)

rburema:
@Ghostkeeper No I did this during one of the testing and tooling sessions after @nallath talked about how not all cases where covered, even if all the tickets we had at the time where implemented. 

I remember not merging at the time was a conscious decision, but since we didn't close it I assume it's just unfinished in some way (whether that's the PR itself or some other thing that we needed in order to progress I don't know, I'd have to dive in again).

rburema:
I'm not sure why I didn't finish it, all I needed to do was add check to not check the bundled ones.

plugins/PostProcessingPlugin/tests/TestPostProcessingPlugin.py
```diff
@@ -0,0 +1,55 @@
...
+    mocked_trust.signedFileCheck = MagicMock(return_value=True)
+
+    realSignatureFileExistsFor = Trust.signatureFileExistsFor
+    Trust.signatureFileExistsFor = MagicMock(return_value=True)
+
+    with patch("UM.Trust.Trust.getInstanceOrNone", return_value=mocked_trust):
+        assert PostProcessingPlugin._isScriptAllowed("blaat.py")
+
+    # cleanup
+    Trust.signatureFileExistsFor = realSignatureFileExistsFor
```
nallath: It's a bit cleaner to do this with a patch.object. I'll fix that for you

PR #7003, opened 2020-01-20
ci_invalid_import_checker

ninovanhooff:
Added a grep for import plugins.*. Wondering what the best place is to put this code

### DIFF
--- a/scripts/check_invalid_imports.py
+++ b/scripts/check_invalid_imports.py
@@ -1,0 +1,60 @@
+import os
+import re
+import sys
+from pathlib import Path
+
+"""
+Run this file with the Cura project root as the working directory
+"""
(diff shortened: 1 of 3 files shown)


scripts/check_invalid_imports.py
```diff
@@ -0,0 +1,65 @@
...
+        :return: True if checks passed, False when the test fails
+        """
+        passed = True
+        for path_like in Path(root_dir).rglob('*.py'):
+            if not self.check_file(str(path_like)):
+                passed = False
+
+        return passed
+
+    def check_file(self, file_path):
```
konskarm:
```suggestion
    def check_file(self, file_path: str) -> bool:
```

Issue #2158, opened 2017-07-24
Hole and External wall offset


AlucardF0X:
So, since the dawn of 3D Printing, the feature "Model Horizontal Offset" a.k.a. "Size compensation" has been a very valuable feature. 

Except I've ran into a few 3D Models that don't wanna slice correctly.
After printing, the external dimensions of the model are consistently 0.5mm smaller than designated.
The hole dimensions are spot-on. 

If I use any of the settings, it'll make the hole too big, or too small. Oof!

There really needs to be a setting that allows configuration of internal holes and external walls, rather than moving both. So I can independently resize holes (internal walls) and external walls.

A work-around is to size-compensate with 0.5mm, thus make the holes 0.5mm smaller... Then drill out the holes, but that's a pain. 

I don't think there's a slicer that currently supports this at all. So I'm a sitting duck with a few projects I'm trying to complete, just because cura decided to make external walls too small, somehow.

It isn't a "plastic shrinks" problem because I'm pretty sure across 2mm, 0.5mm is NOT plastic shrinkage, that'd be a whopping 25% shrinkage, and that's awesomely impossible. 

![bad diameters](https://user-images.githubusercontent.com/10225169/28544885-ccbf5620-70bc-11e7-9046-9d9970d39dfd.JPG)

PS: Those little pegs are also, 0.5mm smaller than they should be. 
If I try to thicken them by horizontal offset, I'd need to apply 0.25mm horizontal offset. 
That would *thicken* the model by volume, making holes smaller by 0.25mm, and pegs thicker by 0.25mm, but that'd make all my holes that much smaller... Once again not a good fit.

So I'd need to compensate horizontally by some half-value to make the two parts fit? But be off-spec? 

If my peg is 2.5mm and the hole is 3mm, I'd transfer 0.25mm from bigger to smaller, so that'd result in 
2.75mm peg and 2.75mm hole, a perfect fit, but then they'd be off-spec as they're meant to be 3mm...

The wall-inset and horizontal-offset settings only move units from one to the other. 
There isn't a setting that would let me increase the peg, without decreasing the hole. 
There isn't a setting that would let me increase the hole, without decreasing the peg. 

I essentially need the hole to STAY, and the peg to THICKEN. 
Or, if I balance them first with horizontal offset:
I'd need to THICKEN (make it bigger) the hole (wall-wise), and THICKEN the peg... Which there isn't currently a setting for. 

[...]

PR #7010, opened 2020-01-21
Added Hole Horizontal Expansion setting.

smartavionics:
Provide simple solution for https://github.com/Ultimaker/Cura/issues/2158.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1389,7 +1389,19 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
-                "z_seam_type":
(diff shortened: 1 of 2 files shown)


rburema:
I'm going to merge this, but for completions sake I will note a few things:
 - While the possibility that it'll ever come up in a real world example _and_ the behaviour isn't wanted is low: If you put Horizontal Expansion high enough that it merges spikes (creating holes) those holes will also be expanded.
 - From a code-clarity perspective, it would'be been a bit more clear to put this where the other ('normal') horizontal expansion takes place.

smartavionics:
Feel free to move the setting elsewhere if you wish.

rburema:
> Feel free to move the setting elsewhere if you wish.

I know, just a bit strapped for time right now :-)

PR #7013, opened 2020-01-22
CURA-7129_license_window_design

ninovanhooff:
Depends on a Uranium PR with the same name


plugins/Toolbox/resources/qml/dialogs/ToolboxLicenseDialog.qml
```diff
@@ -20,31 +21,63 @@ UM.Dialog
...
+                height: width
+                fillMode: Image.PreserveAspectFit
+                source: licenseModel.iconUrl || "../../images/logobot.svg"
+                mipmap: true
+            }
+
+            Label
+            {
+                id: packageName
+                text: licenseModel.packageName
```
konskarm: This should be changed into the display name to be consistent with the previous page.

PR #7015, opened 2020-01-22
Feature: Wall collapse distance

BagelOrb:
Front-end PR for https://github.com/Ultimaker/CuraEngine/pull/1187

Let all discussion etc be there - not here.


Ghostkeeper:
See https://github.com/Ultimaker/CuraEngine/pull/1187#issuecomment-674143432 for the reason of closure.

PR #7034, opened 2020-01-28
Renkforce printers: Use defaults for retraction hop

dok-net:
Retraction hop enable has default value, the height value is fine by default, too.
This cleans up confusion after merging 7e1162093c7d6814db114877bc68677d49bd1cdb (which is a dead branch?)

### DIFF
--- a/resources/definitions/renkforce_rf100.def.json
+++ b/resources/definitions/renkforce_rf100.def.json
@@ -158,9 +158,6 @@
         "retraction_enable": {
             "value": "True"
         },
-        "retraction_hop": {
-            "value": "1.0"
-        },
         "retraction_min_travel": {
             "value": "1.5"
         },
--- a/resources/definitions/renkforce_rf100_v2.def.json
+++ b/resources/definitions/renkforce_rf100_v2.def.json
@@ -158,9 +158,6 @@
         "retraction_enable": {
             "value": "True"
         },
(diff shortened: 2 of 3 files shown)


Ghostkeeper:
https://github.com/Ultimaker/Cura/commit/7e1162093c7d6814db114877bc68677d49bd1cdb is not intended to be merged any more. It was replaced by this PR in the end: https://github.com/Ultimaker/CuraEngine/pull/1185 The branch on Github is deleted. I still have a local copy though. (I have 2006 local branches in my Cura repo! Wow!)

You chose the original Z hop value assuming that would get merged then? Or are you removing it in case it gets merged?

dok-net:
I choose to inherit from fdmprinter.def.json. That should be fine and avoid the "slow and causes wear" caveat, right?
I'm sorry if I do not understand the question exactly:

> You chose the original Z hop value assuming that would get merged then? Or are you removing it in case it gets merged?

If my answer makes sense to you, though, could you please merge this before 2.0.3 (which must be imminent)?

Ghostkeeper:
Yeah, inheriting from fdmprinter.def.json should be fine for many printers, sure. I was just wondering what your "confusion after merging 7e1162093c7d6814db114877bc68677d49bd1cdb" was about, and hoping I could clarify it.

PR #7037, opened 2020-01-28
Add valueFromContainer() functions for profiles

fieldOfView:
This PR adds CuraFormulaFunctions to get a value for a key from a stack starting at a specified container index. This enables a writer of profiles to eg refer to inherited values for a key from a formula for the same setting, eg:

A quality profile can use this formula to specify a temperature 5 degrees higher than the temperature specified in the material container:
`material_print_temperature = = valueFromContainer("material_print_temperature", 4) + 5`
A setting referencing itself without this function will create an infinite loop.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -393,6 +393,8 @@ class CuraApplication(QtApplication):
         SettingFunction.registerOperator("extruderValues", self._cura_formula_functions.getValuesInAllExtruders)
         SettingFunction.registerOperator("resolveOrValue", self._cura_formula_functions.getResolveOrValue)
         SettingFunction.registerOperator("defaultExtruderPosition", self._cura_formula_functions.getDefaultExtruderPosition)
+        SettingFunction.registerOperator("valueFromContainer", self._cura_formula_functions.getValueFromContainerAtIndex)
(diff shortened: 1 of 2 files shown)


fieldOfView:
Did I break things, or were they already broken?

Ghostkeeper:
You broke things, it seems. It's the type checking that's failing:

    cura/Settings/CuraFormulaFunctions.py:153: error: Cannot determine type of 'extruder_position'

fieldOfView:
Fixed it, thanks.

nallath:
It doesn't quite seem to work for me. 

I've added the `material_print_temperature = = valueFromContainer("material_print_temperature", 4) + 5` to um3_aa0.4_PLA_Draft_print.inst.cfg. Once I enable that profile, I get no value for printing temperature at all 
NoPrintingTemperature

fieldOfView:
I'll look in to it.

fieldOfView:
I made two mistakes in my example:
* there's a space after the second `=` that causes an unexpected indent error making the formula not work
* `material_print_temperature` should be taken from the extruder, not the global stack

@nallath, could you try the following instead?
`material_print_temperature = =extruderValueFromContainer(extruder_nr, "material_print_temperature", 4) + 2.1`
The 2.1 value should make it clear that the formula is working, and changing the material should show the value updating.

fieldOfView:
For testing I prefer using the formula in a `user` profile, so it is still used when I switch materials.

PR #7041, opened 2020-01-29
Cura 7099 sync sideloaded plugins

ninovanhooff:
Has a Companion UM PR

### DIFF
--- a/plugins/Toolbox/src/Toolbox.py
+++ b/plugins/Toolbox/src/Toolbox.py
@@ -21,7 +21,6 @@ from cura.Machines.ContainerTree import ContainerTree
 
 from .CloudApiModel import CloudApiModel
 from .AuthorsModel import AuthorsModel
-from .CloudSync.CloudPackageManager import CloudPackageManager
 from .CloudSync.LicenseModel import LicenseModel
 from .PackagesModel import PackagesModel
 from .UltimakerCloudScope import UltimakerCloudScope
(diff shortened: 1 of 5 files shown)


ninovanhooff:
> This is a singleton, so don't cache it.

@Ghostkeeper What would be the alternative approach? Do you prefer that I cache CuraApplication instead? It doesn't use the Singleton pattern, but can be considered as such.

ninovanhooff:
> Should we turn this into a protected function then?

Thought about this, but probably not. The scenario where the user copies a plugin directly to the corresponding settings folder probably needs to use this

ninovanhooff:
> The original line here explains why this plug-in can't be disabled: If you disable the Toolbox, you've disabled the interface to disable/enable plug-ins, so you couldn't change it back. To prevent this we've made the Toolbox a required plug-in.

Reverted

plugins/Toolbox/src/Toolbox.py
```diff
@@ -377,7 +371,6 @@ def resetMaterialsQualitiesAndUninstall(self) -> None:
     def onLicenseAccepted(self):
         self.closeLicenseDialog.emit()
         package_id = self.install(self.getLicenseDialogPluginFileLocation())
-        self.subscribe(package_id)
```
Ghostkeeper: Hmm, I'd still expect it to subscribe also if you install it via the Marketplace? I'll test if this is still working. Seems that you intended to remove this, so I don't understand. Was this removed just for testing maybe?

PR #7058, opened 2020-02-03
WIP: Delete leftover files from hidden machines

nallath:
CURA-7024

### DIFF
--- a/plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
+++ b/plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
@@ -34,7 +34,6 @@ class VersionUpgrade44to45(VersionUpgrade):
 
         self._extruder_machine_mapping = []  # type: List[ExtruderMachineMapping]
 
-
     def getCfgVersion(self, serialised: str) -> int:
         parser = configparser.ConfigParser(interpolation = None)
         parser.read_string(serialised)
@@ -116,13 +115,13 @@ class VersionUpgrade44to45(VersionUpgrade):
 
     def deleteHiddenMachines(self) -> None:
 
-        # delete extruders associated to hidden machines
+        # Delete extruders associated to hidden machines
         for item in self._extruder_machine_mapping:
             if item.machine_id in self._hidden_machines:
                 Logger.debug("Deleting extruder files for {}".format(item))
                 self._deleteConfigFiles(item.extruder_filename)
 
-        # delete machine files
+        # Delete machine files
         for machine_id, filename in self._hidden_machines.items():
             Logger.debug("Deleting files for hidden machine {}".format(machine_id))
             self._deleteConfigFiles(filename)
--- a/plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
+++ b/plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
@@ -1,7 +1,17 @@
 import configparser
-from typing import Tuple, List
+import os
+from collections import namedtuple
+from pathlib import Path
+from typing import Tuple, List, Dict
 import io
+
+from UM.Logger import Logger
+from UM.Resources import Resources
 from UM.VersionUpgrade import VersionUpgrade
+from UM.VersionUpgradeManager import VersionUpgradeManager, CallableTask
+
+ExtruderMachineMapping = namedtuple("ExtruderMachineMapping", ["extruder_id", "extruder_filename", "machine_id"])
+HiddenMachine = namedtuple("HiddenMachine", ["machine_id", "filename"])
 
 # Settings that were merged into one. Each one is a pair of settings. If both
 # are overwritten, the key wins. If only the key or the value is overwritten,
@@ -15,7 +25,16 @@ _removed_settings = {
     "support_tree_wall_thickness"
 }
 
+
 class VersionUpgrade44to45(VersionUpgrade):
+
+    def __init__(self):
+        self._hidden_machines = {}  # type: Dict[str, str]
+        """machine_id -> stack filename prefix"""
+
+        self._extruder_machine_mapping = []  # type: List[ExtruderMachineMapping]
+
+
     def getCfgVersion(self, serialised: str) -> int:
         parser = configparser.ConfigParser(interpolation = None)
         parser.read_string(serialised)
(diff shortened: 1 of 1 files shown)


plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -15,7 +25,16 @@
     "support_tree_wall_thickness"
 }
 
+
 class VersionUpgrade44to45(VersionUpgrade):
+
+    def __init__(self):
```
nallath:
```suggestion
    def __init__(self) -> None:
```

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -15,7 +25,16 @@
...
+        self._hidden_machines = {}  # type: Dict[str, str]
+        """machine_id -> stack filename prefix"""
+
+        self._extruder_machine_mapping = []  # type: List[ExtruderMachineMapping]
+
+
```
nallath:
```suggestion
```
Only one whiteline between functions in a class.

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +94,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
         # Update version number.
         if "metadata" not in parser:
             parser["metadata"] = {}
+
+        # Check for leftover hidden machines. They should be deleted along with their associated files
+        stack_id = parser["general"]["id"]
+        if parser["metadata"]["setting_version"] == "10" and stack_id is not None:
+            stack_type = parser["metadata"]["type"]
+            if stack_type == "machine" and parser["metadata"]["hidden"]:
+                if not self._hidden_machines:
```
nallath: If there are no hidden machines, the hidden machines need to be deleted? Shouldn't it be the other way around?

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +94,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
...
+                machine = parser["metadata"]["machine"]
+                self._extruder_machine_mapping.append(ExtruderMachineMapping(stack_id, filename, machine))
+
         parser["metadata"]["setting_version"] = "11"
 
         result = io.StringIO()
         parser.write(result)
         return [filename], [result.getvalue()]
+
+    def deleteHiddenMachines(self):
```
nallath:
```suggestion
    def deleteHiddenMachines(self) -> None:
```

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +94,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
...
+
+        # delete extruders associated to hidden machines
```
nallath:
```suggestion
        # Delete extruders associated to hidden machines
```

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +94,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
...
+        for item in self._extruder_machine_mapping:
+            if item.machine_id in self._hidden_machines:
+                Logger.debug("Deleting extruder files for {}".format(item))
+                self._deleteConfigFiles(item.extruder_filename)
+
+        # delete machine files
```
nallath:
```suggestion
        # Delete machine files
```

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +94,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
...
+        for machine_id, filename in self._hidden_machines.items():
+            Logger.debug("Deleting files for hidden machine {}".format(machine_id))
+            self._deleteConfigFiles(filename)
+
+    def _deleteConfigFiles(self, filename: str):
```
nallath:
```suggestion
    def _deleteConfigFiles(self, filename: str) -> None:
```

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +93,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
         # Update version number.
         if "metadata" not in parser:
             parser["metadata"] = {}
+
+        # Check for leftover hidden machines. They should be deleted along with their associated files
+        stack_id = parser["general"]["id"]
+        if parser["metadata"]["setting_version"] == "10" and stack_id is not None:
```
Ghostkeeper: This upgrade step is defined as upgrading from `setting_version` 10 to 11, so checking this is unnecessary.

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -75,8 +93,42 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
...
         parser.write(result)
         return [filename], [result.getvalue()]
+
+    def deleteHiddenMachines(self) -> None:
+
+        # Delete extruders associated to hidden machines
+        for item in self._extruder_machine_mapping:
+            if item.machine_id in self._hidden_machines:
+                Logger.debug("Deleting extruder files for {}".format(item))
+                self._deleteConfigFiles(item.extruder_filename)
```
Ghostkeeper: This is not going to work in the future any more if any of our future upgrade tasks decide to rename the files.

Ghostkeeper:
I have three proposed solutions for the current conundrum:

1. We could use this structure but instead of appending the upgrade task to the end of the upgrade queue insert it in front. I think this'll work, as in, I can't think of a situation why it can't work.
2. In the init the plug-in could gather the files that need to be deleted (before any version upgrading happens). And then in the version upgrade delete them. This has the advantage that no changes to the main program are needed. It'll only work if no other version upgrade renames those files still, but since this only happened in version 4.4 that's not a problem now.
3. We could implement this in the mainline code after all of the version upgrades. This has the disadvantage that it'll chase us forever.

We also have to consider that we might be deleting data on those hidden machines.

PR #7066, opened 2020-02-04
Remove hidden stacks in 4.4 to 4.5 version upgrade

Ghostkeeper:
This implements the discussion we just had on how to do a version upgrade with files referring to different versions.

Replaces https://github.com/Ultimaker/Cura/pull/7058 and https://github.com/Ultimaker/Uranium/pull/575. Those won't be necessary any more if this solution is chosen.


Ghostkeeper:
Please ignore the first commit. This was already merged to master but rebased there but it was still in my history for this branch. Sorry.

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -16,6 +26,79 @@
 }
 
 class VersionUpgrade44to45(VersionUpgrade):
+    def __init__(self) -> None:
+        """
```
ninovanhooff: Thanks for using docstrings :)

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
```diff
@@ -16,6 +26,79 @@
...
+        for root, _, files in os.walk(data_storage):
+            for filename in fnmatch.filter(files, "*.global.cfg"):
+                parser = configparser.ConfigParser(interpolation = None)
+                parser.read(os.path.join(root, filename))
+                if "metadata" in parser and "hidden" in parser["metadata"] and parser["metadata"]["hidden"] == "True":
+                    stack_id = urllib.parse.unquote_plus(os.path.basename(filename).split(".")[0])
+                    hidden_global_stacks.add(stack_id)
+                    # The user container and definition changes container are specific to this stack. We need to delete those too.
+                    if "containers" in parser:
+                        if "0" in parser["containers"]:
```
ninovanhooff: These are the stack entries by "layer", right? eg layer 6 might contain the extruder. Might need a comment here

Ghostkeeper: Yeah they are. I'll clarify which stack entry is which container type.

plugins/VersionUpgrade/VersionUpgrade44to45/VersionUpgrade44to45.py
Ghostkeeper: I'm used to using that for [personal projects](https://github.com/Ghostkeeper/Luna/blob/master/luna/plugins.py). Currently it looks a bit weird with the combination of styles though.

PR #7067, opened 2020-02-04
Add "Always Write Active Tool" machine setting to support Smoothie firmware

grk3010:
Add Smoothie entry to fdmprinter.def.json to allow selecting "Smoothie" Firmware flavor - see https://github.com/Ultimaker/CuraEngine/pull/1197


Ghostkeeper:
See my comment at https://github.com/Ultimaker/CuraEngine/pull/1197

grk3010:
I've implemented the addition of the "machine_always_write_active_tool" parameter and added machine definition files for Smoothieware based printers.  See Ultimaker/CuraEngine#1197.

resources/definitions/smoothie_dual.def.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "machine_heat_zone_length": {
+        	"default_value": 15
+        },
+        "machine_start_gcode": {
+            "default_value": "G28 X0 Y0 ; Home X Y\nM375; Load Bed Leveling Grid\nG1 X300 Y275 F15000 ; Move to bed center\nM280 S3.0 ; Deploy probe pin\n## NOTE: Adjust \"ZX.XX\" after G30 to be the Z-offset (in mm) between the probe and the nozzle\n## This is how you adjust nozzle printing height!\nG30 Z1.25 ; Move down to find bed & set Z home offset\nM280 S7.0 ; Retract probe\nG1 Z15.0 F600 ; move extruder up 15mm\nT0; Activate Extruder 1"
+        },
+        "machine_end_gcode": {
+            "default_value": "G28 X0 Y0; Home X and Y\nM104 S0 ; turn off extruder\nM140 S0 ; turn off bed\nM107; turn off fans\nM84 ; disable motors"
+        }
```
Ghostkeeper, line 52: As far as I understand, this printer is intended to be a "custom" type printer, which could be any printer but with the motion controller swapped out for a Smoothieboard. It's better then to not override any of these settings, but keep them at the fdmprinter defaults.

The Smoothieboard doesn't magically make your printer 600x550x500mm, or give it 2 extruders. I'd expect this printer to only override the g-code flavour, machine_always_write_active_too, and perhaps the start and end g-code. Take a look at how [Custom FFF Printer](https://github.com/Ultimaker/Cura/blob/master/resources/definitions/custom.def.json) looks, for example.

resources/extruders/bondtech_bmgx2m_extruder_0.def.json
```diff
@@ -0,0 +1,21 @@
...
+
+    "overrides": {
+        "extruder_nr": {
+            "default_value": 0,
+            "maximum_value": "1"
+        },
+        "machine_nozzle_offset_x": { "default_value": 0 },
+        "machine_nozzle_offset_y": { "default_value": 0 },
+        "machine_nozzle_size": { "default_value": 0.8 },
+        "material_diameter": { "default_value": 1.75 }
```
Ghostkeeper, line 18: Similarly, I don't think that the Smoothieboard makes your nozzle 0.8mm or have 1.75mm filament.

resources/extruders/bondtech_bmgx2m_extruder_0.def.json
```diff
@@ -0,0 +1,21 @@
+{
```
Ghostkeeper, line 1: Did you intend to make this for the Bondtech extruder kits?

resources/quality/smoothie_dual/smoothie_0.4_fast.inst.cfg
```diff
@@ -0,0 +1,24 @@
+[general]
```
Ghostkeeper, line 1: These quality profiles are really specific to the Mosquito Magnum nozzles, and probably also to your specific printer. I'd prefer you to stick with the defaults. This printer should probably not have any specific quality profiles but just use the global ones.

resources/definitions/smoothie_dual.def.json
grk3010: That make sense. I made all the configuration files match my current printer configuration for testing purposes. I'll make the changes to make it more generic.

resources/extruders/bondtech_bmgx2m_extruder_0.def.json
grk3010: Yes, I made that definition for the Bondtech extruders as that is what I am using, but I can easily configure that based on the custom extruder definitions when setting up the printer profile. 

Deleted these files.

resources/quality/smoothie_dual/smoothie_0.4_fast.inst.cfg
grk3010: removed the quality profiles...

Issue #6531, opened 2019-10-14
Hang with SKR 1.3 + TMC2208 + DRV8825


anjoze:
**Application version**

Cura 4.3
Cura 4.2

**Platform**

Windows 10

**Printer**

Hephestos with SKR 1.3
X and Y: TMC2208
Z and Extruder: DRV8825

**Reproduction steps**
Just print something

**Actual results**

It stops / freeze after 2 or 3 minutes

**Additional information**

Printing from USB
I'm using Repetier Host with no problem, printing for hours.
Tried disable the "anonymous data" as suggested here: https://github.com/Ultimaker/Cura/issues/4016

Does anybody have this configuration working?
Thanks for any help.

Ghostkeeper:
To fix this, we'd need to know what Cura should send differently via USB. We don't have every printer available to test with.

> It stops / freeze after 2 or 3 minutes

Cura stops / freezes, or the printer stops / freezes?

anjoze:
Ok, sorry for the missing info.
The printer freezes and the fans still running.
I'm not sure if the extruder and bed keep ON or OFF. I will test it maybe tomorrow and let you know.
I can click on Abort and the popup message comes, so it's not freezed but I'm unable to close CURA.
When I try to close it it will be stucked until I switch off the printer power supply.

Ghostkeeper:
Could be either that the firmware breaks (which would be an issue for the BQ developers) or that Cura stops sending stuff. We've seen before that if some commands take a long time (such as heating commands) then Cura sends temperature probes to update its temperature as well and these temperature probes queue up too much causing the firmware to break itself. This may be similar, though your description is different with that it does print for 2 minutes.

anjoze:
I've tested and it looks you are right, something is happening with temperature.
As you can see in the picture, the temperature is set to 242/60 but it stops to hold it.
In this test stopped at 14% and start to decrease temperature some seconds before.

On Repetier I'm using Crua 1.3
I have a inductive probe and no more extras like SD card or Screen.

![Freeze](https://user-images.githubusercontent.com/11679993/66833416-b9339400-ef53-11e9-9e9b-6b13e07512a4.png)

GregValiant:
I'm cleaning house.
Is this still a problem with current Cura versions (5.8.0 and up)?  Can I close this?

PR #7078, opened 2020-02-09
Fix freezing on SKR-E3-mini-v1.2 board

bgbnbigben:
When attached to a BigTreeTech SKR-E3-mini-v1.2 board, Cura+USB seems to reliably hang the printer; sometimes you'll hear an alarm sound, other times the printer will just freeze entirely.

The leaveISP() function claims that the serial port is still valid, but as far as I can tell this isn't an actually-valid assumption; it was first introduced in b4df277c but I can't quite track context behind that. This might be true for a bunch of devices but I doubt it's true generally.

My hypothesis here is that leaveISP causes the STM32F103RCT6 to reboot, which then wigs out the power-selector circuit (there's now USB power where there wasn't before), and now all of a sudden the chip is in an entirely different state. But that's mostly a guess.

It doesn't look like the programmer is necessary, however; it simply connects, loads a register for a READ/WRITE command to use, checks the availability of a checksum byte, and then returns. Since we don't care for that, we can just open individual serial connections as/when they're needed with the various different baud rates. This stops my personal SKR-E3-mini-v.2 board from failing, and since this looks like the same fix as https://www.reddit.com/r/BIGTREETECH/comments/dp6k6r/skr_mini_e3_v12_and_curausb/fbv808j/ I'm going to guess that it works for everyone with this board, and shouldn't interfere with other printers since, again, all the STK600 communication is mostly noise.

I suspect this also fixes issue #6531.

### DIFF
--- a/plugins/USBPrinting/AutoDetectBaudJob.py
+++ b/plugins/USBPrinting/AutoDetectBaudJob.py
@@ -5,7 +5,6 @@ from UM.Job import Job
 from UM.Logger import Logger
 
 from .avr_isp import ispBase
-from .avr_isp.stk500v2 import Stk500v2
 
 from time import time, sleep
 from serial import Serial, SerialException
@@ -28,13 +27,7 @@ class AutoDetectBaudJob(Job):
         read_timeout = 3
         tries = 2
 
-        programmer = Stk500v2()
         serial = None
-        try:
-            programmer.connect(self._serial_port)
-            serial = programmer.leaveISP()
-        except ispBase.IspError:
-            programmer.close()
 
         for retry in range(tries):
             for baud_rate in self._all_baud_rates:


Ghostkeeper:
Sorry, I don't reckon we can remove this code. As I replied [here](https://github.com/Ultimaker/Cura/issues/7314#issuecomment-603766968), it's necessary to upload firmware upgrades via USB cable from Cura.

bgbnbigben:
Hmm, just for the sake of my own sanity -- how does the existence of the (closed) programmer allow one to upload firmware? The serial connection is still established, and AFAICT programming mode is left basically immediately (certainly before any firmware upload takes place); could you perhaps point me to the firmware upload code and I can try to wrap my head around that from there?

Cheers~

Ghostkeeper:
The programmer is still active while the baud rate is being detected. If I recall correctly, the presence of the programmer causes Marlin to go into USB printing mode which then allows it to respond to commands for the baud rate detection. But I said that it's necessary for firmware updates because of this commit: https://github.com/Ultimaker/Cura/commit/32cbd27b708fe6311edc538a8af6f15c57294a50 . I trust the commit message more than my memory on this.

Maybe @nallath remembers more on why this fix was applied? It refers to CL-541 which is a refactor ticket for the UM3 network connection.

PR #7086, opened 2020-02-11
Fix the "save to removable drive" not appearing

konskarm:
CURA-7203


rburema:
The right fix, but to the wrong branch, we'd like this in 4.5

PR #7095, opened 2020-02-13
Prevent crashing on PermissionError when importing http.server

Ghostkeeper:
We got a crash report where a PermissionError was raised when http.server is imported. This PermissionError seems to be specific to the user's system. It's crashing somewhere internally in that library when it's trying to read the MIME types supported. Perhaps this user has elevated their computer's security in some way.

Instead of crashing, we don't start the web server. This means that instead of crashing Cura, you get an error in the browser after logging in that it can't reach the localhost page. In Cura you'll remain logged out, as it keeps waiting on the response from the log-in.

Fixes Sentry crash CURA-3Q.

### DIFF
--- a/cura/OAuth2/LocalAuthorizationServer.py
+++ b/cura/OAuth2/LocalAuthorizationServer.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Ultimaker B.V.
+# Copyright (c) 2020 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
 import threading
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This is necessary because the error occurs when importing this:

    from http.server import HTTPServer

Both of these classes import HTTPServer. We could put the try-catch in the files that actually import HTTPServer, but those files can't handle the exception properly (as in, they would need to disable themselves or something but then the parent class also needs to handle disabled AuthorizationRequestServers). So the exception is caught in the place where it can be handled, in this file.

PR #7098, opened 2020-02-14
Fixing default bed temperature for MP Mini Delta

PurpleHullPeas:
This was pointed out to me by new users during Cura 4.5 Beta testing. I would greatly appreciate it if this could be included with the final release of Cura 4.5, but I understand if it needs to wait.

The stock Monoprice Mini Delta 3D printer has trouble reaching 60 C. Setting the bed temperature to 40 C so that it doesn't confuse newcomers to the hobby when the printer sits forever waiting for the bed to reach a temperature that it cannot attain.

### DIFF
--- a/resources/definitions/mp_mini_delta.def.json
+++ b/resources/definitions/mp_mini_delta.def.json
@@ -51,6 +51,7 @@
             "default_value": 0.21,
             "minimum_value": 0.07
         },
+        "material_bed_temperature": { "value": 40 },
         "line_width": { "value": "round(machine_nozzle_size * 0.875, 2)" },
         "material_print_temperature_layer_0": { "value": "material_print_temperature + 5" },
         "material_bed_temperature_layer_0": { "value": "material_bed_temperature + 5" },


Ghostkeeper:
Sorry, we'd prefer this to go through a beta test as well, so we'll merge this to master rather than 4.5. It'll be in the 4.6 release. Thanks for the contribution though!

Ghostkeeper:
Note that the initial build plate temperature is set to the normal build plate temperature +5. This may be why the previous temperature was wrong; it was trying to reach 65 degrees C at the start of the print. The printer is only specced to 60.

PurpleHullPeas:
I understand. The 60 C spec is a lie, as is a lot of manufacturer claims with the printer. The printer struggles to even reach 55 C.

fieldOfView:
> we'd prefer this to go through a beta test as well

This is a bit weird. You publish a beta with new features in order to find issues, and when an issue is found you don't fix it until the next beta which is 2 months away. So the new stable version will ship with a new feature that is known to be broken?

PurpleHullPeas:
@fieldOfView It's my fault for not catching it in the initial commit. I'm so used to the printer profile being packaged with Brian's MPMD 101 profiles (which have bed temperature set there), that I didn't think to check the bed temperature for the default Cura profiles until someone in the FB Group (and later Reddit) encountered the issue.

fieldOfView:
In my opinion, this is exactly why we do betas; to catch bugs before they end up in a stable version.

PR #7101, opened 2020-02-15
Add DXU machine profile

yyh1002:
These are the machine profile files for DXU mod for UM2. Tested on 4.5 beta. 
Thanks.


Ghostkeeper:
The CI testing suite has a few issues with this definition:
* The main definition file is apparently invalid. This means that the JSON failed to parse.
* The Dual version is also invalid.
* There's a number of settings that don't do anything because they override `default_value`, but the setting also has a `value` which overrides `default_value`. AssertionError: Unnecessary default_values for ['gantry_height', 'machine_end_gcode', 'machine_start_gcode', 'speed_layer_0', 'switch_extruder_prime_speed', 'switch_extruder_retraction_amount', 'switch_extruder_retraction_speed'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/dxu.def.json and dxu_dual.def.json
* Your definitions contain an ID field which is not allowed.

yyh1002:
> The CI testing suite has a few issues with this definition:
> 
> * The main definition file is apparently invalid. This means that the JSON failed to parse.
> * The Dual version is also invalid.
> * There's a number of settings that don't do anything because they override `default_value`, but the setting also has a `value` which overrides `default_value`. AssertionError: Unnecessary default_values for ['gantry_height', 'machine_end_gcode', 'machine_start_gcode', 'speed_layer_0', 'switch_extruder_prime_speed', 'switch_extruder_retraction_amount', 'switch_extruder_retraction_speed'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/dxu.def.json and dxu_dual.def.json
> * Your definitions contain an ID field which is not allowed.

Thanks for the very comprehensive hints. I've fixed them, and also simplified the dual version with inherits.

konskarm:
@yyh1002 I would suggest you to add quality profiles because at the moment it is allowed to slice at layer heights with seemingly incompatible nozzles, e.g. slice at 0.6mm layer height with a 0.25mm nozzle, which does not seem possible.
Nevertheless, I will accept the merge request, as the test pass and Cura slices correctly with the DXU. Please consider adding the quality profiles at some time in the future to tighten the layer height options for specific nozzles.

yyh1002:
> @yyh1002 I would suggest you to add quality profiles because at the moment it is allowed to slice at layer heights with seemingly incompatible nozzles, e.g. slice at 0.6mm layer height with a 0.25mm nozzle, which does not seem possible.
> Nevertheless, I will accept the merge request, as the test pass and Cura slices correctly with the DXU. Please consider adding the quality profiles at some time in the future to tighten the layer height options for specific nozzles.

Thanks, I'll look into the benefits of adding quality profiles.

PR #7109, opened 2020-02-16
Acceleration/Jerk override on layer change

kyjak:
When using Jerk/Acceleration settings in Cura, it can have a destroying effect on some printers.
Confirmed case is the Prusa MK3 printer (confirmed by Prusa technical support), which cannot handle these, because they happen too often.
As a workaround, this change request adds possibility for the plugin to set acceleration/jerk at the start of a layer, for example the initial layer, which sometimes needs less acceleration to better bed adhesion.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
+++ b/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
@@ -285,14 +285,14 @@ class ChangeAtZ(Script):
                 "k1_Change_acceleration":
                 {
                     "label": "Change acceleration",
-                    "description": "Select if acceleration has to be changed",
+                    "description": "Select if you want to change max. acceleration. The setting will be overriden when you also use Cura profile acc. settings. Works only with layer number, not height.",
                     "type": "bool",
                     "default_value": false
                 },
                 "k2_acceleration":
                 {
                     "label": "Acceleration",
-                    "description": "Set acceleration string for example 'X1000 Y1000 Z100 E1000'",                    
+                    "description": "Set acceleration string, for example 'X1000 Y1000 Z100 E1000' is a good default, values in mm/s^2. Lower values increase quality/1st layer bonding, but print times get longer.",                    
                     "type": "str",
                     "default_value": "X1000 Y1000 Z100 E1000",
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
As far as I can tell from the implementation, this will only change the acceleration and jerk if Acceleration Control and Jerk Control respectively are disabled in the settings, right?

kyjak:
Dont know how to check for the main settings page if it is enabled/disabled. It should be up to the user knowledge if to use this override at layer or not. If someone uses Curas jerk/acc settings, this setting, if enabled and used, will render useless and will be overruled byt curas jerk/acc realtime settings.

kyjak:
Maybe I should write a better "readme" / hints?

Ghostkeeper:
Yeah there is really no place to put that "readme" currently. But we can at least mention this in the setting descriptions.

kyjak:
I have upgraded the usage help strings, should be fine now.

plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
```diff
@@ -279,6 +281,36 @@ def getSettingDataString(self):
...
                     "enabled": "j1_Change_fanSpeed"
+                },
+                "k1_Change_acceleration":
+                {
+                    "label": "Change acceleration",
+                    "description": "Select if you want to change max. acceleration. The setting will be overriden when you also use Cura profile acc. settings. Works only with layer number, not height.",
+                    "type": "bool",
+                    "default_value": false
+                },
+                "k2_acceleration":
```
ninovanhooff: Looking at other parameters, I expected this to be named k3_ and the next one k4_

Ghostkeeper:
This needs significant changes now that the Change at Z script was largely rewritten for 5.2.

But I also don't think that we should accept this change if it only works when providing a layer number, not when providing a Z height. Most people don't read tooltips, so they will expect it to work fine just like all of the other settings. There is really no good reason why this should only work when pausing at a certain Z coordinate.

kyjak:
Ok, will merge current branch and try to adopt the script, also add layer height not only index. If you want to close this PR in the meantime is fine, can create new one when done.

PR #7112, opened 2020-02-16
Update Rigid3D Printer Profiles

mehmetsutas:
Rigid3D printer profiles are updated. Print settings consolidated, quality and material profiles added.


mehmetsutas:
Build fails but I couldn't find the problem. Cura 4.4.1 runs fine with the new changes but tests fail.

If someone can point out the problem, I can correct it.

konskarm:
Developers, see CURA-7306.

PR #7114, opened 2020-02-16
Ender 5 bed size

mrfomt:
Ender 3 has now 235x235x300 but Ender 5 still has 220x220x300.


Ghostkeeper:
This is a pull request to merge 4.5 into master, which we regularly do anyway. I think you just wanted to make an issue?

We can't verify the specs of the Ender 5 apart from what sources on the internet tell us, so we'd need to have the confirmation of users before we can change them. Preferably these users make those changes themselves in a fork and then supply their change to us via a pull request.

PR #7123, opened 2020-02-17
HMS434 update

maukcc:


Ghostkeeper:
Works fine. It's giving a warning value for the Support Floor Thickness, but I think that's a bit too strict. I'll change it to use `support_z_distance + layer_height` as warning threshold rather than `0.2 + layer_height` since that is more what the warning is supposed to indicate, I think.

PR #7126, opened 2020-02-17
Fix the layout of the head jogging controls with newer Qt versions

fieldOfView:
This PR fixes the Layout of the buttons used to jog the head for output devices that support it (eg USB, OctoPrint) with newer Qt versions than what is currently used in official releases (eg people running from source or unofficial distributions on Linux).

The GridLayout was using incorrect row and column indexes. The Qt versions used in official releases sofar would accept the incorrect indexes, but newer Qt versions apparently truncate the GridLayout, as seen in the screenshot below:
truncated
Please do not git blame to find out who thought the indices within a GridLayout start at 1.

This PR fixes https://github.com/fieldOfView/Cura-OctoPrintPlugin/issues/149

### DIFF
--- a/resources/qml/PrinterOutput/ManualPrinterControl.qml
+++ b/resources/qml/PrinterOutput/ManualPrinterControl.qml
@@ -90,16 +90,16 @@ Item
                     verticalAlignment: Text.AlignVCenter
                     horizontalAlignment: Text.AlignHCenter
 
-                    Layout.row: 1
-                    Layout.column: 2
+                    Layout.row: 0
+                    Layout.column: 1
                     Layout.preferredWidth: width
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Not seeing that issue while running from source using Qt 5.13 though!

Ghostkeeper:
Not seeing that issue while running from source using Qt 5.13 though!

fieldOfView:
The code is wrong and silly though. Whoever wrote it must have had an off-day...

fieldOfView:
Is there a reason _not_ to merge this correction?

Ghostkeeper:
On my home computer I was actually able to reproduce the original bug. I can see that it's fixed then!

PR #7133, opened 2020-02-19
Updated pt_BR strings for Cura 4.5

Patola:
Also changed a couple old strings because the translation was not very good.


Ghostkeeper:
Nice! We'll get to work on these on Friday I think.

PR #7136, opened 2020-02-19
Adding Fabx Pro Machine

hussainsail2002:
Hello,

We are an Indian company called 3Ding and we manufacture our own brand of 3D printers.
our website: www.3ding.in 

We would like to add our machines to the cura ecosystem. 

Since we are doing this for the first time I would like to add our machine FabX Pro first.

I have added a definition file , extruder file, Quality file and a mesh for the bed.

Thanks 
Hussain


hussainsail2002:
Is it possible to find out why pull request  has failed

Ghostkeeper:
Yeah, here are the test failures:

```
_ test_validateMachineDefinitionContainer[/__w/Cura/Cura/tests/Settings/../../resources/definitions/fabxpro.def.json] _

file_path = '/__w/Cura/Cura/tests/Settings/../../resources/definitions/fabxpro.def.json'
definition_container = <UM.Settings.DefinitionContainer.DefinitionContainer object at 0x4e4c718>

    @pytest.mark.parametrize("file_path", machine_filepaths)
    def test_validateMachineDefinitionContainer(file_path, definition_container):
        file_name = os.path.basename(file_path)
        if file_name == "fdmprinter.def.json" or file_name == "fdmextruder.def.json":
            return  # Stop checking, these are root files.
    
        from UM.VersionUpgradeManager import FilesDataUpdateResult
    
        mocked_vum = MagicMock()
        mocked_vum.updateFilesData = lambda ct, v, fdl, fnl: FilesDataUpdateResult(ct, v, fdl, fnl)
        with patch("UM.VersionUpgradeManager.VersionUpgradeManager.getInstance", MagicMock(return_value = mocked_vum)):
>           assertIsDefinitionValid(definition_container, file_path)

../tests/Settings/TestDefinitionContainer.py:60:
```

It's saying that the definition is "invalid" meaning that either the JSON didn't deserialise or it depends on JSON files that couldn't be found. I'm not really sure about this one, but it seems to be related to this auxilary output:

```
[MainThread] UM.Settings.DefinitionContainer.deserializeMetadata [371]: Could not load parent definition container fdmprinter of child whatever
_ test_extruderMatch[/__w/Cura/Cura/tests/Settings/../../resources/extruders/fabxpro_extruder_0.def.json] _

file_path = '/__w/Cura/Cura/tests/Settings/../../resources/extruders/fabxpro_extruder_0.def.json'

    @pytest.mark.parametrize("file_path", extruder_filepaths)
    def test_extruderMatch(file_path: str):
        extruder_id = os.path.basename(file_path).split(".")[0]
        with open(file_path, encoding = "utf-8") as f:
            doc = json.load(f)
    
        if "metadata" not in doc:
            return  # May not be desirable either, but it's probably unfinished then.
        if "machine" not in doc["metadata"] or "position" not in doc["metadata"]:
            return  # FDMextruder doesn't have this since it's not linked to a particular printer.
        machine = doc["metadata"]["machine"]
        position = doc["metadata"]["position"]
    
        # Find the machine definition.
        for machine_filepath in machine_filepaths:
            machine_id = os.path.basename(machine_filepath).split(".")[0]
            if machine_id == machine:
                break
        else:
>           assert False, "The machine ID {machine} is not found.".format(machine = machine)
E           AssertionError: The machine ID FabX Pro is not found.
E           assert False

../tests/Settings/TestDefinitionContainer.py:185: AssertionError
```

```
definition_container = <UM.Settings.DefinitionContainer.DefinitionContainer object at 0x4e4c718>
file_path = '/__w/Cura/Cura/tests/Settings/../../resources/definitions/fabxpro.def.json'

    def assertIsDefinitionValid(definition_container, file_path):
        with open(file_path, encoding = "utf-8") as data:
            json = data.read()
            parser, is_valid = definition_container.readAndValidateSerialized(json)
            assert is_valid #The definition has invalid JSON structure.
            metadata = DefinitionContainer.deserializeMetadata(json, "whatever")
    
            # If the definition defines a platform file, it should be in /resources/meshes/
            if "platform" in metadata[0]:
>               assert metadata[0]["platform"] in all_meshes
E               AssertionError: assert 'fabxpro_platform.stl' in ['anycubic_i3_mega_platform.stl', 'printrbot_simple_metal_upgrade.stl', 'z-bolt_classic_platform.stl', 'stereotech_ste320_platform.obj', 'hms_platform.obj', 'bq_witbox_platform.stl', ...]

../tests/Settings/TestDefinitionContainer.py:72: AssertionError
```

Here it's saying that the platform mesh `fabxpro_platform.stl` couldn't be found. Keep in mind that this is case sensitive.

hussainsail2002:
@Ghostkeeper Thank you for reviewing the errors !.

I changed the file name from fabxpro_platform.stl to fabxpro_platform.STL and I think it solved the issue regarding the platform. There seesm to be an issue with the extruder file maybe ?

I am not sure as everything works fine locally and I am not able to reproduce any errors here.

Thanks again.

Ghostkeeper:
It's still failing in one spot, yes:

```
_ test_extruderMatch[/__w/Cura/Cura/tests/Settings/../../resources/extruders/fabxpro_extruder_0.def.json] _

file_path = '/__w/Cura/Cura/tests/Settings/../../resources/extruders/fabxpro_extruder_0.def.json'

    @pytest.mark.parametrize("file_path", extruder_filepaths)
    def test_extruderMatch(file_path: str):
        extruder_id = os.path.basename(file_path).split(".")[0]
        with open(file_path, encoding = "utf-8") as f:
            doc = json.load(f)
    
        if "metadata" not in doc:
            return  # May not be desirable either, but it's probably unfinished then.
        if "machine" not in doc["metadata"] or "position" not in doc["metadata"]:
            return  # FDMextruder doesn't have this since it's not linked to a particular printer.
        machine = doc["metadata"]["machine"]
        position = doc["metadata"]["position"]
    
        # Find the machine definition.
        for machine_filepath in machine_filepaths:
            machine_id = os.path.basename(machine_filepath).split(".")[0]
            if machine_id == machine:
                break
        else:
>           assert False, "The machine ID {machine} is not found.".format(machine = machine)
E           AssertionError: The machine ID FabX Pro is not found.
E           assert False

../tests/Settings/TestDefinitionContainer.py:185: AssertionError
```

You should be able to see these if you're logged into Github?

The problem is that the extruder file refers to `FabX Pro` but it should refer to the file name of the definition file: `fabxpro`.

hussainsail2002:
@Ghostkeeper thanks alot !!, that solved the issue.....

I was not aware that the machine name inside the extruder file should be the same as the definition file. I named it similar to what I had given inside the definition file itself....

The machine will be added to the next version of Cura ? I think Cura 4.5 was released today, 
Thanks again !

Ghostkeeper:
Yeah I expect it'll be in 4.6.

hussainsail2002:
Awesome, Thanks !

hussainsail2002:
I seem to be getting the following error when I have the machine setup locally

**Traceback (most recent call last):
  File "X:\4.5-msi\build\inst\lib\python3.5\site-packages\cura\BuildVolume.py", line 161, in _onSceneChangeTimerFinished
  File "X:\4.5-msi\build\inst\lib\python3.5\site-packages\cura\BuildVolume.py", line 529, in rebuild
  File "X:\4.5-msi\build\inst\lib\python3.5\site-packages\cura\BuildVolume.py", line 1107, in getEdgeDisallowedSize
  File "X:\4.5-msi\build\inst\lib\python3.5\site-packages\cura\Settings\ExtruderManager.py", line 236, in getUsedExtruderStacks
TypeError: unsupported operand type(s) for |=: 'bool' and 'str'**

hussainsail2002:
The error only shows up when I have the quality files turned on, seems there is some error in those

hussainsail2002:
The crash happens due to the following line that is placed in all the quality files:

support_enable = true

also the cooling fan does not get turned off even though the false value is provided

cool_fan_enabled = false

I have a feeling that the syntax for boolean is a bit different ?

hussainsail2002:
So the issue was that the true and false needs to be capital letters for the first letter, it needs to be True instead of true. But for the definition file it does not need to be so....

Ghostkeeper:
Yeah this is the issue with that we parse those setting values as Python expressions, unless the file format specifies otherwise. JSON allows boolean types, but uses a lowercase version. In the JSON files it will still need to be uppercase for the `"value"` property (as opposed to `"default_value"`) because that is a string containing a Python expression. A bit confusing I guess.

hussainsail2002:
Updated quality files on further testing

Ghostkeeper:
Developers, see ticket CURA-7305.

Ghostkeeper:
Why did you name profiles "PLA" or "ABS" rather than letting them be specified by the `material` metadata? This means you can load the PLA material in, but then select an ABS profile with it.

If you specify the material in the quality profile, you'd also see that there are no profiles available for other materials, such as CPE.

hussainsail2002:
> Expensive printer!

Haha , we use original e3d components so the cost picks up a bit...(heavy taxes for imports)

I am still not completely aware of the materials side of things. I saw another 3D printer brand that named their quality profiles in such a way. 

Is it possible to create our own brand of filament, the reason for this being I can fine tune the print temperatures. Also for certain materials such as ABS we need the cooling fan turned off and the raft enabled automatically, also certain materials like TPU may need to print slowly. I am not sure if this would be possible. 

> you'd also see that there are no profiles available for other materials, such as CPE.

Yes you are right and we wish to add as many materials as possible, will try this tomorrow and get back. Thanks a lot for the feedback.

Ghostkeeper:
> I am still not completely aware of the materials side of things. I saw another 3D printer brand that named their quality profiles in such a way.

For an example of profiles that are material-specific (other than just naming) see the Ultimaker 3 profiles: https://github.com/Ultimaker/Cura/tree/master/resources/quality/ultimaker3

Each of these profiles lists a material that it's compatible with:

https://github.com/Ultimaker/Cura/blob/39f41e8e7446d7c1f32f5a5bcd4126fb8b155cb9/resources/quality/ultimaker3/um3_aa0.25_ABS_Normal_Quality.inst.cfg#L11

The profile will only be shown if this material or any material with the same "material type" is selected (material type would be simply "ABS" rather than "Generic ABS"; it's basically the category of materials, as listed in the material manager).

> Is it possible to create our own brand of filament, the reason for this being I can fine tune the print temperatures. Also for certain materials such as ABS we need the cooling fan turned off and the raft enabled automatically, also certain materials like TPU may need to print slowly. I am not sure if this would be possible.

This is possible. You simply need to add them to the fdm_materials repository here: https://github.com/Ultimaker/fdm_materials

We keep these in a separate repository since Ultimaker's firmware also uses it. It's basically Ultimaker's filament database. You can either give these materials the same material type as the generic materials (probably easiest for the users) or a different one (which prevents the quality profiles from being used if people use different filaments).

> Also for certain materials such as ABS we need the cooling fan turned off and the raft enabled automatically, also certain materials like TPU may need to print slowly. I am not sure if this would be possible.

You can put this in the material profiles if you're creating those (would be preferred if it's the same for every profile of that material). It can have the same setting overrides as your quality profiles. Or you can put it in the quality profiles specific to that material.

hussainsail2002:
@Ghostkeeper Thanks !

I think I will use the current generic materials instead of creating a brand and just have the quality settings for them modified accordingly.

Ghostkeeper:
Nice, a lot more specialised profiles I see.

However the original profiles still specify ABS and PLA in their profile names. That shouldn't be necessary any more, right?

hussainsail2002:
@Ghostkeeper Thanks alot for that, I have changed those files.

I would like to know that once these files are released into CURA 4.6 and I need to make some changes will they get updated within the same version or it will happen only in the next version ?

ninovanhooff:
Seeing the length and detail of the discussion, and the last issues being solved, I'll go ahead and merge this.

@hussainsail2002 changes can only be shipped with a new version of Cura. We do release patch versions though, such as 4.4.1. See the releases page to get a feel for our release frequency.

hussainsail2002:
@ninovanhooff thanks !

Ghostkeeper:
A patch version is not likely to update profiles unless those profiles are causing blocking issues with a significant portion of the user base. We release patches only for blocking bugs, generally.

hussainsail2002:
@Ghostkeeper 

I would like to start adding the other machines we have, I would like to know if its possible to share the same quality files ?

Thanks 
Hussain

Ghostkeeper:
Yes, it is. You set the `quality_definition` metadata of the new machines to be the filename (without extension) of the printer you want to share the quality profiles with, e.g.:

    "quality_definition": "fabxpro",

Look to the Ultimaker 3 Extended definition for an example, if necessary.

Issue #5488, opened 2019-03-23
[4.0] Crash when clicking on "Preview"


muesli:
**Application Version**
4.0

**Platform**
Linux

Clicking on "PREVIEW" in the main-window crashes the application. Last output from terminal:

```
2019-03-23 13:22:54,104 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [55]: Loading shader file [/usr/lib/cura/plugins/SimulationView/simulationview_composite.shader]...
Exception: cura killed by signal segmentation fault (core dumped)
```

Johannes128:
I have the same problem (Archlinux, Cura 4.0).

fieldOfView:
Try the AppImage.

Johannes128:
The AppImage works. I will report the bug to the package maintainer of cura on Archlinux. Thank you very much!

rburema:
Thans for reporting anyway!

MathyV:
@Johannes128 has this been fixed in Arch already? I'm running into the same issue on Gentoo but I can't figure out what package / dependency is causing it.

rburema:
@MathyV I'm not sure if anything has been fixed in their repo's, but the AppImage should work for all distro's, since it has the dependencies baked in. It's also the only one we (at Ultimaker at least) support.

MathyV:
The AppImage works for me too, I'm however trying to create a working Gentoo package. I understand you only support AppImage but I'm looking for pointers what could be the possible reason, both the debug and strace don't tell me a lot, it fails in the call to load the shaders, that's it. I'm thinking in the direction of an outdated or too new dependency but I can't pinpoint it.

rburema:
@MathyV I know we've had some problems with Qt, locking us into a specific version on both sides for different reasons. (We're on 5.10.) You can also take a look at [cura-build-environment](https://github.com/Ultimaker/cura-build-environment) and [cura-build](https://github.com/Ultimaker/cura-build) for inspiration.

MathyV:
Ah, I previously locked PyQt5 on 5.9 because 3.x didn't build with 5.10. I'll check, thanks for the pointer.

rburema:
@MathyV You might even specifically need 5.10.0

Johannes128:
@MathyV No, up until now, the issue still persists on my machine on Arch... Arch currently uses PyQt 5.12.1 and Qt 5.12.2.

MathyV:
[...]

PR #7139, opened 2020-02-20
Avoid crash caused by KDE qqc2-desktop-style

StefanBruens:
Workaround/fix for #5488
Also see https://bugs.kde.org/show_bug.cgi?id=417900


cura_app.py
```diff
@@ -207,5 +207,11 @@ def exceptHook(hook_type, value, traceback):
     import trimesh.exchange.load
     os.environ["LD_LIBRARY_PATH"] = old_env
 
+# WORKAROUND: CURA-5488
```
nallath:
```suggestion
# WORKAROUND: Cura#5488 
```
CURA-number is used for our Jira tickets (which is an internal tracker). Github uses [reponame]#[issue_nr]

thopiekar:
Additionally, the fix shall be KDE-specific but is applied on every running Linux system.
Don't like my environment workaround/hack in my packaging files, too, but everything that goes into Cura here must be implemented perfectly in my opinion.

Ghostkeeper:
The default style shouldn't be used anyway, as everything should either be Cura-themed or system-themed. But of course there might still be fields that we don't specialise because they normally don't influence things - like inner margins of the slider handles or something - that make KDE crash. If they don't affect the layout of Cura I'm fine with accepting this change for all Linux distributions, or even all platforms. If it *does* affect the layout of Cura, we have a different problem on our hands that apparently the layout is not fully themed.

Ghostkeeper:
It does break a lot of stuff though :cry:

image

credomane:
This patch doesn't quite fix everything. Importing a STL file then importing a second one will actually import the first stl again and the second stl file. Also drop down's are misaligned but not a bit deal. At least the crashing is resolved with this patch.

I use the following snippet to fully fix Cura on KDE. Only bad part is the open file dialog isn't the fancy KDE one anymore and instead a minimal one. Everything works as expected and looks like it should. I actually forgot I applied my snipped below until I updated to Cura 4.5 just an hour ago and had to find it again.

```python
if Platform.isLinux():
    os.environ.pop("DESKTOP_SESSION", None)
    os.environ.pop("KDE_FULL_SESSION", None)
    os.environ.pop("XDG_CURRENT_DESKTOP", None)
```

protist:
FWIW a similar patch is mentioned [here](https://bugs.archlinux.org/task/62133).

Ghostkeeper:
I don't think we can merge this as it is now, since it seems to break the layout for other systems.

onitake:
@Ghostkeeper Can you be more specific how this breaks other systems?
Right now, Cura is practically unusable on many Linux systems because of this rendering issue. A fix is urgently needed.

Ghostkeeper:
I posted a screenshot of it there:

Screenshot

The things I can see on the main screen are:
* There are now margins between the setting category headers.
* Margins above and below every action button is not big enough.
* Drop-downs become too slim.
* Open file button too small.
* There's a red line on the active extruder tab.
* There are margins on the tool bar on the left.

cura_app.py
StefanBruens, line 210:
```suggestion
# WORKAROUND: Cura#5488
```

StefanBruens: Applied without extra whitespace

cura_app.py
```diff
@@ -207,5 +207,11 @@ def exceptHook(hook_type, value, traceback):
     import trimesh.exchange.load
     os.environ["LD_LIBRARY_PATH"] = old_env
 
+# WORKAROUND: Cura#5488
+# When using the KDE qqc2-desktop-style, the UI layout is completely broken, and
+# even worse, it crashes when switching to the "Preview" pane.
+if Platform.isLinux():
+    os.environ["QT_QUICK_CONTROLS_STYLE"] = "material"
```
StefanBruens, line 214:
```suggestion
    os.environ["QT_QUICK_CONTROLS_STYLE"] = "default"
```

StefanBruens:
@Ghostkeeper -- please check again:

Screenshot_20201208_145708

Ghostkeeper:
Yeah, that was it. It looks normal now to me.

PR #7140, opened 2020-02-20
fixed support layer height

skriDude:
Added `resolveOrValue('layer_height')` `support_infill_sparse_thickness` and `infill_sparse_thickness`


Ghostkeeper:
Looks good to me. It also removed some overrides that were unnecessary because the override is the same as the original.

There's a merge conflict but from the context of your message I know how to resolve that: Take the `resolveOrValue` formula.

PR #7144, opened 2020-02-21
CURA-7162/Replace the default logobot.svg with a new placeholder

DonExo:
Replace the old logobot.svg with a new placeholder.


nallath:
RIP

fieldOfView:
goodbye

I know of at least 1 place where Ultibot still rules. I'm not telling where, because then he'll get hit with the anti-toy-hammer.

nallath:
> I know of at least 1 place where Ultibot still rules. I'm not telling where, because then he'll get hit with the anti-toy-hammer.

It still lives within my heart :'(

ChrisTerBeke:
It's not as if developers have left UltiBot easter eggs throughout the ecosystem....

ChrisTerBeke:
The new image doesn't seem to render very sharp when enlarged:
image

DonExo:
> The new image doesn't seem to render very sharp when enlarged:
> image

I've added some property to the Image object in QML. It should be better now.

diegopradogesto:
> It's not as if developers have left UltiBot easter eggs throughout the ecosystem....

Including some snowflakes for next XMas XDDDD

DonExo:
@konskarm you might also want to remove the old logobot.svg file from the system.

PR #7147, opened 2020-02-22
CMake: Fixing warning about path conversion

thopiekar:
According to what I found in the docs[1], there is no "DIRECTORY" available as <type>.
Therefore, correcting it to "PATH", which is according to the docs correct.

At the moment I'm getting the following error[2]:
```
CMake Warning (dev) at CMakeLists.txt:8 (set):
  implicitly converting 'DIRECTORY' to 'STRING' type.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:9 (set):
  implicitly converting 'DIRECTORY' to 'STRING' type.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
[1] https://cmake.org/cmake/help/v3.6/command/set.html
[2] https://launchpadlibrarian.net/464670600/buildlog_ubuntu-focal-amd64.cura_1%3A4.5~202002121434~rev4433~pkg330~ubuntu20.04.1_BUILDING.txt.gz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,8 @@ include(GNUInstallDirs)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
-set(URANIUM_DIR "${CMAKE_SOURCE_DIR}/../Uranium" CACHE DIRECTORY "The location of the Uranium repository")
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I'm not getting these warnings myself. Maybe that got changed in later versions of CMake. According to the documentation you're right though. And with this change it still works for me.

PR #7154, opened 2020-02-24
Prune all sensitive data before sending it to Sentry

ninovanhooff:
CURA-7245

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -32,6 +32,8 @@ from UM.Resources import Resources
 from cura import ApplicationMetadata
 
 catalog = i18nCatalog("cura")
+home_dir = os.path.expanduser("~")
+
 
 MYPY = False
 if MYPY:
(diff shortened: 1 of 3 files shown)


cura/CrashHandler.py
```diff
@@ -83,6 +85,20 @@ def __init__(self, exception_type, value, tb, has_started = True):
         self.dialog = QDialog()
         self._createDialog()
 
+    @staticmethod
+    def pruneSensitiveData(obj):
```
nallath:
```suggestion
    def pruneSensitiveData(obj: Any) -> None:
```

cura/CrashHandler.py
```diff
@@ -83,6 +85,20 @@ def __init__(self, exception_type, value, tb, has_started = True):
...
+        if type(obj) is list:
```
nallath: I think it's a bit better to use isinstance instead (eg; It might happen that you get an ordered dict)

cura/CrashHandler.py
```diff
@@ -83,6 +85,20 @@ def __init__(self, exception_type, value, tb, has_started = True):
...
+            return [CrashHandler.pruneSensitiveData(item) for item in obj]
+        if type(obj) is dict:
+            return {k: CrashHandler.pruneSensitiveData(v) for k, v in obj.items()}
+        if type(obj) is str:
+            return obj.replace(home_dir, "<user_home>")
+        return obj
+
+    @staticmethod
+    def sentry_before_send(event, hint):
```
nallath:
```suggestion
    def sentryBeforeSend(event, hint):
```

cura_app.py
```diff
@@ -42,8 +43,9 @@
             sentry_env = "nightly"
     except IndexError:
         pass
-    
+
     sentry_sdk.init("https://5034bf0054fb4b889f82896326e79b13@sentry.io/1821564",
+                    before_send = CrashHandler.sentry_before_send,
```
nallath:
```suggestion
                    before_send = CrashHandler.sentryBeforeSend,
```

PR #7171, opened 2020-02-27
Add requirements.txt file

ninovanhooff:
The wiki page for bulding from source mentions some pip
dependencies but misses others. I suggest to put them here
and add a `pip3 install -r requirements.txt` to the wiki

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,7 @@
+colorlog
+pyserial
+sentry-sdk
+shapely
+PyQt5==5.10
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
Praise the Lord

fieldOfView:
Many requirements are missing though. Most notably numpy and scipy. But also numpy-stl, requests. NetworkX is an optional dependency of Trimesh so it does not get automatically pulled in.

PR #7182, opened 2020-02-28
[CURA-7016] Add actual public key instead of dummy.

rburema:
You'll need https://github.com/Ultimaker/Uranium/pull/584 to properly test this.


fieldOfView:
So the default build type is now "enterprise"?

rburema:
> So the default build type is now "enterprise"?

That shouldn't be the case.

> Uhm, shouldn't we fix this instead by just setting the build type to "enterprise" in our build system?

That was the idea!

I'd like to revert that change (not the public key, the other commit). Any objections?

ninovanhooff:
Wow, this is quite a wild mistake by me. Sorry people, I was under the impression that I only pushed the Uranium change, as indicated by the commit message mismatch.

rburema:
No problem, don't beat yourself up over this, could happen to anyone!

PR #7187, opened 2020-02-28
Show active postprocessing scripts count

### LINKED ISSUE #7159: Number of active post-processing scripts indicator
I run a post-processing script 100% of the time, so the icon (hammer and spanner next to slice/save) that would normally alert me to an active post-processing script is always there. There is currently no indication (that I know of) that would alert me when I forget to turn off some additional post-processing script that I don't need anymore but don't usually use.

Example: I needed to slow down some layers on a certain engineering model for higher accuracy. Used ChangeAtZ to achieve that, which worked flawlessly. Then I forgot to turn ChangeAtZ off. Hammer and spanner icon is always present, so that didn't remind me. Feedrate preview does not show speed after post-processing scripts do their thing, so that also couldn't alert me to my mistake. Next model took unnecessary extra time to print as well as having inconsistent outer wall due to different print speed.

It would be very nice to have a counter next to the already existing icon that would display the current number of active post-processing scripts, something like this:
Script_counter_example
I think this would be a simple and effective extension of the already existing alert for the user. I would for example always expect to see a "1", and seeing anything other than that would remind me to check the post-processing scripts before issuing the print.

Alternatively, I could always check the post-processing plugins before saving every print. I could also always check the final gcode by loading it into Cura again. Neither of those will work because that slows down my workflow too much. I'm also way too lazy to do that.

All users that permanently use a post-processing script (e.g. for info on the LCD during print) and occasionally use other post-processing scripts.

**Additional context**
-

fieldOfView:
This PR adds a "notification" count to the button that indicates there are postprocessing scripts active. It also adds a descriptive tooltip text, showing the name(s) of the active scripts without having to open the PostProcessing plugin dialog to inspect.

image

While fixing this, I also fixed some missing tooltips for ActionButtons.

Fixes #7159

### DIFF
--- a/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
+++ b/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
@@ -484,15 +484,36 @@ UM.Dialog
         onClicked: dialog.accept()
     }
 
-    Cura.SecondaryButton
+    Item
     {
         objectName: "postProcessingSaveAreaButton"
         visible: activeScriptsList.count > 0
         height: UM.Theme.getSize("action_button").height
         width: height
-        tooltip: catalog.i18nc("@info:tooltip", "Change active post-processing scripts")
-        onClicked: dialog.show()
-        iconSource: "postprocessing.svg"
-        fixedWidthMode: true
+
+        Cura.SecondaryButton
+        {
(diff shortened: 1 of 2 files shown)


nallath:
For our reference: CURA-7268

PR #7199, opened 2020-03-01
Adding 3DI D300 v2

v27jain:
Hi,
We are a 3D printer manufacturer based in India (3Deometry Innovations - www.3deometry.com) and would like to add our machine 3DI D300 v2 to the Cura ecosystem.

PR #7200, opened 2020-03-01
Create 3DI_D300_v2_extruder_0

v27jain:


v27jain:
Is it possible to find out what needs to be changed to make this a success ?

fieldOfView:
Don't put each file into a separate pull request. Add `3DI_D300_v2_extruder_0.def.json` to #7199

fieldOfView:
The file must have an extension `.def.json`

Issue #6099, opened 2019-07-31
Add post processing script to pause print for alphawise


olivier-boesch:
**Is your feature request related to a problem? Please describe.**
the alphawise U20 doesn't recognize the M0 gcode to perform a pause.

**Describe the solution you'd like**
A new script with the M25 gcode (this works).

**Describe alternatives you've considered**
I've written a modified version of the default script : https://drive.google.com/drive/folders/1VDTBfjTMyXb4ArnH09DDlfhRpaMfSOzd?fbclid=IwAR0zfXAu181OiVz2zjPBvi9Ym2yh677pVazcmue4E3SJ-QjubpiVJLkSZTY

or the default script could be modified to let the user enter le right gcode to perform the pause (M0 or M25 or other...)

**Affected users and/or printers**
Alphawise printer owners

Thanks.

Ghostkeeper:
So we now have one script for duet that has `M226`, one for Repetier that has `@pause`, one generic one that puts `M83` and one for filament change that puts `M600`. We should merge them into one script, I'd say.

Vandrasc:
As discussed internally, we will defer this for now as we do not have time for implementation but we do want to address it in the future.

olivier-boesch:
for my own needs, i will continue the dev of this script to make things work good with the U20.
I will be happy to contribute to this script in the future.

fieldOfView:
The BQ script uses the M25 gcode. But there's also an attempt to unify all pause at height scripts here: https://github.com/Ultimaker/Cura/pull/7201/files

PR #7201, opened 2020-03-01
Unify pause at height scripts

### LINKED ISSUE #7198: make the "pause at height" post process  context based
The Post process list is in my opinion a mess. instead of having one entry for "pause of layer height"
we have four entries that aren't even listed directly after another. And the difference is basically the used commands. If you don't read every entry it is possible to use the false one, resulting in abortet prints ,ect.

make it context sensitive to the printer and its code setting.

make a dropdown  menufor the version in the setting of the process

Basically everyone that uses Boards that doesn't use the marlin code set

fieldOfView:
This PR adds an option to the pause at height script to set the pause method (gcode command), instead of having 4 different pause at height scripts. This should make maintenance of these script less work, since fixes don't have to be made in 4 scripts. It also gives feature parity to all the different flavors.

TODO:
* ~automatically suggest the right method for the current printer definition~ see a7e23ba
* ~add repetier-specific parking if repetier flavor is selected~ see c87ee61
* ~make "griffin" a method~ see 87633e7
* ~consider the need to upgrade configurations for projects including a pause at height script~ see 321693d

Fixes #7198


nallath:
I think we should add an upgrade script for this (one way or another). Old projects will otherwise not be loaded correctly.

fieldOfView:
I agree, hence the WIP tag.

fieldOfView:
For the upgrade script, I'll wait for #7247 to be merged (or rejected), since that sets up the basics for a 4.5 to 4.6 upgrade script.

fieldOfView:
Also see https://github.com/Ultimaker/Cura/issues/6099

mahtDFR:
Relevant? #7227

agibson2:
Will this make is so that I don't need to maintain the hacked Octoprint @pause script?  For octoprint, all that should be inserted is a single @pause line.  Octoprint handles the scripts for the actual gcode commands to make pause and resume work.

https://community.ultimaker.com/topic/24105-post-processing-script-for-octoprints-pause-feature/

fieldOfView:
I'm waiting for #7427 to merge, in order not to create a big merge conflict. There needs to be a versionupgrade

nallath:
@fieldOfView https://github.com/Ultimaker/Cura/pull/7227 is merged.

fieldOfView:
Now that there is an upgrade step for 4.6 to 4.7 in master, I can add my work on the Pause at Height scripts to that.

fieldOfView:
Ready for review

ninovanhooff:
Internal reference: CURA-7453

agibson2:
Does this new revised unified script have an option to only insert a single @pause line for Octoprint?  Octoprint handles the commands to send to the printer to implement the pause and resume (resume on the web GUI)?  Looking at the code it seems like the only option for @pause is to use 'repetier' but I see other commands being put in too.

fieldOfView:
> Does this new revised unified script have an option to only insert a single @Pause (lower case p) line for Octoprint?

Did any of the previous scripts? If not, then neither does this unified one. It can easily be added, but please open a new feature request for that.

agibson2:
Feature request added for Octoprint...
https://github.com/Ultimaker/Cura/issues/7842

PR #7205, opened 2020-03-01
Fix watching for changes in AMF and trimesh-loaded files

### LINKED ISSUE #7192: Add reload dialog for AMF models
When overwriting an STL model file which is open in Cura, Cura detects this change and displays a helpful dialog offering to reload the model on the build plate. This doesn't work for models loaded from AMF files. Cura doesn't detect when the AMF source file changes.

It would be nice for Cura to also detect when the AMF source file is changed, and offer to reload the model on the build plate, just like with STL models.

fieldOfView:
This PR fixes the AMF and TriMesh readers to support watching the filesystem for modified files.

Fixes #7192

### DIFF
--- a/plugins/AMFReader/AMFReader.py
+++ b/plugins/AMFReader/AMFReader.py
@@ -118,7 +118,7 @@ class AMFReader(MeshReader):
                     mesh.merge_vertices()
                     mesh.remove_unreferenced_vertices()
                     mesh.fix_normals()
-                    mesh_data = self._toMeshData(mesh)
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
This works! Thank you very much.

I tried applying the same fix to the 3MF reader but there it didn't work the same way, sadly.

fieldOfView:
https://github.com/Ultimaker/Cura/pull/7230 should work for the 3mf reader

PR #7223, opened 2020-03-04
CURA-7255_infill_only_dropdown

ninovanhooff:
Note that per object settings can not be enabled due to a bug introduced in CURA-7211


BagelOrb:
Code looks good besides these super small things.

PR #7227, opened 2020-03-04
PP-plugin -> PauseAtHeight: steppers disarm timeout

ilyko96:
Small change in _PauseAtHeight_ script for *Post-Processing Plugin*, which adds `M18 S{VALUE}` command with a specified in a new _Pause Duration_ field number of seconds.

This overrides the default disarm timeout in firmware by user-defined value, which helps to prevent situation, when firmware disarms steppers and Z-axis falls down bcuz of weight which leads to loosing the Z-position and making it impossible to continue printing.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
+++ b/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
@@ -49,6 +49,17 @@ class PauseAtHeight(Script):
                     "minimum_value_warning": "1",
                     "enabled": "pause_at == 'layer_no'"
                 },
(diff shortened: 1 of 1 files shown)


nallath:
Thanks for contributing this. I hope you don't mind, but i've made some additional changes on 329c5c5c110642cbfa247c462b8c30a92dc3ea4f

I've renamed the setting so that it more correctly describes what it does (Decide when the steppers time out, not when the pause end).
I've also set the defaults of it so that the default behavior won't change. This is to ensure that when people would re-slice with an older version, the results will remain more the same.

PR #7230, opened 2020-03-05
Fix watching for changes in 3mf files

fieldOfView:
This PR fixes the 3MF reader to support watching the filesystem for modified files.

This PR is a followup to #7205

### DIFF
--- a/plugins/3MFReader/ThreeMFReader.py
+++ b/plugins/3MFReader/ThreeMFReader.py
@@ -86,7 +86,7 @@ class ThreeMFReader(MeshReader):
 
     ##  Convenience function that converts a SceneNode object (as obtained from libSavitar) to a scene node.
     #   \returns Scene node.
-    def _convertSavitarNodeToUMNode(self, savitar_node: Savitar.SceneNode) -> Optional[SceneNode]:
+    def _convertSavitarNodeToUMNode(self, savitar_node: Savitar.SceneNode, file_name: str = "") -> Optional[SceneNode]:
         self._object_count += 1
         node_name = "Object %s" % self._object_count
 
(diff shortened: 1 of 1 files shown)


nallath:
I've found one thing that's perhaps a bit weird with this. 

**Steps:**
1. Start up two instances of Cura.
2. Create a project file in Cura A. 
3. Load the project in Cura B.
4. Make a change in Cura A (Rotate the model), save it again.
5. Check Cura B

**Expected:** 
Cura B shows "File has been changed, do you want to reload" message 
After re-load, the model has the same orientation as cura A 

**Actual:**
Message is shown (OK)
Model does not have the same orientation (Maybe ok? IDK?)

Alternatively, you can similar weird-ish results if you remove / add an object to the 3mf. In that case the changes are also not executed (So instead of orientation step, you multiply the object). 

So what to do with this? In some cases, it does improves the behavior (eg; the mesh data is changed), but it's not quite there yet.

Ghostkeeper:
When reloading an STL, it also keeps the current orientation. That's probably why it's also kept for 3MF. There is no disambiguation between the transformations obtained from the project and the transformations applied by the user later.

I don't think there is an easy way to solve this.

nallath:
For our reference; CURA-7309

nallath:
Decided to merge this since it is an improvement over the default behavior (even if it's not completely the right one)

BagelOrb:
So it will be in 4.6?
Awesome!

This should fix my workflow. I should then be able to export from SolidWorks to 3mf and have my files updated without having to reposition and redo all of my support blockers. Or is that still not working?

BagelOrb:
> I don't think there is an easy way to solve this.

We're just talking about saving two transformations: one from the original source file and one from Cura. Saving two transformations in the project file shouldn't be that difficult, right?

Ghostkeeper:
It's in 4.6. We're still working out some kinks though. It's not reloading properly with multiple objects in the 3MF file.

Ghostkeeper:
> Saving two transformations in the project file shouldn't be that difficult, right?

It would also need to be saved in the scene.

BagelOrb:
Ah yes. Although you could also merge the matrices before, and when saving to file, apply the inverse of the original matrix to separate the two matrices again. All in all still not so easy.

PR #7243, opened 2020-03-08
Add Flying Bear Printers

oducceu:
Add Flying Bear Ghost 4S Printer.


oducceu:
Hello! Thanks for review. This is a community driven integration and we test different print settings a lot. These redundand variants and qualities were left as placeholders for future changes. I took your advice and cleared the unnessessary files.

PR #7247, opened 2020-03-09
Version upgrade 4.5 to 4.6 for Maximum Deviation changes

BagelOrb:
Includes value changes for `meshfix_maximum_deviation`, so that its behavior remains the same.

Required for https://github.com/Ultimaker/CuraEngine/pull/1214

Some profiles used to set the meshfix_maximum_deviation to an odd number of microns. I rounded them _down_ after halving their values.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -145,7 +145,7 @@ class CuraApplication(QtApplication):
     # SettingVersion represents the set of settings available in the machine/extruder definitions.
     # You need to make sure that this version number needs to be increased if there is any non-backwards-compatible
     # changes of the settings.
-    SettingVersion = 11
+    SettingVersion = 12
 
     Created = False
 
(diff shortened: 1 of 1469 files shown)


BagelOrb:
This doesn't include any changes to the version "4.5" to "4.6". Should it?
I only need the version upgrade system to update to a new settings_version, which is from 11 to 12.

I don't think doing the whole version 4.6 should be part of this PR.

fieldOfView:
Wow, 1400+ files need to change for such a small update. That puts startup performance into perspective...

BagelOrb:
Haha yeah. I always have the biggest PRs :D

fieldOfView:
It is sort of indicative though of a larger issue that the current approach of having all those quality profiles for all those printers in the main download is not tenable. Wasn't there once a plan to host them externally (Marketplace)?

ChrisTerBeke:
@fieldOfView there was, but it requires some refactoring regarding XML material profiles (they can contain printer-specific settings) so we never came around to it so far.

Ghostkeeper:
This PR is superseded by https://github.com/Ultimaker/Cura/pull/7385.

PR #7251, opened 2020-03-10
Make support render transparently

Ghostkeeper:
This way you can still sort of see what is being supported under there. Handy!

Totally not showing off my new ultra-widescreen monitor here...

However because the layer view is rendered on top of everything else, the Z buffering is a bit messed up. This makes the support always exactly the transparency value in the theme regardless of how many layers of support are drawn. That's nice, because it makes it so that you can always see through the support.

This doesn't work when the colour scheme of layer view is not set to Line Type. There is no differentiation between support and other things then. Support infill seems to not be used at all by layer view, but I've also made the support interface transparent just like the normal support.


fieldOfView:
Not only the skirt gets hidden by transparent support, but also the entire model.
image
image
Note how you can see the Ultimaker logo text behind the model through the supports.

What is the usecase of rendering transparent supports if you cannot see through them to see what is behind the supports?

Ghostkeeper:
Yes, it's a bit weird. It seems that the alpha channel is applied both when rendering support initially and then also applied to the flattened result of the g-code. I haven't completely figured this out yet but it's got to do with how we render the layer view on top of everything else in the scene. That's where it does the flattening in any case.

> What is the usecase of rendering transparent supports if you cannot see through them to see what is behind the supports?

Well, the thing is that you can see through them to see what is behind the supports. In your screenshot you can follow the silhouette of your model through the support.

PR #7264, opened 2020-03-11
Original Prusa Mini Integration

Krypta1:
Add support for the new Original Prusa Mini printer


resources/quality/prusa_3d/prusa_mini/040_nozzle/prusa_mini_040_025.inst.cfg
```diff
@@ -0,0 +1,48 @@
+[general]
+version = 4
+name = 0.25mm Layers
+definition = prusa_i3_mk3
```
Ghostkeeper: This one has the wrong definition.

Ghostkeeper:
These quality profiles don't get loaded since they are all marked as global quality profile:

`global_quality = True`

However they are also nozzle-specific so they won't get put in the global stack either. Did you test these profiles?

skeltoac:
@Krypta1 Please revisit this PR. :)

PR #7269, opened 2020-03-12
Options to Eliminate Stringing

neildmd:
I feel that a combination of retract while combing and z-hop is the current best route to eliminate stringing for most common prints. What we truly need is the ability to retract and hop after the wipe motion has occurred. This should eliminate stringing for parts where z-hop is required due to thing features curling up around the edges.

Another solution would be to incorporate a ramp option for z-hop where we can set a horizontal distance over which the hop takes place and have the ramp always go towards infill just like combing. If z hop could have a timing configuration option in this scenario that could also be very useful e.g. Wait X milli-seconds after retract before hopping.

Just some ideas, feel free to expand or this is already possible please inform me how.


Ghostkeeper:
Closing this pull request before we accidentally merge master into 4.5... Please just open an issue instead.

PR #7310, opened 2020-03-20
Store/retreive SceneNode name in/from <object> attribute

fieldOfView:
This PR stores and retreives the SceneNode name as attributes of the `<object.` node in 3mf files, in accordance with the spec:
https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md#attributes-3
It requires https://github.com/Ultimaker/libSavitar/pull/23 and CI will (should) fail until that PR is merged.

This PR solves the problem that after saving a project and loading it back in to Cura, all model names are changed to `Object n` (where n is a number referring to the order objects are stored in the 3mf file). This makes it hard to find models back in the model list, and makes eg the postprocessing script to display model names useless for projects. Additionally, the PR follows the 3MF spec more closely by supporting a known metadata entry that was previously unsupported.

### DIFF
--- a/plugins/3MFReader/ThreeMFReader.py
+++ b/plugins/3MFReader/ThreeMFReader.py
@@ -88,7 +88,10 @@ class ThreeMFReader(MeshReader):
     #   \returns Scene node.
     def _convertSavitarNodeToUMNode(self, savitar_node: Savitar.SceneNode, file_name: str = "") -> Optional[SceneNode]:
         self._object_count += 1
-        node_name = "Object %s" % self._object_count
+
+        node_name = savitar_node.getName()
+        if none_name == "":
+            node_name = "Object %s" % self._object_count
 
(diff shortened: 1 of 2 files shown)


fieldOfView:
Well shiver me timbers, CI does not protest about missing setName() and getName() methods in libSavitar.

Ghostkeeper:
I don't think we have automated CI that tests the conversion from libSavitar's nodes to Uranium's nodes.

nallath:
SHouldn't this already have been merged if it's counterpart is also merged?

fieldOfView:
I think @Ghostkeeper ran out of time yesterday

Ghostkeeper:
I had been waiting with merging this one since our build server needed to update its build environment. Until that completed (successfully) we shouldn't merge this, since that would cause 3MF reading to break.

It's completed now, so I'm merging this repository in.

PR #7328, opened 2020-03-22
Show support settings when support meshes are present in the scene

fieldOfView:
This PR shows relevant support settings in the advanced settings ("sidebar") when support meshes are present in the scene, even if the "Enable Supports" setting is off. It does this by adding a "disabled" setting "Scene Has Support Meshes", the value of which is controlled by the CuraSceneController. This way a user can change support settings to change how supports are printed (such as pattern, density and the extruder to use) even when there are only custom supports. Currently these settings are only reachable when the "Enable Supports" setting is turned on, which forces automatic support generation to be active.

In order for the setting value not to end up in profiles and projects, the PR does some trickery to manipulate the settingdefinition of the setting. If the team agrees with the functionality, and the general approach of using an application-controlled hidden setting, I will see if I can clean up the nasty bits.


fieldOfView:
Anything I can do/change to add this along?

Ghostkeeper:
Not really. We're trying to keep up with pull requests but it's hard.

Ghostkeeper:
I didn't find any flaws with this. Thanks for the contribution!

I did find another bug that the support meshes don't count towards the used extruders in the front-end, so if you only have a support mesh and set the support extruder to some otherwise unused extruder, it doesn't take that into account. But that's a separate bug that was in there all along. This is an improvement nonetheless.

PR #7330, opened 2020-03-22
Add method to add separators to extension menus

fieldOfView:
This PR allows extensions to add separators between menuitems in their Extensions submenu like this:
`        self.addMenuItem("", lambda: None)`
The method of reusing the existing `addMenuItem` instead of eg an `addMenuSeparator` means that (1) the method is fully backward compatible in the sense that this menuItem just renders as an empty menuitem and (2) confines the changes to Cura instead of requiring a change in both Cura and Uranium.

### DIFF
--- a/resources/qml/MainWindow/ApplicationMenu.qml
+++ b/resources/qml/MainWindow/ApplicationMenu.qml
@@ -74,7 +74,7 @@ Item
                             property var modelText: model.text
                             property var extensionName: name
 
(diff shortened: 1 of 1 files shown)


fieldOfView:
With PR:
image

Without PR (backwards compatible):
image

nallath:
Oh! I'm liking this one. This would also be usefull for my plugins (or well, the orientation one, not the barbarian plugin)

Issue #6688, opened 2019-11-21
two or more extruders are not created for delta


Goodfeat:
Hello. I encountered the error of slicing parts in Cura 4.4. (the program could not finish the cut after the start).
I decided that it was a profile issue from previous versions.

The next step I tried to create a new profile for delta
![create profile](https://user-images.githubusercontent.com/44455128/69306865-a740c180-0c74-11ea-9529-cb065c35b3b8.PNG)

However, when I went into the printer settings, I could not put two extruders.
![not extruders](https://user-images.githubusercontent.com/44455128/69306871-ab6cdf00-0c74-11ea-9964-9127a41b906f.PNG)

Windows 10
Cura 4.4
Delta printer
nozzle 0.4
dual extruder (extr 1 + extr.support)

Goodfeat:
I solved the problem. It turns out that the profile of the delta was not originally intended to be printed by two extruders.

The solution was simple. I added a new line in to the file "deltabot.def.json"
"1": "deltabot_extruder_1"

![add string](https://user-images.githubusercontent.com/44455128/69311631-84181100-0c78-11ea-8840-6b53e83e6531.PNG)

I also had to create an additional file deltabot_extruder_1.def.json

it was not implemented initially due to bugs?

rburema:
I'll ask around if there was a reason for that.

rburema:
Originally, deltabot wasn't in custom, since 'Custom FFF Printer' can also handle this use case. The recommended way to do this (for now) is to use 'Custom FFF Printer', where you _can_ select multiple extruders.

Pull requests are welcome, though!

Goodfeat:
I tried the method described above. It seems to work.

Ghostkeeper:
> I'll ask around if there was a reason for that.

History. Deltabot was made as an easier way to create a custom Delta printer. It's in fact older than Custom FFF Printer and was implemented before we had multi-extrusion support.

fieldOfView:
I don't know why Deltabot ended up in the "Custom" category. Having it there only confuses matters. The definition is a legacy from way back:
https://github.com/daid/LegacyCura/blob/SteamEngine/resources/machine_profiles/DeltaBot.ini

Goodfeat:
Indeed, this is confusing.
This option offers no benefits other than the selected circle-shaped bed.
... and only one extruder

Ghostkeeper:
[...]

PR #7331, opened 2020-03-23
Hide deltabot from Add Printer dialog

fieldOfView:
This PR hides the Deltabot definition from the Add Printer dialog. The Deltabot is used as a generic stub for delta-style printers. We want users to use the "Custom FFF Printer" instead, which is more fully featured (supports multiple extruders) and has the appropriate options to set an elliptical build plate and origin at center.

With this PR, user that have already configured a Deltabot can keep using their configuration, but users wanting to add a new delta-style printer for which there is no premade configuration will only be able to pick the Custom FFF Printer.

Contributes to #6688

### DIFF
--- a/resources/definitions/deltabot.def.json
+++ b/resources/definitions/deltabot.def.json
@@ -3,7 +3,7 @@
     "version": 2,
     "inherits": "fdmprinter",
     "metadata": {
-        "visible": true,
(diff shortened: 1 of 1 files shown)


rburema:
Internal link-back: CURA-7320

PR #7339, opened 2020-03-24
# Pause at height for FlyingBear Ghost

firexx:
# * tested with FBG4S
# * FBG4S don't make a pause on M0 command. but M25 can be used instead.
# * some default parameter changes that make more sense for this printer

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PauseAtHeightFBG.py
+++ b/plugins/PostProcessingPlugin/scripts/PauseAtHeightFBG.py
@@ -1,0 +1,363 @@
+# Copyright (c) 2019 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+# Pause at height for FlyingBear Ghost
+# * tested with FBG4S 
+# * FBG4S don't make a pause on M0 command. but M25 can be used instead.
+
+from ..Script import Script
+
+from UM.Application import Application #To get the current printer's settings.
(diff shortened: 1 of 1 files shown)


firexx:
can't understand how add a new PostProcessing script can cause the beak of build cura...

fieldOfView:
Note that there's an effort going on to unify all Pause at Height scripts into a single one, with a dropdown to select which method to use. Using an `M25` command is one of those methods.

See https://github.com/Ultimaker/Cura/pull/7201

firexx:
it can be also done, but not simple understandable for all users. not all are familiar with g-code commands. but your way is more logical for me also. it was a first idea to implement so like you suggest.

fieldOfView:
For now, the BQ version uses M25.

firexx:
this inserts the M25 on wrong place. so the pause is directly after heating bed and before heating hot end. that is completely wrong but this pause works :-D

what ist about to read also the scripts from users/<username>/AppData/Roaming/cura/<version>/plugins/PostProcessing/scripts

so the plugin for FlyingBear would install own script and i don't have to pull request to whole cura- where the most users of cura don't need that script.

fieldOfView:
> also the scripts from ...

Try `%APPDATA%\cura\"version"\scripts`

firexx:
cool that's work. thank you very much, i will now make a pull request to MKS plugin! Very nice!

PR #7341, opened 2020-03-24
Updated Cloud sign in strings CURA-7314

mahtDFR:
Updated Cloud sign in strings CURA-7314


rburema:
Failing build is due to already fixed earlier branch, merging anyway.

PR #7363, opened 2020-03-27
Fix xray composite shader for opengl 2.1

fieldOfView:
This PR fixes the xray_in_solid_view branch to work on legacy opengl renderers. textureSize() is not available in opengl 2.1, so we calculate the xray error image scale outside the shader (which is also a theoretical performance improvement because now the scale does not get computed over and over for each pixel)

This PR builds on top of https://github.com/Ultimaker/Cura/pull/6569 and has that branch as target.


BagelOrb:
Looks good.
`u_xray_error_scaling` is a bit of an awkward name. `u_xray_error_image_scaling` would maybe have been better.
Merging this into my PR nonetheless.

PR #7373, opened 2020-03-31
Fix xray in solid view

fieldOfView:
This PR contains some fixes for the xray_in_solid_view branch (#6569)

*  Disables the option in the message should hide the message only
*  Adds the patches from @smartavionics so it works on Raspberry PI 4s and, more importantly, legacy opengl renderers (openGL 2.1).

It also fixes resetting the preference, and factors out an unnecessary import.


BagelOrb:
Thanks, man!!

Merging this instantly, because code review and testing will be done in the other PR.

PR #7375, opened 2020-03-31
CURA-7240 De-emphasize bundled packages

rburema:
... because users probably want to manage their own installed plugins and materials, not the already bundled ones.

### DIFF
--- a/plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
+++ b/plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
@@ -31,7 +31,7 @@ ScrollView
                 right: parent.right
                 margins: parent.padding
             }
-            text: catalog.i18nc("@title:tab", "Plugins")
+            text: catalog.i18nc("@title:tab", "Installed Plugins")
             color: UM.Theme.getColor("text_medium")
             font: UM.Theme.getFont("large")
             renderType: Text.NativeRendering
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
image

Looking a bit funny if one of the sections is empty

fieldOfView:
I always sort of dislike the "My" moniker; it is a bit Windows XP. Can I suggest "Marketplace plugins" or "Plugins installed from the Marketplace" ?

rburema:
> Looking a bit funny if one of the sections is empty

> I always sort of dislike the "My" moniker; 

@Yian-Lai If you have some time, can you give your opinion about these?

Yian-Lai:
> > Looking a bit funny if one of the sections is empty
>
> > I always sort of dislike the "My" moniker;
> 
> @Yian-Lai If you have some time, can you give your opinion about these?

Can we hide the whole section if there is no materials/plugins?

I think "My installed XX" is more straightforward, it also covers cases of user manually install plugins/materials, like PPA.

fieldOfView:
> I think "My installed XX" is more straightforward,

But it is semantically incorrect. The plugins/materials are not mine. What is meant is that I installed them. But that is not what it says. If anything it should be "Plugins installed by me". 

> Jaime and I agreed to

Oh. I agree with that too, and like it much better.

rburema:
Closing this in favour of https://github.com/Ultimaker/Cura/pull/7378 (which is the same, but with all the kinks ironed out).

PR #7378, opened 2020-03-31
CURA-7240 De-emphasize bundled packages II: Electric Boogaloo

rburema:
Replaces https://github.com/Ultimaker/Cura/pull/7375 due to some slight miscommunication on my part.

### DIFF
--- a/plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
+++ b/plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
@@ -20,7 +20,6 @@ ScrollView
         width: page.width
         spacing: UM.Theme.getSize("default_margin").height
         padding: UM.Theme.getSize("wide_margin").width
-        visible: toolbox.pluginsInstalledModel.items.length > 0
         height: childrenRect.height + 2 * UM.Theme.getSize("wide_margin").height
 
         Label
@@ -35,6 +34,7 @@ ScrollView
             color: UM.Theme.getColor("text_medium")
             font: UM.Theme.getFont("medium")
             renderType: Text.NativeRendering
+            visible: toolbox.pluginsInstalledModel.count > 0
         }
 
         Rectangle
@@ -50,6 +50,7 @@ ScrollView
             height: childrenRect.height + UM.Theme.getSize("default_margin").width
             border.color: UM.Theme.getColor("lining")
             border.width: UM.Theme.getSize("default_lining").width
+            visible: toolbox.pluginsInstalledModel.count > 0
             Column
             {
                 anchors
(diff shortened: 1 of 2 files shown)


plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
```diff
@@ -106,8 +113,106 @@ ScrollView
...
+        Rectangle
+        {
+            anchors
+            {
+                left: parent.left
+                right: parent.right
+                margins: parent.padding
+            }
+            id: bundledPlugins
+            color: "transparent"
```
fieldOfView: A Rectangle that is "transparent" should be an Item

plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
```diff
@@ -106,8 +113,106 @@ ScrollView
...
+        Rectangle
+        {
+            anchors
+            {
+                left: parent.left
+                right: parent.right
+                margins: parent.padding
+            }
+            id: bundledMaterials
+            color: "transparent"
```
fieldOfView: A Rectangle that is "transparent" should be an Item

plugins/Toolbox/resources/qml/pages/ToolboxInstalledPage.qml
nallath: Nope, it has a border.

fieldOfView: My bad

plugins/Toolbox/src/Toolbox.py
```diff
@@ -757,6 +764,10 @@ def pluginsAvailableModel(self) -> PackagesModel:
     def pluginsInstalledModel(self) -> PackagesModel:
         return self._plugins_installed_model
 
+    @pyqtProperty(QObject, constant=True)
```
Ghostkeeper, line 767: Should this be constant? Some of the properties of these packages can change, such as `is_active`. I'm not sure how this works by reference.

I guess a test will tell. I'll accept this for now as long as we do test whether it's correctly displayed that a plug-in is inactivated.

Ghostkeeper:
I found one bug with this: There is an extra line underneath the list of installed plug-ins and the list of installed materials.
image

This line is not present in the original two lists.

plugins/Toolbox/src/Toolbox.py
Ghostkeeper: Turns out that this works fine. It's probably getting the model by reference and inside of that model things update, but QML is listening to those updates fine.

PR #7384, opened 2020-04-01
Cura 7333 fix reloading 3mf files with many objects

konskarm:
There was an issue where when reloading muliple objects from a 3mf file, object 1 was being replicated everywhere. 

This PR only fixes that issue to make sure that object 1 is not replicated, when the number of objects has not been changed. There is still undefined behavior when the number of objects or their positional indexes inside the file change.

Requires 

- https://github.com/Ultimaker/libSavitar/pull/24
- https://github.com/Ultimaker/Uranium/pull/596

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1574,7 +1574,6 @@ class CuraApplication(QtApplication):
     fileCompleted = pyqtSignal(str)
 
     def _reloadMeshFinished(self, job):
-        # TODO; This needs to be fixed properly. We now make the assumption that we only load a single mesh!
         job_result = job.getResult()
         object_to_be_reloaded = job.object_to_be_reloaded
         if len(job_result) == 0:
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -5,7 +5,7 @@ import os
 import re
 import sys
 import time
-from typing import cast, TYPE_CHECKING, Optional, Callable, List, Any
+from typing import cast, TYPE_CHECKING, Optional, Callable, List, Any, Dict
 
 import numpy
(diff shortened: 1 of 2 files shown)


konskarm:
Merge this along with https://github.com/Ultimaker/Uranium/pull/595

konskarm:
This PR fixes only the issue of object 1 replacing all objects when reloading a 3mf file.

Known issues that still need to be fixed (in a separate PR/ticket): 

1. When reloading a file from which an oject was deleted, the object remains in the scene and it is no longer refreshed. A warning will be logged in this case.
2. When reloading a file in which additional objects have been added (that do not exist in the scene), these objects will NOT be loaded.
3. Strange behavior will be noticed if the objects are rearranged in the file. For example if Object 2 is exchanged with Object 5, when reloading, these two objects will replace each other in the scene applying the transformations of object 2 to object 5 and vise-versa, as there is no way of telling which object is which.

cura/CuraApplication.py
```diff
@@ -1382,22 +1383,65 @@ def reloadAll(self) -> None:
...
+                job.object_to_be_reloaded = self._getObjectIndexInFile(file_name, node.getName())  # The object index in file to be loaded by this specific job
+                job._node = node  # type: ignore
+                job.finished.connect(self._reloadMeshFinished)
+                if has_merged_nodes:
+                    job.finished.connect(self.updateOriginOfMergedMeshes)
+
+                job.start()
+
+    @staticmethod
+    def _getObjectIndexInFile(file_name: str, node_name: str) -> int:
```
nallath, line 1411: I'm not really a big fan of this solution. Right now the name is generated from the objectId, but it's not really a future proof way of checking for it.
Soo, instead of that, I think it's better to add an objectID property that can be used to do this. That way, we keep open the ability to let the user change the names of objects.

PR #7385, opened 2020-04-01
Version upgrade 4.6 to 4.7 for Maximum Deviation changes

Ghostkeeper:
Previously we had a version upgrade from 4.5 to 4.6 that halves the maximum deviation in concert with this PR in CuraEngine: https://github.com/Ultimaker/CuraEngine/pull/1214
That version upgrade was proposed in a pull request here: https://github.com/Ultimaker/Cura/pull/7247

It seems that this won't make it in 4.6 any more, so we need to apply the same changes during the version upgrade to 4.7 instead.
Some differences with #7247 are:
* Now upgrade to setting version 13 instead of 12.
* Give a new name to the version upgrade plug-in to indicate that it updates to Cura 4.7.

### DIFF
--- a/plugins/VersionUpgrade/VersionUpgrade46to47/__init__.py
+++ b/plugins/VersionUpgrade/VersionUpgrade46to47/__init__.py
@@ -1,0 +1,59 @@
+# Copyright (c) 2020 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+from typing import Any, Dict, TYPE_CHECKING
+
+from . import VersionUpgrade46to47
+
+if TYPE_CHECKING:
+    from UM.Application import Application
+
+upgrade = VersionUpgrade46to47.VersionUpgrade46to47()
+
+def getMetaData() -> Dict[str, Any]:
+    return {
+        "version_upgrade": {
+            # From                           To                              Upgrade function
+            ("preferences", 6000012):        ("preferences", 6000013,        upgrade.upgradePreferences),
+            ("machine_stack", 4000012):      ("machine_stack", 4000013,      upgrade.upgradeStack),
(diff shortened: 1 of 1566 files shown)


plugins/VersionUpgrade/VersionUpgrade46to47/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+    return {
+        "version_upgrade": {
+            # From                           To                              Upgrade function
+            ("preferences", 6000012):        ("preferences", 6000013,        upgrade.upgradePreferences),
+            ("machine_stack", 4000012):      ("machine_stack", 4000013,      upgrade.upgradeStack),
+            ("extruder_train", 4000012):     ("extruder_train", 4000013,     upgrade.upgradeStack),
+            ("definition_changes", 4000012): ("definition_changes", 4000013, upgrade.upgradeInstanceContainer),
+            ("quality_changes", 4000012):    ("quality_changes", 4000013,    upgrade.upgradeInstanceContainer),
+            ("quality", 4000012):            ("quality", 4000013,            upgrade.upgradeInstanceContainer),
+            ("user", 4000012):               ("user", 4000013,               upgrade.upgradeInstanceContainer),
```
rburema, line 23: Shouldn't there be an upgrader for Intent in this list as well?

Ghostkeeper: We've never had this, since the user can't create intent profiles unless by manually modifying his configuration folder.

rburema: Ok, if it's on purpose, then I'm fine. We should _really_ keep an eye on this in case we ever make something like that though!

PR #7391, opened 2020-04-02
CURA-7340_fix_simulation_slider_layout

ninovanhooff:
Fixed: ActionPanelWidget's width was removed, 
which prevented a safe area from being layed out.
This in turn caused misalignment of the simulation slider


Ghostkeeper:
Sorry :grin:

Ghostkeeper:
Works perfectly fine, and no more binding loop either.

Issue #2058, opened 2017-07-03
Add Flow to Materials


ethanspitz:
I feel like Flow should be part of materials. Considering part my my calibration workflow is to recalibrate the flow per material, I'd like to be able to set that per material. Thoughts?

Ghostkeeper:
We are sort of limited by the XML material definition, which is not completely under Cura's control.

We had plans to add a cura namespace to the XML files so that Cura could add settings at will. But I see that the issue we had about that in our tracker was deleted for being too old, so I just had to re-create that...

fieldOfView:
> We are sort of limited by the XML material definition, which is not completely under Cura's control.

Which is kind of ridiculous, because it is an Ultimaker defined standard that noone else uses.

nallath:
I don't think it's that hard to add the flow to the material profile. Granted, we'd have to discuss it with some developers outside the Cura team, but that's not a big issue (I'd even go as far as to say it's always a good idea to discuss changes to a spec with multiple people)

Liger0:
Is there any updates? The flow doesn't have any reason to stay part of a profile instead of a profile.

fieldOfView:
For the benefit of those stumbling upon this issue: Check out the Material Settings plugin from the Marketplace. It will let you most settings to the Material pane of the preferences, saved per-material. Do note however that if you change the same setting value in the sidebar, that value will override what is set in the Material

Ghostkeeper:
For the record, this is still something I'm keeping in the back of the mind since it's quite a common request. But it's hard to make material profiles properly understandable to the user and fieldOfView's plug-in does as good a job as any.

nallath:
I think this issue can be closed; There is a way to do this via the material settings plugin.

PR #7392, opened 2020-04-02
FIlament specific flow settings.

Samanafatta:
It would be great to have flow settings that are specific for every filament you have in your list, so that you don't have to write the flow value on every spool you have, in order to remember it.

Thanks


Ghostkeeper:
Please create issues rather than pull requests to merge arbitrary branches.

https://www.quora.com/What-is-difference-between-pull-request-and-issues-in-GitHub?share=1

Ghostkeeper:
Or rather, just see the existing issue about this topic: https://github.com/Ultimaker/Cura/issues/2058

Or install the Material Settings plug-in from fieldOfView in the Marketplace.

PR #7413, opened 2020-04-05
Simplifying mp_mini_delta.def.json

PurpleHullPeas:
MP Mini Delta printer definition updates.

Removing minimum/maximum nozzle size, minimum layer_height, and minimum layer_height_0.

I'm doing this now because I barely knew what I was doing with the Cura json files back when I first added this printer.


Ghostkeeper:
Yeah I agree with these changes. The previous minimums are also a bit against the philosophy of Cura that it should warn the user when he's doing something that will not print nicely, but only actively prevent him if the change is physically impossible, such as a 0mm layer height.

Ghostkeeper:
That works fine. Bring on the 5-metre nozzles!

PR #7441, opened 2020-04-07
CURA-7262 X-Ray in solid-view (colors instead)

rburema:
hue_shift

See #6569 (which this is very much based off of) for the extended discussion. I put this as a separate PR because it didn't feel right forcing this into @BagelOrb s one.


fieldOfView:
Because the pattern deforms with the surface, it now looks like the problem may be on a surface where it is not.

BagelOrb:
What are the (intended) changes of this PR wrt my PR?

rburema:
@BagelOrb  Since @fieldOfView already took care of the legacy-OpenGL-issues (thanks for that b.t.w.!), this is basically a visual update, since UX found that the image, and the way it moved when moving the camera around will probably just visually overload users. This is indeed a compromise; it's technically correct that the _surface_ is not the problem, but o.t.o.h., it'll look more like an intended (albeit alerting) effect instead of a glitch.

(Oh, and the colors also adjust when the model is another color. The other colors are a triadic 'opposite' of the 'main' color, except for colors close to greyscale.)

BagelOrb:
Ok.

fieldOfView:
> the way it moved when moving the camera around will probably just visually overload users.

For the record (and perhaps not surprisingly) I disagree with this. I think the pattern should be in screen space (like it was in Tim's version) so it looks less like a texture on the surface.

fieldOfView:
"Why are parts of my modele I made in sketchup gift-wrapped?"

BagelOrb:
> For the record (and perhaps not surprisingly) I disagree with this. I think the pattern should be in screen space (like it was in Tim's version) so it looks less like a texture on the surface.

Me too. It *should* look like a hole in the model, not something added on the surface of the model.

BagelOrb:
I really don't understand why the Cura teams wants to visualize these errors **on the surface** of the model, when they occur **inside** the model.

To be fair: thie gift wrapping is quite geil!

fieldOfView:
Sure, just ignore the opinion of the person who has done most of the work on this feature.

rburema:
> Sure, just ignore the opinion of the person who has done most of the work on this feature.

Not every long meeting & discussion we have about this, with or without @BagelOrb is visible to you. At some point we have to be decisive.

BagelOrb:
> Sure, just ignore the opinion of the person who has done most of the work on this feature.

It's just my opinion. No worries. The basic functionality got merged, so I think we're better off than half a year ago, before we had the PR :)

rburema:
> It's just my opinion. No worries. The basic functionality got merged, so I think we're better off than half a year ago, before we had the PR :)

Thanks. Once you've rested for a long while, the team and you can have a chat, and we could still eventually decide to change it. For now, please take care!

PR #7453, opened 2020-04-09
Update in print quality

skriDude:


nallath:
These changes have issues with them. Pleas check the logs of the CI.

skriDude:
@nallath From what I understood there were two errors. One about "default_value" which I think is not showing up now after I've changed those to "value". But I don't understand the problem with the other error. Could you help me decipher it?

Ghostkeeper:
It's complaining that you're overriding settings that don't exist in the global stack. In particular: `extruder_nr`, `machine_nozzle_offset_y`, `machine_nozzle_offset_x` and `machine_filament_park_distance`. Those are settings that can't be set globally.

When looking at this profile I'm guessing that you filled this list automatically. It contains a lot of other mistakes too that are not caught by our automated tests. For instance, you're overriding the setting group `experimental` to have a value of `0`. That makes no sense at all! Many settings are also the same as their defaults.

skriDude:
Alright, we've changed how our plugin pulls the parameters data and it seems it's passed the automated test. Let me know if there's anything more that needs to be changed. :)

nallath:
I've changed one more thing (but it's really nitpicky); the spacing of the JSON was a bit inconsistent.

Issue #7252, opened 2020-03-10
Unable to slice/Outside build volume


Tom950:
**Application version**
4.5.0

**Platform**
Windows10
**Printer**
Ender 3 Pro

**Reproduction steps**
1. Load Stl
2. Try to slice

**Screenshot(s)**
![Screenshot_21](https://user-images.githubusercontent.com/55776886/76268737-67aedd00-6245-11ea-850b-059df222cf04.jpg)

**Actual results**
(What happens after the above steps have been followed.)

**Expected results**
Should slice like previous versions which slice fine with same files

**Project file**
https://we.tl/t-F901Jlg09f

**Log file**
2020-03-09 20:24:17,336 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [56]: Loading shader file [F:\Ultimaker Cura 4.5\resources\shaders\transparent_object.shader]...
2020-03-09 20:24:18,311 - DEBUG - [MainThread] cura.AutoSave._onTimeout [60]: Autosaving preferences, instances and profiles
2020-03-09 20:24:21,327 - INFO - [MainThread] CuraEngineBackend.CuraEngineBackend.slice [225]: Starting to slice...
2020-03-09 20:24:21,327 - DEBUG - [MainThread] CuraEngineBackend.CuraEngineBackend.slice [243]: Going to slice build plate [0]!
2020-03-09 20:24:21,334 - WARNING - [Thread-2] UM.Decorators.deprecated_function [20]: <function GlobalStack.extruders at 0x000001C07F8D7620> is deprecated (since 4.4): Please use extruderList instead.

**Additional information**
Settings in 4.5 are identical to 4.4 and under. Only 4.5 refuses to slice

mahtDFR:
Thanks for the report. For some reason the disallowed area for this profile has changed between versions. Generally we don't touch third-party profiles, and I can't see a history of a change made recently by anyone here, so I don't understand how this has happened. Maybe someone else can enlighten us.

The problem is in the definition file. This is located at `<Cura 4.5 installation>/resources/definitions/creality_ender3.def.json`

Find `"machine_disallowed_areas"`

Removing these lines fixes the issue:
`              [[-117.5, 117.5], [-117.5, 108], [117.5, 108], [117.5, 117.5]],
              [[-117.5, -108], [-117.5, -117.5], [117.5, -117.5], [117.5, -108]]`

[...]

PR #7466, opened 2020-04-10
Add Ender-3 Pro without clip boundaries for magnetic bed

kazooless:
In this discussion: https://github.com/Ultimaker/Cura/issues/7252

Some were pointing out the Ender-3 Pro comes with a magnetic bed without clips making the entire surface of the bed printable. The disallowed areas in the standard Ender-3 are not easily removed. @Ghostkeeper seemed to indicate a new "Pro" profile is an acceptable solution. So, here it is.

P.S. I'm still very new to GitHub. I made this change in the 4.6 branch instead of master this time. Is this correct? (preferred by you?)

### DIFF
--- a/resources/definitions/creality_ender3pro.def.json
+++ b/resources/definitions/creality_ender3pro.def.json
@@ -1,0 +1,28 @@
+{
+    "name": "Creality Ender-3 Pro",
+    "version": 2,
+    "inherits": "creality_base",
+    "metadata": {
+        "quality_definition": "creality_base",
+        "visible": true,
+        "platform": "creality_ender3.stl"
+    },
+    "overrides": {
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
> P.S. I'm still very new to GitHub. I made this change in the 4.6 branch instead of master this time. Is this correct? (preferred by you?)

For Github this is fine. You can aim a PR to be merged into any branch. However for the team's workflow it's incorrect. Before we start the public beta we always have a feature freeze. That means that only bugfixes are allowed in 4.6 at the moment. That's necessary to get the release stable. I hope you understand.

I'll change the base of this PR to master. That way it can get merged into 4.7 instead (which we will branch off in about 2 months).

Ghostkeeper:
Seems to work fine. No more clip collision areas. Thanks!

PR #7477, opened 2020-04-11
Adding profile for Anycubic Mega Zero

kad:
Dimensions: 220x220x250
Nozzle: 0.4
Print speed: 20-100 (40 for normal quality)

Vendor URL: https://www.anycubic.com/collections/anycubic-mega-3d-printers/products/mega-zero


resources/definitions/anycubic_mega_zero.def.json
```diff
@@ -0,0 +1,98 @@
...
+        {
+            "default_value": "Anycubic Mega Zero"
+        },
+        "machine_heated_bed":
+        {
+            "default_value": false
+        },
+        "machine_width":
+        {
+            "default_value": 224
```
Ghostkeeper, line 34: On the [spec page](https://www.anycubic.com/products/mega-zero) they mention that the build plate is 220x220. Was this size a deliberate deviation from that?

kad: updated to 220x220, with minor start code tweak.

kad:
@Ghostkeeper CI failure seems to be unrelated to my changes, as many other PRs are failing with same failed upgrade tests.

resources/definitions/anycubic_mega_zero.def.json
Ghostkeeper: Well if it's safe to use a bigger volume size, that's fine too. I was just making sure that it was intentional.

Ghostkeeper:
Yeah the failed tests are not your fault. Something slipped through, because we made a change that depended on a change in a different repository (and so obviously the tests failed in Cura since that change was not merged yet either). But when we merged both branches on both repositories, it turned out there was also an actual test failure. We should've ran the tests locally as well but I didn't...

kad:
@Ghostkeeper I rebased PR with few more additions (max feed and acceleration rates) based on recently published sources of the firmware: https://github.com/ANYCUBIC-3D/MEGA_ZERO/blob/master/Configuration.h#L611

Now CI tests are passing, so if there are no objections, please help with merging it. It would be great to get it into next version, it will simplify my life a bit in terms of Cura upgrades.

Ghostkeeper:
Works perfectly fine. Thanks!

PR #7483, opened 2020-04-12
Objectlist additions

fieldOfView:
This PR adds some information to the object list:
* indicates the extruder(-color) used for the model (for multi-extrusion printers only)
* shows the meshtype if the model is not a normal mesh
* shows the number of per model setting overrides
* indicates models that are outside the build volume

which box is which?
Clicking the mesh-type or per-model-settings-count will open the Per Model Settings tool panel.

The PR also adds a small fix to ensure that support_mesh models are rendered using the color of the support_extruder.

### DIFF
--- a/cura/Settings/SettingOverrideDecorator.py
+++ b/cura/Settings/SettingOverrideDecorator.py
@@ -94,6 +94,12 @@ class SettingOverrideDecorator(SceneNodeDecorator):
     #
     #   \return An extruder's position, or None if no position info is available.
     def getActiveExtruderPosition(self):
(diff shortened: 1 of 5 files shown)


fieldOfView:
I see Test_ObjectsModel needs some updating now that I have added a number of roles to the model.

fieldOfView:
Tests are fixed

Ellecross:
jesus christ. This is awesome. @fieldOfView you are the man.

nallath:
For our reference: CURA-7405

PR #7515, opened 2020-04-16
CURA-7147 Fetch preview image from cloud for monitor page.

rburema:


plugins/UM3NetworkPrinting/src/Models/Http/ClusterPrintJobStatus.py
```diff
@@ -57,6 +58,7 @@ def __init__(self, created_at: str, force: bool, machine_variant: str, name: str
                  build_plate: Union[Dict[str, Any], ClusterBuildPlate] = None,
                  compatible_machine_families: List[str] = None,
                  impediments_to_printing: List[Union[Dict[str, Any], ClusterPrintJobImpediment]] = None,
+                 preview_url = None,
```
ChrisTerBeke, line 61: type is optional string

rburema: oops, that slipped through, let me fix that real quick

rburema: Resolved with #ac5213d

PR #7539, opened 2020-04-20
Cura 7019 move sign in screen in front of add printer in first run wizard

konskarm:


resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -33,7 +33,7 @@ Item
         anchors.top: parent.top
         anchors.horizontalCenter: parent.horizontalCenter
         horizontalAlignment: Text.AlignHCenter
-        text: catalog.i18nc("@label", "Ultimaker Account")
+        text: catalog.i18nc("@label", "Sign in to Ultimaker Cloud")
```
ChrisTerBeke, line 36: We cannot use "Ultimaker Cloud" anymore as product name. Should be either "Ultimaker Account" or "Ultimaker Digital Factory".

konskarm: Thanks Chris. Fixed.

PR #7545, opened 2020-04-21
3MF platform meshes

Ghostkeeper:
This pull request replaces all platform meshes that were using the STL file format with 3MF. This saves about 30MB in storage and the download size of Cura for everyone (except people who checkout the Git history). Hopefully this helps not only our users, but also our testers that need to download this stuff regularly from our build servers.

I did a bit of a weird thing...

So I tried this initially a week or 2 ago on a Turbo Testing & Tooling session. Cura is not a good way to convert models to 3MF since it alters the positions of your models. That would mean that all meshes need to be realigned and such, a big hassle. Using libSavitar directly is a bit involved (I thought) and also contains some Curaisms in how it handles namespaces and how the data needs to be presented and retrieved. I couldn't find ANY other application that saves 3MF files on Linux. Other slicers, CAD software (free ones at least), online tools, nothing. I could have rebooted to Windows where there is 3D Builder from Microsoft. But noooooo I was lazy, so I wrote [my own command line tool](https://github.com/Ghostkeeper/ConvertTo3mf), a small C++ command line application that converts OBJ, STL and STL-binary to 3MF. Now there is a tool.

After converting it with ConvertTo3mf, I ran [ECT](https://github.com/fhanau/Efficient-Compression-Tool) on the result, which saved another 15% or so.

I expect that this pull request will:
* Save some download time for users and QA.
* Save storage space on our build server, reducing the frequency with which we need to clean old builds out.
* Reduce installation size a bit.
* Slightly improve loading time and time to switch printer types. After all, disk I/O is a major bottleneck usually.

If this is merged we need to decide on whether we still want to accept STLs from new pull requests. I expect that other people will follow suit if they see the 3MF files, but some might not due to difficulty of creating 3MF files (unless you have a tool like mine or know that 3D Builder does it). We also need to see if we want to reduce the maximum file size for platform meshes. The largest mesh among these is now 178kB.

I couldn't convert the OBJ files to 3MF. Those files are OBJ because they contain texture coordinates (and STL doesn't support those) which Cura needs to draw the image on the build plate. Cura doesn't load texture coordinates from 3MF files so I couldn't use 3MF there.


nallath:
CURA-7416

PR #7547, opened 2020-04-21
Update sdk version to 7.2.0

ninovanhooff:
CURA-7383

### DIFF
--- a/cura/ApplicationMetadata.py
+++ b/cura/ApplicationMetadata.py
@@ -13,7 +13,7 @@ DEFAULT_CURA_DEBUG_MODE = False
 # Each release has a fixed SDK version coupled with it. It doesn't make sense to make it configurable because, for
 # example Cura 3.2 with SDK version 6.1 will not work. So the SDK version is hard-coded here and left out of the
 # CuraVersion.py.in template.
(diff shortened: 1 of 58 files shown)


fieldOfView:
Could this sort of thing please be done before the beta release instead of before the final release?

Ghostkeeper:
Yeah, that would've been better. However it should not affect any current plug-ins since SDK 7.1 is loaded fine in 4.6 even if the SDK version is a minor version higher. This change is only to allow plug-ins to target 4.6 without targeting 4.5.

fieldOfView:
> it should not affect any current plug-ins

In some of my plugins (like the SidebarGUI plugin), it would be very nice if I could rely on the SDK number to differentiate between Cura releases because I need to apply specific patches for specific versions of Cura. I can't rely on the Cura version, because the Cura version is not available for those running a master branch (eg from a from source checkout from any point in time, or a smartavionics build, or a ppa master build).

Ghostkeeper:
Ah, so now you're able to differentiate between 4.5 and 4.6 then. I hope that at least for the most recent SDK version you're comparing with `>=` rather than `==`? Or were you using `==`?

PR #7552, opened 2020-04-21
Adding the mesh for the FLSUN QQ-S thanks to Eddy Emck, Anycubic Predator added

curso007:


fieldOfView:
This mesh is way too large to include with Cura at 22 MB. Please make a smaller (less detailed) version.

curso007:
OK I will take a look, how I can reduce this mesh.

Ghostkeeper:
We strive for a limit of 1MB per printer. That includes the mesh files (there's a 900kB PNG file there too). Less is better. If all of the 220 printers in Cura would have a 22MB platform mesh, we'd be looking at a download size of almost 5GB in platform meshes alone. Less is better; it's only displayed as flavour backdrop, after all.

The FLSUN QQ-S platform mesh has a LOT of detail with a gearbox at the top, a Bowden tube that probably only gets in the way for the user, a fan down at the bottom and lots of rivets and screws all over the place. Honestly it's probably more clear for the user to cut off the entire top half, so they can properly use their build plate (although it gets rendered in front of the mesh).

The Predator build plate is less bad with this (and actually has the top half cut off by the looks of things), but owes a large part of its file size to all of the rivets and screw wire as well.

curso007:
I reduced the sizes at a acceptable level this time I think.

resources/definitions/predator.def.json
```diff
@@ -6,7 +6,7 @@
         "author": "ksihota",
         "manufacturer": "Anycubic",
         "file_formats": "text/x-gcode",
-        "platform": "predator_platform.obj",
+        "platform": "predator_platform.stl",
 		"platform_texture": "predator_image.png",
```
fieldOfView, line 10: png image was removed, so please remove this reference to it

fieldOfView:
Please use spaces for indentation consistently

resources/definitions/predator.def.json
curso007: Weird the image on my side is present. I will rename to force the push (I hope).

Ghostkeeper:
(Tests concerning this printer definition are passing. It's just that currently on Master the tests are failing due to https://github.com/Ultimaker/Cura/commit/b21b36924c8e4c1646833a6d4e7d437de8fe985b .)

Ghostkeeper:
I did a squash-merge in order to prevent the huge meshes from ending up in the commit history. Github doesn't detect that properly, but they are in there: https://github.com/Ultimaker/Cura/commit/68e8261c2124415814b89809fc8c914e2e4f3bde

Thanks for the help!

PR #7473, opened 2020-04-11
added some Tronxy printers

Spanni26:
Added some definitions and a base extruder for tronxy printers

PR #7553, opened 2020-04-21
Patch for 4.6 that added support for Tronxy machines

64bittuning:
A complete patch from 4.5 that added support for Tronxy machines
I had assembled this "patch" to add Tronxy machines to the 4.5 version of Cura. It is currently live on thingiverse and has received a bunch of positive feedback. I was hoping to get this loaded to Git before you dropped 4.6, but Ive never used Git before.


fieldOfView:
Cura 4.6 has already been released.

fieldOfView:
How does this compare to https://github.com/Ultimaker/Cura/pull/7473 ?

64bittuning:
I am aware that 4.6 has already been released. If I have to, I will rebuild the files in 4.6 and verify that they are functional.

#7473 is extremely bare bones and doesn't have quality profiles, material settings... Etc. My package addresses the Tronxy X5SA machines by build area specifically and is much more complete.

I have the package installed on several machines, it runs and produces greatly improved results over the "custom fff printer" profile that we are forced to use.

I have had this together for a while, but had no idea how to upload to GitHub. Its quite possible that spanni jumped on my idea as I had posted it to The Official Tronxy Facebook group a while ago and he asked if I wanted to collaborate... Meanwhile, I was almost done building the pack.

fieldOfView:
* This PR reverts changes in `fdmprinter.def.json` and `fdmextruder.def.json`, which it should not.
* `tronxy_base.def.json` seems to be missing

Spanni26:
My request to the FB group was a few days before his announcement. So I can't have jumped on any train.
My intention is to make Cura more usable to the tronxy users, not to see my name in the configs.
I take back my PullRequest.

nallath:
Check the results of the CI, there are quite a few issues with this.

fieldOfView:
The files 'resources/materials/generic_abs_175.xml.fdm_material', 'resources/materials/generic_petg_175.xml.fdm_material' and 'resources/materials/generic_pla_175.xml.fdm_material' should not be part of this PR, but should be in a PR on the fdm-materials repository: https://github.com/Ultimaker/fdm_materials

fieldOfView:
In all your *.inst.cfg files, the `settings_version` should be 13 instead of 11.

64bittuning:
*.inst.cfg files have been updated to reflect the correct "settings_version"
uploaded tronxy_base.def.json
revised other definition files with new print head dimensions
removed improperly uploaded fdm* files
removed materials files

64bittuning:
big big big update coming in the next hour or so.

64bittuning:
Downloaded and installed 4.6.
Dropped my package in and verified proper operation.

Ghostkeeper:
It looks like you copy-pasted the quality profiles from one printer to the others. I'm assuming this just by sampling the `***0.2_petg_super.inst.cfg` files.

Please use the `quality_definition` metadata instead. You can reuse another printer's quality files if they are the same. This saves Cura from loading up another 500+ files on start-up to see which printer they belong to. Here is an example of the Ultimaker 3 Extended specifying that it uses the quality profiles of the normal Ultimaker 3:

https://github.com/Ultimaker/Cura/blob/f4f990708fd7b0d2cd4413c741acf6c09a0f9d23/resources/definitions/ultimaker3_extended.def.json#L8

Or perhaps more aptly, here is an example of a specific Creality printer specifying that it uses the same quality profiles as the other Creality printers (which also use the same as creality_base):

https://github.com/Ultimaker/Cura/blob/f4f990708fd7b0d2cd4413c741acf6c09a0f9d23/resources/definitions/creality_ender3.def.json#L6

64bittuning:
Closing this until I can figure out why the profiles dont work without having one setup for each machine.

Ghostkeeper:
Maybe going by the example of `creality_base` would be the most appropriate here.

PR #7559, opened 2020-04-22
Move the getCfgVersion to base version upgrade

nallath:
They can still be overridden, but in the current situation they were
all the same.


nallath:
Note; tests are failing because it needs the matching branch of Uranium to work.

PR #7561, opened 2020-04-22
HMS434 update

maukcc:
material exclusion


maukcc:
That is correct, we do not want any materials to show up. No error state, it picks one of the 2.85 materials for some reason. Preferably we would have it blank.
Can you imagine our frustration every time someone adds a new material?

Ghostkeeper:
Then please instead set `"has_materials": false,` in this printer's metadata. It will be blank then and the user can't change it any more either.

It would also be better to remove the `material = generic_pla` from the HMS434 quality profiles then, or maybe merge those into the global profiles altogether.

Ghostkeeper:
On second thought I'll reopen this to just allow you to change it in this PR. Since you're working from your own master branch you'll probably include the original commit anyway.

maukcc:
We can not do "has_materials": false, as we do supply a .package file to our customers where all of the materials and quality profiles are in.
We just do not want all of the other silly materials to show up.

maukcc:
We went for option 2 :)
Is there anything else you need from our side?

Ghostkeeper:
Thanks! That was probably the best solution in the end.

I'll fix the merge conflicts and preferred quality type and get this in for the next release.

It's really nice how you guys keep this printer so well up to date, by the way. It's good to see an active maintainer.

PR #7575, opened 2020-04-23
Cura 7367 update image and copy in enterprise version

jellespijker:
Removed the logobot and revised the text that is shown when an enterprise user tries to acces the marketplace and is not signed in.


rburema:
Congrats on you first PR as part of the team!

rburema:
Shame about the robot though, Fortunately it's not _completely_ gone, as `ultibot.svg` is still used.

PR #7622, opened 2020-04-29
Tronxy Support pack for 4.6 Version 2 

64bittuning:
This pack has been heavily reduced by 500 files or more.
Some more work to the profiles is required and should be sorted out shortly.
@Ghostkeeper Thanks for the tip on the quality_definition. Once I established that *_base wasnt working, I was able to get the associations to flow more smoothly.


Ghostkeeper:
Thanks for reworking it! This makes it a lot more maintainable and slightly more performant.

resources/definitions/tronxy_x.def.json
```diff
@@ -0,0 +1,267 @@
...
+
+        "has_materials": true,
+        "has_variants": true,
+        "has_machine_quality": true,
+        "variants_name": "Nozzle Size",
+
+        "preferred_variant_name": "0.4mm Nozzle",
+        "preferred_quality_type": "standard",
+        "preferred_material": "generic_pla",
+        "exclude_materials": [
```
fieldOfView, line 24: You are excluding a lot of materials that will not be shown anyway because the filament diameter does not match the filament diameter of the printers: `generic_*` and `ultimaker_*`

64bittuning: Not sure if you watch Dragonball Z but... *This isnt even my final form*
I re-uploaded this as a draft so I could work between two PC's but still utilize the build check on the page. MOST of the files in this are being revamped and should be done this weekend. The large majority of those exclusions were done due to filament diameter. I was unaware that the program would ignore them due to that value. This and various other changes will be rolled into the next file drop. Thanks for the heads up!

64bittuning:
@Ghostkeeper and @fieldOfView 
Everything is all set from my end. Is there anything else I should adjust?

fieldOfView:
Getting into nit-pick territory now, but every `"quality_definition": "tronxy_x",` line has a tab vs spaces inconsistency.

64bittuning:
@fieldOfView  My text editor (Notepad ++) didnt recognize any issues there. Do you have a preference on a viewer where I can spot check all these things? I have repaired that line of the files and re-pushed.
Nit-pick all you want. Ill make the changes.

Ghostkeeper:
You can make whitespace visible in Notepad++, but I don't think it has an option to make it only available before the line, so you'd have to dig through their plug-ins or else deal with the ugly newline symbols everywhere.

Myself I'm using a [Firefox plug-in](https://github.com/glebm/render-whitespace-on-github) that displays it for me in a pull request, and a fully-fledged IDE when I'm working on my own code.

If I see such problems normally I would just fix them myself rather than having to ping-pong with the author of the code, honestly...

64bittuning:
@Ghostkeeper  I dont mind having to go in a clean up a few bits and pieces. The pain was removing the 500+ files from the last pack. But its working now. Hopefully as long as nothing else nit-picky pops up, I'll get the green light.
I made the pack in 4.5 and rebuilt it for 4.6/4.6.1
There is, as far as I know, currently almost 400 people using the package as it was loaded to the support group. I tried to be very meticulous with it on my own machine before sending it out to anyone else.
Cura-Tronxy Pack
Is it possible that I missed something? Absolutely! But from my end, it all seems to be working properly.

resources/definitions/tronxy_x.def.json
```diff
@@ -0,0 +1,167 @@
...
+        "skin_overlap": { "value": 10.0 },
+        "infill_wipe_dist": { "value": 0.0 },
+        "wall_0_wipe_dist": { "value": 0.0 },
+
+        "fill_perimeter_gaps": { "value": "'everywhere'" },
+        "fill_outline_gaps": { "value": false },
+        "filter_out_tiny_gaps": { "value": false },
+
+        "retraction_speed": {
+            "maximum_value_warning": "machine_max_feedrate_e if retraction_enable else float('inf')",
```
Ghostkeeper, line 104: This is the warning value. You can't see the warning if the setting is not enabled, so why does it need to become infinite then?

Ghostkeeper:
I'd like to rename these profiles so that they have a more specific name to Tronxy. Otherwise you'd get the same problem as what we currently have with the Creality printers where the quality profiles are just called `base_something_something`. As you've probably seen, the file names need to be unique globally so to call it `x_0.2_ABS_fine` is going to muddle the namespace a bit.

This is quite easy to do before it's been published. After it's been published though we would need to change that in the configuration folders on the computers of all of our users which is more error-prone.
However that would mean that for you specifically you would get a corrupt configuration folder if you already had a Tronxy printer added.

PR #7623, opened 2020-04-29
CURA-7407 Fix x-ray error colour

rburema:
X-ray errors showed as green and translucent instead of read and solid.

CURA-7407


mahtDFR:
**X Ray View error color wrong.** The error color in the X Ray View turned translucent green for some reason. This version fixes it so it's a more logical solid red.

[[Pull request](https://github.com/Ultimaker/Cura/pull/7623)]

PR #7640, opened 2020-05-01
Make Infill Line Directions visible for cubicsubdiv infill pattern.

smartavionics:
Engine can now rotate this infill pattern so let the user specify an angle.

Only the first angle in the array is used.


Ghostkeeper:
Should be merged together with https://github.com/Ultimaker/CuraEngine/pull/1248.

PR #7655, opened 2020-05-02
Adding ADIMLab Printer (Gantry and Base for future models)

jprad99:
Based off of the Creality_Base and Creality CR-10S4 profiles, but modified to reflect the direct drive extruder, as well as a different start/end gcode.


resources/definitions/adimlab_base.def.json
```diff
@@ -0,0 +1,268 @@
...
+        },
+
+        "has_materials": true,
+        "has_variants": true,
+        "has_machine_": true,
+        "variants_name": "Nozzle Size",
+
+        "preferred_variant_name": "0.4mm Nozzle",
+        "preferred_quality_type": "standard",
+        "preferred_material": "generic_pla",
```
nallath, line 23:
```suggestion
       "preferred_material": "generic_pla_175",
```

resources/definitions/adimlab_base.def.json
```diff
@@ -0,0 +1,268 @@
...
+            "octofiber_pla",
+            "polyflex_pla",
+            "polymax_pla",
+            "polyplus_pla",
+            "polywood_pla",
+            "structur3d_dap100silicone",
+            "tizyx_abs",
+            "tizyx_pla",
+            "tizyx_pla_bois",
+            "ultimaker_abs_black",
```
nallath, line 71: There is no need to add all these materials, since they are 2.85mm (and thus won't be shown)

resources/definitions/adimlab_gantry.def.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "ADIMLab Gantry",
+    "version": 2,
+    "inherits": "adimlab_base",
```
nallath, line 4: If there is only one machine that uses this base, merge it into a single definition.

nallath, line 4: If there aren't actually multiple machines using this, please merge this into a single file

resources/definitions/adimlab_base.def.json
nallath, line 71: All Ultimaker materials are 2.85mm, so they don't need to be excluded. They won't show up.

nallath:
A small reminder, could you have another look at the review comments?

nallath:
Closing this due to inactivity.

PR #7693, opened 2020-05-06
CURA-7426 Layerview can now deal with visibility extruder nrs. > 4

rburema:
You'll need
https://github.com/Ultimaker/Uranium/pull/606
to make the per-layer simulator not crash :-)

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -77,7 +77,7 @@ class SimulationPass(RenderPass):
             self._layer_shader.setUniformValue("u_max_thickness", 1)
             self._layer_shader.setUniformValue("u_min_thickness", 0)
             self._layer_shader.setUniformValue("u_layer_view_type", 1)
-            self._layer_shader.setUniformValue("u_extruder_opacity", [1, 1, 1, 1])
+            self._layer_shader.setUniformValue("u_extruder_opacity", [[1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1], [1, 1, 1, 1]])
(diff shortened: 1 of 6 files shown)


fieldOfView:
So the only reason a matrix is used is that this offers 16 floats? Wouldn't a 1D sampler have been a lot more semantically correct?

rburema:
Yes, but performance wise it would require more calls on the python side & a sampler-lookup on the glsl side. Now it could be that I'm overreacting when it comes to this and it could be it turns out it doesn't matter much, but I rather play it safe, since people are already complaining & this approach alters the least.

fieldOfView:
Perhaps a tad more documentation could be added (both in python and in glsl) because this fixes the problem only up to 16 extruders, and the next dev who is going to fix that may wonder why a Matrix is used at all.

rburema:
That's a good point. I'll try to have a look later today, I'm in the middle of something complicated.

Ghostkeeper:
Note that CuraEngine also supports up to 16:

https://github.com/Ultimaker/CuraEngine/blob/122440013a9c18de97424b1f4644c83dcc234093/src/settings/Settings.h#L11

fieldOfView:
But at least there it is a single #define, which can be changed with a recompile. Cura is fundamentally limited to 16 because there are only 16 floats in a 4x4 matrix.

rburema:
Still better than fundamentally limited because there are only 4 floats in a length-4 vector, while the rest of the program supports 8, confusing users.

rburema:
See: https://github.com/Ultimaker/Cura/commit/23cdf871c1f6b1bb49311d295370314f2669fc14

PR #7708, opened 2020-05-07
Add a preference to use a single instance for loading files

### LINKED ISSUE #7664: Allow to enforce --single-instance via config file
Opening new instances of Cura 4.x takes longer and longer with each version. Fusion 360 seems to not use the --single-instance command line parameter when starting Cura and the default file association that is created by the Cura installer also seems to not use it.

It should be possible to enforce the use of --single-instance even if not specified on the command line by setting a config file setting and/or preference that ensures that --single-instance is always used when starting Cura. This would reverse the default behavior of creating a new instance each time. 

Wrapping the Cura.exe in a batch file would be a workaround but would require additional changes to the default file associations that are not easy to do for end users.

Most users of Cura would benefit from this behavior that was the default in earlier versions of Cura like 15.04.

**Additional context**
<none>

fieldOfView:
This PR adds a preference to use a single instance for loading files which has the same functionality as the --single_instance command line argument. ~It adds a very minimal preference parser that specifically looks for this particular preference instead of spinning up the full preference singleton, so the early boot is not delayed by much for this feature.~ Edit: the preferences are already available at that point, so I can use them as is.

The "single instance" feature is commonly used by users of CAD applications that launch Cura from their CAD application, and who may prefer to sequentially load objects into a single instance of Cura instead of starting up a new Cura for each iteration of the design.

This PR fixes #7664

### DIFF
--- a/resources/qml/Preferences/GeneralPage.qml
+++ b/resources/qml/Preferences/GeneralPage.qml
@@ -72,6 +72,9 @@ UM.PreferencesPage
         var defaultTheme = UM.Preferences.getValue("general/theme")
         setDefaultTheme(defaultTheme)
 
+        UM.Preferences.resetPreference("cura/single_instance")
+        scaleToFitCheckbox.checked = boolCheck(UM.Preferences.getValue("cura/single_instance"))
+
         UM.Preferences.resetPreference("physics/automatic_push_free")
         pushFreeCheckbox.checked = boolCheck(UM.Preferences.getValue("physics/automatic_push_free"))
         UM.Preferences.resetPreference("physics/automatic_drop_down")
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
A similar pre-check also exists for the language to display in the splash screen, doesn't it? Or is that one again too late? We may need to reconsider the order of initialisation to prevent adding a third.

fieldOfView:
I did not know about that. I'll have a look into making them into one.

fieldOfView:
Turns out the Preferences are already fully loaded at this point, so I could simplify the implementation a lot. Thanks @Ghostkeeper for paying attention.

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -72,6 +72,9 @@ UM.PreferencesPage
         var defaultTheme = UM.Preferences.getValue("general/theme")
         setDefaultTheme(defaultTheme)
 
+        UM.Preferences.resetPreference("cura/single_instance")
+        scaleToFitCheckbox.checked = boolCheck(UM.Preferences.getValue("cura/single_instance"))
```
nallath, line 76: I will fix this one for you :) (I'm fairly sure this should be singleInstanceCheckbox.checked)

fieldOfView: It should, thanks!

PR #7716, opened 2020-05-08
Limit platform meshes to 1MB

konskarm:
As a result of Cura and Cookies discussion, we decided to limit the
sizes of platform meshes to less than 1MB. This test checks that
during the CI/CD process and displays an assertion error when a
mesh excludes this limit.


Ghostkeeper:
Works. At least locally for me.

PR #7726, opened 2020-05-09
Add PostProcessingPlugin script DisplayProgressOnLCD

Bostwickenator:
This script is intended increase the accuracy of print progress display for Marlin users. I attempted to implement this as part of CuraEngine but found that without multipass gcode processing outputting time estimates as a percentage of total is not possible. As such a post processing script seems a good approach to implementing this.


Ghostkeeper:
I agree that it's better to write this as post-processing script.

Maybe it would be good to merge this with the `DisplayRemainingTimeOnLCD` as option?

You could display both remaining time and percentage completed. There's plenty of space for both, on the displays that I know of.
You've copied this code from that script, so I wouldn't think it's very hard to merge the two.

Bostwickenator:
I thought it made sense to separate concerns as much as possible here but I could collapse the two into one file and option control the two different behaviors. I'll take the name suggestion too. Thanks for the feedback

Bostwickenator:
Addressed all the comments. I've tried to test all the edge cases I can think of, tiny models, large models, missing lines in the output but I'll appreciate any additional suggestions.

fieldOfView:
You effectively removed the current DisplayRemainingTimeOnLCD script, which means that people that currently use that script will have to instead add your script. Not just once, but after each time they load an older project.

Two ways to get around that:
* Include an upgrade step that migrates any DisplayRemainingTimeOnLCD script to DisplayProgressOnLCD, or
* Keep the DisplayRemainingTimeOnLCD.py name, even if that filename no longer fully covers the functionality; it is not use-facing anyway.

If you want to go the former way, see this commit which does something similar for PauseAtHeight scripts:
https://github.com/Ultimaker/Cura/pull/7201/commits/321693d41023941f8fc5889ff42fbffbe0cb3fed

Bostwickenator:
I do not know how to test this migration script change. But this is roughly what it should look like. Could use `getattr` on line 126 but without being able to test I'm hesitant to try anything not already in the file.

Bostwickenator:
Just checking in do you need any additional changes from me to push this along. Thanks for the guidance thus far.

Ghostkeeper:
No, sorry, it's taking a bit longer due to other tasks we have to do. We have many pull requests also that have been waiting for a longer time.

It would be good to implement this in time for 4.7. That makes it a lot easier to merge due to the version upgrade. We'll try to get it done in time.

plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
```diff
@@ -0,0 +1,131 @@
...
+                                    # calculate may slightly exceed 100, as that is not valid we cap the value here
+                                    output = min(percentage + previous_layer_end_percentage, 100)
+                                    
+                                    # Now insert the sanitized percentage into the GCODE
+                                    lines.insert(percentage_line_index, "M73 P{}".format(output))
+
+                                previous_layer_end_percentage = layer_end_percentage
+
+                # Join up the lines for this layer again and store them in the data array
+                data[layer_index] =  "\n".join(lines)
```
konskarm, line 130:
```suggestion
                data[layer_index] = "\n".join(lines)
```

plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
```diff
@@ -0,0 +1,131 @@
...
+                        total_time = self.getTimeValue(line)
+                        line_index = lines.index(line)
+
+                        if (output_time):
+                            self.outputTime(lines, line_index, total_time)
+                        if (output_percentage):
+                            # Emit 0 percent to sure Marlin knows we are overriding the completion percentage
+                            lines.insert(line_index, "M73 P0")
+
+
```
konskarm, line 81:
```suggestion
```

plugins/VersionUpgrade/VersionUpgrade462to47/VersionUpgrade462to47.py
```diff
@@ -120,6 +120,17 @@ def upgradeStack(self, serialized: str, filename: str) -> Tuple[List[str], List[
                     if "redo_layers" in script_parser["PauseAtHeight"]:
                         script_parser["PauseAtHeight"]["redo_layer"] = str(int(script_parser["PauseAtHeight"]["redo_layers"]) > 0)
                         del script_parser["PauseAtHeight"]["redo_layers"]  # Has been renamed to without the S.
+
+                # Migrate DisplayCompleteOnLCD to DisplayProgressOnLCD
+                if script_id == "DisplayPercentCompleteOnLCD":
```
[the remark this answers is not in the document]
Bostwickenator: Good catch thanks!

Bostwickenator:
Thanks @konskarm. I've accepted your lint fixes and that great catch you made. Appreciate the thorougher review and you taking the time to add suggestions.

Issue #7665, opened 2020-05-03
[4.6.1] Loading file via command line parameter not relative to working directory


sk8nfool:
**Application version**
4.6.1 and possibly other earlier versions

**Platform**
Ubuntu 18.04

**Printer**
N/A

**Reproduction steps**
1. Start with '.../Ultimaker_Cura-4.6.1.AppImage Stand_Link.stl" where ... is path to Ultimaker file

**Screenshot(s)**
Note the "Invalid File" error at bottom of screen
![image](https://user-images.githubusercontent.com/12006794/80920033-35989380-8d3b-11ea-8aaa-3f5c37a53ffc.png)

However just reopening the file works. Note that Cura seems to understand the current directory correctly but not on startup.
![image](https://user-images.githubusercontent.com/12006794/80920085-87d9b480-8d3b-11ea-9326-80b0a76cb2d7.png)

**Actual results**
See above

**Expected results**
file to open

**Project file**
N/A

**Log file**
Logs (cura.log, stdout.log and stderr.log) don't mention open failure or file name

**Additional information**
It appears that the appimage is messing with Cura's idea of the current directory. This has been a chronic problem but not annoying enough to really be called a "bug" .... but it is a bug.

smartavionics:
Then why not do .../Ultimaker_Cura-4.6.1.AppImage ~/Stand_Link.stl to make the filename path absolute?

smartavionics:
Sorry, that assumes the stl is in the home directory, but the point is, use an absolute path name rather than a relative one for the stl file.

smartavionics:
Personally, I wouldn't call it a bug unless the Cura docs actually state that it can open files specified on the command line relative to the launching directory.

sk8nfool:
I understand that documentation sets the ground rules BUT an unqualified
pathname lives in a context and that context needs to be considered unless
overridden. I believe that Linux documentation states that un-rooted
pathnames are relative to the current working directory. I also believe
that Windows documentation says the same thing. So I believe that this
situation is a bug. This is expecially true because it appears that Cura
assumes all un-rooted paths start at the home folder. This simply isn't
true.

Now, all that being said, I don't know what the API for AppImages is nor do
I know what the Windows API looks like.

smartavionics:
> I believe that Linux documentation states that un-rooted
pathnames are relative to the current working directory. I also believe
that Windows documentation says the same thing. So I believe that this
situation is a bug.

[...]

Issue #6194, opened 2019-08-12
Cannot sign-in to Ultimaker Cloud in Cura 4.2.1 from Ubuntu 19.04


darkernoise:
**Application version**

4.2.1
**Platform**

Ubuntu 19.04 64bit

**Printer**

an Ender 3 profile was selected

**Reproduction steps**

Open the official Cura AppImage file in Ubuntu 19.04, running standard Gnome 3 Desktop.
Click the "Sign In" button to the right of the "Marketplace" button
When the dialog box opens, hover over the "Sign In" button and notice that it changes color
Click the button

**Actual results**

Nothing happens

**Expected results**

Sign-in panel should have displayed allowing selection of Ultimaker account.

**Additional information**

https://i.imgur.com/D9O5YgL.png
[CE3_square.zip](https://github.com/Ultimaker/Cura/files/3492681/CE3_square.zip)
[cura_log_clean.zip](https://github.com/Ultimaker/Cura/files/3492723/cura_log_clean.zip)

ChrisTerBeke:
I don't see anything in the logs that point to an error. The authorization callback server is started, and one second later the application is told to shut down (cleanly). Not sure though what is telling Cura to shut down.

darkernoise:
Hello ChrisTerBeke.  I agree. I don't see any issue in the log either.  Though, the issue is that neither of the button presses for "Sign In" Or "Create Account" function.  It is as if the events for those buttons are not registered.  I have tested the same thing in Windows and it works there.  This appears to be an issue with the AppImage on Ubuntu only.

ChrisTerBeke:
I'm running the AppImage on Ubuntu (18.04) myself and it works. Maybe it's something in only the very latest Ubuntu version that's different?

darkernoise:
A little Google-fu led me to this: https://unix.stackexchange.com/questions/417160/mouse-clicks-do-not-work-correctly-in-qt-applications

I am running a Dell XPS 15 with a touchscreen and it sounds like there is an issue with QT support of touchscreens.  I tested by touch-tapping the "Sign In" and "Create Account" buttons and sure enough, they don't work via tapping either. I will do some more research.

darkernoise:
I have followed the threads I could find around QT clicks not registering.  Here is my xinput list:

xinput --list
⎡ Virtual core pointer                    	id=2	[master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer              	id=4	[slave  pointer  (2)]
[...]

PR #7729, opened 2020-05-10
When running from AppImage, chdir to os.environ["OWD"] if it exists and is a directory.

smartavionics:
The AppImage runtime should set OWD to the original current working directory.

Solution for https://github.com/Ultimaker/Cura/issues/7665.

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -118,7 +118,10 @@ if Platform.isWindows() and hasattr(sys, "frozen"):
 # otherwise it doesn't work. This is a workaround on Linux that before we call QDesktopServices.openUrl(), we
 # switch to a directory where the user has the ownership.
 if Platform.isLinux() and hasattr(sys, "frozen"):
-    os.chdir(os.path.expanduser("~"))
+    if "OWD" in os.environ and os.path.isdir(os.environ["OWD"]):
+        os.chdir(os.environ["OWD"])
+    else:
+        os.chdir(os.path.expanduser("~"))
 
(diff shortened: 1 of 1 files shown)


cura_app.py
```diff
@@ -118,7 +118,10 @@ def get_cura_dir_path():
 # otherwise it doesn't work. This is a workaround on Linux that before we call QDesktopServices.openUrl(), we
 # switch to a directory where the user has the ownership.
 if Platform.isLinux() and hasattr(sys, "frozen"):
-    os.chdir(os.path.expanduser("~"))
+    if "OWD" in os.environ and os.path.isdir(os.environ["OWD"]):
```
Ghostkeeper, line 121: Since `OWD` should always be present when running via AppImage, I expect this will re-introduce https://github.com/Ultimaker/Cura/issues/6194 for everyone.

smartavionics: If the current directory is not writeable by the user running Cura then that could cause a problem if being able to create a file in the current directory is needed to avoid #6194. Testing whether it is possible to create a file in the current directory is obviously possible but is it worth the hassle? If you want to close this PR unused, I don't care as I only put it together as someone was banging on about not being able to open files with relative path names from the command line.

Ghostkeeper: Yes, I agree, it's not really worth the hassle there.

Sorry, I'll close this PR then. It's more important that people can click links rather than being able to provide relative paths in the command line arguments.

PR #7735, opened 2020-05-11
fix: remove disallowed areas on ender 3

pmaidens:
The Ender 3's bed is properly set to 235 x 235. However, the `machine_disallowed_areas` does not allow users to use the entire area. Strangely, the disallowed areas are also only set on the y axis.

The Ender 3 is capable of printing in the entire 235 x 235 area, so in my opinion, having disallowed areas does not make sense.

This change removes the disallowed areas, allowing users to choose for themselves how closely to the edge of the bed they would like to place models.

### DIFF
--- a/resources/definitions/creality_ender3.def.json
+++ b/resources/definitions/creality_ender3.def.json
@@ -12,11 +12,6 @@
         "machine_width": { "default_value": 235 },
         "machine_depth": { "default_value": 235 },
         "machine_height": { "default_value": 250 },
-        "machine_disallowed_areas": {
-            "default_value": [
-              [[-117.5, 117.5], [-117.5, 108], [117.5, 108], [117.5, 117.5]],
-              [[-117.5, -108], [-117.5, -117.5], [117.5, -117.5], [117.5, -108]]
-        ]},
         "machine_head_with_fans_polygon": { "default_value": [
                 [-26, 34],
                 [-26, -32],


64bittuning:
I'd have to look to find the exact source, but the Ender 3 can come without a magnetic bed. The disallowed areas were inputted to prevent users from putting material in the areas that the binder clips reside.

IMHO, the profile should be left alone. If you have a magnetic bed, use the Ender3 Pro config.

pmaidens:
That is a good point! That makes a lot of sense why it was added in the first place. It has been so long since I used clips that I totally forgot.

However, I think that it is a bit of a stretch to ask users to select a different printer if they have a non-clip-on bed.

I know I am only a sample size of 1, but personally it never occurred to me to select the Pro to fix the bed size. That is probably due to the fact that there are some pretty significant differences between the Ender 3 Pro and the Ender 3. Selecting a different printer without intimate knowledge of how that would affect things would make me super reluctant to do so.

Personally, I don't think it is a stretch for users to know that they have an obstruction on the bed, and they should not print in the obstructed area. The flip-side is asking people to know that a different printer is properly configured to work as theirs, or asking them to be familiar enough with the code to go in and modify it themselves. It took me several hours to figure out how to remove the disallowed areas, and I am a full time programmer.

Ghostkeeper:
We don't need half solutions. Cura supports proper solutions for printer add-ons like magnetic clips. We can have it ask the user whether they have clips or a magnetic bed, and add/remove the disallowed area as necessary.

For an example of how this works, try adding an Ultimaker Original printer.

This solution however I won't accept. I can't form an opinion about this since I have no Ender 3, but these disallowed areas were added fairly recently after a lengthy discussion. People generally seem to agree with the disallowed areas being necessary.

64bittuning:
@pmaidens alot of people don't grasp gcode, or the simple concept (to us) of X versus Y axis travel. A new user who knows that the machine was provided with these clips, expects that the machine won't hit them. The fine folks making this software were able to accommodate the newest of us by adding the disallowed area as a safety feature.

@ghostkeeper I've been looking for a new project. I'll start looking into adding a menu option that gets pulled in through the FDMprinter

PR #7752, opened 2020-05-13
Update geeetech_A10.def.json

keleticsaba:
Height and dept value was mixed.
XY values also corrected to factory defaults.


Ghostkeeper:
Eh, but they do report 220x220, not 235x235?

keleticsaba:
235 is the real size.220 is propably some safety with 15mm "not printable" area.But with some updated marlin (with the included a10 config file), the plate is centered, and the whole area can be printed. Sometime that "extra" space can be useful.So, vanila marlin2 is a must with that printer.The factory 1.1.8 is garbage, and ther is no offset in menu, to correct that.Ghostkeeper <notifications@github.com> ezt írta (2020. máj. 15. 11:11):
Eh, but they do report 220x220, not 235x235?

Ghostkeeper:
We would normally try to target the printer's own firmware rather than an alternative firmware that people flash themselves. Flashing your own firmware is quite an expert-level operation. I'd expect that if they do that, they can also find the field in Cura to adjust their build plate size to accompany the change.

For other people, the extra 15mm of safety sounds like a good idea.

keleticsaba:
You can be right.But the y-z axis size was mixed up anyway.Ghostkeeper <notifications@github.com> ezt írta (2020. máj. 18. 14:07):
We would normally try to target the printer's own firmware rather than an alternative firmware that people flash themselves. Flashing your own firmware is quite an expert-level operation. I'd expect that if they do that, they can also find the field in Cura to adjust their build plate size to accompany the change.
For other people, the extra 15mm of safety sounds like a good idea.

konskarm:
I resolved the merge conflict (not sure why Github mentions that the entire file was added). I agree with @Ghostkeeper that the size should be kept according to the official specifications (220x220x260). The size can be changed manually in Cura if needed through the machine settings. As such I am reverting it to the original size and accepting the PR.

PR #7763, opened 2020-05-15
Cura 7290 manual account sync

ninovanhooff:
Part of Cura 7290 has already been approved and merged.
This PR fixes some problems and implements:
- date formatting mm/dd/YYYY
- manual sync action only available after re-opening the popup
- Cloud Printer syncing should adhere to the contract that it should never get stuck in SYNCING state

Note:

I refactored the CloudApiClient and ToolpathUploader to use HttpRequestManager. 
Functional changes to those files:
- addition of request timeouts
- addition of json content-type and accept headers to json requests
If you see any other functional changes to those files, you should report them as part of your review.


cura/API/Account.py
```diff
@@ -114,14 +118,15 @@ def setSyncState(self, service_name: str, state: int) -> None:
...
+            self._setManualSyncEnabled(True)
         else:
             self._sync_state = SyncState.SUCCESS
 
         if self._sync_state != prev_state:
             self.syncStateChanged.emit(self._sync_state)
 
             if self._sync_state == SyncState.SUCCESS:
-                self._last_sync_str = datetime.now().strftime("%d/%m/%Y %H:%M")
+                self._last_sync_str = datetime.now().strftime("%m/%d/%Y %H:%M")
```
nallath, line 129: Uh? Why? Only America uses this. I think it's way better to have it Day/month/year. As far as I know, this is also how we handle other places where we show dates (eg; "your print is completed by" in the cloud / connect)

cura/API/Account.py
```diff
@@ -200,20 +225,21 @@ def userProfile(self) -> Optional[Dict[str, Optional[str]]]:
...
 
+    @pyqtProperty(bool, notify=manualSyncEnabledChanged)
+    def manualSyncEnabled(self) -> bool:
+        return self._manual_sync_enabled
+
     @pyqtSlot()
-    def sync(self) -> None:
-        """Signals all sync services to start syncing
+    @pyqtSlot(bool)
+    def sync(self, user_initiated=False):
```
nallath, line 234:
```suggestion
    def sync(self, user_initiated: bool = False) -> None:
```

cura/API/Account.py
```diff
@@ -200,20 +225,21 @@ def userProfile(self) -> Optional[Dict[str, Optional[str]]]:
...
+        self._sync()
 
-        if self._update_timer.isActive():
-            self._update_timer.stop()
-        elif self._sync_state == SyncState.SYNCING:
-            Logger.warning("Starting a new sync while previous sync was not completed\n{}", str(self._sync_services))
-
-        self.syncRequested.emit()
+    @pyqtSlot()
+    def popupClosed(self):
```
nallath, line 241:
```suggestion
    def popupClosed(self) -> None:
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -118,7 +119,7 @@ def _onGetRemoteClustersFinished(self, clusters: List[CloudClusterResponse]) ->
         self._syncing = False
         self._account.setSyncState(self.SYNC_SERVICE_NAME, SyncState.SUCCESS)
 
-    def _onGetRemoteClusterFailed(self):
+    def _onGetRemoteClusterFailed(self, reply: QNetworkReply, error: QNetworkReply.NetworkError):
```
nallath, line 122:
```suggestion
    def _onGetRemoteClusterFailed(self, reply: QNetworkReply, error: QNetworkReply.NetworkError) -> None:
```

cura/API/Account.py
ninovanhooff: As per design: https://jira.ultimaker.com/browse/CURA-7290?focusedCommentId=103361&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-103361

Issue #3345, opened 2018-02-18
Can't set two extruders


mark-hahn:
**Application Version**
3.2.1

**Platform**
Ubuntu 16.0.4 unity

**Steps to Reproduce**
set up machine settings for flashforge creator dual

**Actual Results**
The drop-down for number of extruders only has one extruder with no other option

Here are my machine settings.  It shows one extruder and I cannot select two.
![screenshot from 2018-02-09 17-48-27](https://user-images.githubusercontent.com/811455/36347095-8f8a5742-1403-11e8-819b-72b30c0d550b.png)

Kinglyboiz:
which printer are you using? I am using Tevo Tarantula with dual extruder and have the same issue. I am able to select the customer FDM printer and chose a second extruder, but I'm creating another issue where I cannot print with the second extruder. Cura seems to lock up. 

Running 3.2.1 on MAC OS

ChrisTerBeke:
That's because the printer definition for the Tevo Tarantual has no extruders defined, meaning it will fall back the default: 1 FDM extruder. They (or the person who's maintaining the Cura profiles for Tevo) should update the definition (or you can make a PR to fix it). We usually don't maintain 3rd party machines as we can't test with them.

Details: https://github.com/Ultimaker/Cura/blob/master/resources/definitions/tevo_tarantula.def.json.

Ghostkeeper:
To prevent confusion, I've now hidden this drop-down for printers where no more than 1 extruder is defined: https://github.com/Ultimaker/Cura/commit/aae8a31f00cf9db2b6190a43dbf1658db64311cf

Ghostkeeper:
We've been assuming that if you are technical enough to modify your printer to add additional extruders that it normally doesn't have, you'd also be technical enough to modify the JSON files of your printer definition in Cura's installation.

That's not always the case of course. But you could use Custom FDM Printer as an alternative.

Festivejelly:
This isnt very user friendly. Also assuming that someone knows their way around a printer should also know their way around JSON and how the formatting works is a bit ignorant tbh. I feel you should be able to add this through the UI.

paalex:
For those that decide to tweak Cura on their own, you can find the printer configuration file here (macOS):
```/Contents/Resources/resources/definitions/tevo_tarantula.def.json``` inside ```Ultimaker Cura``` app package

Replace ```tevo_tarantula.def.json``` with your printer's config filename.

Edit this part to add additional extruders:

[...]

PR #7769, opened 2020-05-16
add 2nd extruder to Tevo Tarantula printers

paalex:
addressing #3345 issue

### DIFF
--- a/resources/definitions/tevo_tarantula.def.json
+++ b/resources/definitions/tevo_tarantula.def.json
@@ -11,7 +11,8 @@
         "platform": "prusai3_platform.3mf",
         "machine_extruder_trains":
         {
-            "0": "tevo_tarantula_extruder_0"
+            "0": "tevo_tarantula_extruder_0",
+            "1": "tevo_tarantula_extruder_1"
         }
     },
 
--- a/resources/definitions/tevo_tarantula_pro.def.json
+++ b/resources/definitions/tevo_tarantula_pro.def.json
@@ -11,7 +11,8 @@
         "has_materials": true,
         "machine_extruder_trains":
         {
-            "0": "tevo_tarantula_pro_extruder_0"
+            "0": "tevo_tarantula_pro_extruder_0",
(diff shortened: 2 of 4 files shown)


paalex:
@TheAssassin @FN5969

TheAssassin:
Is there a reason you ping me?

paalex:
@TheAssassin you are marked as the author for the Tevo Tarantula config file ```"author": "TheAssassin"```
If it's not the case, or it's not relevant, my apologies

TheAssassin:
@paalex ah, right, that can well be.

I got the printer nearly 3 years ago, I don't know if there's even an "official" two-printer-head kit or something like that. I never really liked the printer, and right now it's even out of order as I don't have time to fix its extruder.
I don't think I can contribute a lot to this PR. Please feel free to take any changes you want to the config file.

paalex:
@TheAssassin got it, thanks.
I still use and the Tarantula and even got a 2nd extruder for it
I tested these config files with latest stable version of Cura and it loads well and works as expected. Hopefully it would get merged

konskarm:
When adding a second extruder, I would expect to see at least some distance from the first extruder e.g. "machine_nozzle_offset_x": { "default_value": x.x }. Currently I see that both in the tevo_tarantula and the tevo_tarantula_pro extruders 0 and 1 are identical. Are you sure this is correct?

konskarm:
> When adding a second extruder, I would expect to see at least some distance from the first extruder e.g. "machine_nozzle_offset_x": { "default_value": x.x }. Currently I see that both in the tevo_tarantula and the tevo_tarantula_pro extruders 0 and 1 are identical. @paalex are you sure this is correct?
> 
> Also, the second extruder is an upgrade, right? If we accept this PR, it means that the the second extruder will exist for all Tevo Tarantula users, which will cause confusion if they have only one extruder.

Alright, I see that the default number of extruders is 1 and it can be changed in the Print Settings window, similar to the Nozzle offsets, so this PR works fine.

PR #7783, opened 2020-05-18
Adds files for: Layer One Atom 3 and Atom 3 lite

Daniel-Kurth:
Adds definition, extruder, variant and mesh files for the Atom 3 and Atom 3 lite printers.


Ghostkeeper:
The automated tests are failing for a multitude of reasons on this:
* The platform meshes are too big.
* Ineffective `default_value` overrides due to a `value` being present.
* The extruder definitions don't use the printer definition's ID, but some human-readable name.
* Some of these files seem to have incorrect syntax? Though maybe that is just about the setting names starting with an underscore.
* And lastly, you are adding new settings that are not in fdmprinter in order to adjust warning values per profile. This is currently failing the tests. We've been discussing this a bit and we can't guarantee that this will stay working since we'll probably not think of testing it with every release, but we've decided to accept this change. The alternative would slow down the settings system for everyone, not just the Atom users, so we're not likely to implement an alternative ourselves and we'll allow this change in. However we do require you to fix the tests to account for this case then: The test for the instance container needs to look at the definitions in the correct definition container, rather than look at the definitions in fdmprinter.

Also please use 4 spaces rather than tabs for indentation. This is our code style guideline: https://github.com/Ultimaker/Meta/blob/master/general/generic_code_conventions.md

ninovanhooff:
No response after 20 days. Provisionally closing

PR #7785, opened 2020-05-18
CURA-7436_cloud_printer_remove_warning

ninovanhooff:
This implementation might introduce a somewhat redundant variable, since there are already 2 
Qt roles for indicating printer connection type. Interested to learn about reviewer's opinion.

Also this implemenation is written in cura core rather than a custom removal message set from 
a networkingplugin. Since the cloud printer concept is already available in Cura I thought 
this to be OK.


resources/qml/Preferences/MachinesPage.qml
```diff
@@ -136,6 +136,11 @@ UM.ManagementPage
         {
             id: confirmDialog
             object: base.currentItem && base.currentItem.name ? base.currentItem.name : ""
+            text: base.currentItem && base.currentItem.hasCloudConnection?
```
nallath, line 139: This does mean that _all_ cloud printers will have this behavior, which is not quite correct. If someone else makes a plugin that adds a cloud printer in the old way, this message will then also show up (which is not correct, since the printer will not be re-added)

PR #7790, opened 2020-05-19
Upgrade the ironing inset based on the selected ironing pattern 

konskarm:
Previously, the ironing_inset was changed in the profiles to reflect
the changes made in the CuraEngine. This inset though was not taking
into account the ironing_pattern (in the CuraEngine, the inset is
compensated differently depending on the pattern).

This PR fixes that in the version upgrader and the printer
definitions in order to maintain the same behavior for the old
profiles.

Merge along with https://github.com/Ultimaker/CuraEngine/pull/1254

CURA-7318

### DIFF
--- a/cura/BugReportHandler.py
+++ b/cura/BugReportHandler.py
@@ -1,0 +1,102 @@
+
+import os
+import os.path
+import platform
+from collections import OrderedDict
+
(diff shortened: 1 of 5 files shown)


Ghostkeeper:
What's the BugReportHandler doing there?

PR #7804, opened 2020-05-22
Make ActionButtons render.

lonetech:
The background implicitly got a zero width, making the buttons unusable
and near invisible.
This appears to be this bug:
https://community.ultimaker.com/topic/28928-cura-41-stuck-at-welcome-screen-buttons-are-missing/

I am uncertain if the particular implicitWidth I set is perfectly
correct, but usable buttons beat unusable ones.

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -230,5 +230,8 @@ if ApplicationMetadata.CuraDebugMode:
     ssl_conf.setPeerVerifyMode(QSslSocket.VerifyNone)
     QSslConfiguration.setDefaultConfiguration(ssl_conf)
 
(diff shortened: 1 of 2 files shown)


lonetech:
A simpler approach would seem to force the theme that Cura was written for, which corrected more issues (preview crash, button placement, perhaps more). I tested this by adding:

    os.environ["QT_QUICK_CONTROLS_STYLE"] = "material"

near the end of cura_app.py, before the CuraApplication() call. This workaround was noted in [issue #5488](https://github.com/Ultimaker/Cura/issues/5488#issuecomment-588560291).

lonetech:
Closing this as duplicate of https://github.com/Ultimaker/Cura/pull/7139

PR #7816, opened 2020-05-24
Add a repeat option to better support testing out a range of values.

viesturz:
Hi, I got fed up creating a bunch of incremental rules to test a range of values.
And added a repeat thing.

This is my first change in Cura, so not really aware to your style preferences.
Would be happy to clean up any issues.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
+++ b/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
@@ -11,6 +11,7 @@
 # Modified by Jaime van Kessel (Ultimaker), j.vankessel@ultimaker.com to make it work for 15.10 / 2.x
 # Modified by Ruben Dulek (Ultimaker), r.dulek@ultimaker.com, to debug.
 # Modified by Wes Hanney, https://github.com/novamxd, Retract Length + Speed, Clean up
+# Modified by Viesturs Zarins, https://github.com/viesturz, Repeat
 
 # history / changelog:
 # V3.0.1:   TweakAtZ-state default 1 (i.e. the plugin works without any TweakAtZ comment)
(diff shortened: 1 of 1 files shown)


ninovanhooff:
Code quality style looks good. Do you think there will be enough demand for this functionality, @nallath ?

Ghostkeeper:
I'm not sure if this is worth it, really. I see the value in it for people that create their own profiles, like for new materials or printers that Cura doesn't support. This is really quite expert though. And those people will also understand that if you make changes at different heights their test will be invalidated if the heated bed is activated, because the material will be warmer the closer it comes to that heated bed. You're better off cloning your test object and printing them in one-at-a-time mode next to each other for a tuning test, unless temperature is not a factor at all (i.e. to measure printing time it's fine).

The cost of this is complexity. Additional features confuse new users, scare them away. At a glance clicking the "repeat" option doubles the settings and it's hard to see what each setting ends up with in the end. It's mixing up absolute setting values with deltas with the previous repeat. Is there a way in which we can make this a bit simpler to understand for the user? It's hard to see how we can make it easier on the UX.

Ghostkeeper:
Sorry, we've decided to reject the feature in this form, because:
* It adds complexity to the user experience. Every setting now having a "repeat" option isn't immediately clear to even expert users.
* Creating experiments to tune profiles is not seen as the main use case of this script. Tuning profiles would increase the scope of this script enormously, making it untenable to maintain.
* For the purpose of tuning profiles, use case is limited due to the different environment higher up in the model vs. close to the build plate.
* It's not clear to the user how the repeating change is applied. Does it scale with the initial change or is it absolute?

These disadvantages outweigh the advantage of being able to test a range of values, which would indeed be useful for testing profiles. Sorry.

PR #7843, opened 2020-05-29
Adding note about unexpected behavior caused by USB Printing plugin t…

JamesHagerman:
…o the Printers section of the Preferences.

This PR is intended to help address #5207 where the USB Printing plugin can cause unexpected behavior in third-party software due to how it enumerates USB serial devices while looking for available printers.

I've added two clickable links. One to @fieldOfView's comment on #5207 describing the work around and another that opens the Marketplace panel itself.

Imgur

### DIFF
--- a/resources/qml/Preferences/MachinesPage.qml
+++ b/resources/qml/Preferences/MachinesPage.qml
@@ -113,6 +113,23 @@ UM.ManagementPage
                     }
                 }
             }
+
+            Text
+            {
+                textFormat: Text.RichText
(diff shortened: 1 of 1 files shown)


fieldOfView:
This message shows regardless of whether the USB Printing plugin is enabled, and regardless of what printers are added, even if none of the printers even have a USB port. I don't think this is the proper solution.

nallath:
I agree with @fieldOfView. Showing the message in *all* cases is not a good work around. 

A better solution would be to just add a preference field that simply disables the auto detect feature for USB printing plugin.

fieldOfView:
In my opinion, it would be best to disable the USB Printing plugin by default, and add a MachineAction to all printers that support USB printing to optionally enable the USB Printing plugin. This MachineAction could have a warning about it interfering with other serial devices etc.

victornpb:
The actual problem is trying to brute force all serial ports, trying to open then and throw data at them without any regard to what they are until you find something that you recognize as a printer. 

This could cause serious issues, I wonder who thought this was a good idea, this can cause data loss, property damage and even harm someone.

Ghostkeeper:
I agree that this is not a solution. Text that always warns you is quickly ignored; and indeed annoys people and looks messy. We'd prefer a solution such as suggested here: https://github.com/Ultimaker/Cura/issues/5207

Issue #7757, opened 2020-05-14
Only are availlables an ultimaker printers for Cura 4.6.1


Kikicandela:
During and after install the Cura 4.6.1 only it's available the ultimaker printers only, the rest of manufacturer are missing.

![photo5904517582101262872](https://user-images.githubusercontent.com/65359282/81960418-3b328b00-9611-11ea-921f-8d159ffe7b2c.jpg)

Kikicandela:
I answer myself, the printer are availables, the problem is that a vertical scroll bar is missing

Issue #7634, opened 2020-04-30
trying to add a non-network printer in 4.6.1, only Ultimaker BV printers listed 


VOCAPTAIN:
following guidance from your ultimaker support page on how to add a custom printer to Ultimaker Cura

**Application version**
(The version of the application this issue occurs with.)
4.6.1 and 4.6.0
**Platform**
(Information about the operating system the issue occurs on. Include at least the operating system and maybe GPU.)
windows 10 pro 64, Intel i32310M
**Printer**
(Which printer was selected in Cura?)
n/a
**Reproduction steps**

Following guidance from your ultimaker support page on how to add a custom printer to Ultimaker Cura

**Screenshot(s)**
(Image showing the problem, perhaps before/after images.) 
![thumbnail_IMG_3779](https://user-images.githubusercontent.com/64589354/80707816-83668f00-8ae2-11ea-8b02-44d6d59b5aa3.png)

**Actual results**
(What happens after the above steps have been followed.)
![Screenshot (8)](https://user-images.githubusercontent.com/64589354/80708042-e5bf8f80-8ae2-11ea-80e5-a4767173b693.png)

**Expected results**
(What should happen after the above steps have been followed.)

**Project file**
(For slicing bugs, provide a project which clearly shows the bug, by going to File->Save. For big files you may need to use WeTransfer or similar file sharing sites.)

**Log file**
(See https://github.com/Ultimaker/Cura#logging-issues to find the log file to upload, or copy a relevant snippet from it.)

**Additional information**
(Extra information relevant to the issue.)

fieldOfView:
You can scroll the list of printers. There's a scrollbar that is hidden until you use it.

fieldOfView:
Duplicate of https://github.com/Ultimaker/Cura/issues/7614

VOCAPTAIN:
thanks silly me the hidden scroll bar got me :-)

PR #7845, opened 2020-05-29
Add local printer improvements

fieldOfView:
This PR improves the UX of adding a local printer:
* Makes the scrollbar while selecting local printer visible so it doesn't look like only Ultimaker printers are supported
* Makes it more obvious there may be more items in the list by showing at least one more item (Custom)
* Makes the name field look less like a search field by placing it - along with some metadata of the selected printer - to the right of the list
* Makes the list respond to resizing the window

image

Showing the scrollbar is the most important change, and partly based on work by @diegopradogesto here: https://github.com/Ultimaker/Cura/pull/6325

This PR also in part reimplements https://github.com/Ultimaker/Cura/pull/4498

The "Add a networked printer" section remains untouched

### DIFF
--- a/resources/qml/qmldir
+++ b/resources/qml/qmldir
@@ -35,6 +35,7 @@ RadioButton         1.0 RadioButton.qml
 Scrollable          1.0 Scrollable.qml
 TabButton           1.0 TabButton.qml
 TextField           1.0 TextField.qml
+ScrollView          1.0 ScrollView.qml
 
 
 # Cura/MachineSettings
--- a/resources/qml/WelcomePages/AddLocalPrinterScrollView.qml
+++ b/resources/qml/WelcomePages/AddLocalPrinterScrollView.qml
@@ -29,8 +29,6 @@ Item
         "Custom": -1
     }
 
-    property int maxItemCountAtOnce: 11  // show at max 11 items at once, otherwise you need to scroll.
-
     // User-editable printer name
     property alias printerName: printerNameTextField.text
     property alias isPrinterNameValid: printerNameTextField.acceptableInput
--- a/resources/qml/WelcomePages/AddLocalPrinterScrollView.qml
+++ b/resources/qml/WelcomePages/AddLocalPrinterScrollView.qml
@@ -29,7 +29,7 @@ Item
         "Custom": -1
     }
 
-    property int maxItemCountAtOnce: 10  // show at max 10 items at once, otherwise you need to scroll.
+    property int maxItemCountAtOnce: 11  // show at max 11 items at once, otherwise you need to scroll.
 
     // User-editable printer name
     property alias printerName: printerNameTextField.text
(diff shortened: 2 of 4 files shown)


fieldOfView:
See #7757, #7634, #7614

fieldOfView:
Looks like CI is drunk (or Uranium is broken)

nallath:
Uranium doesn't seem to be broken and the tests succeed for me locally. Weeeird.

fieldOfView:
CI must still be drunk.

nallath:
I've kicked it. Maybe a re-run solves the issue.

Yian-Lai:
Hi, 
Great improvement on showing the scrollbar! 
Here are some thoughts for other points:

_Makes it more obvious there may be more items in the list by showing at least one more item (Custom)_
If there are more printers from Ultimaker then Custom won't show. I think making the scrollbar visible would solve this problem already.

_Makes the name field look less like a search field by placing it - along with some metadata of the selected printer - to the right of the list_
I personally think it is inconsistent with the Networked printer list which has one column, so I can't see a strong need for this change.

fieldOfView:
>> Makes the name field look less like a search field by placing it - along with some metadata of the selected printer - to the right of the list
> I personally think it is inconsistent with the Networked printer list which has one column, so I can't see a strong need for this change.

I have seen a number of users who just typed "Creailty Ender 5" in the printer name field. and then wonder why Cura saves .ufp files instead of .gcode files (spolier: they added an Ultimaker S5 named "Creality Ender 5"). The proposed change makes it a lot clearer what printer gets added.

fieldOfView:
Showing the Author of the definition has the additional benefit of at least hinting at the user that the profile has been provided by someone other than Ultimaker. Perhaps this could be a good place for a disclaimer that Ultimaker has not tested these profiles.

Ghostkeeper:
I'd hate showing disclaimers in places that some users have no workaround for though. But yes, it's good to show the author there. Maybe it should say "profile author"?

fieldOfView:
Good news: CI has sobered up!

fieldOfView:
Regarding the horizontal split in the dialog, I would argue that splitting up the wide listview brings the scrollbar closer to the actual list it is scrolling, which narrows the mental divide between the two. The scrollbar has a small enough handle that it is still going to be easy to overlook if it is separated from the list by that much whitespace.

image

nallath:
CURA-7516

resources/qml/qmldir
```diff
@@ -35,6 +35,7 @@ RadioButton         1.0 RadioButton.qml
 Scrollable          1.0 Scrollable.qml
 TabButton           1.0 TabButton.qml
 TextField           1.0 TextField.qml
+ScrollView          1.0 ScrollView.qml
```
nallath, line 38: Super nitpicky, shouldn't it be 1.1? You actually changed it in AddLocalPrinterScrollView.qml to use cura 1.1 instead of 1.0

fieldOfView: To be honest, I don't know what the version of the Cura namespace should be. The change to Cura 1.1 is a vestigial change from @diegopradogesto's work. 

I'm not even sure if registering ScrollView in qmldir is necessary; neither Scrollable.qml nor TabButton.qml exist in the current codebase, and their replacements are not registered but still work.

Personally, given this nitpick, I would change AddLocalPrinterScrollView.qml back to using Cura 1.0. Let me know what you prefer.

nallath: Oh, well. I don't think it's very actively used by us. The idea is to have versioning in your qml. 

I thought it was an intentional change, hence the nitpick. If it doesn't serve a real purpose, it doesn't really matter which one is used.

PR #7857, opened 2020-06-02
Support serial connections with 76800 baud rate 

victornpb:
Some printers are using 76800baud, and cura is not able to detect them.

Why?

Running a 16Mhz ATmega2560 at the more common _115200 baud_ is not recommended as the error rate of 3% would be above what the spec tolerates, because of the clock mismatch 16Mhz is not divisible by 300.
So the next more appropriate rate is `76800 baud` with a much cleaner transmission with an error rate of about 0.2%.

76800 is also referred as 38400 double rate, and some ESP-12 also use 76.8K baud.

Sources:
https://www.robotroom.com/Asynchronous-Serial-Communication-2.html
https://trolsoft.ru/en/uart-calc

### DIFF
--- a/plugins/USBPrinting/AutoDetectBaudJob.py
+++ b/plugins/USBPrinting/AutoDetectBaudJob.py
@@ -18,7 +18,7 @@ class AutoDetectBaudJob(Job):
     def __init__(self, serial_port: int) -> None:
         super().__init__()
         self._serial_port = serial_port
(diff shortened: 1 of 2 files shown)


Ghostkeeper:
I think we simply never had any printer that uses that baud rate.

The tests were failing due to a problem in master at the time you branched off. We can ignore that. (It's fixed in master in the meanwhile.)

PR #7859, opened 2020-06-02
Ignore right click in nonscene components

konskarm:
Right clicking anywhere in Cura opens the popup menu which handles the scene. This can be confusing when accidentally right clicking on a component unrelated to the scene, such as the Prepare-Preview-Monitor buttons or the Print settings panel.

This PR fixes that by converting most elements that reside on top of the scene into UnclickableItems. These items have their own mouse area to consume clicks and mouse wheels, thus not inheriting the mouse area of the scene.

Requires: https://github.com/Ultimaker/Uranium/pull/614

Completed during Turbo Testing and Tooling.


resources/qml/Toolbar.qml
```diff
@@ -4,10 +4,10 @@
 import QtQuick 2.2
 import QtQuick.Controls 2.3
 
-import UM 1.2 as UM
+import UM 1.4 as UM
 import Cura 1.0 as Cura
 
-Item
+UM.UnclickableItem
```
nallath, line 10: Is this change really needed? The item is already filled with objects that "steal" the clicks, right (so no need to add another one).

konskarm: I would expect the same, but as it seams, the objects steal only the left clicks. If you right click on any of the button you can move the scene, or open the popup context menu
image

PR #7866, opened 2020-06-03
Change the panelBorder rectangle anchor to the bottom for lower toolbar buttons.

smartavionics:
This makes the panel grow upwards into clear space rather than downwards which is a problem
when the Cura window height is restricted as the bottom of the panel become inaccessible.

Screenshot_2020-06-03_08-10-27

As discussed in https://community.ultimaker.com/topic/32840-a-problem-in-the-fitting-on-the-screen/.

The panel anchor point for the top toolbar buttons (Move, Scale, Rotate) is not changed.

### DIFF
--- a/resources/qml/Toolbar.qml
+++ b/resources/qml/Toolbar.qml
@@ -96,6 +96,8 @@ Item
                             {
                                 UM.Controller.setActiveTool(model.id);
                             }
+
+                            base.state = (index < toolsModel.count/2) ? "anchorAtTop" : "anchorAtBottom";
                         }
                     }
                 }
@@ -219,4 +221,33 @@ Item
 
         visible: toolHint.text != ""
     }
+
+    states: [
+        State {
+            name: "anchorAtTop"
+
+            AnchorChanges {
+                target: panelBorder
+                anchors.top: base.top
+                anchors.bottom: undefined
+            }
+            PropertyChanges {
+                target: panelBorder
+                anchors.topMargin: base.activeY
+            }
+        },
+        State {
+            name: "anchorAtBottom"
+
+            AnchorChanges {
+                target: panelBorder
+                anchors.top: undefined
+                anchors.bottom: base.top
+            }
+            PropertyChanges {
+                target: panelBorder
+                anchors.bottomMargin: -(base.activeY + UM.Theme.getSize("button").height)
+            }
+        }
+    ]
 }
--- a/resources/qml/Toolbar.qml
+++ b/resources/qml/Toolbar.qml
@@ -246,7 +246,7 @@ Item
             }
             PropertyChanges {
                 target: panelBorder
-                anchors.bottomMargin: -(base.activeY + UM.Theme.getSize("button").height)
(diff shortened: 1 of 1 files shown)


fieldOfView:
I don't think this is the proper solution to the problem. With more tools being added to the toolpanel via plugins, it will become confusing why tools on the bottom throw their panels to the top. With some more QML-side logic, it is possible to make the toolpanels "hug" the bottom of the window as needed, ie only if they would otherwise extend over the bottom of the window.

smartavionics:
Thanks for the comment. However,  I don't agree with your suggestion that people will be confused by the panels going up rather than down. 

This PR fixes the issue, if you think it can be done better using a different strategy, write the code!

fieldOfView:
The alignment works a bit differently and less confusing than I originally thought, but your solution has other issues with small windows:
image

smartavionics:
I agree that is still a problem and this PR is not perfect but in mitigation I would say:

1 - although ugly, the panel is at least usable in the sense that the button at the bottom is accessible.

2 - do people really use windows that shallow?

3 - the PR could be modified to move the anchor point even lower when the panel is tall?

Ghostkeeper:
It would be perfect if it's not just completely at the top or completely at the bottom (based on how far along the list of buttons it is) but adjusts itself to fit. Might be a bit difficult though because the borders are in completely separate parts of the QML element tree.

Yes, plenty of people use windows that small. It's not just a problem of resolution but more that they have small screens with high DPI, so the pixel scaling is turned up in their OS.

smartavionics:
OK, I'll tinker with it some more. Stay tuned...

smartavionics:
Try it now.

When panel is not tall, it anchors at the bottom...

Screenshot_2020-06-04_11-27-28

When the top of the panel would be above the first tool button the anchor is shifted...

Screenshot_2020-06-04_11-27-08

resources/qml/Toolbar.qml
```diff
@@ -246,7 +246,7 @@ Item
             }
             PropertyChanges {
                 target: panelBorder
-                anchors.bottomMargin: -(base.activeY + UM.Theme.getSize("button").height)
+                anchors.bottomMargin: ((base.activeY + UM.Theme.getSize("button").height) > panelBorder.height) ? -(base.activeY + UM.Theme.getSize("button").height) : -panelBorder.height
```
fieldOfView, line 249: Suggestion for readability:
```
                anchors.bottomMargin: 
                {
                    if (panelBorder.height < base.activeY + UM.Theme.getSize("button").height)
                    {
                        return base.activeY + UM.Theme.getSize("button").height
                    }
                    return panelBorder.height
                }
```

smartavionics: I think your suggested code is missing some - signs but I get the idea. Will do.

I didn't realise that in qml you can use return within curly brackets like that.

smartavionics: Now beautiful!

fieldOfView: Gorgeous!

nallath:
For our reference: CURA-7512

PR #7869, opened 2020-06-04
Create SV01.def

Joyce-lujunxu:
Add SOVOL 3D printer


fieldOfView:
The path to the machine definition should be 'resources/definitions'.

PR #7871, opened 2020-06-04
UM3NetworkPrinting: remove duplicate updates for model properties

Gobbel2000:
time_total, time_elapsed and owner were updated twice for the ClusterPrintJobStatus Model. 

(Also 1000th fork)


nallath:
Whoops. I have no idea why that snuck in. It's even been there for two years... Anyhow, thanks for fixing it!

For our internal tracker: CURA-7486

PR #7873, opened 2020-06-04
CURA-7438 Handle the case when a cloud printer is removed from the account

konskarm:
When a cloud printer is no longer received from mycloud, it is safe to
assume that the printer is removed from the account. In such a case,
the new function _devicesRemovedFromAccount(..) will retrieve all the
devices that are to be removed (i.e. they are added in Cura but they
do not exist in the clusters received from mycloud) and will remove
their output device. In addition the function will generate a message
that mentions all the devices that were no longer linked to the
account.

In order to handle the removed printers, the CloudOutputDeviceManager
is enriched with the dictionary self._um_cloud_printers that contains all
the currently existing cloud-enabled devices that were previously added
in Cura. In addition, the metadata field "linked_to_account" is added to all
the cloud-enabled Cura devices to distinguish the ones actually linked 
to the current account. With these additions, the cloud devices that are 
removed from the account will be automagically re-linked to it once the 
devices are once again retrieved from mycloud for the specific account.

Finally, the PR changes the connectionStatusMessage of MachineSelector.qml
into a function so that the popup tooltip text is properly updated every
time the user hovers over the "Printer disconnected" icon and reflect the 
correct message over why the printer seems offline.

Requires: https://github.com/Ultimaker/Uranium/pull/617 in order to properly display links in 
the popup messages.

CURA-7438


plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -31,6 +32,7 @@ class CloudOutputDeviceManager:
     META_CLUSTER_ID = "um_cloud_cluster_id"
     META_NETWORK_KEY = "um_network_key"
     SYNC_SERVICE_NAME = "CloudOutputDeviceManager"
+    META_LINKED_TO_ACCOUNT = "linked_to_account"
```
nallath, line 35: This should be renamed, since it's not clear what this means outside of this context. I think something like "added_by_um_account" or "um_linked_to_account" is better.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -206,19 +228,85 @@ def _onDevicesDiscovered(self, clusters: List[CloudClusterResponse]) -> None:
...
+
+        # Do not report device ids which have been previously marked as non-linked to the account
+        ignored_device_ids = set()
+        for device_id in removed_device_ids:
+            if not parseBool(self._um_cloud_printers[device_id].getMetaDataEntry(self.META_LINKED_TO_ACCOUNT, "true")):
+                ignored_device_ids.add(device_id)
+        # Keep the reported_device_ids list in a class variable, so that the message button actions can access it and
+        # take the necessary steps to fulfill their purpose.
+        self.reported_device_ids = removed_device_ids - ignored_device_ids
+        if len(self.reported_device_ids) == 0:
```
nallath, line 266:
```suggestion
        if not self.reported_device_ids:
```

Don't use len == 0. It's much slower.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -206,19 +228,85 @@ def _onDevicesDiscovered(self, clusters: List[CloudClusterResponse]) -> None:
...
+                device_names
+        )
+        removed_printers_message.setText(message_text)
+
+        # Remove the output device from the printers
+        for device_id in removed_device_ids:
+            device = self._um_cloud_printers.get(device_id, None)  # type: Optional[GlobalStack]
+            if not device:
+                continue
+            output_device_manager = CuraApplication.getInstance().getOutputDeviceManager()
```
nallath, line 299: Move this outside of the for loop.

resources/qml/PrinterSelector/MachineSelector.qml
```diff
@@ -139,12 +146,13 @@ Cura.ExpandablePopup
         {
             id: connectionStatusTooltipHoverArea
             anchors.fill: parent
-            hoverEnabled: connectionStatusMessage !== ""
+            hoverEnabled: getConnectionStatusMessage() !== ""
```
nallath, line 149: Why did you change this into a function? It should work when it's a property, right?

konskarm: Nope, it doesn't get re-evaluated when hovering over it. I had to change into a function, or else it kept the old message in the tooltip, until the machine was changed.

PR #7883, opened 2020-06-05
Set padding to 0 to remove vertical shift on Linux builds.

smartavionics:
This problem has been seen before but I don't think it was ever fixed and one of the users of my builds reported it again so here's a fix for the bad valign issue that you can get in the machine settings dialog (at least on Linux).

Before...

Screenshot_2020-06-05_07-56-36

After...

Screenshot_2020-06-05_15-35-27

FYI I'm using Qt 5.13.2.

### DIFF
--- a/resources/qml/MachineSettings/NumericTextFieldWithUnit.qml
+++ b/resources/qml/MachineSettings/NumericTextFieldWithUnit.qml
@@ -77,6 +77,7 @@ UM.TooltipArea
         anchors.left: fieldLabel.right
         anchors.leftMargin: UM.Theme.getSize("default_margin").width
         verticalAlignment: Text.AlignVCenter
+        padding: 0
         width: numericTextFieldWithUnit.controlWidth
         height: numericTextFieldWithUnit.controlHeight
 


Ghostkeeper:
Need someone to test this on Windows or MacOS. I fear that this might be dependent on the specific desktop widgets that the user is using, so it's just Gnome that has this problem, so this change could misalign the text for those.

smartavionics:
Yes, good point. I'll test it on Windows and MacOS here.

smartavionics:
Looks fine on a Windows 10 laptop.

smartavionics:
Mind you, I am using Qt 5.13.2 for all my builds.

smartavionics:
And no problem on an ancient MacBook Pro.

I can't test this on anything else so I'll leave it to you guys to check if it's ok with Qt 5.10.

nallath:
Since I don't actually have the problem as originally described, could you check if it still works after you add:
`leftPadding: UM.Theme.getSize("narrow_margin").width`
In combination with your change this should ensure that everything works.

smartavionics:
Just tried that on Linux (Qt 5.13.2) and it didn't cause any problems.

nallath:
Cool, will add that as well then, since the text does look nicer if that's there.

PR #7887, opened 2020-06-06
Add silly "ride the nozzle" simulation mode.

smartavionics:
When running the simulation, press Ctrl (Cmd on MacOS, I think) to "ride the nozzle".

Normal simulation...

Screenshot_2020-06-06_11-10-59

Ride that nozzle!...

Screenshot_2020-06-06_11-11-42

Just don't make yourself seasick!

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -16,6 +16,9 @@ from UM.View.GL.OpenGL import OpenGL
 
 from cura.Settings.ExtruderManager import ExtruderManager
 
+from PyQt5 import QtCore, QtWidgets
+
+from copy import deepcopy
 
 import os.path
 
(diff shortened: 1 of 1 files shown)


Asterchades:
Would recommend either changing the hotkey or reconfiguring some of the other key combos. As it stands, [ctrl]+up or down is used to change the layer in the preview, which automatically invokes this view at the same time. Makes a fine-grained examination of the preview extremely awkward.

smartavionics:
> As it stands, [ctrl]+up or down is used to change the layer in the preview

Ah, I didn't know that. I may change the hotkey but I'm not seriously expecting UM to use this PR, it's a joke.

Asterchades:
It is, however, present in your Master builds - which I suspect have a sizeable install base, and will continue to do so at least until libArachne gets incorporated.

Gotta have some fun, though. A little Easter egg like that is a welcome touch to any program, kind-of like the first-person view in Prison Architect (which lacks any real 3D assets to speak of).

smartavionics:
Yeah, well, I'll do something about it.

Ghostkeeper:
People have seen this. I wouldn't be surprised if it does get merged as an Easter egg, as long as it doesn't interfere with anything. Probably won't show up in the change log though...

Ghostkeeper:
Maybe it should play [this sound effect](https://youtu.be/cclgtCvzhW8?t=40) when going down for the second object in one-at-a-time mode :stuck_out_tongue:

PR #7894, opened 2020-06-07
Bugfix for double ";" when using function "putvalue"

louiswouters:
Fixed double ";" when used on an existing line with a comment.
Optimized the construction of the new G-code line by using list.append and join
Updated description (outdated)

### DIFF
--- a/plugins/PostProcessingPlugin/Script.py
+++ b/plugins/PostProcessingPlugin/Script.py
@@ -149,38 +149,38 @@ class Script:
         :return: A line of g-code with the desired parameters filled in.
         """
 
-        def putValue(self, line: str = "", **kwargs) -> str:
-            #Separate the code and comment
+        #Separate the g-code and comment
+        if ";" in line:
+            comment = line[line.find(";"):]
+            line = line[:line.find(";")]
+        else:
             comment = ""
-            if ";" in line:
-                comment = line[line.find(";"):]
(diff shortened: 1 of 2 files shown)


plugins/PostProcessingPlugin/Script.py
```diff
@@ -141,52 +141,46 @@ def putValue(self, line: str = "", **kwargs) -> str:
...
-            result += key + str(value)
-
-        #Put the comment back in.
-        if comment != "":
-            if result != "":
-                result += " "
-            result += ";" + comment
-
-        return result
+        def putValue(self, line: str = "", **kwargs) -> str:
```
konskarm, line 152: putValue(..) is already defined in line 136. Please remove line 152 and unindent the rest of the function.

Ghostkeeper:
We already have a post-processing script to display progress in the LCD. If there's something you want to improve on that, please improve the existing one. It's also good to build separate features in separate pull requests, lest they block each other.

louiswouters:
O, it appended this pull request... I tried to create a separate pull request but it looked like nothing had happened.

konskarm:
@louiswouters, as @Ghostkeeper mentioned, it seems you are doing two thing in this PR, fixing the bug mentioned in the title and also adding a ShowProgress post-processing script (which we already have as DisplayProgressOnLCD). Please split these two into separate PR's and if needed modify the DisplayProgressOnLCD instead of adding an extra script.

PR #7896, opened 2020-06-08
Upload to cloud printer in a single chunk

Ghostkeeper:
Previously the print job was sent with multiple PUT requests. This was necessary since the request was made with the Python Requests library, which freezes the interface while it's running. By breaking the upload up in chunks, it would at least periodically keep updating the interface (4 times per second or so, depending on your internet speed and Ultimaker's). Later on, this was replaced by a `QNetworkRequest` which doesn't freeze the interface in a refactor, but it was still broken up in chunks.

Recently that was again refactored to use Uranium's `HttpRequestManager`, which under the covers also uses `QNetworkRequest`. This also doesn't freeze while uploading, which is good. However for some reason the second chunk would always give a `QNetworkReply::ProtocolUnknownError`, which is a bit of a catch-all error for any HTTP status codes not supported by Qt. I was not able to figure out what was really going wrong there, but it has something to do with the content-range header and making multiple requests.

Instead of fixing that, I've removed the chunking. This simplifies the code a lot and doesn't have any implications for the user, since it still doesn't freeze the interface while the network request is ongoing. It should be slightly faster to upload, and reduce load on the server too.

However the two disadvantages of this are that:
* If the original bug is in the `HttpRequestManager`, it will still be there. If it was a bug here in the `ToolPathUploader` I probably deleted that bug in this PR.
* If internet connection is spotty but restores quickly, the previous code would only try to resend the last chunk. This code will start again from the beginning.

To test, upload UFP files larger than 256kB. This was the chunk size originally.

Fixes CURA-7488.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/ToolPathUploader.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/ToolPathUploader.py
@@ -20,9 +20,6 @@ class ToolPathUploader:
     # The HTTP codes that should trigger a retry.
     RETRY_HTTP_CODES = {500, 502, 503, 504}
 
-    # The amount of bytes to send per request
-    BYTES_PER_REQUEST = 256 * 1024
-
     def __init__(self, http: HttpRequestManager, print_job: CloudPrintJobResponse, data: bytes,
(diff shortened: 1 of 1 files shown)


ChrisTerBeke:
This seems to fix the issue indeed, I was able to upload a Benchy to a cloud printer again when running this branch from source.

PR #7900, opened 2020-06-09
CURA-7437 More robust cluster group size

ninovanhooff:
This PR adds a metadata data field to a machine for (cloud) cluster size. 
This means this info is also available when either Cura or the printer are offline


plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, api_client: CloudApiClient, cluster: CloudClusterResponse, pa
             b"name": cluster.friendly_name.encode() if cluster.friendly_name else b"",
             b"firmware_version": cluster.host_version.encode() if cluster.host_version else b"",
             b"printer_type": cluster.printer_type.encode() if cluster.printer_type else b"",
-            b"cluster_size": b"1"  # cloud devices are always clusters of at least one
+            b"cluster_size": str(cluster.printer_count).encode() if cluster.printer_count else b"1"
```
nallath, line 77: So if print_count is 0, the cluster size is set to 1?

PR #7901, opened 2020-06-09
add SOVOL printer

Joyce-lujunxu:


nallath:
For our reference: CURA-7511

PR #7904, opened 2020-06-09
CURA-7454: Add "remove printers" button in removed printers from account message

konskarm:
This PR adds an additional button in the "removed printers from account" message that allows the user to completely remove these printers from Cura. 
Removed printers message
For that to happen, the user is asked through a question box to confirm the action.
image

To achieve the requirements of this ticket, the following had to be done in Cura:

- Add another "AddPrintersPagesModel" that is exactly the same as the other one with the exception that it does not display a "Cancel" button
- Allow Cura to set an empty active machine, if all machines are removed (when pressing the "Remove printers" button)
- Gracefully handle in the front-end the case where the last active machine is removed 
- Popup the "Add printers" wizard without the option to be cancelled when the last printer is removed

**Note 1**: Tests are failing because it requires the Uranium PR https://github.com/Ultimaker/Uranium/pull/621.

**Note 2**: Before reviewing and component testing this PR, make sure that the tickets that follow
 are merged in the right order.

1. CURA-7438:
    - https://github.com/Ultimaker/Cura/pull/7873 (Cura)
    - https://github.com/Ultimaker/Uranium/pull/617 (Uranium)
2. CURA-7455:
    - https://github.com/Ultimaker/Cura/pull/7880 (Cura)
3. CURA-7454

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -207,6 +207,7 @@ class CuraApplication(QtApplication):
         self._first_start_machine_actions_model = None
         self._welcome_pages_model = WelcomePagesModel(self, parent = self)
         self._add_printer_pages_model = AddPrinterPagesModel(self, parent = self)
+        self._add_printer_pages_model_without_cancel = AddPrinterPagesModel(self, parent = self)
         self._whats_new_pages_model = WhatsNewPagesModel(self, parent = self)
         self._text_manager = TextManager(parent = self)
 
@@ -647,7 +648,7 @@ class CuraApplication(QtApplication):
         return self._global_container_stack
 
     @override(Application)
-    def setGlobalContainerStack(self, stack: "GlobalStack") -> None:
+    def setGlobalContainerStack(self, stack: Optional[GlobalStack]) -> None:
         self._setLoadingHint(self._i18n_catalog.i18nc("@info:progress", "Initializing Active Machine..."))
         super().setGlobalContainerStack(stack)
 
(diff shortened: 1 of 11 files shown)


cura/CuraApplication.py
```diff
@@ -1940,6 +1947,11 @@ def shouldShowWelcomeDialog(self) -> bool:
         # Only show the complete flow if there is no printer yet.
         return self._machine_manager.activeMachine is None
 
+    @pyqtSlot(result = bool)
+    def shouldShowAddPrintersUncancellableDialog(self) -> bool:
```
nallath, line 1951: I don't really see a reason for this to be in the Cura application (or in python for that matter). Both the required properties are already exposed in QML, so why not make it part of that?

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -50,6 +51,7 @@ def __init__(self) -> None:
         self._account = CuraApplication.getInstance().getCuraAPI().account  # type: Account
         self._api = CloudApiClient(CuraApplication.getInstance(), on_error = lambda error: Logger.log("e", str(error)))
         self._account.loginStateChanged.connect(self._onLoginStateChanged)
+        self.removed_printers_message = None  # type: Optional[Message]
```
nallath, line 54: Is this intentionally public?

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDeviceManager.py
```diff
@@ -269,14 +276,14 @@ def _devicesRemovedFromAccount(self, removed_device_ids: Set[str]) -> None:
...
-        removed_printers_message = Message(
+        self.removed_printers_message = Message(
                 title = self.I18N_CATALOG.i18ncp(
                         "info:status",
                         "Cloud connection is not available for a printer",
                         "Cloud connection is not available for some printers",
                         len(self.reported_device_ids)
                 ),
-                lifetime = 0
+                lifetime = 30
```
nallath, line 286:
```suggestion

```
30 is the default, so I think it's better to leave it out altogether (my IDE feels the same way ;))

PR #7911, opened 2020-06-10
CURA-7479: Convert RowLayout into Row in ConfigurationMenu

konskarm:
Fixes the following problem:
image

CURA-7479

### DIFF
--- a/resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
+++ b/resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
@@ -38,7 +38,7 @@ Cura.ExpandablePopup
     headerItem: Item
     {
         // Horizontal list that shows the extruders and their materials
-        RowLayout
+        Row
         {
             anchors.fill: parent
             Repeater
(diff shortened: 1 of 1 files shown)


resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
```diff
@@ -38,16 +38,16 @@ Cura.ExpandablePopup
...
-        RowLayout
+        Row
         {
             anchors.fill: parent
             Repeater
             {
                 model: extrudersModel
                 delegate: Item
                 {
-                    Layout.fillWidth: true
```
ChrisTerBeke, line 49: Layout.preferredWidth: Math.round(parent.width / extrudersModel.count) would probably do the trick as well if you wanna keep the RowLayout component (which is better for re-sizable interfaces).

konskarm: Thanks for the suggestion Chris. I converted it back to RowLayout

PR #7917, opened 2020-06-11
Adjusted correct pause_method for FlyingBear Ghost 4S

adripo:
As per #7339, FlyingBear Ghost 4S uses command M25, instead of marlin M0 to pause printing.

I tried to use the new script for all printers, but the condition on the pause method was wrong.


konskarm:
As I do not own a FBG4S, I cannot validate whether the above change is needed, so I take your word for it.

Issue #7918, opened 2020-06-11
PauseAtHeight getting wrokg next XY position


adripo:
**Application version**
4.7(master)

**Platform**
Wondows

**Printer**
FlyingBear Ghost 4S

**Reproduction steps**
1. Set pauseAtHeight in Post Processing extension
2. Generate gcode
3. Open file and find wrong position after pause

**Additional information**
I find out that the wrong X and Y positions are the values set to max jerk M205, that is the first X and Y values found on the next layer.
I already fixed this by excluding lines that start with command M205.
If you have a better solution please edit my pull request.

GregValiant:
The fix was merged so I'll close this.

PR #7919, opened 2020-06-11
Fixed next XY return position after pause

adripo:
Fixed bug #7918

Excluded lines that start with M205, command used for setting Jerk and not to move.
https://marlinfw.org/docs/gcode/M205.html


fieldOfView:
I think that instead of "blacklisting" `M205,` it is better to "whitelist" `G0` and `G1` (and possibly `G2` and `G3`). There may be other gcode flavors that have other commands that include an X and Y argument.

adripo:
I changed as you suggested

ninovanhooff:
Thanks!

PR #7920, opened 2020-06-11
CURA-6989 / CURA-6747 Add setting: Minimum slope steepness for stair-stepping.

rburema:
See also https://github.com/Ultimaker/CuraEngine/pull/1271


rburema:
Accidentally the last commit!

Rather than play time-machine with GH I'll just leave it here, but _also_ commit it to master. Shouldn't do any harm. Just be aware that the 2nd commit isn't really part of the PR :-)

PR #7930, opened 2020-06-13
Add BeamUp L machine profile files

beamup3d:
definition, extruder, mesh and quality files added


nallath:
You seem to be adding a bunch of profiles with an outdated version number. This is what makes our automated tests complain about it. Could you fix this?

beamup3d:
What should be the version numbers of quality files? I think def and extruder versions are correct?

nallath:
We're currently at version nr 15

Ghostkeeper:
Note also that if you created the profile when the setting version was 10, there will have been a number of differences in the settings list since you created the profile. Mostly deleted and renamed settings, but also stuff like the Maximum Deviation setting halving. Please test the profiles on a modern Cura version!

beamup3d:
Ok we made the updates. Thank you.

ninovanhooff:
I'm seeing these errors when after selecting the Beamup L. No printer mesh is loaded.

```
2020-06-23 17:06:53,274 - DEBUG - [Thread-3] UM.Mesh.MeshData.calculateNormalsFromVertices [548]: Calculating normals took 6.175041198730469e-05 seconds
/home/nvanhooff/PycharmProjects/Uranium/UM/Mesh/MeshData.py:540: RuntimeWarning: invalid value encountered in true_divide
  n[:, 0] /= l
/home/nvanhooff/PycharmProjects/Uranium/UM/Mesh/MeshData.py:541: RuntimeWarning: invalid value encountered in true_divide
  n[:, 1] /= l
/home/nvanhooff/PycharmProjects/Uranium/UM/Mesh/MeshData.py:542: RuntimeWarning: invalid value encountered in true_divide
  n[:, 2] /= l
```

beamup3d:
What should be the file type of mesh 3mf or stl?

Ghostkeeper:
Any file that Cura can open (without external plug-in) is allowed, so you may use either. We prefer 3MF somewhat because it generally has a smaller file size than STL. However if a texture is needed you'll need to use OBJ.

Ghostkeeper:
There were a few issues with the 3MF file's structure that Uranium wasn't handling well. I fixed that up, so the platform mesh loads correctly. However it appears that the platform mesh is not well aligned.

Screenshot from 2020-06-26 13-38-27

I'm guessing that whatever made the conversion to 3MF there didn't take into account the convention in STL to invert the Y and Z axes.

beamup3d:
Thanks. I rotated the mesh.

Ghostkeeper:
Due to a bug in our own code (the Uranium repository), your rotation wasn't applied. Took me a while to figure it out! In the end I had to [apply](https://github.com/Ultimaker/Cura/commit/922986a3f90fc8ffd8119eb58e3eeca546ac7ecd) the transformation to the actual mesh data instead of how your editor saved it. The result is visually the same as before though.

Thanks for supporting Cura!

beamup3d:
Thank you. If there is anything I can do please contact.

PR #7951, opened 2020-06-19
Fixed double ";" when using putValue on a line with a comment

louiswouters:
Refactored the code to use a join so that it doesn't need to check for adding spaces everytime.
We use a join at the end so that every element is separated by a space.

It also fixes the bug where it would grab the comment including ";" and then add another ";" when reinserting the comment

### DIFF
--- a/plugins/PostProcessingPlugin/Script.py
+++ b/plugins/PostProcessingPlugin/Script.py
@@ -141,52 +141,45 @@ class Script:
         All other keyword parameters are put in the result in g-code's format.
         For instance, if you put ``G=1`` in the parameters, it will output
         ``G1``. If you put ``G=1, X=100`` in the parameters, it will output
-        ``G1 X100``. The parameters G and M will always be put first. The
-        parameters T and S will be put second (or first if there is no G or M).
-        The rest of the parameters will be put in arbitrary order.
+        ``G1 X100``. The parameters will be added in order G M T S F X Y Z E.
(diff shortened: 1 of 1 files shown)


louiswouters:
Line 173-175 could possible be removed as well, since I don't think there are other parameters then G M T S F X Y Z E.

plugins/PostProcessingPlugin/Script.py
```diff
@@ -141,52 +141,45 @@ def putValue(self, line: str = "", **kwargs) -> str:
...
+        Any other parameters will be added in arbitrary order.
 
         :param line: The original g-code line that must be modified. If not
             provided, an entirely new g-code line will be produced.
         :return: A line of g-code with the desired parameters filled in.
         """
-
-        #Strip the comment.
-        comment = ""
+        # Strip the comment.
```
nallath, line 151: Uh, how does that change the behavior? As far as I can see, the changed code will do exactly the same.

louiswouters:
PutValue( "G92 X45 ; move", Y=12) would evaluate to G92 X45 Y12 ;; move. This fixes the double semicolon

ninovanhooff:
Assuming the code actually works, I'd say this is an improvement. The kwargs.pop() is easy to gloss over though, please put it on a separate line so it is more clear that the prioritized kwargs will not be iterated a second time in the other for loop

PR #7952, opened 2020-06-19
[WIP] Run CTest in parallel

Ghostkeeper:
Should make the tests a bit faster.


Ghostkeeper:
Okay, well, it's a little bit faster... Like 20 seconds faster on the 3.5 minute run. Oh well, it's something :tada:

PR #7953, opened 2020-06-19
[VERY WIP] Experiment monitor in browser

ChrisTerBeke:
This experimental PR swaps out the monitor tab UI for the web-based Digital Factory UI. Took me about an hour to do including some research on QtWebEngine. With a bit more investment I think this could get quite awesome.

**To do (at least)**
- [ ] Include QtWebEngine in Cura builds.
- [ ] Remove unneeded QML (most of it) and Python (still some needed for other Cura behaviour) code from the network plugin.
- [x] Prevent navigating to 'external' URLs (like support pages) that prevent the user from coming back to the actual Digital Factory web UI. Maybe open these in the default browser still?
- [ ] Deal with locally connected printers (show local web UI instead?).
- [ ] Improve upon hacky way to hide UI elements that don't really fit well in the context of Cura (e.g. DF header bar).

image


ninovanhooff:
Noice

Ghostkeeper:
Note that we've talked about including QtWebEngine in the Cura architecture meetings but rejected this previously. We'd prefer to refrain from including that as a dependency.
This approach also prevents LAN printers from being used on a factory floor that is separated from the internet, which is currently also a requirement.

ChrisTerBeke:
> Note that we've talked about including QtWebEngine in the Cura architecture meetings but rejected this previously. We'd prefer to refrain from including that as a dependency.
> This approach also prevents LAN printers from being used on a factory floor that is separated from the internet, which is currently also a requirement.

Whether to include QtWebEngine is ultimately the decision of the Cura team, but I think there's many added benefits to it, this PR is just one of them.

Regarding the local LAN, this PR could easily be modified to show the local web UI on the printer as well for those cases, as is indicated in the description.

nallath:
I don't really have a specific issue with not adding the QtWebEngine. The one reason we didn't do it is because of build size, but back then we also had significantly less maintenance burden on the various connected components.

PR #7994, opened 2020-06-29
Initial layer Horizontal Expansion for AA0.4 and AA0.25 print core

Sarita-Ultimaker:
To compensate elephant foot -> initial layer horizontal expansion setting is modeified for engineering and AA0.25 core profile for UMS5 and for UM3  for AA0.4 and AA0.25 cores.


Ghostkeeper:
The tests are failing because the settings for the Ultimaker 3 are no longer the same as for the Ultimaker 3 extended. I'll fix that up.

PR #7995, opened 2020-06-29
Cura 7199 remove spaghetti infill

jellespijker:
Removed spaghetti infill.

see also https://github.com/Ultimaker/CuraEngine/pull/1280


plugins/VersionUpgrade/VersionUpgrade462to47/VersionUpgrade462to47.py
```diff
@@ -27,6 +36,11 @@ def upgradePreferences(self, serialized: str, filename: str) -> Tuple[List[str],
 
         # Update version number.
         parser["metadata"]["setting_version"] = "15"
+        
+        # Remove deleted settings from the visible settings list.
```
nallath, line 40: You should also remove them from the profiles.

jellespijker: Good catch, thx.

PR #7999, opened 2020-06-30
Value of skin_outline_count reverts to 0 when concentric top/bottom used

jellespijker:
Part of CURA-6684

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1510,8 +1510,7 @@
                 {
                     "label": "Extra Skin Wall Count",
                     "description": "Replaces the outermost part of the top/bottom pattern with a number of concentric lines. Using one or two lines improves roofs that start on infill material.",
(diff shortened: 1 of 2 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1510,11 +1510,11 @@
                 {
                     "label": "Extra Skin Wall Count",
                     "description": "Replaces the outermost part of the top/bottom pattern with a number of concentric lines. Using one or two lines improves roofs that start on infill material.",
-                    "default_value": 1,
+                    "default_value": "0 if top_bottom_pattern == 'concentric' else 1",
```
nallath, line 1513: Nope, sorry, you can't do that! default_value is what is used by the engine to get a value. Since it doesn't have a python interpreter, these must be "simple" values (eg not setting functions)

jellespijker: noted at changed

PR #8001, opened 2020-06-30
Add a print job name setting to general preferences

ninovanhooff:
Technical implementation done.

Todo: add more variables

CURA-5479

### DIFF
--- a/cura/UI/PrintInformation.py
+++ b/cura/UI/PrintInformation.py
@@ -450,10 +450,10 @@ class PrintInformation(QObject):
 
         if "{machine_name}" in template:
             global_container_stack = self._application.getGlobalContainerStack()
-            if not global_container_stack:
-                self._abbr_machine = ""
-                return
-            active_machine_type_name = global_container_stack.definition.getName()
+            active_machine_type_name = global_container_stack.definition.getName() \
+                if global_container_stack \
+                else "no_machine"
+
             active_machine_type_name = active_machine_type_name.replace(" ", "_")
             output = output.replace("{machine_name}", active_machine_type_name)
 
--- a/cura/UI/PrintInformation.py
+++ b/cura/UI/PrintInformation.py
@@ -254,8 +254,7 @@ class PrintInformation(QObject):
     def _onPreferencesChanged(self, preference: str) -> None:
         if preference == "cura/job_name_template":
             self._updateJobName()
-
-        if preference == "cura/material_settings":
+        elif preference == "cura/material_settings":
             for build_plate_number in range(self._multi_build_plate_model.maxBuildPlate + 1):
                 self._calculateInformation(build_plate_number)
 
--- a/cura/UI/PrintInformation.py
+++ b/cura/UI/PrintInformation.py
@@ -252,11 +252,12 @@ class PrintInformation(QObject):
         self.materialNamesChanged.emit()
 
     def _onPreferencesChanged(self, preference: str) -> None:
-        if preference != "cura/material_settings":
-            return
+        if preference == "cura/job_name_template":
+            self._updateJobName()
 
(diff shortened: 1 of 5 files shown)


fieldOfView:
Hah, two plugins in the Marketplace implementing this feature isn't enough eh?

nallath:
Uh, two? It's only the custom prefix plugin, right?

ninovanhooff:
Internally discussed: this PR has some benefits over plugins. Also: if there are multiple plugins covering somewhat equal functionality, it is at least some indication that we might want to include that functionality in the core because it is likely a good idea that benefits lots of users.

Also discussed: this PR implements basic variables. More variables will be added upon request (either internal or requests by the community)

cura/UI/PrintInformation.py
```diff
@@ -252,11 +252,12 @@ def _calculateInformation(self, build_plate_number: int) -> None:
...
 
     def _onPreferencesChanged(self, preference: str) -> None:
-        if preference != "cura/material_settings":
-            return
+        if preference == "cura/job_name_template":
+            self._updateJobName()
 
-        for build_plate_number in range(self._multi_build_plate_model.maxBuildPlate + 1):
-            self._calculateInformation(build_plate_number)
+        if preference == "cura/material_settings":
```
konskarm, line 258: Shouldn't these be an if-elseif to avoid checking both cases always?

tests/TestPrintInformation.py
```diff
@@ -19,7 +26,7 @@ def getPrintInformation(printer_name) -> PrintInformation:
     mocked_extruder_stack.material = mocked_material
 
     mock_application.getInstance = MagicMock(return_value = mock_application)
-    mocked_preferences.getValue = MagicMock(return_value = '{"omgzomg": {"spool_weight": 10, "spool_cost": 9}}')
+    mocked_preferences.getValue = MagicMock(side_effect=preferencesGetValue)
```
konskarm, line 29:
```suggestion
    mocked_preferences.getValue = MagicMock(side_effect = preferencesGetValue)
```

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -627,14 +627,25 @@ UM.PreferencesPage
...
+                    Label
+                    {
+                        id: jobNameTemplateLabel
+                        text: catalog.i18nc("@label","Print job template:")
+                    }
+
+                    TextField
+                    {
+                        id: jobNameTemplateTextField
+                        width: 200 * screenScaleFactor
```
konskarm, line 645: The width could be bigger to fit in the full initialy template (just a suggestion).

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -627,14 +627,25 @@ UM.PreferencesPage
             {
                 width: childrenRect.width
                 height: childrenRect.height
-                text: catalog.i18nc("@info:tooltip", "Should a prefix based on the printer name be added to the print job name automatically?")
+                text: catalog.i18nc("@info:tooltip. Note variable names themselves (ie. machine_name_short, project_name) should not be translated", "Variables: machine_name_short, machine_name, project_name")
 
-                CheckBox
+                Column
```
konskarm, line 632: Use Row instead of column, so that it looks prettier?

fieldOfView:
> Uh, two? It's only the custom prefix plugin, right?

There is also [GCode Filename Format](https://marketplace.ultimaker.com/app/cura/plugins/rgomezjnr/GcodeFilenameFormat) which tries to do the same thing.

fieldOfView:
I assume there will be an versionupgrade from the `cura/jobname_prefix` to the `cura/job_name_template` preference?

konskarm:
> I assume there will be an versionupgrade from the `cura/jobname_prefix` to the `cura/job_name_template` preference?

Good catch. We will add it.

PR #8053, opened 2020-07-10
Include additional machine defaults for Dagoma DiscoEasy / DiscoUltimate

Sophist-UK:
Values taken from [Dagoma Marlin source code](https://github.com/dagoma3d/Marlin-By-Dagoma/blob/Dagoma-Development/LinuxAddons/Dagoma/conf/serie/E200/Base). Same values are in Marlin for DiscoUltimate.

### DIFF
--- a/resources/definitions/dagoma_disco.def.json
+++ b/resources/definitions/dagoma_disco.def.json
@@ -70,9 +70,9 @@
         "machine_max_jerk_xy": { "value": 20 },
         "machine_max_jerk_z": { "value": 0.4 },
         "machine_max_jerk_e": { "value": 5 },
-		"machine_steps_per_mm_x": 	{ "default_value": 80 },
-		"machine_steps_per_mm_y": 	{ "default_value": 80 },
-		"machine_steps_per_mm_z": 	{ "default_value": 2560 },
-		"machine_steps_per_mm_e": 	{ "default_value": 98 }
+	"machine_steps_per_mm_x": 	{ "default_value": 80 },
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I created using examples from other CURA machine profiles and values from the Dagoma Marlin repo. However I note that some examples used `default_value` and some `value` which seems odd.

Can someone please enlighten me about whether these are correct and how you can find out which to use?

Ghostkeeper:
`value` overrides `default_value`. So if there is already a `value`, the `default_value` won't have any effect. The `value` can contain a Python expression, while the `default_value` cannot. In practice, always overriding `value` is the safest.

The purpose of `default_value` is so that CuraEngine can be provisioned with a value as well when ran from the command line, which is not an issue when running Cura normally. It's also slightly more efficient at start-up.

There is some documentation about this here: https://github.com/Ultimaker/Cura/blob/master/docs/profiles/setting_properties.md

Sophist-UK:
> Changes look good to me! Hopefully they'll result in more accurate printing time estimates for those printers.

They do - I entered them manually into Cura and I noticed that the time estimate was significantly more accurate.

Given the comment about `value` vs. `default_value` do we need to change any of these to the other type?

Ghostkeeper:
No I think those are fine. We also have automated checks to test if any of the setting overrides aren't working.

0r31:
Hi @Sophist-UK , goot idea, i will do the very same for Dagoma delta printers.

PR #8055, opened 2020-07-13
CURA-6711_support_structure_dropdown

ninovanhooff:


ninovanhooff:
> Regarding [617b070](https://github.com/Ultimaker/Cura/commit/617b07061dcaa0d624fc7066f2f3f19004a55f50):
> 
> What if someone is only using support meshes (ie: with `support_enable` off)?
> /sa #7328

Then the resulting value would be equal to the density used for the regular support structure, for example 15. Since the fallthrough `else` case has the same value as `support_enable and support_structure == 'normal'

PR #8060, opened 2020-07-13
Added Definitions for all SVT FDM printers

venkatkamesh:
New brand


venkatkamesh:
could someone elaborate the error caused
thank you

Ghostkeeper:
We have automated tests to check the files you're submitting. These are the tests that failed:

* `arjun300_platform.stl` is too big. Limit is 1MB. Please reduce the resolution to keep the installation size reasonable.
* `diytech400.obj` is too big too.
* `Arjun300_Duplication_Mode.def.json` failed to load. Perhaps broken references to other files? Syntax?
* `arjun300.def.json` failed to load as well.
* `arjun300.def.json` defines a `default_value` for the settings `switch_extruder_prime_speed`, `switch_extruder_retraction_amount` and `switch_extruder_retraction_speed` which is ineffective because those settings have a `value` that overrides them.
* All of these definitions have an `id` metadata which is not allowed (legacy).
* `arjun300_dm_extruder.def.json` refers to a machine `arjun300_dm ` which doesn't exist.
* `arjun300_dual_extruder_0.def.json` and `arjun300_dual_extruder_1.def.json` refer to a machine `arjun_dual` which doesn't exist.
* `diy220_extruder_0` refers to a definition file which doesn't list `diy220_extruder_0` as one of its extruders.
* There are settings in your profiles that don't exist in Cura. Although this error message doesn't elaborate which.
* There are syntax failures in your .cfg files. Perhaps something to do with the extra newline after `[metadata]`? Not sure about this one.
* `setting_version` is 13 instead of the most recent version (15). Please re-test the profiles in the most recent version or apply the version upgrade manually. The Maximum Deviation setting has changed in meaning since then.

venkatkamesh:
Thank you for listing the issues hope this helps our team dev to fix em,
our team wrote def and this is working fine with 4.6v of cura just by merging. 
will update your mentioned comment to team

at present diytech 200 is only listed and working def from our end source.

PR #8068, opened 2020-07-15
Add Makertech printer profiles

Makertech3D:
Profiles are for the Proforge original, Proforge 2S and Axis 3d printers. Profiles also have materials, a material pull request has also been issued. Thanks.


konskarm:
Hi @Makertech3D. Thank you for your PR. 

This is an enormous change. Currently Cura has ~3300 resource files. Adding 1200 extra files is an increase of ~36% for 5 printers. 
I do not see much of change between quality profiles of different Makertech printers (e.g. Generic ABS 0.2 Ultra Detailed profile for Axis is the same as the equivalent one in Proforge2s). Could you please take a look at how Creality printers are structured, and follow the same paradigm to reduce the amount of quality files? Basically, they use a base creality printer on which all qualities are attached to. Then, all the other creality printer definitions inheirt from that, thus eliminating the need for creating quality profiles for each printer.

Also, I have the following remarks:

- I see that the .def.json files have wrong spacing, which makes them difficult to read. E.g.

image

 - Naming conventions are not followed. E.g. quality profile axis_01_gabs_02_UD.inst.cfg should actually be axis1_0.2_ultra_detailed.inst.cfg. See ultimaker resources as examples.

CI/CD is hasn't run for your PR for some reason, but running it locally I get the following:
- ID field shouldn't exist in definition files (e.g. makertech_axis_extruder_0.def.json)
- "category" shouldn't exist as a field in the definitions metadata (e.g. makertech_proforge_2s.def.json)
- When the "value" field exists in a setting, then the setting shouldn't have a "default_value", since it is overriden by value anyway (e.g. "infill_pattern" and "skin_overlap" in makertech_axis.def.json)

konskarm:
Devs, see ticket CURA-7638

Makertech3D:
Thank you for taking a look at this, we'll see about slimming down the changes and get back to you.

konskarm:
Closing this since there hasn't been any recent activity. Please submit a new PR if you manage to address all the aforementioned issues.

PR #8077, opened 2020-07-17
Change "Save..." to "Save Project..." in file menu

mahtDFR:
Quick qml string change. People sometimes send the wrong 3mf on the forum/issue tracker and then there's wasted time asking for it again. Idea from @fieldOfView

### DIFF
--- a/resources/qml/Menus/FileMenu.qml
+++ b/resources/qml/Menus/FileMenu.qml
@@ -30,7 +30,7 @@ Menu
     {
         id: saveWorkspaceMenu
         shortcut: StandardKey.Save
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Note that this used to be called "Save Project...", before this change: https://github.com/Ultimaker/Cura/commit/e53e62ccab47e81aac180fa4c2cf3cf951f5524a

That change was a decision by our UXer at the time (Yi-an). The objective was that the workspace file format ("project file") would be seen as a document for an in-progress slice preparation project source. Similar to how Photoshop has their PSD files as the default for the "Save..." entry, or Blender has their blender projects.

mahtDFR:
There's an internal ticket about this here: CURA-7604

PR #8081, opened 2020-07-18
Add M73 R to progress post-processing script

ftk:
Set the remaining time using M73 R gcode for Marlin firmware that has options LCD_SET_PROGRESS_MANUALLY, USE_M73_REMAINING_TIME enabled

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
+++ b/plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
@@ -1,7 +1,7 @@
 # Cura PostProcessingPlugin
-# Author:   Mathias Lyngklip Kjeldgaard, Alexander Gee
+# Author:   Mathias Lyngklip Kjeldgaard, Alexander Gee, ftk
 # Date:     July 31, 2019
-# Modified: May 22, 2020
+# Modified: July 18, 2020
 
 # Description:  This plugin displays progress on the LCD. It can output the estimated time remaining and the completion percentage.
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
I like the idea with this change. But is there something we can improve with the disambiguation and clash of the three settings here?

There are three issues I can see with this:
* There is a setting called "Time Remaining (M73)" and one called "Percentage", which both place `M73` in the g-code, only one is a percentage and the other is a time remaining.
* The new setting mentions `M73` which will not say anything to a user who is not intimately familiar with g-code. Even most expert users would have to consult a manual to understand what your setting does.
* If the user enables both "Time Remaining (M73)" and "Percentage", only the "Percentage" actually gets displayed on the printer.

I'd like to suggest redesigning these settings a little bit to improve those issues (just a suggestion, see for yourself):
* Create a boolean setting to enable or disable the `M73` command, replacing the existing "Percentage" setting.
* When enabled, it'll output both the progress percentage and the remaining time in one command, such as `M73 P69 R1337`. We can assume that if a printer supports `M73`, it supports all of its parameters or at least ignores the ones it doesn't know, so being able to choose one or the other is not necessary.
* Rename this boolean setting to something like "Update Progress Bar" or something that the user can directly see on their printer. If you think this changes too much between printer firmware, we'd need to use something more cryptic such as "Notify Firmware" or "Report via M-code".

Would you agree?

Ghostkeeper:
You can make another pull request (from the same branch if you like) once the aforementioned issues have been fixed.

PR #8095, opened 2020-07-21
Temperature Tower - new Post Processing Plugin script

jaaaco:
Hi,
I was looking a way to print temperature tower in Cura and I found using ChangeAtZ plugin to manually enter each step a bit tiresome :)

I stumbled upon this unfinished work: https://github.com/asmaurya/Temperature-Tower-Plugin-for-Cura-4.- and decided to make it work.

I hope somebody will find it useful as I did. As this is my first contribution in this repo I'll be happy to incorporate all suggested fixes and changes.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/TemperatureTower.py
+++ b/plugins/PostProcessingPlugin/scripts/TemperatureTower.py
@@ -6,7 +6,6 @@ from typing import List
 from ..Script import Script
 from UM.Application import Application
 
-#  Performs a search-and-replace on all g-code.
 class TemperatureTower(Script):
     def getSettingDataString(self):
         return """{
--- a/plugins/PostProcessingPlugin/scripts/TemperatureTower.py
+++ b/plugins/PostProcessingPlugin/scripts/TemperatureTower.py
@@ -1,0 +1,103 @@
+# Created by Jakub Wietrzyk <jakub@wietrzyk.com> https://github.com/jaaaco
+# Based on https://github.com/asmaurya/Temperature-Tower-Plugin-for-Cura-4.-
+
+from typing import List
+
+from ..Script import Script
+from UM.Application import Application
+
+#  Performs a search-and-replace on all g-code.
+class TemperatureTower(Script):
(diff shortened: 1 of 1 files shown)


kabakakao:
I tried this Script today and it ist working like a charm. I would reale like to see that in the next release

looxonline:
This is neat!

rburema (changes requested):
Hi! Dev here. Sorry to keep you waiting. Occasionally our PR's have a habit of piling up.

I have some questions that I like to have answered.

Most importantly:
- What if there where already temperature commands in the gcode in those layers?
- Shouldn't you use M109 instead? (Or at the very least have that as an option?)
- Error handling, even for 'silly' cases like a very low amount (1 or 2) of layers. (Even if it's just a return and do nothing.)
- Some settings use 'value' while others' use 'default_value'. Is there a reason for this?

Other remarks I have:
- Take a look at what inheriting from the 'Script' class already offers you, for instance (or mostly, really), method(s like) 'putValue' (as used, for example, in the Pause-At-Height script).
- I don't think relying on the comment to provide the line-end is very clear. Sure, the script is very short _now_ but, well.
- What about tool switches? That also might've got temperature commands in the code?
- I know we're making open source and all, but it might be a good idea to inform the author of the code you've based this on you made this PR :-)

Future considerations:
- Fan speed.

nallath:
Friendly reminder for @jaaaco We still have some review questions open.

konskarm:
@jaaaco as an update, the [CalibrationShapes plugin](https://marketplace.ultimaker.com/app/cura/plugins/5axes/CalibrationShapes) includes a TempTower post-processing script. Can you check if that fulfills the same purpose as the script you are introducing in your PR? If it does, then maybe this PR is redundant.

ptrakk:
> * Shouldn't you use M109 instead? (Or at the very least have that as an option?)

I would like this too. every release i go to the file and change the script, it would be nice to see this in uptream.

nallath:
Closing this due to inactivity.

PR #8106, opened 2020-07-24
CURA-7559: Remove all references to Wall Compensation from the resources/ in arachne

konskarm:
LibArachne takes care of generating the walls and the Wall Compensation settings have no effect to it, so they should be removed.
CuraEngine PR: https://github.com/Ultimaker/CuraEngine/pull/1299

CURA-7559


resources/quality/cartesio/abs/cartesio_0.25_abs_high.inst.cfg
```diff
@@ -18,7 +18,6 @@ wall_thickness = 1
 top_bottom_thickness = 0.8
 wall_0_inset = -0.05
 fill_perimeter_gaps = nowhere
-travel_compensate_overlapping_walls_enabled =
```
Ghostkeeper, line 21: I wonder how these empty values ever got through. I'd even expect them to give a syntax error maybe.

PR #8118, opened 2020-07-28
CURA-7608: Ignore machine network metadata when reading a project file

konskarm:
CURA-7608

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -5,7 +5,7 @@ from configparser import ConfigParser
 import zipfile
 import os
 import json
-from typing import cast, Dict, List, Optional, Tuple, Any
+from typing import cast, Dict, List, Optional, Tuple, Any, Set
 
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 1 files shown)


plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -41,6 +41,18 @@
...
+_ignored_machine_network_metadata = {
+    "um_cloud_cluster_id",
+    "um_network_key",
+    "um_linked_to_account",
+    "host_guid",
+    "removal_warning",
+    "group_name",
+    "group_size",
+    "connection_type"
+}  # type: Set[str]
```
fieldOfView, line 53: This mirrors the `ignore_keys` here (but not quite):
https://github.com/Ultimaker/Cura/pull/8116/files
Wouldn't it be better to have a single list?

nallath: That shouldn't have been merged (and it has also been reverted). These keys _should_ not have ended up in a project file (but they should be stored normally). If we remove them by means of the version upgrader, they are removed from locally stored printers as well.
So the list is only there for the project loading.

PR #8142, opened 2020-08-03
CURA-7609: Pick any printer of the same type when opening a project file

konskarm:
The open project dialog will now allow the user to select any of the printers they have in Cura that are of the same type as the printer in the project file (if any). 
With that, the user will not have to create new machines just to open a project file. All there needs to be done now is to update one of the existing similar printers. 

CURA-7609

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -243,7 +243,7 @@ class ThreeMFWorkspaceReader(WorkspaceReader):
 
             definition_container_type = definition_container.get("type")
             if definition_container_type == "machine":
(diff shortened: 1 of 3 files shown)


plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -229,6 +227,7 @@ def preRead(self, file_name, show_dialog=True, *args, **kwargs):
         # Read definition containers
         #
         machine_definition_id = None
+        updatable_machines = []
```
nallath, line 230: I think you should rename this to updatable_machine_names (since the list doesn't actually contain machines). Also, it's missing typing.

nallath:
Closing in favor of https://github.com/Ultimaker/Cura/pull/8144

PR #8146, opened 2020-08-05
Allow scrolling through setting tooltips

Ghostkeeper:
This is something that will by default probably not make any difference for Cura, unless there are settings with extremely long descriptions and people try to read them on very small screens. I'm implementing this because the Settings Guide has an option to put the articles in the setting tooltips. That option is sometimes useless because the articles are just way too long to display on the screen.

I've made a change that allows the user to interact with the setting tooltips and scroll through the text, if it is too long to display on the screen. To allow this, I've also implemented that hovering over the tooltip makes the tooltip stay visible. You can hover over the tooltip while it is still visible (opacity > 0), so you've got the 0.1s duration of the fade animation to move the cursor from the setting to the tooltip. It's only a dozen pixels or so, so this turns out to be fairly easy.

A video of the action: http://dulek.net/work/setting_tooltip_scroll.webm

A few notes about the requirements of this implementation that I held to:
* It should not affect the design if the tooltip is smaller than the screen.
* It must not cause significant performance reduction, not even in theory. This is not so difficult since the tooltip is only created once, so it will now only have one extra MouseArea and one extra ScrollView to render. Clipping is not necessary since it goes off the screen anyway.

The only difference that you need to consider when deciding on whether to accept this change or not, is that the user may move the mouse on top of the tooltip accidentally, causing the tooltip to be in the way. From what I feel, it behaves sort of like how many applications work with expanding menus and it feels natural that the tooltip stays when you hover, so it is also clear that you must move the cursor to get the tooltip to go away.

A bug that I see in this implementation is that there is somehow a tiny horizontal scrollbar at the bottom. I think this is a bug in Qt 5.14 that doesn't appear 5.10.2. Various other components in the interface have those for me, but they are not in the builds. The scrollbar can be moved but does nothing.

### DIFF
--- a/resources/qml/PrintSetupTooltip.qml
+++ b/resources/qml/PrintSetupTooltip.qml
@@ -11,7 +11,7 @@ UM.PointingRectangle
     id: base
     property real sourceWidth: 0
     width: UM.Theme.getSize("tooltip").width
-    height: label.height + UM.Theme.getSize("tooltip_margins").height * 2
+    height: textScroll.height + UM.Theme.getSize("tooltip_margins").height * 2
     color: UM.Theme.getColor("tooltip")
 
     arrowSize: UM.Theme.getSize("default_arrow").width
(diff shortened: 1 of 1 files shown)


nallath:
CURA-7650

PR #8172, opened 2020-08-10
Retranslate ko_kr "ironing_flow"

n7484443:
en:i changed ironing_flow translation to "다림질 압출량". "과정" means process. so i changed it "압출량"-every flow word has been translated to this. 
kr:기존의 다림질 과정에서 flow를 "과정"이 아닌 "압출량"으로 바꾸었습니다.

### DIFF
--- a/resources/i18n/ko_KR/fdmprinter.def.json.po
+++ b/resources/i18n/ko_KR/fdmprinter.def.json.po
@@ -1488,12 +1488,12 @@ msgstr "다림질 라인 사이의 거리."
 #: fdmprinter.def.json
 msgctxt "ironing_flow label"
 msgid "Ironing Flow"
-msgstr "다림질 과정"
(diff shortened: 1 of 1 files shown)


nallath:
These suggestion have been forwarded to one of our resellers to check (Since I only speak English, Dutch & German, I'm in no position to check this). So it might take a bit until we hear back from them.

Ghostkeeper:
The reseller already responded. Unfortunately I had seen the response and applied the change myself without knowing the full context of who suggested the improvement originally, so attribution is wrong now. I'll merge this and fix the attribution as best as possible.

The reseller agrees with the updated terminology, but wanted to make a change to the description. See here for the new description: https://github.com/Ultimaker/Cura/commit/33ea62bb3d863bd234b415c3a54691d7d6b85110

PR #8189, opened 2020-08-13
CURA-7555: Add libArachne parameters in the front-end

konskarm:
The parameters are added only in the Shell category for now. Also, the default values used are not strictly defined yet.

CuraEngine equivalent: https://github.com/Ultimaker/CuraEngine/pull/1305

CURA-7555


resources/definitions/fdmprinter.def.json
```diff
@@ -1053,6 +1053,55 @@
                         }
                     }
                 },
+                "beading_strategy_type":
+                {
+                    "label": "Beading Strategy",
+                    "description": "Beading strategy used by libArachne to generate walls.",
```
fieldOfView, line 1059: Is the lemma "used by libArachne" of any insterest to the user?

It would be nice if the strategies were clarified here instead.

konskarm: It is still a work in progress, so we haven't settled down to the description that will be used in the end. I left the descriptions like that for now so that the settings can be easily found while we still integrate libArachne, by searching 'libArachne' or 'beading'.

jellespijker (approved):

resources/definitions/fdmprinter.def.json
fieldOfView: Seems you forgot to go over it later on...

PR #8195, opened 2020-08-14
Get property speed improvements

nallath:
I was doing some research on finding a way to create a fundamentally different way to handle properties. But during that time I again found some things that were being rather inefficient about it.

This PR should improve the speed of setting loading & value changes.

The one that is easiest to measure it the speed by which the slice message sending is handled. 
Before: 0.65s
After: 0.36s

### DIFF
--- a/cura/Settings/CuraContainerStack.py
+++ b/cura/Settings/CuraContainerStack.py
@@ -393,9 +393,11 @@ class CuraContainerStack(ContainerStack):
         if property_name == "settable_per_extruder":
             # Setable per extruder isn't a value that can ever change. So once we requested it once, we can just keep
             # that in memory.
-            if key not in self._settable_per_extruder_cache:
+            try:
+                return self._settable_per_extruder_cache[key]
+            except KeyError:
                 self._settable_per_extruder_cache[key] = super().getProperty(key, property_name, context)
-            return self._settable_per_extruder_cache[key]
+                return self._settable_per_extruder_cache[key]
 
         return super().getProperty(key, property_name, context)
 
--- a/cura/Settings/CuraContainerStack.py
+++ b/cura/Settings/CuraContainerStack.py
@@ -60,6 +60,8 @@ class CuraContainerStack(ContainerStack):
         import cura.CuraApplication #Here to prevent circular imports.
         self.setMetaDataEntry("setting_version", cura.CuraApplication.CuraApplication.SettingVersion)
 
+        self._settable_per_extruder_cache = {}
+
         self.setDirty(False)
 
     # This is emitted whenever the containersChanged signal from the ContainerStack base class is emitted.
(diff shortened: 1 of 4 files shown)


cura/Settings/CuraContainerStack.py
```diff
@@ -387,6 +389,18 @@ def getExtruderPositionValueWithDefault(self, key):
             value = int(Application.getInstance().getMachineManager().defaultExtruderPosition)
         return value
 
+    def getProperty(self, key: str, property_name: str, context = None) -> Any:
+        if property_name == "settable_per_extruder":
```
jellespijker, line 393: if you're really nit picking on speed and are willing to give in on code readability, you could shorten the "settable_per_extruder" to a shorter string. When trying out various code snippets with timeit I found a speed difference of 0.1 second for 10.000.000 iterations if I shortend it to "spe" ;-)

cura/Settings/CuraContainerStack.py
```diff
@@ -387,6 +389,18 @@ def getExtruderPositionValueWithDefault(self, key):
...
+            # Setable per extruder isn't a value that can ever change. So once we requested it once, we can just keep
+            # that in memory.
+            try:
```
jellespijker, line 396: You could alternatively use:
```
if key in self._settable_per_extruder_cache:
    return self._settable_per_extruder_cache[key]
else:
    ......
```
This will give you a 20% speed increase but for cases that the key isn't present. for cases that the key is present your code is  25% faster. So the question remains do we query more for existing keys then non-existing keys?

cura/Settings/CuraContainerStack.py
```diff
@@ -387,6 +389,18 @@ def getExtruderPositionValueWithDefault(self, key):
             value = int(Application.getInstance().getMachineManager().defaultExtruderPosition)
         return value
 
+    def getProperty(self, key: str, property_name: str, context = None) -> Any:
```
jellespijker, line 392: Did you look into the `functool.lru_cache` decorator for this as well?
Decorator to wrap a function with a memoizing callable that saves up to the maxsize most recent calls. It can save time when an expensive or I/O bound function is periodically called with the same arguments. Source: https://docs.python.org/3/library/functools.html

How often are the same queries made??

jellespijker (changes requested):
Code in overall looks good and is indeed a speed improvement. I got a few suggestions which could give a very small extra increase in speed depending on the actual scenarios. Which you are way more familiar with then me

cura/Settings/CuraContainerStack.py
nallath: Yep. Also discarded it. The LRU cache will only work if for that object the answer will be the same the next time that function is called. We don't have that luxury. (or at least, that's how I understood it)

cura/Settings/CuraContainerStack.py
nallath: Check the description of the commit :)

jellespijker (approved):

PR #8196, opened 2020-08-14
Speed up ci

nallath:
I was getting pretty annoyed with the very cluttered results of the test output. So I removed some of the unneeded things that the CI was doing (such as printing what plugins were installed & generating the translations).

This also splits the build and the test step, which ensures that you don't have to scroll through a wall of text before you reach what you actually want; The reason why your tests failed!


.github/workflows/cicd.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: CI/CD
+name: CI
```
jellespijker, line 2: File should be renamed i.m.o., it should reflect the workflows name. If this workflow isn't responsible for the Continues Delivery it should be named `ci.yml`

docker/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+cd build
+ctest3 -j4 --output-on-failure -T Test
```
jellespijker, line 3: Add newline at the end of a file

jellespijker (changes requested):

PR #8203, opened 2020-08-16
Updated pt_BR strings for 4.7

Patola:


Ghostkeeper:
That's okay. Thanks for helping us correct it!

PR #8231, opened 2020-08-25
Add Fused Form 3D printer profiles 

FusedForm:
New profiles for Fused Form 3D printers


nallath:
A lot of your profiles seem to be exactly the same. If that is the case, it's better to use the intheritance system in Cura. A good example for this is the creality_base; a number of printers share the same profiles (which prevents you from having to copy a lot of them)

FusedForm:
Thanks! I will check and modify them.

FusedForm:
> 
> 
> A lot of your profiles seem to be exactly the same. If that is the case, it's better to use the intheritance system in Cura. A good example for this is the creality_base; a number of printers share the same profiles (which prevents you from having to copy a lot of them)

@nallath  thanks for your feedback. I have modified the files under the inheritance system. Look forward for your feedback.
Thanks for your time

konskarm:
Hi @FusedForm, thanks for the significant file reduction. Things seem generally that they work fine, aside from this issue
image

It seems that all of the qualities appear for all the materials. This is probably because you have set "global_quality = True" in all of them. This should be set only for the global qualities, which are applied to all materials (even the ones you do not have qualities for). If you want global qualities, you need to make separate quality files for them (see [creality global qualities as an example](https://github.com/Ultimaker/Cura/blob/master/resources/quality/creality/base/base_global_adaptive.inst.cfg)).

FusedForm:
Hi @konskarm , thanks for your feedback. This has been a learning process for us.
I hope to have the corrections today.

FusedForm:
> 
> 
> Hi @konskarm , thanks for your feedback. This has been a learning process for us.
> I hope to have the corrections today.

Hi  @konskarm , sorry to disturb. 
For a given quality, if I set the parameter global_quality to false, or erase the definition of global_quality, the quality definition is no longer available for that material. What am I doing wrong here?
imagen

FusedForm:
Look, in this case this material quality is no longer available:

imagen

FusedForm:
Hi @konskarm, I have finished modifying the quality structures and material specific qualities under the global quality structures. Also added some more material definitions according to our experience.
Look forward for your feedback!

konskarm:
Everything seems to work as expected now, so I am accepting this.

Issue #6701, opened 2019-11-23
Super slow travel speed


mruiferreira:
I've updated from cura 4.2 to 4.4 and now when i start printing and the nozzle is travelling from the starting corner to the center of the build plate to start printing, the travel speed is extremely slow you can barely notice the hotend moving at all... downgrading to 4.2

fieldOfView:
Please fill out the issue template. We are missing vital information to do anything about this, like what printer you use. Also what is your start gcode?

mruiferreira:
**Application version**
4.4

**Platform**
Windows 10

**Printer**
Anycubic i3 Mega

**Reproduction steps**
1. Something you did.
Installed cura 4.4 (was using 4.2)

2. Something you did next.
Sliced a model without changing any settings

**Actual results**
The travel speed was extremely slow the print head wouldn't even visibly move.

**Additional information**
Downgraded back to 4.2 without changing any settings and the speed is correct in this version.

smartavionics:
Please save the project file from 4.4 (File -> Save) and zip or rename the .3mf file and attach to this issue, thanks.

fieldOfView:
> Also what is your start gcode?

Pugnator:
The same issue here, Anycubic i3Mega, nozzle and the bed moves very slow at the very beginning, when hotend and bed are calibrating.Below are the first 70 lines of Gcode I used

```;FLAVOR:Marlin
;TIME:985
;Filament used: 1.18923m
;Layer height: 0.3
;MINX:65.758
;MINY:20.871
;MINZ:0.2
;MAXX:144.241
;MAXY:189.128
;MAXZ:0.275
;Generated with Cura_SteamEngine 4.4.0
M140 S60
M105
M190 S60
M104 S220
M105
M109 S220
M82 ;absolute extrusion mode
G21 ;metric values
G90 ;absolute positioning
M82 ;set extruder to absolute mode
M107 ;start with the fan off
G28 X0 Y0 ;move X/Y to min endstops
G28 Z0 ;move Z to min endstops
G1 Z15.0 F100 ;move the platform down 15mm
G92 E0 ;zero the extruded length
G1 F200 E3 ;extrude 3mm of feed stock
G92 E0 ;zero the extruded length again
G1 F100
G0 Y20 F100
M117 Printing...
G5
G92 E0
G92 E0
G1 F2400 E-6
;LAYER_COUNT:1
;LAYER:0
M107
M204 S3000
M205 X10 Y10
[...]

PR #8281, opened 2020-08-31
New speed tokens represented in mm/min for start/end G-code.

skarasov:
G0 and G1 commands uses feedrate in mm/min. See https://reprap.org/wiki/G-code#G0_.26_G1:_Move
Token {speed_travel} represented in mm/sec.
Fix units of {speed_travel} token in start/end G-code
Issue #6701


Ghostkeeper:
It's missing dozens of other speeds here. Like the outer wall speed, printing speed, infill speed...

I think we need to consider what this does to existing workflows though. Some people use these speeds in comments to support features of the firmware that processes the g-code or a preprocessor (see the recent Octolapse issues) and those should remain in mm/s rather than mm/min. And people use it for their personal administration. Making it mm/min is unintuitive then.

skarasov:
Correct.
I see one option to fix the bug and keep exsisting tokens untouchable - create new tokens represented in mm/min (ex. {speed_travel_mmmin}) and replase them in all start g-code.
?

Ghostkeeper:
That's a better solution indeed, I think.

PR #8296, opened 2020-09-01
Adding Anycubic Kossel Linear and Linear Plus

FoxExe:
Configs for support Anycubic Kossel linear and Linear plus printers.

- start/stop g-code tested on original firmware and Marlin v1.1.9
- .stl model found on the network. Original author unknown, possibly J. Harris


nallath:
If the original author of the model isn't known, we're legally not able to accept it (since we might not have the rights to do so).

FoxExe:
Model distributed "as-is". Creative commons license.

FoxExe:
Found already present same model in repository. Removed old and changed configs.
Added Anycubic Kossel Pulley

resources/definitions/anycubic_kossel_pulley.def.json
```diff
@@ -0,0 +1,16 @@
...
+	"inherits": "anycubic_kossel",
+	"metadata": {
+		"visible": true,
+		"machine_extruder_trains": {
+			"0": "anycubic_kossel_extruder_0"
+		}
+	},
+	"overrides": {
+		"machine_name": {
+			"default_value": "Anicubic kossel Pulley"
```
Ghostkeeper, line 13: Typo here in the machine name.

resources/definitions/anycubic_kossel_linear_plus.def.json
```diff
@@ -0,0 +1,35 @@
...
+	"metadata": {
+		"visible": true,
+		"platform": "kossel_pro_build_platform.3mf",
+		"machine_extruder_trains": {
+			"0": "anycubic_kossel_extruder_0"
+		}
+	},
+	"overrides": {
+		"machine_name": {
+			"default_value": "Anicubic kossel Linear Plus"
```
Ghostkeeper, line 14: Typo here in the machine name.

Ghostkeeper (approved):
For the rest it does seem to work fine. I'll correct the typos and change the tabs into spaces as per our code style...

Vipcioo:
Hi everyone.
I am continuing in this thread, because it concerns this model as much as possible.

I am very happy that the Kossel delta printer has been officially added to the list in Cura.
Of course, I could have achieved a similar effect by creating a new printer with a round bed.


But with both methods, my aesthetic problem has always been "The Ugly Kossel Plus Table Preview".
Artifacts form on the view of the bed that look bad at best and interfere with work at worst.




I've been checking this for many versions backwards, say 2 years.
Now I can ask when the official model has been added to the printer list.

It seems that any bed that is not square has an overlay image which in effect generates changes that change when zoomed in/out. The "Prepare" view is the worst. It's a little better in "Preview", but still visible.

Maybe someone has an idea how to eliminate it?

Ghostkeeper:
Ah, yeah that is [Z-fighting](https://en.wikipedia.org/wiki/Z_fighting). I've moved the visualisation of the platform mesh slightly down to correct it. The same was done in the Johann Kossel printer definition with which the Anycubic Kossel Linear Pro printer shares the platform mesh.

PR #8330, opened 2020-09-06
PostProcessingPlugin FilamentChange use Marlin Advanced Pause Feature

Sekisback:
Modified the original script. Now it is possible to choose if you want to use Marlin Advanced Pause Feature for FilamentChange
[FilamentChange.txt](https://github.com/Ultimaker/Cura/files/5179689/FilamentChange.txt)


Ghostkeeper:
Closed this because it represented an extremely dangerous merge from Master into 4.7.

PR #8333, opened 2020-09-06
Pause-at-height should work for relative extrusion.

### LINKED ISSUE #8320: Post Processing Script "Pause at height" does not work with relative extrusion
If "Relative Extrusion" is set under "Special mode" then this value is not respected by the Pause at height script.

This line will reset the script to absolut extrusion and the rest of the script will be broken.
https://github.com/Ultimaker/Cura/blob/8c0e21f0d58549b645cbd5e5823acdfee0c44781/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py#L509

A proper solution is if the script checks if the extrusion is absolute or relative and only switches back to absolute when needed.

rburema:
Should fix #8320

Not tested yet, so draft. If any of you want to test this, feel free :-) If properly smoke-tested, you can consider this not a draft.


Ghostkeeper:
I think it refers to start g-code and end g-code, rather than post-processing scripts. Not sure if it happens for those though.

Ghostkeeper:
This has since been fixed in a slightly different way in PauseAtHeight.

PR #8351, opened 2020-09-09
Fix default filament diameter for custom printers

towe96:
Almost all new consumer-grade printers have switched to 1.75 mm filament.
Reflecting this change in the default printer profile for Cura reduces the chance of extrusion issues when newcomers set up their own machine in Cura.


nallath:
The problem with this change is that all people that currently have a custom printer that is 2.85 will suddenly change to 1.75mm (and thus, breaking their configuration)

Ghostkeeper:
Yes, this would require a version upgrade and a change in all current printers that don't override this setting.

Issue #8331, opened 2020-09-06
Change the "Display Progress On LCD" plugin to have an option to update every X lines or at time intervals.


Ravenex:
**Is your feature request related to a problem? Please describe.**
The "Display Progress On LCD" plugin is a fantastic feature I have been desiring since I started 3D printing. Though on really large prints it is much less useful as a single layer can take a very long time. Checking the time remaining still leaves a lot of uncertainty about how long it has been since the layer started and the message changed.

Additionally the message doesn't show until after the first layer is complete. So when loading a program I sliced days before if I don't remember the print time and no longer have setup in Cura I can calculate the finish time until I see the first layer print.

**Describe the solution you'd like**
I would also like additional options: "Update every X lines", and "Update every X seconds". I understand the later could add a fair amount of processing but it would be worthwhile to trade processing time for it on prints with long layers or on production prints where scheduling is important. It is possible the first option would be enough to cover the second option with a fine enough setting since any one line is limited in time.

Also, I would the time progress message output at the beginning of the first layer.

**Affected users and/or printers**
I think this would benefit all users but in particular users with large prints or production prints.

konskarm:
I added the 3rd-Party category, since the "Display Progress on LCD" post-processing script is mainly maintained by the community. 

@Ravenex your suggestions sound reasonable, but considering that this PP script is contributed to us, I do not think that your requested feature will get high enough priority within our team. Nevertheless, we would be happy to review the code in case you or someone else decides to add this feature!

Ghostkeeper:
Updating at time intervals would require porting the time estimation algorithm to Python. We already [have such a port, albeit in a modified state](https://github.com/Ultimaker/Cura/blob/master/scripts/check_gcode_buffer.py), but it's going to take minutes to execute that on a medium-sized g-code file. During this time, the Cura interface will be frozen. Windows and Ubuntu will indicate that the application crashed. Windows will start sending data to Microsoft again, etc.

Ravenex:
What about the every x lines variant? Would this be more reasonable?

I'm also wondering if this could be built as a standalone post processing program instead of a post processor inside Cura.

KimmoHop:
@Bostwickenator seems to be contributor of the script?

[...]

PR #8403, opened 2020-09-17
M73 R and update interval in Display Progress On LCD PP plugin

KimmoHop:
This updates post-process plugin to export remaining time in M73 R in addition to M117 status line.
Update remaining time with M117 can be updated periodically in addition to every layer.
Discussed in #8331

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
+++ b/plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
@@ -1,7 +1,7 @@
 # Cura PostProcessingPlugin
-# Author:   Mathias Lyngklip Kjeldgaard, Alexander Gee
+# Author:   Mathias Lyngklip Kjeldgaard, Alexander Gee, Kimmo Toivanen
 # Date:     July 31, 2019
-# Modified: May 22, 2020
+# Modified: Sep 09, 2020
 
 # Description:  This plugin displays progress on the LCD. It can output the estimated time remaining and the completion percentage.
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):
Looks good to me from a review perspective.

I'll look a bit into the setting descriptions. It would be good if they are a bit less technical. And I'll give it a good test, of course.

plugins/PostProcessingPlugin/scripts/DisplayProgressOnLCD.py
```diff
@@ -26,9 +26,19 @@ def getSettingDataString(self):
                 "time_remaining":
                 {
                     "label": "Time Remaining",
-                    "description": "When enabled, write Time Left: HHMMSS on the display using M117. This is updated every layer.",
-                    "type": "bool",
-                    "default_value": false
+                    "description": "Select to write remaining time on the display using M117 status line message (almost all printers) or using M73 command (Prusa and Marlin 2 if enabled).",
+                    "type": "enum",
```
Ghostkeeper, line 30: I see that this change from `bool` to `enum` needs a version upgrade, actually. It's a bit much to ask you to develop a version upgrade plug-in so I'll see what I can do.

nallath: Isn't it way easier to just add a new setting? 

That way there is also no need of a version upgrade script, which makes it a lot cleaner. 

So my proposal is to basicly let the orginal setting remain what it is ("Should time be written at all") and add a new setting that is enabled when the original setting is true ("So now that you want to write time, how should I do it?")

KimmoHop: I can't really say which way is better. Is there any other reason to bump the version?
Does each plugin have their own version tracking and upgrade system, or is it something global affecting all components?

If the version change is for future 4.9, it would give plugin developers an opportunity to go through their UIs and improve usability, if there is a need.
If this plugin is changed to old checkbox + new controls (I'm slightly worried about usability and wording), would it be on path to 4.8 or 4.9?

nallath: In either case it would be 4.9

I'm not against bumping the version number, mind you. But I don't really see a very compelling reason to do so in this case. I will discuss this with the other developers as well so find out what they think.

Ghostkeeper: It would indeed always be 4.9 since 4.8 is feature-frozen at the moment, with the beta imminent.
I agree that splitting it up into two settings is nicer. I'll just create a checkbox and a drop-down.

KimmoHop: So it's OK to add settings but NOT OK to change setting types.
Are there restrictions to removing settings? If not, then one could drop existing setting and introduce new setting to go around type change ;)

Ghostkeeper: What the version upgrade does is to ensure that people with existing profiles keep their behaviour the same. Dropping the existing setting and adding a new one under a different name effectively resets the setting to its (new) default. So it would still need a version upgrade, or users would see that their setting got magically disabled in the upgrade.

PR #8407, opened 2020-09-18
Add support for Creality CR-6 SE

MathieuMH:
Based on the profile that was shared within the Creality CR-6 SE community I wanted to merge this with the main branch in order to have it available by default in upcomming releases.

### DIFF
--- a/resources/definitions/creality_base.def.json
+++ b/resources/definitions/creality_base.def.json
@@ -176,7 +176,7 @@
 
         "skirt_brim_speed": { "value": "speed_layer_0" },
 
-        "line_width": { "value": "machine_nozzle_size" },
+        "line_width": { "value": "machine_nozzle_size * 1.1" },
 
         "optimize_wall_printing_order": { "value": "True" },
 
@@ -185,8 +185,8 @@
         "material_flow": { "value": 100 },
         "travel_compensate_overlapping_walls_0_enabled": { "value": "False" },
 
-        "z_seam_type": { "value": "'back'" },
-        "z_seam_corner": { "value": "'z_seam_corner_weighted'" },
+        "z_seam_type": { "value": "'sharpest_corner'" },
+        "z_seam_corner": { "value": "'z_seam_corner_inner'" },
         
         "infill_sparse_density": { "value": "20" },
         "infill_pattern": { "value": "'lines' if infill_sparse_density > 50 else 'cubic'" },
@@ -215,7 +215,7 @@
         
         "retraction_hop_enabled": { "value": "False" },
         "retraction_hop": { "value": 0.2 },
-        "retraction_combing": { "value": "'off' if retraction_hop_enabled else 'noskin'" },
+        "retraction_combing": { "value": "'off' if retraction_hop_enabled else 'infill'" },
         "retraction_combing_max_distance": { "value": 30 },
         "travel_avoid_other_parts": { "value": true },
         "travel_avoid_supports": { "value": true },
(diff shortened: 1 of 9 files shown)


fieldOfView:
I don't think it is a good idea to make changes to `creality_base` under the guise of adding a new machine definition. These changes are applied to all creality printers.

Sebazzz:
> I don't think it is a good idea to make changes to creality_base under the guise of adding a new machine definition. These changes are applied to all creality printers.

My guess is that these definitions were taken from Creality Slicer 4.2 and the Cura base profile has probably updated since then.

nallath:
Even so, we won't accept these changes as they stand now (also because the CI failed on them) 

The added variants are not needed, since the creality-base already has them. As FieldOfView mentioned; changing the base profile when some changes are needed for a different machine is definitely not the way to do it.

Sebazzz:
> Even so, we won't accept these changes as they stand now (also because the CI failed on them)

It was not an argument against or for this change 😄

Ghostkeeper:
Indeed Sebazzz, I recognise some of those changes indeed as things we changed after 4.2 at the request of the community testing various Creality printers. These changes are assumed to apply to all of their printers. We can probably revert all changes to `creality_base` made here and you'd get better results out of the CR-6 SE too.
Needs testing though, of course.

Ghostkeeper:
Aside from that the tests are failing because:
* Setting version is outdated (profiles meant for 4.7 instead of 4.8). Most likely there is no shrinkage compensation factor involved here so just incrementing those to 16 should work.
* The `machine_head_polygon` setting doesn't exist any more. Please remove it.

Sebazzz:
> Needs testing though, of course.

If @MathieuMH did not fix the findings yet by the time I get my CR-6 ([which takes a while because Creality messed up shipping](https://damsteen.nl/blog/2020/08/29/3d-printers-incoming)) I will submit a new profile. I do hope that #8321 is fixed then otherwise I can't test it on Cura 4.7.1, the CR-6 has been affected by that issue.

MathieuMH:
Sorry you had to wait on it.
I removed the `machine_head_polygon ` and update the nozzle cfg files to v16 instead of v15. Checks including CI passed now.

konskarm:
Everything seems fine now. The CR-6 SE printer shows up in the list and it slices without issues. As a result, I am accepting this PR so that the printer is included in Cura 4.8.

looxonline:
> Everything seems fine now. The CR-6 SE printer shows up in the list and it slices without issues. As a result, I am accepting this PR so that the printer is included in Cura 4.8.

@konskarm Is there any chance of you doing the same for my BIQU PR so that it is included in Cura 4.8? I'm currently using it and it works well. https://github.com/Ultimaker/Cura/pull/8276

DreadKnight:
Not sure when the next release is, but would be cool if profile could be shared around as a file, if possible.
I've set up mine manually, but still trying to figure out if there are any major differences, as I'm new to this.

MathieuMH:
@DreadKnight Yes you can. It is simply coping a the file at per PR. 
In this PR select at the top changes files. From the latest commit use the file "resources/definitions/creality_cr6se.def.json" and past it in the correct folder.
Same for the nozzle files. There I suppose you need to revert back the previous version v15 instead of v16.
For example the .4mm nozzle resources/variants/creality_cr6se_0.4.inst.cfg
For 4.8 it is:
`setting_version = 16`
For 4.7 you should use:
setting_version = 15

Ghostkeeper:
> Not sure when the next release is,

No promises of course, but we normally release approximately every 2 months, which would place the stable release towards the end of October and the beta a few weeks before that.

DreadKnight:
Sounds good, thanks for letting me know! Pretty agile development :+1:

PR #8419, opened 2020-09-21
Create KOONOVO

KOONOVO3DPrinter:
Create the KOONOVO 3D printer


Ghostkeeper:
This one fails to be recognised as a printer since it doesn't have the .def.json extension.

Issue #8415, opened 2020-09-20
[4.7.1] Retracting at start of print


FlailAway:
**Application version**
4.7.1

**Platform**
Mint 20+Cinnamon

**Printer**
Any I am guessing

**Reproduction steps**
Save to File

**Actual results**
Open in Text Editor
`
G92 E0

G1 F1500 E-6    

;LAYER_COUNT:55

;LAYER:0
`

**DESIRED, Expected results**
`G92 E0

;  G1 F1500 E-6   // **<-- Stop this line from ever appearing**

;LAYER_COUNT:55

;LAYER:0
`

**Additional information**
On small jobs, I get 20 Skirts because I have set **"Skirt/Brim minimum length = 250** otherwise filament appears at about Layer-2.

Why must there be a mandatory retraction **before** anything is even printed.

Asterchades:
Definitely doesn't apply to "any" printer. The default FFF/FDM and Creality profiles don't include such a command in their start GCode and it isn't generated automatically (like heating).

You've either changed your start GCode manually or one (or more) of the profiles needs fixing. But without specifying which printer you're using or uploading a project for inspection it's impossible to tell which is happening for certain.

FlailAway:
> Definitely doesn't apply to "any" printer. The default FFF/FDM and Creality profiles don't include such a command in their start GCode and it isn't generated automatically (like heating).

Never say "never," eh! Not meaning to be difficult, but have you actually looked? I just defined an Ultimaker 3 printer, changed **nothing** and loaded an .stl. Sliced and saved the GCode to the desktop and here it is,

`;START_OF_HEADER
;HEADER_VERSION:0.1
;FLAVOR:Griffin
;GENERATOR.NAME:Cura_SteamEngine
;GENERATOR.VERSION:4.7.1
;GENERATOR.BUILD_DATE:2020-09-02
;TARGET_MACHINE.NAME:Ultimaker 3
;EXTRUDER_TRAIN.0.INITIAL_TEMPERATURE:205
;EXTRUDER_TRAIN.0.MATERIAL.VOLUME_USED:2829
;EXTRUDER_TRAIN.0.MATERIAL.GUID:506c9f0d-e3aa-4bd4-b2d2-23e2425b1aa9
;EXTRUDER_TRAIN.0.NOZZLE.DIAMETER:0.4
;EXTRUDER_TRAIN.0.NOZZLE.NAME:AA 0.4
;BUILD_PLATE.TYPE:glass
[...]

Issue #5713, opened 2019-05-08
[4.0] Retract in start g-code issue


ilko-k:
Whenever a retraction line is inserted in the start g-code, it is over-retracting like 10mm lenght (I did not measure), even if it is set to 0.1mm. It seems that the value is ignored and does an absolute retraction.

note : it occurs when printing via USB. Will try via SDC soon

ilko-k:
Hey @smartavionics, I'll tell more about that when my actual print wil be finished. For now, I'm not sure what details I could provide, this occurs when I print via USB and with any project. I didn't try by SDC yet but I'll try.

ilko-k:
ok, so here is the two lines that are written after the custom start g-code : 

```
; End of custom start GCode
G92 E0
G1 F2400 E-4
```

An additional retraction is generated whereas it's not needed.

smartavionics:
OK, but that in isolation tells me next to nothing. Without seeing your settings it's very difficult to guess what's happening. Please either attach the complete gcode file or the project file (they will need to be zipped to let you attach them). Thanks.

ilko-k:
ok, I've zipped a project plus the settings I'm using. I'm printing PETG.
[Calicube.zip](https://github.com/Ultimaker/Cura/files/3159533/Calicube.zip)

smartavionics:
Thanks for the files, all is clear now.

> An additional retraction is generated whereas it's not needed.

Well, that retraction (and the prime that follows after moving to where the skirt is printed) are OK, the real problem is your include of this line in your start code:

```
G1 F2000.0 E-1 ; retract 1mm
```

Remember, you are using absolute extrusion mode and so that -1 means wind 21 mm of filament back as you have previously extruded 20 mm of filament!

Hope this helps.

Ghostkeeper:
> An additional retraction is generated whereas it's not needed.

This retraction prevents a big fat line from being drawn across your build plate, which the rest of the print may collide with, leaving a scar across the bottom of your print. The retraction is unretracted a couple of lines later:

```
;TYPE:SKIRT
G1 F2400 E0
```

GregValiant:
Can we close this?  It appears it was a StartUp Gcode issue.

ilko-k:
Oh wow i completely forgot this. Closing. Sorry for the inconvenience.

PR #8441, opened 2020-09-24
Option to enable/disable retract before nozzle first move.

skarasov:
#5713 and #8415
For bowden extruders its unwanted to retract before first move!

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -3657,6 +3657,16 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "retract_before_first_move":
+                {
+                    "label": "Retract Before First Move",
+                    "description": "Retract the filament before first move of nozzle.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": "retraction_enable",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
                 "retraction_amount":
                 {
                     "label": "Retraction Distance",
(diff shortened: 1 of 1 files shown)


skarasov:
Option for CE#1331.

Ghostkeeper:
Why would Bowden extruders not want to retract before the first move?

skarasov:
Bowden/ABS. Retract before first move leads to the need to significantly increase brim size. Because after retract and first move - long time without prime.
Eplanation - retract at (0,0), move to (100,100), teflon tube bends resultin additional retract of filament.

Ghostkeeper:
And this doesn't happen with any other moves? Just the first move? Maybe you'll need to disable retractions altogether?

skarasov:
It does not happen with other moves! At least not in such scale and solved by retraction adgustments.
First move - very long move. Bend of teflon tube = big influence on retract.
IMG_20200929_163929

Ghostkeeper:
Ah, right so according to that theory it should apply to all travel moves that move great distances.

skarasov:
If you dont seen this effect, it does not mean - no such effect!
Ultimaker printers - not ALL printers!

You have at least 5 reqests for this option. So why not?

Ghostkeeper:
There needs to be a clear use case for why a user would change a setting. There are literally millions of adjustments possible within a print as to which moves get retracted and which don't, when to avoid objects or to lift the nozzle, etc. We can't make a case for a setting unless there is a realistic reason for a user to modify it other than gut feeling.

The Bowden tube flexing _is_ a valid use case though. It could result in a slight accidental retraction or unretraction of the filament whenever the print head moves closer or further away from the feeder.

We've discussed this in Ultimaker but decided that control over this retraction will not solve that issue. The problem is not unique to the first layer but will occur in the same way e.g. when moving from the prime tower to the model or just when printing 2 models. And on the first layer this is perhaps the least concerning case because there you still have a brim or skirt to catch the difference. On other layers you won't have that! If the Bowden tube flexing is enough of a problem to warrant an extra Brim line, that means you're missing out on at least an extra contour on other layers. What's more, disabling retraction won't cause it to prime more after the travel move, the material will still be retracted due to the flexing of the Bowden tube.

For those reasons we've decided to reject this pull request. Sorry!

PR #8463, opened 2020-09-29
Fix connecting to printer via USB

rrrlasse:
If the baudrate was anything else than 115200, detection of baudrate could fail
because readline() with a wrong baudrate would never return, even if timeouts
were set.

Update: Sometimes detection would succeed anyway if restarting Cura enough times, or pulling the USB cable out and in enough times. The problem may be that readline() would read an infinite stream of random data that usually, but not always, would happen to have an EOF character by coincidence.

The fix in the PR still applies

### DIFF
--- a/plugins/USBPrinting/AutoDetectBaudJob.py
+++ b/plugins/USBPrinting/AutoDetectBaudJob.py
@@ -71,7 +71,9 @@ class AutoDetectBaudJob(Job):
                 timeout_time = time() + wait_response_timeout
 
                 while timeout_time > time():
-                    # If baudrate is wrong, then readline() might never return, even with timeouts set. Using read_until with size limit seems to fix this.
+                    # If baudrate is wrong, then readline() might never
+                    # return, even with timeouts set. Using read_until
+                    # with size limit seems to fix this.
                     line = serial.read_until(size = 100)
                     if b"ok" in line and b"T:" in line:
                         successful_responses += 1
--- a/plugins/USBPrinting/AutoDetectBaudJob.py
+++ b/plugins/USBPrinting/AutoDetectBaudJob.py
@@ -71,7 +71,8 @@ class AutoDetectBaudJob(Job):
                 timeout_time = time() + wait_response_timeout
 
                 while timeout_time > time():
-                    line = serial.readline()
+                    # If baudrate is wrong, then readline() might never return, even with timeouts set. Using read_until with size limit seems to fix this.
+                    line = serial.read_until(size = 100)
                     if b"ok" in line and b"T:" in line:
                         successful_responses += 1
                         if successful_responses >= 1:


rrrlasse:
Note that I'm not a Python developer or have much insight in Cura or serial ports. See this PR as advisory only.

Ghostkeeper:
Hmm, but with this, if the baud rate is correct, it will wait until it receives 100 bytes, which will probably not happen since the response to `M105` isn't that long.

rrrlasse:
> Hmm, but with this, if the baud rate is correct, it will wait until it receives 100 bytes, which will probably not happen since the response to `M105` isn't that long. Maybe it just needs to specify the `size` optional parameter of `readline`.

No, read_until() will "Read until an expected sequence is found (‘\n’ by default), the size is exceeded or until timeout occurs." according to https://pyserial.readthedocs.io/en/latest/pyserial_api.html

plugins/USBPrinting/AutoDetectBaudJob.py
```diff
@@ -71,7 +71,8 @@ def run(self) -> None:
                 timeout_time = time() + wait_response_timeout
 
                 while timeout_time > time():
-                    line = serial.readline()
+                    # If baudrate is wrong, then readline() might never return, even with timeouts set. Using read_until with size limit seems to fix this.
```
StefanBruens, line 74:
```suggestion
                    # If baudrate is wrong, then readline() might never
                    # return, even with timeouts set. Using read_until
                    # with size limit seems to fix this.
```

rburema:
Sorry this took so long (yikes!), it seems this PR fell between the cracks somehow.

Since it's so close to the next release, we decided to put it in master, so it will be in the release _after_ that. Better late than never I suppose...

(Devs: It's internal ticket CURA-8323 now.)

Ranchdawg:
USB Cura 4.13 repeatedly attempts to connect to my Anet A8 Plus.  I end up disconnecting USB and printing with micoSD card.

rburema:
Hi @Ranchdawg 

Attempts but not succeeds, even though it should? The that's a bug: You can open a separate issue, where we'd need the logs and other relevant information (for example, did this behaviour _start_ in 4.13?) to even begin diagnosing the problem. (USB support isn't really a feature we focus on these days, so it might get deferred, since we already have a sizeable backlog.)

Attempts and succeeds (but you don't want it)? This is expected behaviour: _However_ if you're sure you don't need to connect to any printers via USB, you can also Disable the (built-in) plugin that's responsible for that via the Marketplace.

Ranchdawg:
Got it to work by hitting reset on my printer On Apr 15, 2022 2:18 AM, Remco Burema ***@***.***> wrote:
Hi @Ranchdawg
Attempts but not succeeds, even though it should? The that's a bug: You can open a separate issue, where we'd need the logs and other relevant information (for example, did this behaviour start in 4.13?) to even begin diagnosing the problem. (USB support isn't really a feature we focus on these days, so it might get deferred, since we already have a sizeable backlog.)
Attempts and succeeds (but you don't want it)? This is expected behaviour: However if you're sure you don't need to connect to any printers via USB, you can also Disable the (built-in) plugin that's responsible for that via the Marketplace.

PR #8465, opened 2020-09-29
Fabtotum profiles fix

PaoloP74:
This PR temporarily disables the quality profiles of the Fabtotum printer until they are corrected.
Since version 4.3 of Cura the profiles are not loaded correctly. Pending a correction I propose that they be disabled.
I have removed the non-machine related settings from the printer definition and removed the minimum retraction distance that Cura will calculate from the profiles.

### DIFF
--- a/resources/definitions/fabtotum.def.json
+++ b/resources/definitions/fabtotum.def.json
@@ -8,7 +8,7 @@
         "manufacturer": "FABtotum",
         "file_formats": "text/x-gcode",
         "platform": "fabtotum_platform.3mf",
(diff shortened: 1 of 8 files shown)


Ghostkeeper:
These profiles were not assigned to any nozzle. Apparently in 4.3 there were nozzle profiles added for FABtotum but it was never considered which nozzle the existing quality profiles were made for.
I've corrected that for 4.8 instead: https://github.com/Ultimaker/Cura/commit/2ec5ffba0d26b88de95c6e3845853cc815b806bb

PR #8476, opened 2020-10-02
Remove redundant variable

aerotog:
While working on a patch to make Cura work with my printer I noticed this variable is unnecessary.


Ghostkeeper (approved):
Yup. This variable was at one point necessary to track the number of successful responses to validate a safe connection. But the number of successful responses needed to be reduced from 3 to 1 for some printers. Since 2018 this variable is no longer necessary.

PR #8499, opened 2020-10-05
CURA-7735: Make symlink following optional when checking the plugin signatures

konskarm:
Merge along with https://github.com/Ultimaker/Uranium/pull/642

CURA-7735


rburema:
Note: Build failure unrelated to this ticket. Building 4.6 on Jenkins still works (well Windows failed, but that's because 'no space left on device').

PR #8508, opened 2020-10-07
Update creality_cr10max.def.json

Jip-Hop:
We need the same fix as in https://github.com/Ultimaker/Cura/pull/7526 for the Creality CR-10 Max as well, otherwise a triangle is drawn before each print.


konskarm:
I cannot really test this, since I do not own a CR10max, but considering it was reviewed and tested for CR10S, I'll take your word for it that it is working. From a code perspective, it looks good.

PR #8509, opened 2020-10-07
Update definition files for Hellbot Magna 1 and Hellbot Magna dual 

DevelopmentHellbot:
Change in the Author's name, added heated bed in both printers.


konskarm (changes requested):
Fix formatting in file resources/definitions/hellbot_magna_dual.def.json for the "machine_heated_bed" setting.

DevelopmentHellbot (commented):
Added new printers: 
Magna 2 230 
Magna 2 230 dual
Magna 2 300 
Magna 2 300 dual 

Added definition files, extruder files, bed meshes and bed images

konskarm:
> Deleted the files to verify that everything is ok. Apologies, i'm new to this process.

Don't worry about making many commits, we can squash them into one when we merge this. 

Generally it looks good, just please rename all extruder files to `...extruder_1.def.json` instead of `...E1.def.json`. Also, we start the counting of extruders from 0 instead of 1. So in this case we should have the following:

- `/extruders/hellbot_magna_2_230_E1.def.json` -> `/extruders/hellbot_magna_2_230_extruder_0.def.json`
- `/extruders/hellbot_magna_2_230_dual_E1.def.json` -> `/extruders/hellbot_magna_2_230_dual_extruder_0.def.json`
- `/extruders/hellbot_magna_2_230_dual_E2.def.json` -> `/extruders/hellbot_magna_2_230_dual_extruder_1.def.json`
- `/extruders/hellbot_magna_2_300_E1.def.json` -> `/extruders/hellbot_magna_2_300_extruder_0.def.json`
- `/extruders/hellbot_magna_2_300_dual_E1.def.json` -> `/extruders/hellbot_magna_2_300_dual_extruder_0.def.json`
- `/extruders/hellbot_magna_2_300_dual_E2.def.json` -> `/extruders/hellbot_magna_2_300_dual_extruder_1.def.json`

It is clearer this way and it follows our naming conventions.

nallath:
Our CI is still reporting issues with the changes that you made.

Ghostkeeper:
What you're seeing there is that our main test suite failed. We have a test suite for the main code base as well as a test suite for several plug-ins, and then another test suite for code style. The main code base contains the tests for the resources such as your printer. That's why it's complaining there. If running from command line, you need to run it in verbose mode to see the actual error, or look into the test logs it generates (those are always verbose).

    ctest --verbose

It's currently complaining that the printer definition still refers to the old name of the extruder definitions:

```
>       assert extruder_trains[position] == extruder_id, "The extruder referenced in the machine definition must match up."
E       AssertionError: The extruder referenced in the machine definition must match up.
E       assert 'hellbot_magna_2_230_dual_E1' == 'hellbot_magn...al_extruder_0'
E         - hellbot_magna_2_230_dual_E1
E         ?                          ^^
E         + hellbot_magna_2_230_dual_extruder_0
E         ?                          ^^^^^^^^^^
```

So if you change the definition files from

    "0": "hellbot_magna_2_230_E1"

into

    "0": "hellbot_magna_2_230_dual_extruder_0"

And such for each printer, then the issue should get resolved.

PR #8510, opened 2020-10-08
Mark disallowed areas as disallowed areas

Ghostkeeper:
I've added a feature to libnest2d to be able to mark items as disallowed areas. Other items won't collate with disallowed areas to minimise the used surface area of the group together. But they will still avoid the position of the disallowed areas and pack inside them.

Also improve the filtering a bit so that it's no longer dependent on the order. Doesn't really matter in this case but it should be more robust.

Contributes to issue CURA-7754.
This change depends on a change to pynest2d: https://github.com/Ultimaker/pynest2d/pull/2 which in turn depends on a change to libnest2d: https://github.com/Ultimaker/libnest2d/pull/1


rburema:
Tests failing, but that's true in the same way for the branch this is merged into. As this fixes another bug, the failing tests do not prohibit merging in this case.

PR #8523, opened 2020-10-12
Fix some montior page hacks

nallath:
It only fixes a few of em, but hey, it's progress!

Cura-7745

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/MonitorContextMenuButton.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/MonitorContextMenuButton.qml
@@ -18,7 +18,7 @@ Button
         width: base.width
     }
     contentItem: Label {
-        color: enabled ? UM.Theme.getColor("monitor_text_primary") : UM.Theme.getColor("monitor_text_disabled")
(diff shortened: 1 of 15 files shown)


plugins/UM3NetworkPrinting/resources/qml/MonitorContextMenuButton.qml
```diff
@@ -18,7 +18,7 @@ Button
         width: base.width
     }
     contentItem: Label {
-        color: enabled ? UM.Theme.getColor("monitor_text_primary") : UM.Theme.getColor("monitor_text_disabled")
+        color: enabled ? UM.Theme.getColor("text") : UM.Theme.getColor("monitor_text_disabled")
```
fieldOfView, line 21: You probably want to use an alternative for monitor_text_disabled too then.

nallath: Yeah, I started this in order to find some quick replacements, but unfortunately, we don't really have disabled text quite the same way in other parts. 

I will spend some extra time on this. The sheer amount of TODO's in this code is just painfull.

Ghostkeeper (approved):
The theme is somewhat out of scope anyway.
I'll test tomorrow, but looks fine from a review perspective.

PR #8532, opened 2020-10-13
Compatibility for newer python-zeroconf versions

Gobbel2000:
python-zeroconf Version 0.24.0 added support for multiple addresses and IPv6 addresses. The attribute ServiceInfo.address (which is still used here) is therefore deprecated and fully removed as of Version 0.27.0. This change makes it compatible with newer versions, while retaining backwards compatibility for Versions <= 0.23.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
+++ b/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
@@ -155,8 +155,8 @@ class ZeroConfClient:
         return True
 
     @staticmethod
-    def _getAddress(info: ServiceInfo) -> Optional[str]:
-        """Retrieve IPv4 address from the ServiceInfo as string, None if there is no address."""
+    def _getAddress(info: ServiceInfo) -> Optional[bytes]:
+        """Retrieve IPv4 address from the ServiceInfo as bytes, None if there is no address."""
 
         # Check the version of python-zeroconf
         if Version(zeroconf_version) >= Version([0, 24, 0]):
--- a/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
+++ b/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
@@ -155,7 +155,7 @@ class ZeroConfClient:
         return True
 
     @staticmethod
-    def _getAddress(info: ServiceInfo) -> str:
+    def _getAddress(info: ServiceInfo) -> Optional[str]:
         """Retrieve IPv4 address from the ServiceInfo as string, None if there is no address."""
 
         # Check the version of python-zeroconf
--- a/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
+++ b/plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
@@ -3,7 +3,7 @@
 from queue import Queue
 from threading import Thread, Event
 from time import time
-from typing import Optional
+from typing import List, Optional
 
 from zeroconf import Zeroconf, ServiceBrowser, ServiceStateChange, ServiceInfo
 from zeroconf import __version__ as zeroconf_version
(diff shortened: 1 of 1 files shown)


plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
```diff
@@ -152,6 +154,16 @@ def _onServiceAdded(self, zero_conf: Zeroconf, service_type: str, name: str) ->
 
         return True
 
+    @staticmethod
+    def _getAddress(info: ServiceInfo) -> str:
```
konskarm, line 158: Info address can be None as it seems 
```suggestion
    def _getAddress(info: ServiceInfo) -> Optional[str]:
```

konskarm (changes requested):
This change should satisfy mypy failed tests

plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
```diff
@@ -152,6 +154,16 @@ def _onServiceAdded(self, zero_conf: Zeroconf, service_type: str, name: str) ->
...
+        """Retrieve IPv4 address from the ServiceInfo as string, None if there is no address."""
+
+        # Check the version of python-zeroconf
+        if Version(zeroconf_version) >= Version([0, 24, 0]):
+            return next(iter(info.addresses), None)
```
konskarm, line 163: Next expects a string, not None

Gobbel2000:
Yes, if the service doesn't provide an address, _getAddress() will return None. I added your change to the typing.

plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
```diff
@@ -129,20 +131,20 @@ def _onServiceAdded(self, zero_conf: Zeroconf, service_type: str, name: str) ->
...
             if new_info is not None:
                 info = new_info
 
-        if info and info.address:
+        if info and self._getAddress(info):
             type_of_device = info.properties.get(b"type", None)
             if type_of_device:
                 if type_of_device == b"printer":
-                    address = '.'.join(map(str, info.address))
+                    address = '.'.join(map(str, self._getAddress(info)))
```
Gobbel2000, line 147: I just noticed that I made it so that _getAddress() returns the address in byte form (or None) because it gets called four times and only once is the string actually needed. But when converting to a string here, the typing doesn't like the possibility of None being returned, even though that is ruled out by the test above. I haven't used the typing system enough to know what the best way to deal with this is.

plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
Gobbel2000: None is the default value here, in case info.addresses is empty. As far as I know it accepts any type.

plugins/UM3NetworkPrinting/src/Network/ZeroConfClient.py
konskarm: The problem here is that map expects an iterable, but _getAddress() returns Optional[bytes]. So if map sees None here, it will raise a 'NonType is not iterable' exception. Mypy complains because of that. So this case needs to be handled (by making next to not return None, and surrounding line 147 with a try-except StopIteration).

Also, I notice that `'.'.join(map(str, b"helloWorld"))` will output `'104.101.108.108.111.87.111.114.108.100'`. Are you sure this should return bytes? 
Note: I cannot get it to invoke this line of code locally, so I am asking to be sure.

Gobbel2000: Even if _getAddress() raised StopIteration instead of returning None there, info.address on older zeroconf versions would still return None, and you would have to add a try-except to all the four places where _getAddress() is called. Instead I went with a list of addresses which still satisfies the boolean checks, we just need to take the first item when converting to a string here.

Zeroconf stores the addresses as bytes, so for example `'192.168.178.1'` would be `b'\xc0\xa8\xb2\x01'`. When iterating over a bytes object, the individual elements are represented as int, so this line converts the address bytes from _getAddress into its standard string representation. 

I have tested this with old and new zeroconf versions, with and without giving the ServiceInfo an address and all works as expected for me.

Gobbel2000:
Commit cfccf949 solves this as well with the difference being, that old zeroconf versions (< 0.23.0) will not work anymore using that change, as these versions _only_ provide `ServiceInfo.address`. If that support is not needed, this pull request can be closed.

PR #8533, opened 2020-10-13
disable material shrinkage percentage

jellespijker:
Contributes to CURA-7771


rburema:
Checks where cancelled during build, which is why they're 'failing'.

PR #8542, opened 2020-10-14
Double Extruder profiles

FusedForm:
Hi, this is just an addition of double extruder machines and a second extruder definition.


konskarm:
Your local Cura master branch is not updated, since I see that FusedForm 300, 600, 600plus, mini, and STD already exist in our codebase. Please update it so that the PR is informed, because at the moment it shows that all FusedForm printers are about to be added from scratch.

konskarm:
Also it seems that tests are failing, so please have a look at the test output

FusedForm:
Thanks, I will look at it

FusedForm:
I have updated my repository and the files. Thanks for your help.
I look forward for your feedback.
Thanks in advance for your support!

FusedForm:
Hi @konskarm, I have updated my repository and the configuration files. Thanks for your help.
I look forward for your feedback.
Thanks in advance for your support!

konskarm:
I notice that the fusedform_doppia_base.def.json is exactly the same as fusedform_base.def.json, just with the extra extruder train. I do not see a need for that, so I would suggest you to remove the fusedform_doppia_base.def.json completely and make all doppia printers inherit from the fusedform_base. Then just override the machine_extruder_trains in the doppia machines to include the second extruder (which, as far as I see, is already overridden) .
Of course if you plan to radically change the fusedform_doppia_base in the future, then disregard the above statement.

Also, I noticed that the FF300 and FF300 Doppia have different sizes (360x300x320 vs 320x360x320). Is this intentional?

FusedForm:
Hi @konskarm, thanks for the feedback.
I have erased the doppia base definition, added the two extruder trains in the base definition and used the extruder qty in the definitions for it to use either 1 or 2 extruders.

The dimensions of the mahines can change because of the second extruder. That modification is intentional and depends on our components packing inside the machines.

Thanks again

konskarm:
Hi @FusedForm. Sorry for the slow responses, we've been quite busy with the 4.8 release. 

What I meant with my comment was to have only one extruder train in the `fused_form_base`, so that single extrusion machines will always report one extruder, and override the `machine_extruder_trains` in the FusedForm doppia machines. This way it is more logical, because we don't run the risk of single extruding FusedForm printers to report at any point that they have two extruders, even though they are limited to one through the `"machine_extruder_count":  { "value": 1 }` .

Sorry if I was not clear enough.

FusedForm:
Hi @konskarm,
yes, I was playing a lot with those options and I had

FusedForm:
Hi @konskarm,
yes, I was playing a lot with those options and I had

FusedForm:
Hi @konskarm,
don´t worry about the delay in the responses, it must have been very busy for you. I understand.

About the overriding of the extruder trains, I did try to override the extruder trains as you mentioned (tried various options) but I had errors given by the Github compliation algorithm, so finally I decided to do it that way to have a straightforward process at least regarding to the verification algorithm.

By our side I think it is not much of a problem if it remains in this way, let me know if it is needed by your side to be changed.

Let me know if we should change it the other way.

Thanks again for your time and help

konskarm:
I see that Creality follows the way I mentioned (see examples of `creality_base.def.json`, which has one extruder train, and `creality_cr-x.def.json` which overrides the trains with its own). I would be surprised if this is giving you errors, but maybe I am missing something there. 
If it's not too much work, I will insisting on that because I want to avoid an unforseen future error from coming up due to that.

FusedForm:
Hi @konskarm,
Congratulations on the release 4.8.
I have checked the creality definitions and what I understood is that they are defining the "creality_cr-x.def.json" with the "fdmprinter" as a base definition, not using the "creality_base" as the base definition for the "creality_cr-x.def.json".  So what I understand is that "creality_cr-x" and "creality_base" would be something like two base definitions in paralell, not the "creality_cr-x" overriding the "creality_base" definition.

The way I found it it work without generating "extruder nr errors" was to define the second extruder "fusedform_base_extruder_1.def"  pulling the FF300 doppia as machine definition in the metadata section of the "fusedform_base_extruder_1.def" . 
In this way I could effectively override the extruder trains definitions following a single "fusedform_base" machine definition on all the machines.

Let me know your thoughts.
Thanks!

konskarm:
Reading your comment makes me realize that you were correct in the first place to have a separate `fusedform_base_doppia.def.json`. It is the most appropriate way, it just should inherit from `fusedform_base.def.json` and only override the machine_extruder_trains. I am **terribly** sorry that I didn't see it earlier.

Fortunately the change is not that big. The following have to be done (I also ran the tests locally and tried slicing, everything seems to work as expected):

- Create a `fusedform_doppia_base_extruder_0.def.json` which is exactly the same as your `fusedform_base_extruder_0.def.json`, with the difference that it has
```
    "metadata": {
        "machine": "fusedform_doppia_base",
        "position": "0"
    },
```
- Rename `fusedform_base_extruder_1.def.json` to `fusedform_doppia_base_extruder_1.def.json` and change its metadata as follows (similar to fusedform_doppia_base_extruder_0):
```
    "metadata": {
        "machine": "fusedform_doppia_base",
        "position": "1"
    },
```
- Add a `fusedform_doppia_base.def.json`:
```
{
    "version": 2,
    "name": "fusedform_doppia_base",
    "inherits": "fusedform_base",
    "metadata": {
        "author": "Juan Blanco",
        "manufacturer": "Fused Form",
        "visible": false,
        "machine_extruder_trains": {
            "0": "fusedform_doppia_base_extruder_0",
            "1": "fusedform_doppia_base_extruder_1"
        },
        "quality_definition": "fusedform_base"
    }
}
```

- Change `fusedform_300_doppia.def.json`, `fusedform_600_doppia.def.json`, `fusedform_600plus_doppia.def.json`, and `fusedform_std_doppia.def.json` to inherit "fusedform_doppia_base" and not mention any machine extruder trains (since its now properly covered by the fusedform_doppia_base). So e.g. FF300 Doppia should look like this:
```
{
    "version": 2,
    "name": "FF300 Doppia",
    "inherits": "fusedform_doppia_base",
    "metadata": {
        "visible": true
    },

    "overrides": {
        .....
    }
```

With these changes the inheritance is now as expected. Again, my apologies for the extra time you had to put into it.

FusedForm:
Hi @konskarm, thanks for your support. 
Don´t worry regarding changes we had to make, the idea is to make a good and stable definition of the machines.

I made the changes suggested, let me know if they are OK. Thanks in advance for your support!

FusedForm:
Hi again @konskarm 
Ok, these ones work well running in Cura, I had some problems in Quality definitions. This would be the definitive one.
Thanks for your help, let me know your feedback.

konskarm:
I assume you mean that the qualities were not properly showing up. This was because `"quality_definition": "fusedform_base"` was missing from the "metadata" field in the `fusedform_doppia_base.def.json`. I see that you added it explicitly in the separate doppia machines, which also works fine. 

For the record, the `fusedform_doppia_base` could also omit the `"overrides"` field completely, since it's exactly the same as the overrides from `fusedform_base`, which it inherits. Nevertheless, it doesn't hurt to have it.

Upon inspecting and testing them again, I see that the doppia machines slice without a problem, so I am merging this. The machines will show up from the next release (4.9) onwards.

PR #8551, opened 2020-10-16
Fix not a group host message

nallath:
Cura-7360

### DIFF
--- a/plugins/UM3NetworkPrinting/src/UltimakerNetworkedPrinterOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/src/UltimakerNetworkedPrinterOutputDevice.py
@@ -85,6 +85,8 @@ class UltimakerNetworkedPrinterOutputDevice(NetworkedPrinterOutputDevice):
 
         self._timeout_time = 30
 
+        self._num_is_host_check_failed = 0
+
     @pyqtProperty(str, constant=True)
     def address(self) -> str:
         """The IP address of the printer."""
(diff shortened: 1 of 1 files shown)


konskarm (approved):

plugins/UM3NetworkPrinting/src/UltimakerNetworkedPrinterOutputDevice.py
```diff
@@ -293,8 +295,16 @@ def _updatePrinters(self, remote_printers: List[ClusterPrinterStatus]) -> None:
...
         if len(self._printers) < 1 and self.isConnected():
+            self._num_is_host_check_failed += 1
+        else:
+            self._num_is_host_check_failed = 0
+
+        # Since we request the status of the cluster itself way less frequent in the cloud, it can happen that a cloud
+        # printer reports having 0 printers (since they are offline!) but we haven't asked if the entire cluster is
+        # offline. (See CURA-7360)
+        # So by just counting a number of subsequent times that this has happened fixes the incorrect display.
+        if self._num_is_host_check_failed >= 6:
```
jellespijker, line 307: Where does the number 6 comes from?

Ghostkeeper: IIUC from the difference in the intervals between the checks. One check is done every 10 seconds and the other every 2 seconds. That would allow up to 5 failed checks in between the 10 second interval. Perhaps we should calculate this from the constants it's derived from.

PR #8552, opened 2020-10-16
Add ConsoleLogger as pre-loaded plug-in

Ghostkeeper:
This way it gets loaded only once on start-up, which prevents errors that the plug-in was already loaded.

This PR depends on https://github.com/Ultimaker/Uranium/pull/648 . Please merge at the same time.
Contributes to issue CURA-7501.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -756,7 +756,7 @@ class CuraApplication(QtApplication):
             self._plugin_registry.addPluginLocation(os.path.join(QtApplication.getInstallPrefix(), "lib" + suffix, "cura"))
         if not hasattr(sys, "frozen"):
             self._plugin_registry.addPluginLocation(os.path.join(os.path.abspath(os.path.dirname(__file__)), "..", "plugins"))
-            self._plugin_registry.loadPlugin("ConsoleLogger")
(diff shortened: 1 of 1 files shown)


fieldOfView:
Instead of blindly appending to a public list, wouldn't it be better to have a `addPreloadedPlugin` method that checks if the plugin isn't already in the list? Or make the list a set.

Ghostkeeper:
I originally had it as a set (due to the normal plug-in loading sequence using a set too), but figured it would be more powerful as a list, because that allows you to determine an order too if there is an interdependency between these pre-loaded plug-ins in the future.

Calling a public `addPreloadedPlugin` function is just as blind in my opinion as appending it to a public list, and then just having a list is the more brief solution and common in Python. Is there a reason why this list should be made private / write-only? It's not necessarily an implementation detail.

nallath:
De add PreloadedPlugin could ensure that it's not double

Ghostkeeper:
The `loadPlugin` function already ensures that it's not double, of course, and (still) posts an appropriate warning in the log if it is loaded double. The `addPreloadedPlugin` function can only ensure it's not double within the pre-loaded functions, not within all of the functions.

PR #8576, opened 2020-10-20
Add "preparing" printer status in the monitor page

konskarm:
To make sure that the printer status reflected in Cura is the same as the one in the Digital Factory when the printer is preparing for the print.

Contributes to CURA-7745

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
@@ -359,6 +359,10 @@ Item
                     {
                         return catalog.i18nc("@label:status", "Idle")
                     }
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
```diff
@@ -359,6 +359,10 @@ Item
                     {
                         return catalog.i18nc("@label:status", "Idle")
                     }
+                    if (!printer.activePrintJob && printer.state == "pre_print")
+                    {
+                        return catalog.i18nc("@label:status", "Preparing...")
```
Ghostkeeper, line 364: This re-uses an existing translation so it doesn't break the string freeze for 4.8, in case anyone is wondering.

Issue #8575, opened 2020-10-20
Post Processing Plugin | Pause at height | Repetier (@pause) 1mm to hight after continue


EGOiST1991:
Fixed it myself - see in the comments.
Bug: 
The Post Processing Plugin "Pause at height" has a bug with the Repetier Method where the Z-height is 1mm to high after continue.

**Application version**
4.7.1

**Platform**
Windows 10 | Intel Coree i5-4960k | Geforce 1070 

**Printer**
Anycubic i3 Mega

**Reproduction steps**
1. Open a File
2. Extensions
3. Post Processing Scripts
4. Pause at Height
5. Layer Number (not heigth)
6. Pause at Layer (any will do)
7.Method: Repetier

**Screenshot(s)**
https://i.imgur.com/Z1q1mXJ.png

**Actual results**
;TIME_ELAPSED:207.767029
;TYPE:CUSTOM
;added code by post processing
;script: PauseAtHeight.py
;current layer: 3
M83 ; switch to relative E values for any needed retraction
G1 F300 Z1.65 ; move up a millimeter to get out of the way
G1 F9000 X190 Y190
G1 F300 Z15.65
M84 E0
press @pause now and filament change ; Do the actual pause
@pause
G1 F300 Z1.65
G1 F9000 X0 Y0
G1 F2400 ; restore extrusion feedrate
M82
G92 E125.5039
G92 E125.5039
;LAYER:3
M106 S127.5
M204 S1800
M205 X8 Y8
;TYPE:WALL-INNER

**Expected results**
mycomments:
---- Here i am at 0.65mm
G1 F300 Z1.65 ; move up a millimeter to get out of the way
---- here the script from cura decides to go a mm up to get out of the way
G1 F9000 X190 Y190
G1 F300 Z15.65
----- 14mm up again
M84 E0
press @pause now and filament change ; Do the actual pause
@pause
----- Now im in the Pause
G1 F300 Z1.65
**----- Here i am 1 mm to hight now!!!! Here should stay 0.65mmF**
G1 F9000 X0 Y0
----- Jetzt bist du wieder bei x0 y0 aber 1mm zu hoch. Hier fehlt die Bewegung auf 0.65mm!!!!
G1 F2400 ; restore extrusion feedrate
M82

**Project file**
Any Gcode/STL will do - all will have the same result :) Test it yourself

**Log file**
No need, obvious bug in the Script

**Additional information**
[...]

PR #8578, opened 2020-10-20
Post Processing Plugin | Pause at height | Repetier (@pause) 1mm to high

EGOiST1991:
Bug:
The Post Processing Plugin "Pause at height" has a bug with the Repetier Method where the Z-height is 1mm to high after continue.
Discuss: https://github.com/Ultimaker/Cura/issues/8575

The wrong line starts here:
458: prepend_gcode += self.putValue(G = 1, Z = current_z + 1, F = 300) + "\n"
The correct line:
458: prepend_gcode += self.putValue(G = 1, Z = current_z, F = 300) + "\n"

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
+++ b/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
@@ -455,7 +455,7 @@ class PauseAtHeight(Script):
                         prepend_gcode += self.putValue(G = 1, E = -retraction_amount, F = 6000) + "\n"
 
                     #Move the head back
-                    prepend_gcode += self.putValue(G = 1, Z = current_z + 1, F = 300) + "\n"
+                    prepend_gcode += self.putValue(G = 1, Z = current_z, F = 300) + "\n"
                     prepend_gcode += self.putValue(G = 1, X = x, Y = y, F = 9000) + "\n"
                     if retraction_amount != 0:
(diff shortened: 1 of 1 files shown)


fieldOfView:
It will probably help getting this PR accepted if you add some more descriptive text what this PR does (in addition to a link to the original discussion).

EGOiST1991:
> 
> 
> It will probably help getting this PR accepted if you add some more descriptive text what this PR does (in addition to a link to the original discussion).

Thanks for your help. Is my edit okey?

tigercjn (approved):
Yes, this old line is causing the head to be 1mm high after it moves back after a pause. The fix to remove the extra 1mm will solve the problem. I have tested this now with the uploaded file as I was having the same problem using Octoprint, the same bug was causing the issue, the fix has now solved the problem, thanks heaps.

Ghostkeeper (approved):
I can indeed see this happening, verify that it is wrong behaviour and verify that this PR fixes the issue.

I'm merging this for the 4.9 release. Thanks for the analysis and the fix!

EGOiST1991:
This fixed it for Layers (what i am using)
But not for mm.
For mm:

The wrong line starts here:
`494: prepend_gcode += self.putValue(G = 1, Z = current_z + 1, F = 300) + "\n"`
The correct line:
`458: prepend_gcode += self.putValue(G = 1, Z = current_z , F = 300) + "\n"`

Its the same error differnt line for method "griffin"

PR #8601, opened 2020-10-23
CURA-7682: Print thin walls with libArachne

konskarm:
In this PR: 

- "Enable Variable Line Widening" has been removed in favor of "Print Thin Walls" that already existed
- Default/minimum/maximum values and descriptions of "Minimum Wall Line Width" and "Minimum Feature Size"have been updated to fit the purpose of the widening beading meta-strategy of libArachne

More specifically:

- **Minimum Feature Size**

  - The default value has been set to 0, because enabling Print Thin Walls should attempt to print all thin features of the model. In fact, these thin features that are smaller than the outer wall line width will be widened to the value of the outer wall line width.
  - The maximum value has been set to the Outer Wall Line Width. It doesn't make sense to try and ignore features thicker than the outer wall, since a wall can definitely fit there.

- **Minimum Wall Line Width**

  - The minimum value has been set to half the machine nozzle size, since attempting to print lines with a smaller width than that can be unreliable and problematic.
  - The minimum_value_warning has been set to the min_feature_size, since libArachne will anyway default the wall width of the thin features to the thickness of these features, if the minimum wall line width is set to be smaller than the feature size (it doesn't make sense to try and replace a thin feature with an even thinner wall).
  - The maximum_value_warning has been set to the outer wall line width, because using a greater value will seem weird (it will replace thin features with walls that are wider than the rest of the outer walls). We may have to turn this into a maximum_value in the future, but there is no reason to limit the users to that, yet.

Requires https://github.com/Ultimaker/CuraEngine/pull/1347.

CURA-7682


Ghostkeeper (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -1379,6 +1347,31 @@
...
+                    "limit_to_extruder": "wall_0_extruder_nr",
+                    "enabled": "fill_outline_gaps"
+                },
+                "min_bead_width":
+                {
+                    "label": "Minimum Wall Line Width",
+                    "description": "Width of the wall that will replace thin features (according to the Minimum Feature Size) of the model. If the Minimum Wall Line Width is thinner than the thickness of the feature, the wall will become as thick as the feature itself.",
+                    "unit": "mm",
+                    "value": "wall_line_width_0",
+                    "minimum_value": "0.5 * machine_nozzle_size",
```
Ghostkeeper, line 1368: Ah, I initially misread this, but Cura sets hard limits (`minimum_value` and `maximum_value`) only for physical or technical impossibilities. Examples of this are negative speeds (physical impossibility), layer height more than the printer height (technical constraint), or more than 20 gradual infill steps (technical impossibility due to integer overflows).
This limit here is just because we think it won't print well, so the user should be allowed to choose minimum wall line widths of lower than half a nozzle size, but should be warned when they do so.

Ghostkeeper (changes requested):

resources/definitions/fdmprinter.def.json
Ghostkeeper: From our discussion on Friday, the takeaway I intended to give is that I thought that half a line width was a good default for this setting. The minimum value should probably be 0.001mm, or maybe 0.01mm if rounding errors become a technical limitation.

Ghostkeeper:
Test failure seems to come from Master, not this branch.

resources/definitions/fdmprinter.def.json
konskarm: Let's make it into a minimum_value_warning then. It makes more sense than to strictly restrict the users.

PR #8616, opened 2020-10-25
Updated pt_BR strings for 4.8.

Patola:


Ghostkeeper:
Thanks! We'll review the translations in a week or so, when most of the other translations should also be in.

Devs, see our translation ticket for 4.8: CURA-7782.

resources/i18n/pt_BR/cura.po
```diff
@@ -1351,20 +1351,20 @@ msgstr[1] "Novas impressoras detectadas na sua conta Ultimaker"
 #, python-brace-format
 msgctxt "info:status Filled in with printer name and printer model."
 msgid "Adding printer {name} ({model}) from your account"
-msgstr ""
+msgstr "Adicionando impressora {name} {model} da sua conta"
```
Ghostkeeper, line 1354: Missing the brackets here. We'll add them back.

Ghostkeeper (approved):

PR #8620, opened 2020-10-26
Adding TwoTrees Bluer printer

WashingtonJunior:
Adding TwoTrees Bluer definition and platform files.


WashingtonJunior:
Adding TwoTrees Bluer definition and platform files.

nallath:
Please don't re-use the custom_extruder. This might cause unintended changes if that file is ever changed.

konskarm:
The printer slices fine and I do not see any issues with the added files, so I am merging this.

WashingtonJunior:
Thank you!!!

konskarm:
Note that the printer will be available in 4.9, since the 4.8 is scheduled for next week.

PR #8659, opened 2020-10-30
Skip running the tests for the old arrange algorithm

konskarm:
So that the output of the CI/CD testing step is not cluttered with deprecation warnings.

Done during TTT


jellespijker:
New tests need to be written at a later time.

PR #8661, opened 2020-10-30
CURA-7557 / CURA-7677 Rename Optimize Wall Printing Order to Order Inner Walls By Inset.

rburema:
See Engine-PR: https://github.com/Ultimaker/CuraEngine/pull/1350

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1273,8 +1273,8 @@
                 },
                 "optimize_wall_printing_order":
                 {
-                    "label": "Optimize Wall Printing Order",
-                    "description": "Optimize the order in which walls are printed so as to reduce the number of retractions and the distance travelled. Most parts will benefit from this being enabled but some may actually take longer so please compare the print time estimates with and without optimization. First layer is not optimized when choosing brim as build plate adhesion type.",
+                    "label": "Order Inner Walls By Inset",
+                    "description": "Order inner wall printing by inset-index, instead of by (hole) region.",
                     "type": "bool",
                     "default_value": false,
                     "settable_per_mesh": true


resources/definitions/fdmprinter.def.json
```diff
@@ -1273,8 +1273,8 @@
                 },
                 "optimize_wall_printing_order":
```
jellespijker, line 1274: I think leaving this called `optimize_` will create confusion later on. It will certainly be raised once or twice during a discussion.
However chancing this requires multiple additional changes so I'm not sure what will be the best approach.
Can you explain why you left it called `optimize_`?

resources/definitions/fdmprinter.def.json
```diff
@@ -1273,8 +1273,8 @@
...
                 {
-                    "label": "Optimize Wall Printing Order",
-                    "description": "Optimize the order in which walls are printed so as to reduce the number of retractions and the distance travelled. Most parts will benefit from this being enabled but some may actually take longer so please compare the print time estimates with and without optimization. First layer is not optimized when choosing brim as build plate adhesion type.",
+                    "label": "Order Inner Walls By Inset",
```
jellespijker, line 1276: Just to be clear, the term inset in this context isn't the same as our regular used inset term in the code.

jellespijker (commented):
Tests are failing on the arrange.

Unrelated to these changes.

jellespijker:
> Tests are failing on the arrange.
> 
> Unrelated to these changes.

FYI I just merged the PR which skips the arrange tests to master.

resources/definitions/fdmprinter.def.json
jellespijker: I think we should change it, since keeping the same name for something that will change the "original" behaviour could break backward compatibility if somebody imports an "old" 3mf file

rburema: Backwards compatibility. It still gives more or less the same behaviour as optimize ... it just isn't optimized any more. (Hence also the switch). I'll put it in Cura & Cookies.

resources/definitions/fdmprinter.def.json
rburema: Do you have a proposal for what it can be called?

resources/definitions/fdmprinter.def.json
rburema: Also it would require a version upgrade to actually rename the 'internal name'.

jellespijker: True indeed, lets discuss in Cura & Cookies

resources/definitions/fdmprinter.def.json
jellespijker: Sorry, this wasn't a change request. Inset is the proper name i.m.o. I just wanted to make sure I was on the same page with you.

PR #8677, opened 2020-11-02
Fix PickingPass glViewport ctypes.ArgumentError (crash on adding support blocker)

myfreeweb:
At least with Qt 5.15 Wayland backend and a HiDPI screen, width and height are floats, so adding a support blocker crashes.

### DIFF
--- a/cura/PickingPass.py
+++ b/cura/PickingPass.py
@@ -44,7 +44,7 @@ class PickingPass(RenderPass):
                 return
 
         width, height = self.getSize()
-        self._gl.glViewport(0, 0, width, height)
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
Python gives a warning about having to automatically cast, but doesn't crash on this, right?

Ghostkeeper:
Turns out this is already fixed by this commit, a week before this PR was made: https://github.com/Ultimaker/Uranium/commit/0abc64523009e8e8d711905ad914c6348a8992cf

All functions in this area require integers as per their typing, so it's best to use integers all the way down the line rather than use a cast at the bottom of the stream.

PR #8678, opened 2020-11-03
Gracefully handle exceptions in arrange

nallath:
CURA-7785


rburema (approved):

rburema:
Does more good than harm (as in, it doesn't crash), but OTOH, loading any subsequent 'loose' models confuses the heck out of the arrange algo. Merging this but flagging the ticket.

PR #8679, opened 2020-11-03
Added ideagen3D Sapphire Plus

Firedrops:


fieldOfView:
The definition references a png that is missing.

Another minor issue is that the definition files mix tabs and spaces. Please use spaces only.

konskarm:
Two things to be mentioned here:

- `has_machine_quality` should be set to false, so that it can use the global qualities. This field tells the printer to search for its own qualities, which are not provided in this PR.
- The buildplate is twisted by 90degrees 
image

Firedrops:
Apologies, and thanks for pointing those out. They have been fixed.

The original plan was to include our printing profiles (has_machine_quality), but we haven't been able to figure out how to convert .curaprofile to .inst.cfg, and other printers' .inst.cfgs are very barebones vs what we were expecting. For now, we'll just leave it false and send printer profiles manually.

Bed mesh has been fixed as well.

konskarm:
The machine quality is now fixed. The printer is using the global qualities.
Unfortunately, I still see the buildplate loading at 90degrees.

image

By the way, regarding the quality profiles, don't worry about it, take your time to work on them and you can add them in a separate PR in the future.

Firedrops:
Ah this is bizarre! I've been rotating it in Microsoft's 3D Builder. I guess that doesn't work.

I've rotate the file in Solidworks, and have tested it this time, and it appears to be working. However, the texture I've also supplied (and referenced) doesn't show up. Will that only work with .OBJ, and not .3MF meshes?

Ghostkeeper (approved):

Ghostkeeper:
The build plate is oriented correctly now, and it indeed has an embossed texture. I'm removing the platform texture from our source then.

The texturing extension of 3MF is not supported by Cura. But apparently this 3D model doesn't contain any texture coordinates of itself in the file either, so even if the textures would be supported it wouldn't show up. For textures, your only option is OBJ right now indeed.

PR #8681, opened 2020-11-03
Send None instead of empty string.

rburema:
5 min fix. Helping the data team out a bit. Should we merge this in 4.8?

### DIFF
--- a/plugins/SliceInfoPlugin/SliceInfo.py
+++ b/plugins/SliceInfoPlugin/SliceInfo.py
@@ -125,7 +125,8 @@ class SliceInfo(QObject, Extension):
             data["schema_version"] = 0
             data["cura_version"] = self._application.getVersion()
             data["cura_build_type"] = ApplicationMetadata.CuraBuildType
-            data["organization_id"] = user_profile.get("organization_id", None) if user_profile else None
+            org_id = user_profile.get("organization_id", None) if user_profile else None
+            data["organization_id"] = org_id if len(org_id) > 0 else None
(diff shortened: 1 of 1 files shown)


plugins/SliceInfoPlugin/SliceInfo.py
```diff
@@ -125,7 +125,8 @@ def _onWriteStarted(self, output_device):
             data["schema_version"] = 0
             data["cura_version"] = self._application.getVersion()
             data["cura_build_type"] = ApplicationMetadata.CuraBuildType
-            data["organization_id"] = user_profile.get("organization_id", None) if user_profile else None
+            org_id = user_profile.get("organization_id", None) if user_profile else None
+            data["organization_id"] = org_id if len(org_id) > 0 else None
```
nallath, line 129:
```suggestion
            data["organization_id"] = org_id if org_id else None
```

>>> bool("0") 
True
>>> bool("False") 
True
>>> bool("") 
False

nallath (approved):

konskarm:
It also needs a small change in https://github.com/Ultimaker/Cura/blob/f78fd9b3a44314729f5c1980f59c89310608bd9e/cura/OAuth2/AuthorizationHelpers.py#L130 so that it returns None.

rburema:
@konskarm Feel free to make the change ;-)

konskarm:
Done in https://github.com/Ultimaker/Cura/commit/881e4406a44ae45cb3ec54960dfc6f9b4d876078.

rburema:
@konskarm Thanks!

PR #8685, opened 2020-11-03
Also add wall_line_count when adding a infill mesh

### LINKED ISSUE #8589: Overlap modifier "infill mesh only" creates internal shells
**Application version**
4.7.1 official and Ultimaker_Cura-master.20201002014541.AppImage (nightly)

**Platform**
Linux Mint 20 - GTX 1050TI

**Printer**
Custom printer - (Printer is an Ender3 Modified to Direct Drive with BMG and e3dv6 clones.)

**Reproduction steps**
1. Load model and create a modifier
2. Set modifier to overlap with "infill mesh only" and change infill density.
3. Slice and check. 

**Screenshot(s)**

Cutting Mesh:
image

Infill Mesh Only:
image

**Actual results**

Where the modifier is overlapping there are shells before the new infill value is applied. However when changing to "cutting mesh", the infill is modified without creating new shells as expected, but it cuts the model in the external perimeters (I guess this is expected by using cutting mesh).

**Expected results**
No internal additional shells should be created when using "infill mesh only" when the idea is changing only the infill density.

**Project file**

[CuraProject.zip](https://github.com/Ultimaker/Cura/files/5418742/CuraProject.zip)

**Log file**
N/A

**Additional information**
seems like this [PR](https://github.com/Ultimaker/Cura/pull/6574) is related

nallath:
CURA-7793

Fixes #8589

### DIFF
--- a/plugins/PerObjectSettingsTool/PerObjectSettingsTool.py
+++ b/plugins/PerObjectSettingsTool/PerObjectSettingsTool.py
@@ -103,17 +103,20 @@ class PerObjectSettingsTool(Tool):
                     new_instance.resetState()  # Ensure that the state is not seen as a user state.
                     settings.addInstance(new_instance)
 
-        for property_key in ["top_bottom_thickness", "wall_thickness"]:
+        for property_key in ["top_bottom_thickness", "wall_thickness", "wall_line_count"]:
             if mesh_type == "infill_mesh":
                 if settings.getInstance(property_key) is None:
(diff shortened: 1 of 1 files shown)


rburema (approved):

nallath:
That's odd. It does for me. When I add an infill mesh only it now shows three settings:

image

rburema:
> That's odd. It does for me. When I add an infill mesh only it now shows three settings:

Can you try to load the model I attached to the ticket in jira?

PR #8689, opened 2020-11-04
Remove the reset of one_at_a_time on extruder switch

### LINKED ISSUE #8682: [4.8 beta] "Discard changes" causes print sequence go blank (settings updated to match the availablity of extruders)
**Application version**
4.8 beta

**Printer**
At least Custom fff, Anycubic kossel linear plus, wanhao duplicatori3.
Only with my old settings-profiles created in 4.3 (or possibly even older but imported and exported from 4.3)
The printer profiles are created fresh in 4.8

**Reproduction steps**
1. open attached project
2. change any setting and then hit discard current changes

**Screenshot(s)**
Discard
Discarded
discarded value

**Project file**
[Settings updated.zip](https://github.com/Ultimaker/Cura/files/5481762/Settings.updated.zip)

**Log file**
[stderr.log](https://github.com/Ultimaker/Cura/files/5481770/stderr.log)

**Additional information**
Apart from this, the profiles work as they should. If it's simply something wrong with my profiles, I can edit my files if I know what to change.
After uploading this, I changed quality type fast to normal in all the profile files, it made no difference.

nallath:
Instead solve it by ensuring that the order of the models isn't
changed if the setting is disabled

Fixes #8682

CURA-7827

### DIFF
--- a/cura/Settings/GlobalStack.py
+++ b/cura/Settings/GlobalStack.py
@@ -301,6 +301,16 @@ class GlobalStack(CuraContainerStack):
                 return False
         return True
 
+    def isInOneAtATimeMode(self) -> bool:
+        """
+        Gets whether the one-at-a-time mode is activated, provided that there is only one active extruder
+        """
+        one_at_a_time_mode_is_active = self.getProperty("print_sequence", "value") == "one_at_a_time"
+        extruders_enabled_count = self.definitionChanges.getProperty("extruders_enabled_count", "value")
+        if extruders_enabled_count:
+            one_at_a_time_mode_is_active &= extruders_enabled_count == 1
+        return one_at_a_time_mode_is_active
+
     def getHeadAndFansCoordinates(self):
         return self.getProperty("machine_head_with_fans_polygon", "value")
 
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -162,7 +162,8 @@ class BuildVolume(SceneNode):
             self.rebuild()
 
             self._scene_objects = new_scene_objects
-            self._onSettingPropertyChanged("print_sequence", "value")  # Create fake event, so right settings are triggered.
+            self._onSettingPropertyChanged("print_sequence", "value")
+            self._onSettingPropertyChanged("extruders_enabled_count", "value")  # Create fake event, so right settings are triggered.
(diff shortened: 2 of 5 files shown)


konskarm (approved):
The problem described in https://github.com/Ultimaker/Cura/issues/8682 is fixed and also the original problem that this issue was trying to fix (CURA-6914) is also not reappearing.

konskarm:
Turns out that it doesn't fix the original issue completely. 
**Reproduction steps:**

1. load model
2. disable extruder 2
3. change print sequence to one-at-a-time
4. re-enable extruder 2

image

The convexHullDecorator is not resetted.

konskarm:
The problem described above is fixed in https://github.com/Ultimaker/Cura/pull/8689/commits/bdfccd8beace6bcda7b2e7c299ec4b4575844a9d. 

**Note:** If you enable one-at-a-time and then enable the second extruder, the global stack will still report that it has one-at-a-time print sequence, despite the fact that it will print everything all at once, since the setting is not resetted to all-at-once (compared what the original fix https://github.com/Ultimaker/Cura/commit/9266c753c30868ca3b6badda41da1b0f6d9b9857 was attempting to do.

cura/Settings/GlobalStack.py
```diff
@@ -301,6 +301,16 @@ def isValid(self) -> bool:
                 return False
         return True
 
+    def isInOneAtATimeMode(self) -> bool:
+        """
+        Gets whether the one-at-a-time mode is activated, provided that there is only one active extruder
+        """
+        one_at_a_time_mode_is_active = self.getProperty("print_sequence", "value") == "one_at_a_time"
+        extruders_enabled_count = self.definitionChanges.getProperty("extruders_enabled_count", "value")
```
nallath, line 309: Isn't it just easier to look at the enabled property? Now you are effectively re-implementing the enabled check.

konskarm: Indeed, I just figured that it is clearer using the extruders enable count

konskarm:
As we discussed with @nallath, I am closing this since it doesn't seem like a viable solution. Another solution is coming through https://github.com/Ultimaker/Cura/pull/8698.

PR #8698, opened 2020-11-05
CURA-7827: Correct print sequence when there are more than one enabled extruders

konskarm:
Fixes https://github.com/Ultimaker/Cura/issues/8682 while still maintaining the correct behavior for CURA-6914.

CURA-7827


cura/Settings/MachineManager.py
```diff
@@ -862,6 +858,19 @@ def correctExtruderSettings(self) -> None:
                 title = catalog.i18nc("@info:title", "Settings updated"))
             caution_message.show()
 
+    def _correctPrintSequence(self) -> None:
```
Ghostkeeper, line 861: It would help to document why Print Sequence is a special case here. Won't this also happen with other similar settings? Like with all enum-settings, or with all global settings or something?

cura/Settings/MachineManager.py
```diff
@@ -862,6 +858,19 @@ def correctExtruderSettings(self) -> None:
...
+        """Resets the Print Sequence setting when there are more than one enabled extruders."""
+
+        if self._global_container_stack is None:
```
jellespijker, line 864: why not change it to:
```py
if self._global_container_stack is None and  self.numberExtrudersEnabled < 2:
    return
```
Then you can remove  the other `if` statement

jellespijker (changes requested):

cura/Settings/MachineManager.py
konskarm: Better?

jellespijker (approved):

cura/Settings/MachineManager.py
jellespijker: Fine by me.
Clear documentation through out the code

jellespijker (approved):

jellespijker:
Testing to be added at the next TTT session see CURA-7833

PR #8731, opened 2020-11-11
Fix tooltips on actionbuttons

fieldOfView:
This PR fixes tooltips on ActionButtons such as the buttons that constitute the toolbar and buttons in the ActionPanel etc. The tooltips were broken in 5f574d4b64ab0114b19b843acbc70e006bbe69d1, and this PR applies a fix on top of that. I am not entirely sure what the original commit was meant to fix, so there may be a more direct way to fix the tooltips than this roundabout fix.

### DIFF
--- a/resources/qml/ToolTip.qml
+++ b/resources/qml/ToolTip.qml
@@ -34,6 +34,9 @@ ToolTip
         NumberAnimation { duration: 100; }
     }
 
+    onAboutToShow: show()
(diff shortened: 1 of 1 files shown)


nallath:
They were supposed to fix the tooltips in the machine list if you have a cloud connected printer. I'll have a look if they still work.

nallath:
Still works!

nallath (approved):

PR #8741, opened 2020-11-12
Add warning icon to show what extruder makes the config Not Supported

fieldOfView:
This PR adds icons to the Configuration Menu to show which extruder(s) is/are causing the configuration to be "Not Supported" in the quality dropdown. It can be very unclear why a configuration is unsupported, and where to fix this. The warning icons in this PR bring the attention to the configuration menu, where the problem can be fixed.

image
In the example above, the left extruder causes the "Not Supported" state, and the right extruder causes slicing to be outright inhibited because the material and printcore type are not compatible.

This functionality has been extensively tested in the Sidebar GUI plugin since Cura 4.0.


Ghostkeeper:
There is some ongoing discussion about this since the UX designers just made something else that was supposed to go there, apparently. They're considering ways in which to incorporate both.

fieldOfView:
Perhaps if the UX designers can share their "something else", I can work on a solution with them.

fieldOfView:
I'm going to go out on a limb here and guess that anything that Lorenzo is working on is not going to make it into Cura 4.9. So I would say this is still an improvement for Cura 4.9 until his redesign happens.

fieldOfView:
I guess not. There's no use then to keep this open.

Ghostkeeper:
@lorenzoromagnoli Is there any plan on your end regarding the change you made to improve this behaviour?

fieldOfView:
Since I have seen no action in the past two releases towards a better design, I'm reopening this. It is still an improvement for 4.11, as it would have been for 4.8, 4.9 and 4.10 in the mean time.

fieldOfView:
It currently looks like this:
image

fieldOfView:
I have updated the PR to use the UM.StatusIcon that was introduced in Cura 4.11:
image

I still think it is a good idea to show the user which extruder is causing the profile to go "Not supported".

fieldOfView:
:cake:

This PR is now a year old. Users still don't really have a clue why their configuration is unsupported. This situation could have been improved since Cura 4.8. 

I'll fix the merge conflicts one last time.

lorenzoromagnoli:
Sorry for the enormous delay @fieldOfView, I think I'm the main person responsible for this enormous wait. 
I had some weird notification settings in Github and I didn't receive the update about you replacing the icons in the PR. 

Since indeed I think this is a valuable contribution I'd say please fix the conflict and I'll push the team to merge this. 
Ideally I'd like those warning badges to show up in the top right corner of the extruder icon, similarly to how we show the printer connection statuses, and other warnings on the printer interface. 

Screen Shot 2021-11-16 at 14 30 22

Would you be able to adjust that? Else I'll have a look and maybe fix the position myself.

fieldOfView:
Do you want the "badges" in the top right (as in your mockup), or in the bottom right (like the printer connection statusses)? Or should the printer connection statusses move to the top-right of the icon, like most of the other badges (post processing active, updates available)?

lorenzoromagnoli:
I think I'd like to standardise on this pattern : 

- badges requesting attention (warnings, error, notifications ) on top 
- status badges, which are there to add a meaning to the icon could stay on the bottom

Which practically means that the badge next to the icon could stay on the bottom, but the one on the extruder should follow the same positioning as the post processing and notifications button.

Does it makes sense to you?

fieldOfView:
When you explain it, it makes sense. When you don't, it looks inconsistent. But that may be just me.

fieldOfView:
For good measure, I have added a tooltip that explains the status of the extruder/configuration:
image

The sizing of the status icon is a bit funky, to make the badge size match the machine connection status badge size.

lorenzoromagnoli:
I believe that's because the UM Status icon uses the default size of icon (24px- 2em). 

For those small badges the intention was to use the small icons sizes designed at 12px (1em) https://github.com/Ultimaker/Cura/tree/master/resources/themes/cura-light/icons/low

fieldOfView:
I'm using the reusable StatusIcon qml component. I would assumed that uses the correct icon.
https://github.com/Ultimaker/Uranium/blob/master/UM/Qt/qml/UM/StatusIcon.qml

fieldOfView:
So do you want me to create new icons in the "low" folder, or should I use the StatusIcon qml component like I do now?

lorenzoromagnoli:
I just realised that some default size icons ended up in the low size folder. I'm going to try fix that first in the meantime I ask the developers to answer your previous question I'm not sure what the preference would be in this case.

lorenzoromagnoli:
I prepared these 2 PR to move those warning icons out of the low folder and to the default folder where they belong.
https://github.com/Ultimaker/Uranium/pull/752
https://github.com/Ultimaker/Cura/pull/10857

as part of that I added the actual low resolution icons that you should use in the low folder. 

resources/themes/cura-light/icons/low/BadgeWarning.svg
resources/themes/cura-light/icons/low/BadgeCross.svg
resources/themes/cura-light/icons/low/BadgeCheck.svg

if you could use those instead of the reusable component you will see an improvement in readability.

fieldOfView:
For consistency, I have removed the colors from the SVGs @lorenzoromagnoli provided, and used the naming scheme that conforms with the CloudBadge icon that is used in the connection status. The color of the bade is now themeable using the theme.json file.

In a followup PR, the badges in this PR and in the printer connection status should be abstracted into a reusable component for consistency and to reduce code duplication.

fieldOfView:
image

Ready for review.

resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
```diff
@@ -63,6 +69,109 @@ Cura.ExpandablePopup
                         anchors.verticalCenter: parent.verticalCenter
                     }
 
+                    MouseArea // Connection status tooltip hover area
+                    {
+                        id: tooltipHoverArea
+                        anchors.fill: parent
+                        hoverEnabled: true //getConnectionStatusMessage() !== ""
```
Ghostkeeper, line 76: This was some old code? Maybe the commented code should be removed. Or maybe you want to replace it with this:
```suggestion
                        hoverEnabled: tooltip.text !== ""
```
Or this:
```suggestion
                        hoverEnabled: badge.source !== ""
```

After all, you only want the tooltip to appear when there is an icon.

I'd be fine with either of these 3 suggestions.

Ghostkeeper (changes requested):
One small remark. I'll proceed with testing in the meanwhile.

resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
fieldOfView: Indeed, this is leftover code from copy/pasting. Thanks for spotting it.

fieldOfView: Fixed in c805aac

Ghostkeeper (approved):

Ghostkeeper:
I found a bit of a bug here. When the extruder is disabled, the main icon fades but the badge doesn't. It looks a bit weird to me.

image

Do you agree that it should fade too?

fieldOfView:
Good one. Does it still block slicing or cause the profile to become "Unsupported" when it is disabled? If not, it should not show a badge at all.

Fading the badge icon could make it look ugly (with the underlying reel icon showing through). 

I'll have a quick look when my computer is done rendering what it is currently rendering.

Ghostkeeper:
It doesn't block slicing if the extruder with invalid configuration is disabled. So indeed maybe it shouldn't show the badge at all then.

Ghostkeeper:
I completed testing and that was the only issue I found with it.

Thanks for the useful feature. Nice QOL thing. It's not very nice how this went with this PR lingering for so long, but I'm glad we're pushing this through now.

Ghostkeeper (approved):
Works correctly now!

fieldOfView:
Fortunately the PR aged like wine, not like milk...

PR #8788, opened 2020-11-17
Add Z-Hop and filament retraction to Timelapse

racerxdl:
This adds a z-hop and filament retraction for timelapse post processing.

On my Ender 3 Pro that reduces to zero the stringing ocurring when parking the head.

With the original script:
image

With the modified script:
image

The test settings was:
image

Tested on Cura 4.8

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/TimeLapse.py
+++ b/plugins/PostProcessingPlugin/scripts/TimeLapse.py
@@ -73,7 +73,7 @@ class TimeLapse(Script):
                     "description": "Filament retraction distance for camera trigger.",
                     "unit": "mm",
                     "type": "int",
-                    "default_value": 1
+                    "default_value": 0
                 },
                 "zhop":
                 {
--- a/plugins/PostProcessingPlugin/scripts/TimeLapse.py
+++ b/plugins/PostProcessingPlugin/scripts/TimeLapse.py
@@ -81,7 +81,7 @@ class TimeLapse(Script):
                     "description": "Z-hop length before parking",
                     "unit": "mm",
                     "type": "float",
-                    "default_value": 0.8
+                    "default_value": 0
                 }
             }
         }"""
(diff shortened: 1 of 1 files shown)


plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -66,6 +66,22 @@ def getSettingDataString(self):
...
                     "default_value": 9000,
                     "enabled": "park_print_head"
+                },
+                "retract":
+                {
+                    "label": "Retraction Distance",
+                    "description": "Filament retraction distance for camera trigger.",
+                    "unit": "mm",
+                    "type": "int",
+                    "default_value": 1
```
nallath, line 76:
```suggestion
                    "default_value": 0
```

In order to keep the script behaving the same if the user didn't change anything, this should be 0. We try to ensure that scripts like this stay the same.

Another option would be to create an upgrade script that checks if someone had the script loaded (and then set this value to 0), but that's a bit more complicated.

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -66,6 +66,22 @@ def getSettingDataString(self):
...
+                },
+                "zhop":
+                {
+                    "label": "Z-Hop when parking",
```
nallath, line 80:
```suggestion
                    "label": "Z-Hop Height When Parking",
```

The "normal" z-hop setting is called "Z Hop Height", so it makes sense to use a similar naming.

plugins/PostProcessingPlugin/scripts/TimeLapse.py
```diff
@@ -66,6 +66,22 @@ def getSettingDataString(self):
...
+                    "description": "Z-hop length before parking",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 0.8
```
nallath, line 84:
```suggestion
                    "default_value": 0
```

Same as before; Should be zero or an upgrade script needs to be made.

nallath (changes requested):

PR #8802, opened 2020-11-20
[ CURA-7871] Invert colors at lowest print height.

### LINKED ISSUE #8735: Color for contact zone to platform
In the past, I asked that the area of the object under the bed be colored in a different color as it is used to cut objects. This has been implemented in version 4.8.
Along with this they said they adjusted the height where the bed is drawn.

The current problem is: If I place an object on the bed, I want to see all the parts that are in contact with it for cases where I don't want to cut objects.

THIS IS VERY IMPORTANT when we rotate objects and use automatic functionalities to deposit them on the bed ensuring that it is not slightly tilted resulting in layers in the air and a failed print.

https://github.com/Ultimaker/Cura/issues/8469

Perhaps it would be good to have a darker blue or another color to indicate the parts of the objects that are in contact with the bed.
Do not confuse the red zone indicating overhang.
I leave you a riddle so you can see how confusing it can be. What parts touch the bed? 1,2,3?
imagen

Only 2 and 3. apparently...

imagen

rburema:
Instead of inverting them if below 0 (build-plate).

Fixes #8735


nallath (approved):

Ghostkeeper (approved):

plugins/SolidView/SolidView.py
```diff
@@ -111,7 +120,18 @@ def _onPropertyChanged(self, key: str, property_name: str) -> None:
...
+        return 90.0
+
+    def _retrieveLowestPrintHeight(self) -> float:
+        min_height = 0.0
+        for extruder in Application.getInstance().getExtruderManager().getActiveExtruderStacks():
+            init_layer_height = extruder.getProperty("layer_height_0", "value")
+            tolerance_setting = extruder.getProperty("slicing_tolerance", "value")
+            if tolerance_setting == "middle":
+                init_layer_height /= 2.0
+            min_height = max(min_height, init_layer_height / 2.0)
```
Ghostkeeper, line 133: This calculation is incorrect. It's dividing the initial layer height by 2 twice. As a result, the height at which it shows inverted colours is a quarter of the initial layer height by default, rather than half.

To test it I used a staircase model with 10mm steps, and an initial layer height of 21mm. It showed that the base step of the stair would be printed as normal.
image

While actually it was missing the first step since the cross section is taken at 10.5mm height:
image

So I'll remove this second division by 2 here. I hope you'll agree @rburema .

rburema: If you think it then works for all tolerances, please do! :-)

PR #8806, opened 2020-11-20
New Printer

venkatkamesh:


resources/definitions/diy220.def.json
```diff
@@ -0,0 +1,89 @@
...
+        "preferred_variant_name": "0.4 mm",
+        "machine_extruder_trains":
+        {
+            "0": "diy220_extruder_0"
+        }
+    },
+    "overrides": {
+        "machine_name": { "default_value": "Diytech 220" },
+        "machine_start_gcode" : {
+            "value": "\"\"  if machine_gcode_flavor == \"UltiGCode\" else \"G21 ;metric values\\nG90 ;absolute positioning\\nM82 ;set extruder to absolute mode\\nM107 ;start with the fan off\\nG28 Z0 ;move Z to bottom endstops\\nG28 X0 Y0 ;move X/Y to endstops\\nG1 X15 Y0 F4000 ;move X/Y to front of printer\\nG1 Z15.0 F9000 ;move the platform to 15mm\\nG92 E0 ;zero the extruded length\\nG1 F200 E50 ;extrude 10 mm of feed stock\\nG92 E0 ;zero the extruded length again\\nG1 F9000\\n;Put printing message on LCD screen\\nM117 Printing...\""
```
nallath, line 25: Does the machine support UltiGcode? If not, why have two different start gcodes?

resources/definitions/diy220.def.json
```diff
@@ -0,0 +1,89 @@
...
+    "version": 2,
+    "name": "Diytech 220",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Venkatkamesh",
+        "manufacturer": "Sri Vignan Technologies",
+        "weight": 3,
+        "file_formats": "text/x-gcode",
+        "platform": "ultimaker3_platform.obj",
```
nallath, line 11: Uh? Why is this machine using the same model as the Ultimaker?

resources/definitions/diy220.def.json
```diff
@@ -0,0 +1,89 @@
...
+            "value": "48"
+        },
+        "machine_use_extruder_offset_to_offset_coords": {
+            "default_value": true
+        },
+        "machine_gcode_flavor": {
+            "default_value": "Marlin"
+        },
+        "machine_disallowed_areas": {
+            "default_value": [
```
nallath, line 64: Did you just copy the values from the Ultimaker2? Are you sure that these disallowed areas are needed?

resources/definitions/kosher.def.json
```diff
@@ -0,0 +1,61 @@
...
+        "speed_travel": {"maximum_value": "300", "value": "200"},
+        "optimize_wall_printing_order": { "value": "True" },
+        "material_diameter": { "default_value": 1.75},
+        "retraction_amount": {"default_value": 6.5},
+        "retraction_speed": { "default_value": 30},
+        
+		"adhesion_type": { "default_value": "skirt" },
+        "machine_gcode_flavor": { "default_value": "Marlin"},
+	    "ironing_enabled":{"label": "Enable Ironing",
+                    "description": "Go over the top surface one additional time, but this time extruding very little material. This is meant to melt the plastic on top further, creating a smoother surface. The pressure in the nozzle chamber is kept high so that the creases in the surface are filled with material.",
```
nallath, line 48: Why override the description with the same value?

resources/definitions/kosher.def.json
```diff
@@ -0,0 +1,61 @@
...
+                    "type": "bool", "default_value": true, "limit_to_extruder": "top_bottom_extruder_nr", "settable_per_mesh": true },
+        "machine_disallowed_areas": {
+            "default_value": [
```
nallath, line 51: Same as before; Are you sure that these disallowed areas are correct?

resources/definitions/kosher_duplication.def.json
```diff
@@ -0,0 +1,53 @@
...
+        "gantry_height": {"value": "48"},
+        "line_width": {"value": "machine_nozzle_size"},
+        "speed_travel": {"maximum_value": "300", "value": "200"},
+        "optimize_wall_printing_order": { "value": "True" },
+        "material_diameter": { "default_value": 1.75},
+        "retraction_amount": {"default_value": 6.5},
+        "retraction_speed": { "default_value": 30},
+        
+		"adhesion_type": { "default_value": "skirt" },
+        "ironing_enabled":{"label": "Enable Ironing",
```
nallath, line 45: Same as before; Don't override settings with the same value. THis could cause issues in the future.

resources/definitions/diy220.def.json
venkatkamesh: Using ultimaker board 
In title itself we defined as did(do it yourself)  model
Functions are same as ultimaker

resources/definitions/diy220.def.json
venkatkamesh: Please make a single comment on same def 
This definition is since 2.7 update
Thank you

resources/definitions/kosher.def.json
venkatkamesh: Thank you 
We will update it

resources/definitions/kosher_duplication.def.json
venkatkamesh: Could you please point it out?

resources/definitions/diy220.def.json
nallath: Huh? Using Ultimaker board? Is this an Ultimaker clone? 

I've also not been able to find the printer available anywhere. Is it even a printer that is sold somewhere?

resources/definitions/diy220.def.json
venkatkamesh: Yes

resources/definitions/diy220.def.json
venkatkamesh: We have sold diy models to educational institutions around 60qty
Between sorry it's mks board using marlin architecture 
Other are industrial models sold for industries around our state.

venkatkamesh:
@nallath 
we had found our mistakes and solved. we appreciate on spending time on our definitions
 Thank you so much

resources/definitions/kosher_duplication.def.json
Ghostkeeper: You are overriding the label of a setting. If the label of this setting is changed in fdmprinter, this one won't be translated any more. I'd say this is not what you aimed for. There is no reason why you'd change the label of a setting for a specific printer, I'd think.

Overriding the value of a setting with the same value as what it was before can be done fairly harmlessly, but it's also useless since it just ends up on the same setting value then.

venkatkamesh (commented):
Dear friend , it was mistake from us
we have directly copied the default def code of ironing option without completely reading the description,
and yesterday we have finally cleaned up all the code.

Thank you

resources/definitions/kosher_duplication.def.json
```diff
@@ -0,0 +1,51 @@
+{
+    "version": 2,
+    "name": "Kosher_Duplication",
```
Ghostkeeper, line 3: Is there an underscore in the printer name? Seems a bit weird.
I can't find any information on this printer online to verify it against. I'd guess it's just a duplication mode that the printer can do with 2 print heads. Then maybe just a space is what you intended.

resources/definitions/kosher_mirror.def.json
```diff
@@ -0,0 +1,51 @@
+{
+    "version": 2,
+    "name": "Kosher_Mirror",
```
Ghostkeeper, line 3: Another underscore.

Ghostkeeper (changes requested):
These are not blocking changes. I'll test this printer out and make the requested changes myself if it works otherwise.

resources/definitions/arjun_duplication.def.json
```diff
@@ -0,0 +1,49 @@
...
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Venkat Kamesh",
+        "manufacturer": "Sri Vignan Technologies",
+        "weight": 3,
+        "file_formats": "text/x-gcode",
+        "has_material": true,
+        "has_variants": true,
+        "preferred_variant_name": "0.4 mm",
```
Ghostkeeper, line 13: Found some mistakes when testing. This preferred variant name doesn't exist for this printer, so it chooses an arbitrary one. I think you meant to use `0.4 mm Nozzle` so I'll make a correction there.

Ghostkeeper:
With those corrections it works! I think it's pretty safe to say what you intended. Sorry it took so long to review. But it's in for the 4.9 release now.

Issue #8821, opened 2020-11-23
Cannot switch to Czech language in Cura 4.8


sustmi:
**Application version**
Ultimaker_Cura-4.8.0-amd64.exe

**Platform**
Windows 10 (64bit)
(also reproducible on Fedora Linux 33 (64bit) with Ultimaker_Cura-4.8.0.AppImage)

**Printer**
I do not think that it is relevant but I have Creality Ender 3.

**Reproduction steps**
1. Install Cura 4.8
2. Start Cura
3. Go to "Preferences" > "Configure Cura..."
4. In the "General" > "Interface" section open the "Language" dropdown

**Screenshot(s)**
![cura](https://user-images.githubusercontent.com/885946/99937352-6edb3400-2d5d-11eb-944a-076e6e1d4294.png)

**Actual results**
The Czech language ("Čeština" in Czech) is missing in the selection even when you can see on the screenshot that my interface is in Czech.
If I switch to "English" and restart Cura, I am not able to switch back to Czech any more.

**Expected results**
I should be able to select Czech language.
In Cura 4.7 I was able to select Czech language and Cura was still in Czech after update to Cura 4.8, however the option is missing now.

**Project file**
Not relevant

**Log file**
[stderr.log](https://github.com/Ultimaker/Cura/files/5581487/stderr.log)
[stdout.log](https://github.com/Ultimaker/Cura/files/5581492/stdout.log)

**Additional info**
In one user group that I am part of a member that did clean installation of Cura 4.8 was not able to switch to Czech and thought that Cura 4.8 does not support Czech language anymore. He then recommended Prusa Slicer instead of Cura to a newcomer without good knowledge of English language for this specific reason.

sustmi:
Oh, I can see that it was a deliberate decision: https://github.com/Ultimaker/Cura/commit/0821b1bdb1e548eb8c484b69a10bcf852c23ad7b

I will try to add the missing translations.

nallath:
The Czech translation is a comunity translation. Since this is work done on a voluenteer basis, we can't ensure that it's there (unlike the languages that we officially support and pay translators for).

sustmi:
Thank you for your reply @nallath . 

I did not really know how you manage to translate Cura to Czech. Thank you for the insight.

How should I understand the fact that you closed the issue?
Would be my contribution welcome? If I sent you a pull request with the missing translation will you re-enable Czech language in Cura?

[...]

PR #8842, opened 2020-11-27
Add Czech translations for Cura 4.8

sustmi:
Add missing Czech translations and re-enable the option to switch to Czech language.
Fixes https://github.com/Ultimaker/Cura/issues/8821 .


resources/i18n/cs_CZ/cura.po
```diff
@@ -1440,15 +1440,23 @@ msgid_plural ""
 "You are about to remove {0} printers from Cura. This action cannot be undone.\n"
 "Are you sure you want to continue?"
 msgstr[0] ""
+"Chystáte se odebrat {0} tiskárnu from Cury. Tuto akci nelze vrátit zpět.\n"
```
konvoj, line 1443: "Chystáte se odebrat {0} tiskárnu z Cury. Tuto akci nelze vrátit zpět.\n"

resources/i18n/cs_CZ/cura.po
```diff
@@ -1440,15 +1440,23 @@ msgid_plural ""
...
+"Doopravdy chcete pokračovat?"
 msgstr[1] ""
+"Chystáte se odebrat {0} tiskárny from Cury. Tuto akci nelze vrátit zpět.\n"
```
konvoj, line 1446: "Chystáte se odebrat {0} tiskárnu from Cury. Tuto akci nelze vrátit zpět.\n"

resources/i18n/cs_CZ/cura.po
```diff
@@ -1440,15 +1440,23 @@ msgid_plural ""
...
+"Doopravdy chcete pokračovat?"
 msgstr[2] ""
+"Chystáte se odebrat {0} tiskáren from Cury. Tuto akci nelze vrátit zpět.\n"
```
konvoj, line 1449: "Chystáte se odebrat {0} tiskárnu from Cury. Tuto akci nelze vrátit zpět.\n"

resources/i18n/cs_CZ/cura.po
sustmi: Oops. Fixed. Thank you for noticing. :+1:

Ghostkeeper:
This was laid to rest for a while. Sorry about that.
The translation was submitted after the release of 4.8. It would update the translations for the next release, and that is fine. However for the next release we'll again have new strings to translate.

Would it be okay if I e-mail you with an updated translation template when we have a string freeze for the next release? It'll be a template with ~50 or so new texts that need to be translated for a new release. That's the only way we can ensure a complete translation for the next release.
I don't have a date for the next release yet, but it'll be at least a month yet before we release the beta I think.

sustmi:
Yeah, that would be OK.

sustmi:
I think you already have my e-mail address, don't you?
You can even just let me know when the translations are freezed and tell me in which branch/commit I can find the strings to translate.

Ghostkeeper:
Yeah, I grabbed your e-mail address from the commit that you made here. The e-mail address I found ends with `@****rum.cz` (it's public knowledge since your commit is public, but I censored it to not make it any easier on any bots that might look around here).

I expect that we'll freeze the strings for 4.9 in a month or so. You'll get complete instructions then as to which branch to look at and which files to fill in!

PR #8884, opened 2020-12-04
Remove use of `support_tree_enable` setting in UM 2+ Connect profiles

fieldOfView:
This PR fixes the UM 2+ Connect profiles that were published in 1927f78d5a1124f3a3dda8a90e170c1714ac70c1 containing references to a setting that no longer exists. The `support_tree_enable` setting was removed in Cura 4.7, and instead uses the `support_enable` and `support_structure` settings.


Ghostkeeper:
Argh, sorry, I already remove these myself when I saw your message on that commit. I didn't see this pull request yet then.

I removed the usage here: https://github.com/Ultimaker/Cura/commit/000d7e90f511f1baae5b70bbaebe17aed6f8224c so this change is no longer necessary then.

I've changed it slightly differently. I've made tree support the exception rather than normal support, in case we ever get more types of support structure.

fieldOfView:
I'm happy as long as it is fixed...

PR #8890, opened 2020-12-06
Create Thumbnails in gcode via PostProcessingPlugin

Gravedigger7789:
Added the ability to have thumbnails embedded into gcode similar to other slicers.

Screenshot 2020-12-06 134342

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/CreateThumbnail.py
+++ b/plugins/PostProcessingPlugin/scripts/CreateThumbnail.py
@@ -1,0 +1,109 @@
+import base64
+import re
+
+from UM.Logger import Logger
+from cura.Snapshot import Snapshot
+from PyQt5.QtCore import QByteArray, QIODevice, QBuffer
+
+from ..Script import Script
+
+
+class CreateThumbnail(Script):
+    def __init__(self):
+        super().__init__()
+
+    def _createSnapshot(self, width, height):
+        Logger.log("d", "Creating thumbnail image...")
+        try:
+            return Snapshot.snapshot(width, height)
+        except Exception:
+            Logger.logException("w", "Failed to create snapshot image")
+
+    def _encodeSnapshot(self, snapshot):
+        Logger.log("d", "Encoding thumbnail image...")
+        try:
+            thumbnail_buffer = QBuffer()
+            thumbnail_buffer.open(QBuffer.ReadWrite)
+            thumbnail_image = snapshot
+            thumbnail_image.save(thumbnail_buffer, "PNG")
+            base64_bytes = base64.b64encode(thumbnail_buffer.data())
+            base64_message = base64_bytes.decode('ascii')
+            thumbnail_buffer.close()
+            return base64_message
+        except Exception:
+            Logger.logException("w", "Failed to encode snapshot image")
+
+    def _convertSnapshotToGcode(self, encoded_snapshot, width, height, chunk_size=78):
+        gcode = []
+
+        encoded_snapshot_length = len(encoded_snapshot)
+        gcode.append(";")
+        gcode.append("; thumbnail begin {} {} {}".format(
+            width, height, encoded_snapshot_length))
+
+        chunks = ["; {}".format(encoded_snapshot[i:i+chunk_size])
+                  for i in range(0, len(encoded_snapshot), chunk_size)]
+        gcode.extend(chunks)
+
+        gcode.append("; thumbnail end")
+        gcode.append(";")
+        gcode.append("")
+
+        return gcode
+
+    def getSettingDataString(self):
+        return """{
+            "name": "Create Thumbnail",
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):
This was discussed with a couple of developers and we agree that while embedding PNGs in the g-code is not ideal, it'd be good to merge this for compatibility. The image seems to be encoded in the same way as PrusaSlicer, another popular slicer, so it should be easy for firmware/Octoprint to support this if they already support PrusaSlicer. With two major slicers writing this the same way, let's hope it becomes a bit of a standard.

I've reviewed the code and it looks okay. I'll give it a good test.

Ghostkeeper:
Seems to work well. Thanks!

markwinger:
I am using fluidd instead of octoprint.  It supports thumbnails from gcode but I have not been able to get cura to do it.

How do I get this change to try it?

markwinger:
I tried it.  Put the file in the scripts directory, and in showed up for post processing.  Enabled it and it did insert data for the thumbenail but fluidd did not display it.  The thumb nails in Fluidd are work with Prusa Slicer and Super Slicer.
I loaded prusa slicer and generated a file.  I noticed the slicer gcode first line says:
; generated by PrusaSlicer 2.3.0+win64 on 2021-02-04 at 22:38:39 UTC
but the curafile says:
;Generated with Cura_SteamEngine 4.8.0 down about 10 lines.
So I inserted the prusa slicer line to the cura one and it works.

Added:
gcode.append("; generated by PrusaSlicer 2.3.0+win64 on")
after line 39 and now it works. *Just a work around.  Needs a change in fluidd"

jneilliii:
This is great seeing this coming in the next release. I've had a couple of users using this approach for a while with my PrusaSlicer Thumbnails plugin in OctoPrint with much success. With this being included with the post-processing scripts I could sunset the Cura Thumbnails plugin that is dependent on the ufp file format. Thanks for everyone's hard work on making a great slicer.

daniel-kukiela:
Hi!
I have a question.
Some time ago I wrote this: https://github.com/daniel-kukiela/LotmaxxSnapshot - it works with Lotmaxx printers and I don't know if any other printers are using it, but that's quite possible. Can it be of any value here to add it as an alternative? Currently, it works in a way that if you add a specific comment to the starting gcode, it'll put one or 2 encoded images there, so Lotmaxx users can add it there and forget (will be executed only with given printers), but this behaviour can be, of course, changed.

jneilliii:
@daniel-kukiela this is interesting, that looks like the MKS format that is also included in their plugin when sending files to their printers. I was wanting to create an OctoPrint plugin to go along with my others that support this format, but haven't yet had a chance to reverse engineer the what I think is c encoded images for tft displays. If you have any good ideas/python function to convert back to png format that would make that happen a lot sooner.

daniel-kukiela:
So Lotmaxx adds this preview in their slicer for Windows (which is an older Cura, branded, with some additions), but there's nothing for Mac and Linux users as well as for Windows users which want to use the newest Cura. That's why I wrote it.
For this purpose, I reverse-engineered this format and yes, I can write some code to convert it to png no problem. I've been thinking of creating an OctoPrint plugin for this format too, just haven't had a chance to do so. I don't know OctoPront (I have PIs, just still trying to find some time to install OctoPrint and connect to my printers), so it'll be more challenging for me to write the whole plugin, but if you can do the GUI and integration part in general, I can write any other code parts for this, for example, mentioned MKS (I didn't know the name of this format) to png converter. Or I can just hand you some functions - whatever works best for you.

Also, I'm still curious if there is any interest in adding it into Cura.

fieldOfView:
@daniel-kukiela to get a plugin published in the Marketplace, you would have to submit it via https://contribute.ultimaker.com/

daniel-kukiela:
Yeah, I should probably do this, but after seeing this PR I thought that there might be an interest in adding into Cura more than just one format of embedded thumbnails (so you can choose right for your printer). My current plugin is yet another (and there's probably more?).

jneilliii:
> but if you can do the GUI and integration part in general, I can write any other code parts for this, for example, mentioned MKS (I didn't know the name of this format) to png converter. Or I can just hand you some functions - whatever works best for you.

If you provided those functions to reverse this format to png I could have a plugin up and ready for everyone in a day or two. I have a lot of experience making these plugins if you didn't already know...

https://plugins.octoprint.org/by_author/#jneilliii

daniel-kukiela:
I apologize, but I didn't. I'm yet to start my journey with OctoPrint. We should move this topic out of here - which way of contact do you prefer? I'll write a function soon and send it to you, I just have some questions like the parameters of the function (what will you geed it with, or what input should I expect).

jneilliii:
Let's move it to the above issue in one of my other thumbnail plugins that already support gcode extraction.

PR #8917, opened 2020-12-09
Creation of new profiles and addition of new materials

AButler31:
-Update to include experimental profiles for 1mm and 1.2mm nozzles
-Addition of new profiles for COPA and PLA HR870 for 0.4mm, 0.6mm and 0.8mm nozzles.


resources/variants/strateo3d_standard_10.inst.cfg
```diff
@@ -0,0 +1,20 @@
+[general]
+name = Standard 1.0 BETA!
```
nallath, line 2: Why add the BETA! in a profile? 

If it's not yet tested, I think it's better to leave it out of an official Cura release. Changing this later on might even cause issues.

nallath (changes requested):
Please write commit messages in English. We might need to check why certain changes are made.

resources/variants/strateo3d_standard_10.inst.cfg
AButler31: The profiles for these nozzles have been tested on certain parts but are still being improved to accommodate more complexe prints . We sell 1mm and 1.2mm nozzles and would like to make our profiles available to our clients who are interested in printing with these nozzles. Would changing 'BETA!' to 'Experimental' be acceptable ?
Thanks

PR #8924, opened 2020-12-10
Added Eryone ER20 3d printer

Eryone:
**Hi,** 
We are Eryone 3d printer manufacturer from Shenzhen of China. We want to add the ER20 3d printer to the next version of Ultimaker Cura. Here are the details of our printer: https://eryone.com/ER_20/show/90.html

We added these files:
-eryone_er20.def.json
-eryone_er20_extruder_0.def.json
-eryone_er20_plateform.stl
-eryone_er20_draft.inst.cfg
-eryone_er20_normal.inst.cfg
-eryone_er20_high.inst.cfg

**Greeting**


resources/quality/eryone_er20/eryone_er20_high.inst.cfg
```diff
@@ -0,0 +1,79 @@
...
+
+[metadata]
+setting_version = 16
+type = quality
+quality_type = high
+weight = 1
+global_quality = True
+
+[values]
+acceleration_enabled = True
```
nallath, line 14: Why not add this setting to the machine definition? It seems to always be the same. There are quite a few other settings where this is also the case. 

This makes it a lot easier to maintain the profile in the future; since that way the quality profile only has the settings that it actually changes.

Eryone: Hi, nallath
Thanks for your suggestions! You are correct.
Later I will update it. 
Greeting

konskarm:
The definition works without a problem so I am merging this. The printer will show up in the next Cura release.

adamheffernan:
When's the next release planned?

Ghostkeeper:
It's a bit up in the air at the moment. Due to the Arachne project and the relatively few other features we've built, it doesn't make sense to release at our regular schedule. It's definitely not going to be within a month, but not more than 3 months from now, I'd guess.

adamheffernan:
Thanks @Ghostkeeper

PR #8961, opened 2020-12-17
Small parameter update for printing improvement

FusedForm:
Small changes in printing parameters for printing performance improvement


Ghostkeeper (approved):
It's too bad that the FusedForm website doesn't list the build volume dimensions on its website. Just the outside machine dimensions. That might've prevented some of these mistakes!

PR #8963, opened 2020-12-18
Tronxy profile updates

64bittuning:
Over the last few months, I have gathered a lot of feedback from the community regarding the Tronxy profiles that I built several months back. I've been building several profiles and trying to tweak much of what was built, and I believe I now have a product that will work well for a large majority of the community.

Hopefully I'm on time this time, for the next update!


64bittuning:
No longer supporting Tronxy and their machines.

Please contact me. These profiles need to be removed from Ultimaker Cura

Ghostkeeper:
We're happy to merge or keep the "best known" profiles, even if you don't want to maintain them any more. It's not required of you to maintain them.

We're not about to delete profiles from Cura upon request though. For one that would need a version upgrade, but also it kind of allows for anti-competitive behaviour among printer manufacturers. If the printers are out in the field, I see no reason why we'd delete these profiles.

64bittuning:
Well, for one... It took a solid month and a half for these updates to even be looked at. 
Two, Tronxy has specifically stated to their users NOT to use these profiles.
Three, as it is my intellectual property, I should have control over whether or not it is posted.

I would much rather have this discussion in private though, which is why I requested to be contacted.

Ghostkeeper:
You don't have control over these profiles once posted, and neither does Tronxy. Ultimaker's lawyers have considered them to be data files, which do not fall under any copyright. If they are wrong and a judge would not consider them data files, they would be under Cura's LGPL license. In the second case you will have copyright, but by publishing it licensed anyone to publish them again under the conditions of the LGPL. In either case, we are free to include them in Cura against Tronxy's will.

Sorry that it takes so long to include updates into Cura. We have too long of a backlog. It wouldn't be released yet anyway though since there hasn't been a new release of Cura yet.

PR #8986, opened 2020-12-21
Store auth & refresh key in keyring instead of in preferences

nallath:
People tend to share configuration folders, which just isn't secure.

CURA-7180

https://github.com/Ultimaker/cura-build-environment/pull/93


rburema:
Note that the tests are failing because this pulls a new package into the build-environment!

cura/OAuth2/SecretStorage.py
```diff
@@ -0,0 +1,20 @@
...
+    def __delitem__(self, key):
+        if key in self._stored_secrets:
+            del self._stored_secrets[key]
+            keyring.delete_password("cura", key)
+
+    def __setitem__(self, key, value):
+        self._stored_secrets.append(key)
+        keyring.set_password("cura", key, value)
+
+    def __getitem__(self, key):
```
nallath: Missing typing

cura/OAuth2/SecretStorage.py
```diff
@@ -0,0 +1,20 @@
+import keyring
+
+
+class SecretStorage:
```
nallath, line 10: Missing documentation. What does this class do, What is it's point? How should it be used?

cura/OAuth2/SecretStorage.py
jellespijker: You jumped the gun here @nallath ;-) Still very much in the "in progress" column at the moment.
Anyhow. I removed the SecretStorage class since I now Use a descriptor for the authenicationresponsemodel attributes

cura/OAuth2/SecretStorage.py
jellespijker: You jumped the gun here @nallath ;-) Still very much in the "in progress" column at the moment.
Anyhow. I removed the SecretStorage class since I now Use a descriptor for the authenicationresponsemodel attributes

rburema (approved):

StefanBruens:
This references keyring.errors.NoKeyringError, which was added with python-keyring 10.6.0, released Jan. 2018.

Ghostkeeper:
Is that a problem? We're using version 23.0.1 of the "keyring" package. See https://github.com/Ultimaker/cura-build-environment/blob/c27ab096b87ffb20d5a63f3f67f0c079599eb64f/projects/python_packages.cmake#L106

StefanBruens:
Not everyone is using these huge, mostly redundant application bundles.

Ghostkeeper:
But everyone is using at least the same dependencies, right? Aren't you? If not, you're kind of on your own.

StefanBruens:
Defnitely not. Its almost impossible to use the same versions, and sometimes it is also a bad idea to do so (E.g how long has Cura been using on an unsupported non-LTS Qt 5.10?).

For python-keyring version currently in use, see https://repology.org/project/python:keyring/versions

Also, requirements.txt is not mentioned/referenced in README.md (Cura and cura-build), nor are the versions checked from CMake or in any other way during the build.

Ghostkeeper:
How people choose to package Cura for Linux distributions is up to them. We only support the AppImage here.

That is a decision of priority. We are a team of 5 programmers, and Linux is just 1% of our users. We support Linux through this AppImage since it is easier to maintain, precisely for the sort of reason you are bringing up here.

PR #9007, opened 2020-12-23
fix typo: "Feedrate" and "Flowrate" to "Feed Rate" and "Flow Rate"

michaelpeterswa:
Makes g-code slightly easier to read given that "feedrate" and "flowrate" aren't words on their own. It appears that within formal texts (manuals, textbooks, etc...) that this is the appropriate change to make. In casual circles though, it seems that either way is acceptable. Ref: https://www.reddit.com/r/ChemicalEngineering/comments/992u4r/flowrate_or_flow_rate/


michaelpeterswa:
Also of note, given that there are hundreds of similar changes to make, this change only applies to Creality Printers.

Ghostkeeper:
I disagree with this change. "Feedrate" is an industry term, used like this in many places.

See also the [Wikipedia article](https://en.wikipedia.org/wiki/Speeds_and_feeds) which calls it "Feed rate" but provides "feedrate" as alternative spelling. Other sources: [Marlin consistently calls it "feedrate"](https://marlinfw.org/docs/gcode/M220.html) and in [formal texts](https://link.springer.com/article/10.1007/s11431-018-9404-9) it's also often called "feedrate". Colloquially, looking at the first few pages of search results for both terms, it seems that "feed rate" and "feedrate" are approximately equal in use. Either search term results in a mix of the two uses in any case.
There is a grammar convention in English to not compound words like this. But then this rule is violated as often as not. Take for instance "cottonmouth", "website", "bookstore", "classmate", "snowman", "sailboat", "playground", "overkill"...

It would be good to make it consistent within Cura at least though. Currently in Cura I see 6 uses of "feed rate" and 373 uses of "feedrate". So it would be easier to change it in the other direction.
For "flowrate" vs. "flow rate" the balance is more equal (37 vs. 19 respectively). And most of these uses are in the ChangeAtZ script anyway so it's not a very good sample. But it's probably better to keep it consistent with the use of "feedrate" I guess.

michaelpeterswa:
I agree, I was probably just a little bit too trigger-happy with this one.

Thanks for taking the time to check it out.

PR #9014, opened 2020-12-24
PostProcessingPlugin FilamentChange added Marlin Advanced Pause Feature!

Sekisback:
with this modification, the plugin can use M600 from MarlinFW


plugins/PostProcessingPlugin/scripts/FilamentChange.py
```diff
@@ -74,22 +93,26 @@ def execute(self, data: List[str]):
...
-        if y_pos is not None:
-            color_change = color_change + (" Y%.2f" % y_pos)
+            if x_pos is not None:
+                color_change = color_change + (" X%.2f" % x_pos)
 
-        color_change = color_change + " ; Generated by FilamentChange plugin\n"
+            if y_pos is not None:
+                color_change = color_change + (" Y%.2f" % y_pos)
+                
+            color_change = color_change + " ; Generated by FilamentChange plugin\n"
```
Ghostkeeper, line 115: Might as well just unindent this line and remove the `if` case of the if-else clause. They are (and should stay) the same. And this one has no whitespace at the end of the line.

plugins/PostProcessingPlugin/scripts/FilamentChange.py
```diff
@@ -19,6 +22,14 @@ def getSettingDataString(self):
             "version": 2,
             "settings":
             {
+                "marlin": 
+                {
+                    "label": "Use Marlin Config?",
+                    "description": "Use configuration from Marlin Advanced Pause Feature!",
+                    "type": "bool",
+                    "default_value": true
```
Ghostkeeper, line 30: We don't want to default to `true` here. Existing FilamentChange users would see their Filament Change settings suddenly not working any more. It's also good to give this more time in the field before it gets to be the default. Not everyone has the feature in their printer either.

plugins/PostProcessingPlugin/scripts/FilamentChange.py
```diff
@@ -19,6 +22,14 @@ def getSettingDataString(self):
             "version": 2,
             "settings":
             {
+                "marlin": 
+                {
+                    "label": "Use Marlin Config?",
+                    "description": "Use configuration from Marlin Advanced Pause Feature!",
```
Ghostkeeper, line 28: We should use a more... neutral tone.
```suggestion
                    "description": "Use configuration from Marlin's Advanced Pause Feature.",
```

Ghostkeeper (changes requested):

Ghostkeeper:
These are things that I could fix myself pretty easily, but I'll let you fix them first if you like. The feature itself would be nice for many people.

Sekisback:
please feel free to change/fix these things to your needs, glad i could help

Ghostkeeper:
I changed the three things that I remarked about myself. Thanks for the contribution!

PR #9036, opened 2020-12-29
[WIP/RFC] Show nozzle coordinates in layer view

fieldOfView:
This PR adds a "toolhint" to the virtual nozzle in SimulationView which shows the nozzle coordinates. The toolhint moves with the nozzle.
shown here with the sidebar GUI enabled, but it also works with the standard GUI

This could be an alternate solution to the recurring question to add the Z height to the layer slider.

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -35,6 +35,7 @@ class SimulationPass(RenderPass):
         self._disabled_shader = None
         self._old_current_layer = 0
         self._old_current_path = 0
+        self._old_head_info = {}  # type: Dict[str, Union[List[float], bool]]
         self._switching_layers = True # It tracks when the user is moving the layers' slider
         self._gl = OpenGL.getInstance().getBindingsObject()
         self._scene = Application.getInstance().getController().getScene()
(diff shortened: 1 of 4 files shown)


Ghostkeeper:
I think a lot of people would love this sort of information. It feels like it would be useful for some things. In practice, it's not all that useful, probably. The only real use case I can think of would be to know at what Z coordinate to continue a previously aborted print, and that's not a common use case. Yet it gives some feeling of having control, which is good.

One possible consideration is which coordinate system to use for this. You're now using Cura's front-end coordinate system, which aligns with the coordinates of the Move Tool but not with the coordinates of some of the settings or what people find in the g-codes. *I think your current choice makes the most sense.* But not everyone would agree. Especially people who dig through the g-code coordinates.

Perhaps it would also look better using the small font size. It's less prominent then, and it doesn't need to be prominent.

nallath:
I think it makes most sense to display g-code coordinates. My rationale for this is that this is called the simulation view. As such, the data that you get from it should reflect what is happening if it's printed (To me that is also the reason why the nozzle is in the animation in the first place)

Having it in g-code coordinates would make it easier to find the right bit of g-code if you are manually debugging things. I would see myself using this.

fieldOfView:
> I think it makes most sense to display g-code coordinates. 

I understand your rationale, but this is not how it is done in other areas in Cura. Layer numbers don't match gcode layer numbers for example. I think it would be confusing to show a different coordinate system than what is used in the translate tool.

alexjx:
Is it possible to add more information? For example, current feedrate? Fan speed? That will be very helpful when reviewing detail, especially when diagnosing print issues. I know it might require parsing and tracking gcode. But to show coordinate we must have done the parsing right?

Ghostkeeper:
> I know it might require parsing and tracking gcode. But to show coordinate we must have done the parsing right?

When you load a g-code file into Cura, it's parsing it. That takes a long time though. After a normal slice, the slicing engine just gives the information the front-end needs on a silver platter, so no parsing is needed. This includes all information that the layer view currently shows: Vertex coordinates, layer thickness, line width, line type and feedrate. Those wouldn't be very hard to show. Other information, like fan speed, is not communicated yet so that would be harder.

But with any such decision it's important to consider not to overwhelm the user with information it doesn't need. An interface becomes very confusing if you show all possible information.

lorenzoromagnoli:
@fieldOfView what would be the use of showing the nozzle coordinates in the preview? And which coordinate system are you proposing to use?

lorenzoromagnoli:
If the PR is intended to mainly show the z height next to the layer count I'd rather only add that instead of showing also x and y which might be confusing for the user. As far as I know, and please correct me if I'm wrong, the z height would be the same in gcode coordinate system and in Cura UI.

If that's the case I'd rather move the z-offset indication closer to the play/progress element in the bottom of the screen and give a more professional look to that element instead of adding a new component. 

Something like this.
Screen Shot 2021-01-27 at 08 49 39

What do you think?

nallath:
I think moving it to the bottom component makes it more confusing. You don't control the layer height with that component, so it's a bit weird to leave it there.

fieldOfView:
Sorry to let my own proposal linger.

This started out as a way to solve the oft-returning request for showing the z coordinate of the layer next to the layer number. But the current solution mostly fails at that because it only shows this height when simulating the layer, not when browsing through layers. I agree with @nallath that @lorenzoromagnoli's proposal only makes this more confusing, especially since the layer slider and the toolpath slider are so far away from eachother in the normal UI (it could work slightly better in the sidebar gui in my opinion).

Regarding the coordinate system to use, that is one of the strange issues (from a user standpoint) with Cura: the different coordinate systems and layer numbers. Please pick one and stick with it. For this proposal I picked the coordinate system that is exposed in the UI most (though the origin marker contradicts it) over the coordinate system in gcode, to match the choice that was made in postprocessing scripts.

I'll close this PR until I have time to revisit it, so that it does not stay open without a resolution longer than necessary.

PR #9061, opened 2021-01-02
Cultivate3D Elevator with Medusa Machine Definitions

c3D-Dan:


fieldOfView:
What material is this supposed to use? It specifies generic_pla as default material, but also excludes generic_pla.

Why do you have the "base" definition when this only adds a single printer?

c3D-Dan:
Base : There is an intention to add more printers.
Excludes :  are overridden.  Ill remove them.

Ghostkeeper:
Please don't add materials to this repository. Any updates to the existing material profiles need to go into https://github.com/Ultimaker/fdm_materials . Those files are shared between Cura and Ultimaker's 3D printer firmware.

c3D-Dan:
> Please don't add materials to this repository. Any updates to the existing material profiles need to go into https://github.com/Ultimaker/fdm_materials . Those files are shared between Cura and Ultimaker's 3D printer firmware.

Perhaps your could advise of an alternative means to prevent standby temperature from dropping when used with our machine alone?   Editing the materials was the only means I could find thus far.  Any help greatly appreciated.

Ghostkeeper:
It's fine to edit the material profiles! Just edit the ones in the [fdm_materials](https://github.com/Ultimaker/fdm_materials) repository (in a separate pull request) rather than copy-pasting the profiles here. If we were to merge this, there would be two files for each profile and they get out of sync.

c3D-Dan:
Thank you for the advice.   I will do this in the coming days.

nallath:
@c3D-Dan Could you have another look at this PR?

PR #9101, opened 2021-01-10
Introduce DWMaker machines

alexjx:


Ghostkeeper:
This got somewhat political due to DWMaker possibly infringing the non-commercial clause of the CC-NC license on Ultimaker's hardware designs. For that reason, we're asked to reject this pull request. I'm sorry! This is between Ultimaker and DWMaker.

PR #9112, opened 2021-01-11
Adding Twotrees Full Range of 3d Printer profiles

Lester3DGadgets:
Model to add from Two Trees:
Bluer
Bluer Plus
Sapphire Pro
Sapphire Plus


Lester3DGadgets:
Im adding all 4 Printers models under a proper structures but it seems the model bluer was having conflict with another 1 added by different contributor.

PR #9114, opened 2021-01-12
Adding Twotrees Full range printer profiles

Lester3DGadgets:
Profile for:
Bluer / Bluer Plus / Sapphire Pro / Sapphire Plus

Nozzle size : 0.2mm to 1mm

Material: PLA / ABS / PETG / TPU

PR #9132, opened 2021-01-15
add definitions for Two Trees Sapphire Plus

G4Cab:
add definitions for Two Trees Sapphire Plus


konskarm:
@G4Cab there is another PR (#9114) that is attempting to add the full range of Two Trees machine definitions, including the one you are trying to add with this PR. If you think that that PR covers yours, maybe we can close this one.

konskarm:
Closing this in favour of https://github.com/Ultimaker/Cura/pull/9114, which adds the full range of TwoTrees printers.

PR #9150, opened 2021-01-20
add definitions for LONGER fdm printer series

lowkeyjoe:
- add definisions for LONGER LK1, LK1 Pro, LK1 Plus, LK4, LK4 Pro, LK5, LK5 Plus, and Cube2.
- have tested in longer fdm printer


Ghostkeeper:
These profiles seem to be copied from the Creality profile set. Aside from the definitions, I wasn't able to find any changes. Is it better to simply say that `"quality_definition": "creality_base"` and then remove the duplicates?

lowkeyjoe:
> These profiles seem to be copied from the Creality profile set. Aside from the definitions, I wasn't able to find any changes. Is it better to simply say that `"quality_definition": "creality_base"` and then remove the duplicates?
I'm sorry bring some trouble to you and Cura team. So Many LONGER users eagerly want to use CURA to experience new features and improve printing quality. So We made a inital and rough version for them immediately referring to CREALITY profile.  These profiles not the final version, and we will continue to update them later.  If modified definition to creality base, we can't freely make it. At last, thanks so mush for your and Cura team's hard work to bring us brilliant software.

Pingu94:
Hi guys, 
I'm a Longer printer user. I've found Longer cura profles with all resources files on the official Longer3d website and I was wondering if someone has already asked for a pull request for official support, so I jumped here. I'm pretty new here, please let me know if I can help in any way.

Ghostkeeper:
We'd still prefer to refer to the `creality_base` profiles then as long as they are the same. You have to understand that these 148 files are being read on every start-up for everyone (to determine their metadata to see which printer they belong to). As long as they are the same as the Creality profiles, this is only wasting everyone's time.

If you are going to make modifications later, you can also separate them from the Creality profiles later. As of now I have no guarantees that you'll come around to making these modifications, even with best intentions.

lowkeyjoe:
> We'd still prefer to refer to the `creality_base` profiles then as long as they are the same. You have to understand that these 148 files are being read on every start-up for everyone (to determine their metadata to see which printer they belong to). As long as they are the same as the Creality profiles, this is only wasting everyone's time. Reading every file at every launch is a problem of Cura as a whole, I know, but it's something we'll have to deal with for the time being.
> 
> If you are going to make modifications later, you can also separate them from the Creality profiles later. As of now I have no guarantees that you'll come around to making these modifications, even with best intentions.

Thank your reply. I didn't realize the launch time issue of CURA before. The 148 files really will influence the launch process. I 
agree with your idea. I will modified profiles to squeeze quantity of files

lowkeyjoe:
Hi, I have updated LONGER profiles, delete many files. Please deal with it when you are free

resources/definitions/longer_base.def.json
```diff
@@ -0,0 +1,265 @@
...
+            "ultimaker_pp_transparent",
+            "ultimaker_pva",
+            "ultimaker_tough_pla_black",
+            "ultimaker_tough_pla_green",
+            "ultimaker_tough_pla_red",
+            "ultimaker_tough_pla_white",
+            "ultimaker_tpu_black",
+            "ultimaker_tpu_blue",
+            "ultimaker_tpu_red",
+            "ultimaker_tpu_white",
```
Ghostkeeper, line 115: Many of these materials are 2.85mm filaments which will be filtered by Cura anyway, and don't need to be in this list.

Are you sure these have been tested and seen as incompatible with the Longer printers? By excluding these materials, users aren't able to select the material profiles for them, and that can be seen as anti-competitive if there was no empirical test.

Ghostkeeper (changes requested):
Looks good for the most part. Just a bit of weirdness left with the excluded materials. (This is an actual thorough review.)

I see you've constrained the profiles to a subset, most likely just what you could test in practice. Thanks!

resources/definitions/longer_base.def.json
lowkeyjoe: Yes, you are right. I have removed all material restriction. Is there any other issue?

Ghostkeeper: No, that's all I could find. I've reviewed the changes and tested all printers and profiles. Seems to slice correctly now. So thanks for the contribution!

lowkeyjoe: Thanks for your hard work.

BradleyFord:
Is it possible for this to be extended to include the LK4 X?

PR #9161, opened 2021-01-21
Restore UFP Thumbnail

rburema:


plugins/UFPWriter/UFPWriter.py
```diff
@@ -72,25 +61,23 @@ def write(self, stream, nodes, mode = MeshWriter.OutputMode.BinaryMode):
...
-        #
-        #     thumbnail.write(thumbnail_buffer.data())
-        #     archive.addRelation(virtual_path = "/Metadata/thumbnail.png",
-        #                         relation_type = "http://schemas.openxmlformats.org/package/2006/relationships/metadata/thumbnail",
-        #                         origin = "/3D/model.gcode")
-        # else:
-        #     Logger.log("d", "Thumbnail not created, cannot save it")
+        # Attempt to store the thumbnail, if any:
+        backend = CuraApplication.getInstance().getBackend()
+        snapshot = None if getattr(backend, "getLatestSnapshot", None) is None else backend.getLatestSnapshot()
```
Ghostkeeper, line 66: Hmm, could consider making this `getLatestSnapshot` function part of `UM.Backend` in the form of a stub so that we don't have to check this.
I think it's fine to leave it for now, but putting it out there if you agree it's better.

Ghostkeeper: Actually, thinking about the scope of the class, it's really not supposed to be part of a back-end application so I'm not convinced this should be changed any more.

Ghostkeeper (approved):

PR #9182, opened 2021-01-25
options to restrict hole expansion

alexjx:
Corresponding to CuraEngine change: https://github.com/Ultimaker/CuraEngine/pull/1397

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1403,6 +1403,32 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "hole_xy_offset_max_size":
+                {
+                    "label": "Max Hole Diameter to Expand",
+                    "description": "The max diameter of the holes will be expanded.",
+                    "unit": "mm",
+                    "type": "float",
+                    "minimum_value": "0",
+                    "default_value": 0,
+                    "limit_to_extruder": "wall_0_extruder_nr",
+                    "settable_per_mesh": true,
+                    "children":
+                    {
+                        "hole_xy_offset_max_length":
+                        {
+                            "label": "Max Hole Length to Expand",
+                            "description": "The length of the outline of the holes will be expanded.",
+                            "unit": "mm",
+                            "type": "float",
+                            "minimum_value": "0",
+                            "default_value": 0,
+                            "value": "hole_xy_offset_max_size * math.pi",
+                            "limit_to_extruder": "wall_0_extruder_nr",
+                            "settable_per_mesh": true
+                        }
+                    }
+                },
                "z_seam_type":
                 {
                     "label": "Z Seam Alignment",


nallath:
We currently already have a setting that is _very_ close in naming of the settings you are adding (Small Hole Max Size). Could you see if you can somehow combine these so that it's less confusing (either by merging them, or some other solution).

martinsykes1:
Same issue - I have downloaded a stackable storage tray (https://www.thingiverse.com/thing:2562937) which connects with lugs on the top that fit into holes on the bottom and I want to expand the holes so that the trays stack easily. However, it also treats the much larger tray compartments as holes and expands them so the walls between the compartments become too thin to print

Ideal solution would be to use the small hole max size parameter to control the use of the additional offset as well as the print speed

alexjx:
Hi @rburema, I was intended to follow @nallath 's suggestion to combine the two options, so we shall have one less option in the already crowded option-universe of Cura. However, I didn't get any time to look into it since then.  

Given it a quick look, the "Small Hole Max Size" option have a speed reduction behavior applied. By reusing it, we are overloading the option to have two combined behaviors, which I'm reluctant to have. Since we could have small hole we want to expand but not small enough to slow down.

Regarding to the name, I don't have a candidate right now. To the confusion part, how about nesting the options under the expansion?

Jia

rburema:
Hi @alexjx No worries, I just thought you where waiting for us and might not have seen the other notifications.

> By reusing it, we are overloading the option to have two combined behaviors, which I'm reluctant to have.

I agree, I don't think they should be the same setting, no.

> To the confusion part, how about nesting the options under the expansion?

Maybe? I actually think the real problem is, the original name we have in there is too vague. (Since it's just about speed.) Perhaps that should be changed instead, since it's only about the speed. It would still be good to have that change in/with this PR however, as on it's own it would make little sense.

martinsykes1:
Just wanted to clarify that my comment is nothing to do with speed. We need some way to limit the hole expansion so it doesn't expand holes which are already above a certain size. In my example, expanding the large 'holes' which make the storage tray compartments causes the dividing walls to become too thin to print.

rburema:
@martinsykes1 True but I was referring to the other setting (not comment) this one could be confused with.

Ghostkeeper:
I think overloading the Small Hole Max Size to change two settings isn't really a problem. It's similar to many other structures of Cura, such as Top Surface Skin Layers which separates out N layers to get different speed/line width/flow/etc. This Small Hole Max Size creates a concept of "small holes" which can get some separate settings from the rest.

I'd prefer that over having "Small Hole Max Size for Fan Speed Change" and "Small Hole Max Size for Horizontal Expansion Change" or something of the like. That would feel more complex.

alexjx:
I don't think "Top Surface Skin Layers" are the same here. Since all those properties it has impact upon are attributes of "top surface skin layers".  It's logically coherent. "small hole to slow speed" and "small hole to expand" could be different things to the end user.

Assuming we don't want to involve much post processing procedures, otherwise slow down or compensation are all meaningless. Let's consider if we want to print some electronic cases, which have a window to mount a display, while some standoff to support PCB. If we just define "small hole" to be "internal boundary". We will be slow down the window as well as the standoff.  I don't think this is what I want. I want the case to print as fast as possible, while I only slow down so the sandoff won't be drag away which is necessary.

PR #9202, opened 2021-01-29
Adding eryone thinker series 3d printer profile

Eryone:
Hi,
We are Eryone 3d printer manufacturer from Shenzhen of China. We want to add the thinker series 3d printer to the next version of Ultimaker Cura. Here are the details of our printer: 
https://eryone.com/EryoneThinkerSE/show/86.html

Greeting


jellespijker:
Looks good! Thank you for your contribution. I will merge it into master so it should be in for the next release.

PR #9203, opened 2021-01-29
Don't take snapshot if no main window present.

Ghostkeeper:
When autoslicing is enabled, it was trying to create a snapshot before the main window is present.

You'll need the Uranium branch of the same (fix_is_visible) name if you don't want this to crash on slicing!


Ghostkeeper (commented):
Bit of defensive programming can't hurt here. Nice one.

PR #9213, opened 2021-01-31
Update ideagen3D_sapphire_plus.def.json

Firedrops:
Added Restore Bed Levelling after Homing to start G-Code.

### DIFF
--- a/resources/definitions/ideagen3D_sapphire_plus.def.json
+++ b/resources/definitions/ideagen3D_sapphire_plus.def.json
@@ -1,5 +1,5 @@
 {
-    "version": 3,
+    "version": 2,
     "name": "ideagen3D Sapphire Plus",
     "inherits": "fdmprinter",
     "metadata":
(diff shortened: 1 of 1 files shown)


resources/definitions/ideagen3D_sapphire_plus.def.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": 2,
+    "version": 3,
```
Ghostkeeper, line 2: This is a version number of the file format, and should be left at 2 since it's the most recent version.

```suggestion
    "version": 2,
```

Ghostkeeper (changes requested):

resources/definitions/ideagen3D_sapphire_plus.def.json
konskarm, line 2: Although @Ghostkeeper's comment was originally resolved, the version is now again changed to 3. Please change it back to 2 and the tests won't fail.


```suggestion
    "version": 2,
```

konskarm (approved):

PR #9232, opened 2021-02-03
Retain metadata from 3MF files through save/load loops

Ghostkeeper:
Both metadata on the scene and metadata on the objects should be retained when loading 3MF and then re-saving it.

This PR depends on the following PRs in other repositories:

* https://github.com/Ultimaker/libSavitar/pull/32
* https://github.com/Ultimaker/Uranium/pull/670

Contributes to issue CURA-7615.


Ghostkeeper:
I don't know what made the tests fail here. It's not failing for me locally, at least not that way (just with typing in the new DF code). Perhaps they were failing back in December that way but changes to Uranium fixed them.

PR #9241, opened 2021-02-05
CURA-7996: Fix trying to open deleted files from the recent files list

konskarm:
CURA-7996


cura/CuraApplication.py
```diff
@@ -1940,6 +1945,13 @@ def checkIsValidProjectFile(self, file_url):
         try:
             result = workspace_reader.preRead(file_path, show_dialog=False)
             return result == WorkspaceReader.PreReadResult.accepted
+        except FileNotFoundError:
+            result_message = Message(text = self._i18n_catalog.i18nc("@info:status Don't translate the XML tag <filename>!",
+                                                                     "Failed to load <filename>{0}</filename>. No such file or directory.",
```
Ghostkeeper, line 1950: Looks like you wrote this kind of automatically, but I don't think project files can ever be a directory. I'll adjust this message to "File is moved or deleted."

resources/qml/Dialogs/AskOpenAsProjectOrModelsDialog.qml
```diff
@@ -29,6 +29,7 @@ UM.Dialog
     modality: Qt.WindowModal
 
     property var fileUrl
+    property var addToRecent: true
```
Ghostkeeper, line 32: Would be good to document this property, for posterity.

Ghostkeeper (changes requested):
I'll apply this change myself to prevent ping-ponging.

PR #9245, opened 2021-02-05
[CURA-7979] Add 'Outer Wall Lock Factor' setting.

rburema:
Most of this work is in the Engine, see: https://github.com/Ultimaker/CuraEngine/pull/1405

Arachne produced artifacts in straight outer walls if the inner walls had curves in them (or appeared and dissapeared behind them). 'Hard' fixing that would reduce the flexibility of Arachne, so a setting is introduced how much the outer walls should be locked into place (as much as possible at 100%, not at all at 0%).


jellespijker (changes requested):
Not longer needed in the front-end I will therefore close this PR.

BagelOrb:
Why was it decided to not put this setting in the frontend?

For visual quality you might want this, but for structural qualities it might be better to not lock the width of the outer wall.

PR #9250, opened 2021-02-06
Microswiss direct drive profile

RFBomb:
I have created a custom printer profile for this project:

https://www.thingiverse.com/thing:4685332

This allows anyone with an Ender 3 to more easily set up a dual-nozzle printer. It also changes the base settings for the retracts since this is based off the direct drive system. 
For those users that are using just the direct drive system with a single nozzle, the changes would mainly be the startup GCODE and retraction settings. 

This pull request stores the file into the CUSTOM section, since that is what it is, even if it was based off the ender 3 profile. 
I have also taken the liberty of adding in various descriptions inside the base file to summarize it for anyone that wanted to modify it for their own needs.


resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
```diff
@@ -0,0 +1,118 @@
+{
+    "name": "Ender-3 MicroSwiss DirectDrive",
+    "version": 2,
+    "inherits": "creality_ender3",
+    "metadata": {
+        "quality_definition": "creality_base",
+        "visible": true,
+        "platform": "creality_ender3.3mf",
+		"author": "GitHub.com/RFBomb",
```
Ghostkeeper, line 9: We indent with spaces, not tabs. You indent with a mix of them, it seems.

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
```diff
@@ -0,0 +1,118 @@
...
+		"retraction_extrusion_window": { "default_value":  "retraction_amount * 2" },
+        "retraction_speed": { "default_value": 15.0 }, 
+        
+
+        "machine_heat_zone_length":  { 
+            "description__": "The distance from the tip of the nozzle in which heat from the nozzle is transferred to the filament.",
+            "default_value": 16.0 
+        },
+        
+        "_ExtruderSwitchingSettings_Comment": "Settings to retract material when switching over to the other nozzle",
```
Ghostkeeper, line 64: There's a bunch of these that are not read by Cura. Could you remove them?

resources/extruders/MicroSwiss_DirectDrive_extruder_0.def.json
```diff
@@ -0,0 +1,32 @@
+{
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "_MachineComment": "Ender3_MicroSwiss_DirectDrive",
+		"machine": "creality_base",
```
Ghostkeeper, line 7: Since the `creality_base` doesn't use these extruders, this causes the automated tests to fail. Similar for extruder 1.

Ghostkeeper (changes requested):

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
RFBomb: I used VS-Code to make these changes, and am personally used to using tabs in my other work. I must've had the setting incorrect. I'll fix these and re-commit

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
RFBomb: They aren't read by cura, so shouldn't be a problem. (I formattted them purposefully to make it dissimilar to standard scheme). I found a few threads on how to use GitIgnore to ignore specifics lines, so I'll get that set up so these lines are no longer committed. They were there for my/anyone else that might modify them organization purposes mostly anyway.

resources/extruders/MicroSwiss_DirectDrive_extruder_0.def.json
RFBomb: This one I'm actually confused about, but thanks for pointing out why the build failed. I wasn't sure why it had failed, since this is part of the printer profile I'm currently using just fine in cura currently. So the file _works_, but understand it may not be _correct_. 
Do you think that the only issue is this line: ` "machine": "creality_base",` specifies creality_base instead of `"name": "Ender-3 MicroSwiss DirectDrive",` (as specified by the printer's .def.json file ?

PS: As with the above comment, I'll remove the extraneous comments from the commits. I'll also convert the spaces in the _name_ printer.def.json file to underscores to match the correct scheme (_creality_ender3_).

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
RFBomb: OK so after some test commits to my fork,  I found these items in the build test log:
`[MainThread] UM.Settings.DefinitionContainer._preprocessParsedJson [299]: Unable to override setting _Comment_Cura_Keywords` -> Apparently the test doesn't like them, but (other than writing to the log?) Cura doesn't seem to fail to use the files. As I said above, I'll remove them.

I also saw this : 
```
assert is_valid  # The definition has invalid JSON structure.
E           assert False
```
in reference to my file `Ender3_MicroSwiss_DirectDrive.def.json`. I take it this means that the structure is invalid? I'm a bit unsure why at the moment if that is the case. Other than a few extra spaces in a spot or two, the structure appears valid to me...

RFBomb:
@Ghostkeeper 
I've converted this to a draft instead of a PR so I can validate the profile a bit more.
Thanks for the pointers as to why it was erroring out during the testing. I was a bit confused as to why the test kept calling out invalid JSON, when all the other validators I put it through said it was valid. But it was a fluke, and ended up being the _Comments I had put in to help myself organize. I didn't expect it to fail because of them. 

Once I validate it a bit more I'll change it back into a PR. If you (or any of the team) have any recommendations prior to that, I'm open to hear them.

resources/extruders/MicroSwiss_DirectDrive_extruder_0.def.json
Ghostkeeper: The tests validate whether the link goes both ways, indeed, which is not required for the printer to be added but may play into other things. I'm actually not sure why it needs that, but the issue arises from the metadata being necessary at all for which there is probably a good reason. Even if it's not used for adding the printer it could be used elsewhere.

Ghostkeeper:
The manufacturer seems to be incorrect to me, still. This is the header under which the printer will be sorted, so it'd go in the same category as Custom FFF Printer then. Seems like just Creality will be better. Sure, it's officially a mix of manufacturers for different parts, but people would look under Creality I reckon since that is the base printer.

Other than that, it looks okay to me at a glance. Did the profiles work out okay?

RFBomb:
The profile itself is working ok. I'm actually still fine tuning it a little bit. I've messed with PrusaSlicer and some firmware settings trying to fine tune things, such as StartupGCODE. Since I developed this with my custom 2nd extruder in mind, I had modified the purge line on the start GCODE used by the ender 3 to dynamically choose to purge the active nozzles and perform a nozzle wipe across that line if a nozzle was inactive. But all that requires If-Else statements, which Cura doesn't support afaik. 

Now that I have gotten it to a point where I like it though, I just need to backport it over to Cura in this profile. Should be easily doable, I just haven't had a chance to set it up. Once I test that it I'll get it converted back to a PR. My plan here would be to do a second purge line with Nozzle 2, but have it be commented out to avoid causing issues for people with only 1 nozzle configured.

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
RFBomb: Complete.

RFBomb:
Alright, these latest pushes I made to my repo have the following changes:

- Improved Startup GCODE
     -  Startup GCODE now has proper code to purge Nozzle 1 and Nozzle 2. This should also allow both extruders to wipe any material off of them prior to starting printing. 
     - End of startup gcode sets both nozzle 1 and nozzle 2 temperatures to 0. This is followed by setting the temp for the current extruder to the first layer temperature. These 3 lines effectively cooldown any unused extruder, while keeping the first one heated up. There is also a command to ensure that the extruder Cura expects to be active is selected. 
     - Nozzle 2 offsets are built into the startup gcode. If the user has firmware offsets, there is a note to adjust the X value for the Nozzle 2 purge. 
     - Since Nozzle the printer profile is setup for 1 extruder by default, all references to Nozzle 2 are preceeded with `;` in order to ignore them. Should they want a second nozzle, they can uncomment those lines.

RFBomb:
Before this is merged, I have a question:

I believe I have achieved the best print results by using the Nozzle Offsets to affect GCODE (as opposed to having the offsets in the firmware - having it as part of gcode effectively allows it to wipe the previously selected nozzle when moving the currently selected nozzle into position, something not as easily done in firmware). 

But this results in a small problem: By having nozzle offsets set up in cura, that grid of 'printable area' is created. Both of my nozzles can reach X0 and X220. But due to the grid cura created, it won't allow me to print with nozzle 1 if X  is less than 24 (because my nozzle 2 offset is -24). If appears that the all grids are based off the most negative / most positive offsets when determining which nozzle has access to which area, as opposed to assuming that nozzle 1 can hit 0,0 and everything else is in relation to that. 

Is this type of logic intentional for that grid? or is it something that ought to be looked into? (My method of fixing it would just be to de-couple the printable area from the offset and have it as 4 settings in the extruder. X_Min_Printable, X_Max_Printable, etc. The defaults would be 0 for min and the bed size for max. Then if someone needed to restrict it, they can set up that location according to their printer's specs. (likely useful for IDEX printers especially)

Ghostkeeper:
Yes, that's intentional by default. The build volume is then the build volume that can be reached by *any one* nozzle. This assumes that the build volume is limited by the gantry and print head, rather than by the build plate.

If the build volume is limited by the build plate, you can add this metadata entry to the printer definition:

```
"nozzle_offsetting_for_disallowed_areas": false,
```

It will then still produce different positions for the different nozzles, but the borders on the side of the build volume won't move with the used nozzles any more.

Ghostkeeper (approved):
These changes look fine to me. But would you still like to apply my earlier suggestion for correcting the build volume?

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
```diff
@@ -0,0 +1,95 @@
...
+                [-26, 34],
+                [-26, -32],
+                [32, -32],
+                [32, 34]
+            ]
+        },
+        
+        
+        
+        "machine_firmware_retract":  { "default_value": true },
```
konskarm, line 46: This should override the `value` and not the `default_value` (according to the failing tests).
```suggestion
        "machine_firmware_retract":  { "value": true },
```

resources/variants/MicroSwiss_Ender3_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.2mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_0.3.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.3mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_0.4.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.4mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_0.5.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.5mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_0.6.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.6mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_0.8.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 0.8mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

resources/variants/MicroSwiss_Ender3_1.0.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
+name = 1.0mm Nozzle
+version = 4
+definition = Ender3_MicroSwiss_DirectDrive
+
+[metadata]
+setting_version = 16
```
konskarm, line 7:
```suggestion
setting_version = 17
```

konskarm (changes requested):

RFBomb:
Hey thanks!

I was wondering why it was failing, but working fine in my computer.

There are a few settings that I had set up that likely shouldn’t be
defaulted to (such as my last upload where I changed the swap distance to 0
because I use Marlin’s filament Swap in the firmware settings).

That’s one of the reasons I haven’t pushed for this PR to be actually
merged.

konskarm:
Also, have a look at @Ghostkeeper's suggestion https://github.com/Ultimaker/Cura/pull/9250#issuecomment-790716112.

resources/definitions/Ender3_MicroSwiss_DirectDrive.def.json
```diff
@@ -20,8 +20,11 @@
         "extruders_enabled_count": { "value": 1 },
         "machine_use_extruder_offset_to_offset_coords": { "value": true },
         
-        "machine_width": { "value": 230 },
-        "machine_depth": { "value": 230 },
+        "Machine_Dimensions_Comment_1": "Cura Does not calculate Disallowed Areas Properly. Cura shifts zero of build plate to negative-most nozzle.",
+        "Machine_Dimensions_Comment_2": "Workaround is to extend build plate by negative-most offset, then use G92 to put in offsets in start gcode.",
+        "Machine_Dimensions_Comment_3": "Since Nozzle 2 has X offset of -24, Build Width increased by 24mm.",
```
Ghostkeeper, line 25: This setting doesn't exist. Could you please remove it? Cura gets confused by these things, e.g. when plug-ins list all of the settings.

Comments can currently only be added in Git commit messages, if necessary.

jellespijker:

GregValiant:
@RFBomb This has been sitting idle for 3+ years.  Is it still valid and being worked on or can this be closed?

RFBomb:
This can be closed, my bad for not updating it. 

It also relied on some other cura fixes that were not accepted that have to do with print head offsets (I'm currently working around the issue in my settings -- I was told it 'isn't an issue' when I reported it and provided the fix). 

I've no interest in furthering development on this, feel free to close.

GregValiant:
Thank you.

PR #9255, opened 2021-02-09
setting for support_tree_max_diameter

BagelOrb:
New setting to limit the width of the base of tree support.

I was trying to optimize the hell out of one print and I figured that this setting might help.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4508,6 +4508,20 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "support_tree_max_diameter":
+                {
+                    "label": "Tree Support Trunk Diameter",
+                    "description": "The diameter of the widest branches of tree support. A thicker trunk is more sturdy; a thinner trunk takes up less space on the build plate.",
+                    "unit": "mm",
+                    "type": "float",
+                    "minimum_value": "support_tree_branch_diameter",
+                    "minimum_value_warning": "support_line_width * 5",
+                    "default_value": 15,
+                    "limit_to_extruder": "support_infill_extruder_nr",
+                    "enabled": "support_enable and support_structure=='tree'",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true
+                },
                 "support_tree_branch_diameter_angle":
                 {
                     "label": "Tree Support Branch Diameter Angle",


BagelOrb:
It took me 5 minutes to write this PR. Ask @Ghostkeeper whether he thinks this will be a valuable addition.

Ghostkeeper:
My opinion is that making the trunks wider towards the bottom generally has very little influence on the print time beyond the point where the branches are largely clustered together.
Sometimes it saves material because the branches merge together more. Sometimes it costs more material because the individual branches are wider. So while this setting makes some sense in theory, in practice the benefit is probably small due to these two opposing effects, and we'll have a setting that nobody is really going to adjust.

Just my opinion though. Testing may prove otherwise.

BagelOrb:
Thing is that currently the angle is very low to prevent the base from getting too wide.

Now a user could set the angle a lot wider, while still not getting extraordinarily wide bases.

gitbisector:
> Thing is that currently the angle is very low to prevent the base from getting too wide.
> Now a user could set the angle a lot wider, while still not getting extraordinarily wide bases.

Have my vote to put this in. I was looking for just this setting today. Needed some support very high up in a model and that is generating a lot of wasted support material right now.  Don't want to reduce down the angle too much either as that impedes support not as far up.

BagelOrb:
> Sometimes it saves material because the branches merge together more.

So with this setting you can safely increase the branch widening angle, so that we can benefit even more from this effect you mentioned.

Still this PR would benefit from some tests showing the benefits on some example models.

BagelOrb:
<!DOCTYPE html>

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 7.3.5.2 (Linux)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>


Model | Branch angle | Trunk diam | Branch diam angle |   | print time
-- | -- | -- | -- | -- | --
Debailey_x10.stl | 30 | 5 | 15 |   | 4h51
  | 40 | 5 | 10 |   | 4h09
  | 40 | 15 | 5 | defaults | 3h47
  | 40 | 999 | 5 | without feature | 3h47
  | 40 | 5 | 5 | more strict | 3h46
DJ hot girl and hair | 40 | 5 | 5 | more strict | 10h50
  |   | 999 |   | without feature | 11h06
  |   | 15 |   | defaults | 11h06


</body>

</html>


Tbh these results dont find 3D models out in the wild where the current defaults offer an improvement, but with a smaller max trunk diameter you can actually shave off quite some time. I'm not sure whether this warrants this PR to be merged.

If the branch diameter angle default would have been higher, the setting can easily protect the tree branches from getting too wide, but I don't see good reason to increase that setting default. 

Please discuss in the team and decide whether to close or merge this PR.

BagelOrb:
Perhaps it makes sense to have the base of the trunks have an angle of 5*, while the top has an angle of 15* so that they can merge faster. We could then have a cutoff diameter of something like 4mm where it changes from 15* to 5*.

BagelOrb:
@ devs see CURA-9533

rburema (approved):

Ghostkeeper:
Note that the tips of the branches currently already have an angle of 45° to close up.

PR #9270, opened 2021-02-11
Update Rigid3D definitions / Add Mucit2 definition

mehmetsutas:
Rigid3D definitions reviewed and updated.

Definition for Rigid3D Mucit2 added.


nallath:
Are you sure about these changes? They seem a bit drastic in some cases. The acceleration in z direction is set to be 10x as low as before.

mehmetsutas:
@nallath Yes. There were problems with the former values and they have been updated in order to match them with the values of loaded firmware. Thank you for checking and warning.

nallath (approved):

PR #8427, opened 2020-09-22
add Syndaver AXI machine

synsuiter:
Add Syndaver AXI Machine and Extruder Profile

PR #9277, opened 2021-02-15
Syndaver AXI 2 Machine profile

synsuiter:


konskarm:
Hmmm this machine seems to already exist in our codebase, which was added in PR https://github.com/Ultimaker/Cura/pull/8427. Not sure why Github doesn't recognize it though. 
It will be available to be used in our next Cura release (4.9). I am closing this PR since there is nothing else to be done here.

PR #9289, opened 2021-02-18
Add line width to simulation view

Ghostkeeper:
It's nice to show this when working with variable line width.

image

Implements CURA-7741.


Ghostkeeper:
Tests failing seems to be a problem from previous commits on the Arachne_rebased branch. Or rather, from the Master branch which only fail due to being outdated on Arachne.

jellespijker:
Why not merge it in the master branch?
Granted it will not show that much difference on the current engine generated GCode but we also have different line widths there.

jellespijker:
Will do a merge of master in arachne after this

Ghostkeeper:
We cherry-picked this to Master.

khani3s:
I was looking for a way to display flow rate on the simulation view. 
And looking at this PR, the implementation seems to be quite straightforward.

Could I just go ahead and implement it and open a PR?
( I can't find out if its already planned or not, and I also don't find out much information of how a feature is selected)

nallath:
I've had a quick glance over our product backlog and I couldn't find it in there. 

So feel free to create a PR for it. Chances are pretty high that it gets accepted. Adding an extra view here isn't going to burden us and there is some value to it.

fieldOfView:
Correct me if I am wrong, but getting the flow rate into simulation view is a whole lot harder than getting the line width in there, because the line width was already part of the layer data that is passed from CuraEngine to Cura but the flow rate is not.

Ghostkeeper:
> Correct me if I am wrong, but getting the flow rate into simulation view is a whole lot harder than getting the line width in there, because the line width was already part of the layer data that is passed from CuraEngine to Cura but the flow rate is not.

The flow rate can be calculated in the shader from the line width, layer height and speed, which are all available in the layer data. The speed is inaccurate since it doesn't know about accelerations and jerks, but it'll then be just as inaccurate as the "speed" colour scheme.

I had originally considered if I should add this as well since it's quite useful and not so hard, but due to the time frame in which this was implemented I didn't get that far.

PR #9346, opened 2021-02-28
feat: Adds files for Layer Atom 3 line of printers

Daniel-Kurth:
The commit adds:
- Machine definition files for the Layer One Atom 3 and Atom 3 Lite Printer
- One shared extruder definition file for the Atom 3 and the Atom 3 Lite
- A mesh file for the Atom 3 and Atom 3 Lite
- Variant files for [0.2; 0.4; 0.6]mm Nozzles for [PTFE;  full metal] hotends


Daniel-Kurth:
The test failed because I declared my own settings in the machine definition , which I then adjust via the variants file and thus these settings are not part of the fdmprinter.def.json file. The idea behind this is, that I can set limits for the retraction values for all metal and ptfe lined hot ends. This should prevent the user from setting a too high retraction value for an all metal hot end variant which will cause clogging. 

I hope this is ok

Ghostkeeper:
I don't think that individual printers adding settings is acceptable. This can cause Cura to crash if, e.g. a profile is created on your printer but then used on a different printer.

It's useful to have definitions that can add settings sometimes. We use it for a tool that guides material manufacturers to create profiles. And I also used it in the Settings Guide: https://github.com/Ghostkeeper/SettingsGuide/blob/master/resources/settings_guide_definitions.def.json
But then the new settings are on a setting stack separate from the global stack and the extruder stacks, so that the profile system is not involved.

I think our test suite should test for such usage and decline it explicitly. I'll take a look if I can add that test now.

Daniel-Kurth:
Okay thank you for giving a reason why it is not acceptable to add setting in the way I did. But from what I understand you would deem it ok if I were to create a separate settings file?

Ghostkeeper:
It's not acceptable because it can cause Cura to crash, or at least give unexpected errors.

It would not crash if the settings were defined in fdmprinter or fdmextruder. I would be against adding it there too though, because it makes the setting system more complex for everyone while the change is only needed for one printer.

Ghostkeeper (approved):
Changes look fine to me now in the code review. I'll give this a thorough test and merge it if it all works.

Ghostkeeper:
(Developers: Also requires this PR: https://github.com/Ultimaker/fdm_materials/pull/154 )

PR #9358, opened 2021-03-03
[CURA-7820] Reword the non-manifold message, add link.

rburema:
UX found that users where confused by the word non-manifold. Reword the message and add an explanatory link.

(There's still something off there, as the message doesn't always seem to show. I often have to load the clearly broken testmodel multiple times in order to see it.)

### DIFF
--- a/plugins/SolidView/SolidView.py
+++ b/plugins/SolidView/SolidView.py
@@ -67,7 +67,7 @@ class SolidView(View):
         self._xray_checking_update_time = 30.0 # seconds
         self._xray_warning_cooldown = 60 * 10 # reshow Model error message every 10 minutes
         self._xray_warning_message = Message(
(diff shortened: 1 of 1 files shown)


plugins/SolidView/SolidView.py
```diff
@@ -67,20 +67,27 @@ def __init__(self):
         self._xray_checking_update_time = 30.0 # seconds
         self._xray_warning_cooldown = 60 * 10 # reshow Model error message every 10 minutes
         self._xray_warning_message = Message(
-            catalog.i18nc("@info:status", "Your model is not manifold. The highlighted areas indicate either missing or extraneous surfaces."),
+            catalog.i18nc("@info:status", "The highlighted areas indicate either missing or extraneous surfaces. Fix your model and import it again into Cura."),
```
fieldOfView, line 70: For consistency, I think Cura uses the verb "open" instead of "import" for 3d models.

rburema: Fixed!

konskarm (approved):

PR #9359, opened 2021-03-03
JGAurora A6 3D printer profile

criss-rasc:
Adding a JGAurora A6 3D printer profile


nallath (changes requested):
There is something not quite right with these files; Only two profiles show up (0.16 & 0.2).

criss-rasc:
> 
> 
> There is something not quite right with these files; Only two profiles show up (0.16 & 0.2).
> 
> Developers see CURA-8120

I have tried to find what is wrong but without any luck. I can use some suggestion.

nallath:
You are missing the extruder quality profiles.

nallath:
Take a look at the ultimaker profiles. You only defined global quality profiles, not extruder ones.

Ghostkeeper:
The tests are now failing because of the state at which the branch was split off only, not because of the new definitions.

Ghostkeeper:
I did a squash merge of this PR here: https://github.com/Ultimaker/Cura/commit/4bf440b86efef17a535299b93e82b3baefcaf16f

So it is added now. Thanks for the update!

PR #9362, opened 2021-03-04
Update Geeetech A10 definition

TheTRCG:
Removed overwriting temperature gcode in the definition
Removed repeated gcode
Added comments for the gcode


TheTRCG:
Oh yeah I also tested this gcode on my printer

Ghostkeeper (approved):
Seems to work correctly from Cura's side. Thanks!

PR #9386, opened 2021-03-11
XTLW printers in cura base 

JediRAMA:
Hi, we have community of XTLW Climber 7 printers. I like cura (now arachne is great!!!)
Seller send cura profile for this printer. Would be great if you can add it in cura by default 

[XTLW Climber 7.zip](https://github.com/Ultimaker/Cura/files/6124705/XTLW.Climber.7.zip)

May the force be with you!


Ghostkeeper:
To create a pull request, you need to:
* Create a fork of Cura. This is then your own version of Cura that you can edit.
* Clone (=download) your fork to your computer.
* Make the changes you desire. In your case, that involves adding the printer profiles in the resources folder.
* Commit the changes using Git and push them to your fork on Github.
* Create a pull request to merge changes from your fork into the mainline Cura here.

For more detailed instructions, see: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

You now created a request for us to merge the changes we made for Cura 4.9 into the branch for 4.8. We're not going to do that, so I'm closing this PR to prevent accidental damage.

Ghostkeeper:
I also ran your profiles through our automated tests to give you some better feedback on the actual changes, and there are a number of issues you'll need to fix before we can merge them:
* We limit the file size of build plate meshes to 1MB each. You'll have to reduce the size or the resolution of those STLs. This is to limit the download and installation size of Cura. There are some 400 printers in Cura so if each has a 20MB mesh like you we'd already have a download/installation size of 8GB just on those 3D models.
* There's a `category` metadata entry, which is no longer allowed. The metadata entry is removed.
* There's an `id`, which is no longer allowed.
* Extruders need to refer to an ID of a printer, not a name.
* `hsa_materials` -> `has_materials`
* `retraction_hop_height_after_extruder_switch` doesn't exist.
* You're overriding some `default_values` for a setting which already has a `value`, so that override has no effect. This makes me wonder if you actually created those profiles and tested them, or just blindly copied them from elsewhere. Because you can't possibly have found this to work.
* `setting_version` is outdated. It should be 16 and the upgrade for 4.8 should be applied. This makes me doubt you actually tested them with Cura 4.8.
* Please put the quality files in a separate subfolder for the printer, like the other profiles do.

PR #9400, opened 2021-03-14
New Updated Profiles for Deltacomb 3D Printers Series

kaleidoscopeit:
- Improved profiles
- Introducing "Intents"
- Some fixes


resources/quality/deltacomb/ABS/deltacomb_DBE0.25_ABS_A.inst.cfg
```diff
@@ -6,12 +6,12 @@ definition = deltacomb_base
 [metadata]
 setting_version = 16
 type = quality
-quality_type = a
+quality_type = D005
```
Ghostkeeper, line 9: Changing the quality type will destroy the user's custom profiles, which depends on the quality type, so this requires a version upgrade.

Ghostkeeper (changes requested):
Thanks for the update! Very few printers will have a set of profiles as expansive as these. And you've fixed a lot of things that we should've seen in a code review earlier...

Let me know if you need help with the version upgrade. There is already a version upgrade plug-in for upgrading from Cura 4.9 to 4.10, so it can just be added to that one.

resources/quality/deltacomb/ABS/deltacomb_DBE0.25_ABS_A.inst.cfg
kaleidoscopeit: Really thankyou Ghostkeeper for your support, but as for now, Deltacomb users are limited and directly supported by me, then a version upgrade is not mandatory. From now on, I'll not change the quality_type naming any more.

Ghostkeeper: It's not just you. If their configuration gets corrupt that'll be seen as a bug in Cura by the user. We can't just say "oh blame Kaleidoscopeit" then, since we (attempt to) guard the quality of the application.

We'll need to transfer the user's old profiles to the new profile type names, or not have new names.

kaleidoscopeit: Ok! Well, can you give me some advice where I can find, in the source code, a configuration file which drives the upgrade plugin? Thankyou

kaleidoscopeit: Well, I tried to understand how to configure the Version Upgrade plugin and I made a small change, hoping to help in some way. Furthermore, hoping not to be annoying, I've pushed some "last minute" updates I working on. 
Thankyou again for your time.

Ghostkeeper: Seems you already found where to make the changes. You mimicked the implementation that was done for the TwoTrees profiles, which had a similar change.

Further setting changes are no problem at all. And I see there are some merge conflicts. I should be able to resolve those pretty easily.

Ghostkeeper (approved):
Looks good from the code review. I'll test these changes thoroughly.

Ghostkeeper:
All of the profiles are loaded correctly as far as I can tell. The version upgrade also successfully retained all my settings and custom profiles I made for Deltacomb printers to test with. Well done!

This will be included in the next release of Cura.

kaleidoscopeit:
Thank you Ghostkeeper

PR #9417, opened 2021-03-18
FLSUN Q5 Printer profile

kreuzhofer:
Add FLSUN Q5 printer profile with working startup and end gcodes. Tested it and it works fine.


Ghostkeeper (approved):
Looks good to me. I'll give it a quick test.

Ghostkeeper:
Thanks for the profile!

PR #9421, opened 2021-03-19
Add AXI2 machine profile

synsuiter:
This machine is slightly different from the Syndaver AXI Original. We would like it added into the Cura Ultimaker Supported Printers option.

### DIFF
--- a/resources/definitions/syndaveraxi2.def.json
+++ b/resources/definitions/syndaveraxi2.def.json
@@ -1,0 +1,46 @@
+{
+    "version": 2, 
+    "name": "SyndaverAXI2",
+    "inherits": "fdmprinter",
+    "metadata": 
+{
+        "type": "machine",
+        "visible": true,
+        "author": "Syndaver3D",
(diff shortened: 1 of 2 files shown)


resources/definitions/syndaveraxi2.def.json
```diff
@@ -0,0 +1,46 @@
...
+        "machine_center_is_zero": { "default_value": false },
+        "machine_head_with_fans_polygon": {
+            "default_value": [
+                [ 0, 0 ],
+                [ 0, 0 ],
+                [ 0, 0 ],
+                [ 0, 0 ]
+            ]
+        },
+        "gantry_height": { "value": "286" },
```
nallath, line 36: Are you very sure about this? This should be the distance from the tip of nozzle to the gantry system. The pictures of the machine that i've found show that it's not that distance.

synsuiter: @nallath Yes, it looks like you are correct. I will get that fixed and pushed y'alls way.  Thank you for the catch.

PR #9433, opened 2021-03-23
Premultiplied alpha blending

Ghostkeeper:
This implements a change to our rendering system. It is a good change under water, I think, but it has some unexpected consequences, so I'm asking for your opinions on this. Because the final result is better in some ways but worse in others.

Current/old situation
----
A while ago I [changed the opacity of support to 50%](https://github.com/Ultimaker/Cura/pull/7251). I saw this as an "easy win", low-hanging fruit, which wasn't implemented perfectly but it was such a simple change to make and provided a marginal improvement so it was okay. My colleagues agreed and it was merged.

However this implementation had some problems. You could indeed see the shape of the solid model through the support, but what you were really looking at is the "shadowy" silhouette of the input 3D mesh being rendered behind the layer view.
image
This looks weird for two main reasons:
* If the solid parts of layer view differs significantly from the input 3D mesh, you'd see the difference. This is very easy to see in the brim which is not shown at all behind the support.
* The shadowy silhouette of the input mesh still had transparency by itself, and so the pass over layer view actually made the solid part of layer view transparent. You could faintly see the build plate grid through the solid parts of layer view.

This fix
----
In this branch, and an accompanying branch in Uranium, I implemented pre-multiplied alpha blending.

image

For a theory behind pre-multiplied alpha blending, see https://apoorvaj.io/alpha-compositing-opengl-blending-and-premultiplied-alpha/
In my own words: Pre-multiplied alpha blending multiplies the colour of the image with the alpha channel, which would make the image darker for translucent objects. It then changes the blend function in OpenGL accordingly to restore the brightness of the original colour. (The blend function is essentially the formula that OpenGL will use to blend multiple rendered images together with alpha channels, usually something like `L_out = L1 * alpha + L2 * (1 - alpha)`.) The default blend function of OpenGL is intended for when the output is always fully opaque, but doesn't work well if this is not the case. Because we were rendering with Uranium's composite passes feature, our render results weren't completely opaque if support was transparent. This pre-multiplied alpha technique does work if the output is translucent.

However this then does require a change to the shaders: All output needs to be pre-multiplied with the alpha channel. As far as I could see there are three shaders in Cura and Uranium that may render translucent objects. So this pull request changes all of those. If we add more shaders in the future we'll need to do that too. You'll quickly see that the image becomes white or too bright if you don't.

The result looks impressive and pretty accurate as far as I can tell. Pre-multiplied alpha is in theory not accurate if there are multiple layers on top of each other with different opacities. In practice, since only our support is transparent in layer view, and only the modifier meshes are transparent in solid view, this hardly occurs. Other parts such as the aforementioned shadowy silhouette are drawn on a completely different layer and then blended later so they don't affect it.

One thing that I did notice is that the nozzle mesh gets rendered with incorrect colour if it's in front of the semi-transparent lower layers: 
image
This is because the nozzle mesh is very transparent but the lower layers are almost-but-not-quite-100% opaque. I don't consider this to be a huge issue since nobody really needs to see a lot of detail in the nozzle mesh.

Performance
----
I could really only theorise about the performance impact of this change. My GPU is modern-ish (NVidia RTX2070) so I noticed no change in performance even with very heavy models. Cura also doesn't keep rendering as fast as it can; it only updates if the scene changes, so just measuring frame rate is also not feasible.

The pre-multiplied alpha itself hardly has a performance impact at all. It is one extra colour multiplication per pixel, which is not significant.

However when rendering solid objects, we tell OpenGL to mask pixels behind other pixels. It calculates the colour (since it calculates all fragments in parallel), but then masks it since it needs to be behind another object. This doesn't result in more fragments needing to render, but it does mean that the render batch needs to be split up into a translucent bit and a non-translucent bit even though it is one mesh.
I implemented this in the geometry shader. The geometry shader doesn't emit any vertices for translucent bits when rendering the solid bits, and doesn't emit any vertices for solid bits when rendering the translucent bits. This means that the vertex shader and this check for translucency gets calculated twice (=an extra matrix multiplication and a few if checks for each vertex). And the number of render batches increases by 2. We already have a dozen or so in each frame, but this now increases by 2, so there is more to blend in the composite. All in all, the performance impact shouldn't be huge compared to rendering a full frame, but it could have a measurable impact.

Problems with the result
----
However this change does come with some unforeseen consequences which is why I'd really like to hear your opinions about it.

As it turns out, a normal print really does have a LOT of layers. And all of these layers add up to become pretty much 100% opaque if you look at it from the wrong angle. This is what it really means to have transparent layer view. It should not come at a surprise. But looking at it in person is something else entirely of course.
image

Secondly, having all this transparency also makes the layer view quite "blurry" from some angles. It's sometimes harder to make out how the support is structured without looking at it from multiple angles. I think this can be fixed by doing something with the shaders but I'm currently not really sure how, so I'd like to get your opinions first.

image

This PR is accompanied by a PR in Uranium: https://github.com/Ultimaker/Uranium/pull/677

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -180,14 +180,19 @@ class SimulationPass(RenderPass):
                     prev_line_types = prev_line_types[0:layer_data._attributes["line_types"]["value"].size]
                     layer_data._attributes["prev_line_types"] =  {'opengl_type': 'float', 'value': prev_line_types, 'opengl_name': 'a_prev_line_type'}
 
-                    layers_batch = RenderBatch(self._current_shader, type = RenderBatch.RenderType.Solid, mode = RenderBatch.RenderMode.Lines, range = (start, end), backface_cull = True)
-                    layers_batch.addItem(node.getWorldTransformation(), layer_data)
-                    layers_batch.render(self._scene.getActiveCamera())
+                    # Draw the layer view twice, first just the solid bits, then the transparent bits.
+                    for transparent_pass in [0, 1]:
+                        self._layer_shader.setUniformValue("u_transparent_pass", transparent_pass)
+                        blend_mode = RenderBatch.BlendMode.Normal if (transparent_pass == 1) else RenderBatch.BlendMode.NoBlending
+                        render_type = RenderBatch.RenderType.Transparent if (transparent_pass == 1) else RenderBatch.RenderType.Solid
+                        layers_batch = RenderBatch(self._current_shader, type = render_type, blend_mode = blend_mode, mode = RenderBatch.RenderMode.Lines, range = (start, end), backface_cull = True)
+                        layers_batch.addItem(node.getWorldTransformation(), layer_data)
+                        layers_batch.render(self._scene.getActiveCamera())
 
-                    # Current selected layer is rendered
-                    current_layer_batch = RenderBatch(self._layer_shader, type = RenderBatch.RenderType.Solid, mode = RenderBatch.RenderMode.Lines, range = (current_layer_start, current_layer_end))
-                    current_layer_batch.addItem(node.getWorldTransformation(), layer_data)
-                    current_layer_batch.render(self._scene.getActiveCamera())
+                        # Current selected layer is rendered
+                        current_layer_batch = RenderBatch(self._layer_shader, type = render_type, blend_mode = blend_mode, mode = RenderBatch.RenderMode.Lines, range = (current_layer_start, current_layer_end), backface_cull = True)
+                        current_layer_batch.addItem(node.getWorldTransformation(), layer_data)
+                        current_layer_batch.render(self._scene.getActiveCamera())
 
                     self._old_current_layer = self._layer_view._current_layer_num
                     self._old_current_path = self._layer_view._current_path_num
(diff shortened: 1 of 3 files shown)


fieldOfView:
Personally I don't really see the benefit of transparency, given the disadvantages. Not with the current implementation (which can make it seem parts of objects will not get printed because they are hidden behind supports that show the buildplate instead of the model through the supports), and not with the proposed implementation which makes the supports hard to "read" and removes the transparency for many support parts that overlap as seen in the screenshots.

Ghostkeeper:
Thanks for the input! What would you see as the ideal situation? No transparency at all?

I can imagine that it is still easier to see what's inside of the support if all of the transparent parts would be drawn as a single transparent layer on top of the solid parts. So the transparent parts don't sum up with each other (so they are then just as easy to read as they are now) but they are still transparent w.r.t. solid parts.

fieldOfView:
Given the two options that are there now (current and this PR), I would prefer no transparency at all. 

I think the solution you propose (" all of the transparent parts would be drawn as a single transparent layer, so the transparent parts don't sum up with each other ") would be the closest to your original intention of being able to see through the supports while still being able to see the supports. But not as low hanging fruit as originally envisioned.

Ghostkeeper:
I'd like to keep the branch open for a while to see if I can make a reasonably simple implementation some day that would implement that. As it is now I'm convinced that the current state of this branch is not an improvement but it may be the basis for an improvement.

This rendering system is quite complex and I still don't really understand the nuances in how it orders and groups things.

I'll keep this PR open for a while longer too to allow other people to also give feedback if they want.

PR #9434, opened 2021-03-23
Cura 8094 onboarding flow

jellespijker:
Updated two screens in the onboarding flow.

image
image

If you want to ensure that the onboarding flow is shown, remove the content of the local storage folder, but not the folder itself.


resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -135,16 +176,28 @@ Item
...
-                id: createAccountButton
+                id: createAccountLabel
+                width: parent.width
                 anchors.horizontalCenter: parent.horizontalCenter
-                text: catalog.i18nc("@button","Create account")
-                onClicked: Qt.openUrlExternally(CuraApplication.ultimakerCloudAccountRootUrl + "/app/create")
+                horizontalAlignment: Text.AlignHCenter
+                text:
+                {
+                    var t1 = catalog.i18nc("@text", "Create a free Ultimaker acount")
```
ChrisTerBeke, line 188: Maybe nitpick, but: Ultimaker Account is a name so Account should be capitalized as well

image

jellespijker: Thx, and I will let Lorenzo known as well

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
             spacing: 20 * screenScaleFactor
 
-            Image  // Cloud image
+            // Cloud image
+            Image
             {
                 id: cloudImage
                 anchors.horizontalCenter: parent.horizontalCenter
                 source: UM.Theme.getImage("first_run_ultimaker_cloud")
+                scale: 0.666
```
Ghostkeeper, line 78: This makes the size of this image depend on the internal coordinates of the SVG. For maintainability it would be better to explicitly give this a width or a height.
You also need to make this width or height depend on the screenScaleFactor so that the image is still in the same proportion as the rest of the interface if people use 200% screen scaling on high DPI screens or something.

The theme sizes automatically scale with the screenScaleFactor. Anything hard-coded, such as the spacing on line 70, needs to be multiplied by `screenScaleFactor`.

I realise that this also used to be an issue before your changes. It would be good to fix it anyway though now that we discover it!

resources/qml/WelcomePages/WelcomeContent.qml
```diff
@@ -7,56 +7,76 @@ import QtQuick.Controls 2.3
...
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.verticalCenter: parent.verticalCenter
+        spacing: UM.Theme.getSize("thick_margin").height
 
         Image
         {
             id: curaImage
             anchors.horizontalCenter: parent.horizontalCenter
             source: UM.Theme.getImage("first_run_welcome_cura")
+            scale: 0.666
```
Ghostkeeper, line 29: (Same issue as above: This image doesn't get scaled with the screen scale factor and its width depends on the source image.)

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
-                    full_text += "<p>" + t + "</p>"
+                    id: marketplaceColumn
+                    width: Math.round(parent.width / 3)
+                    spacing: UM.Theme.getSize("default_margin").height
+
+                    Image
+                    {
+                        id: marketplaceImage
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        source: UM.Theme.getIcon("plugin_48px")
```
Ghostkeeper, line 97: This image won't scale with the pixel scaling factor right now, so it will appear out of proportion for people on screens with different DPI.

It needs to either:
* use a size from the theme, which automatically scales,
* calculate its size from something that scales, such as line 90 `parent.width / 3`.
* explicitly multiply a hard-coded size with the `screenScaleFactor` variable.

Also, the width of the image currently depends on the source image which is bad practice. It would be more flexible in our design to give this image an explicit width.

I'm thinking of something along the line of `width: 48 * screenScaleFactor` but it should be considered if this needs to be a consistent size across the whole interface of Cura, i.e. a theme entry.

resources/themes/cura-light/icons/group_48px.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
Ghostkeeper, line 1: Please don't name vector images as to how many pixels they are meant to cover. They will not cover that many pixels on many screens. There are no pixels involved in the file, and different SVG renderers will have different pixels since they all do anti-aliasing differently.

resources/themes/cura-light/icons/plugin_48px.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
Ghostkeeper, line 1: (Same issue as above: Don't name vector images after how many pixels they were designed for.)

resources/themes/cura-light/icons/spool_48px.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
Ghostkeeper, line 1: (Same issue as above: Don't name vector images after how many pixels they are designed for.)

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
+                    }
+                    Label
+                    {
+                        id: marketplaceTextLabel
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        text: catalog.i18nc("@text", "Add material settings and\nplugins from the Marketplace")
```
Ghostkeeper, line 103: Rather than including a newline in the source text, it's better to put this label in an appropriately sized box and let it wrap automatically. Setting the `wrapMode` property to `Text.Wrap` should do the trick.

The newline won't be translated very cleanly. The translator doesn't know what sort of width it can take up per line.

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
+                {
+                    id: syncColumn
+                    width: Math.round(parent.width / 3)
+                    spacing: UM.Theme.getSize("default_margin").height
+
+                    Image
+                    {
+                        id: syncImage
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        source: UM.Theme.getIcon("spool_48px")
```
Ghostkeeper, line 121: (Same issue as above: This image doesn't get scaled with the screen scale factor and its width depends on the source image.)

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
+                {
+                    id: communityColumn
+                    width: Math.round(parent.width / 3)
+                    spacing: UM.Theme.getSize("default_margin").height
+
+                    Image
+                    {
+                        id: communityImage
+                        anchors.horizontalCenter: communityColumn.horizontalCenter
+                        source: UM.Theme.getIcon("group_48px")
```
Ghostkeeper, line 145: (Same issue as above: This image doesn't get scaled with the screen scale factor and its width depends on the source image.)

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
+                    {
+                        id: syncImage
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        source: UM.Theme.getIcon("spool_48px")
+                    }
+                    Label
+                    {
+                        id: syncTextLabel
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        text: catalog.i18nc("@text", "Backup and sync your\nmaterial settings and plugins")
```
Ghostkeeper, line 127: (Same issue as above: Don't add newlines in the source text, but put it in an appropriately sized box and let it wrap.)

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -135,16 +176,28 @@ Item
...
                 anchors.horizontalCenter: parent.horizontalCenter
-                text: catalog.i18nc("@button","Create account")
-                onClicked: Qt.openUrlExternally(CuraApplication.ultimakerCloudAccountRootUrl + "/app/create")
+                horizontalAlignment: Text.AlignHCenter
+                text:
+                {
+                    var t1 = catalog.i18nc("@text", "Create a free Ultimaker Account")
+                    var t = "<a href='https://notusedref'>" + t1 + "</a>"
+                    return t
+                }
```
Ghostkeeper, line 191: Since the whole text is the same link, what's wrong with using the `onClicked` event like before? It wouldn't need to use all that rich text formatting then.

We normally use the `TertiaryButton` QML element for this sort of thing. It is styled to look like a link. Re-using that would make it look consistent with the rest of Cura's interface regardless of theme.

resources/qml/WelcomePages/CloudContent.qml
```diff
@@ -63,64 +63,105 @@ Item
...
+                    {
+                        id: communityImage
+                        anchors.horizontalCenter: communityColumn.horizontalCenter
+                        source: UM.Theme.getIcon("group_48px")
+                    }
+                    Label
+                    {
+                        id: communityTextLabel
+                        anchors.horizontalCenter: communityColumn.horizontalCenter
+                        text: catalog.i18nc("@text", "Share ideas and get help\nfrom 48,000+ users in the\nUltimaker Community")
```
Ghostkeeper, line 151: (Same issue as above: Don't add newlines in the source text, but put it in an appropriately sized box and let it wrap.)

resources/themes/cura-light/theme.json
```diff
@@ -123,6 +128,11 @@
             "italic": true,
             "family": "Noto Sans"
         },
+        "tiny": {
+            "size": 0.85,
+            "weight": 40,
+            "family": "Noto Sans"
+        },
```
Ghostkeeper, line 135: Need to think twice about adding this new size. We've had great efforts to restrict the available font sizes to make the interface look consistent, and having many different font sizes is detrimental to that.

This is ultimately part of the design, but in an interest to keep the design consistent for the future, it would be good to reconsider this font size. Is "small" not good enough, or "default"? How is this going to look in a Japanese, Chinese or Taiwanese font?

This is something you'll need to discuss with the designer. Make sure that he knows beforehand which set of font sizes he wants to use and *then make him use it across the entire interface* instead of giving unfettered freedom to use whichever font size is the flavour of the day. This makes maintenance easier in the future and keeps the interface consistent. We currently have 5 font sizes: Small, default, medium, large and huge. Do we really want to add a 6th? I'd think that the default, which is 0.1em larger, should do fine.

(Not to mention that "tiny" is bigger than "small" apparently.)

resources/qml/WelcomePages/WelcomeContent.qml
```diff
@@ -7,56 +7,76 @@ import QtQuick.Controls 2.3
...
+                font: UM.Theme.getFont("huge_bold")
+                renderType: Text.NativeRendering
+            }
+
+            Label
+            {
+                id: textLabel
+                anchors.horizontalCenter: parent.horizontalCenter
+                horizontalAlignment: Text.AlignHCenter
+                text: catalog.i18nc("@text", "Please follow these steps to set up Ultimaker Cura.\nThis will only take a few moments.")
```
Ghostkeeper, line 53: (In this case the text contains a newline, but it would not be a problem for the translators since each sentence is on its own line. They'll just do the same. You improved the situation here. Just be aware that the lines may not be equally long in a different language so if space is tight this might become a problem.)

Ghostkeeper (changes requested):
The changes I'm requesting are mostly small and arising from inexperience of working with the cross-platform & translated QML, the way we write it.

There is one opinionated issue here, about adding a new font size which I think we should avoid. This needs some discussion, I think. The rest are all very simple changes.

resources/qml/WelcomePages/WelcomeContent.qml
```diff
@@ -7,56 +7,91 @@ import QtQuick.Controls 2.3
...
 
-    Column  // Arrange the items vertically and put everything in the center
+    // Arrange the items vertically and put everything in the center
+    Column
     {
-        anchors.centerIn: parent
-        width: parent.width
-        spacing: 2 * UM.Theme.getSize("wide_margin").height
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.verticalCenter: parent.verticalCenter
```
nallath, line 21: Nitpicking: `anchors.centerIn: parent` is a shorter way of saying that.

nallath (approved):

PR #9437, opened 2021-03-23
Add preference to enable loading multiple objects into a single instance

fieldOfView:
This PR adds a preference to the General pane of the preferences to clear the buildplate before loading a model via the "single instance" method. This way a user can choose to iterate on a single model (the current behavior, and still the default option) or to load multiple models into the single instance.

See https://community.ultimaker.com/topic/35858-command-line-to-open-an-stl-model-while-keeping-existing-ones/?tab=comments#comment-282714 and https://www.reddit.com/r/Cura/comments/mbkw2j/cant_open_multiple_models_at_the_same_time/

### DIFF
--- a/cura/SingleInstance.py
+++ b/cura/SingleInstance.py
@@ -18,6 +18,8 @@ class SingleInstance:
 
         self._single_instance_server = None
 
+        self._application.getPreferences().addPreference("cura/single_instance_clear_before_load", True)
+
     # Starts a client that checks for a single instance server and sends the files that need to opened if the server
     # exists. Returns True if the single instance server is found, otherwise False.
     def startClient(self) -> bool:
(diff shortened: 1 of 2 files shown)


nallath (approved):

nallath:
Hmm, I can't test this properly, since the single instance doesn't seem to work for me in the first place.

fieldOfView:
For me it took some restarting of Cura to get it to work.

nallath:
Ah, it's the --single-instance CLI flag that is not quite working. It does work if I do it via the preferences.

nallath:
Hmm, it's not the CLI. It's just that it only works every so often.

nallath:
Whoops. This was left for way to long. I messed something up in our tracking. 

Just to be on the safe side, i re-tested this and it works fine.

PR #9440, opened 2021-03-24
[CURA-8014] On new install or update, have highlights not just changelog.

rburema:
Place images in resources/images/whats_new/N.format and texts in resources/texts/whats_new/N.format where N is the ordinal and format the respective allowed formats.


cura/UI/WhatsNewPagesModel.py
```diff
@@ -1,22 +1,97 @@
...
 #
 class WhatsNewPagesModel(WelcomePagesModel):
 
+    image_formats = [".png", ".jpg", ".jpeg", ".gif", ".svg"]
+    text_formats = [".txt", ".htm", ".html"]
+    image_key = "image"
+    text_key = "text"
+
+    @classmethod
+    def _collectOrdinalFiles(cls, resource_type: int, include: List[str]) -> Tuple[Dict[int, str], int]:
```
nallath, line 23: Shouldn't this be a static method?

cura/UI/WhatsNewPagesModel.py
```diff
@@ -1,22 +1,97 @@
...
+                        continue
+                    page_no = int(base)
+                    highest = max(highest, page_no)
+                    result[page_no] = os.path.join(folder_path, filename)
+        except FileNotFoundError:
+            Logger.logException("w", "Could not find 'whats_new' folder for resource-type {0}".format(resource_type))
+        return result, highest
+
+    @classmethod
+    def _loadText(cls, filename: str) -> str:
```
nallath, line 42: Same goes for this one. Shouldn't this be static?

resources/qml/WelcomePages/ChangelogContent.qml
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2019 Ultimaker B.V.
```
nallath, line 1:
```suggestion
// Copyright (c) 2021 Ultimaker B.V.
```

resources/qml/WelcomePages/WhatsNewContent.qml
```diff
@@ -21,39 +25,170 @@ Item
...
+    {
+        anchors
+        {
+            top: topSpacer.bottom
+            bottom: whatsNewDots.top
+            horizontalCenter: parent.horizontalCenter
+            margins: UM.Theme.getSize("default_margin").width
+        }
+
+        height: parent.height - (UM.Theme.getSize("default_margin").height * 2)
```
nallath, line 52: Why not use anchors? This way feels a bit prone to errors.

resources/qml/WelcomePages/WhatsNewContent.qml
```diff
@@ -21,39 +25,170 @@ Item
...
+
+        StackLayout
+        {
+            id: whatsNewViewport
+
+            anchors
+            {
+                top: parent.top
+                horizontalCenter: parent.horizontalCenter
+                margins: UM.Theme.getSize("default_margin").width / 2
```
nallath, line 64: Right now this means that you only have a topMargin (since top is the only anchor that you defined). Is there a specific reason that you did that?

resources/qml/WelcomePages/WhatsNewContent.qml
```diff
@@ -21,39 +25,170 @@ Item
...
+        anchors
+        {
+            bottom: bottomSpacer.top
+            horizontalCenter: parent.horizontalCenter
+        }
+
+        delegate:
+            Rectangle
+            {
+                implicitWidth: UM.Theme.getSize("thin_margin").width
```
nallath, line 155:
```suggestion
                width: UM.Theme.getSize("thin_margin").width
```
Implicit width is for when you are creating re-usable components (and defining what it's width should be if nothing is set)

resources/qml/WelcomePages/WhatsNewContent.qml
```diff
@@ -21,39 +25,170 @@ Item
...
+                implicitHeight: UM.Theme.getSize("thin_margin").height
```
nallath, line 156:
```suggestion
                height: UM.Theme.getSize("thin_margin").height
```

nallath (changes requested):

cura/UI/WhatsNewPagesModel.py
rburema: Yup, changed, commit pending.

cura/UI/WhatsNewPagesModel.py
rburema: Also.

resources/qml/WelcomePages/WhatsNewContent.qml
rburema: Yes, the text(field) seems to go on to the bottom of the slightly-more-grey rectangle in the designs. I found out that this is for a good reason, because if the text ends up large enough to need scrolling, a grey margin without border looks really weird. I'll make it into topMargin to prevent everyone reading this code from having the same question though :-p

resources/qml/WelcomePages/WhatsNewContent.qml
rburema: Managed to decomplicate the situation a little bit (I think). Try review again? :-)

resources/qml/WelcomePages/WhatsNewContent.qml
```diff
@@ -21,39 +25,170 @@ Item
...
+
+    Rectangle
+    {
+        anchors
+        {
+            top: topSpacer.bottom
+            bottom: whatsNewDots.top
+            left: parent.left
+            right: parent.right
+            horizontalCenter: parent.horizontalCenter
```
konskarm, line 50: Left and right anchors can't work at the same time with the horizontalCenter. In this case it makes more sence to remove the horizontalCenter, since the left and right ones also define the width of the rectangle.
```suggestion
```

Issue #6939, opened 2020-01-08
Brim height setting


kyjak:
Large prints rip off the bed,even in heated chambers, due to the brim failing sometimes.

I'd like Cura to set a Brim height and to print 2 or more layers of brim along the model.

I tried a taller 1st layer, but that affects also normal print layer not only brim, not good.

All printers, any Cura version affected.

mahtDFR:
Can you give some details about the printers and Cura versions you've tried using so I can help you more effectively?
It could be solved by fine-tuning your printer's z height settings.

Ghostkeeper:
This is what the raft was designed for. Printing a brim with multiple layers makes it very hard to remove and damages the print when you do so. Instead it's better to print a raft underneath and give it some air gap so that you can remove the adhesion structure after printing.

kyjak:
I feel bad, instead of producing a script for a solution people might use in special cases like me, you try to invent why not do it. I will code the plugin myself and maybe do a PR to Cura.
As for the printer specs - I have a Prusa MK3, printing according to specs, 1st layer is fine, extruding height is low so extrusion pressure and width is high.

Twigho:
Hello,
Maybe I can suggest a little tip : 
- Put the option "skirt/Brim flow" to  120% 
- Create a Draft Shield very close to the model with a custom height

mahtDFR:
Our expectation is that a brim with multiple layers will be hard to remove. If you are able to show us the results of your tests please share with us and we can reconsider.

Twigho:
just an idea like that 
Maybe an option that generates a support block on the brim. (separate of current support)
I will explain with an example :
With materials that are subject to a high retractation, I sometimes use an imported outline as support. Which gives a thing like that :
![01](https://user-images.githubusercontent.com/52711806/72438980-5053f600-37a6-11ea-9fd0-ea31681600cd.jpg)

We can play with the options support join distance and support x/y distance (I used an the half of nozzle size)
On bottom is a support join distance 5mm
![jd5](https://user-images.githubusercontent.com/52711806/72439002-58139a80-37a6-11ea-9788-279970584100.JPG)
and here is 10mm
![jd10](https://user-images.githubusercontent.com/52711806/72439251-dcfeb400-37a6-11ea-856f-3ed2faf9047f.jpg)

It had worked well for me but I couldn't use the support properly for print hard angles. Even using "per object setting"
It had not been difficult to remove, only a few traces due to overextrusion or poor accuracy.

[...]

PR #9451, opened 2021-03-25
Create MultilayerBrim.py

Devoney:
This plugin is meant to encounter warping, by increasing the adhesion of the brim to the heated bed. It does so by dividing the single brim layer into multilayers of smaller height. It respects the original height of the brim and filament extruded, and divides it over the configurable number of layers.

### DIFF
--- a/plugins/PostProcessingPlugin/MultilayerBrim.py
+++ b/plugins/PostProcessingPlugin/MultilayerBrim.py
@@ -1,0 +1,121 @@
+# Copyright (c) 2021 Mike de Klerk
+# The PostProcessingPlugin is released under the terms of the AGPLv3 or higher.
+
+from ..Script import Script
+import re
+import json
+
+class MultilayerBrimProcessor:
+
+    layerCount = None
+    nrOfLayers = None
+
+    def __init__(self):
+        self.layerCount = 0
+
+    def changeLineParameter(self, newLine, parameter, multiplier, offset):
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
This doesn't only affect the brim though, right? It'd affect the whole first layer.

Devoney:
> This doesn't only affect the brim though, right? It'd affect the whole first layer.

That is correct. It could be renamed of course to a more general name, which is probably more correct. I named it to a use case it would probably mostly used for (or at least I think it would).

Ghostkeeper:
We've discussed this feature with the development team, but we've decided to reject it. Sorry.

Basically, this script multiplies the first layer a number of times, and reduces the thickness of these layers so that all of the multiplications fit within the space of the first layer.

In our opinion, you're better off using a raft for this case. The raft is designed to counteract warping by making multiple layers with different orientations in the lines, and has the added bonus that it moves the sensitive part further away from the build plate. This solution introduces a big flow change, too, and defeats the purpose of having a greater initial layer height for better adhesion.

We're looking for a more proper implementation of https://github.com/Ultimaker/Cura/issues/6939, in CuraEngine which is supposed to generate these structures, and using some sort of functionality to make the brim as easy to remove as it is now.

usernametakenaaaa:
Rafts are not a proper solution as they ruin the entire first layer.
In many cases you need the shiny finish of the first layer.

PR #9459, opened 2021-03-27
Cura 7813 simple exit

jellespijker:
As the comments for this method already state, anything of value should already have been saved when calling this. It may not be the prettiest workaround, but why bother cleaning things up before they go in the fires of /dev/null anyway? Now the question is ... will it work in the build?

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -603,6 +603,15 @@ class CuraApplication(QtApplication):
     @pyqtSlot()
     def closeApplication(self) -> None:
         Logger.log("i", "Close application")
+
+        # Workaround: Before closing the window, remove the global stack.
+        # This is necessary because as the main window gets closed, hundreds of QML elements get updated which often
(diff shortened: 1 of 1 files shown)


smartavionics:
If you are trying to fix the crash on exit that has been occuring on Linux, what cured it for me was this...

https://github.com/smartavionics/Uranium/commit/566c30bc14e330c82627b5351619d6c343945618

jellespijker:
He, thx! That might work indeed. @rburema the solution that @smartavionics proposes might be the better solution.

jellespijker:
Just tested it on Linux, where it works like a charm! I will create a new branch, So we can test it in the builds for Windows and Mac.
I know I should be frustrated that I lost 3 days trying to fix the inheritance of QObject children. But I'm not. I really love your fix @smartavionics. The best fixes are the ones with only two lines!

jellespijker:
Will close this PR in favor of PR Ultimaker/Uranium#680 and #9462

PR #9466, opened 2021-03-29
Updated eryone_thinker.def.json

Eryone:
1. We have updated the end gcode of the eryone thinker 3d printer.

2. Ensure we disable bed temp and turn off all fans.
Move the nozzle up by 0.5mm and retract to remove wipe on the model top layer.


nallath:
There are merge conflicts with this pull request. We need you to fix those first.

Eryone:
> There are merge conflicts with this pull request. We need you to fix those first.

Thanks! Got it and now fixed it.

nallath (approved):

PR #9472, opened 2021-03-30
Cura 7813 qml modules version bump

jellespijker:


konskarm (approved):

Ghostkeeper:
Updating this version number will not cause bugfixes. Qt only ships with one implementation of each module. Or two in some cases, since it sees QtQuick 1 as different from QtQuick 2.

The version number in the import statement is used only to two effects:
* It signals to Qt that the QML file may be too modern for it to understand (you can't import QtQuick.Controls 1.15 when using Qt 1.12).
* It signals which syntax is allowed to be used, if newer Qt versions introduce new syntax.

Regardless of the version, it still causes the same version of the Qt module to actually be imported and executed, and bugfixes from Qt 5.15 will still apply even when importing a .0 module. The import is really just a "minimum", if you want to look at it like this.

As such, we should use the lowest version number possible.

Ghostkeeper:
For one, this prevents me from running Cura from source using the version of Qt on Ubuntu now.

PR #9478, opened 2021-03-30
Update Maker Made Maker 300 Printer Profile

skilescm:
Modified printer settings and added G90 to pre-print gcode to eliminate printing errors.


skilescm:
We tested our printer in the beta and made these changes so everything will work properly in the full release. Please merge :)

Ghostkeeper:
The beta is feature frozen since about a week before the beta, so it'd be merged in the next release then, not 4.9. Sorry! This is something our testers need in order to get a more stable release.

nallath:
If you want to add machine quality profiles, you will also need to add quality profiles per extruder.

nallath (changes requested):
If you want to change this printer so that it has specific profiles, you will need to also add quality profiles for the extruder as well.

skilescm:
Closing due to being a duplicate of another PR.

PR #9488, opened 2021-04-01
Cura 8110 upgrader

jellespijker:
Upgrader to upgrade from version 4.8 to 4.9

It introduces versioning for settings visibility and moves relevant settings to the new top/bottom category. It also ensures that the new category is added to the preferences and if any of the settings were previously expanded in their old category. It will expand the new top/bottom category as well.

It will also update the progress script.

Both version upgrades 4.7 -> 4.8 and 4.8 -> 4.9 are now bundled categories as they should be.

Updated the preference and container stacks so we could keep the Cura version 16

See also Ultimaker/Uranium#685


jellespijker:
Tests are failing due to CURA-7180 OAUTH2 - Token storage

plugins/VersionUpgrade/VersionUpgrade48to49/VersionUpgrade48to49.py
```diff
@@ -0,0 +1,129 @@
...
+
+        result = io.StringIO()
+        parser.write(result)
+        return [filename], [result.getvalue()]
+
+    def upgradeSettingVisibility(self, serialized: str, filename: str) -> Tuple[List[str], List[str]]:
+        """
+        Upgrades setting visibility to have a version number and move moved settings to a different category
+
+        This updates the post-processing scripts with new parameters.
```
jellespijker, line 106: Will remove this tomorrow. Copy-paste left-over

jellespijker:
See also Ultimaker/Uranium#685

fieldOfView:
Somewhat related to this version upgrade work: 

What is the API version going to be? With the update of Qt, PyQt & Numpy, plugins may break.

nallath:
> Somewhat related to this version upgrade work:
> 
> What is the API version going to be? With the update of Qt, PyQt & Numpy, plugins may break.

It's something we're still discussing. It _shouldn't_ be changes that break anything, since the updates aren't major version increases either. Due to the work it demands of plugin developers when we do a major version increase, we prefer to keep those to an absolute minimum. 

So yeah. What is your view on this?

fieldOfView:
I was mainly concerned about this change: https://github.com/Ultimaker/Cura/commit/7827b36ab75ecaa9e78c0d8213f8f7280e660ae0. The PyQt update seems to be fairly major, but this change is about a deprecation warning so things should still be working. I have not yet tested.

I'll set up a new virtualenv with the new versions and test with my plugins.

Ghostkeeper:
From our own experience so far, there have been the following breakages due to the upgrade:
* A bunch of problems in the build system, mostly because the OSes we build them on were too old. Especially on Linux this was a problem with xcbkeyboard.
* A segfault on exit due to double deletion of Python & Qt held objects.
* Drop-down menus being disabled on MacOS.
* Drop-down menus holding focus on MacOS.
* A lot of deprecation warnings since Qt 5.15 (not yet in 5.14). We've fixed up most of those, to prevent the log from growing too big.

All of these are "general" issues which, if we solve them ourselves, will be solved for the plug-in developer too, except for the last one. However with the deprecation warnings, if it occurs a few times in plug-in code it's not really a problem.

jellespijker:
@fieldOfView I didn't get a chance to test it with your settings visibility plugin yet. I believe this upgrader should handle the introduction of a Top/Bottom category correctly, but it can't hurt to double check

konskarm (approved):

Issue #3598, opened 2018-03-29
Improvement Request: tweaks/enhancements to Layer View - especially toggles


joyfromprocess:
I have a number of enhancements that I'd love to see for **Layer View**. I have no idea how realistic/difficult these changes would be, but wanted to open a discussion. This is my first time submitting an enhancement request, so please let me know how I can improve future submissions.

1. **Allow the user to separately toggle the visibility of walls vs skin**. They currently both fall under "Show Shell" but there are times when I'd like to look at one without seeing the other. They are already colored separately, just not controlled separately. 
2. Make clear in the **color legend** that **retracted travel and unretracted travel are different shades of blue**. I figured this distinction out by accident, but it's a great piece of information to have. I'm guessing that many users may not realize (like me for months).
3. **Allow the user to show/hide z-seam points**. I know I can turn on coasting and that will show me via small gaps, but I don't actually use coasting. This means that I sometimes need to enable coasting to visualize my z-seam but then remember to turn it back off before the final slice.
4. **Scale the Feedrate color scheme** based on **visible** toolpaths. If I'm hiding travels, I don't want travel speed to blow out my scale. 
5. **Allow the user to toggle visibility of the models** (i.e. the light gray ghost that currently remains there regardless of other settings). Especially when I'm using a bunch of modifier meshes, it would be nice to hide them sometimes.

joyfromprocess:
I just wanted to add another use case where it would be nice to toggle model visibility (item 5 above).

This example model has a back wall that is too thin to slice, so the "Remove Empty First Layers" feature is (appropriately) dropping the print. The model itself is sitting higher on the bed though, so the gray ghost still matches that. None of this is unexpected or a "bug" - it just makes Layer View more visually confusing if model 'ghosts' can't be toggled off.

![model-ghost](https://user-images.githubusercontent.com/35975384/38139020-a87ca828-33fb-11e8-8969-fd10444e4b70.png)

fieldOfView:
Regarding your latest comment: There's a setting named "Remove Empty First Layers" in Mesh Fixes that is turned on by default.

joyfromprocess:
Thanks - I was expecting that behavior per my comment:

> ...the "Remove Empty First Layers" feature is (appropriately) dropping the print...

I was just using it as an illustration of when the model and slice may be out of alignment.

fieldOfView:
[...]

PR #9498, opened 2021-04-03
Limit the range of layer view to only the visible structures

Ghostkeeper:
This pull request limits the range of the colour schemes in layer view to only the structure types that are selected to be visible. For instance, if travel moves are not visible in layer view, their speed will not be counted towards the minimum or maximum of the colour scheme spectrum.

Previously this was a bit of a problem, especially with the speed of travel moves. Travel moves are so fast compared to other parts of the layer view, that the colour spectrum gets pushed all the way into the blue part on the left side of the spectrum, or at least very close together, making it hard to distinguish different speeds.

Before:
image

After:
image

Note in this example that the colour spectrum is now limited between 7mm/s and 70mm/s rather than between 7mm/s and 150mm/s. As a result, the visible colours are more stretched out so that there is more of a difference between them. The minimum speed is 7 (which is reached at the top of the tetrahedron due to minimum layer time). The travel moves are at 150, which is filtered out if the travel moves are not made visible. Instead, the next highest speed is used for the orange colour, 70 (for the infill lines).

This change is applied to line width, speed and layer thickness.

This PR also fixes a bug where if the colour spectrum had a width of 0, all of layer view gets rendered black. This could happen if visualising layer thickness and the initial layer height was equal to the layer height, or if all speeds are exactly equal. That bug became more prominent due to this change, for instance if you disable shell and infill and all you then see is the brim, which has a constant speed, line width and layer thickness.

Implements https://github.com/Ultimaker/Cura/issues/3598 .


plugins/SimulationView/SimulationView.py
```diff
@@ -334,36 +331,51 @@ def getExtruderOpacities(self) -> Matrix:
         return Matrix(self._extruder_opacity)
 
     def setShowTravelMoves(self, show):
```
nallath, line 333: Nitpicking: Boyscouting rules say that you should type this function

Ghostkeeper: Indeed. Done!

nallath (approved):

PR #9501, opened 2021-04-04
Handle storing None values in keyring

jellespijker:
Write an empty string as value to the keyring if None is parsed and
return a None value if an empty string was parsed from the keyring.

CURA-7180_keyring_none_value


rburema:
Type errors not the result of this change!

PR #9507, opened 2021-04-06
CURA-8127 preference-screen OSX fix

rburema:
Upgrade to newer QtQuickControls.


jellespijker:
CT are failing due to code-style mypy, which are unrelated to these changes

jellespijker:
Code looks good.
We accept that the language and themes aren't showing the selected items when first opening the preference dialog. It is also a known issue that the mouse scroll wheel doesn't work for these controls. These will be addressed before the final release in a separate ticket.

PR #9509, opened 2021-04-06
CURA-8127: Post processing script add script fix

konskarm:
Fixed by using the qt quick controls 2

CURA-8127


rburema:
Failing tests have nothing to do with this PR, accepting it.

PR #9518, opened 2021-04-07
Added "Relative Extrusion" mode support for PauseAtHeight.py script

ewaldkleefstra:
The extrusion mode is read from the primary extruder (`extruders[0]`). I am not familiar with multi-extruder setups, but I assume that the relative mode, if enabled, would be active for all extruders anyway (thus validating the use of `extruders[0]`).

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
+++ b/plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
@@ -510,8 +510,7 @@ class PauseAtHeight(Script):
                     extrusion_mode_string = "absolute"
                     extrusion_mode_numeric = 82
 
-                    extruders = list(Application.getInstance().getGlobalContainerStack().extruders.values())
-                    relative_extrusion = extruders[0].getProperty("relative_extrusion", "value")
+                    relative_extrusion = Application.getInstance().getGlobalContainerStack().getProperty("relative_extrusion", "value")
                     if relative_extrusion:
                         extrusion_mode_string = "relative"
                         extrusion_mode_numeric = 83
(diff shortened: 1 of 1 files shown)


plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -507,7 +507,16 @@ def execute(self, data: List[str]) -> List[str]:
                     else:
                         Logger.log("w", "No previous feedrate found in gcode, feedrate for next layer(s) might be incorrect")
 
-                    prepend_gcode += self.putValue(M = 82) + " ; switch back to absolute E values\n"
+                    extrusion_mode_string = "absolute"
+                    extrusion_mode_numeric = 82
+
+                    extruders = list(Application.getInstance().getGlobalContainerStack().extruders.values())
+                    relative_extrusion = extruders[0].getProperty("relative_extrusion", "value")
```
Ghostkeeper, line 514: `relative_extrusion` is a global setting. It cannot be changed in an extruder. If it were, getting it from the first extruder would be incorrect anyway (since you don't know which extruder was active during the pause). Luckily since it's a global setting the solution is simple: Just get it from the global stack:

```suggestion
                    relative_extrusion = Application.getInstance().getGlobalContainerStack().getProperty("relative_extrusion", "value")
```

Ghostkeeper (changes requested):
The failing automated tests stem from the Master branch back when you branched off and can be ignored. They have been fixed in the meanwhile.

I'll give you the opportunity to fix this one mistake below (about getting the setting from an extruder stack), if you'll agree. Otherwise I think we'll merge this anyway and make the change ourselves.

ewaldkleefstra:
Thanks for getting back to me. I have committed your suggested changes

Ghostkeeper (approved):

Ghostkeeper:
Thanks for the fix!

dStankiewicz:
Can we have this changed also for Repetier/OctoPrint method? It seems that now it is not added to block `if pause_method == "repetier":`, only to next `elif pause_method != "griffin":`

Ghostkeeper:
Yes, that should be possible. I ported it for Repetier too.

We should look into reducing the duplication here, but that redo layers feature makes it a bit more complex.

PR #9519, opened 2021-04-07
Added typing of KeyringAttributes

jellespijker:
This should hopefully shut mypy up.

Contributes to CURA-7180


jellespijker:
CT doesn't trip over the Keyring anymore, but still fails on other types.

PR #9535, opened 2021-04-08
CURA-8093: Check client scope when upgrading from 4.8

konskarm:
CURA-8093


rburema:
Partially failing test results for push are against the branch before the typing fix, should be fixed now.

PR #9538, opened 2021-04-08
[CURA-7180] Explicitly give keyring-backend on Mac too.

rburema:


cura/OAuth2/KeyringAttribute.py
```diff
@@ -18,6 +18,9 @@
     import win32timezone
     from keyring.backends.Windows import WinVaultKeyring
     keyring.set_keyring(WinVaultKeyring())
+if Platform.isOSX() and hasattr(sys, "frozen"):
+    from keyring.backends.OS_X import Keyring
```
konskarm, line 22: According to https://keyring.readthedocs.io/en/latest/history.html#v22-0-0 the OS_X is depreceted since v22.0.0 and we should use macOS instead (to avoid the deprecation warning).

```suggestion
    from keyring.backends.macOS import Keyring
```

PR #9539, opened 2021-04-08
adding tripod buildplate shape and assigning it to dagoma delta printers

DethCount:
allows to print 270mm instead of 200mm


Ghostkeeper:
Maybe just "triangular" would be a better description for the shape of the build plate?

A new build plate shape also needs to be implemented in CuraEngine. This PR alone is not going to provide support for triangular build plates.

DethCount:
ok for triangular.

I spent 2 days trying to build Cura on Mac OS but i couldn't make it work.
I'm getting issues with python sip import while on cura-build-environment
container.

Le jeu. 15 avr. 2021 à 23:30, Ghostkeeper ***@***.***> a
écrit :

Ghostkeeper:
Sip is always a bit of a difficult one to deal with in the build system, for some reason. Myself I've just compiled Sip from source and then it's working fine by pointing CMake to the correct directories.

nallath:
I'm closing this due to inactivity.

PR #9546, opened 2021-04-09
CURA-8167: Trim the changelog to show release notes of versions >=4.6.0

konskarm:
The release notes of older versions can be found on GitHub.

CURA-8167


rburema (approved):
Remember to put the old (parts of the) changelog somewhere we can still refer people to it!

PR #9563, opened 2021-04-11
Updated pt_BR translation

Patola:
Fixed some boolean entries description, fixed some verb declinations and added new translations.


jellespijker (approved):
Looks good @Patola. Thank you for your help, it is appeciated.

PR #9569, opened 2021-04-12
Fix the Cura logo

fieldOfView:
This PR fixes the Cura logo which somehow got mirrored in the What's New dialog. It irks me to no end.


ChrisTerBeke:
You weren't aware yet that from next release onward, it's gonna be called Ultimaker Aruc?

PR #9574, opened 2021-04-13
add Farm 2 and Farm 2CE printer, extruder profiles, variants, quality, meshes and images.

saliery999:
Hello ultimaker team,
We glad to introduce new 3d printer family from Ukraine :) and ask to add our 2 new printers models, farm 2 and farm 2ce.
We had tested it locally on our system and its work stable, also we want add 2 images to platform (mesh) file but we don't know where correct this do so i attach how they must be.

farm2 example plate
farm2ce example plate
quality folders
test 1
test 2


Ghostkeeper:
These look like the printers being added here: https://rozetka.com.ua/3d_printers/c1593467/producer=farm-3d-llc/

farm2.def.json
```diff
@@ -0,0 +1,57 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "Farm 2" },
+        "machine_start_gcode" : {
+            "default_value": "G28 ;Home\nG1 Y0 Z3.0 F4000 ;Move the platform\nG92 E0\nG1 F1000 Z0.2\nG1 X190 F1000 E50 Z0.2\nG92 E0"
+        },
+        "machine_end_gcode" : {
+            "default_value": "M104 S0\nM140 S0\nG92 E1\nG1 E-1 F300\nG28\nM84\nM81"
+        },
+        "machine_width": {
+            "default_value": 209
```
Ghostkeeper, line 35: I couldn't find online what the dimensions of this printer are, but this guy claims 215x215x205mm: https://rozetka.com.ua/104976036/p104976036/comments/

I'll believe you here, but you might want to double-check.

Ghostkeeper (approved):

farm2.def.json
saliery999: hello, the correct printer value is x(203) y(199) z(197)

saliery999: its for farm2

saliery999: for farm2ce - x(209) y(199) z(213)

Ghostkeeper (changes requested):
Ah, sorry. These resource files are not in the correct positions. I was moving them for you (because that is easier for both of us) but then found out that I need your help with something still that I can't do myself.

We require the platform meshes to be less than 1MB per printer. So I'd like you to make the following changes:

* Please reduce the file size of the farm2_platform.stl . You can do this by converting it to 3MF (which is usually smaller) for instance with [my own converter application](https://github.com/Ghostkeeper/ConvertTo3mf/releases/tag/v0.1.0). Or you could actually reduce the resolution of the mesh, for instance in Blender with the Decimate mesh modifier. Or you could export it at a lower resolution from whatever CAD software you were using.
* Please move the files to their appropriate resources folders:
  * `farm2.def.json` and `farm2_CE.def.json` need to go into `resources/definitions/`.
  * `farm2CEplate.png.png` and `farm2plate.png` need to go into `resources/images/`. Perhaps also remove the duplicate `.png` from the first one while you're at it?
  * `farm2_ce_extruder_1.def.json` and `farm2_extruder_1.def.json` need to go into `resources/extruders/`.
  * All the `farm2_ce_printhead_0.*.inst.cfg` and `farm2_printhead_0.*.inst.cfg` files need to go into `resources/variants/`.

Our automated checks would have caught these mistakes if the files were in their correct folders.

farm2.def.json
Ghostkeeper: Nice, just wanted to check with you.

saliery999: ok i make changes and add new platform which is used both for 2 printers, but i not sure if they correct open in correct position so i add a screen how they would be
zone

saliery999: i mean platform offset will now not correct when i change file to 3mf

nallath:
Your profiles are still not correct. For one, the version number in the profiles is incorrect (should be 17)

saliery999:
> Your profiles are still not correct. For one, the version number in the profiles is incorrect (should be 17)

ok i think i fixed that ?

saliery999:
ok i fix variants to v17 but now cura 4.9.1 not show them anymore. they says - empty and from now cura not open platform mesh correctly, i think they separately platform into parts and choice one of them, how can i fix this ?

Ghostkeeper:
The error message when you run the automated tests is a pretty good indicator for what's wrong with them, I think. You can run the tests with `ctest --verbose`.

The extruders and variants for the Farm 2CE printer refer to a printer called `farm2_ce`. However your definition is called `farm2_CE`. Computers are case sensitive normally. The case needs to match. Our convention is to keep the profile filenames lowercase, so renaming the file to `farm2_ce.def.json` should fix that.

saliery999:
> The error message when you run the automated tests is a pretty good indicator for what's wrong with them, I think. You can run the tests with `ctest --verbose`. We also run them when reviewing these changes though.
> 
> The extruders and variants for the Farm 2CE printer refer to a printer called `farm2_ce`. However your definition is called `farm2_CE`. Computers are case sensitive normally. The case needs to match. Our convention is to keep the profile filenames lowercase, so renaming the file to `farm2_ce.def.json` should fix that.

ok done but idk how to perform test and also my variants dont work in cura anyway and about platform mesh what i write before

saliery999:
ttf
anyway got this menu not work from cura 4.9.1, all work fine in 4.8

farm2.def.json
Ghostkeeper: That would depend on the way these files are converted then. At some point I converted all existing STLs to 3MF and that went pretty painlessly for the most part: https://github.com/Ultimaker/Cura/pull/7545

Ghostkeeper:
That would be because the profiles are too modern. They won't work in 4.8 either, but you haven't tried it since updating the `setting_version` in these profiles. The profiles work fine in Cura 4.10: 
image

However the build plate mesh is not aligned. I think something has gone wrong in the conversion there. Could you align it and rotate it properly?

saliery999 (commented):
> That would be because the profiles are too modern. They won't work in 4.8 either, but you haven't tried it since updating the `setting_version` in these profiles. The profiles work fine in Cura 4.10:
> image
> 
> However the build plate mesh is not aligned. I think something has gone wrong in the conversion there. Could you align it and rotate it properly?
> 
> Also note that STL and 3MF both don't support texture coordinates. At least, 3MF does with an extension, but Cura doesn't support that 3MF extension. You'd have to use OBJ to make the texture work.s

ok i think i change it correctly. Can you tell me what about images (picture on my meshes) ? i cant test it but i have 2 pictures for 1 mesh and for 2 different machines. Also if it doing by coding, can you preview result? and its possible to try add all in cura v4.10? idk when it will be out

Ghostkeeper:
It won't be added for Cura 4.10 any more. 4.10 is feature-frozen, so that the testers can test the release without new features changing the application from underneath them. This would be for 4.11.

The printer definition contains the mesh file and the image to use as two different metadata fields. So both `farm2.def.json` and `farm2_ce.def.json` can refer to the same mesh. But they can specify different images.

The image is put on the mesh as a texture. This means that the mesh needs to contain texture coordinates, so that the application knows where on the mesh the image is supposed to be displayed, e.g. on the build plate or on the backplate (and where the "buildplate" or "backplate" is in the mesh). Your current mesh doesn't contain any texture coordinates, so it won't get displayed.

saliery999:
> It won't be added for Cura 4.10 any more. 4.10 is feature-frozen, so that the testers can test the release without new features changing the application from underneath them. This would be for 4.11.
> 
> The printer definition contains the mesh file and the image to use as two different metadata fields. So both `farm2.def.json` and `farm2_ce.def.json` can refer to the same mesh. But they can specify different images.
> 
> The image is put on the mesh as a texture. This means that the mesh needs to contain texture coordinates, so that the application knows where on the mesh the image is supposed to be displayed, e.g. on the build plate or on the backplate (and where the "buildplate" or "backplate" is in the mesh). Your current mesh doesn't contain any texture coordinates, so it won't get displayed.
> 
> > ok from now something need from me for add 3d printers ?
> > i passed all checks and all works fine in cura 4.9.1 (v16 profiles)
> 
> The build plate is aligned now. Well done!
> But the mesh doesn't have texture coordinates. So either the texture coordinates need to be fixed, or the textures need to be removed.

ok finally i make UV map and add texture, seems all work fine in cura 4.8.0
testttt

Ghostkeeper:
Yes, it's all working, at last! Sorry for the back-and-forth, but it's necessary to get it working just as nicely as the other printers in Cura. Thanks!

saliery999:
> Yes, it's all working, at last! Sorry for the back-and-forth, but it's necessary to get it working just as nicely as the other printers in Cura. Thanks!

ok finally, nice, thank you too :)

Ghostkeeper:
You seem to have removed your profiles in your fork of Cura. Perhaps accidentally. But that change came after I merged it so your profiles will still be in Cura 4.11.

saliery999:
i think i just wrong clicked revert may be, but i not remove anything, i check my fork all in it, may be just lag?

Issue #8779, opened 2020-11-16
Object Overlap using Multiply object in version 4.8.0


tjones99:
<!--
**Application version**
4.8.0

**Platform**
Windows 10 Enterprise x64 GTX 1060 GPU

**Printer**
Prusa I3 MK2

**Reproduction steps**
1) Load STL
2) select object
3) right click and multiply object 3 times

**Screenshot(s)**
![image](https://user-images.githubusercontent.com/6657463/99208509-972bc700-278e-11eb-8860-031da2a6598c.png)
![image](https://user-images.githubusercontent.com/6657463/99208537-aca0f100-278e-11eb-801c-41619c22b865.png)
![image](https://user-images.githubusercontent.com/6657463/99208543-b1fe3b80-278e-11eb-8b8f-000ebbf0c27a.png)
![image](https://user-images.githubusercontent.com/6657463/99208552-b75b8600-278e-11eb-9d05-ea7b41ee884d.png)

**Actual results**
Objwects overlap.

**Expected results**
Objects should have been spaced without overlap. 

**Project file**

[PI3MK2_Solar Panel Mount Minimal V2 (1).zip](https://github.com/Ultimaker/Cura/files/5543844/PI3MK2_Solar.Panel.Mount.Minimal.V2.1.zip)

**Log file**
[cura.log](https://github.com/Ultimaker/Cura/files/5543846/cura.log)

**Additional information**

tjones99:
Hmmm... I filled everything in... why am I seeing nothing

nallath:
We've seen this issue before, but we couldn't get reliable reproduction steps. Thanks for submitting it!

tjones99:
If there is anything else you need let me know. If it is something to do with my STL file you are welcome to that as well. It’s just a simple Tinkercad file and nothing proprietary.

Ghostkeeper:
Hmm, well I think I got it to reliably reproduce. But there's a couple of things to it:
* "Ensure models are kept apart" needs to be disabled. If this is enabled, the move-apart fixes the problem.
* It only seems to occur in certain build volumes. In my UM3 I can't reproduce the problem, but with the printer from this project file the problem occurs pretty much every time.
* The problem is related to the rotations of the model. It only happens with models that don't have a square bounding box (like the rectangle in the project file). I was able to reproduce the issue every time by simply loading a cube and scaling it non-uniformly.

tjones99:
[...]

PR #9577, opened 2021-04-13
Fix segfault when arranging without any disallowed areas

Ghostkeeper:
So libnest2d is giving a segfault when it's being triggered without any disallowed areas. This is because we then give a single disallowed area for the border around the build plate, and give this disallowed area 0 vertices. Because libnest2d doesn't do defensive coding here, it's going to crash, taking Cura along with it. The crash is in libnest2d's function `libnest2d::_Item<ClipperLib::Polygon>::rightmostTopVertex()`.

This segfault started happening recently in Cura. However it only happens when there are no disallowed areas. Steps to reproduce the error are:
- Use Custom FFF Printer.
- Set the Adhesion Type setting to 'None'.
- Load any two models.

I tracked the segfault down to this commit: https://github.com/Ultimaker/Uranium/pull/684/commits/74ddbaab4bb4737367ad2f82698ee7422305a047 . That commit is slightly mysterious in context (It would end up without `_points` otherwise, right? How did that ever work?) but looks fine to me by itself. It did give the hint that this segfault might be caused by empty polygons though. So for that reason, I figured that filtering out the empty polygons here in the arranger might fix it; and indeed it does. Passing empty polygons as fixed/disallowed areas is useless and invalid anyway. Apparently libnest2d doesn't deal well with it, but we might as well not send those polygons over then.

### DIFF
--- a/cura/Arranging/Nest2DArrange.py
+++ b/cura/Arranging/Nest2DArrange.py
@@ -75,7 +75,7 @@ def findNodePlacement(nodes_to_arrange: List["SceneNode"], build_volume: "BuildV
         # Clip the disallowed areas so that they don't overlap the bounding box (The arranger chokes otherwise)
         clipped_area = area.intersectionConvexHulls(build_plate_polygon)
 
-        if clipped_area.getPoints() is not None:  # numpy array has to be explicitly checked against None
+        if clipped_area.getPoints() is not None and len(clipped_area.getPoints()) > 2:  # numpy array has to be explicitly checked against None
             for point in clipped_area.getPoints():
                 converted_points.append(Point(int(point[0] * factor), int(point[1] * factor)))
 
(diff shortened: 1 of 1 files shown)


rburema (approved):
Devs: See internal ticket CURA-8173

rburema:
Note that, if the reproduction steps are followed, and the model is multiplied, the user will need to manually arrange again. That is, in the specific case of 'no disallowed areas' and 'multiply model', the other model(s) is/are not arranged. The user can then manually click arrange all, and it _does_ work.

We've got a lot on our plate, and it's infinitely better than crashing, so I'll report this as a separate issue to the discussion after standup on Friday.

Ghostkeeper:
Yes, that is a known issue: https://github.com/Ultimaker/Cura/issues/8779

PR #9579, opened 2021-04-14
[CURA-8004] Fix wrong 'uploaded print job' cache.

rburema:
Fix 'uploaded print-job cache' set before the mesh upload.

Since we use that to detect when the mesh is already uploaded, and thus can be reprinted, this could cause problems, since, while we do properly set it to None when an error is returned, if the request never returns to us, or if a reprint is started while the mesh is still uploading, the print-job cache could be set while the mesh wasn't actually there yet. Which could in theory have maybe caused the problems we see.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 Ultimaker B.V.
+# Copyright (c) 2021 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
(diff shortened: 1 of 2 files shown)


Ghostkeeper (approved):

Ghostkeeper:
I don't see any bugs with this any more now. Before this change I could reproduce the original problem (prints not showing up in the queue). Afterwards not; it just gives a lot of messages saying that it's still uploading and you need to wait. No flickering error messages either.

Now I need to find a way to remove dozens of prints from the queue... The butt website is rather slow.

rburema:
> website is rather slow.

It being a bit slow (and if you butt it like that ... slow digestion?) was why we noticed this in the first place :-p

PR #9586, opened 2021-04-16
Support animated images in what's new pages

Ghostkeeper:
Let's hope they don't go too crazy with the file size of these. But animated GIFs are now supported here.

Fixes CURA-8145.


rburema (approved):

fieldOfView:
Animated webp also seems to be supported by QMovie, and that may keep the size of the animated image smaller.

rburema:
@fieldOfView By 'they' we mostly mean marketing ... unless someone puts their own images in the resources folder (and then they only have themselves to blame) we've got it under control :-)

PR #9596, opened 2021-04-17
Update creality_ender4.def.json

MattCatt87:
Overall printer dimensions were entered instead of build volume.


nallath (approved):

nallath:
Good catch! Thanks for fixing it.

PR #9601, opened 2021-04-18
Add LayerFused Base printer definitions

dbjz:
Add LayerFused CoreXY printers to the selection list. Three different versions, Build plate stl with Company Logo.


dbjz (commented):
Moved extruder definition file to be in same directory as fdmextruder.def

Ghostkeeper:
The tests here seem to be failing because the branch was branched off at a moment when they were failing on Master. They aren't failing because of the changes made here :)

resources/definitions/lfextruder.def.json
```diff
@@ -0,0 +1,14 @@
+{
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "LayerFused",
```
konskarm, line 6: There is no machine "LayerFused". This should refer to the file name of one of your printer definitions).

konskarm (changes requested):
We have certain naming conventions for printer and extruder definitions. The name of the file should follow the patter `<printer_family_in_lowercase>_<printer_name_in_lowercase>.def.json`. A similar pattern is used for the extruders. So please rename the following:

- `resources/definitions/LayerFusedX201.def.json` --> `resources/definitions/layer_fused_x201.def.json`
- `resources/definitions/LayerFusedX301.def.json` --> `resources/definitions/layer_fused_x301.def.json`
- `resources/definitions/LayerFusedX501.def.json` --> `resources/definitions/layer_fused_x501.def.json`

In addition, there is no machine "LayerFused" (mentioned in the `lfextruder.def.json`). This should refer to a specific machine (e.g. `layer_fused_x201`). To fix that I suggest the following:

Make a `layer_fused_based.def.json` printer definition, from which inherit all the other printer definitions, and link your extruder to this one (the extruder `layer_fused_extruder.def.json` should mention "machine": "layer_fused_base"). This way you can take advantage of our inheritance system and define only one extruder which is used by all your machines. Please see how Creality printers make use of the inheritance system (`creality_base.def.json` and `creality_base_extruder_0.def.json`).

Also, please merge the current master branch of Cura into yours so that our tests can run again without failing for unrelated reasons.

nallath:
Closing this due to inactivity. Feel free to re-open the pull request

PR #9603, opened 2021-04-18
Renamed "starts" to "seam"

BasF0:
After watching CHEP's https://www.youtube.com/watch?v=D39rIi8zKfk he mentions that the show "Starts" option should be called "Seam". When writing it initially (https://github.com/Ultimaker/Cura/pull/8599) the terminology was a bit fuzzy to me. 

But I think the Cura way is just to call it "Seam" instead of "Starts". We could have a discussion about whether it should be plural or singular "Seams" vs "Seam" and I don't know which would be preferred, but I picked "Seam".

This PR changes it in the GUI and in all related code.

### DIFF
--- a/plugins/SimulationView/SimulationPass.py
+++ b/plugins/SimulationView/SimulationPass.py
@@ -64,7 +64,7 @@ class SimulationPass(RenderPass):
         # Use extruder 0 if the extruder manager reports extruder index -1 (for single extrusion printers)
         self._layer_shader.setUniformValue("u_active_extruder", float(max(0, self._extruder_manager.activeExtruderIndex)))
         if not self._compatibility_mode:
-            self._layer_shader.setUniformValue("u_starts_color", Color(*Application.getInstance().getTheme().getColor("layerview_starts").getRgb()))
+            self._layer_shader.setUniformValue("u_seam_color", Color(*Application.getInstance().getTheme().getColor("layerview_seam").getRgb()))
             
         if self._layer_view:
             self._layer_shader.setUniformValue("u_max_feedrate", self._layer_view.getMaxFeedrate())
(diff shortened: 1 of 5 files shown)


fieldOfView:
There is no need to rename the internal variable names. Only the user-facing text should be changed. Personally I think that the "start" is a more apt name because it only becomes a "seam" when multiple "starts" line up. But I guess many people use the term "seam".

BasF0:
Maybe. But one can still think of the 'seam' as the joining of the perimeter, even if it does not line up in multiple layers or when evaluating a single layer. 

On the other hand, what the code does really is detect perimeter starts. If there were thin features such that the start and end of the perimeter do not line up, then it is no longer a seam and only shows the start.  Not sure if thats a common thing, never had it in my prints as far as I know.

Ah well, up to Ultimaker to decide then, at least I made it easy for them with this PR.

Ghostkeeper:
I actually think that "starts" is a more accurate name for the future, because with Arachne the walls aren't always printed as loops any more. They can be polylines which aren't necessarily completely closed. This feature also shows places where a polyline starts, regardless of whether it then also ends in the place where it starts.

It's also a bit less common of a name though. Less common than "seam".

We always discuss PRs with two developers to decide whether to merge it or not. This one hasn't been discussed yet, I'm just voicing my opinion in passing.

BasF0:
Excellent, yes. I wonder: will / should polyline ends also be visualized in the same way if they don't meet the start.

Ghostkeeper:
Yeah, it visualises that white dot wherever an extrusion starts on a wall.

Ghostkeeper:
We discussed this PR with several developers and decided to reject it for two reasons:
* The starts, if they line up vertically, will often form a seam. However they don't always form a seam.
* When Arachne gets released, this will also apply to contours which are not looping, where they don't apply to any seam in any case.

Still thanks for the suggestion and for thinking along!

Issue #9531, opened 2021-04-08
Unable to rename groups/"erasers"


KenLucke:
### Application Version

4.8.0

### Platform

Mac OS 10.15.7 (Catalina)

### Printer

N/A

### Reproduction steps

Group some objects
Attempt to rename the object group or support blocker ("eraser") via Mesh Tools > Rename

### Actual results

Rename Dialog appears, allows entry, but fails to rename group or support blocker.

### Expected results

Should allow all objects - models, support blockers ("erasers"), and groups - to be renamed, for clarity in the Objects List.

[claws.3mf.zip](https://github.com/Ultimaker/Cura/files/6275222/claws.3mf.zip)

nallath:
Please report issues with the meshtools plugin with the developer of that plugin: https://github.com/fieldOfView/Cura-MeshTools

KenLucke:
Thanks, it's hard to find the plugin Github pages.

nallath:
The marketplace shows the website for most plugins. That's where I got this from.

fieldOfView:
> That's where I got this from.

You had to look it up? For shame!

nallath:
Psh, you have so much plugins and I had the marketplace open :P

fieldOfView:
I looked into the issue of renaming groups using the MeshTools plugin, and it turns out that Cura renames the group right back to `Group #n` here:
https://github.com/Ultimaker/Cura/blob/master/cura/UI/ObjectsModel.py#L155
Shall I create a PR to make that code a bit more permissive of other names (as long as there is a name), or is there a reason to force the group names to `Group #n`?

Ghostkeeper:
Fixed link to where it gets renamed (permalink this time):

https://github.com/Ultimaker/Cura/blob/4bf440b86efef17a535299b93e82b3baefcaf16f/cura/UI/ObjectsModel.py#L171

Ghostkeeper:
This seems to originate from here: https://github.com/Ultimaker/Cura/commit/c839bc175998707ec1ddddb4464298ff2884d812

As far as I can tell, the purpose of this is just to make sure the groups have some recognisable/identifiable name, rather than an empty string or all having the same name when groups are created.

If the group names are unique by default, it's fine that they get given other names. So a pull request to fix this would be welcome! Otherwise I'll propose this to the team, but I must say that I don't expect our PO to give this a lot of priority as you might understand.

fieldOfView:
I'll create a PR that does something smart with the group names then.

PR #9614, opened 2021-04-19
Allow renaming groups to something other than "Group #n"

fieldOfView:
This PR fixes renaming of groups to something other than "Group #n". Before this, a renamed group would always immediately be renamed by to the "Group #n" pattern.

Fixes https://github.com/Ultimaker/Cura/issues/9531

### DIFF
--- a/cura/UI/ObjectsModel.py
+++ b/cura/UI/ObjectsModel.py
@@ -132,9 +132,26 @@ class ObjectsModel(ListModel):
 
             is_group = bool(node.callDecoration("isGroup"))
 
+            name_handled_as_group = False
             force_rename = False
-            if not is_group:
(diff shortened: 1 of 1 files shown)


fieldOfView:
To test, us from wc-eend recommend the Mesh Tools plugin.

fieldOfView:
The test tests the old behavior, where "Group" gets renamed to "Group #1". In Cura, the group never gets named "Group" in the first place, and if it does there is still a check to make sure it is unique.

Shall I fix the test, or make sure that "Group" gets renamed to "Group #1"?

nallath:
I think the test should be fixed.

fieldOfView:
Done

nallath (approved):

PR #9626, opened 2021-04-21
Added Atom 2.0 + Atom Plus printers

v-xup6:


resources/extruders/atom2_extruder_0.def.json
```diff
@@ -0,0 +1,15 @@
+{
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "Atom 2",
```
nallath, line 6:
```suggestion
        "machine": "atom2",
```

This needs to match the ID of the machine _exactly_. The ID is defined by the filname (see https://github.com/Ultimaker/Cura/wiki/Adding-new-machine-profiles-to-Cura)

resources/extruders/atomPlus_extruder_0.def.json
```diff
@@ -0,0 +1,15 @@
+{
+    "version": 2,
+    "name": "Extruder 1",
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "Atom Plus",
```
nallath, line 6:
```suggestion
        "machine": "atomPlus",
```

Same as the other one

nallath (changes requested):

v-xup6:
changes has been made, should be correct now

resources/definitions/atom2.def.json
```diff
@@ -0,0 +1,63 @@
...
+    "machine_end_gcode":      { "default_value": "M104 S0\nG28\nG91\nG1 E-6 F300\nM84\nG90" },
+
+    "layer_height": { "default_value": 0.2 },
+    "default_material_print_temperature": { "default_value": 210 },
+    "speed_print": { "default_value": 32 },
+    "optimize_wall_printing_order": { "value": "True" },
+    "infill_sparse_density": { "default_value": 10 },
+    "brim_width":
+    {
+      "label": "Brim Width",
```
Ghostkeeper, line 35: Why are you overriding the Brim Width and Brim Line Count setting label/description/unit/etc?

If you want to adjust the minimum/maximum/etc. perhaps you should only adjust those.

As it stands, this would prevent us from making changes to these texts and properly translating those.

resources/definitions/atom2.def.json
```diff
@@ -0,0 +1,63 @@
...
+    "platform_offset": [0,0,0],
+    "machine_extruder_trains":  { "0": "atom2_extruder_0"
+    }
+  },
+
+  "overrides": {
+    "machine_name":           { "default_value": "Atom 2" },
+    "machine_shape":          { "default_value": "elliptic" },
+    "machine_width":          { "default_value": 210 },
+    "machine_depth":          { "default_value": 210 },
```
Ghostkeeper, line 19: I can't find a page where this printer is actually sold with specs (the [website](https://atom3dp.squarespace.com/) is giving 404 errors and [this store](https://www.atom3dp.com/en/store/) doesn't list it), but according to [this review](https://all3dp.com/stealth-black-layer-one-atom-2-0-delta-3d-printer-review/) the printer is 22x32cm. Was this a typo, or was it intentional that you're 10mm short of the published dimensions?

resources/definitions/atomPlus.def.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "Atom Plus",
```
Ghostkeeper, line 2: I've been unable to find any information on this printer online to verify its dimensions and such. Do you know of any source of information about this printer?

Ghostkeeper (changes requested):
The tests succeed now. Thanks for the fix!

A few more things are still unclear here though. Could you verify what has been written?

The brim width and line count settings should not be overridden like this.

resources/definitions/atom2.def.json
v-xup6: Atom 2.0 has been phased out by Layer One, so it's hard to find official data, but here's their crowdfunding [page](https://www.zeczec.com/projects/atom2), there's spec [photo](https://farm8.staticflickr.com/7533/15704847653_dbb9e03d1b_o.jpg) about three quarter way down the page.
_Even tho it's officially phased out, there's still a large user base in Taiwan that's on Atom 2.0_
I'm intentionally short 10mm, the bed of the Atom 2.0 is held down with "clips", these clips extend into the build area. When printing at the edges, it's possible that the head hits these clips and cause problems.

resources/definitions/atom2.def.json
v-xup6: I forgot where I saw this example, it was a copy and paste with a value change.
When I only have values that I want to change the profile wouldn't load, I'm not sure why.
But we could remove this segment, that's probably easier.

resources/definitions/atomPlus.def.json
v-xup6: The Plus model was never publicly released, we obtained it as testers, I believe it's only available as a custom build.

resources/definitions/atom2.def.json
Ghostkeeper: I'd expect that changing it to this should work:

```
"brim_width": { "default_value": 4 },
```

resources/definitions/atom2.def.json
Ghostkeeper: Okay, as long as it's intentional that is okay!

v-xup6 (commented):
Changes made accordingly
`"brim_width": { "default_value": 4 }`
worked perfectly, thanks.

Ghostkeeper (approved):

Ghostkeeper:
This seems to work properly now. Thanks for the contributions!

PR #9635, opened 2021-04-21
creasee_base_extruder_0.def.json

Creasee-Andy:


nallath:
It looks like you didn't provide all the files needed to setup a machine. Please have a look at https://github.com/Ultimaker/Cura/wiki/Adding-new-machine-profiles-to-Cura

PR #9652, opened 2021-04-22
Fix mypy issues with the DigitalLibrary plugin

konskarm:


plugins/DigitalLibrary/src/DigitalFactoryFileProvider.py
```diff
@@ -31,9 +31,14 @@ def run(self) -> None:
         Function called every time the 'From Digital Factory' option of the 'Open File(s)' submenu is triggered
         """
         self.loadWindow()
+        print("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAa")
```
fieldOfView, line 34: Personally I like debugging with `print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")` ;-)

konskarm: 'A's help me express my inner screaming

nallath (approved):

PR #9658, opened 2021-04-22
Add Z Position parameter to FilamentChange 

Humsie:
This PR will add a `Z position` parameter to the FilamentChange script (PostProcessingPlugin) so you can control all 3 coordinates where the printhead will park while performing the filament change.

Usefull if you have a low print and need room to extrude the old filament for example. 

Tested on Ender 3 (+octoprint)


plugins/PostProcessingPlugin/scripts/FilamentChange.py
```diff
@@ -72,6 +72,14 @@ def getSettingDataString(self):
...
                     "default_value": 0,
                     "enabled": "not firmware_config"
+                },
+                "z_position":
+                {
+                    "label": "Z Position (relative)",
+                    "description": "Extruder relative Z position. Move the print head up for filament change.",
+                    "unit": "mm",
+                    "type": "float",
+                    "default_value": 0
```
konskarm, line 82:
```suggestion
                    "default_value": 0,
                    "minimum_valiue": 0
```

konskarm (changes requested):
Just a small change. Even though you already filter against negative values, you can set the minimum value to 0 so that there is an indication in the interface

image

plugins/PostProcessingPlugin/scripts/FilamentChange.py
Humsie: Added the minimum value parameter.

konskarm (approved):

Issue #9631, opened 2021-04-21
[4.9.0] Preview crash on MacOS 10.13.6 


jwrw:
**Application version**
4.9.0

**Platform**
MacOS High Sierra (10.13.6)
iMac (27-inch Mid 2010)
Processor 2.93 GHz Intel Core i7
Memory 12GB
Graphics ATI Radeon HD 5750

**Printer**
Ender 3 Pro

**Reproduction steps**
1. Load simple model
2. Slice
3. Go to preview

**Screenshot(s)**
No screen - application froze then crashed

**Actual results**
Application crashed when trying to display the sliced model

**Expected results**
Sliced model preview

**Project file**
This has happened with all models I have tried

**Log file**
Attached

**Additional information**
Tried both an upgrade from 4.8 by leaving originals in place and dragging new app to /Applications - preview crashed
Removed 4.9
Removed all added plugins, deleted 4.8 app and removed cura from ~/Library/Application Support/ and re-copied 4.9 to /Applications - again crash on preview

__qtdiag__
Qt 5.9.7 (x86_64-little_endian-lp64 shared (dynamic) release build; by Clang 4.0.1 (tags/RELEASE_401/final)) on "cocoa" 
OS: macOS High Sierra (10.13) [darwin version 17.7.0]

__python --version__
Python 3.8.5

[cura.log](https://github.com/Ultimaker/Cura/files/6350274/cura.log)

ravgu:
I Confirm that preview is crashing on Mac OS 10.13.6 on a iMac (21.5-inch, Mid 2011) when selecting preview. However sliced a file and printed it successfully. but each time i try to preview the program crashes.

colatuttorf:
I have the same 27" iMac detailed above running the same MacOS and Cura and can report the same issue.  100% crash rate when clicking Preview, which makes the software unusable and crashes the system.

Reverted to 4.8.0 and appears functional again.

jwrw:
It's a retrograde step but if you go into Preferences>General>Viewport behaviour then you can select 'Force layer view compatibility mode', which renders each layer as a 2D plane instead of extruded 'sausages'.  It's not nice, but at least it doesn't seem to crash.  You also don't seem to get the playback option (if you ever really used that).

orkide11:
Idem pour moi.
Installation 4.8 ça marche !

otitserip:
same crash on the same system... back to 4.8.

jwrw:
[...]

PR #9682, opened 2021-04-24
Fix for crashes on Mac 10.13 in 2010 era hardware Bug #9631

jwrw:
Not sure I understand why this works - size_x/y are for the end of the line not the start

### DIFF
--- a/plugins/SimulationView/layers3d.shader
+++ b/plugins/SimulationView/layers3d.shader
@@ -296,8 +296,8 @@ geometry41core =
 
 
         if ((u_show_starts == 1) && (v_prev_line_type[0] != 1) && (v_line_type[0] == 1)) {
-            float w = v_line_dim[0].x / 2;
-            float h = v_line_dim[0].y / 2;
+            float w = size_x;
+            float h = size_y;
 
             myEmitVertex(v_vertex[0] + vec3( w,  h,  w), u_starts_color, normalize(vec3( 1.0,  1.0,  1.0)), viewProjectionMatrix * (gl_in[0].gl_Position + vec4( w,  h,  w, 0.0))); // Front-top-left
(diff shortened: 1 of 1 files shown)


jellespijker:
@evtrados could you take a look at this PR on Monday?

jwrw:
Hi - I'm looking at the shader code generating the geometry for the start blobs and I'm not sure it's right but need a more expert view.  E.g. in the myEmitVertex calls, should "v_vertex[0] + vec3( ... )" really be just "v_vertex[0]" like in all the rest of the code above.

Ghostkeeper:
Well, if this fixes the crash for you, I'd expect that the issue is with accessing the `v_line_dim[0]` field. I'm not sure if that field is properly initialised for the first line of the layer view?

Ghostkeeper:
> Hi - I'm also looking at the shader code generating the geometry for the start blobs and I'm not sure it's right but need a more expert view. E.g. in the myEmitVertex calls, should "v_vertex[0] + vec3( ... )" really be just "v_vertex[0]" like in all the rest of the code above.

If I read the online documentation correctly ([1](https://www.khronos.org/opengl/wiki/Geometry_Shader#Outputs), [2](https://www.geeks3d.com/20111117/simple-introduction-to-geometry-shader-in-glsl-part-2/) and [3](https://knowww.eu/nodes/59b8e93cd54a862e9d7e41a4), I'd even say that `f_vertex` is not used in the output of the geometry shader here at all, and we only need `gl_Position`. Indeed, if I comment the line `f_vertex = vertex;` from the `myEmitVertex` function, it still works exactly the same way for me. This change would have a bit more of a risk though, so I think the current changes are fine. We're planning to include them in a 4.9.1 release. We could look later into a change to make it more performant by omitting the `f_vertex` write but we'll leave it out of scope for now.

Ghostkeeper:
Developers, see issue CURA-8189.

Ghostkeeper (approved):
It still works for me, and the result is provably the same (bar 0.01mm) if everything would be initialised correctly.

Ghostkeeper:
Cherry-picked to 4.9.

PR #9702, opened 2021-04-26
Fix test_wrongServerResponses

gferon:
This test was failing with Python 3.9 while packaging `cura` for Fedora 34.

`authorization_service._onAuthStateChanged()` marks the user as logged-in if `success` is `True` (default value) in `AuthenticationResponse`. (I thikn) setting `success` to `False` produces the expected behavior for the test.

### DIFF
--- a/tests/TestOAuth2.py
+++ b/tests/TestOAuth2.py
@@ -46,7 +46,7 @@ NO_REFRESH_AUTH_RESPONSE = AuthenticationResponse(
     success = True
 )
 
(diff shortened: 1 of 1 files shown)


nallath:
It is a bit weird that it would fail on python 3.9 (and not on 3.8). Do you happen to have the reason that it failed? It's just my own curiosity here. The change in itself is good as far as I am concerned.

nallath (approved):

PR #9716, opened 2021-04-28
Cura 8010 new icons

jellespijker:
New Icons have been added. To both Cura and Uranium (see PR:  ultimaker/uranium#694 )

image

The naming of the icons follows company practices, icons are placed in three different folder sizes. So this is future proof when we need to dynamically scale. This can be the case for this printer symbol, less or more lines depending on the size.
image (1)

Note for reviewers the Jira Ticket contains a conversion table attached as pdf.


konskarm (changes requested):
Apart from the icons added, I would expect to see entries in the theme.json such as `icon_small`, `icon_default`, `icon_medium`, and `icon_large` (or equivalent terms) and use that to set uniform sizes for different icons across Cura.

PR #9726, opened 2021-04-29
Added enhanced Anycubic Mega S/Pro Profil

NilsRo:
Enhanced Anycubic Mega S/P profil to add material support and other optimizations compared to Anycubic Mega profil

https://github.com/NilsRo/Cura_Anycubic_MegaS_Profile


nallath:
As far as I understood from the internet, the anycubic i3 mega S is a slightly upgraded anycubic i3 mega 

Is there a specific reason that the profiles need to be fundamentally different from the I3? If not, i think it would be best to at leas let the profiles inherit from eachother (like the UM3 and UM3E do) so that improvements to one printer can also be used on the other. 

Also; if

NilsRo:
Yes, the most reason is to support material functionality of Cura.

Actually many settings like the temperature are done in the qualities so it is not easy or possible to get this overwritten. That is a general topic with the profile delivered so I changed everything to the actual Cura design. Also there are some slight changes to the printer defaults and maximum values that are not according the specifications (but this is not that important) and updates heatup (done in parallel now). As I does own a S and P I created the printer Profile for this explicitly. But it should work visa verca.

nallath (approved):

nallath:
There was one minor issue with it, but i fixed that for you in 4cba96078390afbe110ae0830091c46a36d8a702

NilsRo:
Thanks for the correction. If it is ok I can do some implementations of e.g. the Mega X as an inherited version. Actually the larger derivates are added manually but the printers components are the same. So should be very easy.

Ghostkeeper:
As long as you can test your profiles too, that's okay with us! It'll help a bunch of people getting nicer prints, I'd bet.

PR #9773, opened 2021-05-05
Update change log and what's new pages for Cura 4.9.1

Ghostkeeper:
This lists the changes for Cura 4.9.1. In several different ways.


fieldOfView:
I think 7 MB is way too large for an image that is only going to be shown once and can be skipped by the user. Animated or not.
I tried if it would be a lot smaller when converted to animated WEBP, but that only reduced it to 6 MB, which is still too large.

nallath (approved):

PR #9782, opened 2021-05-06
Rate limit the sending of materials to networked ultimakers

nallath:
Some ultimakers tend to choke if you try and send a lot of materials in one go.

Previously we would also always send the materials on the main thread, which would also make
the interface freeze a bit. That should no longer happen.

The implementation of this is not really great, but since this is a hotfix i've opted to change
as little as possible.

CURA-8224

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Network/SendMaterialJob.py
+++ b/plugins/UM3NetworkPrinting/src/Network/SendMaterialJob.py
@@ -30,7 +30,7 @@ class SendMaterialJob(Job):
         super().__init__()
         self.device = device  # type: LocalClusterOutputDevice
 
-        self._send_material_thread = threading.Thread(target=self._sendMissingMaterials)
+        self._send_material_thread = threading.Thread(target = self._sendMissingMaterials)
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):

Ghostkeeper:
Would have been nicer as a Job maybe? Is that what you meant as bad implementation?

nallath:
It's already a job, but what happens right now is that the run of the job does a request to get the material list. That callback triggers the sending of the materials. That call is done outside of the job on the main thread.

In the future we should refactor it so that figuring out what materials need to be synced is done outside of the job and that hte materials job just accepts a list of materials that need to be synced as a parameter. That way we don't have to start a thread in the materialjob class (which is a bit dirty).

Ghostkeeper:
Agreed. And agreed that that is a bigger change than necessary now.

I wonder why my tiny commit suddenly causes tests to fail in the definition files... Let's hope it's a fluke.

nallath:
The CI was a copy paste derp from me on another change. Fixed that already.

PR #9786, opened 2021-05-06
Added Mega X Profile

NilsRo:
Added Mega X inherited from new Mega S profile.
Actually simply separated by bed size.


nallath (commented):
Few nitpick comments:
The bed now looks a bit weird 
image

If you anticipate that the profiles are always the same, you can actually tell the mega_x to also inherit it the profiles from the mega_s by adding 
`"quality_definition":      "anycubic_i3_mega_s",`
to the metadata

NilsRo:
> 
> 
> Few nitpick comments:
> The bed now looks a bit weird
> image
> 
> If you anticipate that the profiles are always the same, you can actually tell the mega_x to also inherit it the profiles from the mega_s by adding
> `"quality_definition": "anycubic_i3_mega_s",`
> to the metadata

Fixed that. Have not a checked it from that perspective  (overseen that the bed is 3 dimensional with Mega printer in general). :-) Created a new mesh with correct size now. Also inherited the qualities, thanks for that hint.

I got the necessary changes for Mega X from the guides in the web like https://howchoo.com/anycubic/anycubic-mega-x-cura-setup  It only mentioned to change the bed size as necessary. But I will ask somebody to test it in real life.
It

Ghostkeeper (approved):
Looks good now. Thanks for the contribution!

PR #9410, opened 2021-03-17
Add flow rate to simulation view

khani3s:
WIP

Add Flow rate to simulation view

Flow Rate (AKA volumetric flow rate, volume flow rate, rate of fluid flow, volume velocity or volumetric speed) is used defined on 3dp as `line width * thickness * feedrate`. The unit is mm³/s

Is useful to check if your printing inside your hot end/extruder limits.

This implementation is based on #9289 and does not take into account (I think) material_flow.

TODO

- [x] Define the name of the view (currently is just Flow)
- [x] Add to I18n
- [x] Build
- [X] CI test pass

I need some guide of how to test this on my machine (Ubuntu 18.04.5 LTS). 
I tried the "docker way" and after reading all the doc multiple times (cura-build-enviroment, cura-build, cura, wiki) I can't figure out how to build it and generate the AppImage.
I follow the CI build steps and got the same build output but I couldn't find out a way to get the binary for testing on my local machine.
I already invested +8 hours on the "build" :finnadie:

PR #9799, opened 2021-05-09
Lower QtQuick.Controls version to a version that actually exists

fieldOfView:
This PR lowers the QtQuick.Controls version to a version that actually exists. Qt 5.11 uses QtQuick.Controls 2.4. QtQuick.Controls 2.11 won't exist for several years (if ever). See https://doc.qt.io/archives/qt-5.11/qtquickcontrols2-index.html#versions

This PR is against the 4.9 branch, hopefully for inclusion in 4.9.1. While it seams that the non-existing QtQuick controls version is not a problem for most installs, some people running the 4.9 branch from source or using non-official distributions would run in the problems, eg https://github.com/Ultimaker/Cura/pull/9410#issuecomment-835710782

### DIFF
--- a/resources/qml/Menus/SettingVisibilityPresetsMenu.qml
+++ b/resources/qml/Menus/SettingVisibilityPresetsMenu.qml
@@ -2,7 +2,7 @@
 // Cura is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.10
-import QtQuick.Controls 2.11
+import QtQuick.Controls 2.4
 import QtQml.Models 2.14 as Models
 
 import UM 1.2 as UM


khani3s:
I was using the Qt recommended here:

https://github.com/Ultimaker/Cura/wiki/Running-Cura-from-Source-on-Ubuntu

`PyQt5==5.10`

The requirements.txt version solved the problem:

`PyQt5==5.15.2`

nallath:
The version does exist: https://doc.qt.io/qt-5/qtquickcontrols-index.html#prerequisites

That being said, we should always try to use an as low version as possible.

fieldOfView:
My bad. It does indeed exist. Qt started to use a different versioning scheme for QtQuick controls after Qt 5.11. I'm closing this PR because the change is trivial and the issue was caused by the user having a Qt version that is too old.

PR #9803, opened 2021-05-10
Definitons and HT0.4 quality

ChronosTech:


ChronosTech:
We want to link the temperature settings (for example: "default_material_print_temperature", [...]) present in the `ressources/definitions/strateo3d.def.json` with our high temperature hotend in the `resources/variants/strateo3d_haute_temperature_04.inst.cfg`.

To do so, we followed the trick which is written on the community forum (https://community.ultimaker.com/topic/31468-change-maximum_value_warning-via-variants-file/).

Because it's creating a setting in strateo3d.def.json which isn't defined in the fdmprinter.def.json, it doesn't pass the test.
How shall we proceed to make this upgrade without issues?

Thanks a lot for your support.

Ghostkeeper:
That trick probably works for you locally. However if the new setting ends up in a profile and that profile gets loaded in a different printer, Cura would crash. That's why we disallow it. It also makes it impossible for us to track all of the custom settings down across various printers, in case we need to do a version upgrade.

So no, there is no good way to change the warning or error thresholds per nozzle. You'd have to make the warning a bit more lenient and accept that the user might fill in a wrong setting value there.

The most architecturally sound way to implement this would be to add a new machine setting for everyone, in `fdmprinter`, not just your printer, to describe the temperature tolerance of the extrusion train. I wouldn't necessarily be against adding such a setting for everyone (considering we've had complaints about the 365 degree limit Cura currently imposes by default), but it's something we'd have to discuss in the team if you would go that route.

ChronosTech:
Thanks for your reply.

We deleted the custom setting and set the temperatures (maximum value and warning value) to the highest one we propose.

resources/variants/strateo3d_haute_temperature_04.inst.cfg
```diff
@@ -0,0 +1,20 @@
+[general]
+name = Haute température 0.4
```
Ghostkeeper, line 2: We require profile names to have an English name, if the name is descriptive. The Strateo3D website calls it ["high temp"](https://en.emotion-tech.com/product/1234569352-full-high-temp-hotend-for-strateo3d-0-4mm-nozzle). Could you make that consistent?

I can make this change myself too, but it would be good to have your consent. In any case I'm trying to get this profile update in for the next Cura release.

Note that this would then also need to be changed in other references to this nozzle (the quality and material profiles), because for some reason we chose to make that refer to the user-visible name instead of the nozzle ID in the distant past...

Ghostkeeper (changes requested):
Looks nice for the rest, but there was one issue.

Nice to see that PEKK is being developed there. It's really hard to print with, because of the high temperatures, from what I've read. But cool to see that you guys got it to work!

resources/variants/strateo3d_haute_temperature_04.inst.cfg
ChronosTech: Ok, you can make this change on your own (like strateo3d_hight_temp_04.inst.cfg), but can you also do it for the other references (quality and material profiles) ? 
Or should we change it ourselves?

Ghostkeeper: I can do this for you, yeah, no problem.

Ghostkeeper:
Works like a charm!

PR #9804, opened 2021-05-10
Update whats new page

fvrmr:
Added new png for page 0.html
Updated page 1.html


Ghostkeeper (approved):
The use of PNG images for photography is sub-optimal. I'll file a separate issue to change that to move the shadow out of the image so that JPEG can be used.

PR #9805, opened 2021-05-10
CURA-8212: add version upgrader for two trees definitions

konskarm:
Adds the version upgrader for the TwoTrees Bluer definition which has been replaced in https://github.com/Ultimaker/Cura/pull/9114.

CURA-8212


resources/quality/jgaurora_a6/jgaurora_a6_0.16_optimal.inst.cfg
```diff
@@ -20,7 +20,6 @@ speed_travel = 80
 retraction_amount = 4
 adhesion_type = skirt
 acceleration_enabled = True
-adhesion_type = skirt
```
Ghostkeeper, line 23: Don't ask me how I found this change amidst 3000 files... Looked wrong to me but showing the context here in Github I see that it is duplicate and removing this line was correct.

Ghostkeeper (approved):

PR #9863, opened 2021-05-18
Adding 3DI D300

v27jain:


konskarm (changes requested):
Since the 3DI D300 and D220 are almost identical (only the `name`, `platform`, `platform_offset`, and the machine sizes are different), please consider creating a `3di_base.def.json` and in the `3di_d300.def.json` and `3di_d220.def.json` inherit from the `3di_base`. This way, you don't even need to have two different extruders. You can have a single extruder (`3di_base_extruder_0.def.json`) assigned to the `3di_base`.

You can draw inspiration about the inheritance system from other printer families, such as the creality ones.

resources/definitions/3di_d220.def.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version": 2,
+    "name": "3DI D220",
+    "inherits": "3di_base",
+    "metadata": {
+        "visible": true,
+        "platform": "3di_d220_platform.stl",
+        "platform_offset": [0, -5, 0],
```
konskarm, line 8:
```suggestion
        "platform_offset": [0, -5, 0]

```

resources/definitions/3di_d220.def.json
```diff
@@ -0,0 +1,26 @@
...
+			},
+        "machine_width": {
+            "default_value": 220
+        },
+        "machine_height": {
+            "default_value": 220
+        },
+        "machine_depth": {
+            "default_value": 250
+        },
```
konskarm, line 23:
```suggestion
        }

```

resources/definitions/3di_d300.def.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version": 2,
+    "name": "3DI D300",
+    "inherits": "3di_base",
+    "metadata": {
+        "visible": true,
+        "platform": "3di_d300_platform.stl",
+        "platform_offset": [-200, -5, 173.205],
```
konskarm, line 8:
```suggestion
        "platform_offset": [-200, -5, 173.205]

```

resources/definitions/3di_d300.def.json
```diff
@@ -0,0 +1,26 @@
...
+			},
+        "machine_width": {
+            "default_value": 300
+        },
+        "machine_height": {
+            "default_value": 300
+        },
+        "machine_depth": {
+            "default_value": 300
+        },
```
konskarm, line 23:
```suggestion
        }

```

konskarm (changes requested):
The json files cannot be decoded because of those commas, which also causes the failing tests. Once these are removed, the printers should work without issues.

konskarm (approved):

konskarm (changes requested):
It seems that the buildplate is not aligned properly in the 3DI D300

image

v27jain:
I had checked this in the very first upload. It showed the alignment correctly.
I haven't changed the bed.stl file and the bed offset value since the changes were requested. I will verify it again. Regards.

konskarm:
The platform is now perfectly aligned in the D300. I have just a small request. I cannot find anywhere a reference to the 3DI D220 printer. Could you point me to a page with its specifications? I wanted to ensure that the size is the correct one.

konskarm (approved):

Ghostkeeper:
Seems a bit like this D220 is not released yet, still a prototype printer, maybe? There's no trace of it to be found on the internet at least.

v27jain:
> Seems a bit like this D220 is not released yet, still a prototype printer, maybe? There's no trace of it to be found on the internet at least.

I was waiting for an update on our website. This is the previous version named Noisybot (Diameter 180, Height 250)

https://3deometry.com/3d-printers/3deometry/3di-noisybot/

D220 is an upgraded version of this with (Diameter 220, Height 250).
Will be updating Noisybot to D220 on our website soon.

Ghostkeeper:
We don't normally provide profiles for prototypes or other in-development or custom-built printers. I think we'll wait with this merge until that printer is at least released on the website.

v27jain:
> We don't normally distribute profiles for prototypes or other in-development or custom-built printers. I think we'll wait with this merge until that printer is at least released on the website.

Okay, I will remove the 3DI D220 version now. Will add once its released on our website.

konskarm:
Thanks for your response @v27jain. The printer 3DI 300 slices without issues so I am merging this. The 3DI D300 printer will show up in the available printers in Cura 4.12 onwards.

PR #9866, opened 2021-05-18
Fix Power-loss Recovery and update G-codes for Flying Bear Base

oducceu:
Remove start G-code from Flying Bear Ghost 5 because it was causing positioning problems during Power-loss Recovery

Update machine_start_gcode and machine_end_gcode for Flying Bear Base

Update machine_height for Flying Bear Ghost 4S / 5 to a safe value

### DIFF
--- a/resources/definitions/flyingbear_ghost_4s.def.json
+++ b/resources/definitions/flyingbear_ghost_4s.def.json
@@ -17,7 +17,7 @@
 		"machine_name": 		{ "default_value": "Flying Bear Ghost 4S" },
 		"machine_width": 		{ "default_value": 255 },
 		"machine_depth": 		{ "default_value": 210 },
-		"machine_height": 		{ "default_value": 210 },
+		"machine_height": 		{ "default_value": 200 },
 
 		"machine_steps_per_mm_x": 	{ "default_value": 80 },
 		"machine_steps_per_mm_y": 	{ "default_value": 80 },
--- a/resources/definitions/flyingbear_ghost_5.def.json
+++ b/resources/definitions/flyingbear_ghost_5.def.json
@@ -15,10 +15,9 @@
 
 	"overrides": {
 		"machine_name": 		{ "default_value": "Flying Bear Ghost 5" },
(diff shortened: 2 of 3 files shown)


resources/definitions/flyingbear_ghost_4s.def.json
```diff
@@ -17,7 +17,7 @@
 		"machine_name": 		{ "default_value": "Flying Bear Ghost 4S" },
 		"machine_width": 		{ "default_value": 255 },
 		"machine_depth": 		{ "default_value": 210 },
-		"machine_height": 		{ "default_value": 210 },
+		"machine_height": 		{ "default_value": 200 },
```
konskarm, line 20: Why are you reducing the height? The official build volume mentions the height as 210mm.

resources/definitions/flyingbear_ghost_5.def.json
```diff
@@ -15,10 +15,9 @@
 
 	"overrides": {
 		"machine_name": 		{ "default_value": "Flying Bear Ghost 5" },
-		"machine_start_gcode": 			{ "default_value": "M220 S100 ;Reset Feedrate\nM221 S100 ;Reset Flowrate\n\nG28 ;Home\n\n;Fix X0 Y0 being outside the bed after homing\nG1 Z2.0 F3000 ;Move Z Axis up\nG1 X1.3 Y4.8 ;Place the nozzle over the bed\nG92 X0 Y0 ;Set new X0 Y0\n\n;Code for nozzle cleaning and flow normalization\nG92 E0 ;Reset Extruder\nG1 Z2.0 F3000 ;Move Z Axis up\nG1 X10.4 Y20 Z0.28 F5000.0\nG1 X10.4 Y170.0 Z0.28 F1500.0 E15\nG1 X10.1 Y170.0 Z0.28 F5000.0\nG1 X10.1 Y40 Z0.28 F1500.0 E30\n\nG92 E0 ;Reset Extruder\nG1 Z2.0 F3000 ;Move Z Axis up" },
 		"machine_width": 		{ "default_value": 255 },
 		"machine_depth": 		{ "default_value": 210 },
-		"machine_height": 		{ "default_value": 210 },
+		"machine_height": 		{ "default_value": 200 },
```
konskarm, line 20: Same for the Ghost 5. The official build volume mentions the height as 210mm.

resources/definitions/flyingbear_ghost_4s.def.json
oducceu: Hi! I admire your attention to detail! 😊 

The default Z-coupler on "4S" is terrible. So, one of the most common upgrades for "4S" is a coupler like shown in the picture  image

This coupler reduces the maximum Z-axis travel to a value between 200 and 205.

When the "5" came out, the manufacturer began to supply printers with a new coupler, identical to the one shown in the picture, with the same safe Z max. Despite the official information on the build volume, even if you max tighten bed springs, 210mm on Z could never be achieved. 

If you think the decision to reduce of height should be made and applied by the user, I will remove that change.

konskarm: That's perfectly reasonable then.

konskarm (approved):

konskarm:
I am merging this, since it is a safe change. The new changes will appear in the version 4.11 onwards.

PR #9879, opened 2021-05-20
Update liquid printer

alexgrigoras:
Updated the configurations to support USB connections (used for Duet RepRapFirmware Integration), changed the initial fan value and retraction to smaller values.


Ghostkeeper (approved):
Looks good to me.

Ghostkeeper:
As far as we can see, the profiles still work correctly too (although we can't test USB connectivity without having the actual printer). Thanks for the update!

Issue #9847, opened 2021-05-16
Crash in case of authentication issues on the Ultimaker Account login


sanangeles84:
### Application Version

4.9.1

### Platform

Windows 10

### Printer

Not Important for this issue

### Reproduction steps

1. Suddenly I couldn't open Cura anymore, I got this error many times in the log 
`2021-05-16 17:40:06,932 - DEBUG - [MainThread] cura.OAuth2.AuthorizationHelpers.getAccessTokenUsingRefreshToken [61]: Refreshing the access token for [https://account.ultimaker.com]`
`2021-05-16 17:40:07,040 - DEBUG - [MainThread] cura.OAuth2.AuthorizationService._storeAuthData [250]: Attempting to store the auth data for [https://account.ultimaker.com]`
`2021-05-16 17:40:07,042 - DEBUG - [MainThread] cura.OAuth2.AuthorizationHelpers.parseJWT [114]: Checking the access token for [https://account.ultimaker.com/check-token]`
`2021-05-16 17:40:07,151 - WARNING - [MainThread] cura.OAuth2.AuthorizationHelpers.parseJWT [123]: Could not retrieve token data from auth server: {"errors": [{"id": "016dbcd4-5d8f-4e48-b6b6-4f3dcdbc949d", "code": "tokenExpired", "http_status": "403", "title": "The given token has expired.", "meta": {"error_type": "AuthError", "url": "https://account.ultimaker.com/check-token"}}]}`
  until the software decided to crash after many attempts (323 if I'm not wrong)
`2021-05-16 17:40:07,546 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [66]: An uncaught error has occurred!`

2. In the cura.cfg I found this strange null on access and refresh tokens:
`ultimaker_auth_data = {"success": true, "token_type": "Bearer", "expires_in": 600, "scope": "account.user.read connect.cluster.read connect.cluster.write cura.printjob.read cura.printjob.write drive.backup.read drive.backup.write library.project.read library.project.write packages.download packages.rating.read packages.rating.write", "received_at": "2021-05-16 16:01:04", "access_token": null, "refresh_token": null}`

3. removing completely `ultimaker_auth_data` object from the config file solved the issue

### Actual results

Crashes, each time, until I modified the cfg like explained in "Reproduction steps"

### Expected results

Maybe a plan-b to open cura without Ultimaker Account in case of authentication issues or giving the chance to login again from scratch, instead of crashing after many attempt

[...]

PR #9903, opened 2021-05-24
Fix an crash when starting Cura without a network connection

fieldOfView:
This PR fixes a crash when starting Cura without a network connection if the user was previously logged in to an Ultimaker account.

*Steps to reproduce*
* Start Cura 4.9.1
* Log in to Ultimaker account
* Close Cura
* Disconnect from all networks
* Restart Cura

Result:
Infinite loop of `AuthorizationService._parseJWT -> AuthorizationService._storeAuthData -> AuthorizationService.getUserProfile -> AuthorizationService._parseJWT -> etc`

This PR fixes https://github.com/Ultimaker/Cura/issues/9847


fieldOfView:
Fixes https://github.com/Ultimaker/Cura/issues/9847, and CURA-8255

nallath (approved):

PR #9915, opened 2021-05-27
Update Readme regarding irc network

### LINKED ISSUE #9914: Move the irc channel

**#cura** and **#cura-dev** are what's left over from my freenode journey.
Freenode recently made some drastic policy changes and should be avoided, *imho*.

I'd like to get rid of both freenode channels.
As the channel has been an annoyance ever since (#1277), we should consider moving to a different network.
Many, including freenodes staff, moved to https://libera.chat: their new platform.

I'd like cura to follow them there.

Alternative networks would include [oftc](https://www.oftc.net/), [hackint](https://hackint.org/) and [ircnet](https://www.ircnet.com/).
I'd order them like above.

Another option would be to stay with freenode or start discussions about other platforms.

There are not many users in the channels; but I found them pretty useful, when I tried to find an issue in the code a few years ago.

_No response_

AiyionPrime:
This would resolve #9914.
I think @Ghostkeeper documented the channels a few years ago, likely a good choice for a review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ For crashes and similar issues, please attach the following information:
 
 If the Cura user interface still starts, you can also reach this directory from the application menu in Help -> Show settings folder
 
-For additional support, you could also ask in the #cura channel on FreeNode IRC. For help with development, there is also the #cura-dev channel.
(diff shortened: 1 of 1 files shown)


nallath (approved):

AiyionPrime:
@nallath I think you should consider registering those channels properly.
That grants cura sovereignty over all channels that begin with `#cura-` as well as the `#cura`-channel itself.
https://libera.chat/chanreg#project-registration

PR #9930, opened 2021-06-01
Exclude plugins from backup

nallath:
Backups are synced with your account anyway. Backing up the plugins means that you
quickly reach the limit of the max size (250mb) that the backups allow.

CURA-8269

### DIFF
--- a/cura/Backups/Backup.py
+++ b/cura/Backups/Backup.py
@@ -28,6 +28,8 @@ class Backup:
     IGNORED_FILES = [r"cura\.log", r"plugins\.json", r"cache", r"__pycache__", r"\.qmlc", r"\.pyc"]
     """These files should be ignored when making a backup."""
 
+    IGNORED_FOLDERS = [r"plugins"]
+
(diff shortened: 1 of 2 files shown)


cura/Backups/Backup.py
```diff
@@ -94,8 +97,7 @@ def _makeArchive(self, buffer: "io.BytesIO", root_path: str) -> Optional[ZipFile
         :param root_path: The root directory to archive recursively.
         :return: The archive as bytes.
         """
-
-        ignore_string = re.compile("|".join(self.IGNORED_FILES))
+        ignore_string = re.compile("|".join(self.IGNORED_FILES + self.IGNORED_FOLDERS))
```
rburema, line 100: Note that ignore_string is later used to search/match through the entire absolute path. Technically that would mean that when someone installs Cura to C:/cache/ or D:/plugins/ the backup system won't work anymore.

This is already a risk w.r.t. the way it's done now, though except for cache those all contain dots or underscores. Nevertheless fixing this falls outside of the scope of this ticket, I think.

rburema (approved):

PR #9945, opened 2021-06-06
Rework calculation of min/max flow rates for simulation view colour scheme

smartavionics:
The max value was being calculated as the product of the max line width, max layer thickness
and max feedrate but it should be calculated as the maximum of the products of the width,
thickness and feedrate for each individual line. i.e. calculate the flow for each line
and use the max values of the flows. The min value calculation is similarly changed.

### DIFF
--- a/plugins/SimulationView/SimulationView.py
+++ b/plugins/SimulationView/SimulationView.py
@@ -417,10 +417,9 @@ class SimulationView(CuraView):
         return self._max_flow_rate
 
     def getMinFlowRate(self) -> float:
-        min_flow_rate = self._min_flow_rate
-        if abs(min_flow_rate - sys.float_info.max) < 10:  # Some lenience due to floating point rounding.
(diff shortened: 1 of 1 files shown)


smartavionics:
Without this PR...

Screenshot_2021-06-06_16-51-57

With this PR...

Screenshot_2021-06-06_16-54-18

Note that the variation in flow on the skins is OK and is because the skin lines are flow compensated to reduce the under-extrusion that occurs when short skin lines are printed with the lines pattern.

smartavionics:
Just noticed the min flow is showing as 0 above which is not a bug because the wall overlap compensation is producing some lines with extremely small flows. If you disable the overlap compensation and the wall gap filling the min flow increases to 0.6.

nallath:
For our reference: CURA-8294

nallath (approved):

nallath:
Looks good! I've sliced a bunch of models with it and the results are indeed correct (and the previous ones were not).

PR #9972, opened 2021-06-12
Added Kingroon K3P and K3PS definitions

NoTaMu:
Captura de pantalla (204)
Captura de pantalla (205)
Captura de pantalla (206)
Captura de pantalla (207)

Added Kingroon K3P and K3PS (plus variants).


nallath:
If the only thing that is different between the different versions is the build plate (which it seems to be), I would like to request that the extra versions are not added. 

Furthermore; It seems that this is the only printer where the head is actually modelled in the buildplate 3D model. This might actually make it a bit more unusable, as it would get in the way when looking at certain angles.

And finally; the version numbers don't match up (should be 17, not 16)

NoTaMu:
The KP3 is different to the others, among other things, because it is a bowden and the KP3S is a direct drive. The kp3s v2 and v3 are in fact identical except the hotend design, I can leave one. The v1 version is different in e_steps, I created the others because I'm not sure if they are taken into account for the time estimation.

About the meshes, I recreated the whole printers because if I only leave the bed there isn't a clear orientation indication. Would it be okay if I place the hotend and bed in a "parked" position? Also, I wouldn't like to waste all the time I spent modeling the printers, hehe.

nallath:
Well, Cura will always add an indicator in the front left corner. So it should have some info as to what is the front.

NoTaMu:
I've left only the two base models and updated the buildplates to only include the bottom part of the printers.

NoTaMu:
I've reduced the triangle count of the models and simplified them further to improve the performance on less powerful computers.

NoTaMu:
Hello again, when can I expect this to be merged? This is my first pull request and I don't know if I should do anything else...

resources/definitions/kingroon_base.def.json
```diff
@@ -0,0 +1,271 @@
...
+
+        "has_materials": true,
+        "has_variants": true,
+        "has_machine_quality": true,
+        "variants_name": "Nozzle Size",
+
+        "preferred_variant_name": "0.4mm Nozzle",
+        "preferred_quality_type": "standard",
+        "preferred_material": "generic_pla",
+        "exclude_materials": [
```
Ghostkeeper, line 24: You define a material diameter of 1.75mm. However you exclude a lot of materials here that are not compatible with that. They don't need to be excluded since Cura already filters them out.

resources/definitions/kingroon_kp3.def.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "Kingroon KP3",
```
Ghostkeeper, line 2: I couldn't find any printer that's just "KP3". On [their website](https://www.kingroon.com/collections/fdm-3d-printer-28) Kingroon shows printers called KP3S and KP3L. Was this one meant to be the KP3L? The dimensions match at least.

resources/variants/kingroon_base_0.2.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

resources/variants/kingroon_base_0.3.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

resources/variants/kingroon_base_0.4.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

resources/variants/kingroon_base_0.6.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

resources/variants/kingroon_base_0.8.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

resources/variants/kingroon_base_1.0.inst.cfg
```diff
@@ -0,0 +1,12 @@
+[general]
```
Ghostkeeper, line 1: The base printer is never meant to be visible, so it doesn't need variants. There are variants for the Kingroon KP3 and KP3S specifically.

Ghostkeeper (changes requested):
We try to review these pull requests from old to new, more or less. As you can see in the list of pull requests, yours is up next, since all PRs older than yours are waiting on the author or are otherwise under debate. It's quite common for us to have PRs wait more than a month, as we are still way behind. Though we are catching up a bit right now.

Here is my actual code review for these printer profiles. I think Nallath just glanced at it for a bit to get the most common issues out of the way. However this is a full review. Once all of these issues are handled, it's green to be merged.

Ghostkeeper:
Developers, see ticket CURA-8430.

resources/definitions/kingroon_kp3.def.json
NoTaMu: Their website is under renovation so it is a mess right now. The KP3 was the previous model, it used a bowden style system but it seems they don't sell it anymore. I think the KP3L must be an unreleased prototype, because I've not seen that model anywhere else (the design matches that of the KP3 base, but that one only has one Z arm).

resources/definitions/kingroon_base.def.json
NoTaMu: I'm having troubles finding out what materials I should and shouldn't exclude. Is there any suggestion about how to properly do this?

NoTaMu: I think I have resolved this: I cleared the list and added one by one the materials that were listed by Cura (the ones without qualities, of course). I'm waiting for your review 🙂

Ghostkeeper: The "expected" behaviour is that you'd only exclude materials that you are certain won't work. Like, exclude polycarbonate materials because they require a printing temperature of 280C which is hotter than what the printer can reach. It's an open filament system after all, so people should be able to provide new profiles for the printer. We shouldn't limit people to use a combination of profiles if they really want to use their favourite filament in the Kingroon printers.
So in theory, you'd test those profiles one by one, and exclude them only if you tested that they don't work.

Understandably resources are limited and some guesswork is needed, but that's the workflow we've envisioned.

Ghostkeeper (approved):
Looks good in the review now. I'm giving it a good test to see if anything comes up with any of the profiles, and if nothing weird is happening we can merge it.

PR #9998, opened 2021-06-18
Blocks 3D Printers

braga-dev:
Adding Blocks 3D Printers to Ultimaker Cura.

This printers are developed in Portugal by Blocks. You can check them out in www.blockstec.com


nallath:
A few preliminary remarks:

Materials should not be added in this repo. They should go in www.github.com/ultimaker/fdm_materials

Why let the global qualities depend on the nozzle size? The entire system is designed in such a way that you should not do this. 
A global quality is completely independent of the material and the variant used. 

There is no need to list all the excluded materials. The materials that you list are mostly 2.85 materials, which are already excluded by the fact that your machine has a different diameter. 

Please base the profiles off the generic PETG (and all other materials as well). Generic is a special case. So if you select blocks_PETG and it can't find a profile for it, it will default to the generic one. That way if people create their own versions of petg, they will at least use the default settings that you defined. For an example, have a look at how the Ultimaker printers handle this.

The platform files are far to big; We limit them at 1 MB, but we ask you to keep them well below that.

braga-dev:
Thank you for the feedback Nallath. We are working in solving this issues.

PR #10003, opened 2021-06-19
Updated pt_BR strings for Cura 4.10.

Patola:


jellespijker:
Thank you @Patola , greet work and really appreciated

PR #10009, opened 2021-06-21
add printer creality ender3max

navh:
add printer creality ender3max


nallath (changes requested):
This isn't enough to add a machine to Cura. Please check the results from the CI and check https://github.com/Ultimaker/Cura/wiki/Adding-new-machine-profiles-to-Cura for more info on how to add new machines to Cura

PR #10010, opened 2021-06-21
CURA-8332: Don't crash if keyring access is denied on Mac

konskarm:
Denying access to the keyring was causing Cura to crash on MacOS.

Contributes to issue CURA-8332 and sentry issue CURA-2EW

### DIFF
--- a/cura/OAuth2/KeyringAttribute.py
+++ b/cura/OAuth2/KeyringAttribute.py
@@ -4,7 +4,7 @@ from typing import Type, TYPE_CHECKING, Optional, List
 
 import keyring
 from keyring.backend import KeyringBackend
-from keyring.errors import NoKeyringError, PasswordSetError
+from keyring.errors import NoKeyringError, PasswordSetError, KeyringError
 
 from UM.Logger import Logger
 
@@ -22,6 +22,9 @@ if Platform.isOSX() and hasattr(sys, "frozen"):
     from keyring.backends.macOS import Keyring
     from keyring.backends.macOS.api import KeychainDenied
     keyring.set_keyring(Keyring())
+else:
+    class KeychainDenied(Exception):
+        pass
 
 # Even if errors happen, we don't want this stored locally:
 DONT_EVER_STORE_LOCALLY: List[str] = ["refresh_token"]
@@ -40,12 +43,9 @@ class KeyringAttribute:
                 self._store_secure = False
                 Logger.logException("w", "No keyring backend present")
                 return getattr(instance, self._name)
-            except Exception as e:
+            except KeychainDenied:
                 self._store_secure = False
-                if Platform.isOSX() and hasattr(sys, "frozen") and type(e) == KeychainDenied:
-                    Logger.log("i", "Access to the keyring was denied.")
-                else:
-                    Logger.logException("w", f"Something went wrong while trying to retrieve the password from the Keyring. Exception: {e}")
+                Logger.log("i", "Access to the keyring was denied.")
                 return getattr(instance, self._name)
         else:
             return getattr(instance, self._name)
--- a/cura/OAuth2/KeyringAttribute.py
+++ b/cura/OAuth2/KeyringAttribute.py
@@ -42,10 +42,10 @@ class KeyringAttribute:
                 return getattr(instance, self._name)
             except Exception as e:
                 self._store_secure = False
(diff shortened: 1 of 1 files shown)


cura/OAuth2/KeyringAttribute.py
```diff
@@ -4,6 +4,7 @@
 
 import keyring
 from keyring.backend import KeyringBackend
+from keyring.backends.macOS.api import KeychainDenied
```
fieldOfView: Should this be a platform-dependent import?

On Windows, I am getting the following error on this import `FileNotFoundError: Could not find module 'C:\System\Library\Frameworks\Security.framework\Versions\A\Security' (or one of its dependencies). Try using the full path with constructor syntax.`

konskarm: It should indeed. I changed it in https://github.com/Ultimaker/Cura/pull/10010/commits/951c656096e4bc94d3fdc892b432d6b65c2715f0.

nallath (changes requested):
This will fail if any exception is actually called on another operating system; The KeychainDenied is then not defined.

cura/OAuth2/KeyringAttribute.py
```diff
@@ -39,6 +40,13 @@ def __get__(self, instance: "BaseModel", owner: type) -> Optional[str]:
                 self._store_secure = False
                 Logger.logException("w", "No keyring backend present")
                 return getattr(instance, self._name)
+            except Exception as e:
```
Ghostkeeper, line 43: Catch-all exceptions are often seen as a code smell. What's wrong with just catching `KeychainDenied` with a specific message, and then catching `KeyringError` with a more general message?

This way, if I make a mistake, e.g. rename `self._keyring_name` without changing the reference to it here, then I'll at least know about it. But problems with getting the password would still be caught without crashing the application.

Ghostkeeper (changes requested):

cura/OAuth2/KeyringAttribute.py
konskarm: As Jaime noticed, if I try to catch the KeychainDenied in OS's other than OSX, it will crash since it will not be able to find it (KeychainDenied is specific to OSX). So I thought that catching a generic error and then exlpicitly checking for it is the best way.

cura/OAuth2/KeyringAttribute.py
```diff
@@ -39,6 +40,13 @@ def __get__(self, instance: "BaseModel", owner: type) -> Optional[str]:
...
+                self._store_secure = False
+                exception_message = f"Something went wrong while trying to retrieve the password from the Keyring. Exception: {e}"
+                if Platform.isOSX() and hasattr(sys, "frozen") and type(e) == KeychainDenied:
+                    exception_message = "Access to the keyring was denied."
```
Ghostkeeper, line 47: In this case it probably shouldn't log a warning. The user explicitly denied access, so it is not a recoverable problem, but intentional. More like info?

cura/OAuth2/KeyringAttribute.py
nallath: But you could just define a KeychainDenied object in the else statement. As long as it inherits from Exception, that should work.

cura/OAuth2/KeyringAttribute.py
konskarm: I changed it to log info in this case.

cura/OAuth2/KeyringAttribute.py
konskarm: Unfortunately, this breaks mypy (it considers it is a redefinition of KeychainDenied, even though it is conditional).

Ghostkeeper: I've made that suggestion as well, to define your own class if not using MacOS. But it is a bit clunky so we decided against that and accepted that we'd be catching all exceptions here then.

I did make a PR for the author to fix this issue: https://github.com/jaraco/keyring/pull/517

Ghostkeeper:
The fake exception class is still in, and still giving MyPy a [cross-popping vein](https://tvtropes.org/pmwiki/pmwiki.php/Main/CrossPoppingVeins). Did you intend to remove it?

Ghostkeeper (approved):

PR #10030, opened 2021-06-24
Export all materials to archive

Ghostkeeper:
This is a new application of #9957, which got merged to the 4.10 branch accidentally but shouldn't have been in there.

This is a feature to export all materials to a .umm archive from Cura.

Requirements, as from the ticket, are:

> requirements:
> 
> * ONE save all button
> * button to export all material profiles to a zip to specific location on disk
> * File-format should work with the firmware of both the s-line and UM2+c (just a .ZIP with the .xml files)
> * Default to removable drive if it is available
> * Only visible on UM-printers

The name of the button is taken from the GUI design by @lorenzoromagnoli . The rest of that design is explicitly considered out of scope.

Implements CURA-8055.


lorenzoromagnoli:
Since the rest of the UX guided  flow is not implemented yet, and this will probably go in the next release without the complete flow. I'd suggest to change the label of the button to "Export all" or "Export for printers" so that the button becomes more self explanatory even without the rest of the flow.

cura/Machines/Models/MaterialManagementModel.py
```diff
@@ -264,3 +270,52 @@ def removeFavorite(self, material_base_file: str) -> None:
...
+            if device.__class__.__name__ == "RemovableDriveOutputDevice":
+                self._preferred_export_all_path = QUrl.fromLocalFile(device.getId())
+                break
+        else:  # No removable drives? Use local path.
+            self._preferred_export_all_path = cura_application.getDefaultPath("dialog_material_path")
+        self.outputDevicesChanged.emit()
+
+    outputDevicesChanged = pyqtSignal()  # Triggered when adding or removing removable drives.
+    @pyqtProperty(QUrl, notify = outputDevicesChanged)
+    def preferredExportAllPath(self) -> QUrl:
```
nallath, line 292: Isn't it neater to just turn this into a `getPreferredExportAllPath() ` (so a slot) and only calculate it before the exportAllMaterialsDialog dialog is opened? 

It feels a bit strange that we re-calculate the path every time that a device is added (even if the user might not ever use this). 

If that is changed, you can also remove the extra code that you added in the init.

cura/Machines/Models/MaterialManagementModel.py
```diff
@@ -264,3 +270,52 @@ def removeFavorite(self, material_base_file: str) -> None:
...
+    def _onOutputDevicesChanged(self) -> None:
+        """
+        When the list of output devices changes, we may want to update the
+        preferred export path.
+        """
+        cura_application = cura.CuraApplication.CuraApplication.getInstance()
+        device_manager = cura_application.getOutputDeviceManager()
+        devices = device_manager.getOutputDevices()
+        for device in devices:
+            if device.__class__.__name__ == "RemovableDriveOutputDevice":
```
nallath, line 283: Oof. I'm really not a fan of this :( Unfortunately all the other solutions that I can come up with are worse.

nallath (dismissed):

cura/Machines/Models/MaterialManagementModel.py
Ghostkeeper: An alternative would've been to check the device's short description (`Save to removable drive`, but translated) or its icon (`save_sd`). We can't use the device name because it's arbitrary, named after the actual drive name. The stream is just a file stream. The rest of the real properties are not unique.

The short description is a bit finicky because it's translated. It would work but not be nice code either. The icon is easy but not very descriptive. I opted for this code because it's clear what it does.

We can't use `isinstance` or something like that without importing classes from a plug-in.

One alternative would be to define a new field, like `OutputDeviceTargetType` and have that be a string by which we can identify it. But yeah, the class name is kind of already that. It's not a big improvement.

cura/Machines/Models/MaterialManagementModel.py
Ghostkeeper: Yeah, agreed. Basically the advantages are:

* No performance impact for people not using the sync button (=99.9% of users).
* Simplification of code, not requiring that extra signal to update listeners of the property.

Disadvantages are:

* Every time the dialogue is opened, we need to update the folder.

Since the dialogue is really only opened from this one button, and in the future from a pop-up instead but still only in one spot, the disadvantage is very minor. The advantages are better. I'll change it!

cura/Machines/Models/MaterialManagementModel.py
Ghostkeeper: As discussed online, we're leaving it like this.

rburema (approved):

PR #10031, opened 2021-06-24
CURA-8313: Ignore files-in-use while restoring a backup

konskarm:
Sometimes, while a backup is being restored, one of the files in the current config folder may still be in use. This means that once the backup tries to replace this file with the one in the backup, the entire backup restoration fails (on Windows, due to a permission error) and half of the configuration folder may end up being missing.

To prevent that, we decided to ignore files that are currently in use while restoring a backup (i.e. keep the current version of those files and not replace them with the ones in the backup). This _may_ lead to a slightly wrong configuration (e.g. a plugin may not be restored properly), but it is an acceptable result, as the rest of the configuration folder is restored properly.

CURA-8313

### DIFF
--- a/cura/Backups/Backup.py
+++ b/cura/Backups/Backup.py
@@ -193,14 +193,13 @@ class Backup:
         Logger.log("d", "Removing current data in location: %s", target_path)
         Resources.factoryReset()
         Logger.log("d", "Extracting backup to location: %s", target_path)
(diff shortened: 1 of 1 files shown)


jellespijker (approved):
Both @Vandrasc and @evtrados are okay with the last commit so I will merge it

PR #10038, opened 2021-06-26
Update FilamentChange.py

Poporacer:
The L parameter is used for the filament load length the additional unload parameter is U

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/FilamentChange.py
+++ b/plugins/PostProcessingPlugin/scripts/FilamentChange.py
@@ -106,7 +106,7 @@ class FilamentChange(Script):
                 color_change = color_change + (" E%.2f" % initial_retract)
 
             if later_retract is not None and later_retract > 0.:
-                color_change = color_change + (" L%.2f" % later_retract)
+                color_change = color_change + (" U%.2f" % later_retract)
 
             if x_pos is not None:
                 color_change = color_change + (" X%.2f" % x_pos)


rburema:
Thanks for the PR!

It seems that `U` is only required for a [Marlin-flavoured](https://marlinfw.org/docs/gcode/M600.html) printer though, [RepRap](https://reprap.org/wiki/G-code#M600:_Filament_change_pause) makes no mention of the `U` in their `M600`, only `L`.

Poporacer:
That could be a problem since Marlin uses `L` for Load length and RepRap uses `L` for additional **unload** length. It looks like RepRap does not have a load option? The problem with the Additional Unload length is that if you are using Marlin based firmware, once the filament is loaded, it will then load the additional` L` value and usually will strip the filament. A couple solutions:
1. Label the current "Later Retraction Distance" as "RepRap Later Retraction Distance" and add another option labeled "Marlin Later Retraction Distance" and in the code utilize the appropriate flag values. Possibly add an additional option of "Marlin Load Length"
2. On the tool tip for the "Later Retraction Distance" include the information that this is only for RepRap based systems?

rburema:
Hi @Poporacer 

When looking through the code to see how other scripts handle these things: `PauseAtHeight.py` and `DisplayProgressOnLCD.py` simply let the user choose a method (gcode flavour) from a dropdown (via 'method' and, one 'time remaining' is enabled 'time remaining method' respectively -- the second one has the simpler menu).

nallath:
I think we can close this in favor of  #10403

PR #10044, opened 2021-06-28
CURA-8013: Adjust sizes of header and toolbar

konskarm:
### Changes in the toolbar

* The toolbar button's background is changed to be a circle when hovered
* The spacing between the toolbar buttons is removed

image

### Changes in the Configuration selector:

* The size of the Machine selector gets adjusted when the Cura window gets smaller/bigger
* The text about the material in the configuration selector changes as follows:
    * If it fits, display "Brand, Color, and Type" of material (e.g. `Ultimaker Black PLA`)
    * If "Brand, Color, and Type" doesn't fit, change it to "Color and Type" of material (e.g. `Black PLA`)
    * If "Color Type" doesn't fit either, display only the type (e.g. `PLA`)
    * If "Type" doesn't fit, elide it

**Before:**
image

**After:**
image

image

image

### Finally, the shadow of the PrepareMenu has been replaced with a border.

**After:**
image

CURA-8013


plugins/PrepareStage/PrepareMenu.qml
```diff
@@ -120,24 +112,12 @@ Item
...
                 height: UM.Theme.getSize("stage_menu").height
                 width: UM.Theme.getSize("stage_menu").height
+                border.color: UM.Theme.getColor("lining")
+                border.width: UM.Theme.getSize("default_lining").width
 
                 radius: UM.Theme.getSize("default_radius").width
                 color: openFileButton.hovered ? UM.Theme.getColor("action_button_hovered") : UM.Theme.getColor("action_button")
             }
-
-            DropShadow
```
fieldOfView, line 128: Removing the DropShadow means that the QtGraphicalEffects import can be removed too.

nallath: Which is a good thing, since QtGraphical effects isn't supported by Qt 6.x (and might never be?)

konskarm: Good catch!

nallath (approved):

rburema:
I did notice an issue with the visibility of a disabled extruder in the side-bar, but that's not part of this PR.

PR #10048, opened 2021-06-28
Handle Critical Variant and Material Validation Error

TheFanatr:
#### Don't Validate Materials for Extruder With Unknown Preferred Variant
`variant_node` will be null in this case, so `variant_node.materials.get` query will cause critical failure in the software and prevent launch from continuing.

#### Use F-String Indicator to Translate `{variant_name}` Into Actual Value in Log String
It would seem that the f-string `f` indicator was missing from the string in the log statement for the case when there is no variant with the given name in `setVariantByName`.


TheFanatr:
This was closed when I renamed the branch with GitHub. The intention was to have the new/renamed branch be linked here; instead, the old branch was detected as "deleted" and this PR was closed without consent.

Ghostkeeper:
This man sees a problem in software and makes a fix. You have my respect.
And the side-fix for the string formatting is definitely fine.

However catching that case here is controversial. This hides a problem that we'd rather know about.

We've discussed this and think a better solution would be to validate that the preferred variant exists upon loading the machine definition.

TheFanatr:
Hey @Ghostkeeper, thanks for the reply.

Here are some of my thoughts on what your brought up:

> This man sees a problem in software and makes a fix. You have my respect.

Thanks! Ego boosts always appreciated 😃.

> However catching that case here is controversial. This hides a problem that we'd rather know about.
> 
> We've discussed this and think a better solution would be to validate that the preferred variant exists upon loading the machine definition.

I did add an error log to try to address the hiding, because I felt that silently ignoring this would be incorrect, but yes, I agree that rectifying/handling this in a core location before other validation steps which may rely on machine definition assumptions (such as the presence of the preferred variant) would be better; however, I just want to be clear that I would greatly prefer if this validation step also could try to rectify the issue (such as just setting the preferred variant to the first one found if it's not set), not just decide the definition is inadequate and continue on to the next one, or stop startup, or something like that. Might even be nice to notify the user and let them confirm a corrective action to the relevant configuration. Just being explicit because I'm not sure as to which sort of "validate" you were referring.

Would you like me to attempt to modify my PR to include any of what I just mentioned, or is it being handled already? Also, if you would like me to, where do you think it would be reasonable to include the "assumption validation" or just specifically, the "preferred variant validation" step? I'll do some digging at some point as well if needed to find somewhere, but just asking in case you had a specific location in mind.

Anyways, enjoy the rest of your day.

Ghostkeeper:
> however, I just want to be clear that I would greatly prefer if this validation step also could try to rectify the issue (such as just setting the preferred variant to the first one found if it's not set), not just decide the definition is inadequate and continue on to the next one, or stop startup, or something like that.

Yes, that should be possible and indeed better. The ContainerTree class creates the list of available variants for the printer (ContainerTree -> MachineNode -> `self.variants`). All of these variants are safe to activate (it is what fills the menu of available variants). It will contain just the "empty" container if there are no variants, which is also safe to activate. It should never be empty.

I think letting the user confirm a corrective action is overkill here.

> Just being explicit because I'm not sure as to which sort of "validate" you were referring.

What I meant is that the root cause of this crash is, as you say, that the printer definition file lists a "preferred_variant", yet that variant apparently doesn't exist, or doesn't match the current printer. Before trying to set the `variant_node` to None it should first try to see if the `variant_node` exists within the `machine_node.variants`, and if not explicitly change the `variant_node` to something which does exist.
In short, I think this check should be moved a bit earlier in the same function, to before where the `setVariantByName` call already sets the variant to `None` (which would cause the issue down the line). And it should make sure that it sets the variant to a fallback if the preferred one doesn't exist. It should also log a warning then (which you already implemented more or less).

> Would you like me to attempt to modify my PR to include any of what I just mentioned, or is it being handled already, or should it be done in another possibly more comprehensive PR?

It's fine to modify the existing PR. We are not complete purists of the Git commit history. And if we were we could always squash and merge. I kind of hope that you are able to reproduce the issue and then fix it. We could also try to fix it (based on what I just said above) but it would help a lot if you can provide reproduction steps of the issue you're seeing (also to test your PR)! Are you using a custom definition file?

casperlamboo:
Hi @TheFanatr,

Thank you once again for opening the PR and bringing this to our attention. As @Ghostkeeper pointed out we'd rather not "silently" (from the users perspective) ignore the issue. A "corrective validation" step would be a better solution our opinion indeed. The `f"string"` part of the PR is clearly a mistake which should be fixed.

If you're still up to it you are free to update the PR and put through the discussed changes. Alternatively we could just merge the `f"string"` part and open a bug on the back log to correctly handle "missing variants".

Thanks,
@casperlamboo

GregValiant:
It's been three years and a few new Cura releases.  Is this still an issue(?) or can this be closed?

GregValiant:
There was no response to my push so I'll go ahead and close this.  It can always be reopened.

PR #10051, opened 2021-06-29
Replace extruder icons

nallath:
As per the new design!

CURA-8011


resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
```diff
@@ -70,7 +70,7 @@ Cura.ExpandablePopup
                         font: UM.Theme.getFont("default")
                         color: UM.Theme.getColor("text")
                         renderType: Text.NativeRendering
-
+                        opacity: model.enabled ? 1 : 0.4
```
Ghostkeeper, line 73: Maybe this would be better done through a theme file, so that it can be specialised for e.g. the high contrast themes?

I consider this to be subjective though, so I won't let it block the issue and continue testing.

resources/qml/ExtruderIcon.qml
```diff
@@ -17,29 +17,26 @@ Item
...
-    {
-        id: extruderNumberCircle
-
-        width: height
-        height: Math.round(parent.height / 2)
-        radius: Math.round(width)
-        color: UM.Theme.getColor("toolbar_background")
+        UM.RecolorImage
+        {
+            id: blorp
```
Ghostkeeper, line 27: This is not a very... descriptive ID. It's not used anywhere either. Better remove it.

resources/qml/ExtruderIcon.qml
```diff
@@ -17,29 +17,26 @@ Item
...
+        UM.RecolorImage
         {
-            horizontalCenter: parent.horizontalCenter
-            top: parent.top
-            topMargin: (parent.height - height) / 2
+            id: mainIcon
+            anchors.fill: parent
+
+            source: UM.Theme.getIcon("Extruder", "medium")
+            color: UM.Theme.getColor("text")
```
Ghostkeeper, line 39: My first thought was that this is not text and should not get the text colour assigned. But I understand that you'll want this to have the same colour as the text in the centre of the icon. Maybe this should just be linked to `extruderNumberText.color` then?

Ghostkeeper (changes requested):

resources/qml/ExtruderIcon.qml
nallath: WHoops. My bad. I copied it and it was complaining about a duplicate.

konskarm: blorp, forever in our hearts

PR #10071, opened 2021-07-02
[CURA-8012] restyle buttons

rburema:
Light and dark theme.


resources/qml/TertiaryButton.qml
```diff
@@ -9,13 +9,11 @@ import Cura 1.1 as Cura
...
-    shadowColor: enabled ? UM.Theme.getColor("secondary_button_shadow"): UM.Theme.getColor("action_button_disabled_shadow")
     color: "transparent"
     textColor: UM.Theme.getColor("text_link")
     outlineColor: "transparent"
-    disabledColor: UM.Theme.getColor("action_button_disabled")
+    outlineDisabledColor: "transparent"
+    disabledColor: "transparent"
     textDisabledColor: UM.Theme.getColor("action_button_disabled_text")
-    hoverColor: "transparent"
-    underlineTextOnHover: true
```
Ghostkeeper, line 20: These links are underlined in the design. Should they now not always get underlined (regardless of hover)?

resources/qml/TertiaryButton.qml
```diff
@@ -9,13 +9,11 @@ import Cura 1.1 as Cura
...
-}
+    hoverColor: UM.Theme.getColor("secondary_button_hover")
```
Ghostkeeper, line 18: Tertiary buttons should not use secondary button theme elements. It would be better to duplicate the theme entry if they happen to be the same in the current design.

resources/themes/cura-dark/theme.json
```diff
@@ -87,8 +94,8 @@
         "action_button_active": [39, 44, 48, 30],
         "action_button_active_text": [255, 255, 255, 255],
         "action_button_active_border": [255, 255, 255, 100],
-        "action_button_disabled": [19, 24, 28, 255],
-        "action_button_disabled_text": [200, 200, 200, 80],
+        "action_button_disabled": [85, 85, 87, 255],
+        "action_button_disabled_text": [103, 103, 104, 255],
```
Ghostkeeper, line 98: Are you sure the intent wasn't to just reduce the opacity for this one?

Ghostkeeper (changes requested):
This has an unintended change throughout our interface. The designs have a Primary, Secondary, Tertiary and Quaternary button design, with Quaternary looking like a link. However Cura has always had just a Primary, Secondary and Tertiary in its designs, with Tertiary looking like a link.

This change turns all previously Tertiary buttons into the new Tertiary buttons. However:
* The links should still stay links in the previous designs, so they should become "Quaternary" buttons as shown in the new button designs.
* The new Tertiary button is out of scope for this ticket.
* Changing all Tertiary buttons to Quaternary buttons would arguably be an API break as well.

So I suggest that you keep Tertiary buttons as they are, as links, but change their designs to what the new Quaternary button designs show.

resources/themes/cura-dark/theme.json
rburema: It was not that way in the light theme. To have the light theme have an opaque disabled text and then have the dark theme have it be semi-translucent is just asking for trouble.

resources/qml/PrimaryButton.qml
```diff
@@ -9,8 +9,6 @@ import Cura 1.1 as Cura
 
 Cura.ActionButton
 {
-    shadowEnabled: true
-    shadowColor: enabled ? UM.Theme.getColor("primary_button_shadow"): UM.Theme.getColor("action_button_disabled_shadow")
```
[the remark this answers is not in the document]
rburema: How does it break? Those are present (albeit in a different, deprecated form) in ActionButton?

You can still do anything with those properties, they just don't _do_ anything besides print a deprecation warning.

resources/qml/TertiaryButton.qml
rburema: From the ticket:
> we do not have tertiary buttons in cura (yet).

This is a bit of a lie. We _do_ have (some) tertiary buttons, and they _are_ used differently than links/quaternary. For example, the 'Skip' button in the what's new pages is tertiary, but should not be a link:

> These links ...

... and they aren't really links either. (As evidenced by the fact that the 'old' tertiary button _also_ didn't have underlining in any 'mode' anyway.)

In any case, everything that was a link will stay a link, because _those_ where never using (this) tertiary in the first place. 

So: I just took the ticket and comments to mean that I shouldn't yet add Tertiary buttons in places where we now use links that we maybe _should_ use them.

resources/themes/cura-dark/theme.json
```diff
@@ -5,7 +5,7 @@
     },
 
     "colors": {
-        "main_background": [39, 44, 48, 255],
+        "main_background": [31, 31, 32, 255],
```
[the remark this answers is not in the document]
rburema: I did this to see the buttons in the same way as the background of the design was.

However, since the buttons are never directly on the scene, I'll put this back (since it wasn't really doing what I want either).

resources/qml/ActionButton.qml
```diff
@@ -127,29 +167,14 @@ Button
         }
     }
 
-    background: Cura.RoundedRectangle
+    background: Rectangle
```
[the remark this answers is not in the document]
rburema: Agreed, resolved in: https://github.com/Ultimaker/Cura/pull/10071/commits/e4710dd921245d031074a53e10f11c98d7844e02

resources/themes/cura-dark/theme.json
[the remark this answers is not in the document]
rburema: Resolved in: https://github.com/Ultimaker/Cura/pull/10071/commits/a42fee48340f72842e0a340fc090321eff71ed23

resources/qml/TertiaryButton.qml
rburema: Agreed, resolved in: https://github.com/Ultimaker/Cura/pull/10071/commits/acf3f2d02a1a0cf393cf6e10d381a4a28dd1376e

resources/qml/TertiaryButton.qml
rburema: Sorry, that part about the old button was wrong. I think the rest still stands though?

rburema: Sorry, that part about the old button was wrong. I think the rest still stands though?

rburema: Sorry, that part about the old button was wrong. I think the rest still stands though?

konskarm: I do agree with Ruben Dulek here that the tertiary button should remain the "link" button, as it was traditionally used, and we should introduce the quatertiary for the more subtle action button. If we change the Tertiary buttons to not be links, the existing tertiary buttons no longer look like they link to an external source:

image

This is probably not desirable by plugins that use Tertiary buttons.

resources/qml/PrimaryButton.qml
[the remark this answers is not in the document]
konskarm: I don't think it breaks either. The properties `shadowEnabled` and `shadowColor` of the `ActionButton` still exist, they are just not changed at all in the `PrimaryButton`.

nallath (approved):

resources/qml/TertiaryButton.qml
rburema: Oof, sorry about all of that. I was completely in the wrong here :-(

PR #10078, opened 2021-07-03
Adds FLSun Super Racer Profile

thushan:
Adds the [FLSun Super Racer](https://www.flsun3d.com/product/3.html) profile to Cura.

Verified with a bunch of prints, slim profile.


resources/definitions/flsun_sr.def.json
```diff
@@ -0,0 +1,76 @@
...
+    "name": "FlSun SuperRacer",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Thushan Fernando",
+        "manufacturer": "Flsun",
+        "platform": "flsun_sr.3mf",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
+		"has_machine_quality": true,
```
konskarm, line 12:
```suggestion
		"has_machine_quality": false,
```
The machine states that it has machine qualities but there are no quality profiles provided. If you don't keep it as false, you will end up with unsupported profiles. 
image

If you set it to false, then the global qualities will be able to be used instead.
image

konskarm (changes requested):

resources/definitions/flsun_sr.def.json
thushan: Good catch, resolved as requested.

Will add quality profiles after consulting with FLSun. Good catch :)

resources/definitions/flsun_sr.def.json
```diff
@@ -0,0 +1,76 @@
...
+        "machine_head_with_fans_polygon": {
+            "default_value": [
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0]
+            ]
+        },
+        "z_seam_type": {
+            "default_value": "back"
```
konskarm, line 49:
```suggestion
            "default_value": "'back'"
```
I missed that in the previous review. The enumaration values should be inside single quotes (in addition to the double outside quotes) to be recognized as enums.

konskarm (changes requested):

resources/definitions/flsun_sr.def.json
thushan: Updated with thanks.

konskarm:
Thanks for the quick responses @thushan. Everything works as expected as far as I can see. The models are sliced without issues, so I am merging this. The printer will be available starting from our next release (4.11).

Guilouz:
@thushan @konskarm 3mf file is not Super Racer, here is a more adequate one and use : 

"platform_offset": [0,-35,-40] in definition file

[flsun_sr.3mf.zip](https://github.com/Ultimaker/Cura/files/6874798/flsun_sr.3mf.zip)

Capture d’écran 2021-07-26 à 02 29 45

thushan:
> @thushan @konskarm 3mf file is not Super Racer, here is a more adequate one and use :
> 
> "platform_offset": [0,-35,-40] in definition file
> 
> [flsun_sr.3mf.zip](https://github.com/Ultimaker/Cura/files/6874798/flsun_sr.3mf.zip)
> 
> Capture d’écran 2021-07-26 à 02 29 45

Thanks mate, looks sweet! This is what FLSun requested I use. Do you want to open a new PR with the above work of art as this is closed and it's merged? Happy to do it otherwise. Thanks again!

Ghostkeeper:
See also https://github.com/Ultimaker/Cura/pull/10192 where that model is integrated.

mayassa58:
Hi I'm new to the FLSUN SR just came in yesterday set it up and pulled the Cura 4.10 files from https://drive.google.com/drive/folders/1GPXxRs7cEx7Z3UORi-EPmMg6j4ZM2zd5.  I can print the test nut and bolt but anything else I print that I slice with Cura puts a 1mm flat border around my print.  Not sure what I'm doing wrong and need some help.

I also reformatted my memory to 32 bit like the instructions showed and also 4096 then sliced and saved again with no luck

My Cura works fine with my Aquila, I just must be missing something.

Thanks in advance

mayassa58:
Maybe I found it under build plate adhesion

Guilouz:
This is not the place for this kind of problem, your problem is not related to this PR…

Just change build plate adhesion type to skirt or another one.

mayassa58:
Thanks

PR #10086, opened 2021-07-04
Biqu bx definition

looxonline:
Added the BIQU BX printer definition


konskarm:
Everything looks to be in order. The printer slices without issues, so I am merging this. Kudos for delivering a definition that didn't need any changes!
The Biqu BX printer will be available in Cura starting from 4.11.

PR #10093, opened 2021-07-06
Changed "Deviation" for Anycubic Mega S and X to better support hardware 

NilsRo:
As the printer has a ATMega2560 it is better to have a lower resolution to avoid blobs


resources/definitions/anycubic_i3_mega_s.def.json
```diff
@@ -115,6 +115,8 @@
         "skirt_gap": { "value": 5.0 },
         "skirt_line_count": { "value": 4 },
 
+		"meshfix_maximum_deviation": { "value": 0.05 },
+		"meshfix_maximum_resolution": { "value": 0.5 },
```
konskarm, line 119: A couple of changes:

- Spaces instead of tabs
- The `meshfix_maximum_resolution` is already inherited as 0.5 from the `fdmprinter.def.json`, so no need to override it here with the same value

```suggestion
        "meshfix_maximum_deviation": { "value": 0.05 },
```

PR #10110, opened 2021-07-09
Increase coloured ring thickness in for the medium sized icon.

lorenzoromagnoli:
As reported by some users, the color is quite hard to read for extremely light and dark color. To improve the visibility I made the ring a little thicker. 
It still blends in nicely with the rest of the interface.


Ghostkeeper:
New renders:
image
image

nallath (approved):

PR #10121, opened 2021-07-12
Add AtomStack printer definitions

zhpt:
Hello, we want to add our machine AtomStack to Cura. I have added files in definitions and extruders.
Thank you!


konskarm (changes requested):
Thanks for the pull request @zhpt. I notice the following issues:

1. The tests are failing because the `'category'` should not exist in the definition metadata. 
2. Definitions and extruders should be named as `<manufacturer>_<printer_name>.def.json` and `<manufacturer>_<printer_name>_<extruder>.def.json` correspondingly. So please prepend `atomstack` in front of the definition and extruder files. This will make it easier to maintain in the future.
3. Sometimes there is an underscore `_` between "cambrian" and the printer model while other times there is not (e.g. `cambrianmaxe285_extruder_1.def.json` and `cambiran_maxe285.def.json`. Please keep it consistent.
4. There is a misspelling in the "cam**bri**an" as sometimes it is written as "cam**bir**an".

resources/definitions/cambiran_maxe175.def.json
```diff
@@ -0,0 +1,48 @@
+{
+  "version": 2,
+  "name": "Cambrian MAX E175",
+  "inherits": "fdmprinter",
+  "metadata": {
+    "visible": true,
+    "author": "AtomStack",
+    "manufacturer": "AtomStack",
+    "category": "AtomStack",
```
konskarm, line 9:
```suggestion
```

resources/definitions/cambiran_maxe285.def.json
```diff
@@ -0,0 +1,48 @@
+{
+  "version": 2,
+  "name": "Cambrian MAX E285",
+  "inherits": "fdmprinter",
+  "metadata": {
+    "visible": true,
+    "author": "AtomStack",
+    "manufacturer": "AtomStack",
+    "category": "AtomStack",
```
konskarm, line 9:
```suggestion
```

resources/definitions/cambiran_proe175.def.json
```diff
@@ -0,0 +1,48 @@
+{
+  "version": 2,
+  "name": "Cambrian Pro E175",
+  "inherits": "fdmprinter",
+  "metadata": {
+    "visible": true,
+    "author": "AtomStack",
+    "manufacturer": "AtomStack",
+    "category": "AtomStack",
```
konskarm, line 9:
```suggestion
```

resources/definitions/cambiran_proe285.def.json
```diff
@@ -0,0 +1,48 @@
+{
+  "version": 2,
+  "name": "Cambrian Pro E285",
+  "inherits": "fdmprinter",
+  "metadata": {
+    "visible": true,
+    "author": "AtomStack",
+    "manufacturer": "AtomStack",
+    "category": "AtomStack",
```
konskarm, line 9:
```suggestion
```

konskarm (changes requested):

zhpt:
@konskarm Hello, I have changed those files follow your advice. Please review again. Thank you!

nallath:
Our CI is still reporting issues. Please check the errors that it generated.

Ghostkeeper:
You can find the issues if you run `cmake . && make test` in the source repository (or better yet, from a build directory).

The current issues are:
* `atomstack_cambrianmaxe175_extruder_0.def.json` refers to a printer called `cambrian_maxe175.def.json` which doesn't exist. The reference to that printer needs to be updated (`atomstack_...`).
* `atomstack_cambrianmaxe285_extruder_0.def.json` refers to `cambrian_maxe285.def.json` which doesn't exist.
* `atomstack_cambrianproe175_extruder_0.def.json` refers to `cambrian_proe175.def.json` which doesn't exist.
* `atomstack_cambrianproe285_extruder_0.def.json` refers to `cambrian_proe285.def.json` which doesn't exist.

zhpt:
@Ghostkeeper  Hello, I'm so sorry to take these issues. I fixed these issues just now. Please review again. Thanks a lot!

zhpt:
@Ghostkeeper Hello, I change some parameters on our machine definitions. Please check it. Thank you.

konskarm (changes requested):
The issues have indeed been fixed. I see lots of code duplication now though. It seems that all the profiles are exactly the same apart from the machine sizes, the machine names, and the extruders. 
I would suggest you to make a base profile that is not visible and contains all the information that stays the same among those printers, such as an `atomstack_cambrian_base.def.json` with:

- basic `metadata` that stays the same (`author`, `manufacturer`, `file_formats`, `has_materials`)
- `overrides` that stay the same (`machine_heated_bed`, `machine_extruder_count`, `machine_start_gcode`, `machine_end_gcode`)

Then, your printer profiles can inherit from that and change only the things they need to change (`name`, `visible`, `machine_extruder_trains`, `machine_name`, `machine_width`, `machine_height`, `machine_depth`). This will help in the future maintenance of those profiles because, if you want to e.g. change the `machine_end_gcode`, it will need to be changed only in the base profile instead of in all profiles.

Please have a look at how other printer families follow that paradigm (e.g. the creality profiles inheritting from `creality_base.def.json` or the artillery profiles inheritting from `artillery_base.def.json`).

Also, please remove the `platform_offset` from the profiles. This is only necessary if you have a platform mesh file and you want to place it properly relatively to your printer. Since you do not provide any platform meshes, changing this has no effect.

zhpt:
@konskarm Hello, thanks your check and advices. I have fixed it following your advice. Please check again. Thank you!

Ghostkeeper (approved):
Looks good to me now. I'll give this a spin to see if it works, and if so merge it for the next release.

Thanks for your patience with our comments!

PR #10122, opened 2021-07-12
Add a drop-down to the open-file button if multiple options are available

Ghostkeeper:
If there are multiple sources from which you can open files, the open file button turns into a drop-down menu.

image

The drop-down menu then allows you to select the source from which to open files.

If there is only one option, it shows the old button.

Implements CURA-8008.

### DIFF
--- a/plugins/PrepareStage/PrepareMenu.qml
+++ b/plugins/PrepareStage/PrepareMenu.qml
@@ -78,35 +78,28 @@ Item
         {
             id: openFileButton
 
-            //Make the button square if the contents are.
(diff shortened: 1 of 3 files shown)


nallath (approved):

fieldOfView:
The inconsistent capitalization irks me.

I think that instead of adding a non-functional menu-item as a language construct, it is better to just spell this out: "Open from Disk" and "Open from Digital Library". This will also give translators fewer issues.

lorenzoromagnoli:
Super valid points Aldo! I guess I overdesigned this a little bit. I'll talk with the team and see if we can change it before the release.

PR #10123, opened 2021-07-12
corrected "i18n/de_DE/fdmprinter.def.json.po" -> z_seam-"Back" from "Zurück" to "Hinten"

tispokes:


Ghostkeeper (approved):
Yup, that seems to have been a mistake from the translators.

It translated `Back` in the sense of `Return`, but it should rather have been translated in the sense of `Behind` or `Rear`.

Thanks for the correction!

PR #10130, opened 2021-07-13
Search through Digital Library projects

Ghostkeeper:
This PR features the following changes:

* The `TextField` re-usable widget has a new feature to allow it to display an icon before the typed text. The icon is not part of the placeholder text but part of the background, so it remains visible even during typing.
* There is a search field at the top of the project list in the Digital Library, both when loading a file and when saving a file.
* The search is still paginated just like the original results.
* If there are no search results, show a different placeholder image/text than the standard "you don't have projects yet" message.

It looks like this:
image

And dark mode:
image

Implements CURA-8009.


plugins/DigitalLibrary/src/DigitalFactoryController.py
```diff
@@ -302,6 +310,37 @@ def setSelectedFileIndices(self, file_indices: List[int]) -> None:
...
+
+    @pyqtProperty(str, notify = projectFilterChanged, fset = setProjectFilter)
+    def projectFilter(self) -> str:
+        """
+        The current search filter being applied to the project list.
+        :return: The current search filter being applied to the project list.
+        """
+        return self._project_filter
+
+    def _applyProjectFilter(self):
```
nallath, line 336:
```suggestion
    def _applyProjectFilter(self) -> None:
```

nallath (changes requested):
Some minor changes, it looks good otherwise.

Ghostkeeper:
This failed the component testing stage because the pagination feature seems to have broken. Needs a bit more debugging.

PR #10131, opened 2021-07-13
CURA-8112: Inform the user about their subscription limits

konskarm:
Instead of letting users go through the project creation process only to get rejected with a "subscription limits reached" message, now the "New Library project" button is being replaced with an "Upgrade plan" button when the maximum allowed projects have been reached for the specific account. The button is accompanied by a tooltip that explains the situation to the user. Once clicked, the user is redirected to the subscriptions page.

image

CURA-8112


plugins/DigitalLibrary/src/DigitalFactoryController.py
```diff
@@ -166,6 +170,11 @@ def userAccountHasLibraryAccess(self) -> bool:
             subscriptions = self._account.userProfile.get("subscriptions", [])
             if len(subscriptions) > 0:
                 return True
+        if self._user_has_access:
+            # The user has access even though they have no subscriptions. This means they are an Essential user and they
+            # have limited personal private projects available. In this case, we need to check whether they have already
+            # reached their limit.
+            self._api.checkUserCanCreateNewLibraryProject(callback = self.setCanCreateNewLibraryProject)
```
nallath, line 177: Shouldn't we just always check if a user has a limit? That way if it ever changes, Cura is ready for that.

resources/qml/Cura.qml
```diff
@@ -434,6 +436,11 @@ UM.MainWindow
                     Cura.TertiaryButton
                     {
                         text: model.name
+                        iconSource: {
+                            if (model.icon == null || model.icon == "")
```
nallath, line 440: Nitpicking: add { }.

plugins/DigitalLibrary/src/DigitalFactoryController.py
konskarm: Indeed, there was no need to check for the total number of subscriptions any more, since the variable `_user_has_access` already has covered the case the user has a subscription. I removed the number of subscriptions check in https://github.com/Ultimaker/Cura/pull/10131/commits/3dbbe7276baa3b201698e19bfbacbef45a3d8464.

resources/qml/Cura.qml
```diff
@@ -434,6 +436,13 @@ UM.MainWindow
                     Cura.TertiaryButton
                     {
                         text: model.name
+                        iconSource: {
```
Ghostkeeper, line 439: Code style: Brackets on new line.
```suggestion
                        iconSource:
                        {
```

plugins/DigitalLibrary/src/DigitalFactoryApiClient.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, application: CuraApplication, on_error: Callable[[List[CloudE
         self._http = HttpRequestManager.getInstance()
         self._on_error = on_error
         self._file_uploader = None  # type: Optional[DFFileUploader]
+        self._library_max_private_projects: Optional[int] = None
```
Ghostkeeper, line 58: Aren't all of these projects private?
```suggestion
        self._library_max_projects: Optional[int] = None
```

plugins/DigitalLibrary/resources/qml/SelectProjectPage.qml
```diff
@@ -61,13 +61,28 @@ Item
...
         }
+
+
+        Cura.SecondaryButton
+        {
+            id: upgradePlanButton
+
+            text: "Upgrade plan"
+            iconSource: UM.Theme.getIcon("LinkExternal")
+            visible: createNewProjectButtonVisible && !manager.userAccountCanCreateNewLibraryProject && (manager.retrievingProjectsStatus == DF.RetrievalStatus.Success || manager.retrievingProjectsStatus == DF.RetrievalStatus.Failed)
```
Ghostkeeper, line 80: Why does this need to depend on `DF.RetrievalStatus` and `manager.retrievingProjectsStatus`? Shouldn't it just depend on `manager.userAccountCanCreateNewLibraryProject`? If it fails to retrieve the list of projects, the last known information should count, right?

plugins/DigitalLibrary/src/DigitalFactoryApiClient.py
konskarm: Nope. The ones retrieved by calling the API without filters contains both private and shared projects. We only check against the private (with query parameter `shared=false`).

plugins/DigitalLibrary/resources/qml/SelectProjectPage.qml
konskarm: To check for private projects we need to query the `/projects` API endpoint. So the button should be shown after the GET request for retrieving the private projects has finished.

nallath (approved):

PR #10141, opened 2021-07-15
MP Mini Delta V2 Printer Definition

PurpleHullPeas:
Adding the MP Mini Delta V2. While this printer has the same form factor as the original MP Mini Delta, it is actually made by a different manufacturer (WEEBO instead of Malyan). The different mainboard, firmware, and other minor changes calls for different Start Gcode, as well as some other settings tweaks. 

I also made some minor changes to the original MP Mini Delta def. 

I am behind on the process for adding printers to Cura manually (via file copy), so I have not tested this def outside of making sure the Start/End gcode makes the old V1 def work for the V2. I can dig into this more if the build test cases report errors.

### DIFF
--- a/resources/definitions/mp_mini_delta.def.json
+++ b/resources/definitions/mp_mini_delta.def.json
@@ -3,7 +3,7 @@
     "name": "MP Mini Delta",
     "inherits": "fdmprinter",
     "metadata": {
-        "author": "MPMD Facebook Group",
+        "author": "MPMD V1 Facebook Group",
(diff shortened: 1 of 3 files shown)


Ghostkeeper (approved):
Looks good to me from a review perspective. I'll test if it all loads properly in Cura and if so merge it. Thanks for the contribution, Reddit!

Ghostkeeper:
Cute little printer. Thanks for the profile updates!

PR #10143, opened 2021-07-15
Update and rename resources/definitions/creality_ender3.def.json to r…

Nicoo-G:
…ecursos/definiciones/creality_ender3.def.json


Ghostkeeper:
"Definiciones"? No, that sabotages this definition file. Sorry!

PR #10150, opened 2021-07-18
Z-Bolt series profiles

kotovskii:
Hello. We're adding Cura profiles support for our Z-Bolt Limited and S series 3D printers.


resources/definitions/z-bolt_base.def.json
```diff
@@ -0,0 +1,279 @@
...
+        "file_formats": "text/x-gcode",
+        "first_start_actions": ["MachineSettingsAction"],
+        "has_machine_quality": true,
+        "has_materials": true,
+        "has_variants": true,
+        "variants_name": "Toolhead",
+        "preferred_variant_name": "Direct E3DV6 0.40mm",
+        "machine_extruder_trains": { "0": "z-bolt_extruder_0" },
+        "exclude_materials": [
+            "ultimaker_abs_black",
```
nallath, line 19: It's really not needed to exclude these material profiles! They are 2.85 mm, so they won't show up for this printer anyway.

resources/definitions/z-bolt_base.def.json
```diff
@@ -0,0 +1,279 @@
...
+        "machine_extruder_count":            			{ "value": 1 },
+        "machine_start_gcode":                          { "default_value": "G21 ;metric values\nG90 ;absolute positioning\nM82 ;set extruder to absolute mode\nM107 ;start with the fan off\nG28 X0 Y0 Z0 ;Home all\nM140 S{material_bed_temperature_layer_0}\nM104 S160\nM190 S{material_bed_temperature_layer_0}\nM109 S160\nG4 P15000; pause\nG29;auto bed level\nG1 Z3 F1200 ;move the platform up 3mm\nG1 X0 Y0 F12000\nG1 Z0 F1200 ;move the platform down\nM104 S{material_print_temperature_layer_0}\nM109 S{material_print_temperature_layer_0}\nG1 Z3 F6000 ;move the platform up 3mm\nG92 E0;zero the extruded length\n;intro line\nG1 X0 Y0 F5000\nG1 Z0.2 F2000\nG1 X50 E10 F1500\nG1 X100 E18 F1000\nG92 E0\n" },
+        "machine_end_gcode":                            { "default_value": "G91 ;relative positioning\nG1 E-6 F1200 ;start retract\nG1 Z+3 E-10 F5000 ;continue retract\nG90 ;absolute positioning\nG28 X0 Y0\nM104 S0\nM140 S0\nM82\n" },
+
+        "layer_height_0":								{ "value": "layer_height if layer_height >= 0.24 else 0.2"},
+		"line_width":                                   { "value": "machine_nozzle_size" },
```
nallath, line 198: Please fix the identation everywhere. It's a lot easier for the future if it's formatted the same e

resources/extruders/z-bolt_extruder_0.def.json
```diff
@@ -1,14 +1,16 @@
 {
     "version": 2,
-    "name": "Extruder 1",
+    "name": "Toolhead",
     "inherits": "fdmextruder",
-    "metadata": {
+    "metadata":
+    {
+        "machine": "z-bolt_base",
```
nallath, line 7: Did you check if older profiles still work due to this change?

resources/definitions/z-bolt_base.def.json
kotovskii: It was easier for me to develop and test profiles with this indentation, however it's not needed now since profiles are ready. Fixed it.

resources/definitions/z-bolt_base.def.json
kotovskii: Removed Ultimaker material profiles from blacklist.

resources/extruders/z-bolt_extruder_0.def.json
kotovskii: Probably no. We want to remove Original and Original Plus profiles, since these printers are no longer officially supported by Z-Bolt, and also printing profiles are outdated. 
New profiles for Limited and Limited Plus are mostly compatible with Original and Original Plus and have been successfully tested by our team.

nallath: If you want to remove them, you will also need to create a version upgrader script.

Ghostkeeper: A version upgrader would make sense if a new profile would be used for the same printer. But in this case it's just that there's a wholly new printer.

Since you say that the profiles for the new printer work on the older printer as well, I think the most logical solution would be to keep the old definition file, but make it inherit from the new printer and use the same quality profile set. That way a version upgrade is only necessary to change the quality profiles, but the definitions can just stay in place.

kotovskii: Old profiles are outdated, and adding (also possibly maintaining and updating) new quality profiles and variants will be very difficult. Limited series have same build volume as older Original ones, it's basically just a redesign in all-metal frame with small updates.

Ghostkeeper: Yes, so what I meant was to just have one set of profiles, your new profiles, but mark the Original and Original Plus printers to use that same set of profiles (by overriding the `quality_definition` metadata entry).
When those people transition to the new set of profiles, a transformation needs to be made to their configuration. Cura stores which profiles you had active in a "machine instance" resource. That entry would need to change in a version upgrade, to make them use the new set of profiles, rather than the old ones that you don't want to support any more.

resources/quality/z-bolt/bowden/ABS/v6 0.3/z-bolt_bowden_v6_0.30_ABS_extrafine.inst.cfg
```diff
@@ -0,0 +1,23 @@
+[general]
```
Ghostkeeper, line 1: A space in the folder name is not reliable. We do percent-encoding of these paths so that could lead to corruptions if using special characters like the space.

resources/quality/z-bolt/bowden/ABS/volcano 0.6/z-bolt_bowden_volcano_0.60_ABS_sprint1.inst.cfg
```diff
@@ -0,0 +1,23 @@
+[general]
+version = 4
+name = Sprint [1]
```
Ghostkeeper, line 3: You might want to give these profiles different names than "various numbers of sprint" or re-weight the user-visible names if the Draft profile isn't really intended for quick drafts.

resources/quality/z-bolt/direct/ABS/v6 0.3/z-bolt_direct_v6_0.30_ABS_extrafine.inst.cfg
```diff
@@ -0,0 +1,23 @@
...
+variant = Direct E3DV6 0.30mm
+
+[values]
+speed_print = 25
+retraction_amount = 1.1
+adhesion_type = brim
+
+cool_fan_speed_min = 35
+cool_fan_speed_max = 60
+cool_min_layer_time_fan_speed_max = 20
```
Ghostkeeper, line 20: It may be worth considering - for your own sanity - to override this setting in the ABS profile instead of in ALL of the quality profiles. For ABS you're using 20, and for all of the rest 15 so the printer definition could just set it to 15 and have ABS override it.

Ghostkeeper (changes requested):
Nice to see some progress with the indentation. But the main issue here is not yet addressed: The breaking of existing configurations.

The automated tests are currently also failing because `z-bolt_base` doesn't define `machine_extruder_trains`, a required metadata entry.

This here is a full review. If these issues are fixed, we'll still have to do a test, but you're probably in the clear by then.

resources/definitions/z-bolt_S600.def.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "Z-Bolt S600",
```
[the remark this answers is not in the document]
kotovskii: Those are manufactured by personal request only, so they should not be on site.

resources/definitions/z-bolt_base.def.json
```diff
@@ -0,0 +1,199 @@
...
+		"speed_infill":					{ "value": "speed_print" },
+		"speed_wall":                                   { "value": "math.ceil(speed_print * 0.6)" },
+		"speed_wall_0":                                 { "value": "math.ceil(speed_print * 0.6)" },
+		"speed_wall_x":                                 { "value": "math.ceil(speed_print * 0.8)" },
+		"speed_roofing":                                { "value": "math.ceil(speed_print * 0.4)" },
+		"speed_topbottom":                              { "value": "math.ceil(speed_print * 0.5)" },
+		"speed_prime_tower": 				{ "value": "speed_topbottom" },
+		"speed_support": 				{ "value": "speed_print" },
+		"speed_support_interface": 			{ "value": "speed_topbottom" },
+		"speed_travel":                                 { "value": 120, "maximum_value": 200 },
```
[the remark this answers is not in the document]
kotovskii: No, we want our customers to use firmware built-in feedrates and acceleration values.

resources/quality/z-bolt/bowden/ABS/v6 0.3/z-bolt_bowden_v6_0.30_ABS_extrafine.inst.cfg
kotovskii: Will be fixed soon, thanks.

resources/definitions/z-bolt_base.def.json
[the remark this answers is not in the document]
Ghostkeeper: This has nothing to do with the acceleration values.

If you want users to use the feedrate defined in the firmware, you'd need to set all speeds to 0 which would prevent Cura from outputting any `F` parameters with its movements. The current proposal would be wrong then as well, because this defines the feedrate to be 120mm/s (F7200), not to be firmware-defined but Cura-defined.

Either way, if the firmware imposes a limit on feedrate of 200mm/s in the X direction, setting `machine_max_feedrate_x` to 200 would limit the user in the same way as with the `speed_travel` set to 200, but offers more of an explanation as to why the feedrate is higher than possible. And it would also properly limit the other speeds, such as the infill speed. Important to note though is that the X and Y feedrates are normally limited separately by firmware, so the printer could still move diagonally at 282.8mm/s. If this limit works differently on your printer, you can override the maximum_value with a different formula.

resources/quality/z-bolt/direct/ABS/v6 0.3/z-bolt_direct_v6_0.30_ABS_extrafine.inst.cfg
kotovskii: How do we add cool_min_layer_time and other settings to material profiles?

Ghostkeeper: The ABS material profile is here: https://github.com/Ultimaker/fdm_materials/blob/master/generic_abs_175.xml.fdm_material

The material profile format has a number of specific settings (like `print cooling` instead of Cura's `cool_fan_speed`), but `cool_min_layer_time` is not one of them. Cura's own settings can be set by adding the `cura` namespace to the XML tag, like so:

```
<cura:setting key="cool_min_layer_time">10</cura:setting>
```

Ghostkeeper:
Thanks for the fixes!

I've marked the issues that you fixed as Resolved now. A few are remaining though. And the automated tests are failing too.

The Classic and Plus printers were deleted now with your recent changes. That will similarly break existing users' workflow, since their existing profiles will now be based on a non-existing definition instead of on old profiles. It will cause the user to be faced with a message that their printer was corrupt and the loss of all of their custom profiles. This is not an acceptable solution either.

resources/quality/z-bolt/direct/ABS/v6 0.3/z-bolt_direct_v6_0.30_ABS_extrafine.inst.cfg
kotovskii: Thank you a lot for this answer, we had a lot of struggle with those advanced material profile settings.

jellespijker:

GregValiant:
I see the Z-Bolt Classic and 3+ were added to Cura.  I'll close this.  Let me know if that is premature.

PR #10151, opened 2021-07-18
Updated Maker Made 300x Profiles

austinsk8s:
Modified printer settings to add G90 to pre-print Gcode and eliminate printing errors. Added a profile for each supported quality/material combination.

### DIFF
--- a/resources/definitions/maker_made_300x.def.json
+++ b/resources/definitions/maker_made_300x.def.json
@@ -1,6 +1,6 @@
 {
     "version": 2,
-    "name": "Maker Made 300x",    
+    "name": "Maker Made 300x",
     "inherits": "fdmprinter",
     "metadata": {
         "visible": true,
@@ -11,7 +11,7 @@
         "has_materials": true,
         "has_variants": false,
         "preferred_quality_type": "normal",
-        "has_machine_quality": false,
+        "has_machine_quality": true,
         "preferred_material": "generic_pla",
         "machine_extruder_trains":
         {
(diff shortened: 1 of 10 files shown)


resources/definitions/maker_made_300x.def.json
```diff
@@ -11,7 +11,7 @@
         "has_materials": true,
         "has_variants": false,
         "preferred_quality_type": "normal",
-        "has_machine_quality": false,
+        "has_machine_quality": true,
```
nallath, line 14: Did you check if older profiles / project files still work after this change?

austinsk8s: Old profiles specifically for the 300x machine? The original profile would cause major issues with printing due to the default start and end gcode, so I made sure this new one works. The original 300x profile was problematic anyways.

Ghostkeeper: What Nallath meant was to see if users who added their Maker Made printer to Cura in Cura 4.10 would not get configuration errors or otherwise broken profiles due to this change. It's not about the print quality (we trust that your improvements make that better) but about ensuring that we don't break users' existing data.

Ghostkeeper: Specifically, users from previous versions will have the global quality profiles active, which are no longer valid profiles after this update.

jellespijker:

GregValiant:
There hasn't been any update on this for years.  Can this be closed, or is there still interest

PR #10169, opened 2021-07-22
Fix crash on exit on MacOS if using Cura's translations

### LINKED ISSUE #9837: Cura crashes on exit if language is not English
### Application Version

4.9.1

### Platform

macOS 10.15.7

### Printer

Creality Ender 3V2

### Reproduction steps

I copied the Program from the downloaded DMG to my programs folder, keeping my previous installation of Cura (4.8), so Cura 4.9.1 gets automatically renamed to "Ultimaker Cura **2**".

Then I start Cura 4.9.1 and Quit the program (it is not necessary to change any settings or work in the program to reproduce the error).

The same issue already happened with Cura 4.9.0 (4.8 works fine).

### Actual results

The program crashes with the attached error.
[error.txt](https://github.com/Ultimaker/Cura/files/6487927/error.txt)
[cura.log](https://github.com/Ultimaker/Cura/files/6487930/cura.log)

### Expected results

The program exits without any error.

- [X] Log file
- [ ] Project file

The code gets created without any issues when using 4.9.1, it is just the program exit that is faulty.

The logs are attached above at "Actual Results".

Ghostkeeper:
If you're using Cura in a different language, it was crashing on MacOS due to being moved to the special "Ultimaker Cura" menu that MacOS creates. These were the reproduction steps:

* Change Cura's language in its preferences menu.
* Restart Cura to complete the language change.
* Exit Cura. MacOS gives an error message then.

To fix this, certain menu entries should not be translated. Translating them causes Cura to update the menu entry's name upon closing down Cura, but the menu entry can't be found in its original place any more. The Qt documentation warns us that signals on the moved Menu elements won't work well any more.

The items changed below are the items that are moved by MacOS. Only those should not be translated any more. The rest should still be translated. MacOS itself should translate the moved menu items into the user's system language (which is out of Cura's control and may be different from the language preference in Cura).

Fixes CURA-8245 and fixes #9837.

### DIFF
--- a/resources/qml/MainWindow/ApplicationMenu.qml
+++ b/resources/qml/MainWindow/ApplicationMenu.qml
@@ -109,7 +109,7 @@ Item
             //For more information, see:
             //- https://doc.qt.io/qt-5/macos-issues.html#menu-bar
             //- https://doc.qt.io/qt-5/qmenubar.html#qmenubar-as-a-global-menu-bar
-            title: (Qt.platform.os == "osx") ? "P&references" : catalog.i18nc("@title:menu menubar:toplevel", "P&references")
+            title: (Qt.platform.os == "osx") ? "&Preferences" : catalog.i18nc("@title:menu menubar:toplevel", "P&references")
 
             MenuItem { action: Cura.Actions.preferences }
         }
--- a/resources/qml/MainWindow/ApplicationMenu.qml
+++ b/resources/qml/MainWindow/ApplicationMenu.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Ultimaker B.V.
+// Copyright (c) 2021 Ultimaker B.V.
 // Cura is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.7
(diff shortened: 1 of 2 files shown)


resources/qml/MainWindow/ApplicationMenu.qml
```diff
@@ -91,7 +101,15 @@ Item
...
-            title: catalog.i18nc("@title:menu menubar:toplevel", "P&references")
+
+            //On MacOS, don't translate the "Preferences" word.
+            //Qt moves the "preferences" entry to a different place, and if it got renamed can't find it again when it
+            //attempts to delete the item upon closing the application, causing a crash.
+            //In the new location, these items are translated automatically according to the system's language.
+            //For more information, see:
+            //- https://doc.qt.io/qt-5/macos-issues.html#menu-bar
+            //- https://doc.qt.io/qt-5/qmenubar.html#qmenubar-as-a-global-menu-bar
+            title: (Qt.platform.os == "osx") ? "&Preferences" : catalog.i18nc("@title:menu menubar:toplevel", "P&references")
```
nallath, line 112:
```suggestion
            title: (Qt.platform.os == "osx") ? "P&references" : catalog.i18nc("@title:menu menubar:toplevel", "P&references")
```
I assume it should be the same in both cases right?

nallath (changes requested):

resources/qml/MainWindow/ApplicationMenu.qml
Ghostkeeper: It doesn't have to be the same letter in both cases. It's not the same in every language either.

nallath (approved):

PR #10182, opened 2021-07-25
Add Weedo X40 printer

x40-Community:
The new Weedo X40 printer was tested with the Cura profile for several months by the X40 community.
The initial upload does not yet contain an intent profile.


resources/quality/weedo_x40/0.4_generic_ABS_Draft.inst.cfg
```diff
@@ -0,0 +1,38 @@
...
+[metadata]
+setting_version = 17
+type = quality
+quality_type = draft
+weight = -1
+material = generic_abs_175
+variant = 0.4mm Nozzle
+
+[values]
+layer_height = 0.24
```
nallath, line 15: Layer height should not be set in extruder profiles.

resources/definitions/weedo_x40.def.json
```diff
@@ -0,0 +1,334 @@
...
+            "ultimaker_cpe_plus_white",
+            "ultimaker_cpe_red",
+            "ultimaker_cpe_transparent",
+            "ultimaker_cpe_white",
+            "ultimaker_cpe_yellow",
+            "ultimaker_nylon_black",
+            "ultimaker_nylon_transparent",
+            "ultimaker_pc_black",
+            "ultimaker_pc_transparent",
+            "ultimaker_pc_white",
```
nallath, line 134: You don't need to exclude all of these, they are 2.85 materials. Since the machine is 1,75, they will never show up anyway.

resources/quality/weedo_x40/0.4_generic_ABS_Extra_Fine.inst.cfg
```diff
@@ -0,0 +1,38 @@
+[general]
+version = 4
+name = Default - Extra Fine
```
nallath, line 3: Why prepend all the names with default? If everything is default, it doesn't change anything to add it.

I'd suggest renaming this to just "Extra Fine"

resources/quality/weedo_x40/0.4_generic_PLA_Extra_Fine.inst.cfg
```diff
@@ -0,0 +1,38 @@
...
+
+speed_print = 30
+speed_layer_0 = =round(speed_print / 2.5, 2)
+speed_print_layer_0 = =speed_layer_0
+speed_infill = =speed_print
+speed_support_interface = =speed_print
+speed_wall = =round(speed_print / 2, 2)
+speed_wall_0 = =speed_wall
+speed_wall_x = =speed_wall
+speed_topbottom = =speed_wall
```
nallath, line 38: Why set this for every profile? If you want this to be default for your printer, just change it in the definition. This makes it a lot easier to see what the profile actually changed.

I'd advise to double check all the settings in the profiles and check if you really need to set them.

resources/definitions/weedo_x40.def.json
```diff
@@ -0,0 +1,334 @@
...
+        },
```
mah115, line 224: You included M301 PID values for your custom cooling system.
M420 not needed, G29 will automatically enable bed leveling
M107 need P0/P1 or it'll only control the left fan
Typrically M109 is used for hotend temperature, not M192

x40-Community:
Desired changes introduced

x40-Community (commented):
Desired changes introduced

Ghostkeeper (changes requested):
Doing a full review now, and including the latest changes.

Developers, see ticket CURA-8469.

plugins/PostProcessingPlugin/scripts/X40Snapshot.py
```diff
@@ -0,0 +1,80 @@
+from UM.Logger import Logger
```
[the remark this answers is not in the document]
Ghostkeeper: We'd like to request that this script gets merged with the existing post-processing script. And I can help out a bit with the transfer if necessary; just ask :)

This is how I envision it working nicely for others and for the Weedo users:
* In CreateThumbnail.py, add a new parameter to specify the format, a drop-down with two options, "Octoprint" and yours. Octoprint should be the default. Your option could be called "Weedo" since your W220 etc. commands seem to be specific to Weedo printers. Or maybe you could call it "W221".
* In the `_encodeSnapshot` function you can read this parameter and choose between PNG+base64 for Octoprint and JPG+Hex for Weedo.
* The `_convertSnapshotToGcode` needs to then choose different prefixes, etc. for Weedo but can otherwise stay mostly the same.

This solution is desirable because of two sides:
* It reduces code duplication.
* It reduces user confusion as to which post-processing script to choose.

In early versions of Cura we had various flavours of Pause at Height for different g-code flavours and it was causing people to be confused as to which one they should choose. Each script had different parameters and different features. Combining them made things more clear for users and also allowed us to give all of these pause methods the same features, resolve bugs all at once instead of needing to test all scripts separately, etc.

mah115 (changes requested):
The start gcode needs to be fixed.

- Should not set extruder PID values with M301 in start gcode, it could mess up someone's printer if they try to print the pre-sliced gcode.
- Should not set M92 in start gocde, other people's extruders may require a different value, especially if they have mods.
- M109 is used for hotend temperature, not M192
- The following gcode is not necessary and dangerous because some printers have an offset less than 352, and this will crash the extruder.  After G28 the right extruder should already be in its home position.
```
T1 S ; select right extruder
G1 X352 F3000 ; Move right nozzle to parking position
```

x40-Community:
> The start gcode needs to be fixed.
> 
>     * Should not set extruder PID values with M301 in start gcode, it could mess up someone's printer if they try to print the pre-sliced gcode.
> 
>     * Should not set M92 in start gocde, this not a universal setting.  The default values are spot on for my machine.  Also, other people's extruders may require a different value, especially if they have mods.
> 
>     * M109 is used for hotend temperature, not M192
> 
>     * The following gcode is not necessary and dangerous because some printers have an offset less than 352, and this will crash the extruder.  After G28 the right extruder should already be in its home position.
> 
> 
> ```
> T1 S ; select right extruder
> G1 X352 F3000 ; Move right nozzle to parking position
> ```

The commands M301 and M92 are definitely not removed from the start code. Hundreds of print profiles have been created on this basis and not every user has the specialist knowledge to set it in the firmware accordingly.
You can remove it for yourself without any problems!
I will examine the effects of M191, as long as it lasts.

plugins/PostProcessingPlugin/scripts/X40Snapshot.py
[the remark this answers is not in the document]
Ghostkeeper: This script has been removed completely now. No support for these screenshots then, but it should work without too.

Ghostkeeper:
Good changes so far! Two of my previous review comments are still standing though.

Ghostkeeper:
(Failing tests are unrelated to this PR.)

Ghostkeeper (approved):
All review comments have been processed. Looks good now!

I'll give this a good test. Quite a few profiles to go through now.

Ghostkeeper:
Looks like it's working fine. All combinations of materials, nozzles and quality levels are slicing without errors. Thanks for the contribution!

PR #10188, opened 2021-07-26
Graphic Buffers Update

bremco:
Front-end implementation of https://github.com/Ultimaker/Uranium/pull/714.
(See also this discussion https://github.com/Ultimaker/Uranium/issues/664.)

Due to technical details (see the link above), meshes ranged by index are now rendered by specifying the range of vertex-indices in the shaders, rather than specifying the element range beforehand. (Manually, as `gl_primitiveID` couldn't be used because of the legacy support... also this gives more control, potentially.)

Potential bonus:
- The compatibility mode, (for legacy OpenGL/shaders,) could be made compatible with Simulation-View relatively easy now. Perhaps I could even do that in this PR. Keeping this relatively small for now is better though!

Also not in this PR (yet?):
- RenderBatch's are still recreated on each draw. As so much of its internals are cached, either directly or via the classes it uses this might not present a problem in practice.
- In Simulation-View, the range to show is recalculated for each render, even when it makes a lot more sense to cache those calculations. An earlier attempt has been made at this, then reverted, however. Rather than complicate this PR by also attempting that, leave it alone for now.
- The geometry shaders are still used. That has a little less to do with _this_ particular update (though related), but for completeness sake, I'd like to say that we could maybe also pre-calculate/upload the 'tubes' to the graphics card instead of re-calculating them in the shaders every redraw (especially since the common wisdom is that 'geometry shaders are slow' ... not sure how true that actually is though).


rburema:
Closing this in favour of https://github.com/Ultimaker/Cura/pull/10259 due to laptop & account related shenanigans.

PR #10192, opened 2021-07-26
Improve FLSUN Super Racer profile

Guilouz:
- Added quality profiles based on profiles provided by FLSUN
- Improved machine settings based on FLSUN's Marlin sources
- Replaced Super Racer STL with correct one 

Capture d’écran 2021-08-07 à 14 54 21

Capture d’écran 2021-08-07 à 14 53 55


resources/definitions/flsun_SuperRacer.def.json
```diff
@@ -1,18 +1,19 @@
...
+        "platform": "flsun_SuperRacer.stl",
+        "platform_offset": [0,-35,-40],
         "file_formats": "text/x-gcode",
         "has_materials": true,
-        "has_machine_quality": false,
+        "has_machine_quality": true,
         "preferred_quality_type": "fast",
         "machine_extruder_trains": {
-            "0": "flsun_sr_extruder_0"
+            "0": "flsun_SuperRacer_extruder_0"
```
nallath, line 16: If you rename this, you will also need to create an upgrade script.

Since this requires a fair amount of python coding, i would not readily recommend doing that.

nallath:
The STL file that you provided is a lot larger than the previous one. We hold a max of 1mb for a model, but ask you to stay well below that.

Also, the CI fails, so please resolve those issues.

resources/definitions/flsun_SuperRacer.def.json
Guilouz: Why ? The release containing the profile has not yet been published, I tested this configuration before creating this PR and it works perfectly.

nallath: My bad! I didn't check if they were already published. You're completely right.

Guilouz:
> The STL file that you provided is a lot larger than the previous one. We hold a max of 1mb for a model, but ask you to stay well below that.
> 
> Also, the CI fails, so please resolve those issues.

It's fixed.

Guilouz:
All is clear now, ready to be merged.

Guilouz:
@konskarm you can merged it :)

konskarm:
Hi @Guilouz! My apologies for the delayed response. Since you are changing the FLsun Super Racer so that it no longer uses global qualities but it uses it's own, this means that a VersionUpgrade script is needed (from 4.11, which is coming soon, to 4.12, which is where the updated profile will appear). 

This is necessary because if a user has an FLSun Super Racer in 4.11, that printer will still be using global qualities in 4.12. So the versionUpgrader should make sure to map the previously used global profile that the user had selected to the newly added machine quality profiles.

You could try and create the VersionUpgrade for the FLSun Super racer improvements! A similar thing happened with TwoTrees Bluer printers in 4.10, where quality profiles were introduced for them. So the `VersionUpgrader49to410.py` script is a good example on how to do that.  I'd suggest you have a look at the `VersionUpgrader49to410.py` and try to create a `VersionUpgrader411to412.py` that fulfills the quality profile mapping. 

If you are unable to do that, we can do it, but that means that your PR will take time to be moved forward until one of our devs has enough time to create the versionUpgrader.

konskarm (changes requested):
A VersionUpgrade script is required to map the previously used global qualities to the newly added FLsun quality profiles.

Guilouz:
@konskarm  Not really understand why to use VersionUpgrade because FLSUN Super Racer printer profile is not yet included in any version of Cura (not present in 4.10). So why can't this be included in version 4.11 directly?

konskarm:
Oh you are right! I thought this printer existed already in 4.10 (even though I was the one that merged it after 4.10). Disregard my comment and review, this is ready to be merged.

konskarm (approved):

konskarm:
@Guilouz How come is the width and depth increased a little bit and the height is decreased? In the product's page it is mentioned as 260x330mm.

resources/quality/flsun_sr/flsun_sr_fine.inst.cfg
```diff
@@ -0,0 +1,86 @@
...
+support_angle = 60
+support_brim_enable = True
+support_brim_width = 4.0
+support_interface_density = 60
+support_interface_enable = True
+support_interface_height = 0.96
+support_interface_pattern = grid
+support_interface_skip_height = 0.12
+support_roof_density = 60
+support_wall_count = 1
```
konskarm, line 52: This setting is mentioned twice, so this line can be removed.
```suggestion
```

resources/quality/flsun_sr/flsun_sr_normal.inst.cfg
```diff
@@ -0,0 +1,86 @@
...
+support_angle = 60
+support_brim_enable = True
+support_brim_width = 4.0
+support_interface_density = 60
+support_interface_enable = True
+support_interface_height = 0.6
+support_interface_pattern = grid
+support_interface_skip_height = 0.2
+support_roof_density = 60
+support_wall_count = 1
```
konskarm, line 52: This setting is mentioned twice, so this line can be removed.
```suggestion
```

Guilouz:
> @Guilouz How come is the width and depth increased a little bit and the height is decreased? In the product's page it is mentioned as 260x330mm.

On the product page yes but not in reality in their Marlin firmware, take a look : 

  // Print surface diameter/2 minus unreachable space (avoid collisions with vertical towers).
  #define DELTA_PRINTABLE_RADIUS 132    // (mm)
  
  // Distance between bed and nozzle Z home position
  #define DELTA_HEIGHT 320

#define MANUAL_Z_HOME_POS DELTA_HEIGHT
  
 // Travel limits (mm) after homing, corresponding to endstop positions.
#define X_MIN_POS -(DELTA_PRINTABLE_RADIUS)
#define Y_MIN_POS -(DELTA_PRINTABLE_RADIUS)
#define Z_MIN_POS 0
#define X_MAX_POS DELTA_PRINTABLE_RADIUS
#define Y_MAX_POS DELTA_PRINTABLE_RADIUS
#define Z_MAX_POS MANUAL_Z_HOME_POS

So diameter is 264mm and height is 320mm.

Guilouz:
@konskarm Changes are done.

konskarm:
I don't see any other issues with it so I will merge it. Since we released the 4.11 beta today, these changes did not make it in. However, if we see that there are only a few slices with this profile in 4.11beta, I'll cherry-pick the changes for the 4.11 stable without a version upgrade script. So this will be mildly inconvenient for the few beta users. 

On the other hand, if there are many slices in the beta, I will leave these changes for 4.12 and I will make a version upgrade script to map the global qualities to the new qualities that you added.

If I hadn't been confused for the necessity of the version upgrade script, this would've made it to 4.11 and we would've avoided this situation. My apologies for this.

thushan:
@Guilouz initially I went with the firmware too but sending the profile to FLSun to review, they advised to use the dimensions given that I had instead, but agree the firmware is the source of truth.

Guilouz:
@thushan When we see how FLSUN coded the firmware of the SR... We can have doubts about their words...

I have already share this changes on many facebook groups and many people used it without issue, and is in line with the firmware values.

Guilouz:
@thushan If you have a SR, you can check my Marlin firmware on my profile page ;)

thushan:
> @thushan If you have a SR, you can check my Marlin firmware on my profile page ;)

I do indeed, will check it out. Been using the updated profile since your first round of changes (wasn't too worried about the STL).

Ghostkeeper:
A version upgrade plug-in was implemented here: https://github.com/Ultimaker/Cura/pull/10449

PR #10195, opened 2021-07-27
Fix up spelling 

aidbish:
The word information was incorrect


nallath:
Ah! That slipped through our review. Thanks for taking the time to fix it!

PR #10204, opened 2021-07-29
[CURA-8442] Wall min flow doesn't make sense for Arachne.

rburema:
Any walls we get from Arachne should be printed, as such thin wall considerations are already taken into account when that is run.

See backend PR: https://github.com/Ultimaker/CuraEngine/pull/1475


jellespijker (approved):
No further comments. Looks good and upgrader has been taken into account.

PR #10206, opened 2021-07-29
Anycubic i3 Mega Dual Extruder

gadgev:
Anycubic i3 Mega Dual Extruder with single nozzle


resources/definitions/anycubic_i3_mega_dual.def.json
```diff
@@ -0,0 +1,127 @@
...
+        "travel_avoid_supports": { "value": "True" },
+        "travel_avoid_distance": { "value": "3 if extruders_enabled_count > 1 else machine_nozzle_tip_outer_diameter / 2 * 1.5" },
+        "wall_0_inset": { "value": "0" },
+        "wall_line_width_x": { "value": "round(line_width * 0.3 / 0.35, 2)" },
+        "wall_thickness": { "value": "line_width * 3" },
+
+        "optimize_wall_printing_order": { "value": "True" },
+        "retraction_combing": { "default_value": "all" },
+        "initial_layer_line_width_factor": { "value": "120" },
+        "zig_zaggify_infill": { "value": "gradual_infill_steps == 0" },
```
nallath, line 125:
```suggestion
        "zig_zaggify_infill": { "value": "gradual_infill_steps == 0" }
```
With the comma this is invalid JSON. Please actually test if the machine loads before contributing it to us.

nallath (changes requested):

resources/definitions/anycubic_i3_mega_dual.def.json
gadgev: The machine is loaded but without variants. With placing variants,  stops loading and I can not understand where the problem is.

Ghostkeeper: Maybe we can help if you show the error message you were getting?

gadgev:
> I've tried adding the machine in Cura, but all the profiles are marked as "not supported" because you didn't add any extruder qualities.
> 
> Is that intentional?

Yes, this is part of the problem because the variants and the quality of the extruders are connected in some way. But even with qualities get the same.

Issue #6267, opened 2019-08-27
[4.2.1] Creality Ender 3 Bed Size Incorrect


kazooless:
**Application version**
4.2.1

**Platform**
I use MacOS, but I'm sure this applies to all OS'

**Printer**
Creality3D Ender 3

**Reproduction steps**
Add printer, go to machine settings, look at bed size definition. Print any bed center template or leveling stl, it will not be centered. Instead, it is closer to X0,Y0. 

**Actual results**
220x220
 Print any bed center template or leveling stl, it will not be centered. Instead, it is closer to X0,Y0. 

**Expected results**
235x235
 Print any bed center template or leveling stl, it will be centered.

**Additional information**
~~I've actually been seeing this since 3.6.0 but I was new to 3D printing back then.~~
(Checking v4.0.0, the 235x235 was defined correctly. I haven't gone back to check other versions yet.)

 The Ender-3 is advertised as 220x220x250 but the bed is in reality 235x235. The heating element is what is 220x220.

Liger0:
the cr10s is 310x 310 but because of the switches and gantry you can't use 5mm on each side so only 300x300 is usable. probably it's similaron the ender.

Ghostkeeper:
According to the [website of the manufacturer](https://creality3d.shop/collections/ender-series-3d-printer/products/creality-ender-3-3d-printer-economic-ender-diy-kits-with-resume-printing-function-v-slot-prusa-i3-220x220x250mm), the build volume is 220x220x250mm. I'm inclined to believe that more.

Keep in mind that some variation is always introduced by the manufacturing and your assembly. So for some builds of the printer it might be bigger than for others. I'm guessing that the manufacturer took the safe side of the numbers.

Ghostkeeper:
> the cr10s is 310x 310 but because of the switches and gantry you can't use 5mm on each side so only 300x300 is usable. Probably it's similar on the ender.

Cura has a mechanism to allow for that, by the way. With the `machine_disallowed_areas` setting, areas can be specified where no objects may be placed.

kazooless:
I'm 100% positive this is not some variation amongst different packages. You can look at their GitHub repository and it has all of the mechanical drawings and such. I have one and I've worked on a friend's and I've worked with people in the FB forums. The physical bed is 235x235 and every single one of them can move that nozzle past all 4 edges. 

[...]

PR #10216, opened 2021-08-01
Add profile for Ender3 v2

pcartwright81:
Add profile for Ender3 v2

### DIFF
--- a/resources/definitions/creality_ender3v2.def.json
+++ b/resources/definitions/creality_ender3v2.def.json
@@ -1,0 +1,33 @@
+{
+    "name": "Creality Ender-3 v2",
+    "version": 2,
+    "inherits": "creality_base",
+    "metadata": {
+        "quality_definition": "creality_base",
+        "visible": true,
+        "platform": "creality_ender3.3mf"
+    },
+    "overrides": {
+        "machine_name": { "default_value": "Creality Ender-3 v2" },
+        "machine_width": { "default_value": 220 },
+        "machine_depth": { "default_value": 220 },
+        "machine_height": { "default_value": 250 },
+        "machine_disallowed_areas": {
+            "default_value": [
+              [[-117.5, 117.5], [-117.5, 108], [117.5, 108], [117.5, 117.5]],
+              [[-117.5, -108], [-117.5, -117.5], [117.5, -117.5], [117.5, -108]]
+        ]},
+        "machine_head_with_fans_polygon": { "default_value": [
+                [-26, 34],
+                [-26, -32],
+                [32, -32],
+                [32, 34]
+            ]
+        },
+        "machine_start_gcode": {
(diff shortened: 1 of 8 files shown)


konskarm:
I notice that the only difference between the Ender3 and the Ender3v2 files is the size of the printer. Keep in mind that the Ender3 used to also be 220x220x250mm but it was changed in commit https://github.com/Ultimaker/Cura/commit/726561bf309616107ad874d4e508b3a3ec9e5701 after the discussion in https://github.com/Ultimaker/Cura/issues/6267. 

Is there a change the same situation will arise in the Ender3v2? Because if that's the case, I think the Ender3v2 users can be covered by the Ender3 printer profile so that we don't have to have two different printer profiles that are exactly the same.

emarforio:
I for one would appreciate to have the "duplicate" config. I am very new to 3D-printing and just bought my first printer, an Ender 3 V2, and were a bit confused and put off by not finding the exact model at first. So even if the config is the same it is nice to show newbies like me that their device is supported out of the box.

pcartwright81:
> I for one would appreciate to have the "duplicate" config. I am very new to 3D-printing and just bought my first printer, an Ender 3 V2, and were a bit confused and put off by not finding the exact model at first. So even if the config is the same it is nice to show newbies like me that their device is supported out of the box.

Agree.  It was a little confusing at first why I had to manually copy the profile and set the dimensions.  Didn't know how to respond to the pull request comment.  Was hoping someone would say whether or not the dimensions needed to be updated or not.

mdehaas:
I do not agree that because users are confused (and not willing to investigate further) there is a need to create duplicate profiles. This causes changes to the profile (and possibly code) need to be applied on multiple files. This will eventually lead to drift (forgotten or orphaned profiles). The solution in my opinion should be that if a profile applies to multiple vendors/series/models/editions there should be a possibility to add more than one of the above to the profile.
This however will lead to a lengthy list in the configuration dialog so a new 'user journey' has to be modelled around that.

Ghostkeeper:
For one, it would be nicer to inherit from the Ender 3 (v1) profile if it's just the size that is different, and then override just that size. That way if other things change in the Ender 3 profile, like an improved start g-code, it would also change for the v2.

The profiles are already shared among all Creality printers anyway, so it wouldn't lead to duplication there, or outdatedness.

Another possibility instead of this would be to add a machine action that allows the user to choose which edition of the printer they have. The Ultimaker Original does something like that, where you get another screen after adding the printer where the user can select whether they have the heated bed installed in their printer or not.

mattpker:
To solve the issue of confusion, the profile for "Ender-3" could be renamed to "Ender-3 (v1/v2)".

Medduus:
in  the endeer slicer that is based on cura u can get the profiles for the v2

amweiss:
> I notice that the only difference between the Ender3 and the Ender3v2 files is the size of the printer. Keep in mind that the Ender3 used to also be 220x220x250mm but it was changed in commit [726561b](https://github.com/Ultimaker/Cura/commit/726561bf309616107ad874d4e508b3a3ec9e5701) after the discussion in #6267.
> 
> Is there a change the same situation will arise in the Ender3v2? Because if that's the case, I think the Ender3v2 users can be covered by the Ender3 printer profile so that we don't have to have two different printer profiles that are exactly the same.

This is correct, the bed size between the Ender 3 and Ender 3v2 is the same. It's actually 235x235x250 but only prints to 220x220x250. I would agree with the above suggestion to just rename the profile to include "(v1/v2)"

digitalfrost:
Can be closed since Rename Ender-3 profile to include v2 #12442 has been merged.

Issue #8909, opened 2020-12-08
Add Monotonic infill pattern, like in Prusaslicer alpha (2.3.0)


Ntrino:
**Is your feature request related to a problem? Please describe.**
Yes, when a hole is going through the top layer, the infill pattern creates a visible line stemming from the hole, not following the path of the rest
Monotonic infill pattern, like in Prusaslicer alpha (2.3.0) seems to solve this problem

Left - Cura | Right - Monotonic infill pattern (Prusaslicer)
![Untitled](https://user-images.githubusercontent.com/24448925/101551411-d065f880-3a15-11eb-9ba9-825ce7932e83.png)

**Describe the solution you'd like**
Adding the same Monotonic infill pattern feature that Prusaslicer alpha (2.3.0) has

**Describe alternatives you've considered**
No other settings in Cura seem to address this issue

**Affected users and/or printers**
I would assume everyone

**Additional context**
Please add it : )

Liger0:
That would be an awesome addition applied on the optional "Top surface skin layer".
I'm not sure if it increases the printing time, in which case leaving the basic top layers printing with the old pattern is more convenient. Maybe it would also be useful for the very bottom layer.

rburema:
We've made a ticket on the backlog (now known to us internally as CURA-7928), but have not prioritised it yet.

smartavionics:
I have added a monotonic skin lines setting to my latest release https://github.com/smartavionics/Cura/releases/tag/20210214. It's not a sophisticated implementation and doesn't attempt to minimise the travel moves required but it does appear to work. If you try it out and have any feedback, please open an issue at https://github.com/smartavionics/Cura rather than commenting here.

![IMG_20210214_141838890](https://user-images.githubusercontent.com/585618/107924246-8dbb2200-6f6a-11eb-8e57-5d44e6df6f0e.jpg)

Minner1337:
I would also love to see this in an official release. +1

gohigher14:
This is such a super option to have on cura. Prusaslicer slicer has this already , so please add this also to cura. That's would be a great improvement.

kylane:
@rburema - Has it progressed in the priority list as yet? Really need this.

smartavionics:
Still available at https://github.com/smartavionics/Cura/releases :)

rburema:
@kylane We discussed it in refinement with the team today. It seems we do really want this. Don't count on it for the next version though.

TheAdamWoods:
[...]

PR #10222, opened 2021-08-02
Monotonic ordering of skin, top surface and ironing

Ghostkeeper:
This adds a number of settings that allow the skin, top surface and ironing to be printed in a monotonic order.

For a more thorough explanation, see the CuraEngine sister-PR. Please contain the discussion in that PR as well.

https://github.com/Ultimaker/CuraEngine/pull/1476

Contributes to CURA-7928. Implements https://github.com/Ultimaker/Cura/issues/8909.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -6405,6 +6405,17 @@
                     "settable_per_mesh": true,
                     "enabled": "roofing_layer_count > 0 and top_layers > 0"
                 },
+                "roofing_monotonic":
+                {
+                    "label": "Monotonic Top Surface Order",
+                    "description": "Print top surface lines in an ordering that causes them to always overlap with adjacent lines in a single direction. This takes slightly more time to print, but makes flat surfaces look more consistent.",
+                    "type": "bool",
+                    "default_value": false,
+                    "value": "skin_monotonic",
+                    "enabled": "roofing_layer_count > 0 and top_layers > 0 and roofing_pattern != 'concentric'",
+                    "limit_to_extruder": "roofing_extruder_nr",
+                    "settable_per_mesh": true
+                },
                 "roofing_angles":
                 {
                     "label": "Top Surface Skin Line Directions",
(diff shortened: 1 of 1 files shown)


fieldOfView:
Would it be hard to implement this for the bottom too? I think many people will want their bottoms clean too...

Ghostkeeper:
Best keep the discussions confined to the CuraEngine ticket! But this currently applies to all skin, top surface and ironing, which includes the bottom skin, yes.

rburema (approved):
(Also see the engine review of course! There I _do_ request a number of minor changes.)

Just to check: There is no way `infill_enable_travel_optimization` is going to be 'on' for the same sort of lines that monotonic is going to be 'true' for right? Otherwise we should prevent it here in the `enabled`s.

resources/definitions/fdmprinter.def.json
```diff
@@ -1519,6 +1519,16 @@
                     "limit_to_extruder": "top_bottom_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "skin_monotonic":
+                {
+                    "label": "Monotonic Top/Bottom Order",
+                    "description": "Print top/bottom lines in an ordering that causes them to always overlap with adjacent lines in a single direction. This takes slightly more time to print, but makes flat surfaces look more consistent.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": "(top_layers > 0 or bottom_layers > 0) and top_bottom_pattern != 'concentric'",
```
nallath, line 1528:
```suggestion
                    "enabled": "(top_layers > 0 or bottom_layers > 0) and (top_bottom_pattern != 'concentric' or top_bottom_pattern_0 != 'concentric')",
```
If the top_bottom_pattern_0 is not concentric, but the other pattern is, I'd argue that this setting should still be active, right?

Ghostkeeper:
> Just to check: There is no way infill_enable_travel_optimization is going to be 'on' for the same sort of lines that monotonic is going to be 'true' for right? Otherwise we should prevent it here in the enableds.

Currently `infill_enable_travel_optimization` only applies to infill, and these settings don't apply to infill. If there is ever a skin-version of that setting, we should make that setting invisible if monotonic is enabled since currently monotonic is not optimising travel moves at all (using combing or otherwise). If we implement monotonic to optimise which monotonic segment is closest, we can re-enable the travel optimisation again also for monotonic.

nallath (approved):

rburema:
> Currently `infill_enable_travel_optimization` only applies to infill, and these settings don't apply to infill.

Good, just as I thought, then.

PR #10246, opened 2021-08-06
Official Arjun Pro 300

venkatkamesh:


nallath:
I couldn't find any information regarding this printer, or why there is a "duplication" and a "mirror" printer. Could you explain a bit more what the logic of that is? As far as I can see at first glance these machines are pretty much identical.

venkatkamesh:
> I couldn't find any information regarding this printer, or why there is a "duplication" and a "mirror" printer. Could you explain a bit more what the logic of that is? As far as I can see at first glance these machines are pretty much identical.

yes arjun300 and arjun pro 300 had slight differences like 

- integrated direct drive extruder, since the bed size is same but the space occupation by dd_extruder, we have write this def
- independent z bed alignment. for that we used bltouch so start code is changed.
- mirror mode and duplication mode - as this is supported with Dual x carriage and to enable these definition got required. 
- by model the functional implementations are changed.
- we supplied machines to industries and institutions in india. we had nearly 200products of different models were supplied as of 1 year time.
  
Thank you for your time on inspecting this. please let us know if any changes are required.

Ghostkeeper:
Hmm, but is there any publicly available information on this printer? Is it being sold anywhere?

resources/definitions/arjunpro_duplication.def.json
```diff
@@ -0,0 +1,49 @@
+{
+    "version": 2,
+    "name": "Arjunpro 300 Duplication",
```
Ghostkeeper, line 3: You've spelled it `Arjunpro` here, but `Arjun Pro` earlier.

resources/definitions/arjunpro_mirrored.def.json
```diff
@@ -0,0 +1,49 @@
+{
+    "version": 2,
+    "name": "Arjunpro 300 Mirror",
```
Ghostkeeper, line 3: You've spelled it `Arjunpro` here, but `Arjun Pro` earlier. I'd think the `Arjun Pro` name is more what most people would think the real name of the printer is, right?

Ghostkeeper (changes requested):
A few small remarks in this code review. Mostly looks fine.

resources/definitions/arjunpro_mirrored.def.json
venkatkamesh: actually yes, but on printer make it was printed as arjun 300 pro instead arjun pro 300.
our team head forced us to change and to use like arjunpro 300 as definition /title .
most of our customers are from industries and colleges/institutes.

below are images of our arjun and arjun pro
WhatsApp Image 2021-09-22 at 8 20 28 PM
20191106181942_2J2A0897 JPG
WhatsApp Image 2021-09-22 at 8 12 58 PM

Ghostkeeper:
Is there any publicly available information on this printer? Is it being sold anywhere?

We don't normally add any printers that are still in development.

venkatkamesh:
> Is there any publicly available information on this printer? Is it being sold anywhere?
> 
> We don't normally add any printers that are still in development.

this is not a prototype this is fully functional new model we have sold 8 ((since 2 months)of this new model. if you need we can personally send the bill copies. thank you

PR #10259, opened 2021-08-09
Graphics Buffer Update -- New PR

rburema:
_The HD of my private laptop is fried. While I resolve that, it's easier to work on the laptop from work. Since switching accounts back and forth can be a big pain and lead to confusion, I'm developing these from my work-account._

_I now return you the text of the original PR_:

Front-end implementation of ~originally: Ultimaker/Uranium#714.~ new: Ultimaker/Uranium#721
(See also this discussion Ultimaker/Uranium#664.)

Due to technical details (see the link above), meshes ranged by index are now rendered by specifying the range of vertex-indices in the shaders, rather than specifying the element range beforehand. (Manually, as gl_primitiveID couldn't be used because of the legacy support... also this gives more control, potentially.)

Potential bonus:

The compatibility mode, (for legacy OpenGL/shaders,) could be made compatible with Simulation-View relatively easy now. Perhaps I could even do that in this PR. Keeping this relatively small for now is better though!
Also not in this PR (yet?):

RenderBatch's are still recreated on each draw. As so much of its internals are cached, either directly or via the classes it uses this might not present a problem in practice.
In Simulation-View, the range to show is recalculated for each render, even when it makes a lot more sense to cache those calculations. An earlier attempt has been made at this, then reverted, however. Rather than complicate this PR by also attempting that, leave it alone for now.
The geometry shaders are still used. That has a little less to do with this particular update (though related), but for completeness sake, I'd like to say that we could maybe also pre-calculate/upload the 'tubes' to the graphics card instead of re-calculating them in the shaders every redraw (especially since the common wisdom is that 'geometry shaders are slow' ... not sure how true that actually is though).


Ghostkeeper (dismissed):
A few minor things, but also a major question:

I don't really understand the need for this type change count you're adding. It seems like it's added to the index where the current layer ends? But why does that matter?

rburema:
@Ghostkeeper Thanks for taking this up btw!

> ..., but also a major question:
> I don't really understand the need for this type change count you're adding. It seems like it's added to the index where the current layer ends? But why does that matter?

So, in the shader(s), since properties are either set globally or on a per-vertex basis*, we duplicate one vertex whenever we switch line-type, because otherwise you'd get a line where the two vertices have different types (think a vertex with the extrusion line type on one end and a vertex with travel on the other end).

However, as this is a workaround for a reality in rendering** (and at least _somewhat_ specific to SimulationView), anywhere else it's just the vertices or line-segments you expect. In order to start with the proper line segment in the layer-view, we need to take the number of type-changes into account.

Possibly, this could also have happened within the ranges itself, but that would complicate Uranium with knowledge it, as a library, shouldn't have, so to speak. (Since the RenderBatch, and the newly introduced ranges, can be used for more than just the simulation-view.) 

(*There are .... ways around that, but I don't think it's pleasant, possibly less performant, and would require rewriting (not just) the shaders.)
(** I understand it's not _just_ OpenGL that does it this way.)

rburema:
Also, I can't really reproduce the stutter here. That must've been something unique to my old Linux laptop then?

rburema:
Requested documentation added! :-)

Ghostkeeper (approved):
Thanks for the fixes. Looks good now.

The documentation is good to add to the actual code, rather than just in this PR. In these comments it will get lost in time. Or lost in SaaS, whichever is first to go. So thanks :)

Ghostkeeper:
Loaded up a big 20cm cube with layer height 0.06mm and 100% density gyroid infill. This PR changed the frame rate from 1.3fps to 5.8fps.

It did appear to vary a lot more than it did before. Some frames still took about a second to render (though I didn't measure this) but most frames are a lot faster, and the slow frames now are comparable to the normal frames before.

Other tests I performed:
* Loading 1GB model in prepare mode, though I didn't really measure performance analytically. The frames were already quite smooth and I didn't notice any change. The noticeable change is definitely in layer view only.
* Lay Flat On Face still works fine.
* Overhang rendering still works fine.
* X-ray view still works fine.
* Selection still works fine.
* Non-manifold rendering with polka dot pattern still works fine.
* Layer view compatibility mode: A new bug arose here.

So I found 1 bug. This is how Compatibility Mode for layer view used to look. The top 5 layers are rendered with full line width, the rest are rendered with thin lines only:

Screenshot from 2021-11-08 15-36-33

However this is how Compatibility Mode for layer view now looks. The lines-rendering is not stopped at the layer slider:

Screenshot from 2021-11-08 15-35-53

This will need fixing before this change can be merged. It works with any model in layer view once compatibility mode is enabled in the preferences and Cura is restarted.

rburema:
Should be fixed now! I thought I had already done this, but apparently not.

I almost added the bottom slider into the compatibility-view, but I ran into the different rendering of the top layers. Let's not add more risk to this now that it's almost properly finished though.

Thanks for the thoroughness, as always.

Ghostkeeper (approved):

Ghostkeeper:
I'm now getting an error when loading up layer view in compatibility mode. I'm guessing you didn't get this error since you probably ran your code to test the draw range in compatibility mode, so I'll see if I can quickly find what the issue is.

```
QOpenGLShader::compile(Fragment): 0(23) : error C7011: implicit cast from "int" to "float"
0(23) : error C7011: implicit cast from "int" to "float"

*** Problematic Fragment shader source code ***
#ifdef GL_KHR_blend_equation_advanced
#extension GL_ARB_fragment_coord_conventions : enable
#extension GL_KHR_blend_equation_advanced : enable
#endif
#define lowp
#define mediump
#define highp
#line 1

#ifdef GL_ES
#ifdef GL_FRAGMENT_PRECISION_HIGH
precision highp float;
#else
precision mediump float;
#endif // GL_FRAGMENT_PRECISION_HIGH
#endif // GL_ES
varying lowp vec4 v_color;
varying float v_line_type;
varying highp float v_vertex_index;

uniform int u_show_travel_moves;
uniform int u_show_helpers;
uniform int u_show_skin;
uniform int u_show_infill;

uniform highp vec2 u_drawRange;

void main()
{
if (u_drawRange.x >= 0 && u_drawRange.y >= 0 && (v_vertex_index < u_drawRange.x || v_vertex_index > u_drawRange.y))
{
discard;
}
if ((u_show_travel_moves == 0) && (v_line_type >= 7.5) && (v_line_type <= 9.5))
{  // actually, 8 and 9
// discard movements
discard;
}
// support: 4, 5, 7, 10, 11
if ((u_show_helpers == 0) && (
((v_line_type >= 3.5) && (v_line_type <= 4.5)) ||
((v_line_type >= 6.5) && (v_line_type <= 7.5)) ||
((v_line_type >= 9.5) && (v_line_type <= 10.5)) ||
((v_line_type >= 4.5) && (v_line_type <= 5.5)) ||
((v_line_type >= 10.5) && (v_line_type <= 11.5))
))
{
discard;
}

// skin: 1, 2, 3
if ((u_show_skin == 0) && (
(v_line_type >= 0.5) && (v_line_type <= 3.5)
)) {
discard;
}
// infill:
if ((u_show_infill == 0) && (v_line_type >= 5.5) && (v_line_type <= 6.5))
{
// discard movements
discard;
}

gl_FragColor = v_color;
}
***
2021-11-15 14:49:25,861 - ERROR - [MainThread] UM.View.GL.ShaderProgram.setFragmentShader [147]: Fragment shader failed to compile: 0(23) : error C7011: implicit cast from "int" to "float"
0(23) : error C7011: implicit cast from "int" to "float"

2021-11-15 14:49:25,861 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: Exception: Could not set fragment shader from '/home/trin/Gedeeld/Projects/Cura/plugins/SimulationView/layers_shadow.shader'.
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: Traceback (most recent call last):
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/View/GL/OpenGL.py", line 197, in createShaderProgram
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:     shader.load(file_name, version = version_string)
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/View/GL/ShaderProgram.py", line 82, in load
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:     raise InvalidShaderProgramError("{0} is missing a shader [{1}, {2}]".format(file_name, vertex_key, fragment_key))
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: UM.View.GL.ShaderProgram.InvalidShaderProgramError: /home/trin/Gedeeld/Projects/Cura/plugins/SimulationView/layers_shadow.shader is missing a shader [vertex41core, fragment41core]
2021-11-15 14:49:25,862 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: 
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: During handling of the above exception, another exception occurred:
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: 
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: Traceback (most recent call last):
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/View/GL/OpenGL.py", line 203, in createShaderProgram
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:     shader.load(file_name, version = "")
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/View/GL/ShaderProgram.py", line 99, in load
2021-11-15 14:49:25,863 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]:     raise InvalidShaderProgramError(f"Could not set fragment shader from '{file_name}'.")
2021-11-15 14:49:25,864 - ERROR - [MainThread] UM.View.GL.OpenGL.createShaderProgram [205]: UM.View.GL.ShaderProgram.InvalidShaderProgramError: Could not set fragment shader from '/home/trin/Gedeeld/Projects/Cura/plugins/SimulationView/layers_shadow.shader'.
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [66]: An uncaught error has occurred!
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]: Traceback (most recent call last):
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/Qt/Bindings/MainWindow.py", line 255, in _render
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     renderer.render()
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "/home/trin/Gedeeld/Projects/Uranium/UM/Qt/QtRenderer.py", line 164, in render
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     render_pass.render()
2021-11-15 14:49:25,864 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:   File "/home/trin/Gedeeld/Projects/Cura/plugins/SimulationView/SimulationPass.py", line 65, in render
2021-11-15 14:49:25,865 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]:     self._layer_shader.setUniformValue("u_active_extruder", float(max(0, self._extruder_manager.activeExtruderIndex)))
2021-11-15 14:49:25,865 - CRITICAL - [MainThread] cura.CrashHandler.__init__ [69]: AttributeError: 'NoneType' object has no attribute 'setUniformValue'
```

rburema:
Thanks for  the fix & merge!

I shouldn't have ignored the feeling I made it a float for a reason then, it seems.

> Not sure if varying was the right choice, but I don't want to change that.

Same. I'm never 100% sure what legacy OpenGL (pre 3.x) does and doesn't accept (also since it's mostly untestable, since everyone has their own implementation that might or might not accept newer things seemingly arbitrarily), so I'm hesitant to change from `varying` as well. I know that works at least.

rburema:
I think I may have a fix for it: https://github.com/Ultimaker/CuraEngine/pull/1530

We should decide if this is worthy of a patch or not this week.

smartavionics:
Please see the comments I left at https://github.com/Ultimaker/Cura/commit/5e60cc6208c16d966505e7a97d9bbbb0644d2716

PR #10265, opened 2021-08-10
Add new machines: TeraForm Inventor, TeraForm Polaris

dimalott:
Added definition, extruder, mesh, and quality resource files for 2 new machines:

TeraForm Inventor: large scale 3D printer, operating in linear mode
TeraForm Polaris : large scale 3D printer, operating in polar mode

Files have been tested locally on Cura version 4.8-4.10.

About us:

AI SpaceFactory is a NASA-award winning company focused on research and development of large-scale additive manufacturing and construction technologies for Earth and Space applications. Our first commercial robotic 3D printer, the TeraForm, can 3D print objects up to 7.5m diameter. TeraForm is compatible with g-code generated with Cura, eliminating the need for any complex robotic programming. We look forward to working with the Ultimaker/Cura community to make large scale 3D printing accessible to everyone.


resources/definitions/teraform_polaris.def.json
```diff
@@ -0,0 +1,150 @@
...
+            "default_value": 60
+        },
+        "retract_at_layer_change":
+        {
+            "default_value": true
+        },
+        "retraction_combing": 
+        {
+            "default_value": "off",
+            "resolve": "off"
```
nallath, line 139: You only have a single extruder, so there is no need to change the resolve function.

resources/definitions/teraform_polaris.def.json
```diff
@@ -0,0 +1,150 @@
...
+        },
+        "zig_zaggify_infill":
+        {
+            "value": true,
+            "settable_per_mesh": true
+        },
+        "infill_wipe_dist":
+        {
+            "enabled": false,
+            "settable_per_mesh": false
```
nallath, line 122: Why make this no longer settable per mesh?

resources/definitions/teraform_polaris.def.json
```diff
@@ -0,0 +1,150 @@
...
+            {
+                "lines": "Lines",
+                "gyroid": "Gyroid"
+            },
+            "default_value": "lines"
+        },
+        "zig_zaggify_infill":
+        {
+            "value": true,
+            "settable_per_mesh": true
```
nallath, line 117: This won't work. It's not settable per mesh because the engine won't handle that correctly. Setting it to true here won't change that.

resources/definitions/teraform_inventor.def.json
```diff
@@ -0,0 +1,149 @@
+  {
+    "version": 2,
+    "name": "TeraForm Inventor",
+    "inherits": "fdmprinter",
+    "metadata":
+    {
+        "visible": true,
+        "author": "SpaceFactory",
```
nallath, line 8: I can't find anything about this printer. Is this a printer that is commercially available? If not, there isn't really a reason to add it to the distribution.

resources/definitions/teraform_inventor.def.json
```diff
@@ -0,0 +1,149 @@
...
+            "value": true
+        },
+        "machine_nozzle_size": { "default_value": 12 },
+        "machine_use_extruder_offset_to_offset_coords": { "default_value": false },
+        "layer_height":
+        {
+            "minimum_value": "1",
+            "minimum_value_warning":"2",
+            "maximum_value_warning":"10",
+            "settable_per_extruder": false
```
nallath, line 45: No need to override this, it's already false in the fdmprinter.def.json

resources/definitions/teraform_inventor.def.json
```diff
@@ -0,0 +1,149 @@
...
+            "settable_per_mesh": false
+        },
+        "fill_outline_gaps":
+        {
+            "default_value": false,
+            "settable_per_mesh": false
+        },
+        "ironing_enabled": 
+        {
+            "default_value": false,
```
nallath, line 97: That is already the default value. No need to set it again.

resources/extruders/teraform_x_120.def.json
```diff
@@ -0,0 +1,15 @@
...
+    "inherits": "fdmextruder",
+    "metadata": {
+        "machine": "teraform_inventor",
+        "position": "0"
+    },
+
+    "overrides": {
+        "extruder_nr": { "default_value": 0 },
+        "machine_nozzle_size": { "default_value": 12 },
+        "material_diameter": { "default_value": 12 }
```
nallath, line 13: 12 mm diameter? Uh. Since the machine defines that is has materials, I guess you should also add material profiles in fdm_materials with a diameter of 12 mm?

nallath (changes requested):
* The provided 3D models are far to large. We use a max of 1 MB, but we ask people keep them a lot smaller
* A large number of settings have unneeded overrides. I've not put comments on all of them, but i suggest that all of them are re-checked.

resources/definitions/teraform_inventor.def.json
rburema: I think it might be? Their website ( https://www.aispacefactory.com/about ) mentions how they 'develop advanced construction technologies', so that seems to imply that they won't just keep the hardware in-house.

dimalott: Yes - printer will be commercially available next month.

resources/definitions/teraform_polaris.def.json
dimalott: Our printer runs a post-processor which works much better when wipes are disabled. When 'settable_per_mesh' is false, does this mean the user does not have the option to enable?

resources/definitions/teraform_polaris.def.json
dimalott: Thanks for reviewing @nallath. Working on the fix.

resources/definitions/teraform_polaris.def.json
dimalott: Noted.

resources/definitions/teraform_inventor.def.json
dimalott: Ok.

resources/extruders/teraform_x_120.def.json
dimalott: Thank you - was wondering where to add the material... We have submitted a pull request to fdm_materials

resources/definitions/teraform_inventor.def.json
dimalott: Fixed

resources/definitions/teraform_inventor.def.json
dimalott: Fixed

resources/definitions/teraform_polaris.def.json
dimalott: Fixed

resources/definitions/teraform_polaris.def.json
dimalott: Hi @nallath, mesh file sizes have been reduced to 1 meg in the latest commit

resources/definitions/teraform_polaris.def.json
nallath: If you simply want to disable the setting at all, you should just set "enabled" to false. That prevents the user from changing it. 

The settable_per_mesh property defines if it can be changed via the per objects setting tool

dimalott:
Hi @nallath the autochecks failed - are you able to spot the problem on your side? Any more changes we can do to the files? 

Best, David @ SpaceFactory

Ghostkeeper:
You should be able to see the test failures, as long as you're logged into Github anyway. Otherwise you can also run the tests locally using `cmake && ctest --verbose`.

It's currently complaining about two things:
* `teraform_polaris_platform.stl` is still too big, 1024.79KB, more than 1024KB. Converting that to 3MF would probably save a lot of space without compromising quality.
* You're overriding some `default_value` properties of settings while the `value` overrides that in a parent definition container. That makes the setting overrides ineffective. `Unnecessary default_values for ['infill_pattern', 'machine_firmware_retract', 'relative_extrusion', 'wall_0_wipe_dist', 'wall_thickness'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/teraform_inventor.def.json` and `Unnecessary default_values for ['infill_pattern', 'machine_firmware_retract', 'relative_extrusion', 'wall_0_wipe_dist', 'wall_thickness'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/teraform_polaris.def.json`. Changing those setting overrides from `default_value` to `value` would fix the issue and cause the correct setting values to be used in your profiles.

resources/definitions/teraform_inventor.def.json
Ghostkeeper: We don't normally publish profiles for printers that have not yet been published, not even if the contributor promises out of the good of their hearts that they will soon publish the printer.

jellespijker:

Joeydelarago:
Hey @dimalott, I'm going to put this PR into draft until the issues the bot has flagged are fixed. Hit the "Ready for review" button after the issues are fixed and we will get back to you as soon as possible.

Thanks!

GregValiant:
It is July of 2026.  Can this be closed or is there still interest.  Many of the UM people who commented (above) have moved on.

Issue #7842, opened 2020-05-29
Add Octoprint @pause to Pause At Height script


agibson2:
**Describe the solution you'd like**
Octoprint has a feature where if it sees an @pause (lower case p) line then it sends it's own commands for pause and resume and also handles the actually resuming using the octoprint server's web GUI.  The current script rewrite has a repetier mode that sends @pause but it also sends other commands too.  Octoprint just needs the @pause (lower case p) put into the gcode and nothing else.

Any chance to get a new mode for pause at height to just put an @pause (lower case p) in the gcode and nothing else for Octoprint?

I maintain another script to do that but it would be great to have this integrated so that other users can take advantage of it without using a third party script.

I maintain a separate script here to do it...
https://community.ultimaker.com/topic/24105-post-processing-script-for-octoprints-pause-feature/

mahtDFR:
One for @fieldOfView I think

Ghostkeeper:
Doesn't the Repetier version work?

fieldOfView:
The repetier version also does funky head parking, as an inheritance from the original pause at height for repetier. The request for OctoPrint is to just have a single `@pause`. I am in dubio whether to make an OctoPrint flavor, or to make the parking optional for all flavors.

PS: @agibson2 the secret to entering `@pause` without it turning into a mention of the user named Pause is to surround it by backticks: `.

Ghostkeeper:
If making parking an option for all flavours is a bit of extra work, we can also do that later and then merge the OctoPrint pause with the Repetier pause. It wouldn't be very difficult to do.

agibson2:
Any chance of this getting implemented?  It would be really nice to have an integrated solution for Octoprint.

nallath:
Well, this isn't the kind of feature that will get attention from the main (eg; Paid by Ultimaker) developers. So the best chance is that someone makes a pull request. We would be more than happy to include it in that case.

fieldOfView:
Here's a PR: https://github.com/Ultimaker/Cura/pull/10288.
Disabling the head parking will make the repetier method suitable for OctoPrint. I have therefor also renamed the repetier method to repetier/octoprint.

I have not tested this PR extensively. Hopefully some of the posters in this thread can do that.

agibson2:
I would like to test it. Which files are changed? I would like to copy them to my C:\Program Files... to test it.  (I posted this to the PR on accident and deleted the comment there).

fieldOfView:
[...]

PR #10288, opened 2021-08-16
Make parking optional in all "methods" of Pause at Height

fieldOfView:
This PR makes parking the head optional (for all methods, except Griffin which handles the parking in the firmware so it is never added to the gcode). Not adding parking gcode makes the repetier `@pause` method suitable for use with OctoPrint (which similarly handles the parking in most cases). Not adding parking gcode can also be beneficial in modern Marlin flavors, since those too can handle the parking by themselves.

Contributes to #7842


nallath:
For our reference: CURA-8487

nallath (approved):

Ghostkeeper:
This was also tested by @agibson2 by drop-in replacing the script in Cura 4.10.

Issue #10223, opened 2021-08-03
Add material used to Gcode for e.g. Filament Manager Plugins in OctoPrint


NilsRo:
### Is your feature request related to a problem?

My filament runs out.  :-)

### Describe the solution you'd like

Add material used (including all important information like color,... ) to GCode. Actually it is missing in the Start-/Endcode variables: http://files.fieldofview.com/cura/Replacement_Patterns.html

material GUID is available but useless as it is the same GUID for custom materials.

### Describe alternatives you've considered

Material has to be selected manually in the Plugin before every print (if OctoPrint is used)

### Affected users and/or printers

All printers if OctoPrint is used, also helpful to check for which material the GCode files is created.

fieldOfView:
> material GUID is available but useless as it is the same GUID for custom materials.

The material GUID is only the same if you keep duplicating materials to create a new one. If you üse the "Create" button instead of "Duplicate", you get a new GUID for each newly created material. For materials that you "Duplicate"d before, there's an "Unlink" button somewhat hidden below the material properties:
![unlink material](https://user-images.githubusercontent.com/2448634/78973760-d2875880-7b10-11ea-8311-69a34cef23d2.png)

NilsRo:
I like to get it linked as for example I created 4 new materials out of PETG like to adapt some properties only. Or Made a brand PETG material and like to duplicate it for every color I own. What you describe is a workaround as the management of the materials is higher as it does not share basic properties (as I understood the behaviour in the correct way).

So If I duplicate green PETG, yellow PETG, ... from generic PETG I like to change the print temperature for generic PETG only.

Ghostkeeper:
The GUID is meant to indicate a type of spool. For instance, all PLA Red spools by Ultimaker have the same GUID, but the PLA Green spools by Ultimaker have a different GUID. The "shared properties" are also the ones pertaining to types of spools, e.g. filament cost and weight. You can still have multiple profiles for the same spool, e.g. to print PLA Red with different temperatures or fan speeds.

It does not indicate which individual spool is being used. Cura has no information on that. It would be rather cumbersome to have to tell Cura that the spool got changed, something which any material manager would already know anyway.

[...]

PR #10289, opened 2021-08-16
Add {material_type} and {material_name} as replacement patterns

fieldOfView:
This PR adds two new replacement patterns that users can add to their start gcode snippets: `{material_type}` and `{material_name}`. They map directly to metadata of the XmlMaterialProfile objects.

Contributes to https://github.com/Ultimaker/Cura/issues/10223 and https://github.com/fieldOfView/Cura-OctoPrintPlugin/issues/244

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -411,8 +411,10 @@ class StartSliceJob(Job):
         extruder_nr = stack.getProperty("extruder_nr", "value")
         settings = self._all_extruders_settings[str(extruder_nr)].copy()
 
-        # Also send the material GUID. This is a setting in fdmprinter, but we have no interface for it.
+        # Also send the material GUID, type and name. These are settings in fdmprinter, but we have no interface for it.
         settings["material_guid"] = stack.material.getMetaDataEntry("GUID", "")
+        settings["material_type"] = stack.material.getMetaDataEntry("material", "")
+        settings["material_name"] = stack.material.getMetaDataEntry("name", "")
 
         # Replace the setting tokens in start and end g-code.
         extruder_nr = stack.getProperty("extruder_nr", "value")
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -353,10 +353,19 @@ class StartSliceJob(Job):
             result[key] = stack.getProperty(key, "value")
             Job.yieldThread()
 
-        result["print_bed_temperature"] = result["material_bed_temperature"] # Renamed settings.
+        # Material identification in addition to non-human-readable GUID
+        result["material_id"] = stack.material.getMetaDataEntry("base_file", "")
+        result["material_type"] = stack.material.getMetaDataEntry("material", "")
+        result["material_name"] = stack.material.getMetaDataEntry("name", "")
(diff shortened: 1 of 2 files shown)


pedrolamas:
I noticed a new beta version of Cura was just released, and this PR seems to be such a simple change with such great value that I hope that it gets merged in one of the next releases! 🙂

fieldOfView:
It will not be part of Cura 4.11, but perhaps the next version.

pedrolamas:
I wanted to try this, so I went ahead and made these changes manually in my local Cura 4.10 install, but I couldn't make it work (using an Ender-3 printer profile with generic PETG)...

After changing the files, I added the following to both the printer Start G-code and the Extruder Start G-code:

```text
SET_MATERIAL TYPE="{material_type}" NAME="{material_name}" GUID="{material_guid}"
```

This only showed the GUID on the output gcode, but the material type and name were empty!

I tried searching for "material_guid" on the code and found a reference [here](https://github.com/Ultimaker/Cura/blob/44913c0d8798e6f4f566af2737b09120854db7b1/plugins/XmlMaterialProfile/XmlMaterialProfile.py#L1139), so I tried adding these lines just above [this one](https://github.com/Ultimaker/Cura/blob/44913c0d8798e6f4f566af2737b09120854db7b1/plugins/XmlMaterialProfile/XmlMaterialProfile.py#L535) and now it works:

```
common_setting_values["material_name"] = meta_data["name"]
common_setting_values["material_type"] = meta_data["material"]
```

This is probably not the best way to fix the issue, but I thought best to point it out anyway before this gets merged!

fieldOfView:
It wouldn't get merged if it doesn't work. Thanks for checking (apparently I did not check it enough), I'll have another look.

fieldOfView:
Hehe, I tested the wrong thing. I tested if the variables were sent to CuraEngine properly (which does nothing with the values), instead of looking if the replacement patterns in the gcode were actually working. I'll rework this PR.

pedrolamas:
Well, I for one am grateful for what you already did here as it pointed me in the right direction to get it somewhat working locally for me!

Now I can put the send the material type directly to Klipper and it will set the correct Pressure Advance!

fieldOfView:
You'll be able to use `{material_name,0}`, `{material_type,1}` and even `{material_guid,initial_extruder_nr}` or `{material_name,support_extruder_nr}`. Extruder count starts at 0.

fieldOfView:
After some deliberation I decided the material name and type don't need to be settings that are sent to CuraEngine (because CuraEngine will not use them anyway). This simplifies the implementation, with the disadvantage that the patterns are more "hardcoded" into the replacement pattern code. The alternative would be to hardcode them into the XMLMaterialProfile class and having them as non-user-settable settings. I chose for code-simplicity (but let me know if the setting route is preferred).

I have also added `{material_id}` and `{material_brand}`. The latter is usefull to do `SET_MATERIAL NAME="{material_brand} {material_name}"` (ie: get the full name of the material including the brand). The former actually maps to the base_file metadata, so it establishes a link to the fdmmaterial.xml file, instead of the pure id which also contains the machine_id and variant_id.

pedrolamas:
Thanks @fieldOfView, I removed my hacked changes and replaced them with your new changes, and they worked perfectly!!

fieldOfView:
Thanks for testing!

nallath (approved):

PR #10294, opened 2021-08-17
Add XYZprinting Printer Definitions

heed818:
Hi,
I just add Definitions for XYZprinting Printers,
Please help to check,thank you.


resources/quality/xyzprinting/abs/xyzprinting_da_vinci_jr_pro_xplus_copper_0.40_abs_fine.inst.cfg
```diff
@@ -0,0 +1,32 @@
...
+quality_type = fine
+material = generic_abs
+variant = Copper 0.4mm Nozzle
+
+[values]
+cool_fan_speed = 100
+cool_fan_speed_0 = 0
+material_bed_temperature = 70
+material_bed_temperature_layer_0 = 70
+material_print_temperature = 240
```
nallath, line 18: It's generally better to not set the material_print temperature in the quality profile. If you do that, changing the setting in the material profile won't work.

It's better to not set it all and use something like this for the other ones :
material_final_print_temperature = =material_print_temperature (and if that is always the case, it's also better to set that in the machine definition and not in a specific profile)

resources/quality/xyzprinting/anti_bact/xyzprinting_da_vinci_jr_pro_xeplus_copper_0.40_antibact_coarse.inst.cfg
```diff
@@ -0,0 +1,29 @@
+[general]
+version = 4
+name = Coarse Quality
+definition = xyzprinting_da_vinci_jr_pro_xeplus
+
+[metadata]
+setting_version = 17
+type = quality
+quality_type = coarse
+material = xyzprinting_antibact_pla
```
nallath, line 10: You didn't add that material, so this will never become active.

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xeplus_global_0.40_coarse.inst.cfg
```diff
@@ -0,0 +1,27 @@
...
+setting_version = 17
+type = quality
+quality_type = coarse
+weight = -3
+global_quality = True
+
+[values]
+adhesion_type = brim
+brim_width = 10.0
+cool_fan_speed = 100
```
nallath, line 16: It's really strange to set this in the global quality. In most cases the cool_fan_speed is very much material dependant!

Also, if it's always 100, why not just set that in the definition. It keeps you from repeating yourself over and over again (DRY principle: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xplus_global_0.20_normal.inst.cfg
```diff
@@ -0,0 +1,27 @@
...
+[values]
+adhesion_type = brim
+brim_width = 10.0
+cool_fan_speed = 100
+cool_fan_speed_0 = 0
+layer_height = 0.2
+layer_height_0 = 0.35
+material_bed_temperature = 40
+material_diameter = 1.75
+material_print_temperature = 210
```
nallath, line 22: Setting the material temperature in a global quality profile? That doesn't really make sense. This setting will _never_ be used, since material profiles always override it.

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xeplus_global_0.40_coarse.inst.cfg
```diff
@@ -0,0 +1,27 @@
+[general]
+version = 4
+name = Coarse Quality
+definition = xyzprinting_da_vinci_jr_pro_xeplus
```
nallath, line 4: This profile is exactly the same as the one for the other printer. Why duplicate it? You can just create a single profile and use the definition = xyzprinting_base

nallath (changes requested):

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xplus_global_0.20_normal.inst.cfg
heed818: Because in quality, we add files by different printers, and same material with different printers have different temperature.
That we set temperature in quality.

resources/quality/xyzprinting/abs/xyzprinting_da_vinci_jr_pro_xplus_copper_0.40_abs_fine.inst.cfg
heed818: Thanks for your advisement,it is more convenient :)

resources/quality/xyzprinting/anti_bact/xyzprinting_da_vinci_jr_pro_xeplus_copper_0.40_antibact_coarse.inst.cfg
heed818: image

But i have added them in material folder.
Should i add other setting for them??

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xeplus_global_0.40_coarse.inst.cfg
heed818: OK,We will move them to XYZprinting_base.
Thank you :)

resources/quality/xyzprinting/anti_bact/xyzprinting_da_vinci_jr_pro_xeplus_copper_0.40_antibact_coarse.inst.cfg
heed818: OK,the pull request of the fdm_materials is on-going,thank you :)

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xeplus_global_0.40_coarse.inst.cfg
heed818: hi, the file is a base for the printer's quality

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xplus_global_0.20_normal.inst.cfg
Ghostkeeper: In the material profile you can set the temperature differently for different printers or even for different nozzles for those printers. For an example, see here: https://github.com/Ultimaker/fdm_materials/blob/7a3e7e81c09c968372e79872d95b9e5cf1ccd7b0/generic_pla.xml.fdm_material#L105

This temperature is indeed overridden by the material profile at the moment, because the printing temperature is always evaluated for a certain extruder. For more information, refer to the diagram at https://github.com/Ultimaker/Cura/blob/master/docs/profiles/container_stacks.md#a-machine-instance . Settings evaluated for extruders start at the top of the extruder stack. This setting override is in the `quality` header of the global stack, which is underneath the material profiles of the exturder stacks.

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xeplus_global_0.40_coarse.inst.cfg
Ghostkeeper: No, it is a global quality profile. It's not used by other quality profiles, just matched by the `quality_type`. Adding it to the parent printer `xyzprinting_base` allows it to be selected just as well.

resources/quality/xyzprinting/abs/xyzprinting_da_vinci_jr_pro_xplus_copper_0.40_abs_fine.inst.cfg
Ghostkeeper: Maybe for you, but for users it breaks their material profiles.

resources/quality/xyzprinting/xyzprinting_da_vinci_jr_pro_xplus_global_0.20_normal.inst.cfg
heed818: hi @Ghostkeeper  ,thank you
i have add the setting in fdm_materials(XYZprinting) help everything is going right,thanks.

heed818:
Hi @nallath 
Will this commit been merged to 4.12?
thanks.

heed818:
> Trying these, I'm getting a bunch of errors and improperly evaluated functions for the two bed temperature settings (Build Plate Temperature and Build Plate Temperature Initial Layer). The error is this:
> 
> ```
> An exception occurred in inherit function =max(extruderValues('material_bed_temperature')): maximum recursion depth exceeded
> ```
> 
> The source of this error is this setting override in 12 of the files:
> 
> material_bed_temperature = =material_bed_temperature
> 
> This causes it to set the build plate temperature to the build plate temperature, causing an infinite loop. This then causes Cura to eventually put it at 0 as fallback (after wasting a dozen times 1000 deep recursion nesting).
> 
> Could you please fix this to have a proper build plate temperature?
> 
> Did you test these profiles before submitting them to us?

Sorry about this wrong setting.
And we have try to place the files in cura 4.11 program file, and it can feel the respond delay while executing and select the related printers. 
i just update the file to remove them.
Once again, I apologize for my carelessness.
Thank you.

Ghostkeeper (approved):

Ghostkeeper:
I'm encountering something else with these profiles. It's rather subtle and I didn't see it at first. But all of the material-specific qualities are marked as global qualities.

Here's a visualisation to help see why that's incorrect: https://github.com/Ultimaker/Cura/blob/master/docs/profiles/container_stacks.md

There's a quality profile in the "global" section, and another in the extruder-specific section where the material is also located. The global quality can override global settings (like layer height) while the extruder-specific one can't. Cura will match one of the global qualities with the printer to put there, but that can't depend on the material or nozzle, because the user might have different materials or nozzles in their multiple extruders.

In this case, there are multiple global qualities that match and Cura selects an arbitrary one. For me, a PETG quality profile got selected as global quality profile underneath my ABS profile.

I can fix this myself though, as it's clear what you wanted to do. And we've had this waiting way too long already (our fault) so I'll do my best to resolve the issue.

PR #10315, opened 2021-08-20
add anycubic vyper

ThatGuyZim:
Thanks for the heads up nallath, files are as follows


anycubic_vyper_platform.stl
```diff
@@ -0,0 +1,4286 @@
+solid CATIA STL
```
nallath, line 1: Could you add it as a binary STL (or better; a 3MF)? That helps keeping the size down quite a bit.

nallath (changes requested):

anycubic_vyper_platform.stl
ThatGuyZim: Uploaded anycubic_vyper_platform.3mf for you

ThatGuyZim: can't use .3mf files for some reason, bed positioned wrong in the print area

nallath:
Oh, one more: Is there a specific reason that the .def.json is called "anycubic_vyper_" (and not "anycubic_vyper")?

ThatGuyZim:
I had a few versions of the file before this, you can remove the second underscore

anycubic_vyper_.def.json
```diff
@@ -0,0 +1,47 @@
+{
+  "version": 2,
+  "name": "Anycubic Vyper",
+  "inherits": "fdmprinter",
+  "metadata": {
+    "visible": true,
+    "author": "ThatGuyZim",
+    "manufacturer": "Anycubic",
+    "file_formats": "text/x-gcode",
+    "platform": "anycubic_vyper_platform.3mf",
```
nallath, line 10: This still references the 3mf.

ThatGuyZim: line changed to reference stl instead

nallath (approved):

nallath:
The files are still in the wrong location, but I will fix that for you. For the next time, they should be located in the resources folder

nallath:
It's merged in: 1740f8e88075c96388206f8b3258228a444f09d4

PR #10318, opened 2021-08-20
pull ZONESTAR printer resource again (definitions and extruders)

ZONESTAR3D:
Hi guys,
I tried to pull the ZONESTAR 3d printer resource again, please help to check. thanks!


ZONESTAR3D:
delete P802QR2 first, please check again

resources/definitions/zonestar_Z10.def.json
```diff
@@ -0,0 +1,47 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "ZONESTAR Z10" },
+        "layer_height": { "default_value": 0.2 },
+        "machine_extruder_count": { "default_value": 8 },
+        "machine_heated_bed": { "default_value": true },
+        "machine_center_is_zero": { "default_value": false },
+        "machine_height": { "default_value": 400 },
+        "machine_gcode_flavor": { "default_value": "RepRap (Marlin/Sprinter)" },
+        "machine_depth": { "default_value": 310 },
+        "machine_width": { "default_value": 310 },
```
Ghostkeeper, line 33: [This sales page](https://www.aliexpress.com/i/32906728737.html) claims a size of 300x300x400mm. You have written 310x310x400mm here. Is that correct? It seems you've made a mistake here, or is that sales page wrong?

resources/definitions/zonestar_Z9.def.json
```diff
@@ -0,0 +1,48 @@
...
+    "overrides": {
+        "machine_name": { "default_value": "ZONESTAR Z9" },
+        "layer_height": { "default_value": 0.2 },
+        "machine_extruder_count": { "default_value": 8 },
+        "machine_heated_bed": { "default_value": true },
+        "machine_center_is_zero": { "default_value": false },
+        "machine_height": { "default_value": 400 },
+        "machine_gcode_flavor": { "default_value": "RepRap (Marlin/Sprinter)" },
+        "machine_depth": { "default_value": 310 },
+        "machine_width": { "default_value": 310 },
```
Ghostkeeper, line 33: According to [this sales page](https://www.electroca.com/product/zonestr-z9-large-size-aluminum-frame-dual-extruder-mixing-color/) the printer's build volume is 300x300x400mm. Did you make a mistake here, or is the sales page incorrect?

resources/definitions/zonestar_Z9.def.json
```diff
@@ -0,0 +1,48 @@
...
+        "machine_extruder_trains":
+        {
+            "0": "zonestar_mixvtool_0",
+            "1": "zonestar_mixvtool_1",
+            "2": "zonestar_mixvtool_2",
+            "3": "zonestar_mixvtool_3",
+            "4": "zonestar_mixvtool_4",
+            "5": "zonestar_mixvtool_5",
+            "6": "zonestar_mixvtool_6",
+            "7": "zonestar_mixvtool_7"
```
Ghostkeeper, line 20: According to [this review](https://zeroair.org/2019/01/09/zonestar-z8xm2-multi-material-3d-printer-kit-review/), the printer has a maximum of 2 print heads. Is there a printer with 8 print heads, or did you make a mistake here?

resources/definitions/zonestar_z5.def.json
```diff
@@ -0,0 +1,47 @@
...
+        "machine_extruder_trains":
+        {
+            "0": "zonestar_mixvtool_0",
+            "1": "zonestar_mixvtool_1",
+            "2": "zonestar_mixvtool_2",
+            "3": "zonestar_mixvtool_3",
+            "4": "zonestar_mixvtool_4",
+            "5": "zonestar_mixvtool_5",
+            "6": "zonestar_mixvtool_6",
+            "7": "zonestar_mixvtool_7"
```
Ghostkeeper, line 20: According to [this sales page](https://www.priceboon.com/product/zonestar-z5m/), this printer is a 2-in-1-out style mixing printer, so it should only have 2 extruders?

resources/definitions/zonestar_z5x.def.json
```diff
@@ -0,0 +1,40 @@
...
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "ZONESTAR",
+        "manufacturer": "ZONESTAR",
+        "file_formats": "text/x-gcode",
+        "platform_offset": [ 0, 0, 0],        
+        "machine_extruder_trains":
+        {
+            "0": "zonestar_Z5X_extruder"
```
Ghostkeeper, line 13: According to [this sales page](https://www.madethebest.com/products/zonestar-z5x-series-large-high-precision-resolution-ultra-silent-easy-install-optional-dual-extrusion-auto-mix-color-3d-printer) the printer has two extruders. Are you missing one?

resources/definitions/zonestar_Z10.def.json
ZONESTAR3D: yes, please keep 310x310x400mm, we set a little small than the real size on sales page

resources/definitions/zonestar_Z9.def.json
ZONESTAR3D: yes, please keep 310x310x400mm, we set a little small than the real size on sales page

resources/definitions/zonestar_Z10.def.json
```diff
@@ -0,0 +1,47 @@
...
+        "machine_extruder_trains":
+        {
+            "0": "zonestar_mixvtool_0",
+            "1": "zonestar_mixvtool_1",
+            "2": "zonestar_mixvtool_2",
+            "3": "zonestar_mixvtool_3",
+            "4": "zonestar_mixvtool_4",
+            "5": "zonestar_mixvtool_5",
+            "6": "zonestar_mixvtool_6",
+            "7": "zonestar_mixvtool_7"
```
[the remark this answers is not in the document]
ZONESTAR3D: we choose to use VTOOLs to replace of the "real extruders", the customer will be easy  to do mixing color print with the Virtual Extruder. Please keep these settings if possible, thanks.

resources/definitions/zonestar_Z9.def.json
ZONESTAR3D: we choose to use VTOOLs to replace of the "real extruders", the customer will be easy  to do mixing color print with the Virtual Extruder. Please keep these settings if possible, thanks.

resources/definitions/zonestar_z5.def.json
ZONESTAR3D: we choose to use VTOOLs to replace of the "real extruders", the customer will be easy  to do mixing color print with the Virtual Extruder. Please keep these settings if possible, thanks.

resources/definitions/zonestar_z5x.def.json
ZONESTAR3D: Z5X has only one extruder, please keep these settings, thanks!

resources/definitions/zonestar_Z10.def.json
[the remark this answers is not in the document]
Ghostkeeper: What do you mean with "VTOOLs"? That is not a term in Cura's architecture.

If the user wants to change the colour of their materials, they are expected to change the material profile, rather than the extruder an object is printed with. This works better with Cura's workflow and profile system. Right now, a custom profile that the user creates would save settings for each extruder separately and won't be available if any of the extruders has a different material type in.

resources/definitions/zonestar_Z9.def.json
Ghostkeeper: What do you mean with "VTOOLs"? That is not a term in Cura's architecture.

If the user wants to change the colour of their materials, they are expected to change the material profile, rather than the extruder an object is printed with. This works better with Cura's workflow and profile system. Right now, a custom profile that the user creates would save settings for each extruder separately and won't be available if any of the extruders has a different material type in.

resources/definitions/zonestar_z5.def.json
Ghostkeeper: What do you mean with "VTOOLs"? That is not a term in Cura's architecture.

If the user wants to change the colour of their materials, they are expected to change the material profile, rather than the extruder an object is printed with. This works better with Cura's workflow and profile system. Right now, a custom profile that the user creates would save settings for each extruder separately and won't be available if any of the extruders has a different material type in.

resources/definitions/zonestar_z5x.def.json
Ghostkeeper: I beg to differ. It says it right there on the page with great pride:

> 2-IN-2-OUT(Z5XR2): The printer has 2 extrusion feeders and 2 nozzle. It can use one nozzle to print the main object and another nozzle to print support filament, for exmple, the main extruder to print PLA and another nozzle to print PVA. Of couse, you can use it to print two color PLA or (ABS) filament, etc.
> 
> 2-IN-1-OUT(Z5XM2): The printer has 2 extrusion feeders and 1 nozzle. The filament will be mixed in the hotend and then flow out from one nozzle, you can use it to mix two colors filament together and become a new color, usually we suggest you use only one type filament in the same time. With ZONESTAR's patented technology, you can easily set the mixed ratio of two filaments on the control panel when printing from SD card, and even let the machine automatically adjust the mixed ratio to become a multi-color object.

If this printer does not have 2 extruders, what "patented technology" is that website referring to that tunes the mixing ratio?

PR #10327, opened 2021-08-23
Store metadata in database

nallath:
Instead of re-reading the metadata from the files on every boot, which is pretty slow on windows (10+ seconds, even on machines with a fast SSD), we store the profiles in a database. By using the modified time of the files, we can check if the database needs updating.

Creating the database is about as fast as what it used to do, but every subsequent run, it knocked my boot time from 10s on windows to 0.6s. It also increases the booting time on Linux, but there it's only 300ms faster.

CURA-6096

Needs to be combined with https://github.com/Ultimaker/Uranium/pull/729

### DIFF
--- a/cura/Settings/CuraContainerRegistry.py
+++ b/cura/Settings/CuraContainerRegistry.py
@@ -46,6 +46,7 @@ class CuraContainerRegistry(ContainerRegistry):
 
         self._add_to_database_handlers["variant"] = self._addVariantToDatabase
         self._add_to_database_handlers["quality"] = self._addQualityToDatabase
+        self._add_to_database_handlers["intent"]  = self._addIntentToDatabase
 
         self._get_from_database_handlers["variant"] = self._getVariantFromDatabase
         self._get_from_database_handlers["quality"] = self._getQualityFromDatabase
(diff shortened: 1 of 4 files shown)


Ghostkeeper (dismissed):
I think these handlers are overly verbose and duplicate too much code. What you really need for these classes is the following information:
* The name of the table (e.g. `qualities`).
* The fields available, in the order in which they appear in the table.
* The data type for each field.

So I'd envision that we basically only need to have two fields being overridden by the subclasses:

```
class QualityDatabaseHandler(DatabaseMetadataContainerController):
    def __init__(self) -> None:
        self.table_name = "qualities"
        self.fields = {
            "name": "text",
            "quality_type": "text",
            "material": "text",
            "variant": "text",
            "global_quality": "bool",
            "definition": "text",
            "version": "text",
            "setting_version": "text"  # Or int?
        }
```

Dictionaries in Python have an order, so this preserves the order in which fields occur in the table.

The methods in the parent class can then be made generic to generate the correct queries in the `__init__`, and do the correct conversions based on that table. If this is not possible (perhaps for `QualityDatabaseHandler_convertMetadataToInsertBatch`) they can still override it. But it should reduce code duplication a lot then, and if there's ever a problem with the conversion in the future then it might just be a simple data change to this dict, which is a much safer change.

Ghostkeeper:
> Previously it was considered to be an implementation detail and you couldn't rely on it.

Previously you would've had to use an `OrderedDict`, which would also have been fine.

jellespijker:
It is always a simple underscore `_` that will cause all that trouble. I remember renaming these variables using PyCharm. Strange that it didn't do it everywhere.
@Ghostkeeper your changes makes sense, since @nallath already reviewed my changes I'm moving it to component testing. Maybe @rburema could do that on his Windows machine.

PR #10328, opened 2021-08-23
Retain the last saved filename after saving a file

konskarm:
Cura was not retaining the last saved filename selected by the user after they were saving their file. This PR fixes that by making sure that the filename selected by the user in the save file dialog will update the name of the workspace (project) all around Cura.

**Note:** The stored filename is forgotten when the user loads another project file, creates a new project (Ctrl+N) or clears the buildplate (Ctrl+D).

Requires https://github.com/Ultimaker/Uranium/pull/728

CURA-8358

### DIFF
--- a/plugins/DigitalLibrary/src/DigitalFactoryController.py
+++ b/plugins/DigitalLibrary/src/DigitalFactoryController.py
@@ -603,8 +603,8 @@ class DigitalFactoryController(QObject):
         self._saveFileToSelectedProjectHelper(filename, formats)
 
     def _saveFileToSelectedProjectHelper(self, filename: str, formats: List[str]) -> None:
-        # Indicate we have started sending a job.
-        self.uploadStarted.emit()
+        # Indicate we have started sending a job (and propagate any user file name changes back to the open project)
+        self.uploadStarted.emit(filename)
 
         library_project_id = self._project_model.items[self._selected_project_idx]["libraryProjectId"]
         library_project_name = self._project_model.items[self._selected_project_idx]["displayName"]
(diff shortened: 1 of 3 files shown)


nallath (approved):

Ghostkeeper (changes requested):

nallath (approved):

nallath (approved):

plugins/DigitalLibrary/src/DigitalFactoryController.py
```diff
@@ -603,8 +603,8 @@ def saveFileToSelectedProject(self, filename: str, formats: List[str]) -> None:
         self._saveFileToSelectedProjectHelper(filename, formats)
 
     def _saveFileToSelectedProjectHelper(self, filename: str, formats: List[str]) -> None:
-        # Indicate we have started sending a job.
-        self.uploadStarted.emit()
+        # Indicate we have started sending a job (and propagate any user file name changes back to the open project)
+        self.uploadStarted.emit(filename if "3mf" in formats else None)
```
Ghostkeeper, line 607: This change causes DigitalLibrary to require a higher API version number, meaning it's no longer backwards compatible with Cura 4.9.

plugins/DigitalLibrary/src/DigitalFactoryOutputDevice.py
```diff
@@ -105,8 +105,10 @@ def _onUserAccessStateChanged(self, logged_in: bool) -> None:
         self.enabled = logged_in and self._controller.userAccountHasLibraryAccess()
         self.enabledChanged.emit()
 
-    def _onWriteStarted(self) -> None:
+    def _onWriteStarted(self, new_name: Optional[str] = None) -> None:
         self._writing = True
+        if new_name:
+            self.setLastOutputName(new_name)  # On saving, the user can change the name, this should propagate.
```
Ghostkeeper, line 111: This change causes DigitalLibrary to require a higher API version number, meaning it's no longer backwards compatible with Cura 4.9.

PR #10331, opened 2021-08-24
CURA-8380: Align link style among os in whats new

konskarm:
The following things are fixed by this PR:

1. The links in the What's new pages were using the OS-style links and they were inconsistent between OS. In addition to the inconsistency, the OS-styled link in Windows was hard to see in the dark theme. This is now fixed by making the links in that text area use a specific style. 
**Note: ** The style is applied by prepending the CSS style, since the text in the What's new pages is retrieved from html files.

link_visibility

2. The background color of premium plugins and materials was making the "Search materials" link very hard to see in the dark theme. This is now changed to a darked shade of grey

premium_background

3. The "No plugins/materials installed" text in the Installed tab in the Marketplace was invisible in the dark theme. This is now changed so that the text color will be consistend with the disabled plugins' text color

label_visibility

CURA-8380

### DIFF
--- a/plugins/Toolbox/resources/qml/components/ToolboxDownloadsShowcase.qml
+++ b/plugins/Toolbox/resources/qml/components/ToolboxDownloadsShowcase.qml
@@ -8,7 +8,7 @@ import UM 1.1 as UM
 
 Rectangle
 {
-    color: UM.Theme.getColor("secondary")
+    color: UM.Theme.getColor("toolbox_premium_packages_background")
     height: childrenRect.height
     width: parent.width
     Column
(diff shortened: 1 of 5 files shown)


Ghostkeeper (approved):
Don't really like it that we have to add a whole new theme entry for the detail of the highlight panel of the Marketplace window. Theme entries are supposed to be generic, such as "the colour of links" or "primary accent colour". A better description for the new theme entry would've been "a secondary background colour that is safe for links to appear on".

But since that is a long and also overly specific description, I'll have to accept this. That is not constructive criticism after all. But if you have a better idea to make the theme entry more generic, I'm open to suggestions.

konskarm:
> Don't really like it that we have to add a whole new theme entry for the detail of the highlight panel of the Marketplace window. Theme entries are supposed to be generic, such as "the colour of links" or "primary accent colour". A better description for the new theme entry would've been "a secondary background colour that is safe for links to appear on".
> 
> But since that is a long and also overly specific description, I'll have to accept this. That is not constructive criticism after all. But if you have a better idea to make the theme entry more generic, I'm open to suggestions.

I went for the overly specific since there were more specific toolbox colors. But I am open to suggestions.

PR #10332, opened 2021-08-24
Fix missing "star" icon in Cura 4.11

fieldOfView:
This PR adds an alias for the previously existing "star" icon that was used in the header for the profile dropdown to show a profile had changed settings.

The PR is against version 4.11, because if the alias is not there in Cura 4.11 then it makes little sense to add it back to the next version of Cura.

### DIFF
--- a/resources/themes/cura-light/icons/deprecated_icons.json
+++ b/resources/themes/cura-light/icons/deprecated_icons.json
@@ -167,6 +167,10 @@
         "new_icon": "Check",
         "size": "default"
     },
+    "star": {
+        "new_icon": "StarFilled",
+        "size": "default"
+    },
     "pencil": {
         "new_icon": "Pen",
         "size": "default"


nallath (approved):

nallath:
Whoops. We missed that. 

I will wait for CI to complete before merging it.

fieldOfView:
"cross2" is also missing by the way. I'll see if I can find more before I open another PR.

fieldOfView:
All of these are missing:
```
    basic
    buildplate
    category_shield
    category_unknown
    cross2
    drop_down_button
    group
    link
    play
    print_time
    printer_cloud_connected
    printer_cloud_not_available
    printer_connected
    printsetup
    quick
    scale_max
    setting_per_object
    star_empty
    star_filled
    ulti
    view_xray
    viewMode
```
That's so many that it does not look like it happened on accident. Is there a reasoning behind which icons were included and which icons were left out?

nallath:
Those were left out on accident. We manually went over the changes, but we just missed out a whole bunch.

The idea was to have a full list so that we would have better backwards compatibility.

PR #10356, opened 2021-08-30
Create anet3d_a8_v2.def.json

ntmds847:
Hi, this is Frank from Anet, A8 v2 is our latest released 3D printer. You can find its info here: https://anet3d.com/pages/a8-v2

You can contact me at shop@anet3d.com or ntmds847@hotmail.com

### DIFF
--- a/resources/definitions/anet3d_a8_v2.def.json
+++ b/resources/definitions/anet3d_a8_v2.def.json
@@ -1,0 +1,31 @@
+{
+    "version": 2,
+    "name": "Anet A8 V2",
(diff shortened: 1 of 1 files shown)


konskarm (changes requested):
Since this is almost exactly the same as the `anet3d_a8.def.json`, please make this file inherit from the `anet3d_a8.def.json` and change only the fields that need to be changed (e.g. the `machine_height`). This way, if for any reason the `anet3d_a8` gets updated, the v2 will get updated too.

PR #10357, opened 2021-08-30
Fix dev being (un-)plugged while still processing prev change

jellespijker:
The number of output devices can change while it is still processing a
previous change. Resulting in a RuntimeError because the dictionary
changed size during iteration.

I made two changes which should fix this:
- don't use a loop but a list comprehension. This is a faster method since
  it is done on the C side. So less change for the error to occur.
- Put a try, except block around it. Only handle the change in dictionary
  size. All other RuntimeError are still (re)raised.

https://sentry.io/share/issue/3665a9b68cbf44faa54a27e9fdd9d702/

### DIFF
--- a/cura/Settings/MachineManager.py
+++ b/cura/Settings/MachineManager.py
@@ -179,11 +179,16 @@ class MachineManager(QObject):
             self.setActiveMachine(active_machine_id)
 
     def _onOutputDevicesChanged(self) -> None:
-        self._printer_output_devices = []
-        for printer_output_device in self._application.getOutputDeviceManager().getOutputDevices():
-            if isinstance(printer_output_device, PrinterOutputDevice):
-                self._printer_output_devices.append(printer_output_device)
-
+        try:
+            output_devices = [d for d in self._application.getOutputDeviceManager().getOutputDevices() if
+                              isinstance(d, PrinterOutputDevice)]
+        except RuntimeError as e:
+            if "dictionary changed size during iteration" in str(e):
+                Logger.warning("The number of output devices changed while processing the previous changed number of output devices.")
+                return
+            else:
+                raise RuntimeError from e
+        self._printer_output_devices = output_devices
         self.outputDevicesChanged.emit()
 
     @pyqtProperty(QObject, notify = currentConfigurationChanged)


cura/Settings/MachineManager.py
```diff
@@ -179,11 +179,16 @@ def setInitialActiveMachine(self) -> None:
             self.setActiveMachine(active_machine_id)
 
     def _onOutputDevicesChanged(self) -> None:
-        self._printer_output_devices = []
-        for printer_output_device in self._application.getOutputDeviceManager().getOutputDevices():
-            if isinstance(printer_output_device, PrinterOutputDevice):
-                self._printer_output_devices.append(printer_output_device)
-
+        try:
+            output_devices = [d for d in self._application.getOutputDeviceManager().getOutputDevices() if
```
nallath, line 183: Isn't making a shallow copy a better idea? That way the list itself won't change size during iteration.

jellespijker: It was my understanding that a shallow copy creates a copy of the refs pointing to the actual data. In other words the normal python = als known as assignment. The actual data can still change. If we're going to copy I would suggest using a deep copy.

I personally think it's best to catch the exception and return if it occurs. Because that piece of code should run again, automatically, from the start. Since it is the handler for when the number of devices change.

nallath: I think you're confusing pointers from C++ with refs in python. Some sample code:

```python3
class SomeCoolObject:
    def __init__(self, name):
	    self.name = name
	
    def __repr__(self):
	    return self.name
        
first_list = [SomeCoolObject("beep"), SomeCoolObject("boop")]

copied_list = first_list.copy()

print(f"Lists before change {first_list} and {copied_list}")

for item in copied_list:
    first_list.remove(item)
    
print(f"Lists after change {first_list} and {copied_list}")
```

This results in the following prints:
```
Lists before change [beep, boop] and [beep, boop]
Lists after change [] and [beep, boop]
```
And no exceptions being thrown

jellespijker: @nallath  You then remove the reference ;-) not the underlying object.
Made a small change to your code to illustrate

```python
class SomeCoolObject:
    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return self.name

first_list = [SomeCoolObject("beep"), SomeCoolObject("boop")]

copied_list = first_list.copy()

print(f"Lists before change {first_list} and {copied_list}")
for item in copied_list:
    item.name = "new name"

print(f"Lists after change {first_list} and {copied_list}")
```

output
```
Lists before change [beep, boop] and [beep, boop]
Lists after change [new name, new name] and [new name, new name]
```

vs

```python
class SomeCoolObject:
    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return self.name

first_list = [SomeCoolObject("beep"), SomeCoolObject("boop")]

from copy import deepcopy
copied_list = deepcopy(first_list)

print(f"Lists before change {first_list} and {copied_list}")

for item in copied_list:
    item.name = "new name"

print(f"Lists after change {first_list} and {copied_list}")
```

output:
```
Lists before change [beep, boop] and [beep, boop]
Lists after change [beep, boop] and [new name, new name]
```

jellespijker: on the other hand..... do we care if the underlying data changes??? Lets do a call later on

nallath: That was my point. We don't care that the underlying data changed. The original code also didn't care about that. We're just trying to prevent it from tripping up on the size of the list being changed.

jellespijker: Convinced me.

jellespijker: Will do a small change later today

PR #10360, opened 2021-08-30
Add extruder to current version for version upgrade script

nallath:
This should remove the Unable to upgrade file of type extruder of version 2 messages

CURA-8509

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -471,7 +471,8 @@ class CuraApplication(QtApplication):
                 ("definition_changes", InstanceContainer.Version * 1000000 + self.SettingVersion):              (self.ResourceTypes.DefinitionChangesContainer, "application/x-uranium-instancecontainer"),
                 ("variant", InstanceContainer.Version * 1000000 + self.SettingVersion):                         (self.ResourceTypes.VariantInstanceContainer, "application/x-uranium-instancecontainer"),
                 ("setting_visibility", SettingVisibilityPresetsModel.Version * 1000000 + self.SettingVersion):  (self.ResourceTypes.SettingVisibilityPreset, "application/x-uranium-preferences"),
-                ("machine", 2):                                                                                 (Resources.DefinitionContainers, "application/x-uranium-definitioncontainer")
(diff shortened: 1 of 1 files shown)


cura/CuraApplication.py
```diff
@@ -471,7 +471,8 @@ def __setLatestResouceVersionsForVersionUpgrade(self):
                 ("definition_changes", InstanceContainer.Version * 1000000 + self.SettingVersion):              (self.ResourceTypes.DefinitionChangesContainer, "application/x-uranium-instancecontainer"),
                 ("variant", InstanceContainer.Version * 1000000 + self.SettingVersion):                         (self.ResourceTypes.VariantInstanceContainer, "application/x-uranium-instancecontainer"),
                 ("setting_visibility", SettingVisibilityPresetsModel.Version * 1000000 + self.SettingVersion):  (self.ResourceTypes.SettingVisibilityPreset, "application/x-uranium-preferences"),
-                ("machine", 2):                                                                                 (Resources.DefinitionContainers, "application/x-uranium-definitioncontainer")
+                ("machine", 2):                                                                                 (Resources.DefinitionContainers, "application/x-uranium-definitioncontainer"),
+                ("extruder", 2):                                                                                    (Resources.DefinitionContainers, "application/x-uranium-definitioncontainer")
```
Ghostkeeper, line 475: Hmm, but why was it a problem in the first place? We don't have any version upgrade for these two resource types, right? Nothing that parses definition files whatsoever...

Ghostkeeper: I'm worried that something is unnecessarily going through 1.8MB of JSON files now at every start-up.

Ghostkeeper: After doing some research, it is indeed unnecessarily going through all of these JSON files. This is coming from `ContainerInterface.deserialize` which always tries to perform a version upgrade for all containers.

If you ask me we should probably not do that there, or skip it if the container originates from the installation. However I'm limited by the timebox here to solve that. As such, this will have to do. Just realise that this might be unnecessarily adding to the start-up time.

Ghostkeeper (approved):

PR #10361, opened 2021-08-30
CURA-8520: Fix missing icons in deprecated icons list

konskarm:
Some icons from older Cura versions were not mapped to any of the new ones, which could cause plugins to have no icons in some places. 

- `basic` -> Not used, and there was no icon replacement designed for it. We decided to use the `PrintQuality.svg` which is similar.
- `buildplate` -> The icon existed only in the `UM3NetworkPrinting` plugin but was also used in the `ConfigurationItem.qml` in Cura. So the icon was moved from the `UM3NetworkPrinting` plugin to Cura.
- `category_shield` -> Remapped to `Shield.svg`.
- `category_unknown` -> Remapped to `Sputnik.svg`.
- `cross2` -> Remapped to `CancelBlock.svg`.
- `drop_down_button` -> Remapped to `ChevronSingleDown.svg`.
- `group` -> This icon existed in the welcome wizard in the cloud content and was wrongfully replaced by the "PrinterTriple.svg" in the code. Now, a `Group.svg` was added and was put back in the code.
- `link` -> Remapped to `Link.svg`.
- `play` -> It was readded to the theme and mapped to the `Play.svg`.
- `print_time` -> Remapped to `Clock.svg`.
- `printer_cloud_connected`, `printer_cloud_not_available` -> In 4.11 these two were both replaced by `CloudBadge.svg`(low) and they are distinguished by recoloring them (blue for available, grey for not available). So this is now being remapped to the `CloudBadge.svg` and it is up to the plugin dev to recolor it.
- `printer_connected` -> Remapped to `CheckBlueBG.svg` (low).
- `printsetup` -> There wasn't any new icon designed to replace this one, so we will use the `Sliders.svg`.
- `quick` -> Remapped to `Quick.svg`.
- `scale_max` -> Remapped to `ScaleMax.svg`.
- `setting_per_object` -> There wasn't any new icon designed to replace this one, so we will use the `MeshType.svg`.
- `star_empty` -> Remapped to `Star.svg`.
- `star_filled` -> Remapped to `StarFilled.svg`.
- `ulti` -> Not used and there was no icon replacement designed for it. We decided to use the `PrintQuality.svg` which is similar.
- `view_xray` -> Not used and there was no icon replacement designed for it. We decided to use the `CancelBlock.svg` which is similar.
- `view_normal` -> Remapped to `Infill0.svg`.
- `viewmode` -> Remapped to `Eye.svg`.

Many thanks to @fieldOfView's [comment](https://github.com/Ultimaker/Cura/pull/10332#issuecomment-904948163) for finding those and making a list out of them.

CURA-8520


fieldOfView:
> `play` -> Was only used in the `SimulationView` plugin, so there is no need for a remapping

I fail to see this logic. The whole idea of this missing icon mapping is for use in _external_ plugins; I would assume that for the code that ships with Cura you have already done the replacement in the actual code. Just removing an icon IMHO is a form of breaking the API (so it would require a major API version number update).

nallath:
Yeah, I agree with @fieldOfView 

The play could have been used by other plugins, so it needs to have a remap.

konskarm:
> > `play` -> Was only used in the `SimulationView` plugin, so there is no need for a remapping
> 
> I fail to see this logic. The whole idea of this missing icon mapping is for use in _external_ plugins; I would assume that for the code that ships with Cura you have already done the replacement in the actual code. Just removing an icon IMHO is a form of breaking the API (so it would require a major API version number update).

I think you misunderstood me. The icon itself is a resource in the SimulationView plugin in the first place. It does not exist in the theme at all, and as such, it cannot be remapped to something else.

fieldOfView:
See https://github.com/Ultimaker/Cura/blob/4.10/resources/themes/cura-light/icons/play.svg

nallath:
> See https://github.com/Ultimaker/Cura/blob/4.10/resources/themes/cura-light/icons/play.svg

I stand corrected yet again!

konskarm:
Indeed, you are absolutely right. It was probably made into a plugin-specific resource in 4.11 then and I missed that. I'll update it.

Ghostkeeper (approved):

PR #10371, opened 2021-09-01
add prime and optimus new device profile in cubicon.

harry572:
modified cubicon default profile.


nallath:
Our tests are failing. Please have a look at what needs to change.

For the next time, it's also better to add a machine in a single commit and make changes to existing machines in another. That way it's easier to revert things if it's ever needed.

nallath:
Friendly reminder; In order to get this merged some changes will need to be made.

PR #10381, opened 2021-09-02
Use secrets instead of random

nallath:
Since we're no longer stuck on python 3.5, we can use secrets instead of random which provides better randomness.

CURA-8401
SEC-207

### DIFF
--- a/cura/OAuth2/AuthorizationHelpers.py
+++ b/cura/OAuth2/AuthorizationHelpers.py
@@ -3,7 +3,7 @@
 
 from datetime import datetime
 import json
-import random
+import secrets
 from hashlib import sha512
 from base64 import b64encode
 from typing import Optional
@@ -143,7 +143,7 @@ class AuthorizationHelpers:
         better to leave it at 32
         """
 
-        return "".join(random.choice("0123456789ABCDEF") for i in range(code_length))
+        return "".join(secrets.choice("0123456789ABCDEF") for i in range(code_length))
 
     @staticmethod
     def generateVerificationCodeChallenge(verification_code: str) -> str:
--- a/cura/OAuth2/AuthorizationHelpers.py
+++ b/cura/OAuth2/AuthorizationHelpers.py
@@ -139,11 +139,11 @@ class AuthorizationHelpers:
     def generateVerificationCode(code_length: int = 32) -> str:
         """Generate a verification code of arbitrary length.
 
-        :param code_length:: How long should the code be? This should never be lower than 16, but it's probably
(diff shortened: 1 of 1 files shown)


cura/OAuth2/AuthorizationHelpers.py
```diff
@@ -143,7 +143,7 @@ def generateVerificationCode(code_length: int = 32) -> str:
         better to leave it at 32
         """
 
-        return "".join(random.choice("0123456789ABCDEF") for i in range(code_length))
+        return "".join(secrets.choice("0123456789ABCDEF") for i in range(code_length))
```
fieldOfView, line 146: Why not use `secrets.token_hex(code_length/2)`?

Ghostkeeper: As long as code_length is divisible by 2 at least.

Currently we never specify `code_length` in our code, only the default value of 32. However a plug-in could theoretically use it differently.

So I guess we have two choices:
* Change the parameter to `code_bytes_length` or something like that, indicating that it's the number of bytes in the code rather than the number of hexadecimal characters. And then use `token_hex`.
* Leave it as-is.

I think it's pretty safe to assume that no external plug-in is calling this function specifically, being only a small part of the OAuth workflow, so then we could change it. What do you think?

fieldOfView: You could do 'secrets.token_hex(math.ceil(code_length/2))[:code_length]'. But honestly I think your first option (and changing the default to 16) is the better option.

nallath: We don't explicitly define what length _means_, so all we would have to update is the documentation. So we could also use `secrets.token_hex(code_length)` right? 

The oauth still works even if the code is now a bit longer.

Ghostkeeper (approved):
While technically an API break, it's not going to hurt anyone so we'll accept it.

cura/OAuth2/AuthorizationHelpers.py
ChrisTerBeke: As per the OAuth2 PKCE extension specification, the verification code should be between 43 and 128 characters (https://datatracker.ietf.org/doc/html/rfc7636#section-4.1).

nallath: We used to have "(.Y.)" as the hardcoded verification code and even that was accepted :P

ChrisTerBeke: That was the `state` parameter, not the PKCE verification code (our server would definitely not allow that :P )

PR #10395, opened 2021-09-05
Correctly calculate the unreachable area for the extruders

### LINKED ISSUE #10354: Dual Extrusion - Build Volume / Disallowed Area is Incorrectly Calculated
### Application Version

4.10

### Platform

Windows 10

### Printer

Custom Ender 3 with 2 extruders

### Reproduction steps

Set up a custom printer with 2 nozzles
 - Nozzle 1 X = 0, Y = 0
- Nozzle 2 X = -24, Y = -24

Nozzle 1 (black) can reach full area. nozzle 2 can reach ( 0 , 0 ), but cannot reach past ( X_Max - 24 ) or ( Y_Max - 24 ) 

### Actual results

The Disallowed Areas prevent putting an object less than 24 mm into the build volume, even though both nozzles can reach that location.

Capture

### Expected results

Nozzle 2 should be able to reach ( 0 , 0 ). 
I would expect Nozzle 2 max X = (Stroke Limit - 24)

Capture

- [ ] Log file
- [ ] Project file

Attached is a screenshot of my hot end setup, the Nozzle Offset page for Extruder 2. (Extruder 3 was only mocked in as an additional extruder to test various values). Likewise, the Y Nozzle Offset was greatly exaggerated to showcase the Disallowed Area. 

Capture

As you can see, even though the The left-hand nozzle, which I have set up as Extruder # 2 (T1) is -24mm away from T0, the 'reachable area' is getting defined much further in on the left-hand side. Based off this calculation, CURA seems to believe that the Left nozzle (T1) can reach the right-most side of the bed (which it can not), but it is unable to reach the left-most side of the bed. 

Likewise, I have set the Y offset for E2 to -50. As you can see, CURA thinks that now between 0 and 50 are unusable, but 50-200 are usable. The reverse should be true. If the bed Y-Max is 200, and Y nozzle 2 is -50 offset compared to nozzle 1, then when Nozzle 1 is at 200, nozzle 2 can only reach maximum 150. 

The GCODE calculation itself is perfectly correct. To reach (0,0) with nozzle 2, the print head will move to (24,0). BUT I cannot slice if anything to be printed by nozzle 2 is less than 0 due to this 'Disallowed Area' calculation. 

Attempting to work around the issue and modify the nozzle offsets in Cura so that Nozzle 1 is offset compared to Nozzle 2 (which is now set to (0,0), you get an identical image. 

Now, changing the offset from (-24) to (24) for Nozzle 2 does two things:

- The 'Disallowed Area' is now 'correct'.
- The print will be incorrect because the GCODE will move to X=0 instead of X=24 to print with Extruder 2 at X=0. (in reality, this puts E2 at -24, off the print bed altogether.) Meaning this setting isn't usable (well, it can be if you turn 'Apply extruder offsets to gcode' setting to OFF', but this is ON by default). 

### Code Base 

I found this code inside of **CuraEngine\gcodeExport.cpp**

```
Point GCodeExport::getGcodePos(const coord_t x, const coord_t y, const int extruder_train) const
{
    if (use_extruder_offset_to_offset_coords)
    {
        const Settings& extruder_settings = Application::getInstance().current_slice->scene.extruders[extruder_train].settings;
        return Point(x - extruder_settings.get<coord_t>("machine_nozzle_offset_x"), y - extruder_settings.get<coord_t>("machine_nozzle_offset_y"));
    }
    else
    {
        return Point(x, y);
    }
}
```

and this code in  **BuildVolume.py -- _computeDisallowedAreasStatic function** 

```
{get details for some extruder }
offset_y = -offset_y  # Y direction of g-code is the inverse of Y direction of Cura's scene space.

... { loop through and compare all active extruders to determine unreachable borders } ...

other_offset_y = -other_offset_y
left_unreachable_border = min(left_unreachable_border, other_offset_x - offset_x)
right_unreachable_border = max(right_unreachable_border, other_offset_x - offset_x)
top_unreachable_border = min(top_unreachable_border, other_offset_y - offset_y)
bottom_unreachable_border = max(bottom_unreachable_border, other_offset_y - offset_y)
```

### SUMMARY

- The above code for generating gcode will move the head POSITIVE if the offset is NEGATIVE. This is correct functionality, and this occurs for both X and Y axes. (you wind up subtracting a negative in my examples, meaning you add.).
- The code for determining the "Unreachable_Border" is likely incorrect, and is the reason that the display engine doesn't match up to expectation.
- Here are my proposed changes:

Replace 
```
offset_y = - offset_y  # Y direction of g-code is the inverse of Y direction of Cura's scene space.
```
with 
```
offset_x = - offset_x  # X direction of g-code is the inverse of X direction of Cura's scene space.
```

and 
```
other_offset_y = -other_offset_y
left_unreachable_border = min(left_unreachable_border, other_offset_x - offset_x)
right_unreachable_border = max(right_unreachable_border, other_offset_x - offset_x)
```
with
```
other_offset_x = -other_offset_x
left_unreachable_border = max(left_unreachable_border, other_offset_x - offset_x)
right_unreachable_border = min(right_unreachable_border,  other_offset_x - offset_x)
```

Changing the Min/Max statements for Left/Right unreachable border brings them in line with the Top/Bottom unreachable border. Top = min, while Bottom = Max. If bottom = 0 and left = 0, Left should also use Max().

Then, removing the offset_y inversion, and replacing it with an offset_x inversion should resolve the issue where the disallowed areas seem inverted for both Y and X. 

Normally, I would test this by compiling and verifying my proposed code. But I followed everything in the wiki regarding setup for compiling on Windows, and it I cannot for the life of me get it to compile. VS2019 keeps complaining about nearly half of the python imports being unavailable, CMAKE issues, etc, etc. 
That, and I wanted to propose the changes with a lengthy argument for them to discuss, instead of just submitting a pull request for it.

RFBomb:
I have made a discussion on the issue linked below:

https://github.com/Ultimaker/Cura/issues/10354

Basically, by comparing nozzle offsets during this calculation, you wind up mirroring disallowed areas, restricting the build volume despite all nozzles being able to reach some locations. This is further discussed by my comments here:

https://github.com/Ultimaker/Cura/issues/10354#issuecomment-913214851

This is the changes requested in that issue.

Fixes #10354

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -952,20 +952,11 @@ class BuildVolume(SceneNode):
 
             # Only do nozzle offsetting if needed
             if nozzle_offsetting_for_disallowed_areas:
-                # The build volume is defined as the union of the area that all extruders can reach, so we need to know
-                # the relative offset to all extruders.
-                for other_extruder in ExtruderManager.getInstance().getActiveExtruderStacks():
-                    other_offset_x = other_extruder.getProperty("machine_nozzle_offset_x", "value")
-                    if other_offset_x is None:
-                        other_offset_x = 0
-                    other_offset_y = other_extruder.getProperty("machine_nozzle_offset_y", "value")
-                    if other_offset_y is None:
-                        other_offset_y = 0
-                    other_offset_y = -other_offset_y
-                    left_unreachable_border = min(left_unreachable_border, other_offset_x - offset_x)
-                    right_unreachable_border = max(right_unreachable_border, other_offset_x - offset_x)
-                    top_unreachable_border = min(top_unreachable_border, other_offset_y - offset_y)
-                    bottom_unreachable_border = max(bottom_unreachable_border, other_offset_y - offset_y)
+                    #Calculate the unreachable areas for this nozzle.
+                    left_unreachable_border = min(left_unreachable_border, - offset_x)
+                    right_unreachable_border = max(right_unreachable_border, - offset_x)
+                    top_unreachable_border = min(top_unreachable_border, - offset_y)
+                    bottom_unreachable_border = max(bottom_unreachable_border, - offset_y)
             half_machine_width = self._global_container_stack.getProperty("machine_width", "value") / 2
             half_machine_depth = self._global_container_stack.getProperty("machine_depth", "value") / 2
 


nallath:
I don't think that the changes here are correct. Are you sure you didn't just want to enable nozzle_offsetting_for_disallowed_areas?

RFBomb:
Looking through the code,
Disabling `nozzle_offsetting_for_disallowed_areas` Simply turns off the calculation altogether. 

The mesh for each nozzle is calculated using the following variables:

```
# Add the border around the edge of the build volume.
left_unreachable_border = 0
right_unreachable_border = 0
top_unreachable_border = 0
bottom_unreachable_border = 0
```

since the default of this is 0, and the nozzle offset is only applied by doing the calculation, we would still want to do the calculation to determine unreachable area. It basically has the same effect of just setting all the nozzle offsets to 0,0 which will allow you to slice into an inaccessible location. 

I believe the mistake with the original code is that it compares each nozzle to every other nozzle, instead of comparing the nozzle to where the printer says 0,0 is location. This resolves that calculation. From a user standpoint, and from the GCode generation standpoint, it is always offsetting with reference to 0. Calculating where nozzle 2 is doesn’t make any sense when determine if nozzle 1 can reach some location.

RFBomb:
To add more info on this:

I performed a search of the commits against this file using this git command:
`git log -S "left_unreachable_border =" --follow -p "cura\BuildVolume.py"`

I came up with just 1 result, which is when the code I modified was first introduced back in 2016. 
https://github.com/Ultimaker/Cura/commit/47bb8f9601799282b499b37a438aa8fe7caf2b52
Other than 2 minor bugfixes (set found offset value to 0 if value not set, and inversion of Y value due to screen space), this code has not changed at all.

--
Breaking down the explanation:
`It's a bit counter-intuitive, so let me explain. The total build volume is defined as the rectangle around the union of the areas that all extruders can reach. The nozzle offset for the left extruder though is 0,0, so what should the border on the right side be? It should be 18 since there is a nozzle to the right of the left nozzle which has a relative x-offset of 18mm. Therefore we need to compute the relative offsets with each of the other nozzles and create sort of a bounding box around it, and the distance to the edge of that bounding box is the unreachable border on each of the sides.`

- Example has 2 nozzles.
- Left extruder is assumed to be the nozzle that hits 0,0
- `nozzle to the right of the left nozzle which has a relative x-offset of 18mm`
     - Right nozzle has offset of 18mm compared to left nozzle with offset of 0mm
     - This means that the right nozzle cannot reach anything below 18mm, so a bounding box was required. (I agree)
- What this assumes (I believe incorrectly) is that the left nozzle is thus also unable to reach the last 18mm of the build volume on the right-hand side. While certainly this is a possible scenario, I don't think this would be the standard. 

--
Now, the code that was added in https://github.com/Ultimaker/Cura/commit/47bb8f9601799282b499b37a438aa8fe7caf2b52 I think was definitely a step in the right direction, as it added the min/max functions, but I think it may have gone too restrictive in the bounding box by imposing offsets from each extruder onto all other extruders. 

I'd be curious to hear from @Ghostkeeper regarding this. Its entirely possible that there is a need for the stricter bounding box, and I'm just not aware of it. 

The reason I went for the changes I proposed instead of changing the initialization of the 'unreachable' variables is because the toggle `nozzle_offsetting_for_disallowed_areas` was added for some reason, and I'm unsure why. Modifying the initial values of the variables off of zero would have unintended consequences affecting machines that were explicitly ignoring the offsets already, so I wanted to avoid that modification. That said, that may be the correct way to do it, and just keep the stricter bounding box available by using the `nozzle_offsetting_for_disallowed_areas` setting.


```
#Add the border around the edge of the build volume.
left_unreachable_border = 0
right_unreachable_border = 0
top_unreachable_border = 0
bottom_unreachable_border = 0
```
to 
```
#Add the border around the edge of the build volume.
left_unreachable_border = min(0, - offset_x)
right_unreachable_border = max(0, - offset_x)
top_unreachable_border = min(0, - offset_y)
bottom_unreachable_border = max(0, - offset_y)
```

Ghostkeeper:
I made a comment in your issue report that answers your questions and tries to explain where your line of thinking went wrong: https://github.com/Ultimaker/Cura/issues/10354#issuecomment-915121668

PR #10403, opened 2021-09-07
FilamentChange: Fix unload on marlin based printers

barnybug:
See related #10038. This fixes the issue by referencing the gcode flavor set for the machine.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/FilamentChange.py
+++ b/plugins/PostProcessingPlugin/scripts/FilamentChange.py
@@ -7,6 +7,8 @@
 from typing import List
 from ..Script import Script
 
+from UM.Application import Application
+
 class FilamentChange(Script):
 
     _layer_keyword = ";LAYER:"
@@ -106,7 +108,11 @@ class FilamentChange(Script):
                 color_change = color_change + (" E%.2f" % initial_retract)
 
             if later_retract is not None and later_retract > 0.:
-                color_change = color_change + (" L%.2f" % later_retract)
+                gcode_flavor = Application.getInstance().getGlobalContainerStack().getProperty("machine_gcode_flavor", "value")
+                param = "U" if gcode_flavor == "RepRap (Marlin/Sprinter)" else "L"
+                # Reprap uses 'L': https://reprap.org/wiki/G-code#M600:_Filament_change_pause
+                # Marlin uses 'U' https://marlinfw.org/docs/gcode/M600.html
+                color_change = color_change + (" %s%.2f" % (param, later_retract))
 
             if x_pos is not None:
                 color_change = color_change + (" X%.2f" % x_pos)


nallath (approved):

plugins/PostProcessingPlugin/scripts/FilamentChange.py
```diff
@@ -106,7 +108,11 @@ def execute(self, data: List[str]):
                 color_change = color_change + (" E%.2f" % initial_retract)
 
             if later_retract is not None and later_retract > 0.:
-                color_change = color_change + (" L%.2f" % later_retract)
+                gcode_flavor = Application.getInstance().getGlobalContainerStack().getProperty("machine_gcode_flavor", "value")
+                param = "U" if gcode_flavor == "RepRap (Marlin/Sprinter)" else "L"
```
nallath, line 112: I thought that this was correct, but after playing around with this a bit, I think it might not be quite right.

There are some types that are marlin based (eg; UltiGcode, griffin, RepRap (Volumetric)). These will now use the L parameter and not the U parameter.

I think it might be better to change the style to be more like what pause at height does; It lets the user pick a certain firmware style and change the behavior based on that.

nallath (changes requested):

plugins/PostProcessingPlugin/scripts/FilamentChange.py
barnybug: Would it be better to capture these differences in the global gcode flavor setting so plugins can reliably adapt automatically? (Or is that not possible currently?)

I could switch the logic to instead to use 'U' by default and 'L' only on a particular flavor if that makes more sense. Was going for keeping the current behaviour for most to avoid any unintended breakage.

nallath: The pause plugin does something like this:

```
"pause_method":
{
    "label": "Method",
    "description": "The method or gcode command to use for pausing.",
    "type": "enum",
    "options": {"marlin": "Marlin (M0)", "griffin": "Griffin (M0, firmware retract)", "bq": "BQ (M25)", "reprap": "RepRap (M226)", "repetier": "Repetier/OctoPrint (@pause)"},
    "default_value": "marlin",
    "value": "\\\"griffin\\\" if machine_gcode_flavor==\\\"Griffin\\\" else \\\"reprap\\\" if machine_gcode_flavor==\\\"RepRap (RepRap)\\\" else \\\"repetier\\\" if machine_gcode_flavor==\\\"Repetier\\\" else \\\"bq\\\" if \\\"BQ\\\" in machine_name or \\\"Flying Bear Ghost 4S\\\" in machine_name  else \\\"marlin\\\""
},
"machine_gcode_flavor":
{
    "label": "G-code flavor",
    "description": "The type of g-code to be generated. This setting is controlled by the script and will not be visible.",
    "type": "enum",
    "options":
    {
        "RepRap (Marlin/Sprinter)": "Marlin",
        "RepRap (Volumetric)": "Marlin (Volumetric)",
        "RepRap (RepRap)": "RepRap",
        "UltiGCode": "Ultimaker 2",
        "Griffin": "Griffin",
        "Makerbot": "Makerbot",
        "BFB": "Bits from Bytes",
        "MACH3": "Mach3",
        "Repetier": "Repetier"
    },
    "default_value": "RepRap (Marlin/Sprinter)",
    "enabled": false
},

```

> Would it be better to capture these differences in the global gcode flavor setting so plugins can reliably adapt automatically? (Or is that not possible currently?)

What do you mean? 

> Was going for keeping the current behaviour for most to avoid any unintended breakage.

Fair point. That is usually our strategy.

barnybug: Sorry, lacking full explanation!

I was hoping it could just use the machine_gcode_flavor alone to determine whether to use 'L' or 'U'.

Do you know what's the reason pause method defaults, but allows it to be overridden?

Ghostkeeper: > Do you know what's the reason pause method defaults, but allows it to be overridden?

Because in general there aren't just those 9 g-code flavours, but thousands of variations of firmware which have various levels of support for all sorts of things and released at various moments along the firmwares' development. So recently we've been stopping hard-coding things into the g-code flavour, and offer per-feature support for variations in g-code (and printers in general). The default can still be selected via the g-code flavour of course, so that the user doesn't have to know unless their firmware is outside of the norm.

barnybug:
@Ghostkeeper  @nallath Thanks for the input both. I've updated the PR with suggested approach.
Let me know what you think!

nallath (approved):

nallath:
Thanks for the patience and your contribution! It will be part of the 4.12 release :)

PR #10431, opened 2021-09-11
Update the thinker printer profile

Eryone:
Improved the print setting


resources/quality/eryone_thinker/eryone_thinker_extra_fast.inst.cfg
```diff
@@ -0,0 +1,37 @@
...
+quality_type = verydraft
+weight = -3
+global_quality = True
+
+[values]
+acceleration_print = 1500
+acceleration_travel = 3000
+infill_sparse_density = 10
+initial_layer_line_width_factor = 100.0
+layer_height = 0.3
```
konskarm, line 18: Not sure if this intended, but the extra fast quality has the same layer height as the fast quality, which results two seemingly same quality choices in the recommended tab. 
image

Our usual approach suggest that each new quality profile has a different layer height attached to it. Even though it is not enforced, it is something we expect to see so I'll discuss with the team to see if having two qualities with the same layer height is something that it is accepted.

Ghostkeeper: We've discussed this and indeed we'd prefer to have unique layer heights for all quality levels. This is A) a mental thing and B) what our interface is designed around. People associate profiles mostly with their layer height, so they identify profiles with layer heights as well. This is exasperated by the design of the interface which mentions the layer height prominently in the top bar, in the Recommended mode and in the profile selection drop-down.

An alternative to this would be to add an intent profile for the 0.3mm layer height that indicates the difference between these two profiles.

dapostol73: @Ghostkeeper As I setup the profile, the intent was to have a profile that would reduce time greatly for the user and who be the max speed the Printer is cable of.  As the stock printer comes with a 0.4 nozzle, we can not exceed 0.3mm for layer height, hence no change.  Can you point me to documentation or explain how I can create an "intent" profile?

Eryone: @Ghostkeeper @konskarm Thanks for the reminder! 
We will update the profile and use the different layer heights to convert the old one.

konskarm:
Everything looks good now and works without issues. Also, models slice normally with the new Extra Fast quality profile. As a result, this PR is accepted. The new changes should appear in Cura 4.12 onwards.

PR #10437, opened 2021-09-12
Add Profile for Creality Ender 6

HOELLII:
Is there a reason this printer isn't yet added to Cura?

### DIFF
--- a/resources/definitions/creality_ender6.def.json
+++ b/resources/definitions/creality_ender6.def.json
@@ -1,0 +1,48 @@
+{
+    "name": "Creality Ender-6",
+    "version": 2,
+    "inherits": "creality_base",
+    "overrides": {
+        "machine_name": { "default_value": "Creality Ender-6" },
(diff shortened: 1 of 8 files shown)


HOELLII:
Updated variants to version 17. Check again.

HOELLII:
Why does the check fail at the json file?
It is the Original File from the Creality Slicer 4.8.2 which is based on Cura.

I mean, i can remove the part with "machine_head_polygon", but i don't know if that's the right way.

Anyone has an idea or suggestion?

nallath:
The machine_head_polygon was merged with machine_head_with_fans_polygon in 4.5, So if you change the setting to machine_head_with_fans_polygon it should work.

Ghostkeeper:
For a while we had both polygons. Until at some point we figured only the shape with fans is really relevant for collision checking, and we removed the one without.

nallath (approved):

PR #10449, opened 2021-09-14
Version upgrade for 4.11 to 4.12: Upgrade Flsun stacks to use printer-specific profiles

Ghostkeeper:
A small change with big consequences.

This adds a version upgrade. The setting_version is upped to 19 now. In version 19, FLSun has printer-specific profiles.
A mapping is made for existing printers to select the quality profile which is closest to the quality profile they previously had.

Implements the remainder of CURA-8510.


nallath (changes requested):
We need to skip the 500018 version here. That's also the one we use for Arachne. So this upgrade should upgrade to 19

nallath (approved):

Ghostkeeper:
Our QA had something similar when upgrading from a recent Master version. What didn't go quite right is that https://github.com/Ultimaker/Cura/pull/10192 was merged without version upgrade and the version upgrade was only added later. So if you came from this intermediary master-version you could already have the "new" setup for this FLSun SR printer while still on `setting_version` 17. It wasn't an atomic merge there.

When coming from Cura 4.11 (or equivalent for your fork) it shouldn't occur. I'm not able to reproduce the error in any case, making a few profiles with and without FLSun SR enabled and then upgrading to current master.

If you're getting this from a state before #10192 was merged, when upgrading to a state after this PR was merged, then I'd love to see that configuration folder! It could be quite a serious bug if something goes wrong there.

PR #10491, opened 2021-09-21
CURA-8421: Introduce app switcher

konskarm:
This PR introduces the ApplicationSwitcher widget. For the time being, the application switcher contains only the Ultimaker platform links in a static model.

CURA-8421


resources/qml/ApplicationSwitcher/ApplicationButton.qml
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        Label
+        {
+            id: applicationDisplayName
+
+            anchors.horizontalCenter: parent.horizontalCenter
+
+            width: base.width - UM.Theme.getSize("default_margin").width
```
nallath, line 92: Why not do this with anchors?

resources/qml/ApplicationSwitcher/ApplicationSwitcher.qml
```diff
@@ -0,0 +1,182 @@
...
+        id: applicationSwitcherButton
+        anchors.verticalCenter: parent.verticalCenter
+        anchors.horizontalCenter: parent.horizontalCenter
+
+        width: Math.round(0.5 * UM.Theme.getSize("main_window_header").height)
+        height: width
+
+        background: Item
+        {
+            anchors.fill: parent
```
nallath, line 28: No need to set this; https://doc.qt.io/qt-5/qml-qtquick-controls2-control.html#background-prop

nallath (dismissed):

resources/qml/ApplicationSwitcher/ApplicationSwitcher.qml
```diff
@@ -0,0 +1,182 @@
...
+
+            Rectangle
+            {
+                anchors.fill: parent
+                radius: UM.Theme.getSize("action_button_radius").width
+                color: applicationSwitcherButton.hovered ? UM.Theme.getColor("primary_text") : "transparent"
```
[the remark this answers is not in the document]
konskarm: Wouldn't it be a bit of an overkill to use states for only the color property of this single element?

resources/qml/ApplicationSwitcher/ApplicationButton.qml
konskarm: The result will be the same, but if you feel like it will look cleaner, I'll use anchors here.

resources/qml/ApplicationSwitcher/ApplicationSwitcher.qml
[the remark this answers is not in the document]
nallath: Hmm, yeah. You're right.

PR #10509, opened 2021-09-24
Set a minimal jerk limit of 20mm/s for UM3, S3 and S5

pkuiper-ultimaker:
A Jerk limit below 20mm/s will not create any mechanical ringing. The start and stop motions due to acceleration when a jerk limit is exceeded creates line width artefacts. Therefore a minimum jerk of 20 is enforced.


Ghostkeeper:
I fixed the overrides that were causing the automated tests to fail for you. The resulting behaviour should be the same.
Please pull in those changes before you continue making profile updates!

pkuiper-ultimaker:
Thanks. I will pull them.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #10528, opened 2021-09-27
Change Ender 5 Plus end gcode to prevent BLTouch collision

mothnox:
If the Ender 5 Plus is turned off without moving the extruder after a print, upon turning it back on, the BLTouch will hit the frame of the printer and enter an error state. This prevents that by changing the end gcode for the Ender 5 Plus to move the extruder to X{machine_width} Y{machine_depth} instead of X0 Y{machine_depth}, where the BLTouch is not positioned to hit the frame when extended.


mothnox:
I didn't change anything in that regard. The Ender 5 Plus was inheriting the end gcode from the base Creality definition file. I just copied that to the Ender 5 Plus file and changed the X value. That said, if I have time later, I can try printing something at max height and see if it's a problem. If it is, it might need to be changed for more than just the Ender 5 Plus.

nallath:
I missed that it was already the case in the base Creality file, sorry for that!

If it was already the case, I don't see an issue with merging this right now. I would consider the raise a separate issue in that case. In case something needs to change there, feel free to open an extra PR :)

nallath (approved):

PR #10534, opened 2021-09-28
Add 2nd Extruder to Duplicator 4s profile

roger-phillips:
Duplicator 4s is a dual extruder and the current profile listed has only one extruder listed.


fieldOfView:
I think this will require updating existing machine instances, so a version update.

nallath:
Yep; existing machine stacks need to get an extra extruder added.

Ghostkeeper (changes requested):
This needs a version upgrade:

* Split any extruder stack for the Duplicator 4s into two extruder stack files.
* Split any custom profile for the Duplicator 4s into 3 profiles instead of 2 (so 2 for the extruders and 1 global subprofile). The two extruder subprofiles need to have the same setting overrides.
* Split any extruder-user profile for the Duplicator 4s into 2 profiles, with the same setting overrides.
* The extruder stacks need to be assigned their two split profiles for the quality-changes and user profiles, and the new extruder definition you added. The rest (material and nozzle) simply needs to be duplicated.

This would need to be implemented in the version upgrade script for 4.12: https://github.com/Ultimaker/Cura/blob/master/plugins/VersionUpgrade/VersionUpgrade411to412/VersionUpgrade411to412.py

nallath:
Closing this due to inactivity. If this machine should still be added, please ensure that the previous comments have been resolved.

PR #10537, opened 2021-09-28
Klema printers

Vladislav0403:
I am a representative of 3DDevice company
I would like to add the profiles of our printers (Klema 180, Klema 250, Klema 500), extruders and table models to your sicer
We'll be very thankful


nallath (changes requested):
Please reduce the meshes to well below 1 MB per mesh.

nallath:
Closing this due to inactivity. Requested changes have not been addressed.

PR #10553, opened 2021-10-03
Add new prints for Vivendino / Formbot, the T-Rex 2+ and the T-Rex 3,  large format IDEX printers.

jim-edwards:
Adding initial support for the Vivendio / Formbot based 3D printers, found here: https://www.formbot3d.com/.  Tested on real hardware, I am going to have additional users on the brand user-group provide additional testing to ensure compatibility.  Further requests will follow with duplicator mode.

It would be nice is "Extruder specific" start code could be used.  These printers have a wipe for the nozzle, which on other slicers I have set to run specific g-code on the start, but only if that nozzle is enabled.  The existing start gcode runs every time the nozzle is selected, while this would run only once, basically if the nozzle is enabled, add this start gcode.

Creating pull to get initial feedback.


Ghostkeeper:
The automated tests are complaining about a couple of things:
* Build plate model files have to be less than 1MB. This is to keep their complexity to a reasonable level, saving download time, start-up time and rendering time.
* `vivedineo_trex2plus.def.json` and `vivedino_trex3.def.json` specify a `default_value` for the Gantry Height setting, but that setting has a `value`, which will always override the `default_value`.
* Definitions have an ID field, which doesn't exist as a field.
* The profiles are outdated. `setting_version` is currently 19, not 17.

jim-edwards:
I am struggling to get the unit tests to run locally.  Running ctest --verbose in my directory says no tests are found.  Trying to build a cmake output first for several days now and have been unable to do so.  I can't get the right dependencies installed.  I started looking at the various cura-build and cura-build-environment repos and I must be missing something that makes this easy..  I suspect I just need some environment setup to build this thing, before I can run the unit tests.  I don't need the unit tests to actually test my JSON files, so I have not been building the software from scratch.  I am not really a python dev, so apologies, but is there an idiots guide to building this thing?

For the inheritance, it seems to work as I would expect, so is that just failing in the unit-test?

nallath:
I run the tests by using `PYTHONPATH=.:../Uranium python3 -m pytest`

Note that doesn't run _all_ the tests (not the plugin tests) but you don't really need those.

jim-edwards:
I believe I fixed the remaining issues, so hopefully this passes the test system :)

jim-edwards:
@Ghostkeeper, what is needed to get this merged?  I believe the issues were resolved.

nallath:
Apologies for the very late reaction!

I've found a number of mistakes with the meshes that you provided.

image

image

jim-edwards:
@nallath, thanks for pointing those out, I have updated the solution to use a single mesh for both units, since they are virtually identical.    The optimization should now have a better/complete mesh.

jim-edwards:
@nallath, thanks for pointing those out, I have updated the solution to use a single mesh for both units, since they are virtually identical.    The optimization should now have a better/complete mesh.

PR #10554, opened 2021-10-03
change at z update

legend069:
hello, I'm unsure of the requirements for creating pull requests, sorry in advanced.

I've fixed a couple typos and added the "enclosure temp" function to be adjusted.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
+++ b/plugins/PostProcessingPlugin/scripts/ChangeAtZ.py
@@ -11,7 +11,7 @@
 # Modified by Jaime van Kessel (Ultimaker), j.vankessel@ultimaker.com to make it work for 15.10 / 2.x
 # Modified by Ruben Dulek (Ultimaker), r.dulek@ultimaker.com, to debug.
 # Modified by Wes Hanney, https://github.com/novamxd, Retract Length + Speed, Clean up
-# Modified by Alex Jaxon, https://github.com/legend069, Added option to modify enclosure temperature
+# Modified by Alex Jaxon, https://github.com/legend069, Added option to modify Build Volume Temperature
 
 
 # history / changelog:
@@ -42,7 +42,7 @@
 # Added support for outputting changes to LCD (untested). Added type hints to most functions and variables. Added more comments. Created GCodeCommand
 # class for better detection of G1 vs G10 or G11 commands, and accessing arguments. Moved most GCode methods to GCodeCommand class. Improved wording
 # of Single Layer vs Keep Layer to better reflect what was happening.
-# V5.2.2    Alex Jaxon, Added option to modify enclosure temperature keeping current format
+# V5.2.2    Alex Jaxon, Added option to modify Build Volume Temperature keeping current format
 #           updated from "Experimental" to "Beta"
 #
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
"Enclosure Temp" is called "Build Volume Temperature" in the rest of Cura. It would be good to call it by the same name.

legend069:
> 
> 
> Hi! Sorry for the wait, replies aren't meant to take this long. Great to see a first-time contributor.
> 
>     * I think it's a really strange decision to change 'Experimental' to 'Beta' unilaterally on your own, without even asking any other developers, as a first time contributor, when that has nothing to do with what you're adding. What made you decide that? I'll be removing that change.
> 
>     * I don't think you should change the gcode comments from what they are now ('CAZO'/'CAZD') to 'ChangeAtZ'. You're replacing different things by the same thing, and I've also seen that people's scripts (done _after_ ours) or metrics rely on these comments being there (I think we had to put them back at one point?)
> 
>     * Capitalisation/Code-style: The variables for 'BuildVolumeTemperature' all start with a capital, unlike, for example, 'bedTemperature'.
> 
> 
> I can make these changes myself if you're OK with it (and we test/review that it all works), but they are extensive relative to the size of the PR, so I'm more comfortable asking you first.

I've done a lot of testing with the script and it does look like it can be moved into the "Beta" stage.
as i believe users are avoiding using it because it is labeled as "experimental".
you're more than welcome to revert this back.

the changes for ('CAZO'/'CAZD') to 'ChangeAtZ was only changed to make it easier to read the gcode once it had been sliced , not everyone knows what "CAZO'/'CAZD" means.
if it is required for other parts of cura i did not think of that.
by reading through the code "CAZO'/'CAZD" was just used for the start/end of the code block that would be inserted, didn't look like it was being used for anything else.


RE:Capitalisation/Code-style, i am used to starting each word with a capital letter that's just my method that i use. makes it easier to read a several word label.


you're more than welcome to make any changed you see fit, adding the build volume feature was a "temp" fix to the main problem I have for a start g code.

thank you.

rburema:
Thanks for the explanation, however:

> I've done a lot of testing with the script and it does look like it can be moved into the "Beta" stage.
> as i believe users are avoiding using it because it is labeled as "experimental".

That may be, but it just doesn't meet our quality standards yet. It's also a bit strange to just decide this unilaterally, t.b.h., no matter how much you personally tested.

> you're more than welcome to revert this back.

I've done so now, in order to test the changes as they will be once this is confirmed to work and merged, but usually, if we change anything, it's just one or two lines, or code style.

Just a heads-up for the general case: It's normally up to the implementor to make sure the team is OK with merging it. We sometimes make exceptions (like now) but you can't rely on that. That said: You can increase the chance we revert stuff we don't want instead of just bouncing it back to you by making more atomic commits. (Then I could have just done a 'git revert' instead of doing it all manually.)

> if it is required for other parts of cura i did not think of that.

Not by Cura itself per-se (though another script _does_ look for a 'ChangeAtZ' string, which _may_ also trip it up.). I know people are using it in their own (post-)post-processing, after it gets out of Cura. (Also plugins might be using it? Less certain of that though.)

>  i am used to starting each word with a capital letter that's just my method that i use. makes it easier to read a several word label.

It's not so much about which style is better, it's about consistency :-)

rburema:
Alright, looks fine. Or at least as fine as 'bed temperature is': Case in point: I found a bug/known issue: Neither bed temp nor volume temp listen to 'target this layer only'.

Anyway, we'll deal with that separately.

Thanks for your contribution! :-)

PR #10558, opened 2021-10-03
Fixed Extruder hitting frame after finished print

Rakhmanov:


rburema:
Perhaps that should be `G28 X Y` as a plain `G28` will home all axis.
See https://marlinfw.org/docs/gcode/G028.html

Have you tested this change with the print still on the build-plate?

Rakhmanov:
I've deleted the branch, cant make a change will close this PR.

Issue #8776, opened 2020-11-15
Support Blocker Unable to Duplicate


esup:
**Application version**
4.8

**Platform**
Windows 10 64 bit

**Printer**
Ender 3 v2

**Reproduction steps**
1. Import model
2. Add support blocker
3. Right mouse click on support blocker
4. Multiple Model 

Nothing happens

**Actual results**
Nothing

**Expected results**
Support blocker should have been duplicated

**Project file**
(For slicing bugs, provide a project which clearly shows the bug, by going to File->Save Project. For big files you may need to use WeTransfer or similar file sharing sites. G-code files are not project files!)

**Log file**
2020-11-15 16:21:11,479 - DEBUG - [MainThread] UM.Operations.OperationStack.push [72]: GroupedOp.(#=1) <UM.Operations.RemoveSceneNodeOperation.RemoveSceneNodeOperation object at 0x000001B0390D84A8>, took 4ms
2020-11-15 16:21:12,055 - DEBUG - [MainThread] UM.Operations.OperationStack.push [72]: GroupedOp.(#=1) <UM.Operations.RemoveSceneNodeOperation.RemoveSceneNodeOperation object at 0x000001B043F60940>, took 10ms
2020-11-15 16:21:12,596 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [406]: file:///C:/Program Files/Ultimaker Cura 4.8.0/resources/qml/ObjectSelector.qml:132:23: Unable to assign [undefined] to QString
2020-11-15 16:21:12,598 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [406]: file:///C:/Program Files/Ultimaker Cura 4.8.0/resources/qml/ObjectSelector.qml:134:49: Unable to assign [undefined] to bool
2020-11-15 16:21:12,600 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [406]: file:///C:/Program Files/Ultimaker Cura 4.8.0/resources/qml/ObjectSelector.qml:135:54: Unable to assign [undefined] to int
2020-11-15 16:21:12,602 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [406]: file:///C:/Program Files/Ultimaker Cura 4.8.0/resources/qml/ObjectSelector.qml:136:43: Unable to assign [undefined] to QString
[...]

PR #10562, opened 2021-10-04
Fix duplicating / multiplicating support blockers

nallath:
CURA-7851

### DIFF
--- a/cura/Arranging/Nest2DArrange.py
+++ b/cura/Arranging/Nest2DArrange.py
@@ -114,7 +114,7 @@ def createGroupOperationForArrange(nodes_to_arrange: List["SceneNode"],
                                    build_volume: "BuildVolume",
                                    fixed_nodes: Optional[List["SceneNode"]] = None,
                                    factor = 10000,
-                                   add_new_nodes_in_scene: bool = False):
+                                   add_new_nodes_in_scene: bool = False)  -> Tuple[GroupedOperation, int]:
     scene_root = Application.getInstance().getController().getScene().getRoot()
     found_solution_for_all, node_items = findNodePlacement(nodes_to_arrange, build_volume, fixed_nodes, factor)
 
@@ -134,8 +134,7 @@ def createGroupOperationForArrange(nodes_to_arrange: List["SceneNode"],
         else:
             # We didn't find a spot
             grouped_operation.addOperation(
-                TranslateOperation(node, Vector(200, node.getWorldPosition().y, -not_fit_count * 20),
-                                   set_position=True))
+                TranslateOperation(node, Vector(200, node.getWorldPosition().y, -not_fit_count * 20), set_position = True))
             not_fit_count += 1
 
     return grouped_operation, not_fit_count
--- a/cura/Arranging/Nest2DArrange.py
+++ b/cura/Arranging/Nest2DArrange.py
@@ -158,5 +158,5 @@ def arrange(nodes_to_arrange: List["SceneNode"],
     """
 
     grouped_operation, not_fit_count = createGroupOperationForArrange(nodes_to_arrange, build_volume, fixed_nodes, factor, add_new_nodes_in_scene)
-    grouped_operation.redo()
+    grouped_operation.push()
     return not_fit_count != 0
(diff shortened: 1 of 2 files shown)


nallath:
This should provide a fix for  https://github.com/Ultimaker/Cura/issues/8776

cura/Arranging/Nest2DArrange.py
```diff
@@ -110,18 +110,11 @@ def findNodePlacement(nodes_to_arrange: List["SceneNode"], build_volume: "BuildV
...
-    :param factor: The library that we use is int based. This factor defines how accuracte we want it to be.
-    :param add_new_nodes_in_scene: Whether to create new scene nodes before applying the transformations and rotations
-
-    :return: found_solution_for_all: Whether the algorithm found a place on the buildplate for all the objects
-    """
+def createGroupOperationForArrange(nodes_to_arrange: List["SceneNode"],
+                                   build_volume: "BuildVolume",
+                                   fixed_nodes: Optional[List["SceneNode"]] = None,
+                                   factor = 10000,
+                                   add_new_nodes_in_scene: bool = False):
```
konskarm, line 117:
```suggestion
                                   add_new_nodes_in_scene: bool = False)  -> Tuple[GroupedOperation, int]:
```

cura/Arranging/Nest2DArrange.py
```diff
@@ -141,8 +134,30 @@ def arrange(nodes_to_arrange: List["SceneNode"], build_volume: "BuildVolume", fi
         else:
             # We didn't find a spot
             grouped_operation.addOperation(
-                TranslateOperation(node, Vector(200, node.getWorldPosition().y, -not_fit_count * 20), set_position = True))
+                TranslateOperation(node, Vector(200, node.getWorldPosition().y, -not_fit_count * 20),
+                                   set_position=True))
```
konskarm, line 138:
```suggestion
                TranslateOperation(node, Vector(200, node.getWorldPosition().y, -not_fit_count * 20), set_position = True))
```

cura/MultiplyObjectsJob.py
```diff
@@ -56,19 +58,33 @@ def run(self) -> None:
...
         found_solution_for_all = True
         if nodes:
-            found_solution_for_all = arrange(nodes, Application.getInstance().getBuildVolume(), fixed_nodes,
-                                             factor = 10000, add_new_nodes_in_scene = True)
+            group_operation, not_fit_count = createGroupOperationForArrange(nodes,
+                                                                            Application.getInstance().getBuildVolume(),
+                                                                            fixed_nodes,
+                                                                            factor=10000, add_new_nodes_in_scene=True)
+        else:
+            group_operation = GroupedOperation()
```
konskarm, line 81: This whole if statement can be simplified as follows:

``` python
group_operation = GroupedOperation()
if nodes:
    group_operation, not_fit_count = createGroupOperationForArrange(nodes,
                                                                            Application.getInstance().getBuildVolume(),
                                                                            fixed_nodes,
                                                                            factor = 10000, add_new_nodes_in_scene = True)
```

cura/MultiplyObjectsJob.py
```diff
@@ -56,19 +58,33 @@ def run(self) -> None:
...
-                nodes.append(new_node)
+                if not current_node.getParent().callDecoration("isSliceable"):
+                    nodes.append(new_node)
+                else:
+                    # The node we're trying to place has another node that is sliceable as a parent.
+                    # As such, we shouldn't arrange it (but it should be added to the scene!)
+                    nodes_to_add_without_arrange.append(new_node)
+                    new_node.setParent(current_node.getParent())
 
         found_solution_for_all = True
```
konskarm, line 74: This variable seems to always be True now. Shouldn't it change based on the outcome of the `createGroupedOperationForArrange()`? 
If it doesn't, then the message `no_full_solution_message` later on can be removed.

konskarm (changes requested):
Some code-style changes and a request to simplify a small part. 
Also, bit of a nitpick, but I think that the new eraser node that is added should be moved at least a tiny bit. Right now it's not visible that it is multiplied if you don't have the Object list in the bottom left corner expanded, because it the new Eraser block is right under the old one.

PR #10570, opened 2021-10-05
CURA-8254: Trigger material profile sync with printer

konskarm:
This PR adds a message on startup which will appear if the user has just installed or updated material packages. The message will only show if the user has printers which support material export.

image

Requires https://github.com/Ultimaker/Uranium/pull/743.

CURA-8254

### DIFF
--- a/cura/Machines/Models/MaterialManagementModel.py
+++ b/cura/Machines/Models/MaterialManagementModel.py
@@ -73,6 +73,8 @@ class MaterialManagementModel(QObject):
                 button_style = Message.ActionButtonStyle.LINK
         )
         sync_materials_message.actionTriggered.connect(self._onSyncMaterialsMessageActionTriggered)
+
+        # Show the message only if there are printers that support material export
         container_registry = cura.CuraApplication.CuraApplication.getInstance().getContainerRegistry()
         global_stacks = container_registry.findContainerStacks(type = "machine")
         if any([stack.supportsMaterialExport for stack in global_stacks]):
--- a/cura/Machines/Models/MaterialManagementModel.py
+++ b/cura/Machines/Models/MaterialManagementModel.py
@@ -7,6 +7,8 @@ from typing import Any, Dict, Optional, TYPE_CHECKING
 import uuid  # To generate new GUIDs for new materials.
 import zipfile  # To export all materials in a .zip archive.
 
+from PyQt5.QtGui import QDesktopServices
+
 from UM.i18n import i18nCatalog
 from UM.Logger import Logger
 from UM.Message import Message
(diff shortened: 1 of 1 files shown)


cura/Machines/Models/MaterialManagementModel.py
```diff
@@ -21,13 +23,72 @@
...
     """Triggered when a favorite is added or removed.
 
     :param The base file of the material is provided as parameter when this emits
     """
 
+    def __init__(self, parent: Optional[QObject] = None) -> None:
+        super().__init__(parent = parent)
+        self._checkIfNewMaterialsWereInstalled()
+
+    def _checkIfNewMaterialsWereInstalled(self):
```
nallath, line 38:
```suggestion
    def _checkIfNewMaterialsWereInstalled(self) -> None: 
```

cura/Machines/Models/MaterialManagementModel.py
```diff
@@ -21,13 +23,72 @@
...
+        a message prompting the user to sync the materials with their printers.
+        """
+        application = cura.CuraApplication.CuraApplication.getInstance()
+        for package_id, package_data in application.getPackageManager().installed_packages.items():
+            if package_data["package_info"]["package_type"] == "material":
+                # At least one new material was installed
+                self._showSyncNewMaterialsMessage()
+                break
+
+    def _showSyncNewMaterialsMessage(self):
```
nallath, line 50:
```suggestion
    def _showSyncNewMaterialsMessage(self) -> None:
```

cura/Machines/Models/MaterialManagementModel.py
```diff
@@ -21,13 +23,72 @@
...
+        )
+        sync_materials_message.actionTriggered.connect(self._onSyncMaterialsMessageActionTriggered)
+
+        # Show the message only if there are printers that support material export
+        container_registry = cura.CuraApplication.CuraApplication.getInstance().getContainerRegistry()
+        global_stacks = container_registry.findContainerStacks(type = "machine")
+        if any([stack.supportsMaterialExport for stack in global_stacks]):
+            sync_materials_message.show()
+
+    def _onSyncMaterialsMessageActionTriggered(self, sync_message: Optional[Message], sync_message_action: Optional[str]):
```
nallath, line 83:
```suggestion
    def _onSyncMaterialsMessageActionTriggered(self, sync_message: Message, sync_message_action: str) -> None:
```

These shouldn't be optional right? As far as I could see it should always pass a message and the action along.

nallath (changes requested):

nallath (approved):

PR #10588, opened 2021-10-08
Improve speed of the inherit button

nallath:
A few unneeded conversions were done. Also moved one check to python from JS

According to my profiler, this should make it about 2x as fast.


Ghostkeeper (approved):
Looks good.

I think the `hasOverrides` function being constant-time instead of linear makes the biggest impact. Good improvement!

PR #10632, opened 2021-10-18
Update fdmprinter.def.json.po

mlapkin:
Fix translation of 'order'.


Ghostkeeper:
Looks good! Apparently the original translator had it translated with the meaning of order as command/instruction rather than sorting/rank. Thanks for the correction!

PR #10634, opened 2021-10-18
Cura 8624 show df links in app switcher depending on account type

jellespijker:
In order to only show the DF links to a user in the App Switcher when they have actual access to them, the DF API needs to be queried. Currently the logic for this resides in the Digital Factory plugin. In order to avoid duplication or scope creep, I introduced an `additionalRights` property to the `API.account` this, dictionary can be updated by other plugins such as the DF plugin and allows us to easily query that information in location where it's logical. It is intended to allow us some flexibility, without breaking the API in the future.

The Application Switcher now queries the additional account rights, which is updated by the DF plugin to only show `My printers`,
`Digital Library` and `Print jobs` when the user has access to the DF.

### DIFF
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 from datetime import datetime
-from typing import Optional, Dict, TYPE_CHECKING, Callable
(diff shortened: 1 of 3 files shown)


jellespijker:
I was indeed considering `accessDFRequired` as an alternative. But you're right in the sense that this  will most certainly be reimplemented in a later ticket.

PR #10650, opened 2021-10-21
Update Eryone thinker and er20 3d printer profile

Eryone:
Update Eryone thinker 3d printer profile
Disabled thinker 3d printer first layer cooling fan
Disabled er20  3d printer first layer cooling fan


resources/quality/eryone_thinker/eryone_thinker_fast.inst.cfg
```diff
@@ -1,33 +0,0 @@
-[general]
```
nallath, line 1: Deleting a file requires an upgrade script; otherwise the configuration would become invalid.

Eryone: Hi, this file was removed by mistake. (It has had a conflict problem before so we removed it then it works. Now we updated it)
Thanks!

dapostol73: Now that the delete has be resolve, anything blocking this merge?

Ghostkeeper (approved):

Ghostkeeper:
Looks good now. Thanks for the contribution!

PR #10668, opened 2021-10-24
Update creality_cr-x.def.json

bob-anthony:
Corrected the CR-X definition file to inherit from creality_base and removed overrides that are no longer needed.

This will allow commonly used quality profiles for other Creality printers such as the Ender 3 to be used. Unfortunately, profiles previously created based on the draft profile will no longer be available.


Ghostkeeper:
This requires a version upgrade to preserve users' profiles when going to the new version. Custom profiles that are made for the Creality CR-X need to change their base (quality) profile.

Ghostkeeper:
It is automated for the user, and partially for the programmer. However Cura still needs to know how to recognise old profiles to upgrade and how to map old profiles to the new profiles. So it does require intervention on your end.

An example of such an upgrade script is this, which did all the changes between Cura 4.11 and 4.12: https://github.com/Ultimaker/Cura/blob/master/plugins/VersionUpgrade/VersionUpgrade411to412/VersionUpgrade411to412.py

nallath:
Closing this due to inactivity. As mentioned before, an upgrade script will be needed before this can be accepted.

PR #10712, opened 2021-10-31
Fix undo and redo for support blocker

### LINKED ISSUE #10648: CTRL+Z history doesn't properly restore support blockers
### Application Version

4.10.0

### Platform

Linux Mint

### Printer

Anet A8

### Reproduction steps

1. Open a test cube STL. 
2. Select it
3. Press E and add support blockers randomly on one face
4. CTRL+Z #N times
5. CTRL+SHIFT+Z  #N times

### Actual results

Support Blockers appear randomly shifted in space

### Expected results

Support Blockers should reapper exactly where they were before being canceled by CTRL+Z

- [ ] Log file
- [ ] Project file

This works with any STL I have tried.

nallath:
The translate wasn't done via an operation, so it wouldn't go back to the same
state after an undo / redo

Since I also ran into this, I decided to fix this in my own time :) 

Fixes #10648

### DIFF
--- a/plugins/SupportEraser/SupportEraser.py
+++ b/plugins/SupportEraser/SupportEraser.py
@@ -6,6 +6,7 @@ from PyQt5.QtWidgets import QApplication
 
 from UM.Application import Application
 from UM.Math.Vector import Vector
+from UM.Operations.TranslateOperation import TranslateOperation
 from UM.Tool import Tool
 from UM.Event import Event, MouseEvent
 from UM.Mesh.MeshBuilder import MeshBuilder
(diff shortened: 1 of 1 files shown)


fieldOfView:
Sorry, my bad.

nallath:
> Sorry, my bad.

Ha, no worries. We didn't notice in our review / component test back then either.

jellespijker (approved):

jellespijker:
The code make sense but the result doesn't. The blockers are re-added again, but scaling goes wrong in between the re-do-steps as can be seen in the animated GIF.

I tried a combination of `scale -> add blocker -> scale -> add blocker`, undoing those with `ctrl+z` and redoing those with `ctrl+shift+z` as a result I end up with a bigger scaled model then it should be and blockers that aren't in the correct relative position to the model. Hanging in mid air.

Peek 2021-11-03 17-08

nallath:
Yeah, I just checked the behavior before this change; the combination of scale + support blocker + scale also caused the same issue there. So i guess that's a seperate issue.

PR #10737, opened 2021-11-04
Add Creality Sermoon D1 Profile to Ultimaker Cura

felixstif:
Hi there,

we work with several Sermoon D1 printers. And we would like to ask you to integrate this into Ultimaker Cura. The corresponding files can be found in the appendix. (Original printer settings from Creality Slicer)

Thanks

Felix

[Cura Sermoon D1 Mod.zip](https://github.com/Ultimaker/Cura/files/7473115/Cura.Sermoon.D1.Mod.zip)


Ghostkeeper:
We can't add profiles for printers without having that printer to test with. And even if I had it, I probably shouldn't spend company time on profiles for a 3rd party printer. We're not going to add this ourselves, but if the community brings forth some profiles we'll gladly merge them so that everyone with that printer can enjoy better print quality.

PR #10742, opened 2021-11-04
Undo PP-29: apply zigzag pattern to top/bottom layers. These layer le…

pkuiper-ultimaker:
…ad to more sagging of bridging patterns. We will undo the chance (for the 4.12 release).


nallath:
For the next time, this should have been based on the 4.12 branch. I will cherry-pick it to there.

PR #10784, opened 2021-11-10
Update atom2.def.json

lin-ycv:
optimize start and end gcode

### DIFF
--- a/resources/definitions/atom2.def.json
+++ b/resources/definitions/atom2.def.json
@@ -22,8 +22,8 @@
     "machine_heated_bed":     { "default_value": false },
     "machine_center_is_zero": { "default_value": true },
 
-    "machine_start_gcode":    { "default_value": "G21\nG90 \nM107\nG28\nG92 E0\nG1 F200 E3\nG92 E0" },
-    "machine_end_gcode":      { "default_value": "M104 S0\nG28\nG91\nG1 E-6 F300\nM84\nG90" },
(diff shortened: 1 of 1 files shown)


resources/definitions/atom2.def.json
```diff
@@ -22,8 +22,8 @@
     "machine_heated_bed":     { "default_value": false },
     "machine_center_is_zero": { "default_value": true },
 
-    "machine_start_gcode":    { "default_value": "G21\nG90 \nM107\nG28\nG92 E0\nG1 F200 E3\nG92 E0" },
-    "machine_end_gcode":      { "default_value": "M104 S0\nG28\nG91\nG1 E-6 F300\nM84\nG90" },
+    "machine_start_gcode":    { "default_value": "G21\nG90 \nM107\nG28\nG1 Y-110 Z15\nG0 Z{layer_height_0}\nG92 E0\nG1 F200 Y-100 E6\nG92 E0" },
+    "machine_end_gcode":      { "default_value": "G28\nG91\nG1 E-6 F300\nM104 S0\nG1 E-1000 F5000\nM84\nG90" },
```
nallath, line 26: Are you sure that -1000 is correct to retract here? That seems like a _very_ big retraction move. The largest that any other print does is -100, and the previous value was -6.

nallath:
For our reference: CURA-8729

resources/definitions/atom2.def.json
lin-ycv: yes, -1000 is the value we use, our experience is that there's some design flaw at the hotend which often causes the filaments to get stuck if left in the hotend after print. we're using a full unload to solve the issue.

Ghostkeeper (approved):

PR #10786, opened 2021-11-10
PP-55 reduce cross3d top layer bridge distance arachne

rijkvanmanen:
The skin orientation is optimized for the cross (3d) pattern to reduce the bridge distance. For more information see PP-55.

### DIFF
--- a/resources/definitions/ultimaker.def.json
+++ b/resources/definitions/ultimaker.def.json
@@ -40,6 +40,9 @@
         {
             "value": false,
             "enabled": false
+        },
+        "skin_angles": {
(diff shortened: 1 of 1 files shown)


nallath (changes requested):
The ticket has "top line direction needs to be changed to [20, 110] for the TPU profiles." as a solution, but these changes are for all material profiles.

pkuiper-ultimaker (approved):

pkuiper-ultimaker:
The change is basically for all materials that use the cross patterns. The top layer needs to align. At the moment only TPU uses this pattern by default. So technically the title is indeed not correct, but the change is. (unless [] for the other materials is not the good default).

PR #10787, opened 2021-11-10
Add Creality Sermoon D1 Profile to Ultimaker Cura

felixstif:
Hi there,

we work with several Sermoon D1 printers. And we would like to ask you to integrate this into Ultimaker Cura. (Original printer settings from Creality Slicer)

Thanks

Felix

### DIFF
--- a/resources/definitions/creality_sermoond1.def.json
+++ b/resources/definitions/creality_sermoond1.def.json
@@ -1,0 +1,32 @@
+{
+    "name": "Creality Sermoon D1",
+    "version": 2,
+    "inherits": "creality_base",
+    "overrides": {
+        "machine_name": { "default_value": "Creality Sermoon D1" },
+        "machine_width": { "default_value": 280 },
+        "machine_depth": { "default_value": 260 },
+        "machine_height": { "default_value": 310 },
+        "machine_head_polygon": { "default_value": [
+                [-26, 34],
+                [-26, -32],
+                [22, -32],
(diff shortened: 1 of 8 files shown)


resources/definitions/creality_sermoond1.def.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "Creality Sermoon D1",
+    "version": 2,
+    "inherits": "creality_base",
+    "overrides": {
+        "machine_name": { "default_value": "Creality Sermoon D1" },
+        "machine_width": { "default_value": 280 },
+        "machine_depth": { "default_value": 260 },
+        "machine_height": { "default_value": 310 },
+        "machine_head_polygon": { "default_value": [
```
Ghostkeeper, line 10: This setting doesn't exist (any more). That's what the automated tests are complaining about. I think you meant to just override `machine_head_with_fans_polygon` instead?

Ghostkeeper (changes requested):

resources/definitions/creality_sermoond1.def.json
felixstif: Thanks for the information. I created the configurations with the template of the Creality Slicer. However, this one was a different version. I hope this is correct now. :)

Ghostkeeper (approved):

Ghostkeeper:
Seems to work correctly now. Thanks for the contribution!

PR #10795, opened 2021-11-11
Create creality_cr100.def

bodyboarder2528:
my first attempt hope its correct


resources/definitions/creality_cr100.def
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Creality CR-100",
+    "version": 4,
```
Ghostkeeper, line 3: The tests indicate (correctly) that this version number is incorrect. It's a version number to indicate the format of this file. The current version is 2. But really, it's best to just leave this out and let it inherit from `creality_base`.
```suggestion
```

Ghostkeeper (changes requested):

resources/definitions/creality_cr100.def
bodyboarder2528: so do i delete that line?

Ghostkeeper: That is my suggested change, yes.

bodyboarder2528: done

Ghostkeeper:
Grr, I'm sorry. How to properly push an additional commit to a PR still kind of eludes me. I did these steps perfectly: https://tighten.co/blog/adding-commits-to-a-pull-request/ But it still shows all commits that were made on master as if they were to get merged too. These commits are already on master, and wouldn't get merged. My only change is this commit: https://github.com/Ultimaker/Cura/pull/10795/commits/37aa01da936feefcd1b8e3918906bbbae69a9560

Ghostkeeper:
Thanks for the contribution!

PR #10797, opened 2021-11-11
Update requirements.txt

Ladnessness:
potential fix for Mac duplication/arrange bug. Appears to work with no downsides that I have detected with this version for me.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,7 +29,7 @@ pywin32==301
 requests==2.22.0
 scipy==1.6.2
 sentry-sdk==0.13.5
-shapely[vectorized]==1.7.1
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
The actual change was already made here: https://github.com/Ultimaker/cura-build-environment/pull/112

Cura's build system uses that build environment. The requirements.txt is only for convenience for developers running from source at the moment, and to allow bots to warn us of security issues in specific versions of our dependencies.

Nonetheless we'd need to update this file sooner or later to make developers have the same version as the builds. Thanks for helping out with that!

Ghostkeeper (approved):

PR #10798, opened 2021-11-11
Ender5 Fixed Extruder hitting frame after print

Rakhmanov:
Using G28 to fix the orientation of homing & the speed after the print.


Ghostkeeper:
Thanks for the fix!

PR #10799, opened 2021-11-11
Pp 60 libArachne iteration3 settings

pkuiper-ultimaker:
These are the proposed settings by the print profile team for the beta 2 release of Arachne. 
Please check https://jira.ultimaker.com/browse/PP-60 for the details and discussions (and PP-43 and PP-30 for even more details).

This pull request changes the Arachne and thin walls settings of ALL machines (so in fdmprinter.def.json) and a few overwrites in ultimaker.def.json. It also removes warnings on the middle line threshold settings. 

We believe these settings will benefit all machines and thus the reputation of Cura as a good slice engine. We think it is good to release this in a beta phase, releasing it later will become more and more difficult.


resources/definitions/fdmprinter.def.json
```diff
@@ -1095,10 +1095,8 @@
                     "description": "The smallest line width, as a factor of the normal line width, below which it will choose to use fewer, but wider lines to fill the available space the wall needs to occupy. Reduce this setting to use more, thinner lines. Increase to use fewer, wider lines. Note that this applies -as if- the entire shape should be filled with wall, so the middle here refers to the middle of the object between two outer edges of the shape, even if there actually is fill or (other) skin in the print instead of wall.",
                     "type": "float",
                     "unit": "%",
-                    "default_value": 50,
+                    "default_value": 90,
                     "minimum_value": "1",
-                    "minimum_value_warning": "min_bead_width / line_width * 100",
```
BagelOrb, line 1100: I don't think this minimum warning should be removed.

resources/definitions/fdmprinter.def.json
```diff
@@ -1264,8 +1258,7 @@
                     "label": "Minimum Wall Line Width",
                     "description": "Width of the wall that will replace thin features (according to the Minimum Feature Size) of the model. If the Minimum Wall Line Width is thinner than the thickness of the feature, the wall will become as thick as the feature itself.",
                     "unit": "mm",
-                    "value": "wall_line_width_0 / 2",
-                    "default_value": "0.2",
```
BagelOrb, line 1268: Don't remove the default_value. Those are used by the engine when using the command line interface. Each setting always needs a default_value, even though it's immediately overwritten by the "value". (Is this still the case? Are we still enforcing this?)

BagelOrb (commented):
Just a couple of comments

resources/definitions/fdmprinter.def.json
fieldOfView: > Is this still the case? Are we still enforcing this?

Yes. Cura sends the evaluated value for each setting to CuraEngine. If no value string is specified, the value evaluates to the default value.

In short: all settings need a default_value.

resources/definitions/fdmprinter.def.json
```diff
@@ -1120,10 +1116,8 @@
                             "type": "float",
                             "unit": "%",
                             "default_value": 50,
-                            "value": "wall_transition_threshold",
+                            "value": "wall_transition_threshold * 8/9",
```
nallath, line 1119:
```suggestion
                            "value": "wall_transition_threshold * 8 / 9",
```
Nitpicking, operators should have white spaces around them.

resources/definitions/ultimaker_original.def.json
```diff
@@ -59,6 +59,9 @@
         },
         "machine_end_gcode": {
             "value": "'M104 S0 ;extruder heater off' + ('\\nM140 S0 ;heated bed heater off' if machine_heated_bed else '') + '\\nG91 ;relative positioning\\nG1 E-1 F300  ;retract the filament a bit before lifting the nozzle, to release some of the pressure\\nG1 Z+0.5 E-5 X-20 Y-20 F9000 ;move Z up a bit and retract filament even more\\nG28 X0 Y0 ;move X/Y to min endstops, so the head is out of the way\\nM84 ;steppers off\\nG90 ;absolute positioning'"
+        },
+               "infill_before_walls": {
```
nallath, line 63:
```suggestion
           "infill_before_walls": {
```

nallath (dismissed):

nallath:
Also; Please follow our commit guide. Don't add ticket numbers in the summary of the commit message. The ticket numbers should be put in the body.

resources/definitions/ultimaker3.def.json
```diff
@@ -143,19 +137,14 @@
         "raft_margin": { "value": "10" },
         "raft_surface_layers": { "value": "1" },
         "retraction_amount": { "value": "6.5" },
-        "retraction_combing": {"value": "'no_outer_surfaces'"},
```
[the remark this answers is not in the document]
pkuiper-ultimaker: This is a clean up action this setting has been moved to ultimaker.def.json so it is needed at this level anymore. Keeps the files simpler.

resources/definitions/fdmprinter.def.json
pkuiper-ultimaker: I behaved strangly (did not calculate the correct value). I think it had to do with the min_bead_width value calculation which uses this setting (recursive?). That is why I removed it. Is there another way to do this?

resources/definitions/fdmprinter.def.json
pkuiper-ultimaker: OK, I will undo the default_value

resources/definitions/fdmprinter.def.json
pkuiper-ultimaker: Will change

resources/definitions/fdmprinter.def.json
BagelOrb: You could just set it to 25%...

resources/definitions/ultimaker3.def.json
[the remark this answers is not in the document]
BagelOrb: Ok

resources/definitions/fdmprinter.def.json
pkuiper-ultimaker: I will undo the change

pkuiper-ultimaker: > You could just set it to 25%...

That is as random as any other number. The maximum warning level was also set at 80% in the past I believe and we past that number with the current default. So might as well not set a warning level at all in my opinion.

resources/definitions/fdmprinter.def.json
```diff
@@ -3231,7 +3225,7 @@
                     "description": "Extrusion width based correction factor on the equalized flow. The target flow (in mm³/s) is reduced by the percentual difference in line width multiplied by this factor. A factor above zero means that thin lines are adjusted to be even faster than simple flow equalization would speed them up to. Conversely, wider lines are slowed down more. A positive value can help to compensate for the width depended back pressure the previous layer exterts on the line.",
                     "type": "float",
                     "unit": "mm³/s",
-                    "default_value": 0.0,
```
BagelOrb, line 3234: I don't think we should be turning on equalization for every printer. Printers with Linear advance shouldn't need flow equalization. Although how they would achieve variable width is an open question.

jellespijker: @BagelOrb FYI your suggestion about using percentage (ratio) instead of mm**3/s is implemented in the Arachne branches.

resources/definitions/ultimaker.def.json
```diff
@@ -40,6 +40,24 @@
...
+            "value": "False"
+        },
+        "retraction_combing": {
+            "value": "'no_outer_surfaces'"
+        },
+        "skin_monotonic" : {
+            "value": true
+        },
+        "speed_equalize_flow_width_factor": {
+            "value": "1.1"
```
jellespijker, line 57: Is now a percentage should be:
```suggestion
            "value": "110.0"
```

jellespijker (approved):
I will change the value of the speed_equalization_flow_factor

pkuiper-ultimaker:
Thanks Jelle!

PR #10810, opened 2021-11-12
Add new Mingda printers model(X,Max,Pro)

jianshu-du:
Add the printers for Magician X, Magician Max and Magician Pro.

### DIFF
--- a/resources/definitions/bfb.def.json
+++ b/resources/definitions/bfb.def.json
@@ -4,7 +4,6 @@
     "inherits": "fdmprinter",
     "metadata": {
         "visible": true,
-        "author": "Ultimaker",
         "manufacturer": "BFB",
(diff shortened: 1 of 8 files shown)


eGit:
Also please add Ender 3 Max.

Ghostkeeper:
Sorry if you closed it because it took too long. It takes us quite a while to work through backlog of pull requests most of the time. We try to do one PR per developer per day, but it's always busy.

eGit:
Should I open a seperate issue for the Creality Ender 3 Max? Those printers are very popular and are on the market since at least Cura 4.10

Ghostkeeper:
If you've made profiles, please do.

Don't create a feature request to add it though. We can't implement it, so it's likely to be closed.

PR #10812, opened 2021-11-12
[CURA-8559] Add new-marketplace search

rburema:
Add a search-bar for the (new) Marketplace.

Works for both Plugins and Materials. Inactive (hidden) when dealing with the Manage Packages tab (the already installed ones).

### DIFF
--- a/plugins/Marketplace/resources/qml/ManagePackagesButton.qml
+++ b/plugins/Marketplace/resources/qml/ManagePackagesButton.qml
@@ -18,7 +18,7 @@ Button
     background: Rectangle
     {
         color: backgroundColor
-        border.color: transparent
+        border.color: "transparent"
         radius: Math.round(width * 0.5)
 
     }
--- a/plugins/Marketplace/resources/qml/ManagePackagesButton.qml
+++ b/plugins/Marketplace/resources/qml/ManagePackagesButton.qml
@@ -18,9 +18,7 @@ Button
     background: Rectangle
     {
         color: backgroundColor
-        border.color: "transparent"
         radius: Math.round(width * 0.5)
-
     }
 
     Cura.ToolTip
--- a/resources/qml/SearchBar.qml
+++ b/resources/qml/SearchBar.qml
@@ -15,6 +15,7 @@ Cura.TextField
     leftPadding: searchIcon.width + UM.Theme.getSize("default_margin").width * 2
 
     placeholderText: catalog.i18nc("@placeholder", "Search")
+    font.italic: true
 
     UM.RecolorImage
     {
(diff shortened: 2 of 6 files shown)


resources/qml/SearchBar.qml
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2021 Ultimaker B.V.
+// Cura is released under the terms of the LGPLv3 or higher.
```
nallath, line 2: This looks like a re-usable component. Shouldn't we also make sure that this component is used in other places where we have a search bar (settings, digital factory)

plugins/Marketplace/resources/qml/ManagePackagesButton.qml
```diff
@@ -18,7 +18,7 @@ Button
     background: Rectangle
     {
         color: backgroundColor
-        border.color: transparent
+        border.color: "transparent"
```
nallath, line 21: So like no border? Why not just remove the border.color all together?

nallath (changes requested):

resources/qml/SearchBar.qml
bremco: Yes that was the intention. See the changes to '26 plugins/DigitalLibrary/resources/qml/SelectProjectPage.qml' in the 1st commit. But I'll try to add it to the settings tomorrow as well :-)

plugins/Marketplace/resources/qml/ManagePackagesButton.qml
bremco: I assumed that the one introducing this had some sort of reason originally. I'll remove it to see what happens tomorrow.

jianshu-du:
@nallath hello，I want to add new printers, but now it shows "1 workflow awaiting approval". Can you help me please.

plugins/Marketplace/resources/qml/ManagePackagesButton.qml
jellespijker: I think I introduced it with some ctrl-c, ctrl-v action

rburema: Ah, so it must've been an actual constant/variable originally.

resources/qml/SearchBar.qml
rburema: Done in https://github.com/Ultimaker/Cura/pull/10812/commits/87f94e680bf0da3f96d03e2a72c6848cf7451c3c

plugins/Marketplace/resources/qml/ManagePackagesButton.qml
rburema: In any case, fixed in https://github.com/Ultimaker/Cura/pull/10812/commits/6df6dab6f01db427a9b73f1012a2d5f28fd9c3fb

nallath (approved):

PR #10814, opened 2021-11-12
Flow equalization width factor as percentage

jellespijker:
This variable makes more sense to as a Ratio.
This was discussed during the eCCB 12/11/21
and implemented as a 5 min fix.

Note: The current default values were 0.5 there
is a new PR which change these 1.0. These values
should be multiplied with a 100

See also PR Ultimaker/CuraEngine#1529


Ghostkeeper (approved):

BagelOrb:
The setting description still mentions mm3/s.

I suggest the following:

"label": "Flow Equalization Ratio"
"description": "Extrusion width based correction factor on the speed. At 0% the movement speed is kept constant at the Print Speed. At 100% the movement speed is adjusted so that the flow (in mm³/s) is kept constant, i.e. lines half the normal Line Width are printed twice as fast and lines twice as wide are printed half as fast. A value larger than 100% can help to compensate for the higher pressure required to extrude wide lines."

jellespijker:
Darn, I was under the impression I changed it. Will create a quick fix using your proposed description

BagelOrb:
That was a quick reply!!

Awesome! Thanks!

jellespijker:
Done check c43ff0707262c86681f585c1c6e60496a70a51ca
I used the GitHub code editor for this, I accidentally found out today that pressing a `.` when you have a file open in GH presents you with the online editor, and this was a good excuse as any to try that.

PR #10831, opened 2021-11-15
Modified Cubicon.

harry572:
Add prime / optimus device.
Modified old devices.
Add meterial profile.


nallath (changes requested):
Materials should not be added to this repository. Instead, they should be added to https://github.com/ultimaker/fdm_materials/

I'm also not quite sure what the benefit of adding these materials is. If you want specific values for a material for a given printer, it's better to use the built in machine tags and add them to generic materials.

PR #10849, opened 2021-11-16
Fix initial layer bed and print head temperature for Snapmaker profile

prueker:
Fixing the initial print temperature for print head and bed for the Snapmaker profile so that users can set the "Printing Temperature Initial Layer" and "Build Plate Temperature Initial Layer" properties and they actually do something.

### DIFF
--- a/resources/definitions/snapmaker2.def.json
+++ b/resources/definitions/snapmaker2.def.json
@@ -26,7 +26,7 @@
             "default_value": true
         },
         "machine_start_gcode": {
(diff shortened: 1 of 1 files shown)


tyeth (commented):
This looks good to me, currently experiencing this problem

Ghostkeeper (approved):

Ghostkeeper:
Indeed, looks good. Thanks for amending this oversight!

PR #10854, opened 2021-11-17
[CURA-8539] HTTP Request Manager Login

rburema:
I think someone forgot to make a pull, since the ticket is in review. Will make a t least a draft PR here.

> The OAuth flow is currently the only place where we still use requests. In all other places we have refactored the code to use HttpRequestManager instead.
We have found the following issues with requests;
Certain companies use ZScaler. This seems to install custom certificates on the machine, which isn't handled correctly by Requests. Qt does handle this correctly
Requests does not support OCSP stapling, Qt does (needed for CURA-8402)
Furthermore, it's also just easier to not have to work with two different network libraries. Do note that our current network implementation does not have blocking network requests and requests does. Although it should make certain things run a bit smoother, it might cause issues when refactoring this.


Ghostkeeper:
Superseded https://github.com/Ultimaker/Cura/pull/10901.

PR #10857, opened 2021-11-17
moved and renamed icons

lorenzoromagnoli:
Those icons are actually of the default size, so I moved them to the default folder. 
I also made this PR to Uranium https://github.com/Ultimaker/Uranium/pull/752 to make sure the right icons are referenced in the notifications


fieldOfView:
You have renamed the icons as "badges", but they are not badges unless they are placed in a circle (right)? The cloud badge (used for printer connection status) comes with a circle of its own. I hate this inconsistency, which means that badges will have to be drawn fundamentally different depending on their content.

Also renaming and moving icons without further measures breaks plugins that try to do the right thing and use the icons introduced in 4.11.

lorenzoromagnoli:
Fair points!
I quickly checked with @nallath who agreed with you and suggested the following:

- Let's add the needed icons with the bg to the low resolution folder and use those for the PR
- Ignore my PR's for now and eventually improve the reusable component for badges and the icon names with the next api change. 

I'm attaching the right icons here so that you can add them to PR.   
[icons.zip](https://github.com/Ultimaker/Cura/files/7561341/icons.zip)

Thanks for taking the time to make this happen :)

PR #10858, opened 2021-11-17
PP-25 close to 100% density infill behaviour

rijkvanmanen:


resources/definitions/ultimaker.def.json
```diff
@@ -43,6 +43,12 @@
         },
         "skin_angles": {
             "value": "[] if infill_pattern not in ['cross', 'cross_3d'] else [20, 110]"
+        },
+        "top_layers": {
+            "value": "math.ceil(round(top_thickness / resolveOrValue('layer_height'), 4))"
```
pkuiper-ultimaker, line 48: Strange formula. The rounding does not seem necessary. Apparently this is also how it is implemented in the fdmprinters.def so I guess it is OK.

pkuiper-ultimaker (approved):

resources/definitions/ultimaker.def.json
nallath: Why wouldn't this formula be correct? It needs to be an integer.

pkuiper-ultimaker: Sure, but why round first and then make it an integer.

Ghostkeeper: Yeah, the `round` call seems to do nothing in this case.

nallath (approved):

PR #10859, opened 2021-11-17
Pp 25 close to 100% density infill behaviour arachne

rijkvanmanen:


rijkvanmanen:
this one should be merged into arachne, not master

PR #10901, opened 2021-11-22
OAuth via HttpRequestManager

Ghostkeeper:
The OAuth log-in flow was using the `requests` library. This was a problem for us for three reasons:
* The `requests` library doesn't properly use the desktop integrations of network requests, like the OS-provided local SSL certificate roots.
* The `requests` library is blocking, including blocking the main thread at start-up.
* The `requests` library is another attack vector for security issues. While there are no known issues with the current version, using a single library for all network requests reduces the attack surface.

This is a series of refactors to change that to use the HttpRequestManager, which uses the QNetworkManager under water. This does properly use local certificate roots and is not blocking. With this, we are no longer impacted by security issues in `requests` if any arise.

We cannot drop `requests` from our build environment until a major SDK increment. This means that plug-ins could still be using `requests` and be prone to not using those SSL certificate roots or security issues in `requests`.

The refactors below mostly have to do with becoming non-blocking. The HttpRequestManager is not blocking, and furthermore cannot be used in a blocking way unless it is certain that it is called from a different thread than the main thread. This is because the HttpRequestManager always executes some of its work on the main (Qt) thread due to usage of QTimers. If a function blocks the main thread while waiting for a request to finish from the HttpRequestManager, this results in a deadlock because the HttpRequestManager needs the Qt (main) thread to process its requests.
As a result of becoming asynchronous, the nested calls become a bit of a callback hell. Any feedback on how this can be improved is welcome.

Implements CURA-8539.

### DIFF
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,15 +1,15 @@
-# Copyright (c) 2018 Ultimaker B.V.
+# Copyright (c) 2021 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
-from datetime import datetime
-from typing import Any, Optional, Dict, TYPE_CHECKING, Callable
 
+from datetime import datetime
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty, QTimer, Q_ENUMS
+from typing import Any, Optional, Dict, TYPE_CHECKING, Callable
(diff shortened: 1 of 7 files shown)


jellespijker (approved):
Some minor remarks and typing issues.
Most of them I have fixed myself.

PR #10925, opened 2021-11-25
New printer settings, Trimaker Nebula Plus

MatiGabaglio:


resources/definitions/trimaker_nebula_plus.def.json
```diff
@@ -0,0 +1,249 @@
...
+            "Vertex_Delta_ABS",
+            "Vertex_Delta_PET",
+            "Vertex_Delta_PLA",
+            "Vertex_Delta_PLA_Glitter",
+            "Vertex_Delta_PLA_Mat",
+            "Vertex_Delta_PLA_Satin",
+            "Vertex_Delta_PLA_Wood",
+            "Vertex_Delta_TPU",
+            "zyyx_pro_flex",
+            "zyyx_pro_pla"
```
Ghostkeeper, line 151: These are a lot of excluded materials. Have you tested all of these, and found that they don't work? We consider blindly excluding third party materials to be quite anti-competitive so please only exclude materials that are provably infeasible.

And I wonder how you tested them, since some of them are 1.75mm diameter and some of them are 2.85mm. Materials that don't have the correct filament diameter are automatically excluded by Cura and don't need to be in this list.

Ghostkeeper (changes requested):

Ghostkeeper:
Failing tests are due to the branch being outdated w.r.t. the build environment and not the fault of the PR itself.

MatiGabaglio:
Hello, I just want to check if any other changes are needed or if the file is ready to merge. Thank you.

Ghostkeeper:
Sorry. We're not allowed to process pull requests at the moment due to a crunch for the 5.0 release. We'll catch up with the backlog shortly.

Ghostkeeper (approved):

Ghostkeeper:
I fixed a syntax error in the JSON file for the Nebula Plus. There was an extra comma there at the end which is not allowed in JSON and caused the file to give an error. Fixed here: https://github.com/Ultimaker/Cura/commit/0993e486f1bbb5e7953c60fb66bcd9b5b2860a79
Other than that, it looks like it works fine. Thanks for the contribution!

PR #10948, opened 2021-11-28
Cura 8564 add onboarding banners with link to explain cura mp

casperlamboo:
Add onboarding banners to the market place.

Fixes  [CURA-8541]

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -2015,33 +2015,6 @@ class CuraApplication(QtApplication):
         show_whatsnew_only = has_active_machine and has_app_just_upgraded
         return show_whatsnew_only
 
-    @pyqtSlot(result = bool)
-    def shouldShowMarketPlacePluginBanner(self) -> bool:
-        return self._preferences.getValue("cura/market_place_show_plugin_banner")
-
-    @pyqtSlot(result = bool)
-    def shouldShowMarketPlaceMaterialBanner(self) -> bool:
-        return self._preferences.getValue("cura/market_place_show_material_banner")
-
-    @pyqtSlot(result = bool)
-    def shouldShowMarketPlaceManagePackagesBanner(self) -> bool:
-        return self._preferences.getValue("cura/market_place_show_manage_packages_banner")
-
-    @pyqtSlot()
-    def closeMarketPlacePluginBanner(self) -> None:
-        Logger.log("i", "Close market place plugin banner")
-        self._preferences.setValue("cura/market_place_show_plugin_banner", False)
-
-    @pyqtSlot()
-    def closeMarketPlaceMaterialBanner(self) -> None:
-        Logger.log("i", "Close market place material banner")
-        self._preferences.setValue("cura/market_place_show_material_banner", False)
-
-    @pyqtSlot()
-    def closeMarketPlaceManagePackagesBanner(self) -> None:
-        Logger.log("i", "Close market place manage packages banner")
-        self._preferences.setValue("cura/market_place_show_manage_packages_banner", False)
-
     @pyqtSlot(result = int)
     def appWidth(self) -> int:
         main_window = QtApplication.getInstance().getMainWindow()
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -572,6 +572,10 @@ class CuraApplication(QtApplication):
 
         preferences.addPreference("general/accepted_user_agreement", False)
 
+        preferences.addPreference("cura/market_place_show_plugin_banner", True)
+        preferences.addPreference("cura/market_place_show_material_banner", True)
+        preferences.addPreference("cura/market_place_show_manage_packages_banner", True)
+
         for key in [
             "dialog_load_path",  # dialog_save_path is in LocalFileOutputDevicePlugin
             "dialog_profile_path",
@@ -2010,6 +2014,33 @@ class CuraApplication(QtApplication):
         # Only show the what's new dialog if there's no machine and we have just upgraded
         show_whatsnew_only = has_active_machine and has_app_just_upgraded
         return show_whatsnew_only
+
+    @pyqtSlot(result = bool)
+    def shouldShowMarketPlacePluginBanner(self) -> bool:
+        return self._preferences.getValue("cura/market_place_show_plugin_banner")
+
+    @pyqtSlot(result = bool)
+    def shouldShowMarketPlaceMaterialBanner(self) -> bool:
+        return self._preferences.getValue("cura/market_place_show_material_banner")
+
+    @pyqtSlot(result = bool)
+    def shouldShowMarketPlaceManagePackagesBanner(self) -> bool:
+        return self._preferences.getValue("cura/market_place_show_manage_packages_banner")
+
+    @pyqtSlot()
+    def closeMarketPlacePluginBanner(self) -> None:
+        Logger.log("i", "Close market place plugin banner")
+        self._preferences.setValue("cura/market_place_show_plugin_banner", False)
+
+    @pyqtSlot()
+    def closeMarketPlaceMaterialBanner(self) -> None:
+        Logger.log("i", "Close market place material banner")
+        self._preferences.setValue("cura/market_place_show_material_banner", False)
+
+    @pyqtSlot()
+    def closeMarketPlaceManagePackagesBanner(self) -> None:
+        Logger.log("i", "Close market place manage packages banner")
+        self._preferences.setValue("cura/market_place_show_manage_packages_banner", False)
 
     @pyqtSlot(result = int)
     def appWidth(self) -> int:
--- a/plugins/Marketplace/resources/qml/OnboardBanner.qml
+++ b/plugins/Marketplace/resources/qml/OnboardBanner.qml
@@ -15,6 +15,7 @@ Rectangle
     property string bannerIcon
     property string bannerBody
     property var onRemoveBanner
+    property string readMoreUrl
 
     visible: bannerVisible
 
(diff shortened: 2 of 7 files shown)


plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,79 @@
...
+// Onboarding banner.
+Rectangle
+{
+    Layout.preferredHeight: childrenRect.height + 2 * UM.Theme.getSize("default_margin").height
+    anchors
+    {
+        margins: UM.Theme.getSize("default_margin").width
+        left: parent.left
+        right: parent.right
+        top: parent.top
```
nallath, line 28: This doesn't need to be anchored to the top, it's in a column layout. Because it's in a column layout, you're not allowed to anchor anything to the bottom or the top!

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    color: UM.Theme.getColor("action_panel_secondary")
+
+    // Icon
+    Rectangle
```
nallath, line 34: If an rectangle doesn't have a background or a border, it's better to use Item (since that doesn't need to be checked if it needs to be drawn)

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,138 @@
...
+import UM 1.6 as UM
+import Cura 1.6 as Cura
+
+// Onboarding banner.
+Rectangle
+{
+    property bool bannerVisible
+    property string bannerIcon
+    property string bannerBody
+    property var onRemoveBanner
```
nallath, line 17: This should be a signal. That way you can just emit the "onBannerHidden" signal and connect any number functions to it from the outside. 

I would also put the visibility toggle in the banner itself (so the outside doesn't have to remember to actually hide it)

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,138 @@
...
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.1
+
+import UM 1.6 as UM
+import Cura 1.6 as Cura
+
+// Onboarding banner.
+Rectangle
+{
+    property bool bannerVisible
```
nallath, line 14: Why duplicate the property here? Visible is already a property of the rectangle.

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,138 @@
...
+    property string bannerIcon
```
nallath, line 15: This should be an alias instead of a string. See https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html#property-aliases

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,138 @@
...
+                    // If it does place it after the body text
+                    readMoreButton.anchors.left = infoText.left;
+                    readMoreButton.anchors.bottom = infoText.bottom;
+                    readMoreButton.anchors.bottomMargin = -(fontMetrics.height + UM.Theme.getSize("thin_margin").height);
+                    readMoreButton.anchors.leftMargin = 0;
+                }
+                else
+                {
+                    // Otherwise place it under the text
+                    readMoreButton.anchors.left = infoText.left;
```
nallath, line 108: In both cases of the if statement, the anchor is to the left. SO there is no need to set it in the onLineLaidOut. Same holds for the infotext.bottom

plugins/Marketplace/resources/qml/OnboardBanner.qml
```diff
@@ -0,0 +1,138 @@
...
+            left: onboardingIcon.right
+            right: onboardingClose.left
+            margins: UM.Theme.getSize("default_margin").width
+        }
+
+        font: UM.Theme.getFont("default")
+        text: bannerBody
+
+        renderType: Text.NativeRendering
+        color: "white"
```
nallath, line 88: Hardcoded value, needs to be pulled from the theme

cura/CuraApplication.py
```diff
@@ -2011,6 +2015,33 @@ def shouldShowWhatsNewDialog(self) -> bool:
         show_whatsnew_only = has_active_machine and has_app_just_upgraded
         return show_whatsnew_only
 
+    @pyqtSlot(result = bool)
+    def shouldShowMarketPlacePluginBanner(self) -> bool:
+        return self._preferences.getValue("cura/market_place_show_plugin_banner")
+
+    @pyqtSlot(result = bool)
```
nallath, line 2022: I think it's better to just use the direct bindings that we have for preferences in qml. You can gte the value like this:
UM.Preferences.getValue("cura/market_place_show_material_banner")

The same also goes for the hide / setting

plugins/Marketplace/resources/qml/Packages.qml
```diff
@@ -11,6 +11,12 @@ ListView
     id: packages
 
     property string pageTitle
+    property bool bannerVisible
+    property string bannerIcon
+    property string bannerBody
```
nallath, line 16: It's not quite the body right? It contains the text of the banner. As such i think it's better to call it "bannerText"

nallath (changes requested):

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: > I would also put the visibility toggle in the banner itself (so the outside doesn't have to remember to actually hide it)

I tried that before. Unfortunately this did not result in the desired result. There is a single banner element for al three pages. If you hide a banner for one page, it should not be hidden from the remaining pages. Hiding the banner this way removes the banner element in its entirety; it also becomes hidden from the other pages.

Is there a way to "reset" the properties if the page changes?

cura/CuraApplication.py
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/e5c8d5456de7e288905ccb92d7673ac6f13590ff

plugins/Marketplace/resources/qml/Packages.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/00135e574dc464f1562c277ea389511dbc6735fb

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/f8d9f1e39c98546a31705d4b6fe489427f750810

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/f1a9bbd7914b3dbdce17e0e2ec4d02b779c6acf1

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/6909085733a46bda1e99fe36c2eaba07ae6ed32d

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/b53c1ba3061b4fc3dbeb46e161f7f186fa399ceb

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/0ca20e9b3d0f6716cb3fff68442661375bc712e3

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: https://github.com/Ultimaker/Cura/pull/10948/commits/c136deb430b1a0f9a82195eea28e6fbc0bcf2c3e

plugins/Marketplace/resources/qml/OnboardBanner.qml
casperlamboo: I also had a bit of difficulties using a signal as you suggested to remove the on boarding banners. This is due to the multiple pages where the banners are located; as there are 3 pages there are 3 places where we listen to the `onBannerHidden` signal. As all these the handlers for all these 3 pages are listening to the signal at the same time all 3 banners are removed by clicking on a single close button. There are several options for resolving this:
- create 3 different signals, a separate signal for each banner
- add a check within the signal listener to check of the signal is emitted from the right page.
I think these stated options add unnecessary complexity. Of course there is probably a very obvious solution, but as I'm new to Qt I'm just missing it. I gladly hear suggestions.

rburema: Signals can have parameters, so you could also use that.

I tried something as well: Just saving the preference (location) string and moving the function to the OnboardBanner base class. I almost had that working but each time something else stopped working, so I decided that it was outside of the scope of this review & decided to accept the code as-is for now.

rburema (approved):

PR #10968, opened 2021-11-30
Detail page for plug-ins on Marketplace

Ghostkeeper:
This redesigns the plug-in page of the Marketplace. The new design looks like this:
image

Additional features that can't be seen from that screenshot:
* The new page is activated by clicking on a plug-in.
* All ActionButton icons have been increased in side, by request of Lorenzoromagnoli.
* A hover colour is added to the package cards in the package list.
* HTTP(S) links are automatically detected in the description and shown as links the user can click on.
* Fixed alignment of the Read More and "..." ellipses when the short abbreviation is cut off.
* The vertical contents of the old card is now in a Column again rather than anchors, to aid with the positioning and display of the existing items now that they may disappear when replaced with the download count.
* The download count is not shown for bundled plug-ins.

This implements CURA-8565.
Currently the detail page for materials is the same. This is a different ticket (CURA-8585).
There are no action buttons on the detail pages yet. This is also a different ticket (CURA-8587).


rburema (approved):
Good work!

Not entirely happy with the "installed"/"not_installed"/"bundled" being plain strings everywhere (instead of enums referring to strings), but it's more annoying to fix than to live with it as is ;-)

PR #10971, opened 2021-11-30
Add a thumbnail to 3mf files (projects & exports)

### LINKED ISSUE #10913: Add thumbnail to 3MF output format

3MF format supports including a thumbnail, so it would be great if the files created by Cura would include that thumbnail that can be easily read in Windows file explorer (and possibly other OS) 

Add a thumbnail of the model to the 3MF output file

Adding the GCode postprocessing script to add thumbnails, however those are not visible in Windows file explorer, only on specific files.

Every Windows user will benefit from this (and possibly other users too)

_No response_

fieldOfView:
This PR adds a thumbnail to project files (edit: and exported files), like they are added to UFP files.

This PR fixes #10913

### DIFF
--- a/plugins/3MFWriter/ThreeMFWorkspaceWriter.py
+++ b/plugins/3MFWriter/ThreeMFWorkspaceWriter.py
@@ -5,8 +5,6 @@ import configparser
 from io import StringIO
 import zipfile
 
-from PyQt5.QtCore import QBuffer
-
 from UM.Application import Application
 from UM.Logger import Logger
 from UM.Preferences import Preferences
@@ -16,7 +14,7 @@ from UM.i18n import i18nCatalog
 catalog = i18nCatalog("cura")
 
 from cura.Utils.Threading import call_on_qt_thread
-from cura.Snapshot import Snapshot
+
 
 class ThreeMFWorkspaceWriter(WorkspaceWriter):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


fieldOfView:
Hmm, or should I move this to ThreeMFWriter instead, so non-project 3MF files also get a thumbnail?

pedrolamas:
That would make more sense to me, as one can also use the `File -> Export` to generate a 3MF file!

Isn't it also required to update the manifest file?

nallath:
> Isn't it also required to update the manifest file?

Yeah. I think so. This is the relevant spec: https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md#61-package-thumbnail-and-object-thumbnail

"The package thumbnail has a relationship from the package root to the thumbnail image, object thumbnails have a relationship from the 3D Model part to the thumbnail image."

> Hmm, or should I move this to ThreeMFWriter instead, so non-project 3MF files also get a thumbnail?

I think that makes more sense. Shouldn't be that hard to change right?

fieldOfView:
Thumbnails are now added to both projects and exported files. I have also added the necessary relation and content-type metadata.

pedrolamas:
I just tested these changes by directly patching my local Cura 4.12.1, and they worked great! 🙂

nallath (approved):

nallath:
It needed a bit more love to get it to work on linux, but I fixed that in 064a845f717b91facf159e4da76234487637148e

PR #10973, opened 2021-11-30
Detail page for materials on Marketplace

Ghostkeeper:
This adds a design for the detail page to specialise for materials. The new design looks like this:
image

This PR depends on another PR: https://github.com/Ultimaker/Cura/pull/10968
Merging it would merge both. Preferably, merge the other one first.

This implements CURA-8585.


rburema (approved):
Good work!

Fixed a few small issues myself. Otherwise I'm kind of on the fence about the UM PVA & Breakaway being hard coded, but it's probably not worth it thinking up a way around it for the benefit that gives.

PR #10978, opened 2021-12-01
Add new Dagoma PRO 430 printer

aschwenck:
Hello everyone,

As a Dagoma employee, I am glad to propose a brand new 3d printer to be added to Ultimaker Cura!
Please let me know if there is something wrong (I've already checked the settings_version ;))


0r31:
Thanks @aschwenck !

We might have some new materials profiles to add. As far as we understand, we have to add them to [fdm_materials](https://github.com/Ultimaker/fdm_materials) right?

Regards,
Orel

0r31:
Hello !

Any update on this one ?

Regards,
Orel

resources/definitions/dagoma_pro_430_base.def.json
```diff
@@ -0,0 +1,207 @@
...
+            "fabtotum_pla",
+            "fabtotum_tpu",
+            "fdplast_abs_tomato",
+            "fdplast_petg_gray",
+            "fdplast_pla_olive",
+            "fiberlogy_hd_pla",
+            "filo3d_pla",
+            "filo3d_pla_green",
+            "filo3d_pla_red",
+            "generic_abs",
```
nallath, line 77: Why exclude _all_ generic materials? This might be a bit confusing, since we generally set the profiles to be used with the generic ones. For instance, take a look at the way that UM printers do this. We set the profiles for the generic materials. If the user selects ultimaker PLA, the profiles for generic are used.

The advantage of using this is that if a user does decide to print with a custom material profile, they will still get the default quality profiles.

resources/definitions/dagoma_pro_430_base.def.json
```diff
@@ -0,0 +1,207 @@
...
+            "generic_abs_175",
+            "generic_asa_175",
+            "generic_bam",
```
nallath, line 80: There is no need to exclude the 2.85 mm materials, since your printer is 1.75. These are excluded by default, so the DRY (Don't Repeat Yourself) principle should be observed here.

resources/quality/dagoma_pro_430_bowden/Steel nozzle 0.4/dagoma_pro_430_bowden_0.4_abs_high_quality.inst.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+version = 4
+name = High Quality
+definition = dagoma_pro_430_bowden
+
+[metadata]
+
+setting_version = 19
+type = quality
+quality_type = high_quality
```
nallath, line 11: Nitpicking; Why not just set it to be "high"? Adding the _quality seems a bit double.

resources/quality/dagoma_pro_430_bowden/Steel nozzle 0.4/dagoma_pro_430_bowden_0.4_abs_high_quality.inst.cfg
```diff
@@ -0,0 +1,58 @@
...
+weight = -1
+material = dagoma_generic_abs
```
nallath, line 13: I think it's better to just use generic_abs_175 here. See previous comment.

resources/quality/dagoma_pro_430_bowden/Steel nozzle 0.4/dagoma_pro_430_bowden_0.4_abs_high_quality.inst.cfg
```diff
@@ -0,0 +1,58 @@
...
+variant = 0.4
+
+[values]
+
+wall_thickness = =machine_nozzle_size * 2
```
nallath, line 18: Don't use machine_nozzle_size for this, since the line width can be different. So this should be =line_width * 2

Take care to change this for all other profiles as well.

resources/quality/dagoma_pro_430_bowden/Steel nozzle 0.4/dagoma_pro_430_bowden_0.4_abs_high_quality.inst.cfg
```diff
@@ -0,0 +1,58 @@
...
+optimize_wall_printing_order = True
+skin_no_small_gaps_heuristic = True
+fill_outline_gaps = True
```
nallath, line 21: This value is always true. So it shouldn't be set in the quality profile but in the definition. 

Note that there are more settings like this. Take care to set as many values in a profiles as deep down the stack as possible!

resources/quality/dagoma_pro_430_bowden/dagoma_pro_430_bowden_global_low_quality.inst.cfg
```diff
@@ -0,0 +1,28 @@
...
+global_quality = True
+
+[values]
+
+layer_height = 0.4
+layer_height_0 = =round(0.38 * machine_nozzle_size, 2)
+
+top_bottom_thickness = =3 * layer_height
+
+acceleration_enabled = True
```
nallath, line 22: All settings below this point should be in the definition, not in the quality.

nallath (changes requested):
The build platform is also too big (above 1 mb). Please reduce the size of the build plate, there isn't really a need for this much detail.

0r31:
Hi @nallath,

Thanks for the feedback!
I will shrink the platform size and let @aschwenck brings corrections regarding your comments.

Regards,
Orel

aschwenck:
Hello,

Can I get feedback on the new profiles

Thanks in advance

Alexandre

0r31:
Hello all,

Any feedback on this PR?
Are you satisfied enough with the modifications made according to previous review request?

Regards,
Orel

nallath:
Apologies! We've been busting our behinds off to meet the 5.0 deadline, so we've been slacking with PR's. I've carved out a bit of time to handle them.

nallath (approved):

0r31:
Hi @nallath ,

Many thanks for the feedback !

Regards,
Orel

PR #10979, opened 2021-12-01
Add a default value to spool cost if defined in fdm material profile

0r31:
Hello everyone,

This PR just adds a default value to the spool cost if defined in material profile.
It is just the very same behavior as for the spool weight.

Regards,
Orel

### DIFF
--- a/resources/qml/Preferences/Materials/MaterialsView.qml
+++ b/resources/qml/Preferences/Materials/MaterialsView.qml
@@ -261,7 +261,7 @@ TabView
                 {
                     id: spoolCostSpinBox
                     width: scrollView.columnWidth
-                    value: base.getMaterialPreferenceValue(properties.guid, "spool_cost")
+                    value: base.getMaterialPreferenceValue(properties.guid, "spool_cost", Cura.ContainerManager.getContainerMetaDataEntry(properties.container_id, "properties/cost"))
                     prefix: base.currency + " "
                     decimals: 2
                     maximumValue: 100000000


fieldOfView:
For this to be useful, wouldn't you also need the amount of material on a spool (weight) to be set in the XML? And, realistically but less strictly required, a currency?

0r31:
Hi @fieldOfView,

It seems to me the spool weight is already handled (https://github.com/Ultimaker/Cura/blob/26d6a02c32506869a7c46794ce6f26b2a8c0957f/resources/qml/Preferences/Materials/MaterialsView.qml#L281) and used (e.g. https://github.com/Ultimaker/fdm_materials/blob/304287c3a69d79a14d33c1fe897bf653a2fc4f15/ultimaker_pla_black.xml.fdm_material#L19).

For the currency, i agree that just setting a scalar value can be disturbing.
Currently, there is a default currency defined (euro).

Regards,

fieldOfView:
It seems so. I am the one who originally implemented spool weight and cost as preferences; this particular detail slipped my attention when it was added.

Personally I find it strange that the dialog would display fields from the xml file which it can not write; You have to manually edit xml files to use this.

The default currency can be edited in the General preferences. Currency issues as well as reseller markups and spools being available in different weights (eg 750 gr and 2200 gr) at different cost per spool were all reasons to originally not put cost and weight information in the material xml file. 

I would suggest as a minimum that you use `spool_cost` instead of `cost` in the xml file.

0r31:
> Personally I find it strange that the dialog would display fields from the xml file which it can not write; You have to manually edit xml files to use this.

It can be overriden (overriden values are set in cura.cfg).
Example :
Default values in xml:
image

UI:
image

Overrides at UI level:
image

In cura.cfg:
`material_settings = {"69fd1f24-f411-470e-8158-9e5552f36147":{"spool_weight":750,"spool_cost":40},"048209a4-25a4-11ec-9621-0242ac130002":{"spool_cost":20,"spool_weight":500},"1c733efc-25a4-11ec-9621-0242ac130002":{}}`

UI after Cura close / reopen steps:
image

> I would suggest as a minimum that you use `spool_cost` instead of `cost` in the xml file.
I've just followed what have been done for spool weight (_spool_weight_ as param in qml, _properties/weight_ in xml)

nallath:
I'm not a big fan of the weight being in there tbh. It implies that if you have a spool of 750 gram and a spool of 2 kg that these are _seperate_ profiles. That doesn't make any sense, since you wouldn't want different material settings if your spool has more material.

0r31:
Hi @nallath ,

I understand the concerns you and @fieldOfView have.
I've juste replicated to cost what has been done for weight.

If you think it is not relevant, you can close the PR.

Regards,
Orel

PR #11005, opened 2021-12-04
think3dim first add  some printer type file

peacerfu:
think3dim first add  some printer type file,include the configure \mesh \ extrudersfile


peacerfu:
i think i put too many file one time.so I will put 1 machine first

Ghostkeeper:
Generally if we see too many badly-written commits like this we'd just squash them and you get 1 commit for the entire PR. No need to worry about making too many changes if you're not so experienced with Git yet.

A more reliable way of working is probably to use Github Desktop instead of supplying the files via Github's web committing interface.

PR #11008, opened 2021-12-04
add zmorph i500 printer

michallesniak2:
I work for the Zmorph company. 
I added the definition files for our machine zmorph i500 to Cura.


michallesniak2:
I also have the files for the material definitions, but I don't know how to add them. 
Can you tell me how should I implement them.

resources/definitions/zmorph_i500.def.json
```diff
@@ -0,0 +1,168 @@
...
+        "machine_extruder_trains":
+        {
+            "0": "zmorph_i500_build_module",
+            "1": "zmorph_i500_support_module"
+        }
+    },
+
+    "overrides": {
+        "machine_name": { "default_value": "Zmorph i500" },
+        "machine_width": { "default_value": 440 },
```
Ghostkeeper, line 46: On the product page of this printer, it says 460: https://zmorph3d.com/products/zmorph-i500/

Was this difference intentional?

resources/definitions/zmorph_i500.def.json
```diff
@@ -0,0 +1,168 @@
...
+        "top_bottom_thickness": { "value": "1" },
+        "travel_avoid_supports": { "value": "True" },
+        "travel_avoid_distance": { "value": "3 if extruders_enabled_count > 1 else machine_nozzle_tip_outer_diameter / 2 * 1.5" },
+        "wall_0_inset": { "value": "0" },
+        "wall_line_width_x": { "value": "line_width" },
+        "wall_thickness": { "value": "0.8" },
+        "meshfix_maximum_resolution": { "value": "(speed_wall_0 + speed_wall_x) / 60" },
+        "meshfix_maximum_deviation": { "value": "layer_height / 4" },
+        "optimize_wall_printing_order": { "value": "True" },
+        "retraction_combing": { "default_value": "all" },
```
Ghostkeeper, line 163: This setting override is not working. It is overridden by the `value` defined in fdmprinter, which is a formula. This is a way to make it override properly:

```suggestion
        "retraction_combing": { "value": "'all'" },
```

Or you could consider adjusting the formula from fdmprinter to make it dependant on whether ironing or monotonic is used.

Ghostkeeper (changes requested):

PR #11015, opened 2021-12-05
Updated Profiles and Start and End G-Code to fix an issue with premature heatup

reibuehl:
Changed the start & end g-code to use {material_print_temperature} instead of the deprecated {print_temperature} to avoid a premature heatup due to Cura inserting a generic heatup section before the custom start g-code.
Fixed travel speed in start and end g-code to fixed values as {travel_speed} resulted in too slow movements due to mm/min vs. mm/s difference.
See discussion at https://community.ultimaker.com/topic/38849-problem-with-generated-g-code-pre-mature-heatup-g-code-inserted-before-startup-g-code/ for details.


resources/quality/tinyboy/tinyboy_fabrikator15_draft.inst.cfg
```diff
@@ -31,7 +31,7 @@ layer_height_0 = 0.3
 material_bed_temperature = 60
 material_diameter = 1.75
 material_print_temperature = 200
-material_print_temperature_layer_0 = 0
+material_print_temperature_layer_0 = 200
```
fieldOfView, line 34: If you just _remove_ the `material_print_temperature_layer_0` line, Cura will _automatically_ inherit the value from `material_print_temperature` (which I think you are after.

resources/definitions/tinyboy_fabrikator15.def.json
```diff
@@ -9,7 +9,8 @@
         "file_formats": "text/x-gcode",
         "platform": "tinyboy_fabrikator15.stl",
         "platform_offset": [-95, -3, -46],
-        "has_materials": false,
+        "has_materials": true,
+    	"preferred_material": "generic_pla_175",
```
fieldOfView, line 13: You may want to fix the indentation.

resources/quality/tinyboy/tinyboy_fabrikator15_normal.inst.cfg
```diff
@@ -30,8 +30,8 @@ layer_height = 0.2
 layer_height_0 = 0.2
 material_bed_temperature = 60
 material_diameter = 1.75
-material_print_temperature = 210
-material_print_temperature_layer_0 = 0
+material_print_temperature = 200
```
fieldOfView, line 33: This temperature is only going to work well for PLA. And since you have specified it in your quality profile, it will override whatever is set in the material itself.

reibuehl: Should I remove both lines from the profiles if I want to leave the temperature settings to the material definition?

resources/definitions/tinyboy_fabrikator15.def.json
reibuehl: Fixed

PR #11022, opened 2021-12-06
 Updated Profiles and Start and End G-Code to fix an issue with premature heatup

reibuehl:
Changed the start & end g-code to use {material_print_temperature} instead of the deprecated {print_temperature} to avoid a premature heatup due to Cura inserting a generic heatup section before the custom start g-code.
Fixed travel speed in start and end g-code to fixed values as {travel_speed} resulted in too slow movements due to mm/min vs. mm/s difference.
See discussion at https://community.ultimaker.com/topic/38849-problem-with-generated-g-code-pre-mature-heatup-g-code-inserted-before-startup-g-code/ for details.


rburema:
Thanks for your contribution!

For our reference: Internal ticket CURA-8837

PR #11023, opened 2021-12-06
Update anycubic_i3_mega_s.def.json

NilsRo:
Optimized initial layer temperature handling with Anycubic Mega S/X and added material information


NilsRo:
> These changes actually make the json files no longer valid. Are you sure that you tested this in Cura? 

Yes. But used Startup optimiser which suppressed reading the smaller configuration. I updated the file but checks still failing...

nallath:
Aah, I see already, you have a setting that doesn't exist in Arachne: travel_compensate_overlapping_walls_0_enabled

NilsRo:
> Aah, I see already, you have a setting that doesn't exist in Arachne: travel_compensate_overlapping_walls_0_enabled

Something to add on your side?

nallath:
> > Aah, I see already, you have a setting that doesn't exist in Arachne: travel_compensate_overlapping_walls_0_enabled
> 
> Something to add on your side?

No, something to remove on your side. The setting travel_compensate_overlapping_walls_0_enabled doesn't exist anymore. As such it's complaining that it found a setting that it doesn't recognize.

nallath:
For our reference: CURA-8971

nallath (approved):

PR #11027, opened 2021-12-07
Cura 8587 disable update install and uninstall

jellespijker:
See also the Uranium pull request Ultimaker/Uranium#756

We can now install, uninstall, update, disable and enable packages from the new marketplace.

Created a  ManageButton which can have a Primary, Secondary or busy state. The Primary state is the primary button, these are positive actions like: install, update, enable. The secondary state is the secondary button, these are subtracting actions, like uninstall and disable. Each of these states have a busy message with a corresponding busy message and spinner.

Materials can't be enabled or disabled, only plugins can. All types of packages can be installed or uninstalled. When Installing a package it will download the  package, present a license agreement and if accepted and the user is logged in subscribe the user to the packages. When uninstalling the plugin the user is unsubscribed if logged in.

Remote packages, both in the install plugin and install materials tab are filtered on packages which aren't already present locally. When going to the manage packages tab it will show the packages in the correct sections. Bundled packages can only be enabled, disabled or downgraded with an uninstall if the user installed an updated version of a bundled package.

Peek 2021-12-07 10-05

### DIFF
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -19,11 +19,23 @@ if TYPE_CHECKING:
 class CuraPackageManager(PackageManager):
     def __init__(self, application: "QtApplication", parent: Optional["QObject"] = None) -> None:
         super().__init__(application, parent)
+
         self._local_packages: Optional[Dict[str, Dict[str, Any]]] = None
+        self._local_packages_installed: Optional[Dict[str, Dict[str, Any]]] = None
+        self._local_packages_to_remove: Optional[Dict[str, Dict[str, Any]]] = None
+        self._local_packages_to_install: Optional[Dict[str, Dict[str, Any]]] = None
+
         self.installedPackagesChanged.connect(self._updateLocalPackages)
 
     def _updateLocalPackages(self) -> None:
-        self._local_packages = self.getAllLocalPackages()
+        self._local_packages_installed = dict([(package_info["package_id"], dict(package_info)) for package in self.getAllInstalledPackagesInfo().values() for package_info in package])
+        self._local_packages_to_remove = dict([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToRemove().values()])
+        self._local_packages_to_install = dict([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToInstall().values()])
+
+        self._local_packages = {}
+        self._local_packages.update(self._local_packages_installed)
+        self._local_packages.update(self._local_packages_to_remove)
+        self._local_packages.update(self._local_packages_to_install)
 
     @property
     def local_packages(self) -> List[Dict[str, Any]]:
@@ -42,6 +54,15 @@ class CuraPackageManager(PackageManager):
             # _updateLocalPackages always results in a list of packages, not None.
             # It's guaranteed to be a list now.
         return set(self._local_packages.keys())
+
+    @property
+    def installed_packages_ids(self) -> Set[str]:
+        """locally installed packages, lazy execution"""
+        if self._local_packages is None:
+            self._updateLocalPackages()
+            # _updateLocalPackages always results in a list of packages, not None.
+            # It's guaranteed to be a list now.
+        return set(self._local_packages_installed.keys())
 
     def initialize(self) -> None:
         self._installation_dirs_dict["materials"] = Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer)
@@ -72,12 +93,3 @@ class CuraPackageManager(PackageManager):
                         machine_with_qualities.append((global_stack, str(extruder_nr), container_id))
 
         return machine_with_materials, machine_with_qualities
-
-    def getAllLocalPackages(self) -> Dict[str, Dict[str, Any]]:
-        """ returns an unordered list of all the package_info installed, to be installed or to be returned"""
-
-        packages = dict([(package_info["package_id"], dict(package_info)) for package in self.getAllInstalledPackagesInfo().values() for package_info in package])
-        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToRemove().values()])
-        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToInstall().values()])
-
-        return packages
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -76,13 +76,10 @@ class CuraPackageManager(PackageManager):
         return machine_with_materials, machine_with_qualities
 
     def getAllLocalPackages(self) -> List[Dict[str, Any]]:
-        """ returns an unordered list of all the package_info installed, to be installed or to be returned"""
+        """ Returns an unordered list of all the package_info of installed, to be installed, or bundled packages"""
+        packages: List[Dict[str, Any]] = []
 
+        for packages_to_add in self.getAllInstalledPackagesInfo().values():
+            packages.extend(packages_to_add)
 
-        class PkgInfo(dict):
-            # Needed helper class because a dict isn't hashable
-            def __eq__(self, item):
-                return item == self["package_id"]
-
-        packages = [PkgInfo(package_info) for package in self.getAllInstalledPackagesInfo().values() for package_info in package]
-        return [dict(package) for package in packages]
+        return packages
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -19,23 +19,11 @@ if TYPE_CHECKING:
 class CuraPackageManager(PackageManager):
     def __init__(self, application: "QtApplication", parent: Optional["QObject"] = None) -> None:
         super().__init__(application, parent)
-
         self._local_packages: Optional[Dict[str, Dict[str, Any]]] = None
-        self._local_packages_installed: Optional[Dict[str, Dict[str, Any]]] = None
-        self._local_packages_to_remove: Optional[Dict[str, Dict[str, Any]]] = None
-        self._local_packages_to_install: Optional[Dict[str, Dict[str, Any]]] = None
-
         self.installedPackagesChanged.connect(self._updateLocalPackages)
 
     def _updateLocalPackages(self) -> None:
-        self._local_packages_installed = dict([(package_info["package_id"], dict(package_info)) for package in self.getAllInstalledPackagesInfo().values() for package_info in package])
-        self._local_packages_to_remove = dict([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToRemove().values()])
-        self._local_packages_to_install = dict([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToInstall().values()])
-
-        self._local_packages = {}
-        self._local_packages.update(self._local_packages_installed)
-        self._local_packages.update(self._local_packages_to_remove)
-        self._local_packages.update(self._local_packages_to_install)
+        self._local_packages = self.getAllLocalPackages()
 
     @property
     def local_packages(self) -> List[Dict[str, Any]]:
@@ -54,15 +42,6 @@ class CuraPackageManager(PackageManager):
             # _updateLocalPackages always results in a list of packages, not None.
             # It's guaranteed to be a list now.
         return set(self._local_packages.keys())
-
-    @property
-    def installed_packages_ids(self) -> Set[str]:
-        """locally installed packages, lazy execution"""
-        if self._local_packages is None:
-            self._updateLocalPackages()
-            # _updateLocalPackages always results in a list of packages, not None.
-            # It's guaranteed to be a list now.
-        return set(self._local_packages_installed.keys())
 
     def initialize(self) -> None:
         self._installation_dirs_dict["materials"] = Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer)
@@ -93,3 +72,12 @@ class CuraPackageManager(PackageManager):
                         machine_with_qualities.append((global_stack, str(extruder_nr), container_id))
 
         return machine_with_materials, machine_with_qualities
+
+    def getAllLocalPackages(self) -> Dict[str, Dict[str, Any]]:
+        """ returns an unordered list of all the package_info installed, to be installed or to be returned"""
+
+        packages = dict([(package_info["package_id"], dict(package_info)) for package in self.getAllInstalledPackagesInfo().values() for package_info in package])
+        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToRemove().values()])
+        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToInstall().values()])
+
+        return packages
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -19,11 +19,13 @@ if TYPE_CHECKING:
 class CuraPackageManager(PackageManager):
     def __init__(self, application: "QtApplication", parent: Optional["QObject"] = None) -> None:
         super().__init__(application, parent)
-        self._local_packages: Optional[Dict[str, Dict[str, Any]]] = None
+        self._local_packages: Optional[List[Dict[str, Any]]] = None
+        self._local_packages_ids: Optional[Set[str]] = None
         self.installedPackagesChanged.connect(self._updateLocalPackages)
 
     def _updateLocalPackages(self) -> None:
         self._local_packages = self.getAllLocalPackages()
+        self._local_packages_ids = set(pkg["package_id"] for pkg in self._local_packages)
 
     @property
     def local_packages(self) -> List[Dict[str, Any]]:
@@ -32,16 +34,16 @@ class CuraPackageManager(PackageManager):
             self._updateLocalPackages()
             # _updateLocalPackages always results in a list of packages, not None.
             # It's guaranteed to be a list now.
-        return list(self._local_packages.values())
+        return cast(List[Dict[str, Any]], self._local_packages)
 
     @property
     def local_packages_ids(self) -> Set[str]:
         """locally installed packages, lazy execution"""
-        if self._local_packages is None:
+        if self._local_packages_ids is None:
             self._updateLocalPackages()
             # _updateLocalPackages always results in a list of packages, not None.
             # It's guaranteed to be a list now.
-        return set(self._local_packages.keys())
+        return cast(Set[str], self._local_packages_ids)
 
     def initialize(self) -> None:
         self._installation_dirs_dict["materials"] = Resources.getStoragePath(CuraApplication.ResourceTypes.MaterialInstanceContainer)
@@ -73,11 +75,17 @@ class CuraPackageManager(PackageManager):
 
         return machine_with_materials, machine_with_qualities
 
-    def getAllLocalPackages(self) -> Dict[str, Dict[str, Any]]:
+    def getAllLocalPackages(self) -> List[Dict[str, Any]]:
         """ returns an unordered list of all the package_info installed, to be installed or to be returned"""
 
-        packages = dict([(package_info["package_id"], dict(package_info)) for package in self.getAllInstalledPackagesInfo().values() for package_info in package])
-        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToRemove().values()])
-        packages.update([(package["package_info"]["package_id"], dict(package["package_info"])) for package in self.getPackagesToInstall().values()])
 
-        return packages
+        class PkgInfo(dict):
+            # Needed helper class because a dict isn't hashable
+            def __eq__(self, item):
+                return item == self["package_id"]
+
+        packages = [PkgInfo(package_info) for package in self.getAllInstalledPackagesInfo().values() for package_info in package]
+        packages.extend([PkgInfo(package["package_info"]) for package in self.getPackagesToRemove().values() if package["package_info"]["package_id"] not in packages])
+        packages.extend([PkgInfo(package["package_info"]) for package in self.getPackagesToInstall().values() if package["package_info"]["package_id"] not in packages])
+
+        return [dict(package) for package in packages]
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 
-from typing import Any, cast, Dict, List, Tuple, TYPE_CHECKING, Optional
+from typing import Any, cast, Dict, List, Set, Tuple, TYPE_CHECKING, Optional
 
 from cura.CuraApplication import CuraApplication  # To find some resource types.
 from cura.Settings.GlobalStack import GlobalStack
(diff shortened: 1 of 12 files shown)


plugins/Marketplace/resources/qml/PackageCard.qml
```diff
@@ -354,28 +320,71 @@ Rectangle
...
+                        busyPrimaryText: catalog.i18nc("@button", "Enabling...")
+                        confirmedPrimaryText: catalog.i18nc("@button", "Enabled")
+                        secondaryText: catalog.i18nc("@button", "Disable")
+                        busySecondaryText: catalog.i18nc("@button", "Disabling...")
+                        confirmedSecondaryText: catalog.i18nc("@button", "Disabled")
+                        enabled: !(installManageButton.busy || updateManageButton.busy)
+                        visible: root.manageableInListView || root.expanded
+
+                        onClicked: {
+                            if (primary_action)
```
rburema, line 338: Maybe we can just do `onClicked: packageData.setEnabledPackageTriggered(primary_action, packageData.packageId)`?

Not a requirement, just thought I'd mention the possibility.

plugins/Marketplace/resources/qml/PackageCard.qml
```diff
@@ -354,28 +320,71 @@ Rectangle
...
+                        confirmedPrimaryText: catalog.i18nc("@button", "Installed")
+                        secondaryText: catalog.i18nc("@button", "Uninstall")
+                        busySecondaryText: catalog.i18nc("@button", "Uninstalling...")
+                        confirmedSecondaryText: catalog.i18nc("@button", "Uninstalled")
+                        enabled: !(enableManageButton.busy || updateManageButton.busy)
+                        visible: state == "confirmed" || root.manageableInListView || root.expanded
+
+                        onClicked:
+                        {
+                            if (primary_action)
```
rburema, line 365: (Same here.)

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +117,217 @@ def hasFooter(self) -> bool:
...
+        :param update: A flag if this is download request is an update process
+        """
+
+        def downloadFinished(reply: "QNetworkReply") -> None:
+            self._downloadFinished(package_id, reply, update)
+
+        def downloadError(reply: "QNetworkReply", error: "QNetworkReply.NetworkError") -> None:
+            self._downloadError(package_id, update, reply, error)
+
+        HttpRequestManager.getInstance().get(
```
rburema, line 207: I think we might need a (second?) `self._ongoing_request = ...` here just to be on the safe side. Though, on the other hand, pressing the install button, immediately switching tabs a few times and then pressing another install button is not likely to happen.

rburema (commented):
Comments up until now. The code also gave me a few ideas of where to component-test, so this is neither approve or reject, as most of my remarks could have explanations.

plugins/Marketplace/Marketplace.py
```diff
@@ -19,18 +17,17 @@
 if TYPE_CHECKING:
     from PyQt5.QtCore import QObject
 
-ROOT_URL = f"{UltimakerCloudConstants.CuraCloudAPIRoot}/cura-packages/v{UltimakerCloudConstants.CuraCloudAPIVersion}/cura/v{CuraSDKVersion}"  # Root of all Marketplace API requests.
-PACKAGES_URL = f"{ROOT_URL}/packages"  # URL to use for requesting the list of packages.
 
-
-class Marketplace(Extension):
+class Marketplace(Extension, QObject):
```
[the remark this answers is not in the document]
jellespijker: Will look into it

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +117,219 @@ def hasFooter(self) -> bool:
...
+        # close dialog
+        dialog = self._license_dialogs.pop(package_id)
+        if dialog is not None:
+            dialog.deleteLater()
+        # reset package card
+        package = self.getPackageModel(package_id)
+        package.is_installing = False
+
+    def _requestInstall(self, package_id: str, update: bool = False) -> None:
+        Logger.debug(f"Request installing {package_id}")
```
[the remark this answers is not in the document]
jellespijker: Well we have multiple logging levels. I'm more in favor of setting and using those correctly. But You're right in the sense that these don't add to much value now. Will prune them

plugins/Marketplace/resources/qml/PackageCard.qml
jellespijker: That is indeed a possibility but I'm curious at what the benefit might be? I'm inclined to leave it be If it's just a styling issue, except of course it is out of line with the rest of the style. If there are other benefits I'm more then willing to change it.

plugins/Marketplace/PackageList.py
jellespijker: I also tried to break it with big packages, such as the CAD plugin and I didn't see any trouble. But sometimes a bit of defensive programming is a good thing. I will change it.

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+    # --- Handle the manage package buttons ---
+
+    def _connectManageButtonSignals(self, package: PackageModel) -> None:
+        package.installPackageTriggered.connect(self.installPackage)
+        package.uninstallPackageTriggered.connect(self.uninstallPackage)
+        package.updatePackageTriggered.connect(self.updatePackage)
+        package.enablePackageTriggered.connect(self.enablePackage)
+        package.disablePackageTriggered.connect(self.disablePackage)
+
+    @pyqtSlot(str)
```
nallath, line 282: This doesn't have to be a pyqtSlot. This is only called by a signal from package.

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+        """Install a package from the Marketplace
+
+        :param package_id: the package identification string
+        """
+        package = self.getPackageModel(package_id)
+        package.is_installing = ManageState.PROCESSING
+        url = package.download_url
+        self.download(package_id, url, False)
+
+    @pyqtSlot(str)
```
nallath, line 293: Same as installPackage

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+
+        :param package_id: the package identification string
+        """
+        package = self.getPackageModel(package_id)
+        package.is_installing = ManageState.PROCESSING
+        self._manager.removePackage(package_id)
+        self.unsunscribeUserFromPackage(package_id)
+        package.is_installing = ManageState.HALTED
+
+    @pyqtSlot(str)
```
nallath, line 305: Same as above.

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+
+        :param package_id: the package identification string
+        """
+        package = self.getPackageModel(package_id)
+        package.is_updating = ManageState.PROCESSING
+        self._manager.removePackage(package_id, force_add = True)
+        url = package.download_url
+        self.download(package_id, url, True)
+
+    @pyqtSlot(str)
```
nallath, line 317: Same as above

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+        package.disablePackageTriggered.connect(self.disablePackage)
+
+    @pyqtSlot(str)
+    def installPackage(self, package_id: str) -> None:
+        """Install a package from the Marketplace
+
+        :param package_id: the package identification string
+        """
+        package = self.getPackageModel(package_id)
+        package.is_installing = ManageState.PROCESSING
```
nallath, line 289: Why not let the package set that itself? It's already emitting the signal, right?

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +126,214 @@ def hasFooter(self) -> bool:
...
+        self.download(package_id, url, False)
+
+    @pyqtSlot(str)
+    def uninstallPackage(self, package_id: str) -> None:
+        """Uninstall a package from the Marketplace
+
+        :param package_id: the package identification string
+        """
+        package = self.getPackageModel(package_id)
+        package.is_installing = ManageState.PROCESSING
```
nallath, line 300: Package needs to set this themselves

cura/CuraPackageManager.py
```diff
@@ -47,3 +63,30 @@ def getMachinesUsingPackage(self, package_id: str) -> Tuple[List[Tuple[GlobalSta
                         machine_with_qualities.append((global_stack, str(extruder_nr), container_id))
 
         return machine_with_materials, machine_with_qualities
+
+    def getAllLocalPackages(self) -> List[Dict[str, Any]]:
+        """ returns an unordered list of all the package_info installed, to be installed or to be returned"""
+
+        class PkgInfo:
+            # Needed helper class because a dict isn't hashable
```
nallath, line 71: Dict can also be hashed if you do this: 

class hashabledict(dict):
    def __hash__(self):
        return hash(tuple(sorted(self.items())))

plugins/Marketplace/LocalPackageList.py
```diff
@@ -44,50 +48,63 @@ def updatePackages(self) -> None:
...
-        return PackageModel(package_info, installation_status = bundled_or_installed, section_title = section_title, parent = self)
+        section_title = self.PACKAGE_CATEGORIES[bundled_or_installed][package_type]
+        package = PackageModel(package_info, section_title = section_title, parent = self)
+        self._connectManageButtonSignals(package)
+        package.can_downgrade = self._manager.canDowngrade(package_id)
+        if package_id in self._manager.getPackagesToRemove() or package_id in self._manager.getPackagesToInstall():
+            package.is_recently_installed = True
+        return package
+
+    def checkForUpdates(self, packages: List[Dict[str, Any]]):
```
nallath, line 74:
```suggestion
    def checkForUpdates(self, packages: List[Dict[str, Any]]) -> None:
```

plugins/Marketplace/Manager.py
```diff
@@ -0,0 +1,32 @@
...
+#  Cura is released under the terms of the LGPLv3 or higher.
+from typing import Optional
+
+from PyQt5.QtCore import QObject, pyqtProperty, pyqtSignal
+
+from cura.CuraApplication import CuraApplication
+from UM.PluginRegistry import PluginRegistry
+
+class Manager(QObject):
+    def __init__(self, parent: Optional[QObject] = None):
```
nallath:
```suggestion
    def __init__(self, parent: Optional[QObject] = None) -> None:
```

plugins/Marketplace/Manager.py
```diff
@@ -0,0 +1,32 @@
...
+        super().__init__(parent = parent)
+        self._manager: "CuraPackageManager" = CuraApplication.getInstance().getPackageManager()
```
nallath: A manager that has a manager? I'm not really a fan. Could we come up with a better name for this class?

plugins/Marketplace/Manager.py
```diff
@@ -0,0 +1,32 @@
...
+        self._plugin_registry: PluginRegistry = CuraApplication.getInstance().getPluginRegistry()
+
+        self._manager.installedPackagesChanged.connect(self.checkIfRestartNeeded)
+        self._plugin_registry.hasPluginsEnabledOrDisabledChanged.connect(self.checkIfRestartNeeded)
+
+        self._restart_needed = False
+
+    def checkIfRestartNeeded(self):
```
nallath:
```suggestion
    def checkIfRestartNeeded(self) -> bool:
```

plugins/Marketplace/Marketplace.py
```diff
@@ -31,9 +27,11 @@ class Marketplace(Extension):
     def __init__(self) -> None:
         super().__init__()
         self._window: Optional["QObject"] = None  # If the window has been loaded yet, it'll be cached in here.
+        self.plugin_registry: Optional[PluginRegistry] = None
```
nallath, line 30:
```suggestion
        self._plugin_registry: Optional[PluginRegistry] = None
```

plugins/Marketplace/Marketplace.py
```diff
@@ -43,6 +41,7 @@ def show(self) -> None:
         If the window hadn't been loaded yet into Qt, it will be created lazily.
         """
         if self._window is None:
+            self.plugin_registry = PluginRegistry.getInstance()
```
nallath, line 44:
```suggestion
            self._plugin_registry = PluginRegistry.getInstance()
```

plugins/Marketplace/PackageModel.py
```diff
@@ -1,16 +1,25 @@
...
+from PyQt5.QtCore import pyqtProperty, QObject, pyqtSignal
+
+from cura.CuraApplication import CuraApplication
 from cura.Settings.CuraContainerRegistry import CuraContainerRegistry  # To get names of materials we're compatible with.
 from UM.i18n import i18nCatalog  # To translate placeholder names if data is not present.
 
 catalog = i18nCatalog("cura")
 
 
+class ManageState(Enum):
```
nallath, line 17: Needs documentation. What do these states mean? Especially halted has a special connotation in this context

plugins/Marketplace/PackageModel.py
```diff
@@ -250,3 +277,148 @@ def isCompatibleMaterialStation(self) -> bool:
...
+    enablePackageTriggered = pyqtSignal(str)
+
+    disablePackageTriggered = pyqtSignal(str)
+
+    recentlyInstalledChanged = pyqtSignal(bool)
+
+    # --- enabling ---
+
+    @pyqtProperty(str, notify = stateManageButtonChanged)
+    def stateManageEnableButton(self) -> str:
```
nallath, line 300: This shouldn't be handled in the model, this is something that the view should handle

plugins/Marketplace/PackageModel.py
```diff
@@ -250,3 +277,148 @@ def isCompatibleMaterialStation(self) -> bool:
...
+        if self._is_enabling == ManageState.PROCESSING:
+            return "busy"
+        if self._package_type == "material" or not self._is_installed:
+            return "hidden"
+        if self._is_installed and self._is_active:
+            return "secondary"
+        return "primary"
+
+    @property
+    def is_enabling(self) -> ManageState:
```
nallath, line 311: I don't think that you need this, but this is probably a consequence of having this handled in the model and not the view.

plugins/Marketplace/PackageModel.py
```diff
@@ -250,3 +277,148 @@ def isCompatibleMaterialStation(self) -> bool:
...
+    @is_active.setter
+    def is_active(self, value: bool) -> None:
+        if value != self._is_active:
+            self._is_active = value
+            self.stateManageButtonChanged.emit()
+
+    # --- Installing ---
+
+    @pyqtProperty(str, notify = stateManageButtonChanged)
+    def stateManageInstallButton(self) -> str:
```
nallath, line 337: Same as above, this shouldn't be handled in the model.

plugins/Marketplace/PackageModel.py
```diff
@@ -250,3 +277,148 @@ def isCompatibleMaterialStation(self) -> bool:
...
+    @is_installing.setter
+    def is_installing(self, value: ManageState) -> None:
+        if value != self._is_installing:
+            self._is_installing = value
+            if value == ManageState.HALTED:
+                self._is_recently_installed = True
+            self.stateManageButtonChanged.emit()
+
+    @property
+    def is_recently_installed(self):
```
nallath, line 365: What does recently mean? During this boot of Cura?

plugins/Marketplace/resources/qml/LicenseDialog.qml
```diff
@@ -0,0 +1,92 @@
+//Copyright (c) 2021 Ultimaker B.V.
+//Cura is released under the terms of the LGPLv3 or higher.
+
+import QtQuick 2.10
+import QtQuick.Dialogs 1.1
+import QtQuick.Window 2.2
+import QtQuick.Controls 2.3
+import QtQuick.Layouts 1.3
+import QtQuick.Controls.Styles 1.4
```
nallath, line 9: Remove this. We shouldn't be using controls styles 1 (because we will need to remove it later!)

plugins/Marketplace/resources/qml/ManageButton.qml
```diff
@@ -0,0 +1,236 @@
...
+        {
+            id: confirmedMessageText
+            anchors.verticalCenter: parent.verticalCenter
+
+            font: UM.Theme.getFont("medium_bold")
+            color: UM.Theme.getColor("primary")
+        }
+    }
+
+    states:
```
nallath, line 112: This needs to be refactored I think. Adding 4 buttons and only ever showing one of them is a bit weird. A much better solution is to use a loader and swap out the button as needed (eg; by setting the source / sourcecomponent).

This will also make the QML much easier to read / expand.

plugins/Marketplace/resources/qml/Marketplace.qml
```diff
@@ -8,11 +8,13 @@ import QtQuick.Window 2.2
 
 import UM 1.2 as UM
 import Cura 1.6 as Cura
+import Marketplace 1.0 as Marketplace
 
 Window
 {
     id: marketplaceDialog
     property variant catalog: UM.I18nCatalog { name: "cura" }
+    property variant manager: Marketplace.Manager { }
```
nallath: Now that I read this, why not just set the properties that you need from the manager on the marketplace directly?

nallath (changes requested):

plugins/Marketplace/resources/qml/ManageButton.qml
rburema: Except you can have up to 3 of them showing.

plugins/Marketplace/PackageList.py
rburema: Jelle will do that now.

plugins/Marketplace/PackageList.py
rburema: Jelle will do that now.

plugins/Marketplace/PackageList.py
rburema: Jelle will do that now.

plugins/Marketplace/PackageList.py
rburema: Jelle will do that now.

plugins/Marketplace/resources/qml/ManageButton.qml
nallath: Not if you look at the states. Every state hides 3 items and show only one.

cura/CuraPackageManager.py
jellespijker: But then the hash is calculated over the complete dict and I'm just interested in the `package_id`

plugins/Marketplace/Manager.py
jellespijker: Already renamed

plugins/Marketplace/PackageModel.py
jellespijker: Simplified it it is now basically a toggle between primary and secondary for all, I will rename these accordingly. Goes for all of these stateManage....

plugins/Marketplace/PackageModel.py
jellespijker: Yes, should I work that into the naming?

plugins/Marketplace/resources/qml/Marketplace.qml
jellespijker: The (Restar)Manager was intended to also house the check for updates API queries. since we will need that functionality again when the old Toolbox is removed. The red dot we're seeing atm is still the old Toolbox checking for updates.

Gonna give a good thought about this now we're doing the overhaul

plugins/Marketplace/resources/qml/LicenseDialog.qml
nallath: Fixed in f993243d57bf030f78f8e1e8383428e990349db0

nallath (approved):

plugins/Marketplace/resources/qml/PackageCardHeader.qml
```diff
@@ -0,0 +1,239 @@
...
+                visible: packageData.isCheckedByUltimaker
+            }
+
+            Control
+            {
+                Layout.preferredWidth: UM.Theme.getSize("card_tiny_icon").width
+                Layout.preferredHeight: UM.Theme.getSize("card_tiny_icon").height
+                Layout.alignment: Qt.AlignCenter
+                enabled: false  // remove!
+                visible: false  // replace packageInfo.XXXXXX
```
Ghostkeeper, line 78: This control should be removed now?

Ghostkeeper (changes requested):
A couple of changes to request, about the last 70 commits (since the previous review).

The most important ones are about the RuntimeErrors being caught now, I think.

plugins/Marketplace/PackageModel.py
```diff
@@ -58,10 +67,43 @@ def __init__(self, package_data: Dict[str, Any], installation_status: str, secti
...
+        self._is_busy = False
+
+    @pyqtSlot()
+    def _processUpdatedPackages(self):
+        self.setCanUpdate(self._package_manager.checkIfPackageCanUpdate(self._package_id))
+
+    def __del__(self):
+        try:
+            self._package_manager.packagesWithUpdateChanged.disconnect(self._processUpdatedPackages)
+        except RuntimeError:
```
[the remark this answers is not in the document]
nallath: The runtime error is caused by the C++ object already being deleted. Setting the ownership to CPP should solve it, but it doesn't.

plugins/Marketplace/PackageModel.py
```diff
@@ -250,3 +288,99 @@ def isCompatibleMaterialStation(self) -> bool:
...
+            except RuntimeError:
+                pass
+
+    def _packageInstalled(self, package_id: str) -> None:
+        if self._package_id != package_id:
+            return
+        self.setBusy(False)
+        try:
+            self.stateManageButtonChanged.emit()
+        except RuntimeError:
```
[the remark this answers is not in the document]
nallath: The runtime error is caused by the C++ object already being deleted. Setting the ownership to CPP should solve it, but it doesn't.

So we can't move it anywhere else and we really have to listen to the runtime error (unless you know of another way to prevent this issue from occuring)

plugins/Marketplace/Marketplace.py
```diff
@@ -43,6 +41,7 @@ def show(self) -> None:
         If the window hadn't been loaded yet into Qt, it will be created lazily.
         """
         if self._window is None:
+            self._plugin_registry = PluginRegistry.getInstance()
```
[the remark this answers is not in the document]
nallath: In my experience it makes it easier to test it, since you can use fixtures to prevent having to set the plugin registry every time

So instead of having to use context managers every time you make a call where you want the pluginregistry to be different, you can just use a fixture that creates a marketplace item with a different plugin registry.

plugins/Marketplace/PackageList.py
```diff
@@ -91,3 +130,171 @@ def hasFooter(self) -> bool:
         """ Indicating if the PackageList should have a Footer visible. For paginated PackageLists
         :return: ``True`` if a Footer should be displayed in the ListView, e.q.: paginated lists, ``False`` Otherwise"""
         return self._has_footer
+
+    def getPackageModel(self, package_id: str) -> PackageModel:
+        index = self.find("package", package_id)
+        return self.getItem(index)["package"]
+
+    canInstallChanged = pyqtSignal(str, bool)
```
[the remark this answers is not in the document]
nallath: I can even raise you one better; the signal isn't needed at all. This can just be handled with a direct call.

plugins/Marketplace/PackageModel.py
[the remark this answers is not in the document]
nallath: Ah, your above suggestion fixes this. We can remove this

plugins/Marketplace/resources/qml/PackageCardHeader.qml
nallath: Yep

PR #11045, opened 2021-12-09
reduce flow for 100% density parts

rijkvanmanen:
Overextrusion needs to be avoided to main high quality. More info: PP-86

### DIFF
--- a/resources/definitions/ultimaker.def.json
+++ b/resources/definitions/ultimaker.def.json
@@ -49,6 +49,18 @@
         },
         "bottom_layers": {
             "value": "math.ceil(round(bottom_thickness / resolveOrValue('layer_height'), 4))"
+        },
+        "roofing_layer_count": {
+            "value": 1
+        },
+        "skin_material_flow": {
+            "value": "0.95 * material_flow"
+        },
+        "roofing_material_flow": {
+            "value": "material_flow"
+        },
+        "infill_material_flow": {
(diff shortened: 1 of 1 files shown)


rijkvanmanen:
I just checked the combing behaviour with for the top surface skin layer (roofing layer). I did saw minor differences between 0 and 1 roofing layer, but nothing strange. Can you @pkuiper-ultimaker do a double check as well?

pkuiper-ultimaker:
> I just checked the combing behavior with for the top surface skin layer (roofing layer). I did saw minor differences between 0 and 1 roofing layer, but nothing strange. Can you @pkuiper-ultimaker do a double check as well?

Old situation (no roofing):
image

New situation (1 roofing layer):
image

There are way to many unretracted travel moves in the top surface layer. NOK

pkuiper-ultimaker (changes requested):
See comments. This changes produces a lot of travel moves in the top layer. NOK

pkuiper-ultimaker:
> > I just checked the combing behavior with for the top surface skin layer (roofing layer). I did saw minor differences between 0 and 1 roofing layer, but nothing strange. Can you @pkuiper-ultimaker do a double check as well?
> 
> Old situation (no roofing): image
> 
> New situation (1 roofing layer): image
> 
> There are way to many unretracted travel moves in the top surface layer. NOK

Latest commit with te removal of the roofing layer now produces an OK result:
image

pkuiper-ultimaker (approved):

PR #11049, opened 2021-12-10
[CURA-8079] Alternate Wall Directions

rburema:
Adds a setting to alternate the direction a wall is printed every other layer (and inset, if applicable).

See also back-end pull: https://github.com/Ultimaker/CuraEngine/pull/1541


resources/definitions/fdmprinter.def.json
```diff
@@ -2907,6 +2907,16 @@
...
+                },
+                "material_alternate_walls":
+                {
+                    "label": "Alternate Wall Directions",
+                    "description": "Alternate wall directions every other layer and inset. Useful for materials that can build up stress, like for metal printing.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": true,
+                    "settable_per_mesh": true,
+                    "settable_per_extruder": true
```
casperlamboo: According to the ticket this setting should be experimental at first

casperlamboo: https://github.com/Ultimaker/Cura/commit/0218f77b3cc01793efb43be6a1470443cfc2c3a3

nallath:
I've cherry-picked the two commits, since i didn't want to be bothered with fixing the merge conflict.

PR #11084, opened 2021-12-17
Pp 50 0.3mm layer height print profiles - update after V&V testing

pkuiper-ultimaker:
Adjust the minimum layer time to 10s for the 0.3mm profiles to improve quality of small features and domes.


pkuiper-ultimaker:
Waiting for beta release to check the response.

PR #11089, opened 2021-12-18
MakerGear M2 V3b printer definition added

wanders78746:
Added printer definition, V3b hotend spec and platform .3mf image for MakerGear M2


nallath (changes requested):
The 3D model is much to big. We keep a max of 1mb per buildplate, but most of them don't even come close to that.

So please simplify the model significantly

Ghostkeeper (approved):
Looks good, and loads without issue.

Sorry for the wait. It will be included in the 5.1 release, finally.

PR #11098, opened 2021-12-20
Fixed some German translations

Sekisback:
Fix some German translations. 
Translated some empty msgstr

### DIFF
--- a/resources/i18n/de_DE/fdmprinter.def.json.po
+++ b/resources/i18n/de_DE/fdmprinter.def.json.po
@@ -57,6 +57,8 @@ msgid ""
 "G-code commands to be executed at the very start - separated by \n"
 "."
 msgstr ""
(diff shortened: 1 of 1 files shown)


Ghostkeeper:
These are currently at the translation bureau to update Cura's translations for Cura 4.13. I think we'll want to update the translations from Lionbridge first and then look into the merge conflicts created by this PR. That should point out the differences between the translation bureau and this PR, if any.

Should be noted that Sekisback recently translated the Settings Guide too and probably made these fixes to fix inconsistencies they found while translating. I'd trust these translations more than Lionbridge's because Sekisback can be considered an expert in the subject matter as well as a native speaker, while Lionbridge doesn't have 3D printing experts.

Ghostkeeper (approved):
Thanks again!

I merged these to our 4.13_translations branch now (resolving merge conflicts) which will get merged to 4.13 today. Once that's merged, Github should pick up that this PR is merged.

PR #11107, opened 2021-12-22
Update cura.po

MarongHappy:


Ghostkeeper:
Please don't create separate pull requests for each file! https://github.com/Ultimaker/Cura/pull/11108 and https://github.com/Ultimaker/Cura/pull/11109 seem to belong to this PR and should be merged at the same time (if merged). Preferably, you'd close the other two and move those commits into the branch represented by this commit.

So this pull request seems to change every single one of Cura's translations into Korean. Could you explain why the previous translations are wrong, and why these are better? Is there some general property of the translation you changed? Like, a change of script? Since I don't know any Korean, I don't know what really changed here. An explanation would help to build some argument for this change and trust/faith that we're doing the right thing.

Also, please don't remove the translations in the comments at the end of the file. That signifies our translation history. If anything, they should be fixed along with the rest. They are there to guide new translations for their word choice, and to automatically restore existing translations if something is placed back that was temporarily removed.

Ghostkeeper:
(See https://github.com/Ultimaker/Cura/pull/11107.)

MarongHappy:
It seems that someone else added some of what I translated in 2018 and uploaded it.
The Korean translation was translated with the wrong Google Translate.
That's why a lot of things seem to have been modified.

MarongHappy:
As I was translating from scratch with a pot file, I did not pay attention to the existing script. Sorry.

MarongHappy:
https://blog.naver.com/eulsoon0/221180644697
My blog path

MarongHappy:
Oh.....

Ghostkeeper:
> It seems that someone else added some of what I translated in 2018 and uploaded it.
The Korean translation was translated with the wrong Google Translate.

The Korean translations were initially created by Brule, Ultimaker's distribution partner in Japan and Korea, back in 2017 (Cura release 2.5, see commit https://github.com/Ultimaker/Cura/commit/eab77417114217906927a7832e8f384e2dc893fb). They were maintained by Bothof, a professional translation service, for 3 years, and then by Lionbridge for 2 years, and for those 5 years also checked for mistakes by Brule. I'm fairly certain none of these 3 parties used Google Translate, and I'm sure that they didn't steal your work from 2018 since these translations were started in 2017 and incrementally added to for the most part.

If these translations were independently started, I think we won't merge them here. If you can point out mistakes in the existing translations we'll gladly fix them though!

MarongHappy:
I'm not saying I stole the translated file.
At the time, I did not know about github, and there were a few people who had already translated cura into cura before me.
I have linked the blog to let you know that I have completed the incomplete cura translation.
This translation is one of them.
Please let me know if anything needs to be corrected. If I have missed something by mistake, I will correct it and upload the file again.

Ghostkeeper:
I wasn't saying you stole these translations (and they are licensed so that you could use them under the terms of the license anyway). I thought you said that the translations were taken from you, since you said:

> It seems that someone else added some of what I translated in 2018 and uploaded it.
The Korean translation was translated with the wrong Google Translate.

> At the time, I did not know about github, and there were a few people who had already translated cura into cura before me.

Be that as it may, it was translated before and I have no information as to why we'd use these translations instead of the ones we already have.

MarongHappy:
In 2018, I translated more than half of the Qura translations, and finally translated all of them into Korean. However, my translation level was not professional.
And I remember that the level of other translators at the time was not much different from mine.
I can't say that the Korean translation I uploaded this time is also perfect.
I think people always make mistakes, but they always try to correct them.

PR #11109, opened 2021-12-22
Update fdmprinter.def.json.po

MarongHappy:


Ghostkeeper:
(See https://github.com/Ultimaker/Cura/pull/11107.)

PR #11115, opened 2021-12-23
increase linewidth to machine nozzle size

rijkvanmanen:
0.3 profiles benefit from larger line width. PP-99


Ghostkeeper (approved):

Ghostkeeper:
Thanks for the increment! Looks good to me, and slices without problems.

PR #11116, opened 2021-12-23
Pp 99 line width fix for .3mm profiles master

rijkvanmanen:


Ghostkeeper:
Ah, I see that this overrides https://github.com/Ultimaker/Cura/pull/11115 in Master?

rijkvanmanen:
In master it is default that the machine nozzle size in the linewidth. So we need the remove a few statements to make this working for master. But as a result the master and 4.13 profile should get the same values.

nallath (approved):

PR #11117, opened 2021-12-23
reduce top/bottom speed by 80%

rijkvanmanen:
The top/bottom speed is reduced by 80% to allow for better cooling of the top layers. This reduces pillowing. The line width is increased by 80% to maintain the flow equalization. PP-95


Ghostkeeper:
This seems to duplicate https://github.com/Ultimaker/Cura/pull/11118.

We merge all changes to 4.13 also into master, so we only need one PR: To 4.13 if it needs to be in 4.13 and master, or just to master if it needs to not be in 4.13 but only for the next release.

Let me know if this assumption was wrong and there needs to be something different in master that is not in https://github.com/Ultimaker/Cura/pull/11118.

rijkvanmanen:
Thanks, this makes it clear. #11117 can indeed be closed.

PR #11153, opened 2021-12-29
[CURA-8591] W.I.P./base-branch for upgrade to Qt 6.2

rburema:
Also see backend PR: https://github.com/Ultimaker/Uranium/pull/764

Assumed is also Python 3.9+ (or 3.10 in practice).


nallath (approved):

Ghostkeeper:
Not an issue
----
The tests are currently failing, now this is to be expected because of this error occurring all over the place:

> E   ModuleNotFoundError: No module named 'PyQt6'

Obviously PyQt6 is not installed yet in the build environment that the CI server uses. I'm also not getting these errors running the tests locally.

Typing is very hard to test since there are no good stubs yet for PyQt6 as far as I could find. Running locally via `ctest --verbose` I'm getting about a million-and-one typing errors, including stuff like `Module "PyQt6.QtCore" has no attribute "pyqtProperty"` which is clearly false. But also good stuff like this one from Uranium: `UM/TaskManagement/TaskManager.py:90: error: Argument 1 to "removePostedEvents" of "QCoreApplication" has incompatible type "None"; expected "QObject"`. I'm not sure how well we can test this other than just verifying that the server gives no issues with the existing stubs. Once there are established stubs for PyQt6 we can improve our tests I think.

Needs investigation
----
We're also seeing these errors appear in the 4.4 to 4.5 upgrade tests:

> E       FileNotFoundError: [Errno 2] No such file or directory: '/github/home/.local/share/UM'

Those are a bit more worrying. Why is this suddenly an issue? Did the `ApplicationIdentifier` change during tests? I am also not getting these failures locally, even if I delete the UM folder (it'll get re-created by the tests).

Needs fixing
----
I am getting this error during the tests in `TestPrinterOutputDevice.py`, which is not a failing test but indicates some sort of failure that wasn't there before. Effectively this breaks that test because it doesn't continue executing the test:

```
Exception ignored in: <function PrinterOutputDevice.__del__ at 0x7f3ffc961d30>
3: Traceback (most recent call last):
3:   File "/home/trin/Gedeeld/Projects/Cura/cura/PrinterOutput/PrinterOutputDevice.py", line 222, in __del__
3:     self.close()
3:   File "/home/trin/Gedeeld/Projects/Cura/cura/PrinterOutput/PrinterOutputDevice.py", line 217, in close
3:     self.setConnectionState(ConnectionState.Closed)
3:   File "/home/trin/Gedeeld/Projects/Cura/cura/PrinterOutput/PrinterOutputDevice.py", line 140, in setConnectionState
3:     cura.CuraApplication.CuraApplication.getInstance().getGlobalContainerStack().setMetaDataEntry("is_online", self.isConnected())
3: AttributeError: 'NoneType' object has no attribute 'getGlobalContainerStack'
```

PR #11195, opened 2022-01-04
Updated strings for pt_BR

Patola:


resources/i18n/pt_BR/fdmextruder.def.json.po
```diff
@@ -4,7 +4,7 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Cura 4.13\n"
+"Project-Id-Version: Cura 4.12\n"
```
Ghostkeeper, line 7:
```suggestion
"Project-Id-Version: Cura 4.13\n"
```

resources/i18n/pt_BR/cura.po
```diff
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Cura 4.13\n"
+"Project-Id-Version: Cura 4.12\n"
```
Ghostkeeper, line 7:
```suggestion
"Project-Id-Version: Cura 4.13\n"
```

resources/i18n/pt_BR/fdmprinter.def.json.po
```diff
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Cura 4.13\n"
+"Project-Id-Version: Cura 4.12\n"
```
Ghostkeeper, line 7:
```suggestion
"Project-Id-Version: Cura 4.13\n"
```

Ghostkeeper (changes requested):
Thanks for the update! Found a little mistake, probably the result of a wrong check-out in Git or something. We'll fix these ourselves.

Ghostkeeper (approved):

PR #11215, opened 2022-01-07
Refuse to listen to QT_PLUGIN_PATH.

rburema:
We don't need it ourselves, and it's a potentially serious attack vector.

CURA-8475

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -15,6 +15,7 @@ if "" in sys.path:
 import argparse
 import faulthandler
 import os
+os.environ["QT_PLUGIN_PATH"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
 
 from PyQt5.QtNetwork import QSslConfiguration, QSslSocket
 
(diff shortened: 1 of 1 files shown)


Ghostkeeper (commented):
Looking deeper into it, this might conflict with Linux: https://github.com/Ultimaker/cura-build/blob/b72857cb5d2cb94e479e65314b1b9149ee966caf/packaging/cura.sh#L6

Either this breaks the Linux builds, or that line should be removed too. I'm running a build now to see which of the two is the case.

Ghostkeeper (changes requested):
Indeed, this breaks the Linux builds. When running the AppImage, we now get this warning:

```
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Aborted (core dumped).
```

Ghostkeeper (approved):

PR #11231, opened 2022-01-10
Integration Elegoo Saturn

fluefiske:
Hey There,

I just bought an ELegoo Saturn, they offer an older version of Cura however I would like to use the latest one. Are you planning to integrate it in the software while currently it is missing.

Thanks for your answer,

André


rburema:
Hi @fluefiske 

First, please note that 'Pull Request's are for people adding code to Cura themselves. Please do not open pull requests if you only have a question.

At Ultimaker, we depend on third party (which may be the company themselves, or someone else) contributions to supply us with new profiles for non-Ultimnaker printers. Since we can't test, and also don't have the time to maintain, such a huge amount of printers ourselves.

If they, or someone else, make a(n actual) pull request, we're happy to merge it, if it passes some quick checks.

PR #11246, opened 2022-01-11
Remove unused import of sentry_sdk

gferon:
This breaks builds without the Sentry SDK. The check below handles the missing library fine, but this line introduced in https://github.com/Ultimaker/Cura/commit/3bfea5b97094615753fccdb7d03be335417f3228 broke it.


rburema:
Thanks!

PR #11251, opened 2022-01-12
Raft Extruders

Ghostkeeper:
See https://github.com/Ultimaker/CuraEngine/pull/1555 for a more detailed description of what this change is for and why we want to make it.

This pull request contains:
* New settings to change the individual layers of the raft (base, interface and surface, or their user-visible names: Base, Middle and Top).
* Updates of the build volume calculations to get the adhesion settings from the correct extruders.
* Updates of the used extruder calculation with the new raft extruders.
* Updates of the initial extruder calculation with the new raft extruders.

Contributes to issue CURA-8868.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5411,7 +5411,54 @@
                     "value": "int(defaultExtruderPosition())",
                     "enabled": "extruders_enabled_count > 1 and (resolveOrValue('adhesion_type') != 'none' or resolveOrValue('prime_tower_brim_enable'))",
                     "settable_per_mesh": false,
-                    "settable_per_extruder": false
+                    "settable_per_extruder": false,
+                    "children":
+                    {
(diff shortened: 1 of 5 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -5411,7 +5411,54 @@
...
+                            "type": "extruder",
+                            "default_value": "0",
+                            "value": "adhesion_extruder_nr",
+                            "enabled": "extruders_enabled_count > 1 and resolveOrValue('adhesion_type') == 'raft'",
+                            "settable_per_mesh": false,
+                            "settable_per_extruder": false
+                        },
+                        "raft_surface_extruder_nr":
+                        {
+                            "label": "Raft Surface Extruder",
```
casperlamboo, line 5452: To be consistent with the rest of Cura's terminology I would say "Raft Top" here.
```suggestion
                            "label": "Raft Top Extruder",
```

PR #11178, opened 2022-01-01
Added files for the Elegoo Neptune 2 printer

therobear:
Added printer profiles, extruder definitions, and mesh files for the Elegoo Neptune 2 series printers.

PR #11255, opened 2022-01-13
Elegoo Neptune 2 Printer Configuration

Toylerrr:
These are the config files that add the Neptune 2 from Elegoo.


Toylerrr:
I realize that there is already a[ pull request ](https://github.com/Ultimaker/Cura/pull/11178) for the Neptune 2 but those were pulled from Elegoos version of cura with no edits. This version has been cleaned up and uses the proper naming conventions of other definitions.
If there are any problems let me know and I will fix them.

therobear:
My pull request also has the stl file for the build plate for the Neptune 2. I have been trying for the past view days to try to fix those errors that you mention but I keep running into issues just trying to get the tests to run.

rburema:
Hi! Sorry for the extreme wait. We've been very busy with 5.0.

rburema:
(Note that unfortunately, this also means this will be in 5.1, as we already have the 5.0 beta out.)

PR #11278, opened 2022-01-17
Add eryone_thinker_directdrive profile

Eryone:
Hi,
We want to add the eryone thinker directdrive profile.
Since many of a lot of the thinker printer users upgrade to this new system. 
So we hope Cura can support this model.

Greeting


dapostol73:
Any issues with getting this pull request merged in?

rburema:
Sorry for the extreme wait. We've been very busy with 5.0, and hope to resume a more normal pacing shortly.

Sadly this also means this will be in 5.1, since the 5.0-beta is already out.

PR #11286, opened 2022-01-18
Refactoring model code

nallath:
I've found some old code that could be simplified quite a bit. There is even more to clean up, but I want to keep the work smallish.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -667,6 +667,7 @@ class CuraApplication(QtApplication):
     def setGlobalContainerStack(self, stack: Optional["GlobalStack"]) -> None:
         self._setLoadingHint(self._i18n_catalog.i18nc("@info:progress", "Initializing Active Machine..."))
         super().setGlobalContainerStack(stack)
+        self.showAddPrintersUncancellableDialog.emit()
 
     showMessageBox = pyqtSignal(str,str, str, str, int, int,
                                 arguments = ["title", "text", "informativeText", "detailedText","buttons", "icon"])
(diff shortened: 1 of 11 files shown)


cura/CuraApplication.py
```diff
@@ -674,22 +668,7 @@ def getGlobalContainerStack(self) -> Optional["GlobalStack"]:
     def setGlobalContainerStack(self, stack: Optional["GlobalStack"]) -> None:
         self._setLoadingHint(self._i18n_catalog.i18nc("@info:progress", "Initializing Active Machine..."))
         super().setGlobalContainerStack(stack)
-
-    showMessageBox = pyqtSignal(str,str, str, str, int, int,
-                                arguments = ["title", "text", "informativeText", "detailedText","buttons", "icon"])
-    """A reusable dialogbox"""
-
-    def messageBox(self, title, text,
```
Ghostkeeper, line 682: Removing this is a bit of an API break. Should we keep it in? Maybe deprecate it? I think it's quite likely that existing plug-ins use this feature.

Ghostkeeper (changes requested):
Good changes. I especially like that it's no longer calling a function through CuraApplication to get the model. Much simpler and more how it's intended to be used.

We do need to look into the API breakage though, especially the `messageBox` function. I suggest that we deprecate that function but keep it functional for now. The other deleted functions have a similar problem but are likely not actually used by any plug-ins (like, which plug-in would create the welcome screens...). To be very principled we should deprecate those too before deleting them but I'm also fine with directly removing them.

PR #11301, opened 2022-01-20
Update StartSliceJob.py

zjzh:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -210,7 +210,7 @@ class StartSliceJob(Job):
 
                 # Filter on current build plate
                 build_plate_number = node.callDecoration("getBuildPlateNumber")
-                if build_plate_number is not None and build_plate_number != self._build_plate_number:
+                if None is not build_plate_number != self._build_plate_number:
                     continue
 
                 children = node.getAllChildren()


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -210,7 +210,7 @@ def run(self) -> None:
 
                 # Filter on current build plate
                 build_plate_number = node.callDecoration("getBuildPlateNumber")
-                if build_plate_number is not None and build_plate_number != self._build_plate_number:
+                if None is not build_plate_number != self._build_plate_number:
```
nallath, line 213: This will _always_ return false. 

`build_plate_number != self._build_plate_number` will always return a bool. You then compare that with None, so that will never be the case.

If you now set build_plate_number to 1, self._build_plate_number to 1, the result will still be False (whereas it should be True!)

nallath (changes requested):

plugins/CuraEngineBackend/StartSliceJob.py
zjzh: Thanks for your comments, but it will not always return false. For example, build_plate_number to 1, self._build_plate_number to 2, the result will be True.
For the above example, if build_plate_number to 1, self._build_plate_number to 1, it should be False.

rburema: In any case, what's important here is that the suggested change is _less_ readable, not more, in my opinion (and I guess @nallath 's too, since neither of us could apparently intuitively read this).

This is a good example of how something being more 'Pythonic' doesn't automatically make it more readable, since in  this case that word seems to mean 'assumes in-depth knowledge of Python'. To a more casual reader of this code, it is unclear whether the comparison is made between `build_plate_number` alone, and then the result of that to the `!=` statement, or whether the result of the `!=` statement is compared to `None` via `is not`, when in fact, _neither_ of these is what's happening (what is happening is actually closer to the original statement than either of those).

PR #11328, opened 2022-01-24
Cura 8378 improve checkbox visibility

Joeydelarago:
The goal of this ticket is to create a single CheckBox that can be used throughout Cura. 

To see the CheckBox component refer to [Ultimaker/Uranium#773](https://github.com/Ultimaker/Uranium/pull/773)

The styles.qml CheckBoxStyle has been updated to the new design as well for use until the old QtControls checkboxes are replaced and the _style_ is no longer in use

Screenshot 2022-01-24 at 16 46 56


resources/qml/Widgets/CheckBox.qml
```diff
@@ -1,77 +0,0 @@
-// Copyright (c) 2020 Ultimaker B.V.
```
nallath, line 1: I don't think you should delete this file.

Personally I would leave this file in and just make it a default UM.Checkbox. That way you don't have to change all the usage of Cura.Checkbox in the rest of the code ;) (And, although not an argument in this case, it doesn't break the API/SDK)

resources/themes/cura-light/icons/low/CheckMark.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nallath, line 1: Why not use Check.svg? 

I think it's rather confusing to have both CheckMark and Check in the icon set, both which only have very subtle differences.

resources/themes/cura-light/icons/default/CheckBoxBorder.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nallath, line 1: This file isn't used (and I also don't think it's needed, since it's pretty easy to create this in QML without using an SVG)

resources/qml/CheckBoxWithTooltip.qml
```diff
@@ -28,11 +28,11 @@ CheckBox
         {
             anchors.verticalCenter: parent.verticalCenter
             anchors.horizontalCenter: parent.horizontalCenter
-            width: Math.round(parent.width / 2.5)
-            height: Math.round(parent.height / 2.5)
+            height: UM.Theme.getSize("checkbox_mark").height
```
nallath, line 31: This component should not be of type Checkbox but of type UM.Checkbox (or even Cura.Checkbox!)

Right now there is quite a bit of code duplication.

nallath (changes requested):

resources/themes/cura-light/icons/low/CheckMark.svg
Joeydelarago: Do plugins use this icon? The CheckMark.svg width/height is 2x Check.

resources/qml/Widgets/CheckBox.qml
Joeydelarago: I think it is better to have one source of truth for the styling of checkboxes, although I agree that it would be nicer to not have to change the usage of Cura.CheckBox, it could lead to de-syncing of the CheckBox style between all menus (Defined in Cura) and the settingVisibilityMenu.qml (Defined in Uranium). 

I'll keep in mind the API/SDK breaks in the future for non major releases. More than happy to make this change if you believe this would make more sense.

resources/themes/cura-light/icons/low/CheckMark.svg
Ghostkeeper: If they do, we can add that to the deprecated icon names (with the replacement) or even just remove them since the next update is a major SDK update.

resources/themes/cura-light/icons/low/CheckThin.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
```
Ghostkeeper, line 1: Hmm, but now nothing uses this `CheckThin` file, right?

If it's not used, I would say we remove this. Not even plug-ins are using it since they would need to use the new name then.

Ghostkeeper (changes requested):
One more thing from the new commits. Maybe I'm misunderstanding your intention here though.

resources/themes/cura-light/icons/low/CheckThin.svg
Joeydelarago: CheckThin is being used by the tick version of this icon. The component is defined in Uranium.
Screenshot 2022-01-28 at 17 16 28
e check badge pop up in Uranium

Ghostkeeper: Ah, I probably had the wrong branch checked out on Uranium when I checked this then, since it's a different repository.

PR #11331, opened 2022-01-24
jade printer by COLIBRI3D is added

duesk:
we are a 3d printing company and we develop a new 3d printer profile for Cura. i add a few files and i did the performance tests in the latest version of cura


Ghostkeeper:
Is there anywhere I could find information about this printer? Like a sales page?

Colibri3D's website appears to be empty (as of this writing): https://colibri3d.com/categoria-producto/impresoras-3d/

duesk:
ready the printer appears on the site, it was out of stock but we already have more. I also share some videos of our users

https://youtu.be/UDtCPsotIXs
https://youtu.be/9GHFN5jnQuA

duesk:
why does this problem appear? : The version of this profile is not up to date!
assert 17 == 19
  + where 17 = int('17')
  + and 19 = CuraApplication.SettingVersion

nallath:
> why does this problem appear? : The version of this profile is not up to date! assert 17 == 19
> 
>     * where 17 = int('17')
> 
>     * and 19 = CuraApplication.SettingVersion

Because the setting version of Cura is 19 right now, not 17.

Ghostkeeper (changes requested):
It looks like these quality profiles were simply copied from the Creality printers. Are they actually the same, and if so, why not just set the quality_definition to `creality_base` to continue getting updates in the future?

nallath:
Closing this due to inactivity

PR #11335, opened 2022-01-25
Fix Transitioning angle description with what it's actually for

BagelOrb:
The description was wholly wrong. Check out the new description to understand the setting for what it actually does!

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1149,7 +1149,7 @@
                 "wall_transition_angle":
                 {
                     "label": "Wall Transitioning Threshold Angle",
-                    "description": "When to create transitions between even and odd numbers of walls. A wedge shape with an angle greater than this setting will not have transitions and no walls will be printed in the center to fill the remaining space. Reducing this setting reduces the number and length these center walls, but may leave gaps or overextrude.",
+                    "description": "When to create transitions between even and odd numbers of walls. A wedge shape with an angle greater than this setting will not have transitions and no walls will be printed in the center to fill the remaining space. Reducing this setting reduces the number and length of these center walls, but may leave gaps or overextrude.",
                     "type": "float",
                     "unit": "°",
                     "default_value": 10,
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1148,8 +1148,8 @@
                 },
                 "wall_transition_angle":
                 {
-                    "label": "Wall Transition Angle",
-                    "description": "When transitioning between different numbers of walls as the part becomes thinner, two adjacent walls will join together at this angle. This can make the walls come together faster than what the Wall Transition Length indicates, filling the space better.",
+                    "label": "Wall Transitioning Threshold Angle",
+                    "description": "When to apply transitioning between even and odd number of walls. A wedge shape with an angle higher than this setting will not have transitioning applied and no odd number gap filler walls will be genertated. Reducing this setting reduces the number and length of gap filler walls at the cost of small overfills and underfills.",
```
Ghostkeeper, line 1152: I'd also like to suggest reducing the amount of industry slang (or at least the more obscure slang) in this description. What do you think?

```suggestion
                    "description": "When to create transitions between even and odd numbers of walls. A wedge shape with an angle greater than this setting will not have transitions and no walls will be printed in the center to fill the remaining space. Reducing this setting reduces the number and length these center walls, but may leave gaps or overextrude.",
```

If this change is applied, perhaps we should also change the setting label to `Wall Transition Threshold Angle`.

BagelOrb: You missed an " of " in your suggestion.

I've made some commits to implement your suggestions. They were good suggestions.

BagelOrb: I explicitly chose to make it Wall *Transitioning* Threshold Angle to make clear that this is not a setting governing any property of a transition; the setting governs when transitions will be generated in the first place. I wanted to make this very clearly different from all other settings. 

This setting in some sense determines how much of the toolpaths will be governed by Arachne and how much toolpaths will be normal. You could consider this the main setting of Arachne.

nallath (approved):

PR #11346, opened 2022-01-26
Add Optional Filter to Removable Drive Output Device Plugin

### LINKED ISSUE #10915: Removable Media Whitelist

At [The Construct @ RIT](https://hack.rit.edu/), we use Cura on shared, public accounts on desktops. As a result, it is very common for users to use USB drives to transfer STL files. In fact, we provide a supply of USB drives for transferring files from personal laptops to the desktops. With the provided saving to removable media option, they can result in the file being saved to one of those drives, which a lot of people don’t remove during or after using Cura, instead of the SD cards (we label as PRINTER_SD) or USB drives (we label as PRINTER_USB) for the printer, which can result in user confusion for where the files went, then proceed to re-slice the file and get billed for it.

For our use case, it would be helpful to have an optional, configurable whitelist of case-insensitive drive names to filter for the removable media saving option. Making the whitelist a blacklist may also be a useful option, but we don’t have this use case. Allowing regular expressions, or other rules like “starts with”, “ends with”, or “contains” may also be useful, but we don’t have this use case either since we just use PRINTER_SD and PRINTER_USB.

To get around this, we wrapped the `WindowsRemovableDrivePlugin` class to apply a filter to the results of `checkRemovableDrives`. There is also an external configuration for the whitelist since we didn’t want to hard-code the names to whitelist for. The specific code changes we do currently aren’t public but can be made public on request if it helps our use case, and potentially other makerspaces looking to use Cura. We can also help with the related code changes if some guidance one what to change can be given.

This feature request affects a combination of the users (students, faculty, and alumni) to reduce user confusion and errors, as well as to Lab Managers for maintaining the systems and helping users.

_No response_

zxc8027:
Resolves #10915, which is already closed.

This pull request adds an optional filter to the Removable Drive Output Device plugin for being able to whitelist or blacklist removable devices. To restate the initial issue, The Construct @ RIT has Cura installed on public desktops, where users typically transfer files using USB drives. We have standardized on PRINTER_SD and PRINTER_USB and currently use a modified version of the plugin to make sure only these devices appear instead of the user's devices. The changes made add a more generalized filter that is stored in the normal configuration. A blacklist option is also implemented, such as for those who have a removable hard drive always plugged in they want to be ignored.

A user interface for modifying the filter is not included. We do not need this for our use case and I am not a fan of doing user interfaces. In theory, a new plugin could exist just for this by changing the preferences directly, but it is not included.

**Edit:** The tests appeared to fail due to comment syntax issues for the version upgrade plugins, not the changes I made.

### DIFF
--- a/plugins/RemovableDriveOutputDevice/DriveFilter.py
+++ b/plugins/RemovableDriveOutputDevice/DriveFilter.py
@@ -1,0 +1,80 @@
+# Copyright (c) 2022 Ultimaker B.V.
+# Uranium is released under the terms of the LGPLv3 or higher.
+
+import re
+from typing import Dict, List
+
+
+class DriveFilter:
+    def __init__(self):
+        """Creates the filter.
+        """
+
+        self.filterSteps: List[Dict[str, str]] = []
+
+    def reload(self, filterSteps: List[Dict[str, str]]) -> None:
+        """Reloads the filter.
+
+        :param filterSteps: Steps in the filter to apply.
+        """
+
+        self.filterSteps = filterSteps
+
+    def passesFilter(self, string: str) -> bool:
+        """Returns if a string can pass the filter.
+        A string is considered passing the filter if:
+        1. It passes at least 1 whitelist condition.
+        2. It passes every blacklist condition.
+
+        :param string: String to test.
+        :return: Whether the string passes the filter.
+        """
+
+        # Iterate through the steps and determine if it passes the whitelist and blacklist.
+        stepsPassed = {
+            "whitelistregex": 0,
+            "blacklistregex": 0,
+        }
+        stepsAttempted = {
+            "whitelistregex": 0,
+            "blacklistregex": 0,
+        }
+        for filterStep in self.filterSteps:
+            # Add the base counters.
+            stepType = filterStep["type"].lower()
+            if stepType not in stepsAttempted:
+                stepsAttempted[stepType] = 0
+                stepsPassed[stepType] = 0
+            stepsAttempted[stepType] += 1
+
+            # Add the passed step.
+            stepPassed = False
+            if stepType == "whitelistregex":
+                regexPattern = re.compile(filterStep["pattern"])
+                if regexPattern.search(string) is not None:
+                    stepPassed = True
+            elif stepType == "blacklistregex":
+                regexPattern = re.compile(filterStep["pattern"])
+                if regexPattern.search(string) is not None:
+                    stepPassed = True
+            if stepPassed:
+                stepsPassed[stepType] += 1
+
+        # Return if the whitelist and blacklist pass.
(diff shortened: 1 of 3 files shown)


zxc8027:
Any update on this? 3 weeks have passed since this PR was opened and there have been no comments.

plugins/RemovableDriveOutputDevice/DriveFilter.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def passesFilter(self, string: str) -> bool:
+        """Returns if a string can pass the filter.
+        A string is considered passing the filter if:
+        1. It passes at least 1 whitelist condition.
+        2. It passes every blacklist condition.
+
+        :param string: String to test.
+        :return: Whether the string passes the filter.
+        """
```
Ghostkeeper, line 69: It seems that this whole function can be simplified to a single regex, even. Something like this (untested):

    blacklists = "&".join([("^(" + step["pattern"] + ")") for step in self.filterSteps if step["type"] == "blacklistregex"])
    whitelists = "|".join([step["pattern"] for step in self.filterSteps if step["type"] == "whitelistregex"])
    return re.search("^(" + blacklists + ")&" + whitelists, string) is not None

That kind of goes to show that maybe it's sufficient to provide a single whitelist, or a single whitelist and a single blacklist. I think that would make it a lot easier for people to use this feature, since they don't have to type out a whole JSON structure which is error-prone if you need to just have a simple filter.

zxc8027: I went with the original design to make adding new filter types easier to add. The new design is much more compact but I am concerned about what happens if a new user wants to use Cura at the scale of a makerspace and the whitelist/blacklist regex isn't suitable (no clue what though).

As-is, both ideas have some trade-offs without a UI to act as an optional way to edit it: the current implementation requires the user to understand JSON while the potentially simpler implementation requires the user to understand regex. Given the option between the two, I would assume the Lab Director would prefer editing JSON over editing regex because... it's regex. I could get an answer, but either way, if changes were requested, I wouldn't be making them since I graduated, don't use Cura personally, and don't have a build environment set up anymore.

zxc8027: Any update on this? Without a response covering if this is fine or if the design won't be accepted without changes, I can't make said changes. I'd prefer to have all of this resolved before The Construct @ RIT resumes normal operations with the upcoming school year starting in a few weeks.

zxc8027: Again, any updates on this? With PR 777 merged in Uranium, this is the last outstanding pull request I have from school before I can stop using this GitHub account.

casperlamboo:
Hi @zxc8027,

Thank you for opening the PR. Apologies for taking this long to get back to you. I have bought this up with the team, and we have unfortunately decided not to merge this pr. The usecase for the white/black list would be limited compared to the addition in code we'd have to maintain.

Hope you understand.

Thankyou,
Casper

PR #8649, opened 2020-10-29
Add Fracktal Works printers

ganeshmev:
Looking for Cura with Fracktal Works definitions.

Website: fracktal.in
Email: ganesh.v@fracktal.in

PR #11360, opened 2022-01-28
Added Fracktal Works machines and New machine- Julia V6

ganeshmev:
Added Fracktal Works machines and New machine- Julia V6


nallath:
This PR changes a lot of things that don't need changing. To give a (non exclusive) set of examples; Changing the permissions of files, adding icons, changing other printers, etc.

Apart from that, the tests also seem to be failing. 

Please don't make changes to other files in a PR to add a new printer and make sure that the tests are succeeding.

Ghostkeeper:
Similar to https://github.com/Ultimaker/Cura/pull/8649, this changes a lot of things outside of the scope of the PR, making it dangerous to merge. The culprit seems to be again the way in which merges are done. The commit https://github.com/Ultimaker/Cura/pull/11360/commits/91a0d8f2416e4bf6297c71e547be537f6fa78181 looks like a merge, but is not an actual merge. It doesn't have 2 parent commits.

PR #11406, opened 2022-02-02
Post Processing: Add regex flavor to description in 

zinen:
Please state that CURA uses the native python module [re](https://docs.python.org/3/library/re.html) to interpreted "search and replace" function in "post processing" extension.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
+++ b/plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
@@ -38,7 +38,7 @@ class SearchAndReplace(Script):
                 "is_regex":
                 {
                     "label": "Use Regular Expressions",
-                    "description": "When enabled, the search text will be interpreted as a regular expression.",
+                    "description": "When enabled, the search text will be interpreted as a regular expression using python's re  module.",
                     "type": "bool",
                     "default_value": false
                 }


Ghostkeeper:
I think it's a bit technical to talk about Python, modules and the abbreviation `re`. Many users will hardly know what Python is, let alone what a Python module is or what `re` is. Perhaps just in brackets "(Python flavor)" would be less confusing, while still being sufficiently informative for people technical enough to use non-standard features?

nallath:
On the other hand; How many people know what regex is and don't know what python modules are?

zinen:
> I think it's a bit technical to talk about Python, modules and the abbreviation `re`. Many users will hardly know what Python is, let alone what a Python module is or what `re` is. Perhaps just in brackets "(Python flavor)" would be less confusing, while still being sufficiently informative for people technical enough to use non-standard features?

It may be a bit technical but just "(Python flavor)" would not make it 100% clear to the users if this uses the built-in package to interpret regex or another package. Since python does not support regex without importing a package this problem emerges.

In my search I needed both the python flavor info and after finding that. Which exact package. I never heard of the `re` package before this so everything was *learn as you go*.

I also think my suggestion is a tiny addition with lots of info, right to the point and searchable (e.g. on google or Stack Overflow) to find the correct help.

Ghostkeeper:
Most people working with regex probably know what Python is. But the sort of people that use non-standard regex syntax *that is not in the Python implementation* would be the sort of people that use other languages than Python ;)

PR #11409, opened 2022-02-03
Raft Middle Layers

Ghostkeeper:
This feature adds a setting by which the user can change the number of middle layers to print in the raft.

See https://github.com/Ultimaker/CuraEngine/pull/1572 for a more complete description.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5642,7 +5642,7 @@
                     "enabled": "resolveOrValue('adhesion_type') == 'raft'",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true,
-                    "limit_to_extruder": "raft_interface_extruder_nr"
+                    "limit_to_extruder": "raft_surface_extruder_nr"
                 },
                 "raft_surface_thickness":
                 {
@@ -5692,6 +5692,19 @@
                     "settable_per_extruder": true,
                     "limit_to_extruder": "raft_surface_extruder_nr"
                 },
+                "raft_interface_layers":
+                {
+                    "label": "Raft Middle Layers",
+                    "description": "The number of layers between the base and the surface of the raft. These comprise the main thickness of the raft. Increasing this creates a thicker, sturdier raft.",
+                    "type": "int",
+                    "default_value": 1,
+                    "minimum_value": "0",
+                    "maximum_value_warning": "10",
+                    "enabled": "resolveOrValue('adhesion_type') == 'raft'",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": true,
+                    "limit_to_extruder": "raft_interface_extruder_nr"
+                },
                 "raft_interface_thickness":
                 {
                     "label": "Raft Middle Thickness",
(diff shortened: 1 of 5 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -5686,7 +5686,7 @@
                     "minimum_value": "0",
                     "minimum_value_warning": "raft_surface_line_width",
                     "maximum_value_warning": "raft_surface_line_width * 3",
-                    "enabled": "resolveOrValue('adhesion_type') == 'raft'",
+                    "enabled": "resolveOrValue('adhesion_type') == 'raft' and raft_surface_layers > 0",
```
Joeydelarago, line 5689: `"enabled": "resolveOrValue('adhesion_type') == 'raft' and raft_surface_layers > 0"`

should be added to the following settings:

- "raft_surface_speed"
- "raft_surface_acceleration"
- "raft_surface_jerk"
- "raft_surface_fan_speed"
- "raft_surface_extruder_nr"

resources/definitions/fdmprinter.def.json
```diff
@@ -5735,7 +5748,7 @@
                     "minimum_value": "0",
                     "minimum_value_warning": "raft_interface_line_width",
                     "maximum_value_warning": "15.0",
-                    "enabled": "resolveOrValue('adhesion_type') == 'raft'",
+                    "enabled": "resolveOrValue('adhesion_type') == 'raft' and raft_interface_layers > 0",
```
Joeydelarago, line 5751: `"enabled": "resolveOrValue('adhesion_type') == 'raft' and raft_interface_layers > 0"`

should be added to the following settings:

- "raft_interface_speed"
- "raft_interface_acceleration"
- "raft_interface_jerk"
- "raft_interface_fan_speed"
- "raft_interface_extruder_nr"

tests/TestBuildVolume.py
```diff
@@ -163,6 +163,7 @@ def test_computeDisalowedAreasMutliExtruder(self, build_volume):
 
 class TestUpdateRaftThickness:
     setting_property_dict = {"raft_base_thickness": {"value": 1},
+                             "raft_interface_layers": {"value": 2},
                              "raft_interface_thickness": {"value": 1},
                              "raft_surface_layers": {"value": 1},
```
Joeydelarago, line 168:
```suggestion
                             "raft_surface_layers": {"value": 3},
```

Would be nice to throw a test for multiple surface layers in too. Made it 3 instead of 2 so it is clearer which is the issue when the test fails.

Joeydelarago (changes requested):

resources/definitions/fdmprinter.def.json
```diff
@@ -5692,6 +5692,19 @@
                     "settable_per_extruder": true,
                     "limit_to_extruder": "raft_surface_extruder_nr"
                 },
+                "raft_interface_layers":
```
Joeydelarago, line 5695: Raft Middle Layers only shows up when I search, settings are all default
Screenshot 2022-02-04 at 15 52 47

resources/definitions/fdmprinter.def.json
Ghostkeeper: We can't add it to the `raft_surface_extruder_nr` since that determines which extruder to get the setting from. But the rest, yes.

nallath (approved):

Issue #10917, opened 2021-11-24
Intercept/Block Exporting Prints in Plugins


zxc8027:
### Is your feature request related to a problem?

At [The Construct @ RIT](https://hack.rit.edu/), RIT’s makerspace, we have been using a modified version of Cura 4 for over a year to allow users to slice their prints for our 3D printers while providing some control over what they can do and logging what they print to be billed later. In order to log prints, we must have users enter their information before they can save a GCode or X3G file. We have many use cases for this:
* When a user enters their information, we store the name of their file, the material used, and the weight of their print according to Cura. This allows us to bill them at the end of the semester for the prints they owe. We have two reasons to do this 1) to get back some of the money spent on material and 2) to discourage people from printing large amounts of random things for free.
* With the information stored,  we can figure out patterns of who uses prints and for what types of projects. It also helps us predict filament usage for purchasing and printers to buy.
* If a print causes an issue with the printer, we are able to contact the person who printed to figure out what happened and prevent future failures.
* If a user refuses to fill out the form, the print is not saved and the user can not continue.

### Describe the solution you'd like

For our requirements to make a plugin for payment screens and other systems, we need a way to be able to intercept a request to export a print and potentially either delay it by an unknown period of time or block if the user fails to enter their information or another check fails. We do need to know the exact file name though, as well as the weight and material used for the print. Other details, like the printer and the output file, may be needed for others, but we do not have this use case.

### Describe alternatives you've considered

We’ve gotten around this by wrapping the `LocalFileOutputDevice` and `RemovableDriveOutputDevice` to call an external method, which may actually call to write the file in the future. These modifications have made upgrading versions of Cura difficult and making the payment screen a plugin difficult. The specific code changes we do currently aren’t public but can be made public on request if it helps our use case, and potentially other makerspaces looking to use Cura. We can also help with the related code changes if some guidance on what to change can be given.

### Affected users and/or printers

This feature request affects a combination of the users (students, faculty, and alumni) to reduce user confusion and errors, as well as to Lab Managers for maintaining the systems and helping users.

nallath:
These are the interesting questions that I love to be able to think about :)

There is an "onWriteStarted" signal that you could hook into. But it's not really intended to force people to actually do something interface wise. It also doesn't allow for any kind of exit (eg: Something isn't the case, cancel the write all together).

[...]

PR #11412, opened 2022-02-03
Split requestWrite to Allow Intercepting Writes in RemovableDriveOutputDevice Plugin

zxc8027:
This pull request splits the `requestWrite` method in the `RemovableDriveOutputDevice` plugin to allow intercepting of writes for logging purposes. The specifics of this are covered in #10917.

Example usage for The Construct @ RIT that uses a patch to get this change: https://github.com/TheConstructRIT/Construct-Cura-Plugins/blob/master/ConstructPaymentWindow/__init__.py


Ghostkeeper:
This change, by itself, seems fine to me. Not for the reasons documented, but because performing the write can easily be seen as a separate task so it could be put in a separate function for better separation of concerns.

I don't see how this would help you write plug-ins, since this is only changed in a plug-in already (and your own writer shouldn't depend on the Removable Drive plug-in).

PR #11415, opened 2022-02-04
Fix show hidden settings

casperlamboo:
In the Cura UI shows if there are hidding settings changed in the settings menu by showing an `i`-icon. Previously clicking this icon would show the hidden setting in the menu, this functionality broke over time. This PR re-enables that functionality.

### DIFF
--- a/resources/qml/Settings/SettingView.qml
+++ b/resources/qml/Settings/SettingView.qml
@@ -340,7 +340,7 @@ Item
                     }
                     function onShowTooltip(text) { base.showTooltip(delegate, Qt.point(-settingsView.x - UM.Theme.getSize("default_margin").width, 0), text) }
                     function onHideTooltip() { base.hideTooltip() }
-                    function onShowAllHiddenInheritedSettings()
+                    function onShowAllHiddenInheritedSettings(category_id)
                     {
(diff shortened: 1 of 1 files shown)


casperlamboo:
After some investigation I found the cause of the issue

In this commit https://github.com/Ultimaker/Cura/commit/132ed1d25835cdb45a4be398f1a982067ac09afe#diff-360c9fe6a0e9202839b1fbbfb934f4af1b9da06353c71f05882ce166d62952e8 the signal
https://github.com/Ultimaker/Cura/blob/132ed1d25835cdb45a4be398f1a982067ac09afe/resources/qml/Settings/SettingItem.qml#L37
was handled as a code block
https://github.com/Ultimaker/Cura/blob/132ed1d25835cdb45a4be398f1a982067ac09afe/resources/qml/Settings/SettingView.qml#L152-L159

By Qt-Magic :magic_wand: :tm: the `category_id` variable from the signal was made available in this code block. Then in https://github.com/Ultimaker/Cura/commit/7827b36ab75ecaa9e78c0d8213f8f7280e660ae0#diff-360c9fe6a0e9202839b1fbbfb934f4af1b9da06353c71f05882ce166d62952e8R369 this code block was changed to a callback function, changing the behaviour of the Qt-Magic :magic_wand: :tm: .

nallath (approved):

PR #11431, opened 2022-02-07
Make `SpinBox` a reusable component

casperlamboo:
Forgot to update Readonly `SpinBox`es to Qt 2 when implementing Cura 8684. As the spinbox is used in various places with the same functionality it made sense to create a reusable component.

There is still a feature gap in the implementation; The `SpinBox`es from QtControls 2.x value is defined as an integer. For some use-cases we do require a fractional value in the `SpinBox`. The only two places where this is required are the `material_diameter` and `material_density` fields in the material prefference page.

Cura 8684


Joeydelarago (changes requested):
As discussed, I think a solution is required to allow for non integer values inside spin boxes.  

Possibly a custom Component something like.

Item 
{
TextInput { validator: DoubleValidator }
Button { id: increase }
Button { id: decrease }
}

casperlamboo:
> 



> As discussed, I think a solution is required to allow for non integer values inside spin boxes.
> 
> Possibly a custom Component something like.
> 
> Item { TextInput { validator: DoubleValidator } Button { id: increase } Button { id: decrease } }

Took another route: https://github.com/Ultimaker/Cura/pull/11431/commits/9c2d370e724deba3eed2af0273407eee7cfd4cf9

Joeydelarago (approved):

PR #11435, opened 2022-02-08
Replace QtControls MenuItem with Cura.MenuItem

Joeydelarago:
Update missed MenuItem.


Ghostkeeper:
There was more than just this one. All of them should be fixed in the [CURA-8688_qt6_cleanup](https://github.com/Ultimaker/Cura/tree/CURA-8688_qt6_cleanup) branch now.

PR #11437, opened 2022-02-08
Enhance capabilities of "Special Modes/Print Sequence/One at a Time" mode

rondlh:
Using the "Special Modes/Print Sequence/One at a Time" mode limits the build volume height to prevent collusion of a part with the gantry. This limitation is applied to all models, but it could be ignored for the last model in the print sequence. Implementing this would make this mode much more useful and flexible.


nallath:
Please don't use pull requests to make feature requests. If you want to make a feature request, create a new issue.

rondlh:
> Please don't use pull requests to make feature requests. If you want to make a feature request, create a new issue.

Sorry about that, thanks for pointing this out

PR #11462, opened 2022-02-11
fix disallowed areas taking brim_gap into account

BagelOrb:
The disallowed areas didn't care about the brim gap, which meant that the brim could try to be printed outside of the build volume.

Before:
image

After:
image

[brim_outside_1.3mf.renamed.zip](https://github.com/Ultimaker/Cura/files/8047654/brim_outside_1.3mf.renamed.zip)

PS:
On another note, the disallowed areas seem to be too thin in some cases, because the extra brim lines are assumed to be a single line, instead of taking the Skirt/Brim minimal length into account.
This brim line seems to be outside of the build volume.
image
[brim_outside.3mf.renamed.zip](https://github.com/Ultimaker/Cura/files/8047636/brim_outside.3mf.renamed.zip)


BagelOrb:
FYI: the function ConvexHullDecorator::_add2DAdhesionMargin seems to be quite outdated. It doesn't take into account:
- the initial layer line width factor
- the brim gap
- the extra brim lines of other extruders

Perhaps we could avoid the code duplication between this function and BuildVolume::_computeDisallowedAreasPrinted ?

BagelOrb:
You should also cherry-pick this comment for the engine: 
afb0c52719799a66f2f5440f2e1b6f63039720b6

MariskaMaas:
For Internal Use CURA-8996

Ghostkeeper (approved):

PR #11463, opened 2022-02-11
Added Nylon quality presets for the Artillery Printer

Superbrain8:


nallath:
For our reference CURA-8952

nallath (approved):

PR #11495, opened 2022-02-16
[CURA-8735] Disable gradual infill when using gyroid infill.

Joeydelarago:
Disable gradual infill settings when using gyroid infill because layers to not overlap correctly

Example of issue
Screenshot 2022-02-16 at 16 58 51


nallath (approved):

nallath:
Where is the matching CuraEngine PR? If a setting is disabled in the front-end, it can still be used by the backend. So if you first set the gradual infill settings and then change to gyroid, you will still get the old effect (making it worse, since it's not even clear what is causing it as the setting is now hidden!)

PR #11513, opened 2022-02-19
Add Creality Ender 3 S1 profile - fixes #11512

### LINKED ISSUE #11512: does not have ender 3 s1 in printer list
### Application Version

4

### Platform

mac

### Printer

ender 3 s1 3d

### Reproduction steps

cant slice and prepare file for printer

### Actual results

does not have ender 3 s1 in printer list

### Expected results

can not slice

- [ ] Log file
- [ ] Project file

add this model to printer list

Sebazzz:
This adds a machine profile for the Ender 3 S1 as requested in #11512. I ported the settings from Creality Cura and made up-to-date Cura printer profiles.

### DIFF
--- a/resources/definitions/creality_ender3s1.def.json
+++ b/resources/definitions/creality_ender3s1.def.json
@@ -1,0 +1,38 @@
+{
+    "name": "Creality Ender-3 S1",
+    "version": 2,
+    "inherits": "creality_base",
+    "metadata": {
+        "quality_definition": "creality_base",
+        "visible": true
+    },
+    "overrides": {
+        "machine_name": { "default_value": "Creality Ender-3 S1" },
+        "machine_width": { "default_value": 220 },
+        "machine_depth": { "default_value": 220 },
+        "machine_height": { "default_value": 270 },
+        "machine_head_with_fans_polygon": { "default_value": [
+                [-26, 34],
+                [-26, -32],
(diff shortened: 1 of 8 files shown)


kvuj:
Yup, can confirm print quality is very nice. It seems to have fixed a couple of issues I had.

resources/definitions/creality_ender3s1.def.json
```diff
@@ -0,0 +1,38 @@
...
+                [-26, -32],
+                [32, -32],
+                [32, 34]
+            ]
+        },
+
+        "gantry_height": { "value": 25 },
+        
+        "speed_print": {"value": 50},
+        "speed_layer_0": {"value": 25},
```
mattpker, line 25: I was having an issue on initial layers with a square object the start of the corners were not getting filled in. I tried the base Ender 3 profile and noticed it resolved the issue. So I started comparing the profiles and noticed speed_layer_0 is 20 on the Ender 3 and 25 on this new S1 profile. I set it to 20 on the S1 profile and it fixed the issue.

Might want to update this to 20 just to avoid issues with first layers, as it should not add much time to the total print times.

I am also curious why we are enabling acceleration control, and not just letting the printer use the default values? I think this is causing issues with my Benchy, but I am still testing that theory.

PR #11519, opened 2022-02-21
Add the new printer and improving the base definitions

jianshu-du:
I  want to change the Magician printer start_gcode and end_gcode, thanks


resources/quality/mingda/PETG/mingda_0.3_PETG_low.inst.cfg
```diff
@@ -13,3 +13,6 @@ variant = 0.3mm Nozzle
 [values]
 speed_layer_0 = 15
 wall_thickness = =line_width*4
+
+raft_interface_thickness = =0.2
```
nallath, line 17:
```suggestion
raft_interface_thickness = 0.2
```
A direct value doesn't need the = in front of it. the extra = is to indicate that the value of the setting should be treated as a function.

nallath (changes requested):

resources/quality/mingda/PETG/mingda_0.3_PETG_low.inst.cfg
jianshu-du: ok

nallath (approved):

PR #11530, opened 2022-02-23
Update to Anycubic i3 Mega S profile

AelricRiggs:
Updated the Anycubic i3 Mega S profile based on data reported by M503 (although it looks like some of those changes where made in the meantime), created a more accurate 3D model for the build platform, and created a small profile for use with Mega S' running [knutwurst's firmware.](https://github.com/knutwurst/Marlin-2-0-x-Anycubic-i3-MEGA-S)


nallath:
It seems that the mesh isn't referenced anywhere, so it won't actually be used for anything...

AelricRiggs:
> It seems that the mesh isn't referenced anywhere, so it won't actually be used for anything...

It's referenced in anycubic_i3_mega_s.def.json on the line `"platform": "anycubic_i3_mega_s_platform.3mf"` and it showed up when I tested it.

AelricRiggs:
Closing this pull request, I'm going to commit the profile changes to the profile's upstream git repo instead.

PR #11531, opened 2022-02-23
Marketplace cleanup

nallath:
CURA-8726


plugins/Marketplace/resources/qml/PackagePage.qml
```diff
@@ -40,7 +40,7 @@ Rectangle
                     id: downloadCount
                     Layout.preferredWidth: parent.width
                     Layout.fillHeight: true
-
+                    visible: packageData.downloadCount != "0"
```
casperlamboo, line 43: Is this the only variable we can use to detect if we are on a download page? 

A plugin with 0 downloads, should still display 0 download. Now it doesn't.

plugins/Marketplace/resources/qml/PackageCardHeader.qml
```diff
@@ -32,8 +32,24 @@ Item
...
+            }
+            if(packageData.packageType == "plugin")
+            {
+                return "../images/Plugin.svg"
+            }
+            if(packageData.packageType == "material")
+            {
+                return "../images/Spool.svg"
+            }
+            return "../images/placeholder.svg"
```
casperlamboo, line 51: I think this would be a great candidate for a switch statement
```suggestion
            switch (packageData.packageType)
            {
                case "plugin":
                    return "../images/Plugin.svg";
                case "material":
                    return "../images/Spool.svg";
                default:
                    return "../images/placeholder.svg";
            }
```

plugins/Marketplace/resources/qml/PackageCardHeader.qml
```diff
@@ -32,8 +32,24 @@ Item
         }
         width: UM.Theme.getSize("card_icon").width
         height: width
-
-        source: packageData.iconUrl != "" ? packageData.iconUrl : "../images/placeholder.svg"
+        sourceSize.height: height
+        sourceSize.width: width
+        source:
+        {
+            if(packageData.iconUrl != "")
```
casperlamboo, line 39:
```suggestion
            if (packageData.iconUrl != "")
```

plugins/Marketplace/resources/qml/Packages.qml
```diff
@@ -83,7 +83,16 @@ ListView
         {
             manageableInListView: packages.packagesManageableInListView
             packageData: model.package
-            width: parent.width - UM.Theme.getSize("default_margin").width - UM.Theme.getSize("narrow_margin").width
+            width: {
+                if(verticalScrollBar.visible)
```
casperlamboo, line 87:
```suggestion
                if (verticalScrollBar.visible)
````

**Review by casperlamboo — COMMENTED** [review]
Code looks good, had some very minor remarks. Take them in consideration, most of these are personal preference so feel free to reject if you disagree.

File: plugins/Marketplace/resources/qml/PackageCardHeader.qml (hunk 2 above)
Comment (casperlamboo) [reply plugins/Marketplace/resources/qml/PackageCardHeader.qml:51]: For compactness, in personal project I also write something like
```suggestion
            switch (packageData.packageType)
            {
                case "plugin": return "../images/Plugin.svg";
                case "material": return "../images/Spool.svg";
                default: return "../images/placeholder.svg";
            }
```
I like this as the low complexity is better reflected in the number of lines used. (wasting a large number of lines on something that isn't very difficult to understand is a bit annoying to read imo)

casperlamboo:
Also noticed that some properties are redundant now that the "read more" link in the package card is removed
https://github.com/Ultimaker/Cura/blob/35f1ad4097b97501c9736e2135490e59d5cbcc86/plugins/Marketplace/resources/qml/PackageCard.qml#L37
https://github.com/Ultimaker/Cura/blob/35f1ad4097b97501c9736e2135490e59d5cbcc86/plugins/Marketplace/resources/qml/PackageCard.qml#L34

plugins/Marketplace/resources/qml/PackagePage.qml
nallath: It was the only way i could come up with. I figured that the single user that would not see the download button was an acceptable issue.

casperlamboo: Maybe we can add a comment explaining why we did this?

nallath: Yup! Done

PR #11563, opened 2022-02-27
[CUBICON] Add models (Optimus-C23, Prime-M22)

HUNIBEST-HyVISION:
Add models (Optimus-C23, Prime-M22)


resources/extruders/cubicon_style_neo_a22_extruder_0.def.json
```diff
@@ -3,24 +3,13 @@
     "name": "Extruder 1",
     "inherits": "fdmextruder",
     "metadata": {
-        "machine": "cubicon_style_neo_a22",
+        "machine": "ultimaker2_plus",
```
nallath, line 6: This isn't correct. This extruder isn't for the Ultimaker2_plus

resources/definitions/cubicon_style_neo_a22.def.json
```diff
@@ -1,40 +0,0 @@
-{
-    "version": 2,
```
nallath, line 2: Why are you removing this printer? This will cause configurations to become corrupted.

nallath (changes requested):

HUNIBEST-HyVISION:
This is the test code tested because a build error occurred after the profile merge.

PR #11571, opened 2022-02-28
Cura 7924 sip cmake build

jellespijker:


plugins/3MFWriter/ThreeMFWriter.py
```diff
@@ -15,7 +15,7 @@
 
 from PyQt6.QtCore import QBuffer
 
-import Savitar
+import pySavitar
```
nallath, line 18:
```suggestion
import pySavitar as Savitar 
```

That's a much easier way to do things; No other code needs to change that way.

plugins/Marketplace/PackageList.py
```diff
@@ -22,8 +22,8 @@
 from .Constants import USER_PACKAGES_URL, PACKAGES_URL
 
 if TYPE_CHECKING:
-    from PyQt5.QtCore import QObject
-    from PyQt5.QtNetwork import QNetworkReply
+    from PyQt6.QtCore import QObject
```
nallath, line 25: These changes should not be needed yet, right? This PR should only be about SIP, not about the actual upgrade to PyQt6

nallath (changes requested):

plugins/Marketplace/PackageList.py
jellespijker: I'm a bit reluctant to change this. I tend to agree with you here. But @rburema made those changes as part of 99449dbc344ac949358a77fbd04326dbab24e342 Since I myself can't determine locally what the impact will we if we revert this. Also keep in mind that we are merging these changes to the PyQt6_upgrade branch. So even reverting them here would mean reverting the revert there.

nallath (approved):

PR #11584, opened 2022-03-02
use current filament diameter

Wikunia:
Currently the filament diameter of extruder `0` is used but if one wants to display a gcode which was sliced with different filament diameters the lines of one will be too thick/thin.

This fixes the problem for up to 4 extruders (otherwise a index out of range error gets thrown),

### DIFF
--- a/plugins/GCodeReader/FlavorParser.py
+++ b/plugins/GCodeReader/FlavorParser.py
@@ -53,7 +53,7 @@ class FlavorParser:
 
     def _clearValues(self) -> None:
         self._extruder_number = 0
-        self._extrusion_length_offset = [0,0,0,0] # type: List[float]
+        self._extrusion_length_offset = [0] * 8 # type: List[float]
         self._layer_type = LayerPolygon.Inset0Type
         self._layer_number = 0
         self._previous_z = 0 # type: float
--- a/plugins/GCodeReader/FlavorParser.py
+++ b/plugins/GCodeReader/FlavorParser.py
@@ -355,7 +355,7 @@ class FlavorParser:
 
         Logger.log("d", "Parsing g-code...")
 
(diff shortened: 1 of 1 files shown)


plugins/GCodeReader/FlavorParser.py
```diff
@@ -53,7 +53,7 @@ def __init__(self) -> None:
 
     def _clearValues(self) -> None:
         self._extruder_number = 0
-        self._extrusion_length_offset = [0] # type: List[float]
+        self._extrusion_length_offset = [0,0,0,0] # type: List[float]
```
nallath, line 56:
```suggestion
        self._extrusion_length_offset = [0] * 8 # type: List[float]
```
The max nr of extruders supported is 8 (by means of the custom fff printer). The * 8 is a shorthand to write [0, 0, 0, 0, 0, 0, 0, 0]

plugins/GCodeReader/FlavorParser.py
```diff
@@ -354,7 +355,7 @@ def processGCodeStream(self, stream: str, filename: str) -> Optional["CuraSceneN
 
         Logger.log("d", "Parsing g-code...")
 
-        current_position = Position(0, 0, 0, 0, [0])
+        current_position = Position(0, 0, 0, 0, [0,0,0,0])
```
nallath, line 358:
```suggestion
        current_position = Position(0, 0, 0, 0, [0] * 8)
```

nallath (changes requested):

plugins/GCodeReader/FlavorParser.py
Wikunia: Is there maybe a variable for the number of extruders in the current printer?

nallath (approved):

PR #11589, opened 2022-03-02
disable 100% infill bottom layers trick for spiralize

BagelOrb:
what the title says


MariskaMaas:
For Internal Use CURA-8999

PR #11592, opened 2022-03-03
Make Wall Ordering option Center Last into its own setting

BagelOrb:
It's called "Middle Last" rather than "Center Last" in order to be consistent with the Middle Line Threshold settings.

See backend PR: https://github.com/Ultimaker/CuraEngine/pull/1597


Ghostkeeper:
We don't want to merge this. See our reasoning for it here: https://github.com/Ultimaker/CuraEngine/pull/1597#issuecomment-1070936191

PR #11595, opened 2022-03-03
Leftover menu fixes 

nallath:
CURA-8928 

We forgot to update the qualities with intent menu. Since it uses tooltips, we can't re-use the menu items.

### DIFF
--- a/resources/qml/PrintSetupSelector/Custom/MenuButton.qml
+++ b/resources/qml/PrintSetupSelector/Custom/MenuButton.qml
@@ -18,11 +18,11 @@ Button
     property string labelText: ""
     id: button
     hoverEnabled: true
-    leftPadding:UM.Theme.getSize("wide_margin").width
+    leftPadding: UM.Theme.getSize("default_margin").width
 
     background: Rectangle
     {
-        id: backgroundRectangle
+        id: backgroundRectanglewide_margin
         color: button.hovered ? UM.Theme.getColor("background_2"): UM.Theme.getColor("background_1")
     }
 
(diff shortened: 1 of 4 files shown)


resources/qml/PrintSetupSelector/Custom/MenuButton.qml
```diff
@@ -23,10 +23,7 @@ Button
     background: Rectangle
     {
         id: backgroundRectangle
-        border.width: UM.Theme.getSize("default_lining").width
-        border.color: button.checked ? UM.Theme.getColor("setting_control_border_highlight") : "transparent"
-        color: button.hovered ? UM.Theme.getColor("action_button_hovered") : "transparent"
-        radius: UM.Theme.getSize("action_button_radius").width
+        color: button.hovered ? UM.Theme.getColor("background_2"): UM.Theme.getColor("background_1")
```
casperlamboo, line 30:
```suggestion
        color: button.hovered ? UM.Theme.getColor("background_2") : UM.Theme.getColor("background_1")
```

resources/qml/PrintSetupSelector/Custom/MenuButton.qml
```diff
@@ -18,11 +18,11 @@ Button
     property string labelText: ""
     id: button
     hoverEnabled: true
-    leftPadding:UM.Theme.getSize("wide_margin").width
+    leftPadding: UM.Theme.getSize("default_margin").width
 
     background: Rectangle
     {
-        id: backgroundRectangle
+        id: backgroundRectanglewide_margin
```
casperlamboo, line 27: I think it's bad practice to mix snake case, camel case (and no case?) in the same id. As the id is never referred to I would remove this id entirely.
```suggestion
```

casperlamboo (changes requested):

PR #11599, opened 2022-03-03
[CURA-8943] Update Tool Panel Icons style

Joeydelarago:
Rotate Tool & Scale tool image buttons swapped out for new style _ToolbarButton_ that is used in _Per Model Settings_

Moved Tooltip.qml and ToolbarButton.qml into Uranium so that the plugins/Tools can access these elements.

CURA-8943


resources/qml/ActionPanel/OutputProcessWidget.qml
```diff
@@ -128,7 +128,7 @@ Column
             tooltip: text
             fixedWidthMode: true
 
-            toolTipContentAlignment: Cura.ToolTip.ContentAlignment.AlignLeft
+            toolTipContentAlignment: Text.AlignLeft
```
nallath, line 131: After wrecking my brain a bit more about this, I've found a more elegant solution! The issue that I have with this is that there are more alignment options in text (and of course that using the Text alignment to align content also feels weird).

So the solution that I've found is that adding a Enums.qml, that contains the enum that you need does the trick.

```
import QtQuick 2.15
import QtQuick.Controls 2.10

Item
{
    enum ContentAlignment
    {
        AlignLeft,
        AlignRight
    }
}
```

This needs to be exposed in the qmldir

`Enums                   1.5 Enums.qml`

The orientation can then be set as you would expect:
`property int contentAlignment: UM.Enums.ContentAlignment.AlignLeft`

Although I don't know why this didn't fail when it was in the cura namespace, i suspect that it has to do with it mirroring the exact name of the controls tooltip. When I renamed the tootlip qml to tooltipp, it did work.

nallath (changes requested):

resources/qml/Toolbar.qml
```diff
@@ -53,7 +53,7 @@ Item
                 width: childrenRect.width
                 height: childrenRect.height
 
-                delegate: ToolbarButton
+                delegate: UM.ToolbarButton
                 {
                     text: model.name + (model.shortcut ? (" (" + model.shortcut + ")") : "")
```
casperlamboo, line 58: not really part of this commit but I think the parenthesis here make the code a bit unclear. I think [javascripts template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) could really benefit here
```suggestion
                    text: model.name + (model.shortcut ? ` ("${model.shortcut}")` : "")
```

PR #11601, opened 2022-03-03
Feature request build plate adhesion

rdfry:
Would like to be able to choose build plate adhesion for each object when printing multiple objects. Would save on filament


fieldOfView:
A *pull*-request is not a feature-request. Feature requests should be posted here:

https://github.com/Ultimaker/Cura/issues/new?assignees=&labels=Type%3A+New+Feature&template=featurerequest.yaml

PR #11623, opened 2022-03-07
fix retraction speed and amount for direct drive afterburner.

cbc02009:
The Voron v2.4 uses a direct drive extruder (a change from when the config was initially created) and the default Cura values are causing a lot of filament clogs for our new builders who don't know the values need to be changed.


Fulg:
I hope this can be merged soon. At the time the VORON profile was submitted, VORON V2.4 did not exist, but commercial VORON kits are using the direct drive extruder (and existing users have upgraded to direct drive too).

bythorsthunder:
Would love to see this merged. It is the source of issues occurring daily on Voron printers.

nallath (approved):

nallath:
Sorry that merging this took so long. We've been in crunch mode for the 5.0 release, so PR's like this didn't get the attention that they deserve.

PR #11626, opened 2022-03-07
[CURA-9008] Reduce width of preferences dialog page selection buttons.

Joeydelarago:
Reduce width of profiles list and printers list.
Reduce width of buttons for selecting pages in preferences dialog

CURA-9008

https://github.com/Ultimaker/Uranium/pull/795

### DIFF
--- a/resources/themes/cura-light/theme.json
+++ b/resources/themes/cura-light/theme.json
@@ -637,6 +637,8 @@
 
         "banner_icon_size": [2.0, 2.0],
 
-        "marketplace_large_icon": [4.0, 4.0]
+        "marketplace_large_icon": [4.0, 4.0],
+
+        "preferences_page_list": [8.0, 2.0]
     }
 }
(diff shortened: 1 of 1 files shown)


resources/themes/cura-light/theme.json
```diff
@@ -637,6 +637,8 @@
 
         "banner_icon_size": [2.0, 2.0],
 
-        "marketplace_large_icon": [4.0, 4.0]
+        "marketplace_large_icon": [4.0, 4.0],
+
+        "preferences_page_list": [8.0, 2.0]
```
casperlamboo, line 642: The themes.json is designed to be configurable by anyone. From that point of view I find it a bit confusing that the height of this entry doesn't affect anything. I would change this to `preference_page_list_item` and have the height control this height https://github.com/Ultimaker/Uranium/blob/f28133f20e71f7bee5cc17818ac6595783ba5e3a/UM/Qt/qml/UM/Preferences/PreferencesDialog.qml#L56.

nallath (approved):

PR #11627, opened 2022-03-07
[CURA-9004] materials selection dropdown improvements

Joeydelarago:
Make material and print core menus appear at bottom of button like drop downs.

CURA-9004


casperlamboo (approved):
Looking good!

PR #11639, opened 2022-03-10
More intuitive line thresholds

BagelOrb:
This PR introduces Minimum Wall Line Width settings which control the Middle Line Thresholds in a more intuitive fashion.

Moreover, this PR includes the formulae which properly relate the Minimum Lien Width settings to the Threshold settings.

These formulae can be used to verify the current behavior of the threshold settings more easily.

image


MariskaMaas:
Thanks! For internal Reference CURA-9054

rburema:
Hi Tim! Since I'm going to use this as a base/reference for the Arachne bug tickets PP&M, I will want to merge this as soon as possible. This does mean that I have a question _despite_ merging it a.s.a.p.: What is the reason for a separate 'Minimum Wall Line Width' versus 'Minimum Thin Wall Line Width'? Is it just to keep this PR from also altering the engine?

PR #11641, opened 2022-03-10
increase filter distance

rijkvanmanen:
Adding and splitting of lines results in print artefacts and additional travel moves. Avoiding these by sacrificing some additional line width change is better. The linewidth can never very more then 75% - 150%, so this long filter distance cannot cause huge line width changes. PP-129


resources/definitions/fdmprinter.def.json
```diff
@@ -1165,11 +1165,11 @@
...
                     "type": "float",
                     "unit": "mm",
-                    "default_value": 1.4,
-                    "value": "4 * math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x",
+                    "default_value": 100,
+                    "value": "100",
                     "minimum_value": "wall_transition_length",
                     "minimum_value_warning": "math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x",
-                    "maximum_value_warning": "10 * math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x"
+                    "maximum_value_warning": "999999"
```
nallath, line 1172:
```suggestion
                    "maximum_value": "999999"
```
The max value warning isn't needed in this case. A max value is needed, as in certain operating systems it causes an overflow.

casperlamboo:
Most of the changes suggested by @rijkvanmanen have already been made here https://github.com/Ultimaker/Cura/pull/11725. As a result this PR has somewhat become obsolete as the problems described in [CURA-9072](https://ultimaker.atlassian.net/browse/CURA-9072) and [CURA-9073](https://ultimaker.atlassian.net/browse/CURA-9073) _should_ be resolved after the #11725 merge.

I'm still merging this PR as setting the `maximum_value` to some high number might prevent some issues we get if that value would be less then the new default of 100 for this setting

PR #11668, opened 2022-03-15
Materials preference page

nallath:
CURA-8979
https://github.com/Ultimaker/Uranium/pull/797

### DIFF
--- a/resources/qml/Preferences/Materials/MaterialsView.qml
+++ b/resources/qml/Preferences/Materials/MaterialsView.qml
@@ -97,7 +97,6 @@ Item
         }
         border.width: UM.Theme.getSize("default_lining").width
         border.color: UM.Theme.getColor("thick_lining")
-        visible: base.hasCurrentItem
 
         ScrollView
         {
@@ -240,11 +239,12 @@ Item
 
                             anchors.verticalCenter: parent.verticalCenter
 
-                            width: childrenRect.width + 2 * UM.Theme.getSize("narrow_margin").width
-                            height: childrenRect.height + 2 * UM.Theme.getSize("narrow_margin").height
+                            width: colorSelectorBackground.width + 2 * UM.Theme.getSize("narrow_margin").width
+                            height: colorSelectorBackground.height + 2 * UM.Theme.getSize("narrow_margin").height
 
                             Rectangle
                             {
+                                id: colorSelectorBackground
                                 color: properties.color_code
                                 width: UM.Theme.getSize("icon_indicator").width
                                 height: UM.Theme.getSize("icon_indicator").height
--- a/resources/qml/Preferences/Materials/MaterialsView.qml
+++ b/resources/qml/Preferences/Materials/MaterialsView.qml
@@ -4,6 +4,7 @@
 import QtQuick 2.7
 import QtQuick.Controls 2.15
 import QtQuick.Dialogs 1.2
+import QtQuick.Layouts 1.3
 
 import UM 1.5 as UM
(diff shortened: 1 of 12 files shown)


Ghostkeeper:
**Bug A:**
I'm getting some QML errors when opening the material page now:

```
2022-03-16 14:18:06,457 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [428]: file:///home/trin/Gedeeld/Projects/Uranium/UM/Qt/qml/UM/Preferences/ManagementPage.qml:187:9: QML Connections: Detected function "onItemsChanged" in Connections element. This is probably intended to be a signal handler but no signal of the target matches the name.
2022-03-16 14:18:06,477 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [428]: file:///home/trin/Gedeeld/Projects/Uranium/UM/Qt/qml/UM/Preferences/ManagementPage.qml:189:13: Unable to assign [undefined] to QObject*
2022-03-16 14:28:54,395 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [428]: file:///home/trin/Gedeeld/Projects/Cura/resources/qml/Preferences/Materials/MaterialsView.qml:498:9: QML Column: ScrollBar must be attached to a Flickable or ScrollView
```
These are only happening when I open the material page in the preferences. Other preference pages don't generate these errors.

**Bug B:**

It seems the text boxes for the description and adhesion information catch scroll events, which makes it hard to scroll all the way down the information tab's scrollview. I don't consider this to be a blocking problem, but if you can fix this in a minute it's something you might want to look at.

**Bug C:**
Fractional filament cost doesn't get saved, it seems. It only stores integer number of Euros. Steps to reproduce:

1. For the filament cost, enter €3,50.
2. To be extra sure, unfocus the material cost text box. I had it round down to the nearest integer once here, but cannot now reproduce it.
3. Close the window (using either Enter, Escape or the X button on the window).
4. Open the materials preference page again (e.g. with Ctrl+K). The filament cost is now €3.00.

Be sure to try this both with `,` and `.` as radix. We had a bug here in Cura 4.x where only the correct radix according to your system's locale is allowed. It's acceptable if this bug is not solved yet, but one of the two should work at least.

It works properly for the density and diameter, and for the weight it's immediately clear that you're not supposed to fill in fractional grams here (so out of scope). For the cost it's important to be able to provide fractionals. This is also a problem for the settings in the Print Settings tab. For instance, setting the Retraction Distance to 6.6mm causes it to round down to 6.0mm.

**Bug D:**
The favourite icon can overlap with the name of the material, making it hard to read:
image

**Bug E:**
When changing the filament diameter such that it rounds to a different integer, it pops up with a message saying that the material will no longer be compatible with the current extruder. That's correct.
However when you change the diameter and then press enter, it closes the window and doesn't save the new diameter.

**Suggestion:**
The name of materials can be elided if it's too long, but you can't read the full name in that list then. Should we add a tooltip if it's elided?

**Additional:**

It's been said before, but the use of spin boxes is a bit superfluous here. This is more of a consideration for @lorenzoromagnoli and I'll not consider it part of this ticket. But indeed, nobody would select a fan speed by clicking the + and - buttons repeatedly.

resources/qml/Preferences/Materials/MaterialsView.qml
```diff
@@ -517,97 +515,120 @@ Item
             }
         }
 
-        Column
+        ScrollView
```
casperlamboo, line 518: Why not make this a `ListView` and set the model property directly on this component

PR #11676, opened 2022-03-16
updated splash screens

lorenzoromagnoli:


fieldOfView:
Would this be a good moment to maybe reconsider the (arguably) somewhat ludicrous claim that Cura is "the most advanced 3d software"? I think I can name a couple of "3d software" titles that are a bit more advanced than Cura. I get we're proud of the product, but it is a rather pretentious and baseless claim IMHO. Maybe just making it a bit more specific as "the most advanced 3d _slicer_ software", so at least we're not claiming to compete full on with Autodesk, Dassault, etc.

lorenzoromagnoli:
Good point @fieldOfView! I will double-check this with marketing.

fieldOfView:
Much better, thanks for the effort!

nallath (approved):

PR #11682, opened 2022-03-18
Add printer profile for Modix3D printers

ComedianTF2:
Has printer profiles for the Modix V3 line of printers: the Big40, 60, 120X, 120Z, 180X and Meter

### DIFF
--- a/resources/definitions/modix_base.def.json
+++ b/resources/definitions/modix_base.def.json
@@ -51,7 +51,6 @@
         "material_initial_print_temperature": { "value": "material_print_temperature" },
         "material_final_print_temperature": { "value": "material_print_temperature" },
         "material_flow": { "value": 100 },
-        "travel_compensate_overlapping_walls_0_enabled": { "value": false },
 
         "z_seam_corner": { "value": "z_seam_corner_weighted" },
         
--- a/resources/definitions/modix_base.def.json
+++ b/resources/definitions/modix_base.def.json
@@ -1,5 +1,6 @@
 {
     "name": "Modix Base Printer",
+	"version": 2,
     "inherits": "fdmprinter",
     "metadata": {
         "author": "Modix",
@@ -12,7 +13,8 @@
       
         "has_variants": true,
         "variants_name": "Nozzle Size",
-        "preferred_variant_name": "0.4 mm Nozzle",
+        "preferred_variant_name": "0.4 mm Nozzle"
+		
 	},
 	"overrides": {
         "machine_name": { "default_value": "Modix Base Printer" },
@@ -31,7 +33,7 @@
         "speed_wall_x": { "value": "speed_wall" },
         "speed_topbottom": { "value": "speed_print / 2" },
         "speed_roofing": { "value": "speed_topbottom" },
-        "speed_travel": { "value": 150.0},
+        "speed_travel": { "value": 150.0 },
         "speed_layer_0": { "value": 20.0 },
         "speed_print_layer_0": { "value": "speed_layer_0" },
         "speed_travel_layer_0": { "value": "speed_layer_0 * speed_travel / speed_print" },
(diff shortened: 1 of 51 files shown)


resources/definitions/modix_base.def.json
```diff
@@ -0,0 +1,198 @@
...
+        },
+        
+        "has_materials": true,
+        "has_variants": true,
+        
+        "variants_name": "Nozzle Size",
+        "preferred_variant_name": "0.4 mm Nozzle",
+		"preferred_material": "generic_pla",
+		"exclude_materials": [
+            "ultimaker_abs_black",
```
nallath, line 23: No need to exclude these, they are 2.85 mm materials.

resources/definitions/modix_base.def.json
```diff
@@ -0,0 +1,198 @@
...
+        "skin_overlap": { "value": 10.0 },
+        "infill_wipe_dist": { "value": 0.0 },
+        "wall_0_wipe_dist": { "value": 0.0 },
+
+        "fill_perimeter_gaps": { "value": "'everywhere'" },
+        "fill_outline_gaps": { "value": false },
+        "filter_out_tiny_gaps": { "value": false },
+
+        "retraction_speed": {
+            "maximum_value_warning": "machine_max_feedrate_e if retraction_enable else float('inf')",
```
nallath, line 133:
```suggestion
            "maximum_value_warning": "machine_max_feedrate_e if retraction_enable else float('inf')",
```
No need to set this; The default is already the same (DRY: Don't Repeat Yourself)

nallath: There are other cases where this is also the case. I've not marked all of them.

nallath (changes requested):

ComedianTF2:
Please hold on, I made some mistakes in that commit and need to fix it still

ComedianTF2:
Updated the modix_base.def.json to delete any duplicates from the fdmprinter.def.json

nallath:
The version number of the profiles isn't correct. THat is why the tests are failing. It should be 19

ComedianTF2:
Fixed the version number from Version 2 to Version 19

nallath:
Nope, sorry. Not the definition one. That should be 2. The version number for the profiles (setting_version = 17) should be 19.

ComedianTF2:
> Nope, sorry. Not the definition one. That should be 2. The version number for the profiles (setting_version = 17) should be 19.

Fixed the profiles version from 17 to 19, fixed the base definition from 19 to 2

nallath:
The CI is still failing on some things.

 [MainThread] UM.Settings.DefinitionContainer._preprocessParsedJson [299]: Unable to override setting travel_compensate_overlapping_walls_0_enabled
[MainThread] UM.Settings.DefinitionContainer._preprocessParsedJson [299]: Unable to override setting fill_perimeter_gaps

Unnecessary default_values for ['retraction_prime_speed', 'retraction_retract_speed'] in /__w/Cura/Cura/tests/Settings/../../resources/definitions/modix_base.def.json

ComedianTF2:
@nallath I'm a little confused about the retraction_prime_speed and retraction_retract_speed, as the default values in fdmprinter.def.json are set to 25, rather than the preferred 40.

nallath:
> @nallath I'm a little confused about the retraction_prime_speed and retraction_retract_speed, as the default values in fdmprinter.def.json are set to 25, rather than the preferred 40.

Correct. But the _value_ property of these is also set. So if you only change the default_value it won't actually change anything. If the value is set, that is what is used (and not the default_value).

ComedianTF2:
@nallath Do you know what's causing the issue with the failing check?

ComedianTF2:
@Ghostkeeper Thanks for the info. The 5.0 launch looks very exciting and I understand the crunch taking priority.I'll be re-submitting the profile again tomorrow, hopefully it will work properly.

PR #11690, opened 2022-03-20
USBPrinterOutputDevice.py: Convert estimated time to int

jimduchek:
PrintJobOutputModel.py requires timeElapsed be an int.  USBPrinterOutputDevice.py sets it to a float instead, which causes PILES of exception windows to open up, though Cura does not crash and the print continues fine (albeit displaying some nonsense values for percentage done, etc).  Simple fix.


nallath:
Thanks for taking the time to fix this!

PR #11698, opened 2022-03-22
Fix "double"/non-functional scrollbars on ScrollViews

fieldOfView:
This PR fixes what looks like a vestigial "double" scrollbar in some places. The mechanism seems to be that even though a custom scrollbar is provided, the "default" scrollbar is not hidden because it has a parent outside the scrollview (and the custom scrollbar is inside the scrollview). The resulting "default" scrollbar is non-functional, but its hitarea overlaps the custom scrollview so that one has a very small hit-area.

scrollbar

Both scrollbars showing in the materials pane are affected:
image

The profile selector is also affected:
image

For (some) details: https://doc.qt.io/qt-5/qml-qtquick-controls2-scrollbar.html#attaching-scrollbar-to-a-flickable

### DIFF
--- a/resources/qml/Dialogs/WorkspaceSummaryDialog.qml
+++ b/resources/qml/Dialogs/WorkspaceSummaryDialog.qml
@@ -67,7 +67,7 @@ UM.Dialog
 
             ScrollBar.vertical: UM.ScrollBar
             {
-                parent: scroll
+                parent: scroll.parent
                 anchors
                 {
                     top: parent.top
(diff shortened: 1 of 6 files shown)


nallath:
Hmm. We figured that this was caused by a QT bug that is also fixed in 6.2, but I don't think that was tested yet.

But since this is is a pretty simple change in regards to the code, I don't see a problem with adding this regardless.

Ghostkeeper:
The Qt bug Nallath was referring to is this one: https://bugreports.qt.io/browse/QTBUG-89126

PR #11705, opened 2022-03-24
Disable the variable line strategy option in the UI CURA-9069

pkuiper-ultimaker:
Relates to PP-143
Relates to CURA-9069


resources/definitions/fdmprinter.def.json
```diff
@@ -1083,6 +1083,8 @@
                     "label": "Variable Line Strategy",
                     "description": "Strategy to use to print the width of a part with a number of walls. This determines how many walls it will use for a certain total width, and how wide each of these lines are. \"Center Deviation\" will print all walls at the nominal line width except the central one(s), causing big variations in the center but very consistent outsides. \"Distributed\" distributes the width equally over all walls. \"Inward Distributed\" is a balance between the other two, distributing the changes in width over all walls but keeping the walls on the outside slightly more consistent.",
                     "type": "enum",
+                    "visible": false,
```
nallath, line 1086:
```suggestion
```
No need to also set the visibility to false. Because enabled is false, it's already hidden.

PR #11709, opened 2022-03-24
[CURA-9036] Revert to normal deviation error value.

rburema:
Revert "Set the default value of the Maximum Extrusion Area Deviation to 50.000um2 and remove the warning levels."

This reverts commit 3c4c91947e4eed82babe58550316789ffcde2444.

See also backend PR: https://github.com/Ultimaker/CuraEngine/pull/1621


nallath:
This still leaves "meshfix_maximum_extrusion_area_deviation" set to 50.000 in ultimaker.def.json. Is that intentional?

PR #11718, opened 2022-03-28
PauseAtHeight.py postprocessing script improvments

Cycov:
- Change M18 to M84 when disarming
- Add option to keep steppers engaged for 1 hour (when changing filament for example)
- Add beep option
- Allow newlines in "Additional GCODE" fields


rburema:
Hi! Sorry for the wait. I put the relative/absolute back, as it doesn't do any harm and we do allow for all flavours of gcode to be set into absolute or relative mode. Thanks for your contribution!

PR #11721, opened 2022-03-28
voron2_base.def.json: Do not limit travel speeds to 300mm/s

Haxk20:
VORON V2.4 is an fast machine. There is no reason to limit
the user to 300mm/s. Unlock these speeds by removing maximum_value
which makes cura fallback to machine_max_feedrate_x/y which are
set to the C (299,792,458).

Also set the warning to 301 (As requested by Fulg) as
anything above 300 means beyond the "standarts".

### DIFF
--- a/resources/definitions/voron2_base.def.json
+++ b/resources/definitions/voron2_base.def.json
@@ -107,7 +107,7 @@
         "retraction_combing":                           { "value": "'noskin'" },
         "retraction_combing_max_distance":              { "default_value": 10 },
         "travel_avoid_other_parts":                     { "default_value": false },
-        "speed_travel":                                 { "maximum_value": 300, "value": 300, "maximum_value_warning": 501 },
+        "speed_travel":                                 { "value": 300, "maximum_value_warning": 501 },
         "speed_travel_layer_0":                         { "value": "math.ceil(speed_travel * 0.4)" },
         "speed_layer_0":                                { "value": "math.ceil(speed_print * 0.25)" },
         "speed_wall":                                   { "value": "math.ceil(speed_print * 0.33)" },


Ghostkeeper:
It looks like the firmware limits it to 300mm/s: https://github.com/VoronDesign/Voron-2/blob/99276d51d2dc9aac2b8c9f368ad90dfdf115ac7a/firmware/klipper_configurations/Octopus/Voron2_Octopus_Config.cfg#L34

That's probably why this was there. It may be a fast printer, but if the printer will limit it to 300mm/s, it's good to limit it that way in Cura as well.

hifihedgehog:
> It looks like the firmware limits it to 300mm/s: https://github.com/VoronDesign/Voron-2/blob/99276d51d2dc9aac2b8c9f368ad90dfdf115ac7a/firmware/klipper_configurations/Octopus/Voron2_Octopus_Config.cfg#L34
> 
> That's probably why this was there. It may be a fast printer, but if the printer will limit it to 300mm/s, it's good to limit it that way in Cura as well.

Please ask the Voron devs (CC: @Maglin, @stvptrsn, @FHeilmann, @nemgrea). Most of them would probably agree that it is safe with the latest 2.4r2 LDO kits to raise this limit. Besides, the firmware overrides the gcode if there is an issue.

Haxk20:
> > It looks like the firmware limits it to 300mm/s: https://github.com/VoronDesign/Voron-2/blob/99276d51d2dc9aac2b8c9f368ad90dfdf115ac7a/firmware/klipper_configurations/Octopus/Voron2_Octopus_Config.cfg#L34
> > 
> > That's probably why this was there. It may be a fast printer, but if the printer will limit it to 300mm/s, it's good to limit it that way in Cura as well.
> 
> Please ask the Voron devs (CC: @Maglin, @stvptrsn, @FHeilmann, @nemgrea). Most of them would probably agree that it is safe with the latest 2.4r2 LDO kits to raise this limit. (Likewise with acceleration, where most users run well above 3000 mm^2, which by the way is set to just 1500 mm^2 which is wrong in the current iteration of the Voron 2's base def.json file.) Speaking from experience in the Voron community, most users change their configuration file after initial setup. Besides, if there were an issue, the firmware configuration overrides the gcode if this limit did pose a risk

Exactly this.

The main reason i PRed this is that we have the safe limits in klipper config. Those are always easily accessible but what good are they for when the slicer is heavily limiting you on the speed settings. Which i had to spend a good while to see why it was not allowing the requested speeds and even bigger pain to find this file on linux system. Its just not as easy to find and change.

And going even further. The cura print speeds are already unlocked. Why unlock the setting which you will usually set around 60-300mm/s and limit travel speeds which can easily go from 100-1000mm/s.

fieldOfView:
I would recommend to just *remove* the `maximum_value`, so the maximum falls back to the maximum defined in the `fdmprinter.def.json` file, which refers to the `machine_max_feedrate_x` and `machine_max_feedrate_y`, both of which default to the same speed of light you want to limit it to. The definition should not have "hardcoded" a `maximum_value` in the first place, but specify a `machine_max_feedrate_x` and `machine_max_feedrate_y` (which would have been editable with the Printer Settings plugin for users that think the maximum value is incorrect).

The `maximum_value_warning` should in my opinion be kept at a sane maximum value.  I think a warning of any speed above the speed of sound (or actually well below that) is in order.

Fulg:
My personal opinion (not talking on behalf of VORON Design!) would be to _at least_ remove the `maximum_value`, and fix `maximum_value_warning` to its intended speed of `301`. Although I suppose we could leave it at `501`...

At the very least Cura should not refuse to slice if you choose to travel faster, if your number is too high the firmware will just refuse to honor it, and your print time estimates will be inaccurate. I don't remember why I added a `maximum_value` back then, sorry about that.

Haxk20:
Commit was interactively rebased to include requested changes

Fulg:
After discussing with some of the VORON team, it looks like keeping `maximum_value_warning` at `501` would be best, it just never worked due to `maximum_value` being set at `300`. Current V2.4 machines can do higher than 300mm/s (as mentioned initially here), stock hardware these days start having issues going past 500-600mm/s of travel speed.

Haxk20:
> After discussing with some of the VORON team, it looks like keeping `maximum_value_warning` at `501` would be best, it just never worked due to `maximum_value` being set at `300`. Current V2.4 machines can do higher than 300mm/s (as mentioned initially here), stock hardware these days start having issues going past 500-600mm/s of travel speed.

Rebased to change it back to 501 and explained why.

nallath (approved):

PR #11724, opened 2022-03-29
Adding printer da Vinci Pro EVO

heed818:
Adding printer da Vinci Pro EVO


heed818:
Hi Sir @Ghostkeeper ,do you think these request could able merge to version 5.1 ?
Thanks.

nallath (approved):

PR #11725, opened 2022-03-29
Feat/Fix: transition filter limit

BagelOrb:
Frontend settings for https://github.com/Ultimaker/CuraEngine/pull/1627

Contributes to CURA-9080

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1144,8 +1144,8 @@
                     "description": "If it would be transitioning back and forth between different numbers of walls in quick succession, don't transition at all. Remove transitions if they are closer together than this distance.",
                     "type": "float",
                     "unit": "mm",
-                    "default_value": 1.4,
-                    "value": "4 * math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x",
+                    "enabled": false,
+                    "default_value": 100,
                     "minimum_value": "wall_transition_length",
                     "minimum_value_warning": "math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x",
                     "maximum_value_warning": "10 * math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x"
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -1150,6 +1150,18 @@
...
                     "maximum_value_warning": "10 * math.cos(wall_transition_angle / 180 * math.pi) * wall_line_width_x"
                 },
+                "wall_transition_filter_deviation":
+                {
+                    "label": "Wall Transition Filter Deviation",
+                    "description": "Prevent Wall Transition Filtering from inducing errors on the extrusion width larger than this value.",
+                    "type": "float",
+                    "unit": "mm",
+                    "default_value": 0.1,
+                    "value": ".25 * machine_nozzle_size",
```
Ghostkeeper, line 1160: Do we really need this to be a setting? In what cases would you want to increase or reduce this setting, where that wouldn't be acceptable for a different print? What if we simply hard-code this to 25% nozzle size in CuraEngine, would anyone be bothered by it?

After discussing this with colleagues, we think this solution is best:
* If this is something that can be argued that it really needs to be different for different prints, we'd like to come up with a better name.
* If this is something that is going to be tweaked before the release, we'd like to make this a hidden setting until it is tweaked, to be removed later.
* If 25% of nozzle size is fine for pretty much all cases, we'd like to hard-code it in CuraEngine.

We hope this will make it more clear to use, both for the user and for ourselves in the future.

Ghostkeeper (changes requested):

Ghostkeeper:
The failing tests are due to a changed build environment on the CI server, not due to the changes proposed here. They can be ignored for now (pending CURA-8640).

BagelOrb:
I’ve discussed with Rijk and we think that it makes sense to have this as a configurable setting. 
Some flexible filaments are bad at line width variation and lower layer height also have more back pressure, which means that the flow equalization is less accurate there.
Therefore, it makes sense to reduce the extra induced line width variation by this setting for such situations. This could be included in the profiles at some point in the future.

We had a bit of a brainstorm about the name and description of the setting and decided to change it to what you see in the PR.

Also, we think that now that we have filtering based on line width variation, we don't need filtering based on length any more. We therefore disabled that setting (`wall_transition_filter_distance`) in the frontend. We believe that this setting should be removed from the backend as well, at some point in the future.

PR #11734, opened 2022-03-30
Create sunlu_T3.def.json

Chrystal-1024:


nallath:
Please create a single PR with all the files needed for a machine. We run several tests on them and those only work if they are in the same pull request.

PR #11744, opened 2022-03-31
automatically remove all holes when spiralizing

BagelOrb:
This PR automatically enables Remove All Holes when Spiralize Outer Contour is enabled.

Vase mode is meant for models which are solid, but many people don't know and model their things hollow.
With this automatic setting change these people won't notice that they have modeled it wrong.
Which means there are less times I need to explain stuff to users.
Profit.


BagelOrb:
New default behavior when spiralize is turned on:
image
image

image
image

BagelOrb:
Old behavior (viewer discretion advised):
image

image

BagelOrb:
@ devs CURA-9532

PR #11745, opened 2022-03-31
Better default for the experimental bridge settings. See PP-3 for det…

pkuiper-ultimaker:
This PR implements improved default settings for the experimental bridge settings. 

See ticket PP-3 for details on the rational.


resources/definitions/ultimaker.def.json
```diff
@@ -82,6 +82,67 @@
         },
         "xy_offset": {
             "value": "-layer_height * 0.2"
+        },
+        "bridge_settings_enabled": {
+            "value": false
```
nallath, line 87: But that's the default, right?

pkuiper-ultimaker: Yep, for this pull request we decided to keep the bridge settings off. This improves the settings when people enable it.
We have a ticket open to get bridge settings out of experimental. When this happens we will enable it by default in our profiles.

PR #11754, opened 2022-04-02
Add new printer

Chrystal-1024:


Chrystal-1024:
@nallath Please let me know whether added new printer successfully.

PR #11792, opened 2022-04-12
Cura 8640 PyQt6 upgrade

jellespijker:
In order to get this to work on our build-system and working for all three OSes we did a shit tons of boy scouting in our cmake. We removed old methods with variables and try to be consisted in a target-based approach. The idea is that we don't patch stuff down the line, but that the install should place everything in the correct path in a uniform way across all of Cura's dependencies. Most of these changes are in the other PR's. The CMakeLists.txt changes here are for consistency overal.

> The other changes to qml and in most of the Python side I have were made by @rburema , @casperlamboo @Joeydelarago @nallath and @rburema Maybe they can supplement this comment.

# Part of
- ultimaker/libarcus#134
- ultimaker/pynest2d#17
- ultimaker/libsavitar#41
- ultimaker/libnest2d#3
- ultimaker/libcharon#40
- ultimaker/cura-binary-data#17
- ultimaker/curaengine#1639
- ultimaker/uranium#804
- ultimaker/cura-build-environment#126
- ultimaker/cura-build#280
- ultimaker/cura-rundeck-jobs#1

# Fixes
- 

# Todo
- [ ] Update Readme. But maybe best done in a separate ticket to prioritize with our other work for the 5.0 release.
- [ ] Make sure we us the new flag namespace consistently See https://www.pythonguis.com/faq/pyqt6-vs-pyside6/ and https://www.pythonguis.com/faq/pyqt6-vs-pyside6/ 
- [ ] Fix failing CI/CT due to out date build environment. I think it will be best if we fix this in a separate ticket.


resources/qml/TableView.qml
```diff
@@ -1,7 +1,7 @@
 //Copyright (C) 2022 Ultimaker B.V.
 //Cura is released under the terms of the LGPLv3 or higher.
 
-import Qt.labs.qmlmodels 1.0
+//import Qt.labs.qmlmodels 1.0
```
Ghostkeeper, line 4: Might as well remove this entirely, since the module is gone.
```suggestion

```

plugins/ModelChecker/ModelChecker.qml
```diff
@@ -14,7 +14,7 @@ UM.SimpleButton
 
     width: UM.Theme.getSize("save_button_specs_icons").width
     height: UM.Theme.getSize("save_button_specs_icons").height
-    iconSource: "model_checker.svg"
+//    iconSource: "model_checker.svg"
```
Ghostkeeper, line 17: We should still have an icon on this button, right?

Looks like this property still exists and everything, so I'll suggest just uncommenting it. But I haven't tested that and there was probably a reason why this was disabled.
```suggestion
    iconSource: "model_checker.svg"
```

Ghostkeeper (changes requested):
There's still some TODOs in this code with the two dropshadows. But I think that is kept out of scope for this change.

Ghostkeeper (approved):

PR #11816, opened 2022-04-19
[CUBICON] Add model profile (Optimus-C23Z)

HUNIBEST-HyVISION:
Add model profile (Optimus-C23Z)


HUNIBEST-HyVISION:
The request will be canceled due to a re-request.

PR #11817, opened 2022-04-19
Account sync was broken

nallath:
That was a big oversight on our end! We removed the old plugin, but outright forgot to port the syncing logic.

This does introduce a bit of duplication, as we have a "LicenceDialog" and a "MultipleLicenceDialog". But seeing as we're so close to a release, I've opted to not take the more dangerous route of making sure that only one is used. We could consider fixing this in the future, but it's not too bad as far as I'm concerned (since it's also contained in a single plugin)

CURA-9146

### DIFF
--- a/plugins/Marketplace/CloudApiModel.py
+++ b/plugins/Marketplace/CloudApiModel.py
@@ -1,0 +1,29 @@
+from typing import Union
+
+from cura import ApplicationMetadata
+from cura.UltimakerCloud import UltimakerCloudConstants
+
+
+class CloudApiModel:
+    sdk_version = ApplicationMetadata.CuraSDKVersion  # type: Union[str, int]
+    cloud_api_version = UltimakerCloudConstants.CuraCloudAPIVersion  # type: str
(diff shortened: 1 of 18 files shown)


Ghostkeeper (changes requested):
A lot of these files are missing the license and copyright header.

Because these changes are small and you've seen these issues more often, I'll just fix them up myself and continue with testing.

Ghostkeeper (approved):

Ghostkeeper:
Upon logging in and syncing it currently opens a pop-up about materials being incompatible with my Cura version (a problem with the Marketplace, not in Cura). This also puts the following error in the log several times, probably once for every package that doesn't have an icon:

```[MainThread] UM.Qt.QtApplication.__onQmlWarning [431]: file:///home/ghostkeeper/Projects/Cura/plugins/Marketplace/resources/qml/CompatibilityDialog.qml:109:29: QML Image: Cannot open: file:///home/ghostkeeper/Projects/Cura/plugins/Marketplace/images/placeholder.svg```

PR #11821, opened 2022-04-19
Check multiple settings per update of event loop

nallath:
It seems that in Qt5, the update loop was triggered a lot more often than in qt6.
Since we used that to handle the error checking, this made it so that you sometimes had to wait
for quite some time untill the slicing would start (as that is blocked by the error checker)

### DIFF
--- a/cura/Machines/MachineErrorChecker.py
+++ b/cura/Machines/MachineErrorChecker.py
@@ -53,6 +53,8 @@ class MachineErrorChecker(QObject):
 
         self._keys_to_check = set()  # type: Set[str]
 
+        self._num_keys_to_check_per_update = 10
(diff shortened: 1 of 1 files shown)


cura/Machines/MachineErrorChecker.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, parent: Optional[QObject] = None) -> None:
 
         self._keys_to_check = set()  # type: Set[str]
 
+        self._num_keys_to_check_per_update = 10
```
jellespijker, line 56: Magic number, please add a comment describing the rational behind `10`

jellespijker (approved):

PR #11906, opened 2022-04-23
[5.0b1] Fix creating a single copy of a model

### LINKED ISSUE #11897: Multiply Selected x1 does not work
### Application Version

5.0.0-Beta+1

### Platform

Windows 10

### Printer

Tronxy X5SA Pro and Creality Ender 3

### Reproduction steps

Place single STL on buildplate
Right click and select "Multiply Selected" 
Leave quantity at 1 - Select Okay

### Actual results

Result: no new object

### Expected results

Object is copied

- [ ] Log file
- [ ] Project file

If you change the quantity to anything other than 1 it works correctly even if you return the value to 1

fieldOfView:
This PR fixes the Multiply selected models when the number of copies is left to 1. The `value` of the spinbox does not get initialised to 1, so even though the label says 1, the value is 0. The command was making 0 copies.

Fixes #11897

### DIFF
--- a/resources/qml/Menus/ContextMenu.qml
+++ b/resources/qml/Menus/ContextMenu.qml
@@ -134,6 +134,7 @@ Cura.Menu
                 from: 1
                 to: 99
                 width: 2 * UM.Theme.getSize("button").width
+                value: 1
             }
         }
     }


MariskaMaas:
Awesome! Thanks. For internal reference CURA-9081

Joeydelarago:
Nice Fix!

PR #11917, opened 2022-04-23
Fix thumbnail format

tapir:
All slicers I have tried uses an "x" instead of space in between width and height.
Also DWC (Duet web control) can't view thumbnails without the "x" character.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/CreateThumbnail.py
+++ b/plugins/PostProcessingPlugin/scripts/CreateThumbnail.py
@@ -37,7 +37,7 @@ class CreateThumbnail(Script):
 
         encoded_snapshot_length = len(encoded_snapshot)
         gcode.append(";")
-        gcode.append("; thumbnail begin {} {} {}".format(
+        gcode.append("; thumbnail begin {}x{} {}".format(
             width, height, encoded_snapshot_length))
 
         chunks = ["; {}".format(encoded_snapshot[i:i+chunk_size])


Coolit1966:
Hello everyone I hope I'm in the right place I didn't know how or where to start a new question, But anyway I'm new to the 3D printing world and have jumped in with both feet and I'm still learning I have a Entina X40 IDEX printer which is basically a Weedo X40 IDEX printer knockoff and where Weedo comes from I have no clue but when I bought my printer it came with Cura 4.10 that supported my printer it would also let me preview what I was about to print but ever since me doing the updates even though the printer is still supported for some reason I can no longer see the previews when selecting what I want to print I was hoping someone would be able to walk me through the process to fix this issue please assume I don't know anything about computers since I had my heart attach my memory was a affected and I have forgotten a lot of stuff and following certain directions is a bit tuff for me  some times but I can manage so any help would be much appreciated.
Thanks

x40-Community:
> Hello everyone I hope I'm in the right place I didn't know how or where to start a new question, But anyway I'm new to the 3D printing world and have jumped in with both feet and I'm still learning I have a Entina X40 IDEX printer which is basically a Weedo X40 IDEX printer knockoff and where Weedo comes from I have no clue but when I bought my printer it came with Cura 4.10 that supported my printer it would also let me preview what I was about to print but ever since me doing the updates even though the printer is still supported for some reason I can no longer see the previews when selecting what I want to print I was hoping someone would be able to walk me through the process to fix this issue please assume I don't know anything about computers since I had my heart attach my memory was a affected and I have forgotten a lot of stuff and following certain directions is a bit tuff for me some times but I can manage so any help would be much appreciated. Thanks

If you want use Ultimaker Cura 5.X with the Weedo X40 or Etina X40, than you need the configuration and Cura plugin from the X40-Community. The Weedo Cura 4.10 version is a hacked version and doesn't have much to do with Ultimaker Cura anymore!

PR #12007, opened 2022-04-28
[CURA-9214] Can't export profiles, also fixes exporting materials

Joeydelarago:
selectedNameFilter is now an object instead of a string. The nameFilter has to be manually pulled out using the index in selectedNameFilter.

If only one filter is supplied to nameFilters, the index will always be -1. I assume this is because we are not selecting the file type in the Dialog, it just defaults to the only item. The workaround I've written will work even if this behaviour changes. 

FileDialog now uses currentFolder instead of folder.

CURA-9214

### DIFF
--- a/resources/qml/Preferences/Materials/MaterialsPage.qml
+++ b/resources/qml/Preferences/Materials/MaterialsPage.qml
@@ -246,7 +246,7 @@ UM.ManagementPage
                         break;
                 }
                 messageDialog.open();
-                CuraApplication.setDefaultPath("dialog_material_path", folder);
+                CuraApplication.setDefaultPath("dialog_material_path", currentFolder);
             }
         }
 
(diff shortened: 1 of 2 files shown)


resources/qml/Preferences/Materials/MaterialsPage.qml
```diff
@@ -259,7 +259,9 @@ UM.ManagementPage
             currentFolder: CuraApplication.getDefaultPath("dialog_material_path")
             onAccepted:
             {
-                const result = Cura.ContainerManager.exportContainer(base.currentItem.root_material_id, selectedNameFilter, selectedFile);
+                var nameFilterString = selectedNameFilter.index >= 0 ? nameFilters[selectedNameFilter.index] : nameFilters[0]
```
casperlamboo, line 262:
```suggestion
                const nameFilterString = selectedNameFilter.index >= 0 ? nameFilters[selectedNameFilter.index] : nameFilters[0];
```

resources/qml/Preferences/ProfilesPage.qml
```diff
@@ -354,8 +354,13 @@ UM.ManagementPage
             currentFolder: CuraApplication.getDefaultPath("dialog_profile_path")
             onAccepted:
             {
+
+                // If nameFilters contains only 1 item, the index of selectedNameFilter will always be -1
+                // This fetches the nameFilter at index selectedNameFilter.index if it is positive
+                var nameFilterString = selectedNameFilter.index >= 0 ? nameFilters[selectedNameFilter.index] : nameFilters[0]
```
casperlamboo, line 360:
```suggestion
                const nameFilterString = selectedNameFilter.index >= 0 ? nameFilters[selectedNameFilter.index] : nameFilters[0];
```

casperlamboo (approved):

PR #12049, opened 2022-05-01
Czech translations v5.0

sustmi:
I am finishing the translation. I have last few settings to translate. Then I will read all the new translations again and check for spelling errors and mistakes. But now I need to sleep on it.

It was really a big update this time. There are many new settings and some of them were quite hard to understand to me. I could not simply translate the texts without understanding what the settings really do. That's why it is taking so long.

Hopefully, I will finish it on 2nd of May.


sustmi:
The Czech translation of v5.0 is finished now.

PR #12055, opened 2022-05-02
[CURA-9232] Fix menubar not catching events

Joeydelarago:
The PrintSetupTooltip was catching events while invisible on the top left of the screen. This was causing the menubar not to be clickable.

I've disabled PrintSetupTooltip when not visible so that it does not catch events.

CURA-9232

### DIFF
--- a/resources/qml/PrintSetupTooltip.qml
+++ b/resources/qml/PrintSetupTooltip.qml
@@ -17,6 +17,8 @@ UM.PointingRectangle
     arrowSize: UM.Theme.getSize("default_arrow").width
 
     opacity: 0
+    // This should be disabled when invisible, otherwise it will catch mouse events.
+    enabled: false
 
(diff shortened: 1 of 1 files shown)


resources/qml/PrintSetupTooltip.qml
```diff
@@ -17,6 +17,8 @@ UM.PointingRectangle
     arrowSize: UM.Theme.getSize("default_arrow").width
 
     opacity: 0
+    // This should be disabled when invisible, otherwise it will catch mouse events.
+    enabled: false
```
fieldOfView, line 21: Alternatively, you could simply bind enabled and opacity together, like so:

`enabled: opacity != 0`

This makes the other two lines superfluous. I like one-line-fixes ;-)

nallath (approved):

PR #12113, opened 2022-05-06
[PP-168] [CURA-9293] Fix UM2+C buffer underruns

pkuiper-ultimaker:
Blobs were observed on the spiralized vase print on the UM2+C. After analysis we increased the resolution deviation to prevent microsegments. We found out that this is only needed for the spiralize mode, the quality was decreased for normal prints.

For details see: PP-168


rburema:
V&V team testing has determined that this is the last thing we will need in 5.0 to clear the 'blobs' blocker. I will be cherry-picking these commits (plus add brackets), to a new PR, since changing the base would result in pulling stuff into 5.0 from the main branch.

Once that is done, I will close (this version of!) this PR.

rburema:
Closed in favour of https://github.com/Ultimaker/Cura/pull/12186 (which wholly contains these commits, and/but is based on the 5.0 branch).

Issue #11941, opened 2022-04-24
(5.0) "Manage Printers" and "Machine Settings" dialog issues


GregValiant:
### Application Version

5.0.0 beta

### Platform

Windows 10 Home (onboard video OpenGL 4.0 in compatibility mode)

### Printer

Creality E3Pro

### Reproduction steps

The Enter (default?) and Cancel buttons are missing from both dialogs.  The "dismiss" X on the titlebar has to be used to close the dialog.  There is no indication if the settings have been accepted/saved and no way to discard any changes.

When selecting multiple lines of StartUp or EndGcode - the selected text is no longer highlighted to inform the user that they have indeed been selected.  The "Extruder" textboxes work as expected.

There is no right click menu to copy or paste in any text box in Machine Settings.  CTRL+C or CTRL+V work for copy and paste.  Previous versions suffered from this as well.

### Actual results

See above.

### Expected results

Normal dialog and text box interaction.

### Checklist of files to include

- [X] Log file
- [ ] Project file

### Additional information & file uploads

Log File
[cura.log](https://github.com/Ultimaker/Cura/files/8549932/cura.log)
.

GregValiant:
Additional:
When making changes to either StartUp gcode or Ending gcode, if the "close" X button is immediately clicked to dismiss the dialog then the users' changes are not made.
If the user clicks on any other area first (or tabs out of the text box) then the changes are made when the dialog is dismissed.

Vandresc:
Devs, see CURA-9188 for our internal reference.

GregValiant:
I am not able to export profiles.
Importing is OK.

fvrmr:
The ticket on our backlog is has been closed, since the close button and x should do the same. So that is why the close button is removed. However there is an issue with no saving the values. I created a ticket for that, which is linked to https://github.com/Ultimaker/Cura/issues/12021

About selecting text in the machine settings I will bring that up with the team. 

The right click menu is a duplicate of this: https://github.com/Ultimaker/Cura/issues/11579 and we have a ticket on our backlog for this.

GregValiant:
Thank you @fvrmr .

Madmike-1337:
There are about 30 tickets for this issue spread over several versions. It makes tuning, troubleshooting, and general setup of any Non-Ultimaker printer more difficult to get using. Why is it such a pain for the users and such a low priority for the devs?

nallath:
[...]

PR #12116, opened 2022-05-06
[5.0b1] Fix material "Print Settings" tab

fieldOfView:
This PR fixes setting values on the Print Settings tab of the Materials pane of the preferences. Without this PR, changes will not be saved at all. This PR also disables the input fields on the Print Settings tab for "read-only" materials.

Fixes https://github.com/Ultimaker/Cura/issues/12155

### DIFF
--- a/resources/qml/MachineSettings/NumericTextFieldWithUnit.qml
+++ b/resources/qml/MachineSettings/NumericTextFieldWithUnit.qml
@@ -38,6 +38,7 @@ UM.TooltipArea
 
     property alias textField: textFieldWithUnit
     property alias valueText: textFieldWithUnit.text
+    property alias enabled: textFieldWithUnit.enabled
     property alias editingFinishedFunction: textFieldWithUnit.editingFinishedFunction
 
     property string tooltipText: propertyProvider.properties.description ? propertyProvider.properties.description : ""
--- a/resources/qml/Preferences/Materials/MaterialsView.qml
+++ b/resources/qml/Preferences/Materials/MaterialsView.qml
@@ -601,6 +601,7 @@ Item
                     maximum: 99999
                     unitText: model.unit
                     decimals: model.unit == "mm" ? 2 : 0
+                    enabled: base.editingEnabled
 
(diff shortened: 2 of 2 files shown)


nallath (approved):

nallath:
CURA-9275

Madmike-1337:
This is not fixed. Version 5.1.0 still does not save changes without deliberate user action that is not just  closing the window.  I have been wrestling with this since cura 2.3 or something. A save button or apply button won't kill you. Neither would a note that tabbing to another pane is needed to put the users in control of the interface.

fieldOfView:
@Madmike-1337, the PR fixed a specific issue. That does not mean that there are no remaining issues. There is no use complaining about #11941 here.

nallath:
> A save button or apply button won't kill you. 

A normal way of reporting a bug would similarly also not kill you ;)

Madmike-1337:
> @Madmike-1337, the PR fixed a specific issue. That does not mean that there are no remaining issues. There is no use complaining about #11941 here.

They are the same issue, ergo not fixed.

Madmike-1337:
> > A save button or apply button won't kill you.
> 
> A normal way of reporting a bug would similarly also not kill you ;)

You're correct, it wouldn't kill me.  It also doesn't fix the problem. Sorry for riling feathers.

fieldOfView:
That is not the point. This, where you are posting, is _not_ an issue report. It is a pull request for changes I proposed to Cura, which fixed a different underlying issue (#12155). What you are talking about is already reported as an issue in #11941. 

Unless you think the changes made in this pull request broke things, discussion about your issue here is useless.

Madmike-1337:
> That is not the point. This, where you are posting, is _not_ an issue report. It is a pull request for changes I proposed to Cura, which fixed a different underlying issue (#12155). What you are talking about is already reported as an issue in #11941.
> 
> Unless you think the changes made in this pull request broke things, discussion about your issue here is useless.

Thank you for correcting me on this misconception.

PR #12145, opened 2022-05-09
Fix width of some drop-downs

Ghostkeeper:
The drop-down to select where to open files from had a problem where the items inside had the necessary width to fit their text, but that left space if there were other items next to it.

image

This is now fixed. All items have their width adjusted to fit the width of the pop-up. The pop-up takes on the width of the widest item inside. The code is taken from UM.Menu and works similarly, but with a different width property.

Another issue was fixed where the buttons in the account profile menu weren't as wide as the parent, and the text was wider than the buttons. This made the hover area a bit weird. That's fixed by simply making the buttons as wide as their parent.

### DIFF
--- a/plugins/PrepareStage/PrepareMenu.qml
+++ b/plugins/PrepareStage/PrepareMenu.qml
@@ -106,26 +106,34 @@ Item
                 {
                     id: openProviderColumn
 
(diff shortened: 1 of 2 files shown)


nallath (approved):

nallath:
Cherry-picked these to 5.0 as well.

PR #12147, opened 2022-05-09
hello could someone add the artillery hornet?

Medduus:
hello could someone add the artillery hornet? I could provide the profiles for cura


nallath:
Please don't use pull requests to do feature requests. That is not what they are for.

PR #12162, opened 2022-05-10
Remove extra resource path from Python file

Ghostkeeper:
This was recently added. However it appears that we have the same path multiple times in the resources, leading to the resources being found twice. This would normally get removed by virtue of being a set, but not this time since it's a different way of writing the same path.
For the themes, it finds the following paths (in my unpacked directory): `{'/home/trin/tempy/cura5/UM/../share/uranium/resources/themes', '/home/trin/tempy/cura5/cura/../share/cura/resources/themes', '/home/trin/tempy/cura5/share/cura/resources/themes', '/home/trin/.local/share/cura/5.0/themes', '/home/trin/tempy/cura5/share/uranium/resources/themes'}` For both Cura and Uranium, this points to /home/trin/tempy/cura5/curanium/resources/themes twice. So let's remove this one.

Contributes to issue CURA-9147.

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -349,7 +349,6 @@ class CuraApplication(QtApplication):
 
         app_root = os.path.abspath(os.path.join(os.path.dirname(sys.executable)))
         Resources.addSearchPath(os.path.join(app_root, "share", "cura", "resources"))
-        Resources.addSearchPath(os.path.join(os.path.abspath(os.path.dirname(__file__)), "..", "share", "cura", "resources"))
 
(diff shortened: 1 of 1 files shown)


daid:
Should `addSearchPath` just modify the path with `os.path.normpath` and `os.path.normcase`? that should fix any duplication in all cases.

(also, it's over 9000! that's what I actually wanted to say)

fieldOfView:
🙇‍♂️

Ghostkeeper:
I considered using `normpath` (didn't think of `normcase`) to remove the duplicates. But I figured we shouldn't unnecessarily feed it data either. We could implement both solutions.

casperlamboo (approved):

PR #12393, opened 2022-05-30
Pp 9 disable travel acceleration/jerk

rijkvanmanen:
enabling CURA-8708 for ultimaker printers. For more details see PP-9/PP-54


pkuiper-ultimaker (approved):
Nice that you also remove these jerk warning, was indeed annoying.

nallath (approved):

PR #12399, opened 2022-05-30
PP-86 reduce support interface flow

rijkvanmanen:
The reasoning of PP-86 also holds for the support interface layers, however we forgot about support when we implemented PP-86. When doing PP-108 we found this improvement.

### DIFF
--- a/resources/definitions/ultimaker.def.json
+++ b/resources/definitions/ultimaker.def.json
@@ -68,6 +68,9 @@
         "skin_material_flow": {
             "value": "0.97 * material_flow"
         },
+        "support_interface_material_flow": {
+            "value": "skin_material_flow"
+        },
         "skin_monotonic" : {
             "value": true
         },


resources/definitions/ultimaker.def.json
```diff
@@ -68,6 +68,9 @@
         "skin_material_flow": {
             "value": "0.97 * material_flow"
         },
+        "support_interface_material_flow": {
```
pkuiper-ultimaker, line 71: Skin material flow will become 95% when the other PP-93 is merged. This is because a top layer is defined to be 100% in that case. Is 95% enough to create a good (closed) interface layer?

pkuiper-ultimaker: I think 95% will still give a good enough interface surface. Let's merge it.

pkuiper-ultimaker (approved):

nallath (approved):

PR #12402, opened 2022-05-31
[CURA-8610] Save package metadata

Joeydelarago:
I've added functionality for exporting information about packages used in enabled extruder stacks. This metadata is stored inside Metadata/packages.json inside .3mf files. 

Finding the package that owns a material from only the material container is a bit difficult since we only store the material source file name.  I've added getMaterialFilePackageId inside CuraPackageManager to solve this issue, please suggest some extra comments if it is not clear.

### DIFF
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -53,7 +53,7 @@ class CuraPackageManager(PackageManager):
         super().initialize()
 
     def getMaterialFilePackageId(self, file_name: str, guid: str) -> str:
-        """Get the id of the material package that contains file_name"""
+        """Get the id of the installed material package that contains file_name"""
         for material_package in [f for f in os.scandir(self._installation_dirs_dict["materials"]) if f.is_dir()]:
             package_id = material_package.name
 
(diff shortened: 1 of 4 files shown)


plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -343,6 +343,9 @@ def serialize(self, ignored_metadata_keys: Optional[Set[str]] = None):
 
         return stream.getvalue().decode("utf-8")
 
+    def getFileName(self):
```
rburema, line 346:
```suggestion
    def getFileName(self) -> str:
```

cura/CuraPackageManager.py
```diff
@@ -51,6 +54,26 @@ def initialize(self) -> None:
...
+                    continue
+
+                with open(root + "/" + file_name, encoding="utf-8") as f:
+                    # Make sure the file we found has the same guid as our material
+                    # Parsing this xml would be better but the namespace is needed to search it.
+                    parsed_guid = XmlMaterialProfile.getMetadataFromSerialized(f.read(), "GUID")
+                    if guid == parsed_guid:
+                        return package_id
+                    continue
+
```
rburema, line 74: No `return` (or error handling) for the unhappy path.

PR #12412, opened 2022-06-01
PP-19 - Also increase jerk for support

pkuiper-ultimaker:
Prevent flow ringing in the support material and increase productivity.

Relates to PP-19


rijkvanmanen (changes requested):
Most of these settings are already default in fdm_printer. So this needs to be checked and cleaned further.

resources/definitions/ultimaker.def.json
```diff
@@ -154,6 +151,56 @@
         },
         "bridge_fan_speed_3": {
              "value": "cool_fan_speed_min"
+        },
+        "bridge_fan_speed_3": {
+             "value": "cool_fan_speed_min"
+        },
```
rijkvanmanen, line 157: the bridge_fan_speed_3 is now defined twice, seems to be a mistake

resources/variants/ultimaker3_bb0.8.inst.cfg
```diff
@@ -24,9 +24,6 @@ infill_pattern = ='zigzag' if infill_sparse_density > 80 else 'triangles'
 infill_wipe_dist = 0
 jerk_enabled = True
```
rijkvanmanen, line 25: we could bring the jerk_enabled and the raft_jerk also to ultimaker.def.json, right?

rijkvanmanen (changes requested):
Some more remarks:
* some variant bb04 profiles still contain definitions of jerk_support. Is that intended?
* the intent profiles still contain unnecessarily overwrites of jerk_infill, jerk_wall, etc.
* the quality profiles still contain unnecessarily overwrite of jerk_enabled, raft_jerk
* the pp profiles of the UM2+C and the UM2+ have a lot of jerk overwrites, I guess these can also be cleaned right?

So, this PR is growing into a more general jerk clean up, maybe update the title and description of this PR to reflect this a bit better.

resources/variants/ultimaker3_bb0.8.inst.cfg
pkuiper-ultimaker: Then I would need to disable it for the UM2 machines, which is a larger list of machines. (The fdmprinter default for enable jerk is false). Therefore I chose to enable jerk in UM3, S3 and S5 explicitly.

resources/definitions/ultimaker.def.json
pkuiper-ultimaker: Indeed , this seems to be a merge error. I will correct it.

resources/variants/ultimaker3_bb0.8.inst.cfg
rijkvanmanen: then bring it to the machine json (at least it should not be in the variant profiles, and it is in some quality profiles as well)

pkuiper-ultimaker: Hmm, I thought I removed them all.... I will clean this up as well.

nallath (approved):

PR #12415, opened 2022-06-01
Pp 170 clean up retraction limits

rijkvanmanen:


pkuiper-ultimaker (approved):

nallath:
The changes for PC seem to be really big. It used to be 80 retracts per mm of filament, which are now at 25.

Are we very sure that this won't cause issues? I get the impression that it was set to high for CPE on purpose.

rijkvanmanen:
A standard retraction cube (single wall, no infill) has around 15 retractions/mm. So 25 retractions/mm is basically no limit at all. Therefor 50 or 80 retractions/mm does not make sense as a limit.

PR #12442, opened 2022-06-04
Rename Ender-3 profile to include v2

amweiss:
Since the Ender 3 v2 is the same hardware profile as the Ender 3, make it clear in the profile name.


nallath (approved):

nallath:
Change is included in 5.2

PR #12465, opened 2022-06-07
PP-166 DD core variant files

Andi-ultimaker:
DD core variant files for S3 & S5 printers

Relates to: PP-166

Needs to be pulled together with the following pull request in the Cura repository: https://github.com/Ultimaker/fdm_materials/pull/205

### DIFF
--- a/resources/variants/ultimaker_s3_dd04.inst.cfg
+++ b/resources/variants/ultimaker_s3_dd04.inst.cfg
@@ -1,0 +1,45 @@
+[general]
+name = DD 0.4
+version = 4
+definition = ultimaker_s3
+
+[metadata]
+setting_version = 20
+type = variant
+hardware_type = nozzle
+
+[values]
+brim_width = 7
+machine_nozzle_cool_down_speed = 0.9
+machine_nozzle_id = DD 0.4
+machine_nozzle_size = 0.4
+raft_acceleration = =acceleration_print
+raft_airgap = 0.3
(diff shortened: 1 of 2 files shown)


rijkvanmanen (changes requested):
I think this PR needs a bit more of discussion. Why not only put the machine_nozzle_id, machine_nozzle_size, machine_nozzle_tip_outer_diameter in this file? The rest can be put in the quality profiles and has nothing to do with the nozzle, right? The same reasoning is followed in Colorado.

Andi-ultimaker:
> I think this PR needs a bit more of discussion. Why not only put the machine_nozzle_id, machine_nozzle_size, machine_nozzle_tip_outer_diameter in this file? The rest can be put in the quality profiles and has nothing to do with the nozzle, right? The same reasoning is followed in Colorado.
> 
> Please note that a PR to remove these profiles from private_data is also lacking.

I copied the cfg files from the CC 0.4 core, so whatever was there has been copied in.

Ghostkeeper:
The profiles are designed to allow concentrating setting overrides if they are common in all subprofiles of a configuration. If the nozzle profiles were meant to only have the metadata of the nozzle ID and diameter in them, I don't think we'd make them into profiles. Effectively this is saying that all profiles with DD cores use an initial layer speed of 20mm/s, so then I don't see the need to make that specific to to a specific profile.

Putting them here makes it easier to change for you.

We'll merge this in now though, since we have a feature freeze now.

Ghostkeeper (approved):

PR #12472, opened 2022-06-08
[CURA-8610] Fix warning when saving bundled packages

Joeydelarago:
Before this change packages were only being checked if they were bundled after loading the package_id. However only installed packages were being loaded so this would always be false.

The solution here is to check first if a material file is inside "secure_paths" (these are install directory resource paths). If it is, it must be a bundled material.

CURA-8610

### DIFF
--- a/cura/CuraPackageManager.py
+++ b/cura/CuraPackageManager.py
@@ -1,7 +1,9 @@
 # Copyright (c) 2018 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
+import glob
 import os
 
+from pathlib import Path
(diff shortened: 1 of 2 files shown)


cura/CuraPackageManager.py
```diff
@@ -55,6 +57,24 @@ def initialize(self) -> None:
 
         super().initialize()
 
+    def isMaterialBundled(self, file_name: str, guid: str):
+        """ Check if there is a bundled material name with file_name and guid """
```
nallath, line 61: I think you should also write down the documentation that you put into the PR here as well. The logic that a "secureSearchPath" is bundled is info that you would want here.

nallath (changes requested):

nallath (approved):

PR #12476, opened 2022-06-08
[CURA-9271] Build plate shrinks when using horizontal scaling shrinkage compensation

Joeydelarago:
We were previously both scaling down the buildplate and scaling up the disallowed areas of the models with shrinkage compensation. Doing both reduces the max usable area when scaling.

Screenshot 2022-06-08 at 15 37 33

Screenshot 2022-06-08 at 15 37 42

The solution is to do only one of these options.

Removing the disallowed area works fine when scaling models from their centre. However we scale 2 or more models from the buildplate centre rather than the model centre. This means that the models can go outside the buildplate without it being obvious from the visualisation.

The better solution is to remove the buildplate x/y scaling and keep the disallowed areas. This mean our preview can go above the max z shown in prepare, but this is the good solution without implementing 3d disallowed areas.

Screenshot 2022-06-08 at 14 08 39

Screenshot 2022-06-08 at 14 08 43

Removed shrinkage factor build volume scaling on x and y. This information is shown already by scaling the allowed areas around the objects.

Kept the bounding box scaling for z. This shows the max height a model can be. The objects disallowed area does not show this information.

Also removed scaling on non model disallowed areas on the build plate since this does not need to scale with the build plate anymore.

CURA-9271

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -565,8 +565,8 @@ class BuildVolume(SceneNode):
         self._updateScaleFactor()
 
         self._volume_aabb = AxisAlignedBox(
-            minimum = Vector(min_w, min_h - 1.0, min_d).scale(self._scale_vector),
-            maximum = Vector(max_w, max_h - self._raft_thickness - self._extra_z_clearance, max_d).scale(self._scale_vector)
+            minimum = Vector(min_w, min_h - 1.0, min_d),
+            maximum = Vector(max_w, max_h - self._raft_thickness - self._extra_z_clearance, max_d)
         )
 
         bed_adhesion_size = self.getEdgeDisallowedSize()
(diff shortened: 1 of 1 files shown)


Ghostkeeper (approved):

Ghostkeeper:
I found a bug in this. It was probably a bug for a longer time, but didn't manifest itself as a bug until this other fix.

The model can in some cases exit the build volume. The disallowed area grows properly, but if there is no border around the build plate this doesn't prevent the user from slicing. As a result, they get g-code that exits the build volume which can damage some printers.
image

This can only happen if the build adhesion and Avoid Objects While Printing are both disabled (and other features like draft shield too, but those aren't typically enabled by default). Here is a project file that reproduces the issue:
[shrinkage_compensation_exits_buildplate.zip](https://github.com/Ultimaker/Cura/files/8880267/shrinkage_compensation_exits_buildplate.zip)

I think the current fix is good for the user (so please don't opt for shrinking the build volume instead, then) but I think we should add a zero-width collision area around the edge then to prevent the object from going out of bounds.

casperlamboo (approved):

PR #12512, opened 2022-06-13
Show striped background when 'alternate version' enabled. 

rburema:
See (also) [backend PR](https://github.com/Ultimaker/Uranium/pull/831).

Also shows alternate version in header bar.

(Requirement for nightly builds.)

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -271,6 +271,10 @@ class CuraApplication(QtApplication):
     def ultimakerDigitalFactoryUrl(self) -> str:
         return UltimakerCloudConstants.CuraDigitalFactoryURL
 
+    @pyqtProperty(str, constant = True)
+    def extraVersionText(self) -> str:
+        return ApplicationMetadata.CuraBuildType if ApplicationMetadata.IsAlternateVersion else ""
+
     def addCommandLineOptions(self):
         """Adds command line options to the command line parser.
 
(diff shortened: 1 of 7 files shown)


fieldOfView:
Wouldn't this have been a great use of that `backgroundPattern` in the `headerBackground` that is still there as a placeholder? 
https://github.com/Ultimaker/Cura/blob/5.0/resources/qml/Cura.qml#L196-L205

This seems like a bit of a heavy-handed approach, having to add a provision in all the compositing shaders.

rburema:
@fieldOfView We can still change it. (Especially since the design was just off the cuff.) What do you think @jellespijker (cc bc. involved in the nightlies) @lorenzoromagnoli (UX/UI) ?

I don't think the change is that heavy compared to all the other stuff we do in the shaders (minor even), but I realize that might not be the best argument to keep adding to them.

rburema:
I discussed this with @lorenzoromagnoli and we're going to give the poor shader a rest, but (also) not involve the header any further. We'll add a yellow bar around the field instead.

rburema:
@fieldOfView I know my lack of theming is disturbing, but I also converted it to draft just now. This will all likely change anyway!

lorenzoromagnoli:
We are trying to solve 2 things:
- clearly explain that you are running a nightly build to our users
- Have a way to quickly identify in screenshots that the user is using a nightly build, even when taking a close-up of the UI 

I made a quick mockup that I believe could solve the 2 points above without being too disturbing and off-brand. 
Screen Shot 2022-06-15 at 11 36 00

rburema:
Are you saying I'm disturbing and off-brand? :-p :-p

lorenzoromagnoli:
image

rburema:
> How hard would it be to add the major.minor.patch-prerelasetag+buildmetadata as overlaid text on the build plate :-p

Not _that_ hard, we can already somewhat easily make textures from text due to Qt (for example the 'Enterprise'/build-version tag in the splash screen is a picture), from there we'd just have to get the texture to the shader and use it there.

fieldOfView:
image

PR #12528, opened 2022-06-15
[CURA-8849] Improve intent profile selection

Joeydelarago:
Add new design for intent profile selection. See designs for more information


resources/qml/PrintSetupSelector/Recommended/RecommendedQualityProfileSelectorButton.qml
```diff
@@ -0,0 +1,93 @@
...
+import UM 1.5 as UM
+import Cura 1.7 as Cura
+
+
+Rectangle
+{
+    id: base
+    height: 60
+    Layout.fillWidth: true
+    color: mouseArea.containsMouse || selected ? UM.Theme.getColor("um_blue_1") : UM.Theme.getColor("background_1")
```
casperlamboo, line 17: why not use `background_3` instead of `um_blue_1`. It is already in the theme so it has darkmode-support, and then `um_blue_1` can be removed from the themes.json.

resources/qml/PrintSetupSelector/Recommended/UnsupportedProfileIndication.qml
```diff
@@ -0,0 +1,54 @@
...
+            font: UM.Theme.getFont("default_bold")
+            text: catalog.i18nc("@error", "Configuration not supported")
+        }
+    }
+
+    UM.Label
+    {
+        width: parent.width
+
+        text: catalog.i18nc("@message:text %1 is the name the printer uses for 'nozzle'.", "No profiles are available for the selected material/%1 configuration. Please change your configuration."
```
casperlamboo, line 39: Should there be a space between material and the argument?

```suggestion
        text: catalog.i18nc("@message:text %1 is the name the printer uses for 'nozzle'.", "No profiles are available for the selected material /%1 configuration. Please change your configuration."
```

PR #12531, opened 2022-06-15
Cura 8555 link ufp to 3mf in digital library

casperlamboo:


rburema (approved):
Well done! I have a hard time coming up with anything negative :-)

(Maybe `currentIndex: UM.Backend.state == UM.Backend.Done ? 0 : 1` could have been replaced with `currentIndex: enabled ? 0 : 1` ... but that is almost more a matter of taste.)

PR #12532, opened 2022-06-15
Pp 166 clean up dd variant profile

rijkvanmanen:


rijkvanmanen:
Sorry, the commits are labelled as PP-116 but it should be PP-166.

Andi-ultimaker (approved):

Andi-ultimaker:
> 

Why do we want to add the removed settings to the BASF quality profile? The DD core variant was created independent of any material. Whatever values were there for the CC core were simply copier over.

pkuiper-ultimaker:
> > 
> 
> Why do we want to add the removed settings to the BASF quality profile? The DD core variant was created independent of any material. Whatever values were there for the CC core were simply copier over.

The CC core is also not OK and we need to fix all these variant files. We want to prevent that we run into issues with the BASF support material if we change/clean this in a later release. Therefore we directly adjusted the DD core files.

PR #12534, opened 2022-06-15
Enhance CITATION.cff

StuSerious:
Good afternoon!
Recently, for academic reasons, I started to look into citing Cura.
I noticed that the file was outdated, so here I propose some changes to bring it up to speed.

In addition to bumping CFF to `1.2.0` and Cura to `5.0.0` this PR also:
- fixes the GitHub integration (see before/after below)
BEFORE:
old_citation
AFTER:
new_citation
- fixes failing tests _and_ tests successfully with `cffconvert` (see screenshot below)
cff_test_ok
- adds some more relevant information, like:
  - links to the repo and license;
  - contact information (pulled straight from the website);
  - proper author crediting;
  - keywords.

The one thing that I purposefully did not touch are the .yml declarations at the top and bottom of the file.
I don't know why they are there, but they were there before and they don't break tests and integrations so I guess they are staying! 🤣

Please let me know if there is anything that needs changing or doing better.
I'm also open to feedback on adding (or removing) [more tags](https://github.com/citation-file-format/citation-file-format/blob/1.2.0/schema-guide.md), for now I limited this to what I would consider the bare minimum for this sort of project.

Thank you in advance,
_~S²_


nallath:
I just set this up a while ago when I read that Github supported this. In all honesty, I kind of forgot all about it. Whoops! 

Anyhow, it's clear that these changes are needed and make it easier to cite cura!

Out of curiosity, what are you citing Cura for?

StuSerious:
Thank you for the merge @nallath!

We're quite ready to put the work out there in the public right now, but I'll make sure to open an issue here once the work is published! (If you have any idea of a better place to post/make available for the community do let me know!)

Anyhow, I have every intention of sharing everything for free and without any of those pesky journal paywalls, but once the work is finished and peer reviewed!

If you want to get the gist of what we're up to, you can always [check our website out](https://www.3d4med.eu/)!

PR #12561, opened 2022-06-19
ADD printer wanhao D12

Alakaresh:
[Duplicator 12 Cura5.zip](https://github.com/Ultimaker/Cura/files/8935884/Duplicator.12.Cura5.zip)


nallath:
If you want to add a specific printer, please create a pull request that actually does that.

PR #12574, opened 2022-06-21
Create definition for Creality CR-5 Pro

TheBinaryLoop:


nallath (changes requested):
This doesn't seem to be a full machine definition. When adding this to cura, it seems to fail to find variants. 
image

Please have a look at https://github.com/Ultimaker/Cura/wiki/Adding-new-machine-profiles-to-Cura and  https://github.com/Ultimaker/Cura/wiki/Profiles-&-Settings

jellespijker:

PR #12575, opened 2022-06-21
Create definition for Creality CR-5 Pro H

TheBinaryLoop:


nallath (changes requested):
This profile has the same issue as https://github.com/Ultimaker/Cura/pull/12574

jellespijker:

PR #12585, opened 2022-06-22
[CURA-9412] Project file saving fails on printers without materials

Joeydelarago:
Add check for empty material instance containers before trying to fetch material metadata. This fixes failing to save a project on a printer with no materials (UM2 for example).

CURA-9412

### DIFF
--- a/plugins/3MFWriter/ThreeMFWriter.py
+++ b/plugins/3MFWriter/ThreeMFWriter.py
@@ -12,6 +12,7 @@ from UM.Application import Application
 from UM.Message import Message
 from UM.Resources import Resources
 from UM.Scene.SceneNode import SceneNode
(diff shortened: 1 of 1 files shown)


plugins/3MFWriter/ThreeMFWriter.py
```diff
@@ -268,6 +269,10 @@ def _getMaterialPackageMetadata() -> List[Dict[str, str]]:
                 # Don't export materials not in use
                 continue
 
+            if type(extruder.material) is EmptyInstanceContainer:
```
nallath, line 272:
```suggestion
           if isinstance(container, type(ContainerRegistry.getInstance().getEmptyInstanceContainer())):
```

In other places we do it like this, so i think it makes sense to use the same way of doing it.

nallath (changes requested):

PR #12599, opened 2022-06-24
Don't show scary error message when saving custom file

casperlamboo:
CURA-9414


fieldOfView:
A log message will be more visible than a comment in the code.

casperlamboo:
We decided we didn't want to show the error as it is expected behavior when saving a custom material.

rburema (approved):

PR #12600, opened 2022-06-24
[CURA-9401] Toolbar visible on monitor page

Joeydelarago:
Fix a bug where the toolbar was showing on the monitor screen. The fix is to have a boolean in CuraStage objects which says whether they should show the toolbar. I think this will be useful for plugin developers.

The default is not to show the toolbar now. I expect this will not change the functionality of plugins, however if a reviewer thinks a default of true is more sensible please suggest the change.

### DIFF
--- a/cura/Stages/CuraStage.py
+++ b/cura/Stages/CuraStage.py
@@ -15,6 +15,8 @@ from UM.Stage import Stage
 class CuraStage(Stage):
     def __init__(self, parent = None) -> None:
         super().__init__(parent)
+        # True if the toolbar should be visible when this stage is active
+        self._toolbarEnabled = False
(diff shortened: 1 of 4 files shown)


fieldOfView:
I am not sure if I like the default to false.

nallath:
Why not just change the draw order? That's a much simpler way of doing this...

nallath (changes requested):
A much simpler solution is to remove the `z: 1` property from the toolbar qml element.

nallath:
Closing this in favor of https://github.com/Ultimaker/Cura/commit/bafffa994c5b8e0bf0782a0d177456bc26b58560

PR #12651, opened 2022-07-01
[CURA-9357] Remove some very difficult to explain (even to experts) settings.

rburema:
It'll just confuse users. It confused _us_ (both developers and materials experts) a lot of the time :-D

See backend PR: https://github.com/Ultimaker/CuraEngine/pull/1689


Ghostkeeper:
During refinement we discussed this and said we would actually remove them, not just hide them. We are never going to unhide them so we should remove them in order to keep our code base simpler and more maintainable.

Ghostkeeper (changes requested):
Please remove these settings and move the translation code to CuraEngine.

casperlamboo (approved):

PR #12663, opened 2022-07-04
Updated pt_BR strings for Cura 5.1

Patola:


resources/i18n/pt_BR/cura.po
```diff
@@ -4051,28 +3913,27 @@ msgstr "Biblioteca de recorte de polígonos"
 #: /home/trin/Gedeeld/Projects/Cura/resources/qml/Dialogs/AboutDialog.qml:155
 msgctxt "@label"
 msgid "JSON parser"
-msgstr ""
+msgstr "parser JSON"
```
Ghostkeeper, line 3916: We're going to add a capital letter here. You couldn't have known that from the original text, but it's consistent if it would have capital letters.

Ghostkeeper (commented):
Looks good, thanks!

I'll make one correction.

PR #12678, opened 2022-07-06
Adding new Naxe machine into 5.1

Naxecorp:
Adding new machine from Naxe manufacturer
machine def
extruder def
quality
materials


nallath:
I don't know what you're trying to do, but this is not correct. If you want to add a new machine, please create a fork and create a PR from there.

PR #12697, opened 2022-07-07
Add Naxe machines, quality and materials

Naxecorp:
Hello there, 

Here is some files for adding Naxe manufacturer in the Cura software.
We are based in France. 

Thank you all guys


resources/materials/naxe_ABS.xml.fdm_material
```diff
@@ -0,0 +1,30 @@
+<?xml version='1.0' encoding='utf-8'?>
```
nallath, line 1: Materials need to go in https://github.com/ultimaker/fdm_materials

resources/materials/naxe_ABS.xml.fdm_material
```diff
@@ -0,0 +1,30 @@
+<?xml version='1.0' encoding='utf-8'?>
+<fdmmaterial xmlns="http://www.ultimaker.com/material" xmlns:cura="http://www.ultimaker.com/cura" version="1.3">
+  <metadata>
+    <name>
+      <brand>Naxe</brand>
```
nallath, line 5: Is this really a fundamentally different material? It seems like it's better to just add machine settings to generic 1.75 ABS instead

nallath (changes requested):
The STL is also too big. We require buildplate meshes to be well below 1mb.

Naxecorp:
Thank you!
please have a look to my new commit ;)

nallath:
Apologies! I've noticed another mistake while i was testing this. Because you set the material_print_temperature in the quality files to have a certain value (eg; a hardcoded one), it's no longer using whatever is defined in the material. 

So as a rule of thumb, it's better to use 
material_print_temperature = =default_material_print_temperature + 10 
If you want to use the default value of the material but print it at a higher temp.

Naxecorp:
Hi nallath, 
Thanks for reply, I have made the changes. Feel free to comment this out.

PR #12752, opened 2022-07-13
[CUBICON] Add model profile

HUNIBEST-HyVISION:
- Optimus-C23
- Prime-M22


resources/definitions/cubicon_common.def.json
```diff
@@ -5,7 +5,8 @@
     "metadata": {
         "author": "Cubicon R&D Center",
         "manufacturer": "Cubicon",
-        "visible": false
+        "visible": false,
+        "has_materials": true
```
nallath, line 9: This will require a version upgrade script! Without it, we can't accept this change.

resources/definitions/cubicon_style_neo_a22.def.json
```diff
@@ -7,7 +7,8 @@
         "manufacturer": "Cubicon",
         "visible": true,
         "file_formats": "text/x-gcode",
-        "supports_usb_connection": false,
+        "supports_usb_connection": true,
+        "supports_network_connection": true,
```
nallath, line 11: Where is the plugin that supports this?

nallath (changes requested):

resources/definitions/cubicon_common.def.json
HUNIBEST-HyVISION: I will upgrade the version and commit it again.

resources/definitions/cubicon_style_neo_a22.def.json
HUNIBEST-HyVISION: Settings for using the MKS WiFi plugin.

resources/definitions/cubicon_common.def.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": 2,
+    "version": 3,
```
nallath, line 2: That isn't right. The version here is about the schema version. It needs to be 2...

nallath: I think you misunderstood what I was trying to say. The changes that you made require a script to be written that handles upgrades. You can find examples of those scripts here: https://github.com/Ultimaker/Cura/tree/main/plugins/VersionUpgrade

HUNIBEST-HyVISION: I'll change the schema version back to 2.
There are many versions of the plugin version upgrade, which version should it be?

nallath: It's not the definition that needs to change, it's the extruder that need to change. An example of a change like that can be found in https://github.com/Ultimaker/Cura/blob/main/plugins/VersionUpgrade/VersionUpgrade41to42/VersionUpgrade41to42.py

HUNIBEST-HyVISION: I understand that VersionUpgrade is required to change the profile entry.
But I don't know how to do VersionUpgrade.
Is there a guide to refer to?

jellespijker:

casperlamboo:
Hi @HUNIBEST-HyVISION ,

Thanks for opening the PR and adding the Optimus-C23 and Prime-M22 printers to our repository. We did find some issues with your PR using our automatic printer linter. Please resolve the issues as suggested by this bot. Once the issues are resolved we will add your printer to our repository.

thanks,
Casper

PR #12795, opened 2022-07-19
Cura 9342 add missing openssl 1 1 1l shared library

jellespijker:
Make sure dynamic libraries such as OpenSSL 1.1.1l are added to the the AppImage


Ultimaker-Cura.spec.jinja
```diff
@@ -12,6 +13,13 @@ hiddenimports = {{ hiddenimports }}
 datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
 {% endfor %}
 
+# Add dynamic libs in the venv bin/Script Path. This is needed because we might copy some additional libs
+# e.q.: OpenSSL 1.1.1l in that directory with a separate:
+# `conan install openssl@1.1.1l -g deploy && cp openssl/bin/*.so cura_inst/bin`
+binaries.extend([(str(bin), ".") for bin in Path("{{ venv_script_path }}").glob("*.so*")])
+binaries.extend([(str(bin), ".") for bin in Path("{{ venv_script_path }}").glob("*.dll")])
+binaries.extend([(str(bin), ".") for bin in Path("{{ venv_script_path }}").glob("*.dylib")])
```
casperlamboo, line 21: I would avoid code duplication here
```suggestion
for file_type in ["*.so", "*.dll", "*.dylib"]
    binaries.extend([(str(bin), ".") for bin in Path("{{ venv_script_path }}").glob(file_type)])
```

jellespijker: Do what you think is best, but we can also change this after the release...

PR #12829, opened 2022-07-22
Layer.py: Remove whitespaces

digitalfrost:
Remove unnecessary whitespaces before and after brackets.
Don’t use spaces around the = sign when used to indicate a keyword argument.
(PEP8)

### DIFF
--- a/cura/Layer.py
+++ b/cura/Layer.py
@@ -72,7 +72,7 @@ class Layer:
         return self.createMeshOrJumps(False)
 
     # Defines the two triplets of local point indices to use to draw the two faces for each line segment in createMeshOrJump
-    __index_pattern = numpy.array([[0, 3, 2, 0, 1, 3]], dtype = numpy.int32 )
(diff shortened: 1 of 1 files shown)


cura/Layer.py
```diff
@@ -72,7 +72,7 @@ def createJumps(self) -> MeshData:
         return self.createMeshOrJumps(False)
 
     # Defines the two triplets of local point indices to use to draw the two faces for each line segment in createMeshOrJump
-    __index_pattern = numpy.array([[0, 3, 2, 0, 1, 3]], dtype = numpy.int32 )
+    __index_pattern = numpy.array([[0, 3, 2, 0, 1, 3]], dtype=numpy.int32)
```
Joeydelarago, line 75:
```suggestion
    __index_pattern = numpy.array([[0, 3, 2, 0, 1, 3]], dtype = numpy.int32)
```

https://github.com/Ultimaker/Cura/pull/12863#discussion_r929177475

Joeydelarago (changes requested):
We don't use PEP8 conventions for spacing around "=". I think we should, but better to have consistency.

PR #12830, opened 2022-07-22
Layer.py: Simplify if else statement

digitalfrost:


fieldOfView:
Have you considered the performance change? This change introduces a comparison for each polygon.

On a general note, please do provide a description for a PR; let us know how your change improves the code.

digitalfrost:
Yes, I did.
I thought the improved readability was worth the trade off and would facilitate further refactoring…

I also later noticed lines 91 and 93
where the polygons are looped over again you find the same if make_mesh else clause

I think the change I propose makes the code more consistent.

Joeydelarago:
We discussed and while this does shorten the code, given that this results in worse performance I'm going to have to reject the PR, unfortunately.

PR #12866, opened 2022-07-26
Hide cloud interaction buttons if the user has no permissions to them

Ghostkeeper:
Like Animal Farm taught us, some accounts have more access than others. For instance, they may be able to change print jobs only for their own jobs, not for others. Or they may be unable to read the print queue at all. This pull request is intended to hide buttons that would do things that the user doesn't have permission to do anyway.

It adds a request to the sync that requests the user's permissions. This list of permissions gets stored in the account and can be accessed via the Cura API, even from within QML.

Using this new property, the following changes are made:
* The print queue is hidden in the monitor page if the user can't read the print jobs.
* The "move to top" function is hidden if the user can't change other people's print jobs.
* The "pause" function is hidden if the user can't change other people's print jobs.
* The "abort" function is hidden if the user can't change that print job (their own or others, whichever applies).
* The "remove from queue" function is hidden if the user can't change that print job (their own or others).
* The context menu showing the above four options is hidden if all of them are hidden.
* The "manage in browser" link is hidden if the user can't see that management page in their browser.
* The option to sync materials with cloud printers redirects to the USB material syncing workflow if the user doesn't have permissions to sync via cloud.
* The application switcher no longer shows the links to the digital libraries, the print jobs or the printers if the user doesn't have permissions to see those respective pages.

Implements CURA-9220.

### DIFF
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -7,6 +7,7 @@ from PyQt6.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty, QTimer, py
 from PyQt6.QtNetwork import QNetworkRequest
 from typing import Any, Callable, Dict, List, Optional, TYPE_CHECKING
 
+from UM.Decorators import deprecated
 from UM.Logger import Logger
 from UM.Message import Message
 from UM.i18n import i18nCatalog
@@ -319,12 +320,14 @@ class Account(QObject):
 
         self._authorization_service.deleteAuthData()
 
+    @deprecated("Get permissions from the 'permissions' property", since = "5.2.0")
     def updateAdditionalRight(self, **kwargs) -> None:
         """Update the additional rights of the account.
         The argument(s) are the rights that need to be set"""
         self._additional_rights.update(kwargs)
         self.additionalRightsChanged.emit(self._additional_rights)
 
+    @deprecated("Get permissions from the 'permissions' property", since = "5.2.0")
     @pyqtProperty("QVariantMap", notify = additionalRightsChanged)
     def additionalRights(self) -> Dict[str, Any]:
         """A dictionary which can be queried for additional account rights."""
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -1,19 +1,25 @@
-# Copyright (c) 2021 Ultimaker B.V.
+# Copyright (c) 2022 Ultimaker B.V.
 # Cura is released under the terms of the LGPLv3 or higher.
 import enum
 from datetime import datetime
+import json
 from PyQt6.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty, QTimer, pyqtEnum
-from typing import Any, Optional, Dict, TYPE_CHECKING, Callable
+from PyQt6.QtNetwork import QNetworkRequest
+from typing import Any, Callable, Dict, Optional, Set, TYPE_CHECKING
 
 from UM.Logger import Logger
 from UM.Message import Message
 from UM.i18n import i18nCatalog
+from UM.TaskManagement.HttpRequestManager import HttpRequestManager
+from UM.TaskManagement.HttpRequestScope import JsonDecoratorScope
 from cura.OAuth2.AuthorizationService import AuthorizationService
 from cura.OAuth2.Models import OAuth2Settings, UserProfile
 from cura.UltimakerCloud import UltimakerCloudConstants
+from cura.UltimakerCloud.UltimakerCloudScope import UltimakerCloudScope
 
 if TYPE_CHECKING:
     from cura.CuraApplication import CuraApplication
+    from PyQt6.QtNetwork import QNetworkReply
 
 i18n_catalog = i18nCatalog("cura")
 
@@ -78,6 +84,7 @@ class Account(QObject):
         self._logged_in = False
         self._user_profile: Optional[UserProfile] = None
         self._additional_rights: Dict[str, Any] = {}
+        self._permissions: Set[str] = set()  # Set of account permission keys, e.g. {"digital-factory.print-job.write"}
         self._sync_state = SyncState.IDLE
         self._manual_sync_enabled = False
         self._update_packages_enabled = False
@@ -109,6 +116,7 @@ class Account(QObject):
 
         self._sync_services: Dict[str, int] = {}
         """contains entries "service_name" : SyncState"""
+        self.syncRequested.connect(self._updatePermissions)
 
     def initialize(self) -> None:
         self._authorization_service.initialize(self._application.getPreferences())
@@ -321,3 +329,40 @@ class Account(QObject):
     def additionalRights(self) -> Dict[str, Any]:
         """A dictionary which can be queried for additional account rights."""
         return self._additional_rights
+
+    def _updatePermissions(self) -> None:
+        """
+        Update the set of permissions that the user has.
+        """
+        def callback(reply: "QNetworkReply"):
+            status_code = reply.attribute(QNetworkRequest.Attribute.HttpStatusCodeAttribute)
+            if status_code is None:
+                Logger.error("Server did not respond to request to get list of permissions.")
+                return
+            if status_code >= 300:
+                Logger.error(f"Request to get list of permission resulted in HTTP error {status_code}")
+                return
+
+            try:
+                reply_data = json.loads(bytes(reply.readAll()).decode("UTF-8"))
+            except (UnicodeDecodeError, json.JSONDecodeError, ValueError) as e:
+                Logger.logException("e", f"Could not parse response to permission list request: {e}")
+                return
+            if "errors" in reply_data:
+                Logger.error(f"Request to get list of permission resulted in error response: {reply_data['errors']}")
+                return
+
+            if "data" in reply_data and "permissions" in reply_data["data"]:
+                self._permissions = set(reply_data["data"]["permissions"])
+
+        def error_callback(reply: "QNetworkReply", error: "QNetworkReply.NetworkError"):
+            Logger.error(f"Request for user permissions list failed. Network error: {error}")
+
+        HttpRequestManager.getInstance().get(
+            url = f"{self._oauth_root}/users/permissions",
+            scope = JsonDecoratorScope(UltimakerCloudScope(self._application)),
+            callback = callback,
+            error_callback = error_callback,
+            timeout = 10
+        )
+
--- a/cura/API/Account.py
+++ b/cura/API/Account.py
@@ -5,7 +5,7 @@ from datetime import datetime
 import json
 from PyQt6.QtCore import QObject, pyqtSignal, pyqtSlot, pyqtProperty, QTimer, pyqtEnum
 from PyQt6.QtNetwork import QNetworkRequest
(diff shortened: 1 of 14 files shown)


plugins/DigitalLibrary/src/DigitalFactoryApiClient.py
```diff
@@ -71,8 +71,6 @@ def callbackWrap(response: Optional[Any] = None, *args, **kwargs) -> None:
                 has_access = response.library_max_private_projects == -1 or response.library_max_private_projects > 0
                 callback(has_access)
                 self._library_max_private_projects = response.library_max_private_projects
-                # update the account with the additional user rights
-                self._account.updateAdditionalRight(df_access = has_access)
```
jellespijker, line 75: See comment on the removal of this property in Account.py

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
```diff
@@ -209,8 +209,13 @@ Item
             onClicked: enabled ? contextMenu.switchPopupState() : {}
             visible:
             {
-                if (!printer || !printer.activePrintJob) {
-                    return false
+                if(!printer || !printer.activePrintJob)
```
jellespijker, line 212:
```suggestion
                if (!printer || !printer.activePrintJob)
```

plugins/UM3NetworkPrinting/resources/qml/MonitorPrinterCard.qml
```diff
@@ -209,8 +209,13 @@ Item
...
+                {
+                    return false;
+                }
+                if(!contextMenu.hasItems)
```
jellespijker, line 216:
```suggestion
                if (!contextMenu.hasItems)
```

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
```diff
@@ -206,7 +206,11 @@ Item
         onClicked: enabled ? contextMenu.switchPopupState() : {}
         visible:
         {
-            if (!printJob)
+            if(!printJob)
```
jellespijker, line 209:
```suggestion
            if (!printJob)
```

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
```diff
@@ -206,7 +206,11 @@ Item
...
+            {
+                return false;
+            }
+            if(!contextMenu.hasItems)
```
jellespijker, line 213:
```suggestion
            if (!contextMenu.hasItems)
```

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
```diff
@@ -340,6 +342,37 @@ def openPrintJobControlPanel(self) -> None:
...
         QDesktopServices.openUrl(QUrl(self.clusterCloudUrl + "?utm_source=cura&utm_medium=software&utm_campaign=monitor-manage-printer"))
 
+    permissionsChanged = pyqtSignal()
+
+    @pyqtProperty(bool, notify = permissionsChanged)
+    def canReadPrintJobs(self) -> bool:
+        """
+        Whether this user can read the list of print jobs and their properties.
+        """
+        return "digital-factory.print-job.read" in self._account.permissions
```
jellespijker, line 352: What is the difference between `digital-factory.print-job.read` and `cura.printjob.read` see https://docs.api.ultimaker.com/authentication/scopes.html does one superseeds the other? Are we still using the `cura.printjob.read` scope with this new permission scheme

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Ultimaker B.V.
+# Copyright (c) 2022 Ultimaker B.V.
```
jellespijker, line 1: Nothing changed in this file, why update the Copyright?

resources/qml/ApplicationSwitcher/ApplicationSwitcherPopup.qml
```diff
@@ -99,7 +99,24 @@ Popup
...
                 isExternalLink: true
-                visible: modelData.DFAccessRequired ? Cura.API.account.isLoggedIn & Cura.API.account.additionalRights["df_access"] : true
+                visible:
+                {
+                    try
+                    {
+                        modelData.permissionsRequired.forEach(function(permission)
+                        {
+                            if(!Cura.API.account.isLoggedIn || !Cura.API.account.permissions.includes(permission)) //This required permission is not in the account.
+                            {
```
jellespijker, line 109:
```suggestion
                            if (!Cura.API.account.isLoggedIn || !Cura.API.account.permissions.includes(permission)) // This required permission is not in the account.
```

resources/qml/ApplicationSwitcher/ApplicationSwitcherPopup.qml
```diff
@@ -99,7 +99,24 @@ Popup
...
+                                throw "No permission to use this application."; //Can't return from within this lambda. Throw instead.
+                            }
```
jellespijker, line 111:
```suggestion
                                throw "No permission to use this application."; // Can't return from within this lambda. Throw instead.
```

resources/qml/ApplicationSwitcher/ApplicationSwitcherPopup.qml
jellespijker, line 109:
```suggestion
                            if(!Cura.API.account.isLoggedIn || !Cura.API.account.permissions.includes(permission)) // This required permission is not in the account.
```

resources/qml/Preferences/Materials/MaterialsSyncDialog.qml
```diff
@@ -112,7 +120,15 @@ UM.Window
...
                     {
-                        swipeView.currentIndex += 1;
+                        if(Cura.API.account.permissions.includes("digital-factory.printer.write"))
+                        {
+                            swipeView.currentIndex += 1;
+                        }
+                        else
+                        {
+                            //Logged in, but no permissions to start syncing. Direct them to USB.
+                            swipeView.currentIndex = removableDriveSyncPage.SwipeView.index;
```
jellespijker, line 130:
```suggestion
                            // Logged in, but no permissions to start syncing. Direct them to USB.
```

jellespijker (changes requested):
Some style comments. All of them suggestion and provided as a suggestion. Such that they can be easily applied if you agree with them.

One fundamental question; _How does the new permissions interact with the CLIENT_SCOPE?`

It also looks like you replaced the `AdditionalRights` property with the `permissions` property, but that previous property isn't used anymore. Best to remove that in my opinion.

Ghostkeeper:
> One fundamental question; _How does the new permissions interact with the CLIENT_SCOPE?`

They are two different systems that sort of work on top of each other. `CLIENT_SCOPE` stores the OAuth scope, which is a list of the types of access that Cura requests from the account. The permissions are part of the user profile, and store what permissions that user profile has. In my opinion they could be merged together, to let the OAuth scope also determine what permissions the user has. But that's not currently implemented in the cloud. They are two different systems there.
It should be mentioned that the permissions system is part of the "account" API in the cloud, but the available permission keys are currently all about the Digital Factory.

> It also looks like you replaced the AdditionalRights property with the permissions property, but that previous property isn't used anymore. Best to remove that in my opinion.

I couldn't remove it because that would be an API break. I did remove the `df_access` key in there because that was added by a plug-in anyway so no other system should rely on that key being added (the plug-in can be disabled after all). As discussed in person, I'll add a deprecation annotation to this property.

I'll now go through the code style review.

cura/API/Account.py
```diff
@@ -321,3 +329,51 @@ def updateAdditionalRight(self, **kwargs) -> None:
     def additionalRights(self) -> Dict[str, Any]:
         """A dictionary which can be queried for additional account rights."""
         return self._additional_rights
+
+    permissionsChanged = pyqtSignal()
+    @pyqtProperty("QVariantList", notify = permissionsChanged)
+    def permissions(self) -> List[str]:
```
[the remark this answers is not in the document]
Ghostkeeper: As we discussed, this can't be removed without an API break. We've deprecated the function instead.

cura/PrinterOutput/Models/PrintJobOutputModel.py
```diff
@@ -86,6 +88,13 @@ def updateOwner(self, owner: str) -> None:
             self._owner = owner
             self.ownerChanged.emit()
 
+    @pyqtProperty(bool, notify = ownerChanged)
+    def isMine(self) -> bool:
+        """
+        Returns whether this print job was sent by the currently logged in user.
```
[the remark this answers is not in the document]
Ghostkeeper: Done in 125ac98318d74b7d6f6404fe1fd2a74e6aa51906.

plugins/UM3NetworkPrinting/resources/qml/MonitorContextMenu.qml
```diff
@@ -13,10 +13,94 @@ import Cura 1.6 as Cura
...
+            if(!OutputDevice.canWriteOthersPrintJobs)
+            {
+                return false;
+            }
+        }
+        var states = ["printing", "pausing", "paused", "resuming"];
+        return states.indexOf(printJob.state) !== -1;
+    }
+    property bool abortVisible:
+    {
```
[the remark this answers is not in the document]
Ghostkeeper: I'll add the newline before the property, like the rest of the suggested changes.

plugins/UM3NetworkPrinting/resources/qml/MonitorPrintJobCard.qml
Ghostkeeper: This is not in the code style guidelines and I don't really agree with the change.

plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py
Ghostkeeper: The OAuth scope system and the account permissions system are two different systems that work on top of each other in the cloud. We need both. We can assume that the OAuth scope can only be given if the account has that permission, and we always ask for this OAuth scope upon logging in, so it's sufficient to read if the permission is in the account.

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
Ghostkeeper: Probably a mistake because I originally changed something, then undid my changes but forgot to undo my changes to the copyright header.

I'll change it to 2020 which was the last actual change.

resources/qml/ApplicationSwitcher/ApplicationSwitcherPopup.qml
Ghostkeeper: This style is not in our code style guidelines (not specified) and I don't really agree with it, so I'll leave it as is.

PR #12867, opened 2022-07-26
ArrangeObjectsJob.py: Remove if conditional block

digitalfrost:
See #12863
This solution completly removes the problematic conditional block
The flow of the program is now easier to understand as there is context
for the no_full_solution_message.


digitalfrost:
See https://github.com/Ultimaker/Cura/pull/13060 for simpler solution

PR #12878, opened 2022-07-27
CURA-9520_Fix_support_infill_tree_support

pkuiper-ultimaker:
Fixed issue with tree support having an infill density set to 15%. Now it is set to 0%.

Relates to CURA-9520


jellespijker:
Looks good, will merge this now

PR #12881, opened 2022-07-27
Cura 9031 add slice id to griffin gcode header

jellespijker:
Add an UUID to the slice info data. This UUID is generated by CuraEngine and send via Arcus to the front-end

See also the PR on the CuraEngine side

### DIFF
--- a/cura/UI/PrintInformation.py
+++ b/cura/UI/PrintInformation.py
@@ -38,6 +38,8 @@ class PrintInformation(QObject):
 
         self.initializeCuraMessagePrintTimeProperties()
 
+        self._slice_uuid: Optional[str] = None
+
         # Indexed by build plate number
         self._material_lengths = {}  # type: Dict[int, List[float]]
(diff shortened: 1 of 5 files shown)


Ghostkeeper (approved):

cura/UI/PrintInformation.py
```diff
@@ -132,6 +134,14 @@ def setPreSliced(self, pre_sliced: bool) -> None:
             self._updateJobName()
             self.preSlicedChanged.emit()
 
+    @property
+    def slice_uuid(self) -> Optional[str]:
+        return self._slice_uuid
+
+    @slice_uuid.setter
+    def slice_uuid(self, value: Optional[str]) -> None:
+        self._slice_uuid = value
```
Ghostkeeper, line 143: Could consider making this a public field instead. Right now this is just overhead. That's what public fields are for, after all.

jellespijker: Agreed I will change that in a minute

PR #12909, opened 2022-07-31
Use dataclasses for Peripheral

digitalfrost:
Since Peripheral is a data class use dataclasses


digitalfrost:
@jellespijker - do you need anything from me to move this forward?

The checks fail with
```
Error: A branch or tag with the name '310722' could not be found
```

jellespijker:
At this time not, our CI/CT gives a failure due to something else.

I might use this PR to test and debug this somewhere this week if you don't mind.

At the moment we have less people then normal so it might take a while before we actually merge stuff.

Do know that we really appreciate your contributions and I hope to free up some time this week to really look at it.

jellespijker:
Agreed this is indeed a data class and should be treated as such

PR #12929, opened 2022-08-02
MachineErrorChecker.py: Move _error_check_timer initialization to a separate function

digitalfrost:
This simplifies understanding the class `__init__` function


jellespijker:
This is good practice indeed and something we (ourselves) should do more often. 

Thank you for this

PR #12930, opened 2022-08-02
Add new printer

Chrystal-1024:
we need add some new types of machines.please help for us.


Chrystal-1024:
@nallath I have reduced the size of platform meshes.

PR #12935, opened 2022-08-03
Fix Bug in timing MachineErrorChecker execution

digitalfrost:
We want the Logger in line 215 to give the execution time in seconds
so we don't want the start time to be the epoch.
Currently the logger will output something like
time = 1659518458.5s
not like
time = 3.0s

### DIFF
--- a/cura/Machines/MachineErrorChecker.py
+++ b/cura/Machines/MachineErrorChecker.py
@@ -43,7 +43,7 @@ class MachineErrorChecker(QObject):
         self._application = cura.CuraApplication.CuraApplication.getInstance()
         self._machine_manager = self._application.getMachineManager()
 
-        self._start_time = time.time()  # measure checking time
+        self._check_start_time = time.time()
 
         # This timer delays the starting of error check so we can react less frequently if the user is frequently
         # changing settings.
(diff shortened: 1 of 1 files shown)


cura/Machines/MachineErrorChecker.py
```diff
@@ -212,4 +212,4 @@ def _setResult(self, result: bool, keys_to_recheck = None) -> None:
         self._check_in_progress = False
         self.needToWaitForResultChanged.emit()
         self.errorCheckFinished.emit()
-        Logger.log("i", "Error check finished, result = %s, time = %0.1fs", result, time.time() - self._start_time)
+        Logger.log("i", "Error check finished, result = %s, time = %0.1fs", result, time.time() - self._check_start_time)
```
jellespijker, line 215: Use f-string and `info` method
```suggestion
        execution_time = time.time() - self._check_start_time
        Logger.info(f"Error check finished, result = {result}s, time = {execution_time:.2f}s")
```

jellespijker (changes requested):
Yes. Good catch!

Added a quick suggestion to use f-string and the info function

nallath (approved):

PR #12959, opened 2022-08-05
LayerPolygon.py: Refactor check for unknown line types

digitalfrost:
First commit improved the iteration using enumerate.
Second commit uses np.where() instead to improve clarity of intent.

### DIFF
--- a/cura/LayerPolygon.py
+++ b/cura/LayerPolygon.py
@@ -43,7 +43,7 @@ class LayerPolygon:
         if unknown_types:
             # Got faulty line data from the engine.
             for idx in unknown_types:
-                Logger.log("w", "Found an unknown line type at: %s", idx)
+                Logger.warn(f"Found an unknown line type at: {idx}")
                 self._types[idx] = self.NoneType
         self._data = data
         self._line_widths = line_widths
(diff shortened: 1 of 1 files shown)


cura/LayerPolygon.py
```diff
@@ -39,10 +39,12 @@ def __init__(self, extruder: int, line_types: numpy.ndarray, data: numpy.ndarray
...
         self._types = line_types
-        for i in range(len(self._types)):
-            if self._types[i] >= self.__number_of_types: # Got faulty line data from the engine.
-                Logger.log("w", "Found an unknown line type: %s", i)
-                self._types[i] = self.NoneType
+        unknown_types = np.where(self_types >= self_number_of_types)
+        if unknown_types:
+            # Got faulty line data from the engine.
+            for idx in unknown_types:
+                Logger.log("w", "Found an unknown line type at: %s", idx)
```
jellespijker, line 46: Can you use the f-string and the `warn` method here?
```suggestion
                Logger.warn(f"Found an unknown line type at: {idx}")
```

jellespijker (changes requested):
Thank you for these kind of refactors.

I think the use of `np.where` is a smart choice since it faster and shows the intent more clearly

nallath (approved):

PR #12970, opened 2022-08-07
Use is not operator rather than not ... is

digitalfrost:
This is advocated by both PEP8 and the Google Python Style Guide


fieldOfView:
Are there perhaps more instances of this that could/should be fixed?

digitalfrost:
Just looking at [plugins/X3DReader/X3DReader.py](https://github.com/Ultimaker/Cura/pull/12970/files#diff-b54ee0d95e46657e21b099d22d56a555cce5a2479e7193f5e55e96ad5be41015) in this pull request.
Atomic commits for easy merging.

jellespijker (approved):
Thanks you for the contribution.

This is a no brainer... merge!

PR #12980, opened 2022-08-08
ShapeArray.py: Use is not operator rather than not ... is

digitalfrost:
Change to use is not operator rather than not ... is 
in **cura/Arranging/ShapeArray.py** 
as advocated by PEP8 and Google Python Style Guide.

See comment by @fieldOfView in #12970

Atomic commits for easy merging.

### DIFF
--- a/cura/Arranging/ShapeArray.py
+++ b/cura/Arranging/ShapeArray.py
@@ -74,7 +74,7 @@ class ShapeArray:
         # If the child-nodes are included, adjust convex hulls as well:
         if include_children:
             children = node.getAllChildren()
-            if not children is None:
+            if children is not None:
                 for child in children:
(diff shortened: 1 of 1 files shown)


fieldOfView:
Can you add these changes to #12970? The Cura team is swamped with work as is, and having so many PRs will not make their workload lighter. Also, this could have been one commit, right?

digitalfrost:
Merged with #12970

@fieldOfView
The idea of all these commits is to actually make it easier for you to merge the pull request.
Each commit is atomic.  So you can either quickly cycle through the diff of each commit or see all the changes at once by looking at _Files Changed_.  Then you can easily rebase squash into one commit if you don’t like the granularity in the main history.

PR #12981, opened 2022-08-08
X3DReader.py: Iterate directly over list in readIndex()

digitalfrost:
No need for range(len.. and later use of index...

### DIFF
--- a/plugins/X3DReader/X3DReader.py
+++ b/plugins/X3DReader/X3DReader.py
@@ -866,13 +866,13 @@ def readIndex(node, attr):
     v = readIntArray(node, attr, [])
     chunks = []
     chunk = []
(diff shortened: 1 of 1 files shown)


plugins/X3DReader/X3DReader.py
```diff
@@ -866,13 +866,13 @@ def readIndex(node, attr):
     v = readIntArray(node, attr, [])
     chunks = []
     chunk = []
-    for i in range(len(v)):
-        if v[i] == -1:
+    for i in v:
```
digitalfrost, line 869: **i** is not a great variable name but this is consistent with the math style used in the file...

nallath (approved):

PR #12988, opened 2022-08-09
CuraContainerStack.py: iterate directly over IndexTypeMap

digitalfrost:
There is no need for range(len


digitalfrost:
[conan-recipe-version / get-semver](https://github.com/Ultimaker/Cura/runs/7742885336?check_suite_focus=true#logs) fails on:
```Error: A branch or tag with the name '090822' could not be found```

nallath (approved):

PR #12995, opened 2022-08-10
PostProcessingPlugin/.../Stretch.py Use enumerate

digitalfrost:
Use enumerate instead of range(len


digitalfrost:
Checks fail because they cannot find the pull request branch:
``` Error: A branch or tag with the name '100822' could not be found```

jellespijker:
Checks failing is indeed a known issue for external forks. We will look into this in due time

jellespijker:
Change is save enough to merge imo.

PR #13022, opened 2022-08-13
VersionUpgrade.py: Use enumerate

digitalfrost:
Use enumerate instead of range(len


jellespijker:
@casperlamboo since you're currently working on a version upgrader, can you take a look at this PR

nallath (approved):

PR #13056, opened 2022-08-17
Add new theme "Aqua"

x40-Community:
Hello,
I created a new theme with a fresh look. The theme is very popular with users, so I was asked to integrate the theme into the source code.!
Cura_Aqua_Theme


fieldOfView:
Themes support inheritance. If your theme inherits the "cura-light" theme, you don't have to include the fonts or images. You can also simplify the theme.json file by only including the items that are different from cura-light (ie: probably colors only).

fieldOfView:
I see the `"sizes"` section is still in the `theme.json` file. Are all of these sizes different from the `cura-light` sizes? If you just want to adjust a couple of sizes, you only need to include those few, not the whole section. The same is true for the colors by the way.

fieldOfView:
For reference, see this previous PR from long ago and the specific reasoning not to merge it "for now": 
https://github.com/Ultimaker/Cura/pull/2063#issuecomment-315040424

lorenzoromagnoli:
@x40-Community thanks for your theme proposal! 
I discussed the topic with the team and we are still of the idea of not adding more themes to Cura unless needed to support stricter usability requirements. 

Each theme requires some degree of maintenance and testing and since merging your theme might open up the way to other PR we think it is safer to stick to the basic themes for now.

We would love to extend the plugins marketplace with themes to allow the community to customize and share their favourite themes but this is unfortunately not in our roadmap. 

Thanks again for creating the theme and for sharing it with people, making their Cura experience more customized and enjoyable. ❤️

PR #13060, opened 2022-08-18
ArrangeObjectsJob.py: Refactor run method

digitalfrost:
arrange from Nest2DArrange will return a boolean so we can simplify

### DIFF
--- a/cura/Arranging/ArrangeObjectsJob.py
+++ b/cura/Arranging/ArrangeObjectsJob.py
@@ -28,18 +28,20 @@ class ArrangeObjectsJob(Job):
                                  title = i18n_catalog.i18nc("@info:title", "Finding Location"))
         status_message.show()
 
-        found_solution_for_all = None
         try:
             found_solution_for_all = arrange(self._nodes, Application.getInstance().getBuildVolume(), self._fixed_nodes)
         except:  # If the thread crashes, the message should still close
+            found_solution_for_all = False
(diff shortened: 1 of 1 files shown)


cura/Arranging/ArrangeObjectsJob.py
```diff
@@ -28,18 +28,20 @@ def run(self):
                                  title = i18n_catalog.i18nc("@info:title", "Finding Location"))
         status_message.show()
 
-        found_solution_for_all = None
```
jellespijker, line 31: Maybe initialize the variable with False and don't set it a the exception block
```suggestion
        found_solution_for_all = False
```

cura/Arranging/ArrangeObjectsJob.py
```diff
@@ -28,18 +28,20 @@ def run(self):
...
         try:
             found_solution_for_all = arrange(self._nodes, Application.getInstance().getBuildVolume(), self._fixed_nodes)
         except:  # If the thread crashes, the message should still close
+            found_solution_for_all = False
```
jellespijker, line 34:
```suggestion
```

See comment above

jellespijker (changes requested):
Some minor feedback. If you initialize the variable to `False` you ensure that it always be defined.

Yes, your current flow would work, but it is good practice to always initialize a variable such that it is available within the scope. This reduces the risk of an introduction of a bug at a later time.

jellespijker:
Thank you for helping out and suggesting these kinds of simplifications

PR #13082, opened 2022-08-20
OnExitCallbackManager.py: improve documentation

digitalfrost:
- Fix typos
- Improve grammar
- Improve clarity
- Improve layout and shorten line lengths


jellespijker:
Thank you for improving this documentation. For future reference. We value vertical space more than horizontal space. Shortening lines is usually not necessary and we use the guideline of 180 up until 240 characters.

PR #13102, opened 2022-08-24
[CURA-9514] Collapsable printers list

Joeydelarago:
Add new collapsable printers list that groups cloud/LAN printers by printer type.


cura/Machines/Models/MachineListModel.py
```diff
@@ -0,0 +1,96 @@
...
+        for stack in other_machine_stacks:
+            self.addItem(stack)
+
+    def addItem(self, container_stack: ContainerStack, machine_count: int = 0) -> None:
+        if parseBool(container_stack.getMetaDataEntry("hidden", False)):
+            return
+
+        is_online = parseBool(container_stack.getMetaDataEntry("is_online", False))
+        if container_stack.getMetaDataEntry("type") == "abstract_machine":
+            is_online = True
```
casperlamboo, line 88: Why don't we add "is_online" as a property to the meta data of abstract machine? currently we only add cloud printers to the abstract machines so that would always be true.

It is a bit annoying if we manually have to add the abstract machines to online printers in all instances we implement something similar

cura/Settings/AbstractMachine.py
```diff
@@ -14,14 +15,30 @@ def __init__(self, container_id: str) -> None:
...
         application = CuraApplication.getInstance()
         registry = application.getContainerRegistry()
 
-        printer_type = self.definition.getId()
-        return [machine for machine in registry.findContainerStacks(type="machine") if machine.definition.id == printer_type and ConnectionType.CloudConnection in machine.configuredConnectionTypes]
+        machines = registry.findContainerStacks(type="machine")
+        # Filter machines that match definition
+        machines = filter(lambda machine: machine.definition.id == abstract_machine.definition.getId(), machines)
+        # Filter only LAN and Cloud printers
+        machines = filter(lambda machine: ConnectionType.CloudConnection in machine.configuredConnectionTypes or ConnectionType.NetworkConnection in machine.configuredConnectionTypes, machines)
```
casperlamboo, line 36: combine lamda's

```suggestion
        def machinesFilter(machine):
            if machine.definition.id is not abstract_machine.definition.getId():
                return false
            return ConnectionType.CloudConnection in machine.configuredConnectionTypes or ConnectionType.NetworkConnection in machine.configuredConnectionTypes
        machines = filter(machinesFilter, machines)
```
note: not tested

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDeviceManager.py
```diff
@@ -232,6 +232,9 @@ def _createMachineFromDiscoveredDevice(self, device_id: str) -> None:
         self._connectToOutputDevice(device, new_machine)
         self._showCloudFlowMessage(device)
 
+        _abstract_machine = CuraStackBuilder.createAbstractMachine(device.printerType)
```
casperlamboo, line 235: should this have been implemented in the previous ticket? anyway nice catch!

resources/qml/PrinterSelector/MachineListButton.qml
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2018 Ultimaker B.V.
```
casperlamboo, line 1:
```suggestion
// Copyright (c) 2022 Ultimaker B.V.
```

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDeviceManager.py
Joeydelarago: This is for the LAN printers which wasn't in the requirements for the last ticket (by accident I assume).

cura/Settings/AbstractMachine.py
Joeydelarago: I intentionally have them separated. I think it makes it easier to understand the filtering steps.

PR #13106, opened 2022-08-24
CuraSceneController.py : Fix problem with log message and use f-string

digitalfrost:
- Correct log message (we want to log the "select**ed**" build plate)
- Use f-string and debug method for Logger instead of "Old Style" string formatting.


jellespijker:
Thanks, will merge this

PR #13111, opened 2022-08-25
Modify the Ender 6 Profile.

GroovyDrifter:
Remove material temperatures from profile so they can follow those in the material seleced.


GroovyDrifter:
New to all this GIT stuff, hope I did not mess anything

GroovyDrifter:
I changed even more the code to the Ender 6 profile so it is like any other Creality printer, removing any additional settings. In my tests it prints better with this simpler profile if you have original firmware.

nallath:
I don't think that the changes that you made are quite correct. You also seem to have changed the name of the printer to be creality3 and change the size of the printer.

GroovyDrifter:
You're right, I messed things copyng and pasting for the wrong notepad++ tab, sorry... now i think it is fixed.

nallath (approved):

PR #13128, opened 2022-08-26
Remove AbstractMachine

casperlamboo:
Having a separate class for the AbstractMachine complicated things; it's behaviour was extremely similar to the GlobalStack so adding one more stack container type in addition to the many similar setting container types we already have adds complexity to the system. Having these different classes for machines and abstract machines also add complexity to the update script as the abstract machines were stored in a separate folder from the machine types.

Because of these reasons we decided to replace the AbstractMachine by a GlobalStack where the is_abstract_machine property metadata property is set to True.

CURA-9514, CURA-9277

Co-authored-by: joeydelarago <joeydelarago@gmail.com>


cura/Settings/GlobalStack.py
```diff
@@ -344,13 +343,36 @@ def getDefaultFirmwareName(self) -> str:
...
 
-    def setName(self, name: "str") -> None:
+    def setName(self, name: str) -> None:
         super().setName(name)
 
     nameChanged = pyqtSignal()
     name = pyqtProperty(str, fget=getName, fset=setName, notify=nameChanged)
 
 
+def getMachinesWithDefinition(definition_id: str, online_only = False) -> List[ContainerStack]:
```
nallath, line 353: Doesn't it make more sense to put this in the machine manager?

nallath (changes requested):

cura/Settings/GlobalStack.py
casperlamboo: f000b75661765b3087a7e1fb53c4bd37c3d081e8

PR #13131, opened 2022-08-26
[CURA-9422] 'Monitor page' for abstract printer types

rburema:
If an abstract, but cloud connected printer is chosen to monitor, send the user to digital factory instead. (Since we don't know which printer they want, since all we have is a type.)

### DIFF
--- a/plugins/MonitorStage/MonitorMain.qml
+++ b/plugins/MonitorStage/MonitorMain.qml
@@ -126,7 +126,7 @@ Rectangle
             {
                 anchors.horizontalCenter: parent.horizontalCenter
                 anchors.verticalCenter: parent.verticalCenter
-                spacing: UM.Theme.getSize("default_margin").height
+                spacing: UM.Theme.getSize("wide_margin").height
                 padding: UM.Theme.getSize("default_margin").width
                 topPadding: 0
 
--- a/plugins/MonitorStage/MonitorMain.qml
+++ b/plugins/MonitorStage/MonitorMain.qml
@@ -120,16 +120,19 @@ Rectangle
         {
             id: sendToFactoryCard
             color: UM.Theme.getColor("detail_background")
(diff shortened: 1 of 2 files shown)


Joeydelarago (approved):

plugins/MonitorStage/MonitorMain.qml
```diff
@@ -12,7 +12,7 @@ Rectangle
     id: viewportOverlay
 
     property bool isConnected: Cura.MachineManager.activeMachineHasNetworkConnection || Cura.MachineManager.activeMachineHasCloudConnection
-    property bool isAbstractCloudPrinter: Cura.MachineManager.activeMachineIsAbstract // && Cura.MachineManager.activeMachineHasCloudRegistration
+    property bool isAbstractCloudPrinter: Cura.MachineManager.activeMachineIsAbstract
```
rburema, line 15: Oops, I don't think that should have been commented out in the first place. I think it now also may send 3rd party abstract printer users to our Digital Factory? 😅

nallath: Yeah, that's also why this shouldn't live here. The output device that is active for abstract machines should handle this. I'm currently implementing that as part of CURA-8463

plugins/MonitorStage/MonitorMain.qml
```diff
@@ -12,6 +12,7 @@ Rectangle
     id: viewportOverlay
 
     property bool isConnected: Cura.MachineManager.activeMachineHasNetworkConnection || Cura.MachineManager.activeMachineHasCloudConnection
+    property bool isAbstractCloudPrinter: Cura.MachineManager.activeMachineIsAbstract // && Cura.MachineManager.activeMachineHasCloudRegistration
```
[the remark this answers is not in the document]
rburema: Ha, yes it was :-p

PR #13168, opened 2022-08-30
Add button to hide/show connected printers

casperlamboo:
CURA-9514

### DIFF
--- a/cura/Machines/Models/MachineListModel.py
+++ b/cura/Machines/Models/MachineListModel.py
@@ -60,7 +60,7 @@ class MachineListModel(ListModel):
     def showCloudPrinters(self) -> bool:
         return self._show_cloud_printers
 
-    @pyqtSlot(bool, name="setShowCloudPrinters")
+    @pyqtSlot(bool) 
     def setShowCloudPrinters(self, show_cloud_printers: bool) -> None:
         self._show_cloud_printers = show_cloud_printers
         self._updateDelayed()
--- a/cura/Machines/Models/MachineListModel.py
+++ b/cura/Machines/Models/MachineListModel.py
@@ -61,7 +61,7 @@ class MachineListModel(ListModel):
         return self._show_cloud_printers
 
     @pyqtSlot(bool, name="setShowCloudPrinters")
-    def set_show_cloud_printers(self, show_cloud_printers: bool) -> None:
+    def setShowCloudPrinters(self, show_cloud_printers: bool) -> None:
         self._show_cloud_printers = show_cloud_printers
         self._updateDelayed()
         self.showCloudPrintersChanged.emit(show_cloud_printers)
--- a/cura/Machines/Models/MachineListModel.py
+++ b/cura/Machines/Models/MachineListModel.py
@@ -103,21 +103,20 @@ class MachineListModel(ListModel):
 
         if len(abstract_machine_stacks) > 0:
             if self._show_cloud_printers:
-                self.appendItem({ "listType": "HIDE_BUTTON",
-                                  "isOnline": True,
-                                  "isAbstractMachine": False,
-                                  "machineCount": 0
+                self.appendItem({"listType": "HIDE_BUTTON",
+                                 "isOnline": True,
+                                 "isAbstractMachine": False,
+                                 "machineCount": 0
                                  })
             else:
                 self.appendItem({"listType": "SHOW_BUTTON",
-                                  "isOnline": True,
-                                  "isAbstractMachine": False,
-                                  "machineCount": 0
+                                 "isOnline": True,
+                                 "isAbstractMachine": False,
+                                 "machineCount": 0
                                  })
 
         for stack in other_machine_stacks:
             self.addItem(stack)
-        print(self.items)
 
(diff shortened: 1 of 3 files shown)


resources/qml/PrinterSelector/MachineListButton.qml
```diff
@@ -7,81 +7,134 @@ import QtQuick.Controls 2.3
...
+            case "SHOW_BUTTON":
+                showButtonComponent
+                break;
+            case "MACHINE":
+                machineListButtonComponent
+                break;
+            default:
+        }
+    }
+    property var onClicked
```
rburema, line 28: Where is this property actually set? Doesn't the `OnClicked` in the `MachineListButton` (in `MachineSelectorList`) already take care of the clicks?

cura/Machines/Models/MachineListModel.py
```diff
@@ -98,7 +125,9 @@ def addItem(self, container_stack: ContainerStack, machine_count: int = 0) -> No
         for connection_type in [ConnectionType.NetworkConnection.value, ConnectionType.CloudConnection.value]:
             has_connection |= connection_type in container_stack.configuredConnectionTypes
 
-        self.appendItem({"name": container_stack.getName(),
+        self.appendItem({
+                        "listType": "MACHINE",
```
rburema, line 129: Indentation here (`listType`, `name`) is one space less than the ones that already existed (`id`, etc.).

cura/Machines/Models/MachineListModel.py
```diff
@@ -25,10 +25,13 @@ class MachineListModel(ListModel):
     IsOnlineRole = Qt.ItemDataRole.UserRole + 5
     MachineCountRole = Qt.ItemDataRole.UserRole + 6
     IsAbstractMachine = Qt.ItemDataRole.UserRole + 7
+    ListTypeRole = Qt.ItemDataRole.UserRole + 8
```
rburema, line 28: Not sure if this is the right name for it. All the other ones can be applied to the machine, not whether it's a (non-printer) button or not. (Speaking of which, I now just noticed that IsAbstractMachine doesn't have 'Role' appended to it's name.)

I'm OK with keeping it this way if we can't come up with a better name.

cura/Machines/Models/MachineListModel.py
```diff
@@ -50,6 +54,18 @@ def __init__(self, parent=None) -> None:
...
         self._updateDelayed()
 
+    showCloudPrintersChanged = pyqtSignal(bool)
+
+    @pyqtProperty(bool, notify=showCloudPrintersChanged)
+    def showCloudPrinters(self) -> bool:
+        return self._show_cloud_printers
+
+    @pyqtSlot(bool, name="setShowCloudPrinters")
+    def set_show_cloud_printers(self, show_cloud_printers: bool) -> None:
```
rburema, line 64: This is a function, and it's used as a function, so should be camelCased. (Although it looks deliberate; where there any problems with the 'normal' way of doing things?)

resources/qml/PrinterSelector/MachineListButton.qml
```diff
@@ -7,81 +7,134 @@ import QtQuick.Controls 2.3
...
 
-    contentItem: Item
+    Component
     {
-        width: machineListButton.width - machineListButton.leftPadding - machineListButton.rightPadding
-        height: UM.Theme.getSize("action_button").height
-
-        UM.ColorImage
+        id: hideButtonComponent
```
rburema, line 32: I wonder if it would be easier to just have a single show/hide (toggle) button component.

rburema (approved):
Looks good overall. Some low-prio remarks that I'd either like to be addressed (where addressed doesn't necessarily mean fixing).

I can component test most of this while you do that I think.

resources/qml/PrinterSelector/MachineListButton.qml
Joeydelarago: I see what you mean. I think having two components with different text is easier to understand than having one component with swapping text.

resources/qml/PrinterSelector/MachineListButton.qml
Joeydelarago: It needs to be declared in MachineListButton first so that it can be referenced in MachineListButton. If you take it out when this component is loaded OnClicked will be pointing to nothing in MachineListButton before the assignment in MachineSelectorList (causing the sad red text :cry: ).

cura/Machines/Models/MachineListModel.py
Joeydelarago: I agree, the name is not perfect. I changed it to ComponentType, I think this makes it a bit clearer this variable is used to decide which component to use with the model. 

Not much better, but I was getting a bit of analysis paralysis trying to name this so I went with the best I could come up with.

cura/Machines/Models/MachineListModel.py
```diff
@@ -61,7 +61,7 @@ def showCloudPrinters(self) -> bool:
         return self._show_cloud_printers
 
     @pyqtSlot(bool, name="setShowCloudPrinters")
```
casperlamboo, line 63: as name is now the same as the function name the property is redundent 
```suggestion
    @pyqtSlot(bool) 
```

PR #13173, opened 2022-08-31
Autogenerated intent profiles

pkuiper-ultimaker:
Autogenerated intent profiles. Settings are also automatically ordered (= easier diffing in the future).

Relates to: PP-202


rijkvanmanen (changes requested):
Most changes dont effect the end result, so that is perfect. But:
* jerk_layer_0 should not be removed, this will be later part of PP-42
* xy_offset_layer_0 is now removed, but I would prefer consistently adding it and make final value with PP-42

nallath (approved):

PR #13178, opened 2022-08-31
[CURA-9289] Add abstract machines on startup

Joeydelarago:
Check on startup for new abstract machines. This will make sure that if we get into a state where an abstract machine is missing it will always be added back after a restart. 

This will also add abstract machines after an update.

CURA-9289


Joeydelarago:
We went for an alternative solution

PR #13181, opened 2022-09-01
VzBot

ckvsoft:
After the VzBot printer is booming I added it to the resources to make it available for the users. I hope it's ok.


nallath:
I've not looked in-depth to your suggestions, but I did already see that the mesh files are _way_ too big. We only allow a max of 1 mb per build plate. In most cases, you shouldn't need more than 100kb.

resources/definitions/vzbot_base.def.json
```diff
@@ -0,0 +1,165 @@
...
+
+        "has_machine_quality": true,
+        "has_materials": true,
+        "has_variants": true,
+        "variants_name": "Nozzle Size",
+        "preferred_quality_type": "standard",
+        "preferred_variant_name": "0.4mm Nozzle",
+        "preferred_material": "generic_abs",
+        "exclude_materials": [
+            "ultimaker_abs_black",
```
nallath, line 22: You don't need to exclude these materials. They are 2.85 mm materials, so they are automatically filtered out.

resources/definitions/vzbot_custom.def.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "VzBot Custom",
```
nallath, line 2: What is the point of having this machine in there? It doesn't make a lot of sense to me.

ckvsoft: There are modified versions that have a different bed size. So the Platform STL file would not fit. Just like the voron2_custom.

resources/definitions/vzbot_base.def.json
ckvsoft: OK, I did not know that. Unfortunately I couldn't find any exact documentation. I copied that from the voron2. can i remove Thanks

ckvsoft:
What is this error? Does that have anything to do with my stuff?
```
Upload package to 'cura' failed: Conan interactive mode disabled`
```
other
```
Error: .github/workflows/unit-test.yml (Line: 164, Col: 22):
Error: The template is not valid. .github/workflows/unit-test.yml (Line: 164, Col: 22): Error reading JToken from JsonReader. Path '', line 0, position 0.
```

nallath:
The tests failing is an issue on our side. There isn't anything you can do about that. Sorry for the confusion!

ckvsoft:
> The tests failing is an issue on our side. There isn't anything you can do about that. Sorry for the confusion!

Oh, Thanks for the information.

PR #13219, opened 2022-09-06
Add profile for Creality Ender 3 S1 Pro and Plus

izilzty:
I have tested the profile on an Ender 3 S1 Plus and I think it should be correct.


drbytes:
Thanks, this helped my configure my S1.

0xorial:
why not merge this?

drbytes:
> why not merge this?

unit test fails, .. which is a shame since the profile works.

izilzty:
I noticed that setting_version needed to be increased to 20, so I updated it.

izilzty:
The `speed_travel` for S1 Pro and S1 Plus in Creality Slicer is calculated by the expression `120.0 if speed_print < 60 else 250.0 if speed_print > 100 else speed_print * 2.5` and does not use a fixed value, so use the expression to replace it.

jellespijker:

Joeydelarago (approved):

Joeydelarago:
Looks good! 

Thanks for the contribution 👍

PR #13237, opened 2022-09-07
Machine Actions improvement

julianCast:
Screen Shot 2022-09-07 at 4 41 45 PM

[Added machine action buttons dynamic visibility](https://github.com/Ultimaker/Cura/commit/ed14e3bd44890ff15406f7f53c7199f3fa6018e6)
Based on this change, users now will be able to manage the visibility of their machine action plugin implementations. For example, a machine action only visible when loggedIn.

[Added machine action dialog optional](https://github.com/Ultimaker/Cura/commit/75840426d735ddc421b594e6fa2378ed03ad2364)
Based on this change, users will be able to run their machine action plugins code just by clicking the action button, so without having to open a modal.

### DIFF
--- a/cura/MachineAction.py
+++ b/cura/MachineAction.py
@@ -34,8 +34,10 @@ class MachineAction(QObject, PluginObject):
         self._view = None
         self._finished = False
         self._open_as_dialog = True
+        self._visible = True
 
     labelChanged = pyqtSignal()
+    visibilityChanged = pyqtSignal()
     onFinished = pyqtSignal()
 
     def getKey(self) -> str:
@@ -125,8 +127,8 @@ class MachineAction(QObject, PluginObject):
     def getDisplayItem(self) -> Optional["QObject"]:
         return self._createViewFromQML()
 
-    @pyqtSlot(result = bool)
-    def openAsDialog(self) -> bool:
+    @pyqtProperty(bool, constant=True)
+    def shouldOpenAsDialog(self) -> bool:
         """Whether this action will show a dialog.
 
          If not, the action will directly run the function inside execute().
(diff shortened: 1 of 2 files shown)


cura/MachineAction.py
```diff
@@ -114,3 +124,26 @@ def qmlPath(self) -> "QUrl":
     @pyqtSlot(result = QObject)
     def getDisplayItem(self) -> Optional["QObject"]:
         return self._createViewFromQML()
+
+    @pyqtSlot(result = bool)
+    def openAsDialog(self) -> bool:
```
nallath, line 129: I feel that this should be a pyqtProperty and it should probably also be called something like "shouldOpenAsDialog", as openAsDialog feels more like a command to open an action (eg; Open this action as a dialog).

cura/MachineAction.py
```diff
@@ -114,3 +124,26 @@ def qmlPath(self) -> "QUrl":
...
+
+         If not, the action will directly run the function inside execute().
+
+        :return: Defaults to true to be in line with the old behaviour.
+        """
+
+        return self._open_as_dialog
+
+    @pyqtSlot(result = bool)
+    def isVisible(self) -> bool:
```
nallath, line 140: Same as above, this is a property of the machine action, so feel that it should also be a property (and probably just a "visible" one). 

Since it's probably likely that you want this to be dynamic, I would also consider creating a setter that triggers the signal so that the interface is notified of the property changes.

nallath (changes requested):
Good improvement! I have a few minor comments. After that I think this can be fixed.

I don't want to put any time pressure on you, but if we can get this in before Friday, it should still be in the 5.2 release (as we will go into feature freeze by then).

If you don't have time for these changes, I can also make them if you're okay with them, but I also wanted to give you a chance at discussing them if you feel that I'm wrong :)

cura/MachineAction.py
julianCast: Hi! you are right, it should called like that due is not an action. Im on it.

julianCast:
Requested changes added, please let me know if you need something else.

nallath (approved):

julianCast:
Hi! Do you thing will it be available for v5.2 ? Sorry I dont know the process.

nallath:
No, sorry, we already created a 5.2 branch that we sent over to QA. So it will be in 5.3, sorry!

nallath:
Hmm, something is also wrong with it. When I test the changes by going to the preferences/machines page, I get a crash

`TypeError: unable to convert a Python 'NoneType' object to a C++ 'bool' instance`

nallath:
Ah! Found it, you forgot a return statement in the visible property! I've fixed that for you!

PR #13288, opened 2022-09-14
[CURA-9535] Some trimesh operations require networkx

rburema:
Seems networkx itself is low on dependencies itself. It does require some optional ones for e.g. it's linalg sub-package, but those are numpy and scipy, which we already have.


jellespijker (approved):
LGTM

Is trimesh als a dependency in Uranium, if so I would also like to see a PR there

jellespijker:
Nevermind I should read the Jira ticket and associated pr first before commenting

PR #13295, opened 2022-09-15
Pp 230 enable conical support

pkuiper-ultimaker:
Move conical support out of experimental and under the support category.
Enabled conical support by default for ultimaker profiles and increased min. width to 10mm.

Relates to PP-230 and CURA-6710

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -7528,7 +7528,7 @@
                     "unit": "mm",
                     "type": "float",
                     "minimum_value": "0",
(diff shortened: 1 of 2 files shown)


nallath (changes requested):
The settings should not be children. A parent setting indicates that it's a more generic version of a setting (for example: Infill_speed is a child setting of print speed).

pkuiper-ultimaker:
> The settings should not be children. A parent setting indicates that it's a more generic version of a setting (for example: Infill_speed is a child setting of print speed).

OK clear. I have undone this change. They are not children anymore.

PR #13448, opened 2022-10-03
[CURA-9688] Fix View Printers in Digital Factory

MariMakes:
Send the user to the welcome page instead of the printjobs page to get a more comprehensive overview coming from the monitor page.


casperlamboo (approved):
Looks good!

casperlamboo:
Reject because the base of the PR is main instead of 5.2. Changing the base will move commits from the main branch in the 5.2 branch which we don't want. Going to cherry-pick the single commit to the 5.2 branch instead.

PR #13454, opened 2022-10-03
[CURA-9711] libspdlog missing from AppImage

Joeydelarago:


conandata.yml
```diff
@@ -87,6 +87,11 @@
                 src: "bin"
                 dst: "."
                 binary: "CuraEngine"
+            spdlog:
+                package: "spdlog"
+                src: "include"
```
jellespijker, line 92: This shouldn't work. The `include` directory of `spdlog` package contains the header files `*.h` not the shared library.

jellespijker: the `*.so` files are located in `.conan/data/spdlog/1.10.0/_/_/package/<hash>/lib/`

so it should probably be something like:
```suggestion
                src: "lib"
```

but this might give trouble on Windows because on Windows it will be probably a `dll` and stored in the `bin` folder

jellespijker (changes requested):
If the `spdlog` binary is now located in the AppImage, it isn't because of these changes. You are using the `include` directory as the source folder which only contains the `spdlog` header files. On Unix systems the actual shared library will be located in `lib` while on Windows in `bin`.

I would suggest that you tryout creating a new AppImage without these changes. If that AppImage conatins the `spdlog` binaries, then something else went wrong during the running of the cura-installer workflow.

conanfile.py
```diff
@@ -208,6 +208,7 @@ def _generate_pyinstaller_spec(self, location, entrypoint_location, icon_path, e
             else:
                 continue
             if not src_path.exists():
+                self.output.warning(f"Source path for binary {binary['binary']} does not exist")
                 continue
             for bin in src_path.glob(binary["binary"] + ".*[exe|dll|so|dylib]"):
```
jellespijker, line 213: Since it also complains about *.so.1` this glob statement should probably be something like:
```suggestion
            for bin in src_path.glob(binary["binary"] + ".*[exe|dll|so|dylib]*"):
```

PR #13472, opened 2022-10-04
[CURA-9711] fix spdlog binaries missing in linux bulids

Joeydelarago:
Add spdlog to binaries list.

Add warning for missing binary

Update pattern matching for binaries glob in bin paths. It will now match .so.X files that it would not before.

CURA-9711


jellespijker:
@Joeydelarago I think the following should result in a more generic solution for when we add other dependencies
```
        # Make sure all Conan dependencies which are shared are added to the binary list for pyinstaller
        for _, dependency in self.dependencies.host.items():
            if hasattr(dependency.options, "shared") and dependency.options.shared:
                for bin_paths in dependency.cpp_info.bindirs:
                    binaries.extend([(f"{p}", ".") for p in Path(bin_paths).glob("**/*")])
                for lib_paths in dependency.cpp_info.libdirs:
                    binaries.extend([(f"{p}", ".") for p in Path(lib_paths).glob("**/*")])
```

jellespijker:
Started a workflow for the binaries: https://github.com/Ultimaker/Cura/actions/runs/3188584796

PR #13488, opened 2022-10-06
Add ELEGOO printers files

NARUTOfzr:
I optimized files: elegoo_neptune_2.def. json&elegoo_neptune_2d.def.json, and added other ELEGOO printer files.


NARUTOfzr (commented):
Add ELEGOO printers files.
If there is any problem in any of the files, please point out or help to correct the wrong place, thank you in advance!

PR #13512, opened 2022-10-09
Added the BIQU Hurakan

looxonline:
Profile for the BIQU Hurakan printer that will soon be released. Tested and slices fine.


jellespijker:

jellespijker (approved):
LGTM

I do think that the size of the stl could be even more reduced, since it is basically a box. But I'm okay with merging this

jellespijker:
Thank you for your contribution.Your printer should be available in our next release Cura 5.3

I would appreciate it if you can take another look at the stl, see if you can make it a bit smaller. It isn't a show stopper, but Cura is big enough as it is already

looxonline:
> Thank you for your contribution.Your printer should be available in our next release Cura 5.3
> 
> I would appreciate it if you can take another look at the stl, see if you can make it a bit smaller. It isn't a show stopper, but Cura is big enough as it is already

Will see what I can do. Pretty sure I can crunch it down. No point in using space with no need. Re: the suggested accel change. Thanks for keeping it machine specific.

Issue #12343, opened 2022-05-24
Cura 5.0.0 AppImage unable to start


PavelStoyan0v:
### Application Version

5.0.0

### Platform

Zorin OS 16 (Ubuntu)

### Printer

Ender 3 V2

### Reproduction steps

I am trying to run the AppImage for 5.0.0 but Cura doesn't start.
I see that the files in .local/share/cura are separated for each version so probably it isn't a conflict between the old and the new versions. In any case the previous version that I am still using is 4.13.0. 

### Actual results

The loading screen shows and after Loading UI it just disappears.
If I execute the appimage in terminal the final line says:
**QQmlComponent: Component is not ready**

### Expected results

Expected cura to start.

### Checklist of files to include

- [X] Log file
- [ ] Project file

### Additional information & file uploads

[cura.log](https://github.com/Ultimaker/Cura/files/8765834/cura.log)

karcaw:
same issue here

v6-cblais:
Seeing the same issue as well on Zorin16. Found that executing with sudo permitted it to run.

nallath:
We've not been able to figure out what is going wrong here, so we will need more information. 

Could someone provide us with the lines of logging above "QQmlComponent: Component is not ready" and all the lines there when cura does run when with Sudo?

C-Blais:
Happy to help!

Without sudo:
$ ./Ultimaker-Cura-5.0.0-linux.AppImage 
PyInstaller/loader/pyimod03_importers.py:495: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
QApplication: invalid style override 'adwaita' passed, ignoring it.
	Available styles: Windows, Fusion
QQmlApplicationEngine failed to load component
QQmlComponent: Component is not ready



With sudo:
# ./Ultimaker-Cura-5.0.0-linux.AppImage 
PyInstaller/loader/pyimod03_importers.py:495: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
QObject::startTimer: Timers cannot be started from another thread
[...]

PR #13525, opened 2022-10-11
Fix startup on Zorin OS

Tim453:
First unset QT_STYLE_OVERRIDE, then start Cura

Should fix https://github.com/Ultimaker/Cura/issues/12343


jellespijker:
Seems a relative safe and quick fix. Thank you for your contribution.

I will target this PR to our 5.2 branch so it will be included in the upcoming release.

PR #13544, opened 2022-10-12
[CURA-9689] Filter material lists for compatible printers.

rburema:
So that only one identical entry shows up per material, per extruder (per printer).


casperlamboo (approved):
Looks very clear and readable!

PR #13546, opened 2022-10-12
[CURA-9746] Re-add tooltips for intent profile selection in recommended.

rburema:
They where skipped over when transferring the radio bar to the current large button based setup.

### DIFF
--- a/resources/qml/PrintSetupSelector/Recommended/RecommendedQualityProfileSelector.qml
+++ b/resources/qml/PrintSetupSelector/Recommended/RecommendedQualityProfileSelector.qml
@@ -32,7 +32,7 @@ Item
             {
                 profileName: model.name
                 icon: model.icon
-
+                tooltipText: (model.description != undefined) ? model.description : ""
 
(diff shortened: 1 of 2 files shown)


resources/qml/PrintSetupSelector/Recommended/RecommendedQualityProfileSelector.qml
```diff
@@ -32,7 +32,7 @@ Item
             {
                 profileName: model.name
                 icon: model.icon
-
+                tooltipText: (model.description != undefined) ? model.description : ""
```
casperlamboo, line 35: checking against `undefined` is [bad practice](https://stackoverflow.com/questions/3390396/how-can-i-check-for-undefined-in-javascript) in javascript. The recommended pattern is `typeof model.description !== "undefined"`.

However, javascript also has `null` values, to also guard against that I like to write something similar to
```suggestion
                tooltipText: model.description ? model.description : ""
```

casperlamboo (approved):

PR #13596, opened 2022-10-18
Build mo translation files when creating conan package

jellespijker:
Added a build step which uses the Conan package gettext as a tool_requires to convert the po files to mo files and store these in the resources folder

When on Windows the msys2 recipe is used to ensure that we have a bash environment to run gnu gettext

Creating the mo files as part of the conan package will ensures that we no longer need to store them in the cura-binary-data and generate them manually. This should result in less risk of human error during the release cycle

See also:
- https://github.com/Ultimaker/Uranium/pull/858
- https://github.com/Ultimaker/cura-binary-data/pull/26


Joeydelarago (approved):
Doing this manually last time made me understand why repetitive work causes human error!

conanfile.py
```diff
@@ -274,6 +275,11 @@ def requirements(self):
             for req in self._um_data()["internal_requirements"]:
                 self.requires(req)
 
+    def build_requirements(self):
+        if self.settings_build.os == "Windows" and not self.conf.get("tools.microsoft.bash:path", default=False, check_type=bool):
+            self.tool_requires("msys2/cci.latest")
```
rburema, line 280: Maybe a good idea to only do this if translations are required? (Have an option to be able to build without translations.)

conanfile.py
```diff
@@ -284,7 +290,15 @@ def layout(self):
         self.cpp.package.resdirs = ["resources", "plugins", "packaging", "pip_requirements"]  # pip_requirements should be the last item in the list
 
     def build(self):
-        pass
+        if self.settings.os == "Windows":
+            self.win_bash = True  # We need gettext, which requires the bash environment
```
rburema, line 294: I know you're probably doing this because the gettext conan package relies on it, but we have been able to get it to work on Windows without, I think. (Don't ask how ... I've forgotten :-p )

rburema (approved):

rburema (commented):
Results from actual run on Windows:

~~~
checking for tex... no
checking for yacc... yacc
checking for lex... lex
checking whether autoconf is installed... no
configure: error: Autoconf 2.65 or better is required.
    Please make sure it is installed and in your PATH.
automake/1.16.5:
automake/1.16.5: ERROR: Package '546cb51b20c0a40710e1afe718b043478c8f06be' build failed
automake/1.16.5: WARN: Build folder C:\.conan\05add6\1
ERROR: automake/1.16.5: Error in build() method, line 88
        autotools = self._configure_autotools()
while calling '_configure_autotools', line 72
        self._autotools.configure(args=conf_args, configure_dir=self._source_subfolder)
        ConanException: Error 1 while executing /c/.conan/05add6/1/source_subfolder/configure --datarootdir=/c/.conan/c11017/1/res --prefix=/c/.conan/c11017/1 --target=x86_64-windows-msvc
~~~

Joeydelarago (approved):

PR #13655, opened 2022-10-26
Update creality_ender3s1.def.json

lin-ycv:
- Adds UBL adjustment to start gcode
- Adjustments to match the profile shipped with the printer
  - reduce lift from 10 to 2
  - nozzle and bed already waits for preheat after file is selected, no need for M104,190,109


lin-ycv:
some errors were made, needs more testing. closing PR

PR #13666, opened 2022-10-26
added definition for Creality CR-10 Smart

eropple:
Hey there. Noticed that there wasn't a preset for the CR-10 Smart, so I thought I'd add one.

It's a bit of a weird printer (aren't they all?) in that it has the dimensions of a CR-10 (hence the name) but the extruder and hotend match the CR-6 SE more; that said, Cura doesn't do any differentiating there, so I went with the one most people would think of first, and inherited from `creality_cr10`. I don't think it matters much either way. (later edit: I had to change it to inherit from `creality_base` in order to make it work with qualities correctly, even with `quality_definition` defined - that's very strange, but given how few adjustments are made to printers in these profiles I don't think it matters. Pointers would be gratefully appreciated, though.)


jellespijker:

jellespijker (approved):
LGTM

jellespijker:
Thank you for contributing, Your printer should be available in our next release Cura 5.3

PR #13702, opened 2022-11-01
Fix test failture of definition container

parachvte:
Error output:
```
result = VersionUpgradeManager.getInstance().updateFilesData(configuration_type, version,
                                                                         [serialized],
                                                                         [file_name if file_name else ""])
E           AttributeError: 'NoneType' object has no attribute 'updateFilesData'
```

The `DefinitionContainer` is calling `VersionUpgradeManager.getInstance()` to initialize (upgrade) definition container, while  the `VersionUpgradeManager.__instance` singleton is not being initialized in pytest.

I change the `DefinitionContainer` to simply using json to parse and test the definition file.

### DIFF
--- a/tests/Settings/TestDefinitionContainer.py
+++ b/tests/Settings/TestDefinitionContainer.py
@@ -55,9 +55,13 @@ def test_noCategory(file_path):
     :param file_path: The path of the machine definition to test.
     """
     with open(file_path, encoding = "utf-8") as f:
-        json = f.read()
-        metadata = DefinitionContainer.deserializeMetadata(json, "test_container_id")
-        assert "category" not in metadata[0], "Definition located at [%s] referenced a category, which is no longer allowed" % file_path
+        doc = json.load(f)
+
+    if "metadata" not in doc:
+        return
+
+    metadata = doc["metadata"]
+    assert "category" not in metadata, "Definition located at [%s] referenced a category, which is no longer allowed" % file_path
 
 
 @pytest.mark.parametrize("file_path", machine_filepaths)


Joeydelarago:
This change causes the test to ignore the inherited metadata of the definition and any future changes to deserializeMetadata. This could cause this test to pass incorrectly. 

VersionUpgradeManager is initialized when running the whole "Settings" test suite (see[ test configuration file](https://github.com/Ultimaker/Cura/blob/main/tests/Settings/conftest.py)). If you don't want to run all the tests, you can run only the Settings test suite.

Conan will set you up some run configurations in Pycharm for this using [-g PyCharmRunEnv](https://github.com/Ultimaker/Cura/wiki/Running-Cura-from-Source#3-initialize-the-virtual-python-development-environment), but it is only set up for the full "pytest in Test" suite and single files. You can edit the pytest in Test suite to use this script target "/Users/j.delarago/PycharmProjects/Cura/tests/Settings" and that will run only the Settings suite.

I will put a ticket on our backlog to update our tests to point at the suites instead of single files so this will be easier in the future.

parachvte:
Thanks for the explanation!
Now I realize that `pytest` single file fails and the whole Settings test suite `pytest tests/Settings` succeeds.

As my PR, since all machine and extruder definitions are tested, I think it's actually no need to check inherited metadata. The only exception that not being tested is `fdmprinter.def.json` and `fdmextruder.def.json`. And I think they probably not gonna add deprecated keys. (Yeah but the changes are not needed)

nallath:
I think you should be using 

`python3 -m pytest -k TestDefinitionContainer`

That way all the collecting & setting up for pytest is ran, but you do select on running only "TestDefinitionContainer". I don't see a way around first doing the collecting, but that is relatively fast.

PR #13752, opened 2022-11-07
Update creality_ender3s1.def.json

lin-ycv:
Added `M420 S1 Z10` so printer loads bed mesh (S1 comes with ABL probe) also remove redundant heat up command for bed and nozzle, already heated at that point


jellespijker:

lin-ycv:
closing PR, `M420 S1 Z10` is not required on newer 24F4 firmware
redundant heat up command for bed and nozzle can be removed, but no harm in keeping them

PR #13774, opened 2022-11-09
Cura 9424 update loading projects design

casperlamboo:


plugins/PrepareStage/PrepareMenu.qml
```diff
@@ -55,6 +55,50 @@ Item
...
+                    Cura.SecondaryButton
+                    {
+                        id: addPrinterButton
+                        leftPadding: UM.Theme.getSize("default_margin").width
+                        rightPadding: UM.Theme.getSize("default_margin").width
+                        text: catalog.i18nc("@button", "Add printer")
+                        // The maximum width of the button is half of the total space, minus the padding of the parent, the left
+                        // padding of the component and half the spacing because of the space between buttons.
+                        fixedWidthMode: true
+                        width: parent.width / 2 - leftPadding * 1.5
```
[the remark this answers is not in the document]
casperlamboo: Good point!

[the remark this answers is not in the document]
casperlamboo: https://github.com/Ultimaker/Cura/pull/13774/commits/79a7b1e0649b71de2dd5d41b97ad9ec7ce42541e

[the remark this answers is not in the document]
fieldOfView: There's another instance of this 16 lines below

cura/Machines/Models/MachineListModel.py
```diff
@@ -79,17 +78,33 @@ def _onContainerChanged(self, container: ContainerInterface) -> None:
     def _updateDelayed(self) -> None:
         self._change_timer.start()
 
+    def _getMachineStacks(self) -> List[ContainerStack]:
+        return CuraContainerRegistry.getInstance().findContainerStacks(type="machine")
```
jellespijker, line 82:
```suggestion
        return CuraContainerRegistry.getInstance().findContainerStacks(type = "machine")
```

jellespijker (commented):
Incomplete code-review, will cont. once it is finished

Joeydelarago:
I fixed the issue where there were random crashes when changing machines in the preview stage. I did this by not connecting the WorkspaceDialog MachineSelector to any signals. 

This shouldn't matter anyway since we are only choosing a printer to load a project. I would be surprised if anyone has this dialog open for more than a minute, and even if they did it would only be the online status of printers affected.

resources/qml/PrinterSelector/MachineSelectorList.qml
```diff
@@ -21,7 +21,15 @@ ListView
 
     section.delegate: UM.Label
     {
-        text: section == "true" ? catalog.i18nc("@label", "Connected printers") : catalog.i18nc("@label", "Other printers")
+        text: {
+            switch (section)
```
casperlamboo, line 25: Do we want to apply some defensive programming here and add a default case?

plugins/3MFReader/WorkspaceRow.qml
```diff
@@ -0,0 +1,34 @@
...
+    property alias rightLabelText: rightLabel.text
+
+    width: parent.width
+    height: visible ? childrenRect.height : 0
+
+    UM.Label
+    {
+        id: leftLabel
+        text: catalog.i18nc("@action:label", "Type")
+        width: (parent.width / 4) | 0
```
casperlamboo, line 24: I don't really like the `| 0` pattern. It flooring the value, and I believe we use it so we don't have non-integer widths. I rather use `Math.round` because I think rounding makes more sense then flooring, and I think it is more readable. `| 0` is supposed to be a bit faster though, but I don't think it is going to be noticeable.
```suggestion
        width: Math.round(parent.width / 4)
```

plugins/3MFReader/WorkspaceRow.qml
```diff
@@ -0,0 +1,34 @@
...
+        wrapMode: Text.WordWrap
+    }
+    UM.Label
+    {
+        id: rightLabel
+        text: manager.machineType
+        width: (parent.width / 3) | 0
```
casperlamboo, line 31:
```suggestion
        width: Math.round(parent.width / 3)
```

plugins/3MFReader/WorkspaceSection.qml
```diff
@@ -0,0 +1,126 @@
...
+            text: ""
+            anchors.verticalCenter: parent.verticalCenter
+            font: UM.Theme.getFont("default_bold")
+        }
+    }
+
+    Item
+    {
+        id: comboboxTooltip
+        width: (parent.width / 2.5) | 0
```
casperlamboo, line 53:
```suggestion
        width: Math.round(parent.width / 2.5)
```

resources/qml/Widgets/ComboBox.qml
```diff
@@ -45,7 +46,7 @@ ComboBox
         State
         {
             name: "highlighted"
-            when: (base.hovered || control.hovered) && !control.activeFocus
+            when: ((base != undefined && base.hovered) || control.hovered) && !control.activeFocus
```
casperlamboo, line 49: I think it is a bit weird for such a generic component to check for a somewhat component with `id=base` somewhere up in the component tree. I'm not sure if it will break stuff but why not delete that property check all together?
```suggestion
            when: control.hovered && !control.activeFocus
```

plugins/PrepareStage/PrepareMenu.qml
```diff
@@ -55,6 +55,50 @@ Item
...
+                    Cura.SecondaryButton
+                    {
+                        id: managePrinterButton
+                        leftPadding: UM.Theme.getSize("default_margin").width
+                        rightPadding: UM.Theme.getSize("default_margin").width
+                        text: catalog.i18nc("@button", "Manage printers")
+                        fixedWidthMode: true
+                        // The maximum width of the button is half of the total space, minus the padding of the parent, the right
+                        // padding of the component and half the spacing because of the space between buttons.
+                        width: parent.width / 2 - rightPadding * 1.5
```
casperlamboo, line 94:
```suggestion
                        width: Math.round(parent.width / 2 - rightPadding * 1.5)
```

resources/qml/Widgets/ComboBox.qml
Joeydelarago: Moved this logic out into the components that need it.

PR #13778, opened 2022-11-09
Fix internal data plugins deployment

jellespijker:
The conan package from the internal package wasn't deployed, added the logic for the second list item


casperlamboo (approved):
Makes sense!

PR #13785, opened 2022-11-10
CURA-8890-hole-horizontal-expansion-max-diameter

rijkvanmanen:
See also backend PR: https://github.com/Ultimaker/CuraEngine/pull/1768


resources/definitions/fdmprinter.def.json
```diff
@@ -1298,6 +1298,18 @@
                     "limit_to_extruder": "wall_0_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "hole_xy_offset_max_diameter":
```
jellespijker, line 1301: Either describe this as an approximate diameter or change it to area.

resources/definitions/fdmprinter.def.json
```diff
@@ -1298,6 +1298,18 @@
...
+                {
+                    "label": "Hole Horizontal Expansion Max Diameter",
```
jellespijker, line 1303: I think: `"Max Diameter"` should cover it already as child setting

jellespijker (changes requested):
See PR review front-end

PR #13848, opened 2022-11-17
[CURA-9424] Workspace dialog fixes

Joeydelarago:
Fix minor styling issues.

Fix opening the same project file with a new pinter multiple times causing a failure to load. 

CURA-9424

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -600,7 +600,6 @@ class ThreeMFWorkspaceReader(WorkspaceReader):
         self._dialog.setNumUserSettings(num_user_settings)
         self._dialog.setActiveMode(active_mode)
         self._dialog.setUpdatableMachines(updatable_machines)
-        self._dialog.setMachineToOverride(global_stack_id)
         self._dialog.setMaterialLabels(material_labels)
         self._dialog.setMachineType(machine_type)
         self._dialog.setExtruders(extruders)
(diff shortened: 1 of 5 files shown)


plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -96,7 +96,7 @@ UM.Dialog
                         WorkspaceRow
                         {
                             leftLabelText: catalog.i18nc("@action:label", manager.isPrinterGroup ? "Printer Group" : "Printer Name")
-                            rightLabelText: manager.machineName
+                            rightLabelText: manager.machineName == "Create new" ? "" : manager.machineName
```
casperlamboo, line 99: Don't we want to translate the "Create new" string?

resources/qml/RoundedRectangle.qml
```diff
@@ -1,3 +1,4 @@
+
```
casperlamboo, line 1: Don't think we need this empty line in the beginning of the file
```suggestion
```

plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -616,6 +616,7 @@ def preRead(self, file_name, show_dialog=True, *args, **kwargs):
             # The machine included in the project file exists locally already, no need to change selected printers.
             is_networked_machine = global_stack.hasNetworkedConnection()
             is_abstract_machine = parseBool(existing_global_stack.getMetaDataEntry("is_abstract_machine", False))
+            self._dialog.setMachineToOverride(global_stack.getId())
```
casperlamboo, line 619: Do we still need this? I thought the fix from https://github.com/Ultimaker/Cura/pull/13848/commits/4d2f6b9655dd4cecf9eef0f1f808b05e3b500d28#diff-bc57941f6914b8706cf435c14b5d522814678acc5f510b69653cf5a538e9839aR735 already fixed this?

plugins/3MFReader/WorkspaceDialog.qml
casperlamboo: I do think it's a bit cleaner if the machineName is fully set through the manager. So if no compatible machine is found the `ThreeMFWorkspaceReader.py` sets `machineName="Create new".

casperlamboo (approved):

PR #13852, opened 2022-11-19
Support custom gcode macro on filament change plugin

victornpb:
I added 3 new fields to the FilamentChange (PostProcessing Plugin)
 - Enable
   This is consistent with the ChangeAtZ. My use case is that I have to add and then remove this plugin between prints, and every time I want to print dual colors, I have to reconfigure it. So it makes sense to provide a way to disable it when you don't need it, and settings are preserved.
 - G-code Before
   I found it surprising this wasn't an option since it seems like an obvious addition. I wanted to add a beep, so I was using the find-replace plugin but it is very unergonomic and error prone.
 - G-code After
   I have to catch the purge with tweezers when I hit "continue", so it makes sense to add a purging and wiping sequence if you have the appropriate hardware.

The macros have separate "Enable" checkboxes in order to make it easier to toggle the custom Gcode without having to constantly erase and copy/paste G-code into the text boxes. So it is a small quality of life thing.


casperlamboo (changes requested):
Hi @victornpb,

Thanks for opening the PR, all looks good and this should be a nice addition to Cura. I did have some cosmetic remarks though.

I see that you prepend each gcode statement with a linebreak; lines like these are pretty common.
```py
color_change = color_change + "\n" + some_statement
```
However, I like this pattern better
```py
color_change = color_change + some_statement + "\n"
```
I like this pattern better for various reasons
- Each statement is responsible for its "own line" and should clear it
- There is no need for special cases for the first and last added statements

I replied to this PR with suggestion how I _think_ it should look. Please conciser these as suggestions; I have not tested them yet.

Once the remaining issues are resolved the PR can be merged.

Thanks,
@casperlamboo

casperlamboo (approved):

PR #13858, opened 2022-11-20
Auto format printer/profiles with printer-linter

jellespijker:
Runs only master and release branch, will create a co-authored commit if formatting was applied

See the following commit in a test repo https://github.com/Ultimaker/cura-action-tests/commit/489bf00d39c8957c0882167a025579593c6ab905


.github/workflows/printer_linter_format.yml
```diff
@@ -0,0 +1,45 @@
+name: printer-linter-format
+
+on:
+    push:
+        branches:
+            - main
+            - '[1-9].[0-9]'
+            - '[1-9].[0-9][0-9]'
```
Joeydelarago, line 8: Do we also want to run this on PRs?

Joeydelarago (approved):

PR #13873, opened 2022-11-21
Printer linter pr review

jellespijker:
Using the pull-request-comments-from-clang-tidy-reports GH action
https://github.com/marketplace/actions/pull-request-comments-from-clang-tidy-reports

Which is possible because the printer-linter auto
puts the suggested fixes in the same format as
clang-tidy.

It consists of two workflows, to ensure safe handling
of forks PR's. The first workflows is the analysis and
the suggested fixes, these are then upload to the
second workflow which will use the suggested fixes
to past comments on the PR

See: https://github.com/Ultimaker/cura-action-tests/pull/10 for an example PR


jellespijker:
Note: Test are probably failing due to a pi caching issue with the other related PR

.github/workflows/printer-linter-pr-post.yml
```diff
@@ -0,0 +1,80 @@
...
+                        let download = await github.actions.downloadArtifact({
+                            owner: context.repo.owner,
+                            repo: context.repo.repo,
+                            artifact_id: matchArtifact.id,
+                            archive_format: "zip",
+                        });
+                        let fs = require("fs");
+                        fs.writeFileSync("${{github.workspace}}/printer-linter-result.zip", Buffer.from(download.data));
+
+            -   name: Extract analysis results
```
Joeydelarago, line 70: I see you are extracting these results twice using two different versions of the checkout action.  I assume it is for a good reason. What is the reason?

Joeydelarago (approved):

.github/workflows/printer-linter-pr-post.yml
jellespijker: Does copy pasting the example workflow provided by the GH action count as a good reason?

It seemed to do no harm and I assumed they did it for a good reason as well.

But assumption is the mother of all fuck ups so best look at it with a critical eye.

PR #13917, opened 2022-11-28
Made the default skirt/brim extruder -1 (not overridden), this will c…

pkuiper-ultimaker:
…reate the a brim with the same material as the mesh for each mesh. Fixed the brim tower setting to reflect changes in the brim behavior. When the brim is enabled for the whole print it does not enable it for the prime tower anymore. Lastly we reduced the brim inside avoid distance to a smaller value to improve the adhesion of large inner objects.

Relates to PP-36


rijkvanmanen (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -5767,7 +5766,7 @@
                     "description": "If brim is only on outside then parts fully enclosed inside another part will get a brim which might overlap with the internal holes of the outer part. This setting controls how far to stay away from those internal holes. Set to a high value to prevent any brim from being generated for parts enclosed within the holes of other parts.",
                     "unit": "mm",
                     "type": "float",
-                    "default_value": 5,
+                    "value": "2.5",
```
nallath, line 5769:
```suggestion
                    "default_value": "2.5",
```
A setting must always have a default value. The engine can't use this folder otherwise.

nallath (changes requested):

resources/definitions/fdmprinter.def.json
```diff
@@ -5606,8 +5606,7 @@
                             "label": "Skirt/Brim Extruder",
                             "description": "The extruder train to use for printing the skirt or brim. This is used in multi-extrusion.",
                             "type": "optional_extruder",
-                            "default_value": "0",
-                            "value": "adhesion_extruder_nr",
```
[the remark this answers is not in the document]
pkuiper-ultimaker: value is deleted and default value is set to "-1", does that not mean that it defaults to "-1"? The resolve function will make sure the value is -1 if one of the extruders is -1. Is this incorrect?

resources/definitions/fdmprinter.def.json
pkuiper-ultimaker: both value and default value are set to "2.5" should I only choose one of the two?

resources/definitions/fdmprinter.def.json
```diff
@@ -5767,7 +5766,7 @@
                     "description": "If brim is only on outside then parts fully enclosed inside another part will get a brim which might overlap with the internal holes of the outer part. This setting controls how far to stay away from those internal holes. Set to a high value to prevent any brim from being generated for parts enclosed within the holes of other parts.",
                     "unit": "mm",
                     "type": "float",
-                    "default_value": 5,
+                    "default_value": "2.5",
```
nallath, line 5769:
```suggestion
                    "default_value": 2.5,
```

nallath (approved):

PR #13967, opened 2022-12-04
Update Sermoon V1 profile with hotend specifics

lggomez:
Added retraction defaults that go in line with the Creality Sprite hotend extruder settings

PR #13968, opened 2022-12-04
Clarify linter steps for running locally

lggomez:
Updated the README to clarify that Cura defintion files should be their relative paths from the root of the project; I just realized by doing #13967 that running it with an invalid file path yields an empty result instead of an error (although that is a separate issue from this PR):

```shell
XXX@pop-os:~/DEV/Cura$ python3 printer-linter/src/terminal.py "path/to/nowhere" --format --fix
Diagnostics: []
```


Joeydelarago:
Thanks for the contribution, I always love to see more documentation.

PR #13969, opened 2022-12-04
printer-linter: exit on invalid definition file paths

lggomez:
This PR prevents silent failure with empty diagnostics if an invalid file path was provided. Fixes behavior described in #13968


Joeydelarago:
Thanks, this is very helpful information to pass to the user.

PR #13987, opened 2022-12-07
Fixed error in 0.8mm profiles for PC and CPE+

pkuiper-ultimaker:
The 0.15mm mode was generated iso 0.2mm

Relates to: PP-277


rijkvanmanen (changes requested):
I would suggest for this PR to keep the original names and include the name changes in PP-222. Next to this I see one setting change for PC and a lot of settings change for CPEp.

pkuiper-ultimaker:
> I would suggest for this PR to keep the original names and include the name changes in [PP-222](https://ultimaker.atlassian.net/browse/PP-222). Next to this I see one setting change for PC and a lot of settings change for CPEp.

The names of the files have not been changed. They are still the old profile names. 
For some reason CPE+ shows a deleted file (0.15mm) and than an new file (0.2mm), but PC shows a rename of the file.
I think the files are correct, so I do not know what to change.

rijkvanmanen (approved):

jellespijker (approved):

PR #14009, opened 2022-12-10
Added profiles for SnakeOilXY printers

ChipCE:
Added profiles for [SnakeOilXY](https://github.com/SnakeOilXY/SnakeOil-XY) printers. 

By : John Smith (chip@snakeoildev.com)


ChipCE:
I fixed those redundant-override problem. 
I have a question: In future, will values in fdmprinter.def.json change without notice?

jellespijker:
yes fdmprinter.def.json can and will change in the future, since CuraEngine is constantly changing and evolving and our team of print process and materials keep optimizing settings for Cura and CuraEngine. 
These changes are under the care of us, the "Community Software" department. We ensure that all the suggested changes will benefit all our printers (including this one).

If you really think that a setting should always have a certain value even if the current value is the same as currently defined in fdmprinters. Feel free to define them in your profile, but try to motivate your choices. That way future developers know why they deviate from our best-practices

ChipCE:
Thank you for the clarification!
I removed those redundant.

resources/definitions/snakeoil_standard_base.def.json
```diff
@@ -0,0 +1,145 @@
...
+            "ultimaker_pva",
+            "ultimaker_tough_pla_black",
+            "ultimaker_tough_pla_green",
+            "ultimaker_tough_pla_red",
+            "ultimaker_tough_pla_white",
+            "ultimaker_tpu_black",
+            "ultimaker_tpu_blue",
+            "ultimaker_tpu_red",
+            "ultimaker_tpu_white"
+        ]
```
Joeydelarago, line 66: Hey, if you set the "material\_diameter" of the printer to 1.75 it won't show any of these materials because they are all for 2.85 printers. 

This way you don't have to exclude all these materials.

ChipCE: oh nice!
Than you!

jellespijker (approved):
LGTM

Thank you for your contribution. The new printers should be available in Cura 5.3

PR #14060, opened 2022-12-16
PP-286_make-buildplate-adhesion-extruder-optional_extruder

rijkvanmanen:
Since the brim_skirt extruder will be -1 by default, it does make sense to allow the parent setting to be -1 as well. This requires an extra resolve for the raft extruders, since these cant be -1. PP-286


pkuiper-ultimaker (approved):

jellespijker (approved):
LGTM

PR #14061, opened 2022-12-16
PP-282_update-seam-corner-preference

rijkvanmanen:
While waiting on further improving seam location, None gives currently the best results. PP-282


jellespijker (approved):
LGTM;

PR #14062, opened 2022-12-16
PP-288-increase-initial-layer-skin-flow

rijkvanmanen:
This makes adhesion more robust for inaccuracies in the first layer thickness. PP-288


jellespijker (approved):
LGTM;

PR #11518, opened 2022-02-21
add 3d printer profile

ANYCUBIC-3D:
Anycubic Kobra
Anycubic Kobra Max

PR #14083, opened 2022-12-19
Add anycubic kobra go

razzeee:
Hey,

this is the first time I'm doing this, so this will need more work.

This is heavily based on https://github.com/Ultimaker/Cura/pull/11518/files @ANYCUBIC-3D 

Some questions:

1. How do I test this?
2. The quality files were a lot of guess work, there's no way to export them for cura in the correct format is there? I always get `quality_changes` files, which I guess is then based of a referencing definition.

### DIFF
--- a/resources/quality/anycubic_kobra_go/anycubic_kobra_go_abs.inst.cfg
+++ b/resources/quality/anycubic_kobra_go/anycubic_kobra_go_abs.inst.cfg
@@ -1,0 +1,33 @@
+[general]
+version = 4
+name = Normal
+definition = anycubic_kobra_go
+
+[metadata]
+setting_version = 20
+type = quality
+quality_type = abs
+weight = 0
+global_quality = True
+material = generic_abs
+
+[values]
+acceleration_print = 1000.0
+acceleration_travel = 3000.0
+bottom_layers = 6
+brim_width = 4.0
+infill_pattern = lines
+infill_sparse_density = 25
+material_final_print_temperature = 230
+material_initial_print_temperature = 230
+material_print_temperature = 235.0
+material_print_temperature_layer_0 = 235.0
+optimize_wall_printing_order = True
+speed_print = 50.0
+top_layers = 6
+z_seam_corner = z_seam_corner_inner
+z_seam_type = sharpest_corner
+z_seam_x = 111
+
+inset_direction = ='outside_in' if (True) else 'inside_out'
+layer_height = 0.2
--- a/resources/definitions/anycubic_kobra_go.def.json
+++ b/resources/definitions/anycubic_kobra_go.def.json
@@ -1,0 +1,43 @@
+{
+    "version": 2,
+    "name": "Anycubic Kobra Go",
+    "inherits": "fdmprinter",
+    "metadata": {
+        "visible": true,
+        "author": "Kolja Lampe",
+        "manufacturer": "Anycubic",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
+        "preferred_material": "generic_pla",
+        "has_machine_quality": true,
+        "quality_definition": "anycubic_kobra_go",
+        "preferred_quality_type": "pla",
+        "machine_extruder_trains": {
+            "0": "anycubic_kobra_go_extruder_0"
+        },
+        "firmware_file": "MarlinChiron.hex"
+    },
+    "overrides": {
+        "machine_name": {
+            "default_value": "Anycubic Kobra Go"
+        },
+        "machine_heated_bed": {
+            "default_value": true
+        },
+        "machine_width": {
+            "default_value": 222
+        },
+        "machine_height": {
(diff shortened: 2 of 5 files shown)


Joeydelarago:
Hey @razzeee, you can test these by dropping them into your configuration folder. If you open Cura and navigate to "Help">"Show Configuration Folder"  in the toolbar it will open the config folder. This folder contains folders matching the ones from this pull request (definitions, extruders and qualities). Drop your files in the matching folders and they will appear in Cura after a restart. 

These files are already in the correct format for Cura to use. 

To test, please run some prints using **Cura** **5.2.1.**

Good luck!

P.S.

I did notice you have some failing tests as well. It looks like you have added some settings that don't exist in the current Cura version. 

```
"The following setting(s) ['outer_inset_first'] are defined in the quality /home/runner/work/Cura/Cura/resources/quality/anycubic_kobra_go/anycubic_kobra_go_abs.inst.cfg, but not in fdmprinter.def.json"
```

razzeee:
> These files are already in the correct format for Cura to use.

Yeah, but the files provided from anycubic are curaprofiles which seem to be a zip with two files that are of the type `quality_changes` and based on fdmprinter. I'm not sure how to correctly convert those.

Joeydelarago:
Hey @razzeee, you can copy any settings from the quality_changes (Custom Profiles) files to your printers quality files. 

All quality files have a definition, this definition is the printer the qualities can be applied to. All printer definitions inherit from fdmprinter.def.json. This means using `definition=fdmprinter` means that any printer can use this quality. This means that you can copy these settings from quality_changes to quality without any issues.

Also, you need to change your printer definition to end with .def.json 

Good luck

razzeee:
Only difference this has to the provided cura profiles, that I can tell is, that the curaprofiles have different values on global/extruder and these new ones have the same settings on both global and extruder.

razzeee:
Have been printing a bunch of PLA with this and it seems fine.

Joeydelarago:
Hey @razzeee, you have a few failing tests. If you click the details next to the warning it will show the failed tests. 

It looks like you have defined some settings that are no longer in use, among other things.

Are you testing this on the latest Cura version?

razzeee:
The latest stable. I've been looking at the errors, but I'm unsure how to fix it, as I'm not sure what's the new setting to replace it. Then again they might not be essential, if the last stable is ignoring the setting.

resources/quality/anycubic_kobra_go/anycubic_kobra_go_abs.inst.cfg
```diff
@@ -0,0 +1,33 @@
+[general]
+version = 4
+name = Suggested parameters for ABS_V1.0
```
Joeydelarago, line 3: I think you are misunderstanding the use of these qualities. The names should be things like Fast/Normal/Fine and you should use material = generic\_pla/generic\_abs in the metadata.

This way when the user selects generic\_pla/generic\_abs as their material this quality will show for them.

Joeydelarago (changes requested):
Just a small change to the qualities and removing the deprectad setting that is causing the tests to fail. 

Otherwise, looking good 👍

resources/quality/anycubic_kobra_go/anycubic_kobra_go_abs.inst.cfg
razzeee: FYI these names are copied from the profiles that get provide with the printer, will change it when I'm home again

razzeee: Also the other Kobra qualities seem similar to the one I commited, are they wrong/too old?

Joeydelarago (approved):

Joeydelarago (approved):

Joeydelarago:
Looks good 👍 Thanks for the contribution!

inpos:
Hi!
Can add quality config for PETG ?

inpos:
I guess there is a mistake in "machine_end_gcode" at retract.
```
G1 F2400 E-1 ; Retract filament 3mm at 40mm/s to prevent stringing
```
According to [Marlin g-codes doc](https://marlinfw.org/docs/gcode/G010.html) retract g-code is G10 next after M207

Here my example:
```
M207 F2400 S3 ; Set retract options
G10 ; Retract
```

razzeee:
I guess your right, that might be something not soo obvious from the prints?

inpos:
With G1 code there is very short tail with string remains at the end of the print.
However, after G10 code, the tail remains only less

Z3R0C00L13:
I'm new to this, is there any options for adding the AnyCubic Kobra Neo to the list of printers to Cura? Or perhaps some specific machine settings and/or profiles made specifically for the Kobra Neo? I seem to not be able to get good, smooth prints. Also fairly new to this stuff  😊

PR #14161, opened 2022-12-29
[CURA-9522] Material list styled oddly 

Joeydelarago:
This PR fixes two issues. 

1.  The material popups were not overlapping the brand popup. The lining of both were visible. This overlaps them slighly so we only see a single lining width.
2.  The materials list being too wide by one scrollbar width when opened for the first time. When opening the materials list the size of the popup is adjusted, this causes the scrollbar to disappear. For some reason this does not update the width/implicitWidth of the popup even though it is defined based using the scrollbar width. The fix here is to force the width adjustments in the onOpened function.

### DIFF
--- a/resources/qml/Menus/MaterialBrandMenu.qml
+++ b/resources/qml/Menus/MaterialBrandMenu.qml
@@ -94,6 +94,8 @@ Cura.MenuItem
         // We have to keep a count of itemHovered (instead of just a bool)
         property int itemHovered: 0
 
+        implicitX: parent.width - UM.Theme.getSize("default_lining").width
+
         MouseArea
         {
             id: submenuArea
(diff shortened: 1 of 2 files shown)


resources/qml/Menus/MaterialBrandMenu.qml
```diff
@@ -94,6 +94,8 @@ Cura.MenuItem
         // We have to keep a count of itemHovered (instead of just a bool)
         property int itemHovered: 0
 
+        implicitX: parent.width - UM.Theme.getSize("default_lining").width
```
casperlamboo, line 97: I imagine this value should be used in most cases, and as such should be de default value. Why not change the value in https://github.com/Ultimaker/Cura/blob/e99719a08fb32338aafaaa13646e483580bae7b3/resources/qml/Menus/MaterialBrandSubMenu.qml#L22 ?

Joeydelarago: Makes sense 👍 I had to set implicitX: parent.width in the nested MaterialBrandMenu because they require different values to line up correctly.

PR #14202, opened 2023-01-04
[CURA-10118] Change Only_List_USB

MariMakes:
Apply Sovengers suggested fix so more Macusers can print over USB as mentioned in https://community.ultimaker.com/topic/42417-fix-for-connecting-creality-ender-3-neo-to-mac-via-usb/

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -84,9 +84,8 @@ class USBPrinterOutputDeviceManager(QObject, OutputDevicePlugin):
             if container_stack.getMetaDataEntry("supports_usb_connection"):
                 machine_file_formats = [file_type.strip() for file_type in container_stack.getMetaDataEntry("file_formats").split(";")]
                 if "text/x-gcode" in machine_file_formats:
-                    # This is a fix for connecting tty/cu printers on MacOS and Linux
-                    only_list_usb = False if Platform.isOSX() or Platform.isLinux() else True
-                    port_list = self.getSerialPortList(only_list_usb)
+                    # We only limit listing usb on windows is a fix for connecting tty/cu printers on MacOS and Linux
+                    port_list = self.getSerialPortList(only_list_usb=Platform.isWindows())
             self._addRemovePorts(port_list)
             time.sleep(5)
 
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -9,6 +9,7 @@ from re import search
 
 from PyQt6.QtCore import QObject, pyqtSignal
 
+from UM.Platform import Platform
 from UM.Signal import Signal, signalemitter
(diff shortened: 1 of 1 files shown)


fieldOfView:
While I am sure that fixes some thing for some users, other users will get mad because Cura now starts bothering their other (non-usb) serial ports. I think that without a preference, or better yet a full on USB printing configuration dialog, I think this is a situation where you cannot win.

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -83,7 +83,7 @@ def _updateThread(self):
             if container_stack.getMetaDataEntry("supports_usb_connection"):
                 machine_file_formats = [file_type.strip() for file_type in container_stack.getMetaDataEntry("file_formats").split(";")]
                 if "text/x-gcode" in machine_file_formats:
-                    port_list = self.getSerialPortList(only_list_usb=True)
+                    port_list = self.getSerialPortList(only_list_usb=False)
```
Joeydelarago, line 86: FieldOfView made a good point about this annoying users without this issue. I think it would be a good idea to only set this to false when Cura is running on MacOS. You can see an example of this here

[https://github.com/Ultimaker/Uranium/blob/41e98b8686911384ef5088dac7b8d15e52e33cb1/UM/View/GL/OpenGLContext.py#L146](https://github.com/Ultimaker/Uranium/blob/41e98b8686911384ef5088dac7b8d15e52e33cb1/UM/View/GL/OpenGLContext.py#L146)

Also it would be good to add a comment that mentions the issue, so we can change this back in the future if it is fixed.

Joeydelarago (changes requested):

plugins/USBPrinting/USBPrinterOutputDeviceManager.py
```diff
@@ -83,7 +84,9 @@ def _updateThread(self):
             if container_stack.getMetaDataEntry("supports_usb_connection"):
                 machine_file_formats = [file_type.strip() for file_type in container_stack.getMetaDataEntry("file_formats").split(";")]
                 if "text/x-gcode" in machine_file_formats:
-                    port_list = self.getSerialPortList(only_list_usb=True)
+                    # This is a fix for connecting tty/cu printers on MacOS and Linux
+                    only_list_usb = False if Platform.isOSX() or Platform.isLinux() else True
+                    port_list = self.getSerialPortList(only_list_usb)
```
jellespijker, line 89:
```suggestion
                    port_list = self.getSerialPortList(Platform.isWindows())
```

jellespijker (changes requested):
All the other platforms (including possible future platforms) are unix based, therefor a simple toggled can be based on the `Platform.isWindows()` check.

jolaroux:
is this in 5.2.2? On macOS Ventura and having same issue with M1 MBP.

jellespijker:
it's merged to main, it should be in 5.3

MaximumIllusionDEV:
Having issues with this with my Ender 3 Max Neo!

PR #14213, opened 2023-01-05
dont allow minimum speed of zero

rijkvanmanen:
Zero minimum speed results in infinite layer times, causing the speed factor to be NaN. CURA-10127
backend PR: https://github.com/Ultimaker/CuraEngine/pull/1795


jellespijker (approved):
LGTM;

PR #13706, opened 2022-11-01
Elegoo Neptune series Printer Configuration

NARUTOfzr:
These are the config files that add the Neptune series from Elegoo.

PR #14231, opened 2023-01-07
Add Elegoo Neptune series Printer Configuration.

NARUTOfzr:
These are the config files that add the Neptune series from Elegoo.
#13706 #13488 #12656 I have incorporated elegoo-Neptune's machine configuration many times with no good results. In the end, the choice is to close the PR. I hope I didn't cause you any trouble.

#11255 It is not clear whether it was due to editing and deleting the file in this submission, which led to errors in the end. I did not modify these files this time and recreated independent configuration files. If there are any errors, please tell me, I will modify as soon as possible.


casperlamboo:
All looks good 👍  thank you for you contribution

PR #14232, opened 2023-01-07
request to add designing tool 

mspcbs:
I request to add new design tools like text, etc.


jellespijker:
I think you wanted to open an issue (Feature Requests) instead of a Pull Request, which is intended for contributing code.

Please use the new Feature issue template, and make sure that you fill it in as completely as possible. Simply stating: please add design tool is to broad a subject for us to even consider. 

https://github.com/Ultimaker/Cura/issues/new/choose 

I will close this Pull Request now. If your intent was to contribute code feel free to reopen and push your code changes. But also describe in a bit more detail what was added, how it works and why it was added.

Issue #14100, opened 2022-12-20
Retract Continue when using Relative Extrusion


GregValiant:
### Application Version

5.2.1

### Platform

Win10

### Printer

Ender 3 Pro

### Reproduction steps

Enable relative extrusion for a print.
Enable retract continue.
Slice and save the gcode file.

### Actual results

This happens throughout the file.  I hadn't noticed and the extruder was beating itself to death.

G1 F3000 X165.281 Y123.421 E0.18226        ;last extrusion
G1 F2100 **E-6.5**                                             ;normal retraction
;MESH:X-Y Ref Block.stl(1)
G1 F9000 X162.983 Y104.5 Z0.35 **E-7.45300** ;retract continue
G1 X160.123 Y104.5 Z0.35 **E-7.59600** ;retract continue
G1 X69.697 Y104.5 Z0.35 **E-12.11730** ;retract continue
G1 X66.837 Y104.5 Z0.35 **E-12.26030** ;retract continue
G1 X44.657 Y104.5 Z0.35 **E-13.36930** ;retract continue
;TYPE:WALL-INNER
G1 F2100 **E6.5**   ;normal prime - this should be 59.2959 so E6.5 leaves the filament just a tad short of the nozzle.
G1 F3000 X44.657 Y125.5 E2.82354   ;first extrusion

### Expected results

The deltas of the travel moves need to be calculated correctly.
The prime at the end must account for all the delta changes in the E.
G1 F3000 X7.359 Y97.869 E7.89636           ;Last extrusion
G1 F2100 E-6.5                                          ;retract filament -6.5
G1 F9000 X7.36 Y132.129 Z1.4 E-1.21300 ;"retract continue" with a proper additional retraction distance
G1 F2100 E7.713                                      ;prime to make up for all retractions to get the filament back to the nozzle.
G1 F3000 X41.899 Y97.589 E5.97052        ;first extrusion

### Checklist of files to include

- [ ] Log file
- [X] Project file

### Additional information & file uploads

This issue was mentioned in #7396 and #7458.  It looks like @rburema had a hand in making some fixes a couple of years ago.

This snippet is with Absolute Extrusion and appears to be fine.  There are incremental retractions through the travel moves and then the final prime accounts for all the delta changes.
G1 F3000 X222.739 Y154.857 E891.5317 ;Last Extrusion
G1 F2100 E885.0317                   ;retract 6.5
;MESH:NONMESH
[...]

PR #14342, opened 2023-01-20
[CURA-10092] fix/rewrite retract-continue relative mode

rburema:
See #14100 for a description of the problem. Note that the mention of needing to compensate was indeed the main issue, but (unmentioned in that issue) the series of 'continued' retractions was _also_ wrong in relative mode.

Relative mode was just never implemented with regard to this script it seems.

In the end, a (partial) rewrite turned out to be easier.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/RetractContinue.py
+++ b/plugins/PostProcessingPlugin/scripts/RetractContinue.py
@@ -6,11 +6,13 @@ from ..Script import Script
 from UM.Application import Application  # To get current absolute/relative setting.
 from UM.Math.Vector import Vector
 
+from typing import List, Tuple
+
 
 class RetractContinue(Script):
     """Continues retracting during all travel moves."""
 
-    def getSettingDataString(self):
+    def getSettingDataString(self) -> str:
         return """{
             "name": "Retract Continue",
             "key": "RetractContinue",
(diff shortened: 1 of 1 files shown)


plugins/PostProcessingPlugin/scripts/RetractContinue.py
```diff
@@ -27,58 +30,90 @@ def getSettingDataString(self):
...
+    def _getTravelMove(self, travel_move: str, default_pos: Vector) -> Tuple[Vector, float]:
+        travel = Vector(
+            self.getValue(travel_move, "X", default_pos.x),
+            self.getValue(travel_move, "Y", default_pos.y),
+            self.getValue(travel_move, "Z", default_pos.z)
+        )
+        f = self.getValue(travel_move, "F", -1.0)
+        return travel, f
+
+    def _travelMoveString(self, travel: Vector, f: float, out_e: float) -> str:
```
nallath, line 42: Nitpicking, "out_e" might need a bit of documentation. It wasn't immediately clear to me what it meant.

nallath (changes requested):

plugins/PostProcessingPlugin/scripts/RetractContinue.py
rburema: fixed by renaming parameter.

nallath (approved):

PR #14370, opened 2023-01-24
[CURA-10168] Add Interlocking to settings visibility

MariMakes:
Have Interlocking visible in Advanced, but the sub-settings in Expert Setting Visibility

### DIFF
--- a/resources/setting_visibility/expert.cfg
+++ b/resources/setting_visibility/expert.cfg
@@ -384,6 +384,12 @@ roofing_angles
 infill_enable_travel_optimization
 material_flow_dependent_temperature
 material_flow_temp_graph
+interlocking_enable
+interlocking_beam_width
+interlocking_orientation
+interlocking_beam_layer_count
+interlocking_dept
+interlocking_boundary_avoidance
 support_skip_some_zags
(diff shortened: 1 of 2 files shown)


casperlamboo (approved):

resources/setting_visibility/expert.cfg
```diff
@@ -384,6 +384,12 @@ roofing_angles
 infill_enable_travel_optimization
 material_flow_dependent_temperature
 material_flow_temp_graph
+interlocking_enable
```
jellespijker, line 387: @MariMakes and @casperlamboo shouldn't this be placed on level down in advanced

casperlamboo: Ah you're right! It was already there so it was already showing up correctly in the front end. 
https://github.com/Ultimaker/Cura/blob/540fca2409520d133c35ba49199d4092be406d0c/resources/setting_visibility/advanced.cfg#L142
But it shouldn't be duplicated. Will change this directly in main

casperlamboo: Correction, it should be there. If I remove it it will only show up in "advanced" and not in "expert"

Issue #13675, opened 2022-10-27
UFPWriter: SettingFunction is not JSON serializable


fuzebox1:
### Application Version

5.2.1

### Platform

Windows 10

### Printer

custom

### Reproduction steps

Slice
Send to Octoprint

### Actual results

Get error "object of type SettingFunction is not JSON serializable"

### Expected results

Send to Octoprint works as with 5.1.1.

### Checklist of files to include

- [X] Log file
- [ ] Project file

### Additional information & file uploads

Apparently this was a bug in the beta of 5.2, but was "fixed". I did not have the beta installed.
At bottom of log you will see an error from UFPWriter, making octoprint addon not able to proceed with write request.

[cura.log](https://github.com/Ultimaker/Cura/files/9881773/cura.log)

MariMakes:
Hey @fuzebox1,

Thanks for your report 👍
I'm not sure if this is something that we can solve in Cura. It seems to be related to the Octoprint plug-in. 
That is maintained by our wonderful community member @fieldOfView.

Are you sure you have the latest version?
You can find it here:
https://github.com/fieldOfView/Cura-OctoPrintPlugin/releases

fuzebox1:
Hi, I have the latest version, yes.
Fieldofview himself stated it is not a fault in his plugin, but in cura in a discussion thread on the 5.2 beta.
https://community.ultimaker.com/topic/41697-ultimaker-cura-52-released-in-beta/?do=findComment&comment=313331
The issue is said to have been fixed here: https://github.com/Ultimaker/Cura/pull/13473
but i am still getting it.

ckvsoft:
Same mistake
in the Linux Modern app image. Versions 5.2.0 and 5.2.1
Version 5.2 Beta2 works. This problem is not with all my printers.

I don't use Octoprint but send it with Connect Moonraker. But as I said in Beta 2 it works with both installed printers.

If I turn off UFP thumbnails it works.

[cura.log](https://github.com/Ultimaker/Cura/files/9942435/cura.log)

MariMakes:
Hey @fuzebox1 and @ckvsoft,

I brought your issue up to the team. :flex
They believe it was introduced by this change: https://github.com/Ultimaker/Cura/pull/13209

We've added a ticket to the backlog with the intent to improve this. 
For internal reference CURA-9859

Thanks for the report! 👍

shawnp30:
[...]

PR #14373, opened 2023-01-24
PP-71_improved-minimum-layer-time-behaviour

rijkvanmanen:
Align and improve minimum layer time settings. Detailed information can be found in PP-71.


resources/definitions/ultimaker.def.json
```diff
@@ -37,6 +37,11 @@
         "bridge_wall_material_flow": { "value": "wall_material_flow" },
         "bridge_wall_speed": { "value": "bridge_skin_speed" },
         "cool_fan_speed_0": { "value": "cool_fan_speed_min" },
+        "cool_fan_speed_max": { "value": "100" },
+        "cool_min_layer_time": { "value": 6 },
+        "cool_min_layer_time_fan_speed_max": { "value": "cool_min_layer_time + 5" },
+        "cool_min_speed": { "value": "round(speed_wall_0 * 3 / 4) if cool_lift_head else round(speed_wall_0 / 5)" },
```
jellespijker, line 43: It might be good practice to start using `resolveOrValue('<setting_name>')`, this will ensure that the settings can be serialized correctly. I don't think this particular setting will cause problems, but it is good practice imo
 https://github.com/Ultimaker/Cura/blob/main/docs/profiles/getting_a_setting_value.md#evaluating-a-value-property 

and

https://github.com/Ultimaker/Cura/issues/13675#issuecomment-1367995394

jellespijker (approved):

PR #14386, opened 2023-01-26
CURA-10201_fill-narrow-skin-with-walls

rijkvanmanen:
So far only skin with a width of one line was printed with a wall instead the normal pattern. But with this contribution the width of the skin regions that need to be filled with walls is exposed to the user interface.

Small skin region cause jerky motions for a normal lines pattern. Especially when printing at higher speeds you get more vibrations in the printer and the visual quality of the top layer is reduced. Since the printer needs to slow down for every turn also productivity is impacted.

backend: https://github.com/Ultimaker/CuraEngine/pull/1810


casperlamboo (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -1649,6 +1649,19 @@
                     "limit_to_extruder": "top_bottom_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "small_skin_width":
+                {
+                    "label": "Small Top/Bottom width",
```
casperlamboo, line 1654: Capitalize "width" for consistency.
```suggestion
                    "label": "Small Top/Bottom Width",
```

resources/definitions/fdmprinter.def.json
```diff
@@ -1649,6 +1649,19 @@
...
+                    "description": "Small Top/Bottom regions are filled with walls instead of the normal Top/Bottom pattern. This helps to avoids jerky motions.",
+                    "value": "skin_line_width * 2",
+                    "default_value": 1,
+                    "minimum_value": "0",
+                    "maximum_value_warning": "skin_line_width * 10",
+                    "type": "float",
+                    "enabled": "(top_layers > 0 or bottom_layers > 0) and top_bottom_pattern != 'concentric'",
+                    "limit_to_extruder": "top_bottom_extruder_nr",
+                    "settable_per_mesh": true
+                },
```
casperlamboo, line 1664: Add the unit field
```suggestion
                    "unit": "mm"
                },
```

PR #14434, opened 2023-01-31
increase cool_min_temperature

rijkvanmanen:
By default the cool_min_temperature is 15deg below print temperature. However, for PP and TPU this should be 10deg below print temperature. This is already changed for the other ultimaker printers, but I forget the UM3.

PP-71 / CURA-10214


jellespijker (approved):
LGTM

jellespijker:
@rijkvanmanen for a 5.3 fix try to base the branch from 5.3 and merge against that same release branch. That reduces the risk of us accidentally pulling in main in a release branch. I will cherry-pick this specific commit to 5.3

PR #14435, opened 2023-01-31
allow bridge speeds to be lower then cool_min_speed

rijkvanmanen:
With PP-71 the cool_min_speed is increased to a high values in case the lift_head is enabled. This causes the bridge_speeds to be lower than the cool_min_speed. Which is no problem, so the limit can be removed.

PP-71 / CURA-10211


jellespijker (approved):
LGTM

PR #14477, opened 2023-02-04
Add Weedo X40 PRO printer profile

x40-Community:
# Description

Please add the new Weedo X40 PRO printer profile


x40-Community:
Why aren't the finished print profiles included in Cura?

PR #14485, opened 2023-02-05
Create AddCoolingProfile.py

GregValiant:
New post processor

# Description

A new post processor to add a cooling profile By Layer or By Feature.

## Type of change

# How Has This Been Tested?

Cura versions:  5.2.1, 5.2.2, 5.3a, 4.12.1

- [X] Test A - In Cura with Cooling enabled and disabled.  With rafts, skirts, brims.  Various printers.

**Test Configuration**:
* Operating System:  Windows 11

# Checklist:

- [X] I have read the [Contribution guide](https://github.com/Ultimaker/Cura/blob/main/contributing.md) 

### DIFF
--- a/resources/variants/creality/creality_ender3s1pro_0.2.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_0.2.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]
--- a/resources/variants/creality/creality_ender3s1pro_0.3.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_0.3.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]
--- a/resources/variants/creality/creality_ender3s1pro_0.5.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_0.5.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]
--- a/resources/variants/creality/creality_ender3s1pro_0.6.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_0.6.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]
--- a/resources/variants/creality/creality_ender3s1pro_0.8.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_0.8.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]
--- a/resources/variants/creality/creality_ender3s1pro_1.0.inst.cfg
+++ b/resources/variants/creality/creality_ender3s1pro_1.0.inst.cfg
@@ -5,7 +5,7 @@ version = 4
 
 [metadata]
 hardware_type = nozzle
-setting_version = 20
+setting_version = 22
 type = variant
 
 [values]


GregValiant:
Well...maybe the ninth time is the charm.

GregValiant:
Fixed an error in Prime Tower fan speed.

GregValiant:
@rburema ,
When I first designed this I had not anticipated being able to have multiple instances running.  When I added "Remove M106 Lines" as an option - it did not occur to me that it no longer needed all the "By Layer" textfields.  I've chopped that down to 8 and made the adjustments in the code to account for that.
It's still a big file but should be a bit simpler for the user.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                number_of_raft_layers += 1
+                if raft_start_index == 0:
+                    raft_start_index = l_num
+            if layer.startswith(";LAYER:0"):
+                layer_0_index = l_num
+                break
+
+#Is this a single extruder print on a multi-extruder printer? - get the correct fan number for the extruder being used.
+        if is_multi_fan:
+            T0_used = False
```
rburema, line 401: If you make an array of booleans of these, it would simplify the code below.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                lines = data[num]
+                if "T0" in lines:
+                    T0_used = True
+                if "T1" in lines:
+                    T1_used = True
+                if "T2" in lines:
+                    T2_used = True
+                if "T3" in lines:
+                    T3_used = True
+            if T0_used and (T1_used or T2_used or T3_used):
```
rburema, line 415: ... besides the (maybe?) obvious, it would allow to just count the number of extruders used at the end, then set the `is_multi_extr_print` to the right value based on that.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                is_multi_extr_print = True
+            else:
+                is_multi_extr_print = False
+            
+            init_fan = T0_fan
+            if is_multi_fan and not is_multi_extr_print:
+                startup = data[1] #On a multi-extruder printer and single extruder print find out which extruder starts the file.
+                lines = startup.split("\n")
+                for line in lines:
+                    if line == "T1":
```
rburema, line 431: If the `Tx_fans` where an array instead (see above), it would allow for something a bit shorter, like:

~~~
for line in lines:
    for x in Tx_fans:
        if line == "T"+str(i):
            Tx_fans[0] = Tx_fans[i]
~~~

By the way, I think your goal is to make this work for older versions of Cura as well, but if we merge it in, it'll only be in that version.

If you're ok with that, you could also use f-strings instead of string-concatenation by '+'.

So then you could to, for example, `f"T{i}"` instead of `"T"+str(i)` in the above.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                        T0_fan = T1_fan
+                    elif line == "T2":
+                        T0_fan = T2_fan
+                    elif line == "T3":
+                        T0_fan = T3_fan
+            elif is_multi_fan and is_multi_extr_print:
+                startup = data[1] #On a multi-extruder printer and multi extruder print find out which extruder starts the file.
+                lines = startup.split("\n")
+                for line in lines:
+                    if line == "T0":
```
rburema, line 441: Same here...

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                        modified_data += "M106 S0" + prev_fan + "\n" #Turn off the prev fan
+                        modified_data += line + "\n"
+                        modified_data += "M106 S" + str(current_fan_speed) + this_fan + "\n"   #Turn on the current fan
+                        prev_fan = this_fan
+                    if ";LAYER:" in line:
+                        layer_number = str(line.split(":")[1])
+                        modified_data += line + "\n"
+                    if int(layer_number) >= int(the_start_layer):
+                        if ";TYPE:SKIRT" in line:
+                            modified_data += line + "\n"
```
rburema, line 669: Maybe consider making these three _almost_ repeated lines into a (local) function, so you only have to write a call to that function for each 'if' here.

Something like this:

~~~
# Note that this funcion needs to be a local function in order to use `modified_data`, `current_fan_speed` or `this_fan`. If you make it into a normal class function, those parameters will have to be passed as well.
def writeModifiedData(line, fan_sp_type_str):
    modified_data += line + "\n"
    modified_data += fan_sp_type_st + this_fan + "\n"
    current_fan_speed = str(fan_sp_type_str.split("S")[1])

if ";TYPE:SPAM" in line:
    writeModifiedData(line, fan_sp_eggs_spam_spam_eggs)
~~~

(Or, for bonus points: Instead of doing the above, put the `";TYPE:xxx"` and `fan_sp_xxx` in a dictionary `{}` and loop over them with yet another for-loop. -- Though you'd still need to check for the ones that _aren't_ a variation of the same three lines.)

rburema (changes requested):
Hi @GregValiant !

Sorry it took a while. Thanks for the contribution! I understood that you'd also appreciate a bit of feedback in the code, so I've tried to give examples when I talk about something.

I've suggested some more general changes for the first round, that can be applied around the code.

For now, I've only reviewed the AddCoolingProfile, not the other one file, nor the older version of the first one. (Because that's what the PR is called, and what I think you _intended_ to pull in for now.) Are you sure you meant to pull in your main branch? I see you also have `GV-AddCoolingProfile` and `GV-AlterZhops`.

(I usually only point things out a second time when I thought of something new to say about it. If you need it in more detail, please say so.)

Hope this helps. If you need more explanation, please ask!

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+                    "minimum_value": 0,
+                    "maximum_value": 100,
+                    "unit": "%    ",
+                    "enabled": "fan_enable_raft"
+                }
+            }
+        }"""
+
+    def execute(self, data):
+#Initialize variables that are buried in if statements.
```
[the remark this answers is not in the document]
GregValiant: Comment indents adjusted.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+        T0_fan = " P0"; T1_fan = " P0"; T2_fan = " P0"; T3_fan = " P0"; is_multi_extr_print = True
```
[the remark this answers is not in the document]
GregValiant: The support for additional fan circuits is complex but once you get above 1 then it really isn't any more difficult to add more.  I see this post processor being of greater value to industrial users doing large prints.  We both know that all the complaints will come for the dnd and "mini" crowd.
I waffled on an array or list.  In the end it was only a couple of lines of code difference.  It's already as long as War and Peace.and cutting a few lines out just didn't have much benefit.  It runs faster than I thought it would.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
GregValiant: Thanks @rburema.
Congratulations on figuring out which version was the latest one.  Every time I try to make an adjustment to the pull request here on Github I've ended up making a mess.

Personally, I print a lot of PETG and finer control of the cooling fan is something that I had wanted for a while.  Without a post processor like this - the fan gets turned on, maybe ramps up (sometimes at speeds that don't get it spinning), finally gets going, and that's it until it gets turned off at the end.  There is no regard for what a print might require in the middle.  I've attempted to address that.
A couple of years ago I made a feature request for this type of post processor but nobody picked up the ball and ran with it.  As a consequence, I decided to do it myself and so you get stuck with me and my very amateurish coding style.

In my defense, I was teaching myself Python as I went along.  I think this Cooling post processor would be an ambitious endeavor even for someone who knows the language.  I am aware that my coding style is decidedly "un-Pythonic".  I cut my programming teeth on an Analog computer and the last programming class I took was Fortran IV in 1973 (we used punch cards to load a program into the IBM360 that took up a very large room).  I see myself as a decent designer, but a poor coder.  I think the bottom line is that the script runs, works, and does what it is supposed to do.  Considering the number of insertions it can make, it's pretty fast too.
I see this as being of special use in industrial situations and printing things like ABS, ASA etc. where some cooling might be required for specific areas, but not in others.  That's why I went with 4 fans and 4 extruders.  In both cases, once you add a second extruder or fan circuit you any additional ones essentially for free.  Many of your comments appear to apply to both of those situations.  I solved the "multi-extruder" and "multi-fan" situation to the best of my ability.

I'm going through your suggestions now.  It seems that most have to do with "methodology" rather than any real changes to the actual logic, and since debugging is a PITA, I likely won't make any real changes to the code.  If you would like to then absolutely be my guest.  I don't want to be flip here, but tightening the code further makes it more difficult for me to read and understand.  There is also the question of repairing the breakage it would cause.  That takes me back to the "debugging is a PITA" thing.
An example is the way I emulate a multi-dimensional array with the "odd/even" values in the fan_list[ ].    I'm actually proud of my workaround since a multi-dimensional array isn't supported by Python, but it isn't like I'm in love with it.  Throwing it away in favor of a second list wouldn't cause me any heartburn.

In regard to the other pull requests - The Alter Z-hops is a new post processor.  There are prints that require Z-hops for a specific range of layers but would not be required for other areas.  It addresses that.
The others are re-writes.  You get a break on those as at least some of the code is written by an actual programmer rather than some random fireman who scribbles.

I did manage to get messages to the user in "AlterZhops".  There are kick-out messages when "Adaptive Layers" is used or when Z-hops simply aren't enabled in the Cura settings.  They come up too late, but they are there.
Search and Replace is a short (65 lines of code) alteration of the existing post processor.  Even I would have trouble screwing that up.

I'll be going back through this and seeing what I can do given your comments (and what I'm actually capable of).  I'll post an update when I get done and can figure out how to do that on Github without adding 17000 commits to the request.

Thanks again,
Greg

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
```diff
@@ -0,0 +1,803 @@
...
+            T0_fan = " P" + str((extrud[0].getProperty("machine_extruder_cooling_fan_number", "value")))  
+        if is_multi_fan:
+            if extruder_count > 1: T1_fan = " P" + str((extrud[1].getProperty("machine_extruder_cooling_fan_number", "value")))
+            if extruder_count > 2: T2_fan = " P" + str((extrud[2].getProperty("machine_extruder_cooling_fan_number", "value")))
+            if extruder_count > 3: T3_fan = " P" + str((extrud[3].getProperty("machine_extruder_cooling_fan_number", "value")))
+        Logger.log("d", "Add Cooling Profile - T0 fan; " + str(T0_fan) + " ; T1_fan; " + str(T1_fan) + " ; T2 fan; " + str(T2_fan) + " ; T3 fan; " + str(T3_fan))
+#Initialize the fan_list with defaults
+        fan_list = []
+        for q in range(0,8,1):
+            fan_list.append("999999999")
```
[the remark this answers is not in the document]
GregValiant: The "999999999" is just to insure the number is beyond what the actual top layer number might be.    The firmware won't see it.  It could also be "str(Len[data]+1)" which is 15 characters as opposed to 9.  Functionally, there really isn't any difference.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
GregValiant: I was thinking about older versions as well as alternate versions.  Right now it appears to be both backward and sideways compatible.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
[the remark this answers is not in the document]
rburema: I don't really mind much that it's limited. Just thinking about ways to communicate that to the user.

Sure, it's long now, and replacing _one_ manual iteration with a for-loop won't indeed matter much. Doing so consistently however, will make it a lot shorter.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
rburema: Hm, well it doesn't cost us much to keep the string-concatenation like that either way.

plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
rburema: Hi Greg!

Thanks for your explanation.

My grandfather (mom's dad) and my own dad worked with punch-cards as well, back in the day. (My dad, learned FORTRAN on them when he started at IBM I think). I think we used to have a few at home.

Don't worry about being 'un-Pythonic' _specifically_. I shouldn't generalize, but while there often is _some_ reasoning behind such statements, people using those kinds of monikers too easily are often language-snobs :-)

Which brings me to my first point; Yes, I think these scripts are functional in the sense that they do what is being proposed, and my main issue with them is indeed the methodology.

For us, however, that's not just an aesthetic thing, but connected to maintenance and readability. (Ironically enough, for most modern coders, a version adjusted to my comments would be a lot more readable -- which is indeed probably different for those just starting out with modern languages.)

Given your comments, I think I'm going to have to do most of that rewrite myself. We normally don't do that for rewrites of this size. However, I'll see if I have some leftover time somewhen that I can spend on doing just that. Can't promise anything however! Specifically, it might not happen within the next week (or even two) though. -- Alternatively, since these post-processing scripts are properly encapsulated, I could ask the rest if we can maybe make an exception to certain things this time -- I suspect the answer will be 'no' however.

Now that that's out of the way some more specific things:

> Congratulations on figuring out which version was the latest one. Every time I try to make an adjustment to the pull request here on Github I've ended up making a mess.

I think you can adjust the branch after you made the PR. At least I can for my own PR's.

> Another problem was that I tried to get this into 5.3.1 and so it was a rush job. That isn't a problem now.

For future reference, we almost never add features to patch releases like going from 5.3.0 to 5.3.1 -- Unless you meant 5.3 in general. Even then we can't make promises. (Our track record is not exactly a shining example of punctuality...)

> I'll be going back through this and seeing what I can do given your comments (and what I'm actually capable of). I'll post an update when I get done and can figure out how to do that on Github without adding 17000 commits to the request.

Git has a feature called squash. Some repositories require you to use that before your code is merged, but we're never too worried about that.

In any case, I take it we're not that much of a hurry right now, since it seems 5.4 will take a while as well.

Again, as always, thanks for always helping the community out, one way or the other.

cheers,
Remco.

PR #14520, opened 2023-02-09
disable interface extruder if no interface is enabled

rijkvanmanen:
CURA-10254


casperlamboo (approved):

casperlamboo:
Thanks @rijkvanmanen 👍

PR #14570, opened 2023-02-15
[CURA-10290] workaround for conflicting packages on 'linux-modern'

rburema:
Packages conflict. The same commit worked fine the previous day, so it's likely not our fault. We should probably inform maintainers, but until then, try to work around the problem without a blanket ignore of all 'overwrite' type errors. The only clear thing is that this all has to do with 'odbc', which is database related. We use that to cache certain things on start-up, so likely we actually need it.

### DIFF
--- a/.github/workflows/cura-installer.yml
+++ b/.github/workflows/cura-installer.yml
@@ -122,6 +122,10 @@ jobs:
                 if:  ${{ runner.os == 'Macos' }}
                 run: brew install autoconf automake ninja create-dmg # Delete create-dmg when deprecating dmg
 
+            -   name: Hack needed specifically for ubuntu-22.04 from mid-Feb 2023 onwards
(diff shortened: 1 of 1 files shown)


jellespijker:
LGTM;

please fill in a bug report at their repo so the can fix it. Reference this PR and port these changes to the other workflows as well.

 https://github.com/actions/runner-images

rburema:
@jellespijker I'm not sure any other workflows need this. None of the other workflows on that OS pull in these dependencies, as far as I can sus out. If you mean the printer-linter error, I made a proposed fix for that here: https://github.com/Ultimaker/Cura/pull/14581

I'll have to make the minimum reproduction steps for the report on their repo. I'll go do that now.

rburema:
Oops, no, disregard the 'no other' part of the message above... there are two others that I see it potentially needing.

PR #14581, opened 2023-02-16
Prevent execution of spurious clean-up step, which then crashes.

rburema:
The clean-up step for setting up Python and pip assumes that you do at least something with those programs -- and so that cache folders have been made, that need to be cleaned up. If nothing happens, the clean-up step then crashes on the non-existent cache folder.


jellespijker (approved):
LGTM 

thanks for picking this up

PR #14591, opened 2023-02-17
[CURA-10310] Less jitters in support: Walls make a lot of difference.

rburema:
__NOTE__: Frontend 'support' (haha) for [this backend ticket](https://github.com/Ultimaker/CuraEngine/pull/1836).

---

Part of wholly missing support was caused by not having walls. You can see that if brim is enabled, brim will have been generated, meaning the areas are there, but support just stops, because the infill pattern has shifted.

Compensate extra material used by not connecting the zigs any more. That shouldn't be necessary, since, hey, walls.

---

_EDIT: Converted to draft for safety w.r.t. multiple possible solutions, but otherwise 'finished'._

### DIFF
--- a/resources/definitions/ultimaker.def.json
+++ b/resources/definitions/ultimaker.def.json
@@ -120,7 +120,7 @@
         "support_line_distance": { "minimum_value_warning": "0 if support_structure == 'tree' else support_line_width" },
         "support_tower_maximum_supported_diameter": { "value": "support_tower_diameter" },
         "support_tower_roof_angle": { "value": "0 if support_interface_enable else 65" },
-        "support_wall_count": { "value": "1 if support_structure == 'tree' else 0" },
+        "support_wall_count": { "value": "1" },
         "support_xy_distance_overhang": { "value": "0.2" },
         "support_z_distance": { "value": "0" },
         "top_layers": { "value": "math.ceil(round(top_thickness / resolveOrValue('layer_height'), 4))" },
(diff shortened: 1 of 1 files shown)


resources/definitions/ultimaker.def.json
```diff
@@ -120,7 +120,7 @@
         "support_line_distance": { "minimum_value_warning": "0 if support_structure == 'tree' else support_line_width" },
         "support_tower_maximum_supported_diameter": { "value": "support_tower_diameter" },
         "support_tower_roof_angle": { "value": "0 if support_interface_enable else 65" },
-        "support_wall_count": { "value": "1 if support_structure == 'tree' else 0" },
+        "support_wall_count": { "value": "1" },
```
Joeydelarago, line 123: I found that changing this setting definitely has a big effect on the jitters. I'm a bit concerned this will make it difficult to see if your engine changes have improved the support. 

At this point though, if it works it works.

rburema: (You probably already know, just to be sure; this is meant to go with my backend PR!)

It's you can still test with walls off if you like ;-)

I think the engine changes will be _harder_ to see without walls actually! The piece of code we're changing is about the _areas_, not the subsequent infill generation. It really makes it hard to see if the error is yours or if it's just the walls again when its just outright not printing infill in an area for all the layers to the bottom from a certain point onwards.

rburema:
Closed. Walls wall make support harder to remove. Also, these changes where mostly to prevent regressions, and this is not really a regression.

(And also because they where made to work in conjunction with the engine PR, which will also be closed.)

A separate ticket will be made to try to prevent the 'infill not generated in areas that _miss_ the infill walls' in the general case.

PR #14594, opened 2023-02-17
Fix VzBot base

ckvsoft:
The correct value should be -03 not 0.3
Should actually still in the 5.3. Maybe it works out.


ckvsoft:
@Joeydelarago 
Do you think the fix will still make it into 5.3?
Thanks

PR #14778, opened 2023-03-05
Update InsertAtLayerChange.py

tastyratz:
# Description

I removed the period from the G-code to insert label for format consistency. 

I also added a third option to this script called "Skip layers". This allows you to insert gcode but with a specific number of layers skipped (i.e. skip layers 1 will insert gcode every other layer, skip layers 3 will insert every fourth).

## Type of change

- [x] Formatting

# How Has This Been Tested?

As an example, this change allowed me to insert a nozzle cleaning routine in my gcode without having to run it EVERY layer.

I viewed the generated gcode later to make sure it did, in fact, insert my gcode where I wanted and skip specified layers as intended.

**Test Configuration**:
* Operating System: Windows 10 22h2, Cura 5.2.1

# Checklist:


plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
```diff
@@ -26,27 +26,40 @@ def getSettingDataString(self):
...
+                    if count == 0:
+                        if self.getSettingValueByKey("insert_location") == "before":
+                            layer = gcode_to_add + layer
+                        else:
+                            layer = layer + gcode_to_add
+
+                        data[index] = layer
 
-                    data[index] = layer
+                    count = (count + 1) % (skip_layers + 1)
```
wawanbreton, line 63: As this counter is calculated for every iteration, there is no actual need to store it. You can put this formula directly in the `if` (without the `+1` of course) which makes the code a bit simpler.
I expected the `count` variable to be decremented at each layer, maybe this is what you did in the first place ?

wawanbreton:
Hi @tastyratz, thank you for this improvement ! I think the added "skip layers" setting can be useful indeed.
I made a simple remark on your code, which is just to simplify it a bit. Feel free to decline it if you have strong reasons to keep it as is.

tastyratz:
> Hi @tastyratz, thank you for this improvement ! I think the added "skip layers" setting can be useful indeed. I made a simple remark on your code, which is just to simplify it a bit. Feel free to decline it if you have strong reasons to keep it as is.

Thanks! 
I really have no marriage to the method here driving me to a specific way to do it. 
Truthfully, I had some help and am not great with Python. If your way works just as well and tests out OK then I have no objections.

wawanbreton:
Ok, just forget what I said :laughing: 
I wanted to suggest the code change but when writing it I figured out that I had misread it. The `count` variable is indeed incremeted by assigning it again to itself. Sorry for that.
I can then accept this PR.

wawanbreton (approved):

PR #14792, opened 2023-03-06
CURA-9826_add-skirt-height

rijkvanmanen:
See backend: https://github.com/Ultimaker/CuraEngine/pull/1856


jellespijker (changes requested):
The new setting should also be added to the settings_visibility.
I would suggest `expert` mode

jellespijker (approved):

PR #14797, opened 2023-03-06
[5.3] Fix VzBot Base definition file

ckvsoft:
# Description

The correct value should be -03 not 0.3 Should actually still in the 5.3. Maybe it works out.

## Type of change
Fix


nallath (approved):

nallath:
You're just in the nick of time!

ckvsoft:
> 

Thanks, had to re-fork first because I didn't have the 5.3 with me.
I only noticed yesterday that the previous patch 1 was probably the wrong way.
I created a second pull request for main. Hope that fits.

PR #14798, opened 2023-03-06
Update vzbot_base.def.json

ckvsoft:
# Description
This fixes false xy_offset
The correct value should be -03 not 0.3 Should actually still in the 5.3. Maybe it works out.

## Type of change
- [x ] Bug fix (non-breaking change which fixes an issue)
- [ x] Printer definition file(s)


jellespijker:
Currently this PR doesn't change anything was this intended or can this be closed?

jellespijker:
Will close this PR, since there is no actual change to speak of. Feel free to reopen this PR if you think this is in error

PR #14836, opened 2023-03-09
Update cura.po

discip:
language update


rburema:
Hi! Sorry to be so late for this. I see it's closed, but have updated the obviously wrong ones (the bottom ones, that are even wrong according to google translate) in the files we've sent to the translators for 5.5

Thanks!

discip:
@rburema
No problem,
was not sure if that still matched, what it was supposed to, so I closed it due to the fact there have been so much updates in the meantime. 😅

PR #13110, opened 2022-08-25
Add  some WEEDO  printers

WinstonMao:

PR #14926, opened 2023-03-17
Add ENTINA printer definitions

cgobat:
This PR adds printer defintions for the [ENTINA TINA2](https://www.entina3d.com/products/entina-mini-3d-printers-tina-2-basic) and [ENTINA X40](http://wiki.entina3d.com/doku.php/x40/index) 3D printers and their clones.

**Scope:**

- [ ] Translations

# How Has This Been Tested?

All definition files load and behave as expected in local Cura installation.

**Test Configuration**:
* This PR fixes the test failures of #14900, which was originally tested on an older Cura version

# Checklist:


cgobat:
Could I have some clarification about these test failures? It is throwing an `AssertionError: Unnecessary default_values for [...]`, for fields that I *do* want to actually define default values for. Why is this a problem/not allowed?

jellespijker:
Please take a look at the failing unit tests: https://github.com/Ultimaker/Cura/actions/runs/4487096515/jobs/7890198991?pr=14926#step:13:446

jellespijker:
because the definition from which you inherit, already sets a `default_value`,  easy solution is to change then into `value`.

cgobat:
> easy solution is to change then into `value`.

You mean just change all of those `"default_value"` keywords to `"value"`?

nallath:
Hey @cgobat Could you still change those default_values to value? Thanks!

cgobat:
Related: #13110, #15095, #16045

jellespijker:
@wawanbreton and/or @saumyaj3 can either of you take a look at this PR?

resources/definitions/entina_x40.def.json
```diff
@@ -0,0 +1,56 @@
...
+  "name": "ENTINA X40",
+  "inherits": "weedo_base",
+  "metadata":
+  {
+    "visible": true,
+    "author": "ENTINA",
+    "manufacturer": "ENTINA",
+    "file_formats": "text/x-gcode",
+    "machine_extruder_trains": {
+      "0": "weedo_extruder_left_x40",
```
saumyaj3, line 12: These extruders are not available in the database.. Please add them too

resources/definitions/entina_x40.def.json
```diff
@@ -0,0 +1,56 @@
...
+      "1": "weedo_extruder_right_x40"
```
saumyaj3, line 13: this one too

resources/definitions/weedo_f192.def.json
```diff
@@ -0,0 +1,37 @@
...
+    "inherits": "weedo_base",
+    "metadata": {
+        "visible": true,
+        "author": "WEEDO",
+        "manufacturer": "WEEDO",
+        "file_formats": "application/x3g",
+        "machine_extruder_trains":
+        {
+            "1": "weedo_extruder_left",
+            "0": "weedo_extruder_right"
```
saumyaj3, line 13: These extruders do not exist . please add them

saumyaj3 (changes requested):
I see the extruder used in the printers are not available in the PR. Please create them and add it to this PR?

resources/definitions/entina_x40.def.json
cgobat: Thanks, good catch. Fixed now.

Should there be a unit test/check to catch things like undefined extruders and prevent them from slipping through?

PR #15002, opened 2023-03-23
 modified some parameters for Flsun V400

dengdaitianhei:
I am a staff member of flsun. This time I modify some default printing parameters of flsun V400 in order to improve the printing quality.
I've used these configurations with version 5.3 of Cura under Windows with no problems and improved print quality. Hope to get approval soon. Thanks in advance


saumyaj3:
@dengdaitianhei I also noticed, these changes make print a bit faster and lighter. Is that intentional?

PR #15057, opened 2023-03-27
Cura 10317 other inconsistent material profiles in internal builds

jellespijker:
# Description

Not sure how the formatting of the `ln3d` printers in this branch are different from `main` but I think it safe to keep this in


rburema (approved):
lgtm

PR #15074, opened 2023-03-29
Suggested UI changes

ThomasRahm:
I noticed in the UI branch that quite a few changes were not migrated. If any changes were that intentional that this pull request would undo please tell me.
 

### My reasoning for the changes of `fdmprinter.def.json`:

- Stair Step settings do not work with tree support => Do not display when tree support is used  
`Tree Support Branch Distance` is not used by tree support, instead the new easier to understand `Tree Support Branch Density` is used => Removed `Tree Support Branch Distance`  
- `Tree Support Collision Resolution` is no longer used, as the distance for avoidances now grows exponentially instead of linearly. => Removed `Tree Support Collision Resolution`  
- The extra z distance for tree support was by my understanding an UI fix for an engine issue. The new implementation does not have the same problem => Removed the extra z distance when tree supports are used.  
- `Support Horizontal Expansion` increases the overhang area when tree supports are used => Display when tree support is used  
- `Minimum Support Area` can be used to prevent tree support from supporting small areas. Having this at >0 without knowing can cause someone to think that tree supports are bugged => Display when tree support is enabled. Having them apply but not visible will cause bug reports.  
- `Minimum Support Floor Area` does not work with tree supports => Do not Display when tree support is used  
-  `Support Wall Line Count` should be at least one for tree supports, otherwise the output may not be functional => Adjusted minimum value warning

Profiles were changed to also apply the changes above when necessary.

I slightly changed the description for `Tree Support Maximum Branch Angle` and fixed a type `Inital` => `Initial`


ThomasRahm:
I would apply the same behavior `Support Wall Line Count` (should be 1 when tree supports are used) also to `Support Roof Wall Line Count`. Do you agree that this would be beneficial ?

rburema (commented):
Looks (mostly) good to me! Sorry this took a while, I initially wanted to discuss this with the Print Processing & Materials team, but it seems these are mostly procedural rather than related to materials (and only tangentially to any print-process). It looks like I can review them on my own just fine.

W.r.t. the Support Horizontal Expansion (`support_offset`) I had to solve this in another way (`"value": "support_line_width + 0.4 if support_structure == 'normal' else 0.0",`), because if the model has a downward spike that needs to be supported (and no support interface is selected), the tip of the tree can become large enough that the model-spike would fall through it.

So I think I'll just check out this branch locally, change that, and then merge it 'manually' instead of via the github interface.

(I also merged the suggestions of our printer-definition-linter, but I could do that via the github interface already.)

rburema (approved):
The issue mentioned in my last comment was solved by the merge-conflict resolution, so I'm ok to merge anyway.

PR #15525, opened 2023-05-15
Cleanup skin overlap settings

pkuiper-ultimaker:
Part of profile cleanup actions.
Set the skin overlap to 20% for all S line machines and all materials.
Currently a lot of different values are used, but tests show this has no real benefit. 
A default value of 20% is good enough.


jellespijker (approved):
LGTM

PR #15590, opened 2023-05-23
Added Kobra Plus printer profile

Jordonbc:
# Description

This pull request adds a profile for the Anycubic Kobra Plus. I used the Anycubic Kobra Max profile as a base and modified the area along with some specific tweaks I found to give me better prints.

The start gcode also Includes a nozzle wipe and turns on the LED and when finished also turns it off. The nozzle wipe is based on the Anycubic Mega S profile but more or less flipped vertically as it got caught on the hotbed clips.

The profile also does NOT include the M420 S1 gcode command as I have found out it's enabled by default and homing doesn't turn it off (tested with octoprint terminal).

# How Has This Been Tested?

I tested this profile with various prints and managed to print 2 benchys at 0.2 and 0.1 along with a Groot model and a bunch of cosplay models all successfully.

Note: This has only been tested on 1 Anycubic Kobra Plus.

# Checklist:

### DIFF
--- a/resources/definitions/anycubic_kobra_plus.def.json
+++ b/resources/definitions/anycubic_kobra_plus.def.json
@@ -1,0 +1,30 @@
+{
+    "version": 2,
+    "name": "Anycubic Kobra Plus",
+    "inherits": "fdmprinter",
+    "metadata":
+    {
+        "visible": true,
+        "author": "Jordon Brooks",
+        "manufacturer": "Anycubic",
+        "file_formats": "text/x-gcode",
+        "has_machine_quality": true,
+        "has_materials": true,
+        "machine_extruder_trains": { "0": "anycubic_kobra_plus_extruder_0" },
+        "preferred_material": "generic_pla",
+        "preferred_quality_type": "pla",
+        "quality_definition": "anycubic_kobra_plus"
+    },
+    "overrides":
+    {
+        "machine_depth": { "default_value": 302 },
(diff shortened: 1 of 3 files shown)


resources/definitions/anycubic_kobra_plus.def.json
```diff
@@ -0,0 +1,30 @@
...
+    {
+        "visible": true,
+        "author": "Jordon Brooks",
+        "manufacturer": "Anycubic",
+        "file_formats": "text/x-gcode",
+        "has_machine_quality": true,
+        "has_materials": true,
+        "machine_extruder_trains": { "0": "anycubic_kobra_plus_extruder_0" },
+        "preferred_material": "generic_pla",
+        "preferred_quality_type": "pla",
```
jellespijker, line 15: `pla` isn't the best name as a quality type, `standard`, `draft`, `normal` etc. are more inline with best practices.

jellespijker (approved):
See review comment about the naming of the preferred quality type as `pla`. I will merge this and rename this to `normal` which is also used by some other anycubic printers. If you feel like it should have another name instead (draft, default, etc..) feel free to open a new PR mentioning this PR.

When investigating common quality names I came across multiple other definitions which have `pla` so I can understand how this came to be. We discussed this internally and we will add a new check in the printer-linter for best-practices regarding naming of these qualities. 

Thank you for your contribution btw, the printer should be available in Cura 5.5

PR #15632, opened 2023-05-30
PP-81 easy to tear of brim

Frederic98:
# Description
Improve brim removal by setting the `brim_gap` to the initial layer horizontal expansion + 50μm - See PP-81


jellespijker:
@Frederic98 is there a reason why this isn't set in `ultimaker.def.json`?

Frederic98:
We've only been testing on S-line for now, when we're done with other tickets we can test it on our other printers to make sure it works well there too (especially on the UM2(+/connect) without automatic leveling)

jellespijker:
Be sure to link the follow-up ticket, I will merge this for now

PR #15633, opened 2023-05-30
Added material_is_support_material setting 

pkuiper-ultimaker:
The material_is_support_material will be True for support materials profiles (see fdm_material repo).
Added a resolve function to select the active extruder that contains a support material as the default support extruder.

See PP-285


jellespijker (approved):
LGTM

PR #15702, opened 2023-06-09
Lower max fanspeed

pkuiper-ultimaker:
Reduce max. fan speed to prevent low strength for small/thin parts.

**UI bug**
For some reason the UI is not updating correctly when changing the cool_fan_speed. cool_fan_speed_max displays the same value until you open and close the cooling section. Then it refreshes correctly.

PP-351


nallath:
Closing this as i've already cherry-picked the relevant commits into 5.4 (and pulled those into main)

PR #15748, opened 2023-06-14
Fixed spanish translation for lightning infill

Pelochus:
# Description

Lightning infill was translated as "Iluminación" which means **"lighting"** (not to be confused with **lightning**). **Lightning actually means "Rayo" or "Rayos"** (spanish plural for lightning). 

## Fix

**The translation has been fixed as "Rayos"** since it better matches to the infill type, since it has more than one lightning. 

You can read this [link](https://www.wordreference.com/es/translation.asp?tranword=lightning) to verify the translation is correct 

## Type of change

- [x] Translations


Pelochus:
Not a bad suggestion, but I think it feels weird to have that translation. That literally translates to "Lightning shaped infill" which sounds redundant. It doesn't make a lot of sense having translations like "Cubic" -> "Cúbico", "Grid" -> "Rejilla" but then "Lightning" -> "Relleno en forma de rayo". 

If we follow this logic then something like "Cubic subdivision" should be translated to "Relleno en forma de cubo con subdivisiones". However, it is not a bad translation, just sounds and reads weirdly.

Still, if the people at Ultimaker insist I've allowed maintaners to edit this commit if they prefer that translation.

PR #15692, opened 2023-06-07
Fix exclude materials

casperlamboo:
CURA-10599

PR #15840, opened 2023-06-21
Remove exclude materials from 3rd party printers

casperlamboo:
With the changes from https://github.com/Ultimaker/Cura/pull/15692 logic regarding the exclude materials printer setting changed. It has never worked properly before. Now that it works properly again a ton of materials were excluded that were never intended to be excluded.

As we can never be sure if an exclude material was intended or not we decided to be pretty agressive in removing exclude materials. This way the behavior is closed to the behavior prior to the exclude material fix.

There was one case were we decide to leave an exclude material in. If a printer has a material diameter of `175` *and* the material would specifically specify the `175` substring (e.g. `dsm_arnitel2045_175`). When this is the case we were pretty certain it was intended to exclude this material.

CURA-10599


jellespijker (approved):
LGTM

PR #11846, opened 2022-04-21
Update DisplayProgressOnLCD.py

inigoml:
Add support for Layer advance (new "Layer Left" BTREE functionality) when printing from Octoprint.

Issue #6019, opened 2019-07-16
Progress percentage on the software do not match the display on the CR-10S


CGLee65:
The Completed % on the software shows like 4% and the Display on the control box says 0%

rburema:
There's some known (and unkown) issues with progress reporting / time estimation / etc. One thing to keep in mind is that it's an estimation, especially for printers we can't often test with.

Does this discrepancy become wildly out of sync or does it stay within certain limits?

CGLee65:
I figured out that the percentage does not show on the control box when printing via USB cable, but it does show on the monitor screen of cura.  The percentage does show on the box when you print from the TF Card and does not show anything on the monitor page.

Sent from Mail for Windows 10

Ghostkeeper:
The printer doesn't know the progress when printing via USB cable. The software is sending the g-code commands on the go.

There is no way for Cura to communicate progress to a CR-10S via USB cable. The only g-code that Cura could use for that is M532 but that's only supported for Repetier, according to the [RepRap wiki](https://reprap.org/wiki/G-code#M532:_Set_print_progress).

yet-another-average-joe:
Marlin has M73 that displays a percentage on the LCD :

http://marlinfw.org/docs/gcode/M073.html

(it is used by a couple of OctoPrint plugins)

fieldOfView:
M73 support needs to be specifically enabled in Marlin, so it will not work on all printers.

This feels like something that maybe should be implemented in a postprocessing plugin, calculating progress from the ";TIME_ELAPSED" comments that CuraEngine puts into the gcode after each layer.

yet-another-average-joe:
I was just thinking of the Cura Post Processing Plugin, and realized that it could be based on layer # only, and it does not make sense !

I thought that LCD_SET_PROGRESS_MANUALLY was ON by default. I was wrong...

LinFor:
I wrote simple post-processing script to do thins on Marlin: https://github.com/LinFor/Marlin/blob/e-bot-skr/cura-profiles/scripts/EnrichPrintProgress.py
It's just use TIME_ELAPSED layer estimation from Cura and converts it into M73/M117 commands. Maybe later I approximate progress more precisely within layer (by E-movements, for example) to do more reliable progress estimation for short prints or very big layers.

[...]

PR #15866, opened 2023-06-22
Display Info on LCD

GregValiant:
# Description
DisplayInfoOnLCD combines DisplayFilenameAndLayerOnLCD with DisplayProgress.  Those two post processors become options within the single post processor and the two previous files become obsolete.  There is an option to add an M118 line as well as M117 so the info would get bounced back to a print server.
There is another pull request #11846 that involves "Display Progress on LCD" and BTREE firmware.
The Aug 12 commit adds "Countdown to Pauses" option when "Display Progress" is selected.
The Sep 8 commit adds an "Estimated Print Finish Time" message as an option per request #16677.  #6019 
Oct 20 commit adds a comment below the TIME: line and converts "seconds" to Housr:Minutes:Seconds with a pause and/or filament change count.
;TIME:14071
;Filament used: 13.3238m
;Layer height: 0.2
;Cura Time Estimate:  14071sec = 3hr 54min 31sec  with 1 pause(s) and 1 filament change(s)

## Type of change

- [ X] Updated feature

# How Has This Been Tested?

- [ X] Test A - Cura 4.x and 5.x

**Test Configuration**:
* Operating System:  Win10 Home and Marlin

# Checklist:

- [X ] I have read the [Contribution guide](https://github.com/Ultimaker/Cura/blob/main/contributing.md) 
- [X ] I have commented my code, particularly in hard-to-understand areas


GregValiant:
I was researching something else and came across a fact I had completely missed - RepRap can use M566 for Jerk in mm/min.
This PR adds support for M566.

plugins/PostProcessingPlugin/scripts/DisplayInfoOnLCD.py
```diff
@@ -0,0 +1,401 @@
...
+                display_text = lcd_text + str(i)
+                layer_index = data.index(layer)
+                lines = layer.split("\n")
+                for line in lines:
+                    if line.startswith(";LAYER_COUNT:"):
+                        max_layer = line
+                        max_layer = max_layer.split(":")[1]
+                        if self.getSettingValueByKey("startNum") == 0:
+                            max_layer = str(int(max_layer) - 1)
+                    if ";LAYER:" in line:
```
rburema, line 182:
```suggestion
                    if line.startswith(";LAYER:"):
```

plugins/PostProcessingPlugin/scripts/DisplayInfoOnLCD.py
```diff
@@ -0,0 +1,401 @@
...
+        # initialize global variables
+            first_layer_index = 0
+            time_total = 0
+            number_of_layers = 0
+            time_elapsed = 0
+        # if at least one of the settings is disabled, there is enough room on the display to display "layer"
+            first_section = data[0]
+            lines = first_section.split("\n")
+            for line in lines:
+                if ";TIME:" in line:
```
rburema, line 219:
```suggestion
                if line.startswith(";TIME:"):
```

rburema (changes requested):
Hi Greg!

The big problem I have with this is that when you save a project, it saves which (post-processing) scripts where used as well. If it can't find them, it just ignores them -- which means that if the (now) removed scripts are added to a project, they'll no longer work, nor do they use the new script.

There's a number of ways to handle this.
- Writing an upgrade-script would be the most proper, but also the hardest (and I'm not sure if we can do that for upgrade scripts -- though it should be possible). Maybe a bit overkill?
- Leaving the scripts in, but messaging the user that the older scripts are deprecated.
- Only leaving shells of the scripts, and _only_ leave functionality in there to message the user.
- Rewriting the scripts so that they inherit from the new one.
- Not do anything (so reinstate them) and just have them live side-by-side until the next major version of Cura.
- ... or I can ask the rest of the team if it's ok that we just rip these out and just communicate the change well.

plugins/PostProcessingPlugin/scripts/DisplayInfoOnLCD.py
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/DisplayInfoOnLCD.py
GregValiant: Done.

GregValiant:
The project file situation is something I hadn't thought of.  
The new script carries over "Display Filename and Layer on LCD" complete and I don't recall making changes to that script.
The second half of "Display Progress on LCD" is somewhat different with the time fudge factor, and Finish Time message, etc.  

As a user I think I would prefer a quick message to inform me to move to the new post processor.  If the script is left in then there is no incentive to move on and people will keep using it.

I've made changes to the two original scripts and turned them into "dummies" that just give messages.  What is your preference in handling this?

In addition, I made some additions and need to make another commit.  It will have the two changes you requested.

rburema (approved):
Hi Greg,
Sorry for the wait (again... it's been quite busy here).

I think we can accept this now. I _have_ noticed that this doesn't 100% replace the other two scripts (the possibility to set M73 instead of M117/8 is missing, and the update-frequency can only be once per layer), but I think this adds enough (and the consolidation itself is also a good thing), that it's still a good improvement. -- So let's just wait to see if anyone complains about those little thing.

Thanks again for the contribution, and I'll try to take another look at the other PRs again next week. (That big list was/is useful.)

lethuer:
I've observed some bugs in the new plugin: [DisplayInfoOnLCD.py](https://github.com/Ultimaker/Cura/blob/main/plugins/PostProcessingPlugin/scripts/DisplayInfoOnLCD.py)
and noted my obervations here: https://github.com/Ultimaker/Cura/issues/18766

PR #16033, opened 2023-07-06
SpinBox fixed for values above the max value

saumyaj3:
CURA-10096


jellespijker:
@saumyaj3 yes your first bug fix. Don't worry about those merge conflict I will resolve these. They happened due to an automatic workflow

jellespijker (changes requested):

resources/qml/SpinBox.qml
```diff
@@ -97,6 +110,13 @@ Item
                     base.editingFinished();
                 }
             }
+
+            onTextChanged:
+            {
+                var value = spinBox.valueFromText(text);
+                text = spinBox.textFromValue(value);
+                spinBox.value = value;
```
jellespijker, line 118: Otherwise it will only listen to values changed with the `+` and `-` buttons
```suggestion
                base.value = value;
```

jellespijker (changes requested):
You need to set the `value` property of `base` since the `value` of the `spinBox` is calculated from that `value: Math.floor(base.value / base.stepSize)` see L53

PR #16097, opened 2023-07-10
Enable support interface

pkuiper-ultimaker:
This improves the self support interface quality.
These settings are for the AA0.25 and AA0.4 nozzles for all materials. Default interface distance is 0.3mm (rounded to layer heights) and 0.4mm for CPE+

Relates to: PP-10


jellespijker (approved):
LGTM

PR #16110, opened 2023-07-11
Frontend plugin orchestration

nallath:
First rough draft of how the frontend will handle the backend_plugins. 

Originally i added the concept of backend_plugins to uranium, but the extra abstraction made it clumsy quite fast. So i moved all of it over to Cura. The accompanying Uranium PR now essentially only contains boyscouting changes. Since they are still good to do, i've left them in.

https://github.com/Ultimaker/Uranium/pull/886

CURA-10717

### DIFF
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -325,6 +325,8 @@ class CuraEngineBackend(QObject, Backend):
 
         self._createSnapshot()
 
+        self.startPlugins()
+
         Logger.log("i", "Starting to slice...")
         self._time_start_process = time()
         if not self._build_plates_to_be_sliced:
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -70,7 +70,7 @@ class CuraEngineBackend(QObject, Backend):
             os.path.join(CuraApplication.getInstallPrefix(), "bin"),
             os.path.dirname(os.path.abspath(sys.executable)),
         ]
-
+        self._last_backend_plugin_port = self._port + 1000
         for path in search_path:
             engine_path = os.path.join(path, executable_name)
             if os.path.isfile(engine_path):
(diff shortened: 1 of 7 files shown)


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -172,10 +190,25 @@ def __init__(self) -> None:
...
+        It assigns unique ports to each plugin to avoid conflicts.
+        :return:
+        """
+        backend_plugins = CuraApplication.getInstance().getBackendPlugins()
+        for backend_plugin in backend_plugins:
+            if backend_plugin.isRunning():
+                continue
+            # Set the port to prevent plugins from using the same one.
+            backend_plugin.setPort(self._last_backend_plugin_port)
+            self._last_backend_plugin_port += 1
```
jellespijker, line 209: Maybe limit the ports and ensure that they are within this range: https://stackoverflow.com/questions/10476987/best-tcp-port-number-range-for-internal-applications

We should also maybe query before hand if a port is used already before assigning it.

jellespijker (approved):
Overall, the implemented code looks good and neat. It focuses on the implementation of the internal BackendPlugin feature within the Cura application. Here are some comments and suggestions:
1. The readability and comments in the code are well-written which makes it easy to understand the functionality of each part of the code.
2. BackendPlugin.py: Creation of BackendPlugin object which takes care of starting, stopping, validating, etc., for the backend plugin process is neatly implemented. Good abstraction and encapsulation are used here.
3. CuraApplication.py: Storing BackendPlugin list in CuraApplication instance and exposing API's to add and get the list of backend plugins is a good design decision.
4. CuraEngineBackend.py: The use of plugins within the CuraEngineBackend.py is neatly done especially how exceptions are handled in startPlugins function is appreciable.
StartSliceJob.py: Adding the support for plugins in the run function using engine_plugins is a solid move towards extending the functionality

nallath:
> The readability and comments in the code are well-written which makes it easy to understand the functionality of each part of the code.

Did you just use chatgpt to review my code? Jeez :P

PR #12730, opened 2022-07-10
MatterHackers Pulse XE printer definition

nacho319:
These are working definitions for Matterhackers Pulse XE printer with E3D print heads;

this definition has been tested and is working on Cura v5 on macOS Monterey with the
printer connection via OctoPrint.

There is no difference between the E3D Lite6 and E3D V6 print head definition at this
time, but the E3D V6 can handle higher temperatures and more materials.

Also, the printer setup code invokes unified bed leveling before the print on EVERY
print.  (I make no assumptions about the mangement of the printer this way.)  The
end print code also plays the finished printing MatterHackers tune on the printer
at the end of the print as well.

PR #16159, opened 2023-07-15
Definition: Pulse XE E444M

randyzwitch:
# Description

This PR submits the bare minimum of settings needed for the Matterhackers Pulse XE E444M. This supersedes #12730, which seems to be abandoned.

## Type of change

# How Has This Been Tested?

- [x] Benchys
- [x] Ninja Turtle action figure
- [x] Misc guitar tools 

**Test Configuration**:
* Operating System: Windows 11 and direct from memory card on machine

# Checklist:

### DIFF
--- a/resources/definitions/matterhackers_pulsexe_e444m.def.json
+++ b/resources/definitions/matterhackers_pulsexe_e444m.def.json
@@ -1,0 +1,62 @@
+{
+    "version": 2,
+    "name": "Pulse XE E-444M",
+    "inherits": "fdmprinter",
+    "metadata":
+    {
+        "visible": true,
+        "author": "Zwitch Guitars",
+        "manufacturer": "MatterHackers",
+        "file_formats": "text/x-gcode",
+        "has_materials": true,
+        "has_variants": false,
+        "has_machine_quality": false,
+        "preferred_material": "generic_pla",
+        "preferred_quality_type": "normal",
+        "first_start_actions": [ "MachineSettingsAction" ],
+        "machine_extruder_trains":
+        {
+            "0": "matterhackers_extruder"
+        }
+    },
+    "overrides":
+    {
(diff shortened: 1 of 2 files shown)


randyzwitch:
The only concern I have about this is that by convention, MatterControl keeps all the machine settings via software. On my machine, I always run the following line of code to change the Z offset:

`M851 Z-3.13`

I didn't want to hard-code that into the profile G-code, since it would be hard to find. At the same time, this won't work unless 1) the user runs this code or 2) the user has set this permanently on their machine.

Is there a canonical way to handle this situation, or a way via documentation I can give users the heads-up about this?

randyzwitch:
Is there anything else I need to do before this can be reviewed?

saumyaj3:
> The only concern I have about this is that by convention, MatterControl keeps all the machine settings via software. On my machine when using Cura, I always run the following line of code to change the Z offset:
> 
> `M851 Z-3.13`
> 
> I didn't want to hard-code that into the profile G-code, since it would be hard to find. At the same time, this profile definition won't work unless 1) the user runs the above line of code or 2) the user has set the offset permanently on their machine.
> 
> Is there a canonical way to handle this situation, or a way via documentation I can give users the heads-up about this?

Hi randyzwitch!

If this behavior is consistent over all the printers, perhaps you would want to include this in the start g-code. Otherwise, other users have to find or do the offset themselves. I hope this helps.

randyzwitch:
Now that I've thought about it, I think it's more reasonable to not have this inside the g-code. At least in my case, not having the value means the head will be too high off the plate, which I think is better than possibly crashing into the plate and people not knowing why

saumyaj3:
Hey Randy! Thanks for your contribution. Your changes are now part of Cura.

randyzwitch:
Awesome, thanks!

PR #16183, opened 2023-07-17
Automatically scroll down in crash handler

nallath:
It was reaaaaaly annoying me that people would post screenshots all the time. Those screenshots would never include the info you need, because it would only capture the top of the exception / logs.


jellespijker (approved):
LGTM

Issue #15731, opened 2023-06-12
AGPLv3 still in comments?


sp7412:
### Application Version

5.3.1

### Operating System

Windows

### Printer

Ultimaker

### Reproduction steps

I see that Cura is released under GNU Lesser General Public License v3.0 (https://github.com/Ultimaker/Cura/blob/main/LICENSE) and that the license changed from AGPLv3 to LGPLv3 in 2017 (https://github.com/Ultimaker/Cura/commit/9a193ad5c5ab5324ad1335dd9adf80fef24a7977).

But I still see AGPLv3 in the header of this file https://github.com/Ultimaker/Cura/blob/main/plugins/PostProcessingPlugin/PostProcessingPlugin.py#L2

Was this file missed when Cura went from GPLv3 to LGPLv3 in 2017?

### Actual results

N/A

### Expected results

N/A

### Checklist of files to include

- [ ] Log file
- [ ] Project file

### Additional information & file uploads

N/A

fieldOfView:
The PostProcessing plugin was originally a separate project, released under the AGPL:
https://github.com/nallath/postprocessingplugin

It looks like the plugin was merged into the Cura project (6c0fb110fe842758c7cc20dae7edc055d5c8ad8c) after the change in license on the Cura repository. So that leaves the Cura project with a mixed license at the moment.

sp7412:
@nallath

MariMakes:
Hey @sp7412,

Thanks for your report 👍
Excellently spotted 🔎 

I'll bring it up to the team to see how they can update it.

fieldOfView:
Technically, I was never asked permission to change the license of my contributions _to the Postprocessing plugin repository_ after october 2016. My contributions before that date could be seen as owned by Ultimaker. If you want my permission, then just ask.

nallath:
Err. You are right. I overlooked that. Mea culpa! 

So, Do you mind changing it to LGPL?

PR #16237, opened 2023-07-21
Fix wrong licenses

MariMakes:
These seem to be wrong by accident. Let's update them to reflect the rest of Cura which falls under the LGPLv3 or higher.
This resolves CURA-10814 and https://github.com/Ultimaker/Cura/issues/15731

### DIFF
--- a/plugins/PostProcessingPlugin/PostProcessingPlugin.py
+++ b/plugins/PostProcessingPlugin/PostProcessingPlugin.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2018 Jaime van Kessel, Ultimaker B.V.
-# The PostProcessingPlugin is released under the terms of the AGPLv3 or higher.
+# The PostProcessingPlugin is released under the terms of the LGPLv3 or higher.
 
 import configparser  # The script lists are stored in metadata as serialised config files.
 import importlib.util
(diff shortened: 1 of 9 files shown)


jellespijker (approved):
LGTM

fieldOfView:
I don;t think the license was left as AGPL by accident, and just changing it now may not be legal.

The Cura license was originally AGPL. It was changed to LGPL on or around aug 1st, 2017 with the explicit consent of all (?) contributors to the codebase up until that point in time. At the time, the Post Processing plugin was still an external project, licensed under the AGPL license. Its license was never explicitly changed, before it was merged into the Cura repository somehwere in 2018. Technically, all prior contributiors to the codebase of the original repository (https://github.com/nallath/postprocessingplugin) would have to agree with the license change.

cc @nallath

jellespijker:
Good point @fieldOfView . There wasn't any mention of the previous AGPL license in the internal Jira ticket that is why I pressed merge. It seemed straight forward enough. I now actually read the original GH issue mentioned in this PR as well and I see that there are indeed something like 20 contributors the original repository.

I will revert the commit for now and rediscuss this with the team next Tuesday when @nallath is present as well.

jellespijker:
Reverted with: 56b8205ce04c30ff954dba03ab453e64499d9a85

PR #16288, opened 2023-07-26
Create OpenPostProcessorReadMe.pdf

GregValiant:
A Word Document, a PDF file (from the Word Doc), and a PostProcessor.

- The "**PostProcessorReadMe.docx**" file is the editable file.  It contains instructions for the current Post Processors and for my proposed new post processors.  This would not be included in the installation.  It would need to be maintained somewhere by someone who would use it to update the PDF when needed.
- The **"PostProcessorReadMe.PDF"** file is the non-editable version to go in the "scripts" folder.
- **"AAA_PostProcessReadMe.py"** is the Post Processor.  When the user adds it to the list it immediately opens the PDF file.  This works in Windows and there are code lines for MAC and Linux but I'm unable to test that those actually work.

# Description

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* Operating System:  Win10 Home and Cura 5.4.0

# Checklist:

- [X ] I have uploaded any files required to test this change


GregValiant:
Instructions (or at least descriptions) of the Post Processors both current and proposed.
Adding the Post Processor to the list in Cura will immediately open the PDF in the default system viewer.  MAC and LINUX code lines need testing.

        SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
        if Platform.isWindows:        
            text_file = os.startfile(SCRIPT_DIR + "\Post Processor ReadMe.pdf")
        elif Platform.isOSX():
            text_file = open(SCRIPT_DIR + "\Post Processor ReadMe.pdf")
        elif Platform.isLinux:
            text_file = os.system(SCRIPT_DIR + "\Post Processor ReadMe.pdf")

fieldOfView:
A documentation popup is not a postprocessor. How are users supposed to discover this? If anything, the documentation should open via a link in the interface, not by adding a postporcessing script. However in that case it would be more like other places in Cura to open a link to a support webpage.

I do think the Post Processing plugin is due a big UX update, with more documentation per post-processing script which shows what scripts are useful for what printer etc. But it seems post processing may change a lot soon anyway, with postprocessing happening in CuraEngine instead of in Cura.

GregValiant:
Thanks Aldo.  My thought was that it was at least right there at the top of the list and would provide some information to a user regarding the overall function of a particular post processor as well as a closer view of the settings involved.  No one would know it was there until they load a post processor (it could be named "Help").  If it was accessible from the "Help Menu" it would be easier to find.  It would still need some maintenance though.  It is really nothing more than a collection of ToolTips but would provide an overall detailed description of each setting, as well as an overall view of function and any pitfalls to watch for.
If it was accessible from a "Help" button on the Post Process Plugin dialog, that would put it in plain view.

PR #16383, opened 2023-08-04
Add copy, paste, cut functionality

casperlamboo:
This PR adds copy, past and cut of buildplate models to cura.

https://github.com/Ultimaker/Cura/assets/6638028/974e53e7-8f3f-4820-9e7c-125d723c03e7

Copy/paste works even after a model has been deleted for the build plate (no reference is needed to the orginal model). This works by writing the serialized model data to the clipboard. As a consequence a users clipboard might look similar to this after using cura
```
<?xml version="1.0"?>
<model unit="millimeter" xmlns="http://schemas.microsoft.com/3dmanufacturing/core/2015/02" xmlns:cura="http://software.ultimaker.com/xml/cura/3mf/2015/10" xml:lang="en-US">
	<resources>
		<object id="1" name="cube.stl" type="model">
			<metadatagroup>
				<metadata name="cura:extruder_nr">0</metadata>
			</metadatagroup>
			<mesh>
				<vertices>
					<vertex x="-5" y="5" z="-5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="5" y="5" z="-5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="-5" y="5" z="-5" />
					<vertex x="-5" y="5" z="5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="5" y="-5" z="5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="-5" y="-5" z="-5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="-5" y="5" z="5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="5" y="-5" z="5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="5" y="5" z="-5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="5" y="5" z="5" />
					<vertex x="5" y="-5" z="5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="-5" y="5" z="-5" />
					<vertex x="5" y="5" z="-5" />
					<vertex x="-5" y="5" z="-5" />
					<vertex x="5" y="-5" z="-5" />
					<vertex x="-5" y="-5" z="-5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="-5" y="5" z="-5" />
					<vertex x="-5" y="-5" z="-5" />
					<vertex x="-5" y="5" z="-5" />
					<vertex x="-5" y="-5" z="5" />
					<vertex x="-5" y="5" z="5" />
				</vertices>
				<triangles>
					<triangle v1="0" v2="1" v3="2" />
					<triangle v1="3" v2="4" v3="5" />
					<triangle v1="6" v2="7" v3="8" />
					<triangle v1="9" v2="10" v3="11" />
					<triangle v1="12" v2="13" v3="14" />
					<triangle v1="15" v2="16" v3="17" />
					<triangle v1="18" v2="19" v3="20" />
					<triangle v1="21" v2="22" v3="23" />
					<triangle v1="24" v2="25" v3="26" />
					<triangle v1="27" v2="28" v3="29" />
					<triangle v1="30" v2="31" v3="32" />
					<triangle v1="33" v2="34" v3="35" />
				</triangles>
			</mesh>
		</object>
	</resources>
	<build>
		<item objectid="1" transform="1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0 16.52989959716797 5.0 -0.002300000051036477" />
	</build>
</model>
```
After a model has been pasted the newly added objects are automatically selected.

The "cut" functionality is also added which simply works by sequencing the copy and delete selected actions.
https://github.com/Ultimaker/Cura/blob/f07faac4223d3d15d36b486e1fce18a84ea55fb0/cura/CuraActions.py#L191C6-L193

Copy paste cut can be executed by either pressing `ctrl+c`, `ctrl+v`, `ctrl+x` keys or by the right click context menu

Screenshot 2023-08-04 at 13 32 28

CURA-7913

### DIFF
--- a/cura/CuraActions.py
+++ b/cura/CuraActions.py
@@ -1,15 +1,20 @@
-# Copyright (c) 2018 Ultimaker B.V.
+# Copyright (c) 2023 UltiMaker
 # Cura is released under the terms of the LGPLv3 or higher.
 
-from PyQt6.QtCore import QObject, QUrl
-from PyQt6.QtGui import QDesktopServices
 from typing import List, cast
+
+from PyQt6.QtCore import QObject, QUrl, QMimeData
+from PyQt6.QtGui import QDesktopServices
+from PyQt6.QtWidgets import QApplication
+
+import pySavitar as Savitar
 
 from UM.Event import CallFunctionEvent
 from UM.FlameProfiler import pyqtSlot
 from UM.Math.Vector import Vector
 from UM.Scene.Selection import Selection
 from UM.Scene.Iterator.BreadthFirstIterator import BreadthFirstIterator
+from UM.Scene.Iterator.DepthFirstIterator import DepthFirstIterator
 from UM.Operations.GroupedOperation import GroupedOperation
 from UM.Operations.RemoveSceneNodeOperation import RemoveSceneNodeOperation
 from UM.Operations.TranslateOperation import TranslateOperation
@@ -19,6 +24,7 @@ from cura.Operations.SetParentOperation import SetParentOperation
 from cura.MultiplyObjectsJob import MultiplyObjectsJob
 from cura.Settings.SetObjectExtruderOperation import SetObjectExtruderOperation
 from cura.Settings.ExtruderManager import ExtruderManager
+from cura.Arranging.Nest2DArrange import createGroupOperationForArrange
 
 from cura.Operations.SetBuildPlateNumberOperation import SetBuildPlateNumberOperation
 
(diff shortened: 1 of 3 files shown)


cura/CuraActions.py
```diff
@@ -181,5 +187,63 @@ def setBuildPlateForSelection(self, build_plate_nr: int) -> None:
...
+        self.copy()
+        self.deleteSelection()
+
+    @pyqtSlot()
+    def copy(self) -> None:
+        # Convert all selected objects to a Savitar scene
+        savitar_scene = Savitar.Scene()
+        mesh_writer = cura.CuraApplication.CuraApplication.getInstance().getMeshFileHandler().getWriter("3MFWriter")
+        for scene_node in Selection.getAllSelectedObjects():
+            savitar_node = mesh_writer._convertUMNodeToSavitarNode(scene_node)
```
nallath, line 201: I don't think that this is a good idea. This is a protected function.

nallath: I think it would be better to move the entire savitar logic to the 3mf writer. So basicly we would need to add a "toSceneString" (or something like that) function that accepts a list of nodes to the writer

cura/CuraActions.py
```diff
@@ -181,5 +187,63 @@ def setBuildPlateForSelection(self, build_plate_nr: int) -> None:
...
+
+        # Parse the scene from the clipboard
+        scene_string = QApplication.clipboard().text()
+        parser = Savitar.ThreeMFParser()
+        scene = parser.parse(scene_string)
+
+        # Convert the scene to scene nodes
+        nodes = []
+        mesh_reader = application.getMeshFileHandler().getReaderForFile(".3mf")
+        for savitar_node in scene.getSceneNodes():
```
nallath, line 223: Same as above, but inverted; Just add a function that returns a list of nodes given a string.

nallath (changes requested):

cura/CuraActions.py
casperlamboo: https://github.com/Ultimaker/Cura/pull/16383/commits/f8b3fb3d67890b5ee3d2aeac945c9ff136cf62bf

cura/CuraActions.py
casperlamboo: https://github.com/Ultimaker/Cura/pull/16383/commits/f8b3fb3d67890b5ee3d2aeac945c9ff136cf62bf

cura/CuraActions.py
```diff
@@ -181,5 +185,50 @@ def setBuildPlateForSelection(self, build_plate_nr: int) -> None:
...
+        QApplication.clipboard().setText(scene_string)
+
+    @pyqtSlot()
+    def paste(self) -> None:
+        application = cura.CuraApplication.CuraApplication.getInstance()
+
+        # Parse the scene from the clipboard
+        scene_string = QApplication.clipboard().text()
+
+        mesh_reader = application.getMeshFileHandler().getReaderForFile(".3mf")
```
nallath, line 211: Derp. I didn't notice this the previous time. But you can actually disable 3mf reader

If the user did this, there won't be mesh_reader returned here.

nallath (changes requested):

cura/CuraActions.py
fieldOfView: You could make the 3mf reader a required plugin, or you could disable copy/paste if the plugin is not available.

fieldOfView:
In response to the screen recording above: people are going to complain that the resulting objects get auto-arranged, disregarding the original relative positions. One of the main usecases for copy/paste - I imagine - would be to create grids.

cura/CuraActions.py
casperlamboo: https://github.com/Ultimaker/Cura/pull/16383/commits/c393d915d7e85e7437846475d416ead9c3455c0d

nallath (approved):

PR #16394, opened 2023-08-05
Change minimum support area to be 0 when tree supports are used (Elegoo profile)

ThomasRahm:
# Description

Change `minimum support area` to be 0 when tree supports are used. This should prevent confusion like https://www.reddit.com/r/Cura/comments/15ggbuj/cura_54_not_generating_tree_supports/

## Type of change

- Printer definition file(s)


ThomasRahm:
Fixes https://github.com/Ultimaker/Cura/issues/16398

rburema (approved):

PR #16384, opened 2023-08-04
[CURA-10670] Re-enable Small Top/Bottom Width

rburema:
Previously disabled just before the 5.4 release, now fixed and re-enabled.

See also backend PR here: https://github.com/Ultimaker/CuraEngine/pull/1923

PR #16429, opened 2023-08-08
[CURA-10829] Option to not use small skin areas on surface

rburema:
Please review & merge the other one first: https://github.com/Ultimaker/Cura/pull/16384

This one has a backend counterpart as well: https://github.com/Ultimaker/CuraEngine/pull/1925


pkuiper-ultimaker (changes requested):
**small_skin_width**
I'm planning to set this width to a very large value to cover the full skin area with walls. So this warning level will then be triggered, which is strange for our official profiles. My suggestion is to not set a warning level.
"maximum_value_warning": "skin_line_width * 3"

rburema:
@pkuiper-ultimaker Saumya had the same (re)quest(ion). She's going to remove the warning line, then hopefully we can merge it in as is now (though a bit of a regression was noted, which we'll investigate first).

rburema:
@pkuiper-ultimaker Also, question: If we want to fill the full skin area, then why not just set the skin-pattern to concentric?

casperlamboo (approved):

PR #16523, opened 2023-08-20
CURA 9494 nightly builds

jellespijker:
# Description

This PR is based upon #16516 merge that one first. then change the base of this PR to `main`.

Once a day the "All Installer" workflow is triggered for main branch and all the binaries are uploaded to an Nightly GH release page. Binaries from the previous day are overwritten, with the current ones. Since a GH release requires and tagged commit I tagged the `5.4.0` with `nightly` since I was hesitant to do this for the actual tip commit, because that might mess up a lot of watcher bots and spam the GH news feeds.

image

See example run (manually triggered): https://github.com/Ultimaker/Cura/actions/runs/5918809489

The currently unpublished nightly release can be found here: https://github.com/Ultimaker/Cura/releases/tag/untagged-d1fcc56cda6538506161

This will also improve QA's life, because they should be able to install the latest msi each morning by simply running: `msiexec /i https://github.com/Ultimaker/Cura/releases/tag/nigthly/UltiMaker-Cura-5.5.0-alpha-windows-X64.msi` 

This PR doesn't introduce the custom nightly theme, if people feel very strongly about that then I can add the logic for that. But I personally think that it is not needed. The working versions (+<build_info>) are after all present in the 3mf, logs and gcode already, and with this change it will also show up in the title bar and as a warning message.

I added the version number in the Title bar if this is anything other then an actual release or beta, e.q.: 5.4.0 or 5.5.0-beta.1 as soon as there is an `-alpha+1234` it will show up in the title bar. It will also show a warning message each time you start up the nightly build.
We might want to add a link in there as well.

The Titlebar info and the warning message are only shown in actual builds and won't show up when running from source.

image


.github/workflows/installers.yml
```diff
@@ -79,4 +92,136 @@ jobs:
...
+                uses: actions/checkout@v3
+
+            -   name: Download the run info
+                uses: actions/download-artifact@v2
+                with:
+                    name: macos-run-info
+
+            -   name: Set the run info as environment variables
+                run: |
+                    . run_info.sh
```
rburema, line 113: Last little remark: Did you mean to write `./run_info.sh` here? It works this way as well, so I'm not changing it, just a question.

jellespijker: No, the `. <file>` with the space after the dot is basically `source <file>` which reads and executes the content. But `source` is not always present in all Unix systems so the `.` is the safer more portable choice

But the space is there intentional

PR #16555, opened 2023-08-25
CURA-10307_expand-one-at-a-time-shadow

wawanbreton:
Call to _add2DAdhesionMargin is moved to a lower method so that it is always called when appropriate.
This should not cause side-effects, or at least only for one-at-a-time mode which is tagged as experimental.

### DIFF
--- a/resources/definitions/tizyx_evy.def.json
+++ b/resources/definitions/tizyx_evy.def.json
@@ -57,7 +57,7 @@
                 [25, 49],
                 [25, -49],
                 [-25, -49],
-                [25, 49]
+                [-25, 49]
             ]
         },
         "machine_heated_bed": { "default_value": true },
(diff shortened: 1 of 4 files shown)


resources/definitions/tizyx_evy.def.json
```diff
@@ -57,7 +57,7 @@
                 [25, 49],
                 [25, -49],
                 [-25, -49],
-                [25, 49]
```
saumyaj3, line 60: aren't these change unrelated the ticket? Can we perhaps move these changes to another ticket?

wawanbreton: @saumyaj3 That was my original intention, but @casperlamboo told me to include it as it is a quick fix and doesn't require a dedicated ticket.

saumyaj3: Alright then! This looks good :)

saumyaj3 (approved):

PR #13630, opened 2022-10-23
Use ARGV0 to select the executable to execute from AppRun. Intended to run CuraEngine

sree314:
I wanted to run `CuraEngine` from within the AppImage. Since `AppRun` only executes `Ultimaker-Cura`, I modified it to use AppImage's `ARGV0` environment variable to allow users to select the executable to run.

The use, download the AppImage as usual and create a symlink:

```
ln -s Ultimaker-Cura-5.0.0-Linux.AppImage CuraEngine.AppImage
```

Now, running `CuraEngine.AppImage` will launch `CuraEngine` instead of `Ultimaker-Cura`

If the name of the AppImage does not match a file inside the AppImage, `Ultimaker-Cura` will run by default.

PR #16607, opened 2023-08-29
CURA-10951 build plugin as bundled plugin on the GH runners

jellespijker:
Add the gradual flow plugin as a bundled plugin:
- https://github.com/Ultimaker/CuraEngine_plugin_gradual_flow

Needed to build the AppImage on a modern Ubuntu version again for this. Because the Gradual flow plugin is now build using the same dependency versions and options as Cura/CuraEngine.

GCC-13 and statical compiling on glibc<2.34 doesn't play nice. Switched to AppImage Builder, which will also copy the glibc from the system it was compiled. This means that we can run it from Ubuntu 18.04 again. I also made sure that the list of system packages, such as the ones needed for system file dialogs have been packed in the AppImage. All of these measurements should improve portability on different Linux Distro's. I tested it on Manjaro (KDE X11 and Wayland), Ubuntu 20.04 (Gnome X11) running in a virtualbox, Ubuntu 22.04 (Gnome X11, KDE both Wayland and Gnome).

The plugin is packaged and installed from the specified pyinstaller mappings in the conandata.yml.

Most of the changes in Savitar, nest2d, Arcus relate to update the way that the package and build with Conan and our custom SIP and pyprojecttoolchains. Since I needed generate the C++ headers and source files with SIP in a different folder, because conan <= 1.57 didn't create the build folder before the sip-build was run, but the newer versions do. Which would result in a crash in setuptools when the sources are compiled. 

See also:
- https://github.com/Ultimaker/conan-ultimaker-index/pull/8
- https://github.com/Ultimaker/CuraEngine/pull/1941
- https://github.com/Ultimaker/libSavitar/pull/49
- https://github.com/Ultimaker/libnest2d/pull/9
- https://github.com/Ultimaker/pyArcus/pull/6
- https://github.com/Ultimaker/pySavitar/pull/3
- https://github.com/Ultimaker/pynest2d/pull/29
- https://github.com/Ultimaker/CuraEngine_plugin_gradual_flow/pull/3
- https://github.com/Ultimaker/CuraEngine_grpc_defintions/pull/11

See the builds here:
- https://github.com/Ultimaker/Cura/actions/runs/6098787294

Open issues which I recommend to be resolved in a separate PR since I will be off for a few days:

- [ ] SettingsAppender from Uranium doesn't Append the json file (https://github.com/Ultimaker/CuraEngine_plugin_gradual_flow/blob/CURA-10951_gh_build_curapackage/CuraEngineGradualFlow/gradual_flow_settings.def.json) from the plugin. This issue seems to only occur on Linux in the actual AppImage. Everything still runs in the AppDir before it is packaged as an AppImage.
- [ ] Fix calling upon CuraEngine directly as introduced in this PR: https://github.com/Ultimaker/Cura/pull/13630
- [ ] Update the Wiki and README.md on respective repo's to mention the new minimum requirements: >=GCC-12, Conan >=1.59.0,<2.0


packaging/AppImage-builder/create_appimage.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    copied_files = {
+        os.path.join("..", "icons", "cura-icon.svg"): os.path.join("usr", "share", "icons", "hicolor", "scalable", "apps", "cura-icon.svg"),
+        os.path.join("..", "icons", "cura-icon_64x64.png"): os.path.join("usr", "share", "icons", "hicolor", "64x64", "apps", "cura-icon.png"),
+        os.path.join("..", "icons", "cura-icon_128x128.png"): os.path.join("usr", "share", "icons", "hicolor", "128x128", "apps", "cura-icon.png"),
+        os.path.join("..", "icons", "cura-icon_256x256.png"): os.path.join("usr", "share", "icons", "hicolor", "256x256", "apps", "cura-icon.png"),
+        os.path.join("..", "icons", "cura-icon_256x256.png"): "cura-icon.png",
+    }
+
+    # TODO: openssl.cnf ???
```
rburema, line 71: Does this need to happen?

rburema (approved):
LGTM, but let's see what happens when I merge main, since the nigthly branch is now merged.

packaging/AppImage-builder/create_appimage.py
jellespijker: Need to double check, if you can overload the openssl config. But we weren't doing that before, since that file doesn't exist.
I think it only be an issue on Windows

PR #16610, opened 2023-08-29
Refactor object reload to handle 3mf files

pietchaki:
# Description

3mf Files can have multiple models in them, and they can be named individually, but Cura was renaming all these models with its filename, not really an issue by itself but having the original names is better when there are many models in the file.
Also, when the file has multiple models, and user reloads them (right click -> Reload All Models), all objects from that file where being replaced by the first model in the file.
Another minor issue was that when reloading, Cura would read the same file multiple times, one for each model.

! Important: This should be accompanied by https://github.com/Ultimaker/Uranium/pull/898, so that both model reload methods have the same behaviour

## Type of change

# How Has This Been Tested?

- Confirmed other filetypes where not affected by the refactor:
- Opened a OBJ file and reloaded.
- Opened STL file and reloaded.
- Opened gcode file, and reloaded.

- Loaded a 3MF file into Cura, verified the name of the models in the app are the model name in the file.
- Right click the window and selected "Reload All Models", verified each model get replaced as expected.
- Updated the 3MF file removing one of the models, ignored the "File has been modified" message{1}, right click->"Reload All Models". Confirmed it works as before and keep the mesh in the scene.
- Updated the 3MF file, adding a new mesh into it, ignored the "File has been modified" message{1} and reloaded the models via right click. Confirmed it works as before and the new mesh is NOT loaded in the scene{2}. 

{1}: The message about the file being modified and asking to reload is handles by Uranium, and not in Cura. Improvements to that process will be committed to Uranium.
{2}: This should be an improvement, as the newly added model should be loaded when we reload a file. This however should be accompanied with an update to the process described in {1} so that both have the same behaviour.

**Test Configuration**:
* Operating System: Windows 10

# Checklist:


saumyaj3:
Hey @pietchaki,

Thanks for your contribution 👍 
I  was just reviewing the PR, and came across a model where name was not replaced with "object id" and took the file name still. Can you maybe take a look again at the PR?
[bug_array.zip](https://github.com/Ultimaker/Cura/files/12869324/bug_array.zip)

model_names

pietchaki:
@saumyaj3 Thanks for the report, but this file doesn't have name metadata for the objects, and for this reason we name them with the file name. That part is working as intended.

saumyaj3 (approved):

saumyaj3:
Hey @pietchaki I was doing component test of a 3mf file I created. I reloaded all the models and then I did "ctrl+C" "ctrl+V on one of the object. I cannot seem to paste it anymore. I added my model for reference. 

[UMS3_arrow.zip](https://github.com/Ultimaker/Cura/files/13227006/UMS3_arrow.zip)

saumyaj3:
Hey @pietchaki Thank you for your contribution and for your work to make Cura better . 👯 🥇

PR #16612, opened 2023-08-29
[Windows?] Engine-plugins wouldn't always close.

rburema:
The code to terminate them wasn't always called. Especially annoying on Windows, since the sub-program could apparently become a zomboid after termination of the main one(s).

I guess this is still part of the still open CURA-10475 'branch' in a way? There is no other ticket that fits at the moment.


jellespijker:
- Is this also called on a engine crash?
- Maybe we should call them on Cura exit as well?

rburema:
@jellespijker Yes to both, since `stopPlugins()` is called in `_terminate()`.

- It _should_ be called on engine crash via `_onBackedQuit()`, which calls `_terminate()` if the return-code isn't `0` (and just directly `stopPlugins()` if it _is_ `0`).
- It's called on exit of Cura via `close()`, which calls `terminate()`.

jellespijker (approved):
LGTM

PR #16615, opened 2023-08-29
Compile engine plugins with GCC-13 on linux modern and legacy

jellespijker:
seems to be available for 20.04 as well: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

Contribute to CURA-10866


rburema (approved):
I always enjoy these little 'wait we can just _do_ that?' moments :-)

PR #16626, opened 2023-08-30
CURA-10896_improve_add_printer_wizard

wawanbreton:
The add printer wizard window has been made resizable, to help selecting a printer in the list, and editing GCode.
Default size remains unchanged, you just have the option to make it bigger if needed.


resources/qml/Cura.qml
```diff
@@ -826,6 +826,8 @@ UM.MainWindow
     {
         id: addMachineDialog
         title: catalog.i18nc("@title:window", "Add Printer")
+        maximumWidth: undefined
```
saumyaj3, line 829: `
maximumWidth: -1
maximumHeight: -1
`

resources/qml/MachineSettings/GcodeTextArea.qml
```diff
@@ -70,6 +71,7 @@ UM.TooltipArea
             selectionColor: UM.Theme.getColor("text_selection")
             selectedTextColor: UM.Theme.getColor("text")
             wrapMode: TextEdit.NoWrap
+            padding: -border.width //Take care of the border width to fit the text.
```
saumyaj3, line 74: border here is out of scope. creates qml warnings. perhaps can use the borders from rectangle below as : rectid.border

wawanbreton: Oh yes indeed, I just move the code content from below but didn't check the console.. Visually it looked good :laughing:

saumyaj3:
General Comment: The size of the windows when closed is not made to default size back. It would be nice if that happens. 
for example: When I open the window of "Machine Settings" it opens the window in the size which I changed before too and also if I opened tooltip before and closed the window. it opens tooltip.

resources/qml/Cura.qml
```diff
@@ -826,6 +826,8 @@ UM.MainWindow
...
+        maximumHeight: undefined
```
saumyaj3, line 830: these two values gives QML warnings as it expects integer value. 
I don't know if its too much to ask. but is it possible to create it like a resizable window?

wawanbreton:
> The size of the windows when closed is not made to default size back

I could argue that this is somehow logical, but you're right, when you open the window a second time, you expect it to be as it was when you opened it first. I'll see to change it.

resources/qml/Cura.qml
wawanbreton: It took me a while to find the proper way to do it, without changing the default behavior of WizardDialog. Hope you'll like it :)

wawanbreton:
I have made the recommanded changes, but some remaining remarks:
* I have removed the method `WizardDialog.resetModelState()` because it is not used in Cura core anymore, but maybe it could be used by an external plugin ?
* The add printer dialog takes its original size back every time you call it, but I didn't change this behavior for other windows. However, I am no more completely sure about what is best. Dialogs really have a very small size in Cura and this it not very much appealing. I had a discussion with @casperlamboo about this subject, and maybe we could open the topic to the rest of the team. I know it's just windows sizes, but it's bothering me. Maybe we could use the actual screen size to set the default dialogs sizes, or something like that, which would benefit all the users.
* I saw that a few dialogs (at least "add printer" and "what's new") are instantiated at startup and shown/hidden when necessary. Although very easy, it is not optimized, e.g. the "what'ts new" dialog is always instantiated and loaded even if it is never displayed, and it may contain images/videos... Maybe we should make a pass to instantiate the windows only when required, unless there is a good reason not to do it ?

saumyaj3:
> I have made the recommanded changes, but some remaining remarks:
> 
> * I have removed the method `WizardDialog.resetModelState()` because it is not used in Cura core anymore, but maybe it could be used by an external plugin ?
> * The add printer dialog takes its original size back every time you call it, but I didn't change this behavior for other windows. However, I am no more completely sure about what is best. Dialogs really have a very small size in Cura and this it not very much appealing. I had a discussion with @casperlamboo about this subject, and maybe we could open the topic to the rest of the team. I know it's just windows sizes, but it's bothering me. Maybe we could use the actual screen size to set the default dialogs sizes, or something like that, which would benefit all the users.
> * I saw that a few dialogs (at least "add printer" and "what's new") are instantiated at startup and shown/hidden when necessary. Although very easy, it is not optimized, e.g. the "what'ts new" dialog is always instantiated and loaded even if it is never displayed, and it may contain images/videos... Maybe we should make a pass to instantiate the windows only when required, unless there is a good reason not to do it ?

API calls are mentioned here "https://github.com/Ultimaker/Cura/wiki/SDK-Versions" I think we can safely remove it. I do agree with second and third point you mentioned.

saumyaj3 (approved):

PR #16646, opened 2023-09-02
Update creality_ender3.def.json

fooshards:
ender 3 v2 will heat bed and nozzle at same time - saves time from autogenerated generic printer heat up gcode which forces the poor printer to do it sequentially and waste your time.

# Description

This improves print time by allowing the ender 3 v2 profile to utilize both the nozzle heater and bed heater at the same time.
This should probably apply to a whole range of printers, but I do not have those printers sitting next to me like I do the creality_ender3.def.
If you would like to build a buffer into the time between "bed heated" and "extrusion begins" you can do so with a M0 or G04, like a 15 second buffer, to allow the heat to radiate to the entire plate.  This is much more purposeful than "however variable long time it takes to get the nozzle up to temp this time" - which is the current strategy in the generated gcode.  My opinion is that the wait is unnecessary in both scenarios, but my opinions are often stupid, so I defer to the maintainers.

## Type of change

# How Has This Been Tested?

via machine settings -> selecting the printer -> copypastaing the changed content into the "start g-code" block, slicing a file, and printing it on the described stock printer

**Test Configuration**:
* Operating System: Win11 Insider Canary 25941.1000
* Printer: [Stock] Creality Ender 3 V2


GregValiant:
The reason that the heaters didn't start together was that the original Ender 3 power supply wasn't capable of handling the start-up load of both heaters at the same time.  The Ender 3 Pro power supply was/is more robust and could handle it.  The weak point for both was that all the electrical current goes through the mainboard terminals which can cause a high resistance connection at the terminals and subsequently a fire hazard.
Here's mine.

DSCN2577

scotty6435:
This change could easily be rolled out across most if not all Ender 3 models. My v3 SE also pre-heats both natively but when controlled by Cura will pre-heat one-at-a-time

scotty6435:
> _"My v3 SE also pre-heats both natively but when controlled by Cura will pre-heat one-at-a-time"._ The reason for that is in that photo above. It's a safety. Any individual can alter the startup for their particular machine in the Machine Settings. Then it's on them. Your proposed startup has no mention of ABL. When bed-leveling is performed with a hot nozzle it will ooze all over the bed. That also needs to be taken into account. Maybe a user only wants to go to Standby Temperature prior to the ABL routine and then heat the rest of the way after moving to a corner to ooze. I think this is better left to the user to customize their own StartUp Gcode.

Understandable however probably 90%+ of Ender 3s (all revisions) run in this mode and only a small number have ever failed. I'd also wager that their quality control has improved given the relative lack of reports since the initial surge in 2019. I don't really use Cura to monitor as it offers few benefits and increases the risk of a failed print due to disconnect/restart.

fooshards:
It's not that big of a deal, I fixed my own issue but since it was part of the literal defaults I thought I could share back and demystify things for future users.  That fried board is enough convincing to me to not touch the real values, but I still think yall should add clear comments so that the intent is clear, and it isn't being doubted as a misconfiguration by the user as they wait extra-long for the first layer to test.  I'll close this because I dont want my name on the pull request that fries someones poor printer.  Do what ya wanna do, it's your project.

PR #16666, opened 2023-09-05
Printjob icon

saumyaj3:
Printjob icon updated


wawanbreton:
Shouldn't you remove the FoodBeverages.svg as it is not used anymore ? And/or register it in the deprecated_icons.json ?

saumyaj3:
I was considering to remove it, but decided not to do it, as the icon's size is relatively small, and it may still be of use to the plugins. what do you think?

wawanbreton:
Oh sorry I didn't read your message carefully.. You're right about the plugins. We could consider that they should now use the new icon, but some old plugins may not. Then I don't know what should be done...

wawanbreton:
@saumyaj3 Can you add an entry to resources/themes/cura-light/icons/deprecated_icons.json ?

saumyaj3:
> @saumyaj3 Can you add an entry to resources/themes/cura-light/icons/deprecated_icons.json ?

On it!

PR #16683, opened 2023-09-07
[CURA-10870] Prevent non-'hard' crashes from re-setting the window-positions.

rburema:
See Ultimaker/Uranium#900 for more details. (This is the little bit of front-end support we need for that.)

### DIFF
--- a/cura/CrashHandler.py
+++ b/cura/CrashHandler.py
@@ -451,6 +451,9 @@ class CrashHandler:
         os._exit(1)
 
     def show(self):
+        Application.getInstance().getPreferences().setValue("general/last_shutdown_severity",
+                                                            Application.ShutdownSeverity.CrashHandler)
+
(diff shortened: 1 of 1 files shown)


cura/CrashHandler.py
```diff
@@ -451,6 +451,9 @@ def _sendCrashReport(self):
         os._exit(1)
 
     def show(self):
+        Application.getInstance().getPreferences().setValue("general/last_shutdown_severity",
+                                                            Application.ShutdownSeverity.CrashHandler)
```
wawanbreton, line 455: This won't work if Cura crashes within 10 seconds, which is the delay for saving the preferences file. But maybe that's acceptable.

wawanbreton: Sorry, wrong part of code.

rburema:
This design isn't working at all in it's current form. Multiple problems exist with it. As the time-box has run out, I'm closing this.

PR #16711, opened 2023-09-11
Opening the value of setting only if they are not in missing packages

saumyaj3:
CURA-10915

"Open project anyway" should remove all plugin settings of non-installed plugins


plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -1133,7 +1133,15 @@ def _applyUserChanges(self, global_stack, extruder_stack_dict):
                         extruder_stack.userChanges.setProperty(key, "value", value)
                 if parser is not None:
                     for key, value in parser["values"].items():
-                        extruder_stack.userChanges.setProperty(key, "value", value)
+                        value_not_in_missing_package = True
+                        if "PLUGIN" in value:
```
wawanbreton, line 1137: Maybe be more specific ? value.startwith("PLUGIN::")

You could also use a RegExp, it is a bit overkill for the test, but that allows you to extract the plugin name and version at the same time, so that the test below would be more simple.

These are just suggestions, feel free to reject them.

plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -1133,7 +1133,15 @@ def _applyUserChanges(self, global_stack, extruder_stack_dict):
...
+                            for package in self._dialog.missingPackages:
+                                if (package['id']+"@"+package['package_version']) in value:
+                                    value_not_in_missing_package = False
+                                    Logger.log("w", f"Ignoring {key} value {value} from missing package")
+
+                        if value_not_in_missing_package :
+                            extruder_stack.userChanges.setProperty(key, "value", value)
```
wawanbreton, line 1144: Is it also possible that a plugin has settings in other stacks (i.e. global) ?

plugins/3MFReader/ThreeMFWorkspaceReader.py
saumyaj3: I find it a bit complicated to use RegExp here. Do you have a suggestion?

plugins/3MFReader/ThreeMFWorkspaceReader.py
saumyaj3: I did the same for all setting places now.

plugins/3MFReader/ThreeMFWorkspaceReader.py
wawanbreton: In this case it is not really helpful. However you could still use `value.startwith("PLUGIN::")` which will be much faster in case the `PLUGIN` word is not in the string (which is very likely). That implies it is always supposed to be at the beginning though.

saumyaj3: I did that in the new commit :)

wawanbreton: Yeah I just saw that :laughing: :+1:

PR #16713, opened 2023-09-11
Introduces new Dagoma printer

0r31:
Hi,

As a Dagoma collaborator, I submit this PR to add one delta printer (and its profiles -variants, qualities-) called Sigma Pro.

Please don't hesitate to make any useful comments if something is wrong according to your [profile creation guidelines](https://github.com/Ultimaker/Cura/wiki/Profiles-&-Settings).

Regards,
Orel


0r31:
Hi @MariMakes ,

Any update on this PR? Is there something wrong?

0r31:
Hi @nallath ,

Any chance to merge this PR before the next release?

Regards,

wawanbreton:
Hi @0r31,

Sorry for this very long review delay, but we have a lot of similar ones to process.

I see you have done a very nice job, this all looks good to me. I made a quick test with the most recent Cura source code, and it works flowlessly.

I just have a remark about the prime tower, which is set at a fixed position. The XY position has changed in the meantime, due to the addition of a prime tower "base". I would recommend not setting the position at all and let it be calculated by the default formula. This will actually have it at the same place as you have set, but properly move when changing the size or other settings.

Speaking of prime tower, I see that you have a dual color head with a single physical nozzle. In the incoming Cura 5.7, we have integrated a new "interleaved" prime tower type which can greatly reduce the amount of unnecessary primed material. Feel free to test it !

In order to move forwards with this, please:
* Update your branch to the last `Cura/main`
* Either remove the prime tower position or confirm that you want it at this specific position

0r31:
Hi @wawanbreton ,

Many thanks for your feedback !

> In order to move forwards with this, please:
> - Update your branch to the last Cura/main
> - Either remove the prime tower position or confirm that you want it at this specific position

I've just updated the PR branch and followed you advice in removing prime tower position.

>Speaking of prime tower, I see that you have a dual color head with a single physical nozzle. In the incoming Cura 5.7, we have integrated a new "interleaved" prime tower type which can greatly reduce the amount of unnecessary primed material. Feel free to test it !

That's a great news ! I will test it asap !

Regards,

0r31:
(it seems there is an issue in the action runner used for unit-test action. `apt-get update` fails, permission issue on lock file)

wawanbreton:
Thanks for your quick changes ! I had to make some more checks but everything is fine for me.
Don't worry about the unit tests, I know this PR in not the cause of it :smiley:

PR #16812, opened 2023-09-25
Tooltip affects and affected by list fixing

saumyaj3:
CURA-11002

It was noticed that tooltip for settings was showing repeated affects and affected by. 
This PR is fixing that

### DIFF
--- a/resources/qml/Settings/SettingItem.qml
+++ b/resources/qml/Settings/SettingItem.qml
@@ -62,7 +62,7 @@ Item
         var affected_by_list = ""
         for (var i in affected_by)
         {
-            if(affected_by[i].label != "")
+            if((affected_by[i].label != "") && (! affected_by_list.includes("<li>%1</li>\n".arg(affected_by[i].label))))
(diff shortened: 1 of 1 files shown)


fieldOfView:
IMHO it would be a much cleaner (and more performant!) approach to clean the list before presenting it. Now you are formatting the string twice, and doing expensive lookups for corner-cases. It would be much better if this were handled in `settingDefinitionsModel.getRequiredBy` and `settingDefinitionsModel.getRequires`

saumyaj3:
> IMHO it would be a much cleaner (and more performant!) approach to clean the list before presenting it. Now you are formatting the string twice, and doing expensive lookups for corner-cases. It would be much better if this were handled in `settingDefinitionsModel.getRequiredBy` and `settingDefinitionsModel.getRequires`

Hey @fieldOfView Thanks for your honest opinion. Now, that you've mentioned it, I too think I shall do this filtering in root itself. Thanks for the quick review.

PR #16826, opened 2023-09-26
Fix crash when adding some printers

wawanbreton:
CURA-11089

Within some conditions, Cura crashes when adding some specific printers, because they use an "empty" material which has no brand reference, and we try to access it.


rburema (approved):
lgtm!

wawanbreton:
> lgtm!

Which means.... ?

rburema:
> Which means.... ?

@wawanbreton 'looks good to me', sorry, should have realized it's a bit jargon-y 😅

wawanbreton:
Ok thanks, don't worry I'll get used to it :)

PR #16829, opened 2023-09-26
[CURA-10994] Translations for 5.5.x

rburema:


resources/i18n/pt_BR/uranium.po
wawanbreton, line 1: This file should probably be in Uranium PR only

wawanbreton (commented):
Why don't some languages have an updated cura.po ?

wawanbreton:
Gradual flow settings are not translated:

image

wawanbreton:
> Gradual flow settings are not translated:

Apparently the label is properly included in the deployed file, but for some reason it is not translated at display-time

rburema:
> Why don't some languages have an updated cura.po ?

We only translate part of this list. About 11 of the languages are translated 'officially', with English making the 12th. The rest are volunteer translations, of which only 2 are active, Czech and Brazilian Portuguese. (These'll come in later.) The rest (un-active) of _those_ are abandoned, and are selectable as 'partial' translations from the list. Some of these are quite old.

PR #16856, opened 2023-10-01
Remove material_print_temperature from sovol_base_planetary.def.json

mstrent:
# Description

material_print_temperature should not be defined at the printer level as it overrides material default print temperature. Every gcode file comes out at 195* unless this setting is removed here, regardless of your material settings.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [x ] Printer definition file(s)
- [ ] Translations

# How Has This Been Tested?

Tested by editing this file on the installed version of Cura on my PC. After this setting is removed at the printer definition level, the default temperature setting at the material definition level actually affects the final product, as you would expect.


mstrent:
This is a duplicate of, and fixes one less problem than #16107. Though I've only tested the temperature part.

PR #16896, opened 2023-10-05
Profile updates 5.5

pkuiper-ultimaker:
- Restored PETG and ABS profiles. 
- Lowered outer wall speed to 30mm/s for default mode. 
- Reduced gradual flow for PETG and ABS to 1. 
- Reduced inner wall speeds on top surface from 100mm/s to 45mm/s.
- Make Ultimaker PLA blue the default material iso Generic PLA

### DIFF
--- a/resources/definitions/ultimaker_s3.def.json
+++ b/resources/definitions/ultimaker_s3.def.json
@@ -41,6 +41,7 @@
             0
         ],
         "platform_texture": "UltimakerS3backplate.png",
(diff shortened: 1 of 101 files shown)


casperlamboo (approved):

saumyaj3:
Created the same PR: https://github.com/Ultimaker/Cura/pull/16898 . Branch from 5.5 instead of main

PR #17022, opened 2023-10-18
[CURA-10407] Fix description now that we have fractional layer-heights (for top).

rburema:
relates to the changes in https://github.com/Ultimaker/CuraEngine/pull/1955


resources/definitions/fdmprinter.def.json
```diff
@@ -5144,7 +5144,7 @@
                         "support_top_distance":
                         {
                             "label": "Support Top Distance",
-                            "description": "Distance from the top of the support to the print.",
+                            "description": "Distance from the top of the support to the print. Note that this is not rounded.",
```
wawanbreton, line 5147: In this case I would not mention it because this is what one would expect. But too much description does not hurt..

resources/definitions/fdmprinter.def.json
```diff
@@ -5130,7 +5130,7 @@
                 "support_z_distance":
                 {
                     "label": "Support Z Distance",
-                    "description": "Distance from the top/bottom of the support structure to the print. This gap provides clearance to remove the supports after the model is printed. This value is rounded up to a multiple of the layer height.",
+                    "description": "Distance from the top/bottom of the support structure to the print. This gap provides clearance to remove the supports after the model is printed. Apart from the top-layer, this value is rounded up to a multiple of the layer height.",
```
wawanbreton, line 5133: This does not look clear to me... If you have e.g. a layer height 0.2 and Z distance 0.44, you will get 2 "normal" layers and a pseudo 0.04 layer right ? If this is correct, then last part is more confusing than helping.

resources/definitions/fdmprinter.def.json
```diff
@@ -5158,7 +5158,7 @@
                         "support_bottom_distance":
                         {
                             "label": "Support Bottom Distance",
-                            "description": "Distance from the print to the bottom of the support.",
+                            "description": "Distance from the print to the bottom of the support. Not that this is rounded up to the next layer height.",
```
wawanbreton, line 5161: "Not" => "Note"

wawanbreton (changes requested):
1 typo and 2 improvements suggestions

resources/definitions/fdmprinter.def.json
rburema: Feel free to remove it :-)

resources/definitions/fdmprinter.def.json
rburema: I can see how it might be unclear, do you have a suggestion?

wawanbreton: Well, I would not talk about the "normal" layers because what they will do seem obvious. So I would just say something like "The topmost layer might be shorter that regular layers to respect the total distance"

PR #17029, opened 2023-10-19
Remove "no skin in z gap" from profiles

pkuiper-ultimaker:
# Description

The "No skin in z gap" was enabled in the profiles and caused issues in some prints on the skin layers of AA0.8 prints.
Disabled the option.

Relates to: CURA-11187


rburema (approved):
Looks good to me!

PR #17030, opened 2023-10-19
Fix memory leak issues in monitor page

casperlamboo:
# Description

Fix memory leak issues in monitor page. Issue was that QT (incorrectly?) asserted that there was a binding loop between width and height in the `NetworkMJPGImage` component. This caused the height to evalualte to `infinity`, making QT create a buffer with an infinite amount of memory. Solved by calculating a serpeate `img_scale_factor` which both the width and height uses.

CURA-11180

## Type of change

- [ ] Translations

# How Has This Been Tested?

Yes
**Test Configuration**:
* Operating System:
* macos arm64

# Checklist:

### DIFF
--- a/plugins/UM3NetworkPrinting/resources/qml/PrinterVideoStream.qml
+++ b/plugins/UM3NetworkPrinting/resources/qml/PrinterVideoStream.qml
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Ultimaker B.V.
+// Copyright (c) 2023 UltiMaker
 // Cura is released under the terms of the LGPLv3 or higher.
 
 import QtQuick 2.2
@@ -6,7 +6,7 @@ import UM 1.3 as UM
 import Cura 1.0 as Cura
 
 Item {
-    property var cameraUrl: "";
+    property string cameraUrl: "";
 
     Rectangle {
         anchors.fill:parent;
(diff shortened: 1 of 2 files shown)


nallath (approved):

plugins/UM3NetworkPrinting/resources/qml/PrinterVideoStream.qml
```diff
@@ -34,22 +34,29 @@ Item {
...
     Cura.NetworkMJPGImage {
         id: cameraImage
-        anchors.horizontalCenter: parent.horizontalCenter;
-        anchors.verticalCenter: parent.verticalCenter;
-        height: Math.round((imageHeight === 0 ? 600 * screenScaleFactor : imageHeight) * width / imageWidth);
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.verticalCenter: parent.verticalCenter
+
+        readonly property real img_scale_factor: {
+            if (imageWidth > maximumWidth || imageHeight > maximumHeight) {
```
rburema, line 41: Just to double-check: There's no way _both_ of these will be `inf`, `nan` or `0.0` right?

rburema (approved):

plugins/UM3NetworkPrinting/resources/qml/PrinterVideoStream.qml
casperlamboo: Both `imageWidth` and `imageHeight` _might_ be 0 (initially they are always 0 because the image here is only initialized, not loaded https://github.com/Ultimaker/Cura/blob/0e1262126a4400633c2a57f7a1f2e230b7c36fd4/cura/PrinterOutput/NetworkMJPGImage.py#L79). They can never be Inf or NaN

casperlamboo: Both `imageWidth` and `imageHeight` _might_ be 0 (initially they are always 0 because the image here is only initialized, not loaded here)
 https://github.com/Ultimaker/Cura/blob/0e1262126a4400633c2a57f7a1f2e230b7c36fd4/cura/PrinterOutput/NetworkMJPGImage.py#L78-L80
 They can never be Inf or NaN

PR #17036, opened 2023-10-20
Cura 10403 revisit auto arrange strategy

wawanbreton:
Auto-arrange uses the placement algorithm provided by libnest2d. It seems to perform very well with many small objects, but a bit less with medium/big ones. The main issue is that objects are placed one after another, and their relative position can't change once they have been placed. This is a strategy that allows for quick convergence, but does not always give a result.

However, some minor improvements have been performed to improve the global behavior:
* Give a smaller placement area instead of giving the whole build plate with forbidden areas around it. This seems to make the work of libnest2d easier.
* Try two different placement strategies (position by center / by corner), 3 times each to maximize the probability of success. Both strategies work better in different situations, so trying them both really helps.

### DIFF
--- a/cura/Arranging/Nest2DArrange.py
+++ b/cura/Arranging/Nest2DArrange.py
@@ -81,7 +81,6 @@ class Nest2DArrange(Arranger):
         ], numpy.float32))
 
         disallowed_areas = self._build_volume.getDisallowedAreas()
-        num_disallowed_areas_added = 0
         for area in disallowed_areas:
             converted_points = []
 
@@ -96,7 +95,6 @@ class Nest2DArrange(Arranger):
                 disallowed_area = Item(converted_points)
                 disallowed_area.markAsDisallowedAreaInBin(0)
                 node_items.append(disallowed_area)
-                num_disallowed_areas_added += 1
 
         for node in self._fixed_nodes:
             converted_points = []
@@ -109,7 +107,6 @@ class Nest2DArrange(Arranger):
                 item = Item(converted_points)
                 item.markAsFixedInBin(0)
                 node_items.append(item)
-                num_disallowed_areas_added += 1
 
         config = NfpConfig()
         config.accuracy = 1.0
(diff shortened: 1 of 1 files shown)


cura/Arranging/Nest2DArrange.py
```diff
@@ -108,20 +107,25 @@ def findNodePlacement(self) -> Tuple[bool, List[Item]]:
...
-                num_disallowed_areas_added += 1
 
-        config = NfpConfig()
-        config.accuracy = 1.0
-        config.alignment = NfpConfig.Alignment.DONT_ALIGN
-        if self._lock_rotation:
-            config.rotations = [0.0]
+        tried_strategies = [NfpConfig.Alignment.CENTER] * 3 + [NfpConfig.Alignment.BOTTOM_LEFT] * 3
+        found_solution_for_all = False
+        while not found_solution_for_all and len(tried_strategies) > 0:
```
casperlamboo, line 113: I'm not sure about the naming of the variable `tried_strategies`. Since it is past-tense it implies the stragegies are already performed (and rejected since we continued). I would rename the variable to `strategies` instead. 

Additionally, I'm not a big fan of the `while ... len(stragegies) > 0` and `tried_strategies = tried_strategies[1:]`. This is just a more "complicated" loop in my opinion. The `while not found_solution_for_all` is nice though. However, I would instead write something like
```suggestion
        for strategy in strategies
            ....
            if found_solution_for_all:
               break
```
But that is more personal taste

casperlamboo:
Overal looks good! Think this will result in a nice improvement. Had one code style remark which is more personal preference but please do consider the suggested changes!

cura/Arranging/Nest2DArrange.py
wawanbreton: I agree with the improper variable naming, and will change that.
However, about the change of the loop, this is the recommended way. Some purists would say that you should never use a break/continue statement, because they interrupt the flow execution and allow for a "jump", like a "goto" :smile: . I do use them sometimes, but in this case I find it more readable because you can know when the loop will stop without having to look into it.

PR #17046, opened 2023-10-20
Add iso view to snapshot

casperlamboo:


rburema (commented):
If this turns out to work in all cases (which it should, as far as I can figure out ... but fingers crossed for all systems), I like the new method a bit better than the old snapshot, which was a bit of a workaround, due to the fact that for 'normal' rendering we reuse everything, and the snapshot system was a bit 'hacked in' there.

I think it's acceptable (and more clean in a way) to recreate a `QtRenderer` here, since we're not repeating this operation a whole lot anyway.

It does annoy me a bit that the 'normal' snapshot and this version are now so separate, but that's the cost of incremental progress ;-)

PR #17124, opened 2023-10-27
Add setting for specific prime tower raft line spacing

wawanbreton:
Simply add a setting to allow for a different line spacing for the raft of the prime tower's first layer, because in some situations we want extra adhesion for it as it has a low footprint but can get very tall.

This goes along with https://github.com/Ultimaker/CuraEngine/pull/1974

CURA-11233


casperlamboo (changes requested):
Looks good, one small remark: the setting should be added to the correct visibility (https://github.com/Ultimaker/Cura/tree/main/resources/setting_visibility). I think it should go in to both advanced and expert

PR #17164, opened 2023-10-31
Method X and Method XL profiles for ABS-R, ABS-CF and RapidRinse

pkuiper-ultimaker:
Method Profiles


jellespijker (approved):
LGTM;

But I'm missing the meshes and other resources. I will add those myself after I merged this branch

PR #17181, opened 2023-11-02
Intent type translated while reading 3mf

saumyaj3:
CURA-11221

When loading a profile containing engineering or draft intent, the naming will be displayed as Accurate and Quick.

### DIFF
--- a/plugins/3MFReader/WorkspaceDialog.py
+++ b/plugins/3MFReader/WorkspaceDialog.py
@@ -222,8 +222,14 @@ class WorkspaceDialog(QObject):
         return self._intent_name
 
     def setIntentName(self, intent_name: str) -> None:
+        intent_name = intent_name.title()
         if self._intent_name != intent_name:
-            self._intent_name = intent_name
+            if intent_name == "":
+                self._intent_name = "Balanced"
+            elif intent_name == "Quick":
+                self._intent_name = "Draft"
+            else:
+                self._intent_name = intent_name
             self.intentNameChanged.emit()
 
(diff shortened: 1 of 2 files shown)


plugins/3MFReader/WorkspaceDialog.py
```diff
@@ -222,8 +222,14 @@ def intentName(self) -> str:
...
     def setIntentName(self, intent_name: str) -> None:
+        intent_name = intent_name.title()
         if self._intent_name != intent_name:
-            self._intent_name = intent_name
+            if intent_name == "":
+                self._intent_name = "Balanced"
+            elif intent_name == "Quick":
+                self._intent_name = "Draft"
+            else:
+                self._intent_name = intent_name
```
casperlamboo, line 232: doing things this way won't translate the strings, additionaly we have this file that contains (translated) intent names, maybe we can use that?

https://github.com/Ultimaker/Cura/blob/main/cura/Machines/Models/IntentTranslations.py

casperlamboo (changes requested):

plugins/3MFReader/WorkspaceDialog.py
saumyaj3: b584efd021dec1b30488c21ac4a5a36311cf7feb

casperlamboo (approved):

casperlamboo:
Approved, but changes were made against wrong branch. Going to cherry-pick the two commits directly to the 5.6 branch. Otherwise, good!

PR #17186, opened 2023-11-02
 about dialog blemishes

casperlamboo:
Cura-11263

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -152,6 +152,25 @@ class CuraConan(ConanFile):
             }
         return conan_installs
 
+    def _python_installs(self):
+        python_installs = {}
+
+        # list of python installs
+        outer = '"' if self.settings.os == "Windows" else "'"
+        inner = "'" if self.settings.os == "Windows" else '"'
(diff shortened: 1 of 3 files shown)


conanfile.py
```diff
@@ -152,6 +152,25 @@ def _conan_installs(self):
             }
         return conan_installs
 
+    def _python_installs(self):
+        python_installs = {}
+
+        # list of python installs
+        outer = '"' if self.settings.os == "Windows" else "'"
+        inner = "'" if self.settings.os == "Windows" else '"'
```
wawanbreton, line 160: Is that for Mac ? Linux seems to support the 2 syntaxes. I will trust you if you just tell me that is IS required...

wawanbreton (commented):
Looks good, I'll make a build to check that it does work, at least for linux

conanfile.py
casperlamboo: apparently for mac it also works both ways, I'm going to change it to the windows case and if it installs correctly on my mahcine and it won't break the conan package it should work every where 👍

casperlamboo: https://github.com/Ultimaker/Cura/pull/17186/commits/51ab7409058fd38c4541064d93e87fd64902872f

PR #17240, opened 2023-11-06
Rename Method Lab extruder to LABS

wawanbreton:
Simply renaming extruder in config files. This apparently does not require an upgrade script as these files are not publicly released yet.


casperlamboo:
@pkuiper-ultimaker FYI

PR #17241, opened 2023-11-06
Center alligning in case of 3mf write

saumyaj3:
CURA-11166

This seems to be caused by the 3mf writer not saving it correctly. The calibration shape doesn’t seem to center the node correctly (eg; center of the mesh doesn’t align with the center of the model). This is then not correctly stored.


plugins/3MFReader/ThreeMFReader.py
```diff
@@ -233,8 +233,7 @@ def _read(self, file_name: str) -> Union[SceneNode, List[SceneNode]]:
                 if mesh_data is not None:
                     extents = mesh_data.getExtents()
                     if extents is not None:
-                        # We use a different coordinate space, so flip Z and Y
-                        center_vector = Vector(extents.center.x, extents.center.z, extents.center.y)
+                        center_vector = Vector(extents.center.x, extents.center.y, extents.center.z)
```
casperlamboo, line 236: I think we should still swap the coordinates since the 3mf spec defines a different coord system from the one we use.

casperlamboo: This was not an issue

casperlamboo (approved):

PR #17250, opened 2023-11-07
Cura 7647 play simulation time dependent

saumyaj3:
CURA-7647


plugins/SimulationView/SimulationViewMainComponent.qml
saumyaj3, line 1: This part doesn't work ATM. Haven't given much thought here. Suggestions are welcome

cura/LayerPolygon.py
```diff
@@ -186,6 +187,11 @@ def extruder(self):
     def types(self):
         return self._types
 
+    @property
+    def lineLengths(self):
```
jellespijker, line 191: How are you gonna treat the last (or initial) travel move in this regard?

cura/LayerPolygon.py
```diff
@@ -186,6 +187,11 @@ def extruder(self):
...
+        return [math.sqrt(sum((b - a) ** 2 for a, b in zip(self._data[i], self._data[i + 1])))
```
jellespijker, line 192: You might want to consider doing this with Numpy. That is probably faster then using `zip` and list comprehensions

plugins/SimulationView/SimulationView.py
```diff
@@ -400,6 +402,9 @@ def getMinFeedrate(self) -> float:
     def getMaxFeedrate(self) -> float:
         return self._max_feedrate
 
+    def getSimulationTime(self) -> list:
+        return [length / feedrate for length, feedrate in zip(self._visible_lengths[self._current_layer_num], self._current_feedrates[self._current_layer_num])]
```
jellespijker, line 406: See comment above about using Numpy instead of list comprehensions

jellespijker: Keep in mind that feedrate is probably in a non standard unit `mm/min` instead of `mm/s` It might be that it already converted to seconds, but double check if you're a factor 60 of

jellespijker: Also check if you're comparing the same arrays starting from the same offsets, since you calculate length using a sliding view

plugins/SimulationView/SimulationViewMainComponent.qml
jellespijker: Maybe using a Timer? Also not sure about this

jellespijker (changes requested):
Untested and it could be that your are calculating things correctly, but double check your units and if the feedrate that you're using is the same as the length you calculated.

Also check if Numpy will give you some speed benefits  over zip and sliding views, which I think they might

cura/LayerPolygon.py
saumyaj3: I am sorting it out w.r.t.  to visible indexs here: https://github.com/Ultimaker/Cura/blob/efda0aaba4329948485021bbbc4d2e365a5b63c4/plugins/SimulationView/SimulationView.py#L532

cura/LayerPolygon.py
saumyaj3: 457cecba5943c09b1b3b5b2a573a18c0915e0bc0

plugins/SimulationView/SimulationView.py
saumyaj3: yes I filter these arrays w.r.t. visible indexes

saumyaj3: I believe the feedate is already converted to seconds.  So the output here out of simulation time is in seconds

saumyaj3: 457cecba5943c09b1b3b5b2a573a18c0915e0bc0

cura/LayerPolygon.py
saumyaj3: Correction: All travel moves are calculated using 2 consecutive points, so if number of vertices in a layer is 701, the lengths covered will be 700.

plugins/SimulationView/SimulationView.py
```diff
@@ -535,6 +538,8 @@ def calculateColorSchemeLimits(self) -> None:
                     visible_feedrates = numpy.take(polyline.lineFeedrates, visible_indices)
                     visible_feedrates_with_extrusion = numpy.take(polyline.lineFeedrates, visible_indicies_with_extrusion)
                     self._current_feedrates[layer_index] = polyline.lineFeedrates
+                    # if len(polyline.lineLengths) > 0 and len(polyline.lineLengths) == len(polyline.lineFeedrates):
+                    #     self._simulation_time[layer_index] = polyline.lineLengths / polyline.lineFeedrates
```
casperlamboo, line 542: debug code?

casperlamboo (approved):

casperlamboo (approved):

PR #17264, opened 2023-11-08
[CURA-11300] Officially supported languages' translations for v5.6.x

rburema:
Due to various circumstances, an excel spreadsheet, the .tsv export thereof, and a temporary, thrown together python script where involved this time. Please double-check thoroughly.


resources/i18n/nl_NL/fdmprinter.def.json.po
```diff
@@ -2514,19 +2514,19 @@ msgstr "Acceleratie Primepijler"
 
 msgctxt "prime_tower_brim_enable label"
 msgid "Prime Tower Base"
-msgstr ""
+msgstr "Basis van de Prime Tower"
```
casperlamboo, line 2517: This isn't consistent with the previous translations, "Prime Tower" was always translated as "Primepijler". Although I do not like this translation I think we should keep consistent

```suggestion
msgstr "Basis van de Primepijler"
```

resources/i18n/nl_NL/fdmprinter.def.json.po
```diff
@@ -2514,19 +2514,19 @@ msgstr "Acceleratie Primepijler"
...
 
 msgctxt "prime_tower_base_curve_magnitude label"
 msgid "Prime Tower Base Curve Magnitude"
-msgstr ""
+msgstr "Krachtmagnitude van de basiscurve van de Prime Tower"
```
casperlamboo, line 2521:
```suggestion
msgstr "Krachtmagnitude van de basiscurve van de Primepijler"
```

resources/i18n/nl_NL/fdmprinter.def.json.po
```diff
@@ -2514,19 +2514,19 @@ msgstr "Acceleratie Primepijler"
...
 
 msgctxt "prime_tower_base_height label"
 msgid "Prime Tower Base Height"
-msgstr ""
+msgstr "Hoogte van de basis van de Prime Tower"
```
casperlamboo, line 2525:
```suggestion
msgstr "Hoogte van de basis van de Primepijler"
```

resources/i18n/nl_NL/fdmprinter.def.json.po
```diff
@@ -2514,19 +2514,19 @@ msgstr "Acceleratie Primepijler"
...
 
 msgctxt "prime_tower_base_size label"
 msgid "Prime Tower Base Size"
-msgstr ""
+msgstr "Grootte van de basis van de Prime Tower"
```
casperlamboo, line 2529:
```suggestion
msgstr "Grootte van de basis van de Primepijler"
```

resources/i18n/nl_NL/fdmprinter.def.json.po
```diff
@@ -2546,7 +2546,7 @@ msgstr "Minimumvolume primepijler"
 
 msgctxt "prime_tower_raft_base_line_spacing label"
 msgid "Prime Tower Raft Line Spacing"
-msgstr ""
+msgstr "Lijnafstand van het vlot van de Prime Tower"
```
casperlamboo, line 2549:
```suggestion
msgstr "Lijnafstand van het vlot van de Primepijler"
```

casperlamboo (changes requested):

casperlamboo (approved):

casperlamboo (commented):
I'm missing a few files,
- zh_TW/cura.po
- fi_FI/cura.po
- pt_BR/cura.po
- cs_CZ/cura.po

The first 2 are not important since they are not translated officially, but the last two might be of use for our translators

rburema:
None of those are officially translated -- at various points in time, they where all done by volunteer translators.

- For the 'Traditional Chinese' (not to be confused with 'Simplified Chinese', as used on the mainland), we had a volunteer, but I think they only properly responded to Ruben, I'm not sure if I ever received an answer when I wrote the request.
- The Finnish translation is the most out of date -- I'm not sure we ever did that particular one during the time we worked here.
- The Brazilian Portuguese one is done by a very enthusiastic volunteer (at the very least compared to the others...)
- I'm not sure what the status of the Czech one is. I think it might have been a recent addition (or recent resurgence), but I don't think we sent much in their direction for a few versions. (So this one might be on us/me...) -- Seeing them listed in this manner makes that more clear to me. I'll go see if I can clear that up next week.

PR #17272, opened 2023-11-09
[CURA-11293] fix disallowed areas

wawanbreton:
Fixed machine disallowed areas being displayed wrong when the described polygons go around the build plate.

CURA-11293


cura/BuildVolume.py
```diff
@@ -489,20 +489,21 @@ def _buildDisallowedAreaMesh(self, min_w: float, max_w: float, min_h: float, max
...
             return None
 
+        bounding_box = Polygon(numpy.array([[min_w, min_d], [min_w, max_d], [max_w, max_d], [max_w, min_d]], numpy.float32))
+
         mb = MeshBuilder()
         color = self._disallowed_area_color
         for polygon in self._disallowed_areas:
-            points = polygon.getPoints()
-            if len(points) == 0:
+            intersection = polygon.intersectionConvexHulls(bounding_box)
```
casperlamboo, line 497: Why do we do the intersection over the convex hulls of the polygons? If we have a non-convex disallowed area, is it acceptable that the shape becomes convex?

cura/BuildVolume.py
```diff
@@ -489,20 +489,21 @@ def _buildDisallowedAreaMesh(self, min_w: float, max_w: float, min_h: float, max
...
+
+            points = intersection.getPoints()[::-1]
```
casperlamboo, line 499: Think this notation is slightly more readable
```suggestion
            points = intersection.getPoints()
            points.reverse()
```

cura/BuildVolume.py
casperlamboo: I realize the way we generate the mesh ([by generating a fan](https://stackoverflow.com/questions/8043923/gl-triangle-fan-explanation)) doesn't allow for arbitrary polygon triangulation. We could do this by using the [`triangulate_polygon` utility from try `tri_mesh`](https://trimesh.org/trimesh.creation.html#trimesh.creation.triangulate_polygon).

wawanbreton: > Why do we do the intersection over the convex hulls of the polygons? If we have a non-convex disallowed area, is it acceptable that the shape becomes convex?

Because I didn't find an existing method to do it :laughing: but in cases the disallowed area has holes, this is wrong indeed. Is there a way to do it in front-end side ? Or is it to be implemented ?



> I realize the way we generate the mesh ([by generating a fan](https://stackoverflow.com/questions/8043923/gl-triangle-fan-explanation)) doesn't allow for arbitrary polygon triangulation. We could do this by using the [`triangulate_polygon` utility from try `tri_mesh`](https://trimesh.org/trimesh.creation.html#trimesh.creation.triangulate_polygon).

I just re-used the actual method. It could go wrong with very strange disallowed shape areas indeed, but all the actual ones I could see are quite simple.

cura/BuildVolume.py
wawanbreton: It is, but `points` is a `numpy.array` so it does not have a `reverse` method. I'll see if there is some similar thing.

cura/BuildVolume.py
```diff
@@ -494,7 +494,7 @@ def _buildDisallowedAreaMesh(self, min_w: float, max_w: float, min_h: float, max
         mb = MeshBuilder()
         color = self._disallowed_area_color
         for polygon in self._disallowed_areas:
-            intersection = polygon.intersectionConvexHulls(bounding_box)
+            intersection = polygon.intersection(bounding_box)
```
casperlamboo, line 497: If we swap out `intersectionConvexHulls` for the normal `intersection` then we should also update our triangulation strategy

casperlamboo (approved):

PR #17334, opened 2023-11-16
Adding plugin to allow changing of tools at any layer(s)

pearmid:
# Description

This feature allows a user to specify a list of layers on which to change tool head for IDEX printers.
this has the same effect as FilamentChange on IDEX printers without the need for manual intervention

## Type of change

- [ ] Translations

# How Has This Been Tested?

Locally added to Cura 5.5.5 and tested both visual inspection of GCode in addition to real printer testing

**Test Configuration**:
* Operating System: Windows

# Checklist:

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/ToolChange.py
+++ b/plugins/PostProcessingPlugin/scripts/ToolChange.py
@@ -1,0 +1,143 @@
+# Copyright (c) 2023 Jody Pearson.
+# Released under the terms of the LGPLv3 or higher.
+
+from typing import List
+from ..Script import Script
+
+class ToolChange(Script):
+    def __init__(self):
+        super().__init__()
+
+    def getSettingDataString(self):
+        return """{
+            "name": "Tool Change",
+            "key": "ToolChange",
+            "metadata": {},
+            "version": 2,
+            "settings":
+            {
+                "enabled":
+                {
+                    "label": "Enable",
+                    "description": "Uncheck to temporarily disable this feature.",
+                    "type": "bool",
+                    "default_value": true
+                },
+                "layer_number":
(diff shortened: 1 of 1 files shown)


GregValiant:
A few thoughts...
Your script requires that G60 and G61 are enabled in the firmware.  I have no idea how many firmware flavors support them as they are memory hogs.  My older Marlin firmware does not (it would be nice if it did).
There is no "block" to keep the script from running if the machine only has one extruder.
The move "G1 X-50" is only possible if the machine is an "Origin at Center" model.  All others may well crash into the X endstop switch.
The tool change code is being added in front of the layer.  That means the first line of the layer is no longer "LAYER:".  It's possible that following post processors could fail if they count on "LAYER" being the first line of a data[item].
There is no retraction or priming.
The "E" of the new extruder is being set to zero and is not synced to the position in the file.  That's fine if the file is sliced in Relative Extrusion mode, but there isn't a check for that.  The result could be a huge extrusion from 0 to whatever the E location is in the file when the print resumes.
It isn't clear that the before and after gcodes must be single commands rather than multi-line.
If the user has Start and End gcodes for the extruders - they don't get called or added.
There is no consideration of One-at-a-Time mode.
If there is an offset between the extruders then simply switching to the new tool will cause a layer shift by the amount of the offset.
 
This sort of tool change can currently be accomplished with a support blocker configured as a cutting mesh and set to the extruder-of-choice.  That has the advantage that Cura takes all the above into account when it gets to the height of the "Modifier Mesh".  Prime towers can also be utilized.

GregValiant:
This is something that can currently be done using "Insert at Layer Change".
I'll go ahead and close this.

PR #17338, opened 2023-11-16
Cura 11347

casperlamboo:
# Description

With the previous implementation of the formula's in start/end gcode templates we made some oversights. With that implementation the following template (valid) would _not_ be accepted
```
; {filament_cost}, {print_time}, {filament_amount}, {filament_weight}, {jobname} <- post slice variables are not availabe in templates
; {3.14} decimal would be interpreted as object attribute retrieval, `14` is not an attribute of the number `3`
; {any([True])}  function calls would be parsed, and each function-part would be retrieved and resolved individually, instead we would like to interpret them as a whole
```
There is one asterix attached to the current implementation; that is that post slice variables won't work with formula's. This is because the templates are resolved in a two step process. First all template strings are resolved execpt for a string that exactly matches a variable name from the post slice variables (`filament_cost`, `print_time`, `filament_amount`, `filament_weight`, `jobname`). Then after slicing is complete these post slice variables are substituted.

We cannot substitute formula's post slicing since we have lost information about start/end code. We loose information regarding
- the type of gcode (machine or extruder)
- the extruder nr the start/end code belongs to
And vice versa we cannot use the post slice variables before slicing since these variables are generated during the slicing (as the name suggests)

Don't think this is an issue since we only re-introduced this feature in order to not regress compared to previous cura versions. Entering formula's in this version

CURA-11347

## Type of change

- [ ] Translations

# How Has This Been Tested?

Yess

**Test Configuration**:
* Operating System: MacOS

# Checklist:

### DIFF
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -6,7 +6,7 @@ import numpy
 from string import Formatter
 from enum import IntEnum
 import time
-from typing import Any, cast, Dict, List, Optional, Set
+from typing import Any, cast, Dict, List, Optional, Set, Tuple
 import re
 import pyArcus as Arcus  # For typing.
 from PyQt6.QtCore import QCoreApplication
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -102,6 +118,9 @@ def get_value(self, expression: str, args: [str], kwargs: dict) -> str:
         setting_function = SettingFunction(expression)
         value = setting_function(container_stack, additional_variables=additional_variables)
 
+        print("value", value)
+        print("expression", expression)
```
nallath, line 122:
```suggestion
```

Looks like debug prints to me, so we should probably remove em.

nallath (changes requested):

plugins/CuraEngineBackend/StartSliceJob.py
casperlamboo: oops

rburema (approved):

PR #17394, opened 2023-11-21
CURA-11364 add sentry for engine

nallath:
Add a preference to disable sentry crash sending for the engine

CURA-11364

https://github.com/Ultimaker/CuraEngine/pull/1985


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -1088,11 +1094,17 @@ def disableTimer(self) -> None:
...
+        if preference != "general/auto_slice" and preference != "info/send_engine_crash":
             return
-        auto_slice = self.determineAutoSlicing()
-        if auto_slice:
-            self._change_timer.start()
+        if preference == "general/auto_slice":
+            auto_slice = self.determineAutoSlicing()
+            if auto_slice:
+                self._change_timer.start()
+        elif preference == "info/send_engine_crash":
```
casperlamboo, line 1103: Why is this in a `elif`? We don't want to send sentry data when auto slice is enabled?

fieldOfView: This is about handing a change in preference. If the preference that is being changed is anything but "info/send_engine_crash" we don't want to change this particular behavior.

casperlamboo: ah, right. read too fast

plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -1088,11 +1094,17 @@ def disableTimer(self) -> None:
...
+            if CuraApplication.getInstance().getPreferences().getValue("info/send_engine_crash"):
+                os.environ["use_sentry"] = str(1)
+            else:
+                os.environ["use_sentry"] = str(0)
```
casperlamboo, line 1107: Think this notation is cleaner for this instance
```suggestion
            os.environ["use_sentry"] = "1" if CuraApplication.getInstance().getPreferences().getValue("info/send_engine_crash") else "0"
```

PR #17397, opened 2023-11-21
Auto enable interlocking

pkuiper-ultimaker:
Enable interlocking when more then 1 material is used and that non of the materials is a support material and the two model materials are different. 

PP-395


resources/definitions/fdmprinter.def.json
```diff
@@ -76,6 +76,14 @@
                     "type": "str",
                     "enabled": false
                 },
+                "material_type":
+                {
+                    "label": "Material Type",
+                    "description": "The type of material used.",
+                    "default_value": "",
+                    "type": "str",
```
casperlamboo, line 84: can we make this an enum instead? We will only have `model` and `support` right?

Frederic98: The `material_type` is meant as "polymer type", not build vs support.  
So, whatever is in the `<name><material>...</material></name>` tag of the fdm_material.

Maybe we should come up with a better name for it...?

PR #17421, opened 2023-11-24
[CURA-11379] Fix Prime-Tower position for printers with explicit disallowed areas.

rburema:
We tried to to this entirely by formulas (and one other improvement is added to that in this PR), but the disallowed areas (`machine_disallowed_areas` and `nozzle_disallowed_areas`) can be of arbitrary (well, polygonal) shape, so adding those _properly_ will require a feature.

Yet this needs to be fixed in the current release-branch. Since there are only a few printers where this is an actual problem, just fix them manually for now.


resources/definitions/fdmprinter.def.json
```diff
@@ -6640,7 +6640,7 @@
...
+                    "value": "(resolveOrValue('machine_width') / 2 + resolveOrValue('prime_tower_size') / 2) if resolveOrValue('machine_shape') == 'elliptic' else (resolveOrValue('machine_width') - (resolveOrValue('prime_tower_base_size') if (resolveOrValue('adhesion_type') == 'raft' or resolveOrValue('prime_tower_brim_enable')) else 0) - max(max(extruderValues('travel_avoid_distance')) + max(extruderValues('machine_nozzle_offset_x')) + max(extruderValues('support_offset')) + (extruderValue(skirt_brim_extruder_nr, 'skirt_brim_line_width') * extruderValue(skirt_brim_extruder_nr, 'skirt_line_count') * extruderValue(skirt_brim_extruder_nr, 'initial_layer_line_width_factor') / 100 + extruderValue(skirt_brim_extruder_nr, 'skirt_gap') if resolveOrValue('adhesion_type') == 'skirt' else 0) + (resolveOrValue('draft_shield_dist') if resolveOrValue('draft_shield_enabled') else 0), max(map(abs, extruderValues('machine_nozzle_offset_x'))), 1)) - (resolveOrValue('machine_width') / 2 if resolveOrValue('machine_center_is_zero') else 0)",
```
casperlamboo, line 6643: this "value" is stupendously large. Let's quickly resolve this by implementing the prime tower "resolve" function

casperlamboo (approved):

resources/definitions/fdmprinter.def.json
rburema: I was just thinking of consolidating a few parts into larger functions that could in theory be used elsewhere -- but what you said makes more sense, really.

PR #17518, opened 2023-12-04
Model translations while changing build plates

saumyaj3:
while importing models from a 3mf file, coordinate translations shouldn’t matter and the object should be organized on the plate where it finds space.

CURA-11166


cura/CuraApplication.py
```diff
@@ -179,6 +179,7 @@ def __init__(self, *args, **kwargs):
         self._use_single_instance = False
 
         self._single_instance = None
+        self._project_mode = None
```
nallath, line 182:
```suggestion
        self._project_mode: Optional[str] = None
```

Nitpicking, but it's better to type this. Perhaps we should also rename this to "open_project_mode" as that's a bit more verbose. The current name implied to me that it's a boolean (eg "we are in project mode").

cura/CuraApplication.py
```diff
@@ -1999,8 +2000,11 @@ def _readMeshFinished(self, job):
                         center_y = 0
 
                     node.translate(Vector(0, center_y, 0))
-
                     nodes_to_arrange.append(node)
+            # If file extention is 3mf and models are to be loaded from a cura project,
```
nallath, line 2004: This isn't entirely true. Currently we only have project files that are 3mf, but that is not a given. 

It's more correct to say "If the file is a project, but you want to open as a model, the files should be arranged" (so basically removing any reference to 3MF here).

nallath (changes requested):

cura/CuraApplication.py
saumyaj3: fixed 30e4c61620fc31703cf0b4227f419beea882db30

cura/CuraApplication.py
saumyaj3: fixed 30e4c61620fc31703cf0b4227f419beea882db30

PR #17527, opened 2023-12-05
[CURA-11279] Fix arrange in grid

wawanbreton:
I didn't get to the very root cause of the issue, but using absolute coordinates when doing a grid placement fixes it.

### DIFF
--- a/cura/Arranging/GridArrange.py
+++ b/cura/Arranging/GridArrange.py
@@ -242,13 +242,9 @@ class GridArrange(Arranger):
         center_grid_y = coord_grid_y + (0.5 * self._grid_height)
 
         bounding_box = node.getBoundingBox()
-        center_node_x = (bounding_box.left + bounding_box.right) * 0.5
-        center_node_y = (bounding_box.back + bounding_box.front) * 0.5
+        center_node_z = bounding_box.height / 2
 
-        delta_x = center_grid_x - center_node_x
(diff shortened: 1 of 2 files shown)


casperlamboo (approved):

cura/Arranging/GridArrange.py
```diff
@@ -242,13 +242,9 @@ def _moveNodeOnGrid(self, node: "SceneNode", grid_x: int, grid_y: int) -> "Opera
...
         bounding_box = node.getBoundingBox()
-        center_node_x = (bounding_box.left + bounding_box.right) * 0.5
-        center_node_y = (bounding_box.back + bounding_box.front) * 0.5
+        center_node_z = bounding_box.height / 2
 
-        delta_x = center_grid_x - center_node_x
-        delta_y = center_grid_y - center_node_y
-
-        return TranslateOperation(node, Vector(delta_x, 0, delta_y))
+        return TranslateOperation(node, Vector(center_grid_x, center_node_z, center_grid_y), True)
```
casperlamboo, line 247: Nice! didn't know we could set absolute coords; would have done this if I knew

One small nitpick, can we add a named argument for better code readability (so we know what the boolean refers to)?
```suggestion
        return TranslateOperation(node, Vector(center_grid_x, center_node_z, center_grid_y), set_position = True)
```

PR #17547, opened 2023-12-07
Add url scheme support for CURA

casperlamboo:
# Description

This PR add support for url schemes. 

The following url schemes are supported: 
```
cura://
slicer://
```
There is an query-api included with this PR, that at the moment only includes the following end point:
```
cura://open?file=[encode-file-url]
```
When opening cura with this endpoint the include file will be downloaded, and loaded into cura

Cura-11288

## Type of change

- [ ] Translations

# How Has This Been Tested?

yis

**Test Configuration**:
* Operating System: macOS 13.3

# Checklist:


jellespijker (changes requested):
Looks good but my main concern is that I think we don't clean up the registry if we uninstall the NSIS installed Cura version.

The msi should be okay, since that installer keeps track of what's added during the installation process

PR #17705, opened 2023-12-19
Improve oauth

wawanbreton:
Hi, this is the community version of myself :)

I made a few changes to Cura, in order to reuse the OAuth workflow implementation currently in use for Digital Factory, for a plugin which requires OAuth. This allows me to use the actual code and not add dependancies to the plugin.

The PR contains 3 commits:
* https://github.com/Ultimaker/Cura/commit/3c86bf140ceba3e6830b0d483fdba5d17857e852: The OAuth protocol may use a "client secret" with the "client id", but this is currently not used for DF. This commits adds support to use a "client secret".
* https://github.com/Ultimaker/Cura/commit/aac9af577bf6895665eb34dcae4e7fc5967caeff: The `AuthorizationService` class uses signals implemented with `UM.Signal`. The documentation clearly indicates that they should be instanciated by object instance, and not by class (as opposite to Qt signals). The current implementation instanciates them by class, so the signals are emitted to all listeners of all instances of `AuthorizationService`. Thus, if there are two OAuth workflows, they will both receive signals of each other, which creates unexpected behaviors.
* https://github.com/Ultimaker/Cura/commit/b794989468653d04c3e0e24044cb8ea61ce29dff: The `AuthorizationService` class handles the OAuth workflow, then once the authentication token has been received, it asks the server for user profile, which is specific to DF protocol. This commit allows the service to stop once the token has been received. This is clearly not the cleanest way of doing it. The user profile retriving should be done in a specific service implementation, or somewhere else. This is the solution I came up with which is very simple and does not break the API, but this is clearly open to discussion.


wawanbreton:
Hey @nallath could I have your opinion on this PR ?

cura/OAuth2/AuthorizationService.py
```diff
@@ -31,7 +31,10 @@ class AuthorizationService:
     account information.
     """
 
-    def __init__(self, settings: "OAuth2Settings", preferences: Optional["Preferences"] = None) -> None:
+    def __init__(self,
+                 settings: "OAuth2Settings",
+                 preferences: Optional["Preferences"] = None,
+                 get_user_profile: bool = True) -> None:
```
rburema, line 37: I'd prefer if this variable wasn't a boolean, but a callable, defaulted to `self._getUserProfile`, which gives a lot more versatility against almost no cost. If it's not needed, just supply an empty method instead.

(Note that the optional part of the signature of the 'getUserProfile' function doesn't matter much, since when we need the proposed callback, it's called without arguments (other than self).)

wawanbreton: I did my best but it is not possible to have `self.getUserProfile` as default value because `self` is not defined at this point. So giving `None` gives you the default value and otherwise you specify your callback.

rburema: Alright, then I think we can just merge this.

rburema:
I needed to revert my proposed changes in main, because I didn't see you made them here 😅

PR #17718, opened 2023-12-20
Cura 7647 proof of concept

saumyaj3:
Another concept for simulation being time dependent. A frame rate is defined here for toolpaths to move.
CURA-7647


plugins/SimulationView/SimulationView.py
```diff
@@ -360,6 +364,9 @@ def setPath(self, value: float) -> None:
         if self._current_path_num != value:
             self._current_path_num = min(max(value, 0), self._max_paths)
             self._minimum_path_num = min(self._minimum_path_num, self._current_path_num)
+            # update _current time when the path is changed by user
+            if self.getCurrentLayer() in self._cumulative_line_duration  and self._current_path_num < self._max_paths and round(self._current_path_num)== self._current_path_num:
+                self._current_time = self._cumulative_line_duration[self.getCurrentLayer()][int(self._current_path_num)]
```
casperlamboo, line 368: Why only do this when `self.getCurrentLayer() in self._cumulative_line_duration` I would argue, if `_cumulative_line_duration`
```suggestion
            comulative_line_duration = self.comunativeLineDuration()
            self._current_time = comulative_line_duration[int(self._current_path_num)]
```

plugins/SimulationView/SimulationView.py
```diff
@@ -244,15 +243,20 @@ def advanceTime(self, time_increase: float) -> bool:
...
+            total_duration = 0.0
+            for polyline in self.getLayerData().polygons:
+                for line_duration in list((polyline.lineLengths / polyline.lineFeedrates)[0]):
+                    total_duration += line_duration / SimulationView.SIMULATION_FACTOR
+                    self._cumulative_line_duration[self.getCurrentLayer()].append(total_duration)
+
+        # Calculate the total duration using numpy.sum
+        total_duration = (self._cumulative_line_duration[self.getCurrentLayer()][-1])
+
+        return total_duration
```
casperlamboo, line 258: Why return the `total_duration` in a function with the name `cumulativeLineDuration`

I would return the communative duration here
```suggestion
        return self._cumulative_line_duration[self.getCurrentLayer()]
```

And then where it is used also get the `total_duration`
```python
cumulative_line_duration = self.cumulativeLineDuration()
total_duration = cumulative_line_duration[-1]
```

casperlamboo (commented):
Had some remarks and additionally, I have some questions regarding caching. What will happen if a new model is loaded, all caches should be invalidated at that point. Does that happen?

And question number two regarding the caching, this is a bit of a lesser point. I see that caches are store for every layer individually and when a layer is finished that layer data is still stored in cache. This seems a bit wastefull as you not often going back to the same layer. Why not only cache a single layer and updating the cache on a layer change?

plugins/SimulationView/SimulationView.py
saumyaj3: fixed in 0de4f612b399cfe8b7a6c8059e9579489bcd739c

saumyaj3:
The cache is cleared everytime a new model is loaded (or here sliced). Also cache is reloaded when layer is changed.

plugins/SimulationView/SimulationView.py
```diff
@@ -242,20 +243,19 @@ def advanceTime(self, time_increase: float) -> bool:
             self.setTime(self._current_time + time_increase)
         return True
 
-    def cumulativeLineDuration(self) -> float:
+    def cumulativeLineDuration(self) -> list:
```
casperlamboo, line 246:
```suggestion
    def cumulativeLineDuration(self) -> List[float]:
```

casperlamboo (approved):

casperlamboo (approved):

PR #17730, opened 2023-12-22
Property of machinemanager redefined

saumyaj3:
CURA-11459

Due to pyqt update to 6.6. The active machine was shown blank in the UI. 
Provides a fix for that


casperlamboo:
Looks good 👍

PR #17754, opened 2023-12-23
Add RatRig printers configurations

Nu-hin:
# Description

Added configurations for the following RatRig printers:

* RatRig V-Core 3 (200mm, 300mm, 400mm, and 500mm versions)
* RatRig V-Minion

The configurations are compatible with both V-Core 3 and 3.1 versions.

The bed meshes are taken from https://github.com/slic3r/slic3r-profiles with kind permission of the author @HelgeKeck

## Type of change


Nu-hin:
Pipeline fails with 

```
Run actions-ecosystem/action-add-labels@v1
  with:
    labels: PR: Community Contribution :crown:
    github_token: ***
    repo: Ultimaker/Cura
  env:
    CONAN_LOGIN_USERNAME: ***
    CONAN_PASSWORD: ***
Error: HttpError: Resource not accessible by integration
Error: Resource not accessible by integration
```

Seems like it cannot assign labels to the issue, not sure if I can do anything about that.

nallath (approved):

HellAholic:
Hey @Nu-hin,
I see some artifacts on the print bed for one of the definitions
image
In other models they're used in the boundaries/support of the build plate
image
Also for both models (V-Core 3 200mm) and (V-Minion) the printer background seems to be loaded on top of the build plate rather than under it.
image
image

Nu-hin:
It seems like for V-Minion and V-Core 200mm the origin is set incorrectly (below the bed). I will try to fix it ASAP. The artifacts on the V-Minion mesh are, actually, not artifacts, but are indeed the screw holes.

HelgeKeck:
screw holes are partly underneath the build plate

Nu-hin:
> screw holes are partly underneath the build plate

That's because Cura renders with perspective, not orthogonal projection. I rotated V-Minion in the Cura and it seems fine. The only problem is offset.

Nu-hin:
@HellAholic I have found the issue, and submitted #18218 to fix. Thank you very much for finding and reporting the problem.

PR #17870, opened 2024-01-02
.mo files created when they are not present

saumyaj3:
CURA-11497

Switching pf languages was not working. this is a workaround for it


casperlamboo:
This seems a bit double, there is already a conan workflow that should handle this:
https://github.com/Ultimaker/Uranium/blob/a0abe007c51b12862aacf2afa1c9e1e0aa124fdd/conanfile.py#L136-L141
The code is only executed when the [`enable_i18n` flag is provided](https://github.com/Ultimaker/Uranium/blob/a0abe007c51b12862aacf2afa1c9e1e0aa124fdd/conanfile.py#L135). I think we should dive in the workflows, and find out why the flag isn't present.

saumyaj3:
> This seems a bit double, there is already a conan workflow that should handle this: https://github.com/Ultimaker/Uranium/blob/a0abe007c51b12862aacf2afa1c9e1e0aa124fdd/conanfile.py#L136-L141 The code is only executed when the [`enable_i18n` flag is provided](https://github.com/Ultimaker/Uranium/blob/a0abe007c51b12862aacf2afa1c9e1e0aa124fdd/conanfile.py#L135). I think we should dive in the workflows (they are probably relocated here: https://github.com/Ultimaker/cura-workflows/), and find out why the flag isn't present.

I tried building the Cura after removing that flag too.. I still didn't get the .mo files.

PR #17876, opened 2024-01-03
High speed profiles for the layer heights <0.15mm.

pkuiper-ultimaker:
High speed profiles for the layer heights <0.15mm for Ultimaker PLA, T-PLA, ABS and PETG. 
Also removed the constrain that support z gaps need to be multiples of the layer height.

PP-391


wawanbreton (commented):
I see 2 files with layer height 0.2mm with big changes, is that on purpose ?
The changes for the 0.3mm are ok because that's about simplifying them with the now available fractional layer height.

pkuiper-ultimaker:
> I see 2 files with layer height 0.2mm with big changes, is that on purpose ? The changes for the 0.3mm are ok because that's about simplifying them with the now available fractional layer height.

These are from the CC core. I will play it safe and not make those high speed profiles. I will adjust the PR accordingly.

PR #17877, opened 2024-01-03
qt6.6 scroll issues

saumyaj3:
CURA-11468

# Description

qt6.6 changelog: 06540cdb02 Flickable: Proportional wheel scrolling if deceleration is
large
When using a plain "clicky" mouse wheel, Flickable has historically
applied acceleration: the faster you rotate the wheel, the faster it
flicked. The default value of flickDeceleration was 5000, which felt
hard to control, but changing it in QML affected touch-flicking and
wheel behavior simultaneously. So now we decouple those:
flickDeceleration only affects touchscreen flicking, and Flickable
rather uses a new default value of 15000, which is defined to mean that
scroll distance becomes directly proportional to
QWheelEvent::angleDelta() * QStyleHints::wheelScrollLines() (where a
"line" is 24 logical pixels, because Flickable does not have direct
understanding of its contents). You can set the new environment var
QT_QUICK_FLICKABLE_WHEEL_DECELERATION to a value from 1 to 14999 to
restore wheel acceleration behavior from older versions; 5000 was the
old value.

This made scrolling behave abnormally than the previous behavior. The environment variable is now set so that it follows old behavior.

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -17,6 +17,7 @@ import faulthandler
 import os
 if sys.platform != "linux":  # Turns out the Linux build _does_ use this, but we're not making an Enterprise release for that system anyway.
     os.environ["QT_PLUGIN_PATH"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
+    os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
     os.environ["QML2_IMPORT_PATH"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
     os.environ["QT_OPENGL_DLL"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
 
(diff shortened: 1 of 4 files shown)


cura_app.py
```diff
@@ -17,6 +17,7 @@
 import os
 if sys.platform != "linux":  # Turns out the Linux build _does_ use this, but we're not making an Enterprise release for that system anyway.
     os.environ["QT_PLUGIN_PATH"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
+    os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"] = ""  # Security workaround: Don't need it, and introduces an attack vector, so set to nul.
```
casperlamboo, line 20:
```suggestion
    os.environ["QT_PLUGIN_PATH"
    try:
        # try converting to integer
        os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"] = int(os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"])
    except ValueError:
        os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"] = "5000"
```

PR #17878, opened 2024-01-03
Printer name changed to printer type

saumyaj3:
printer names can be different for same type of printer. Also, printer type is coming from cloud here so the type search is changed accordingly

CURA-11432


casperlamboo:
at one point it was already something similar to this: https://github.com/Ultimaker/Cura/blame/64cb060bd1c2ec23d498c81ea50a55957dfad3e6/plugins/UM3NetworkPrinting/src/Cloud/CloudOutputDevice.py#L334. Think the changes make sense

casperlamboo (approved):

casperlamboo (approved):

PR #15853, opened 2023-06-22
Create AddCoolingProfile.py

GregValiant:
Advanced fan control post-processor.  "By Layer" or "By Feature" for up to 4 fan circuits.  This is in response to numerous requests such as #2149 , #5140 , #8527, #8777, #3624
September 10 commit adds an exit if the "print_sequence" is set to "One at a Time" in Cura.
September 14 commit removes the exit code and allows use with "One at a Time".

# Description

A new post-processor to control layer cooling fans.

This fixes... OR This improves... -->

## Type of change

- [ X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ X] Test A
- Tested with Cura 4.x and Cura 5.x

**Test Configuration**:
* Operating System:  Win 10 Home,  Ender 3 Pro with Marlin firmware.

# Checklist:

- [ X] My code follows the style guidelines of this project as described in [UltiMaker Meta](https://github.com/Ultimaker/Meta) and [Cura QML best practices](https://github.com/Ultimaker/Cura/wiki/QML-Best-Practices)
- [ X] I have read the [Contribution guide](https://github.com/Ultimaker/Cura/blob/main/contributing.md) 
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] I have uploaded any files required to test this change

PR #17926, opened 2024-01-07
Add 'Advanced Cooling Fan Control' post-processing script.

rburema:
See discussions in #15853 (which was closed, for which this is a replacement) and (internally) tracking-ticket CURA-11520.

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
+++ b/plugins/PostProcessingPlugin/scripts/AddCoolingProfile.py
@@ -1,0 +1,879 @@
+# January 2023 by GregValiant (Greg Foresi).
+# Functions:
+#    Remove all fan speed lines from the file (optional).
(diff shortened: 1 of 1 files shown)


rburema:
There was a little bug where if you added this post-script, then switched printers, and then added this post-script to the other printer as well, Cura would crash. It was easy to fix, so I did that.

The other code is unchanged from what @GregValiant sent me in the final discussion in #15853

rburema:
@GregValiant Just to be sure: I never mean the comments I write as 'do this or else'. No matter how terse they are, there is almost always rooms for discussion.

(By the way; I'm also reducing the number of hours I work at UM, and I still have to discuss with my leads/colleagues what that means for how much time I'll spend publicly working on things like community PR's and the like.)

In any cases, best whishes' for the new year!

GregValiant:
@rburema 
Did you get the latest and greatest from my last post in #15853 ?  I had made changes that included adding more "lists" as you had requested.  I also fixed a minor bug in the regex replacements.

PR #17946, opened 2024-01-08
Add/reorganize settings for raft walls

wawanbreton:
* Add settings `raft_wall_count`, `raft_interface_wall_count` and `raft_surface_wall_count`
* Move setting `raft_base_wall_count` to child of `raft_wall_count`

Default values for interface and surface layers are 0 to keep retro-compatibility. Value for base layer is linked to the parent.

Contributes to CURA-11228

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -6329,6 +6329,54 @@
                     "settable_per_extruder": true,
                     "limit_to_extruder": "raft_base_extruder_nr"
                 },
+                "raft_wall_count":
+                {
+                    "label": "Raft Wall Count",
+                    "description": "The number of contours to print around the linear pattern of the raft.",
+                    "type": "int",
+                    "default_value": 1,
+                    "minimum_value": "0",
+                    "enabled": "resolveOrValue('adhesion_type') == 'raft'",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": false,
+                    "children":
+                    {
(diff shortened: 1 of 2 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -6329,6 +6329,54 @@
...
+                            "label": "Raft Middle Wall Count",
+                            "description": "The number of contours to print around the linear pattern in the middle layers of the raft.",
+                            "type": "int",
+                            "default_value": 0,
+                            "minimum_value": "0",
+                            "enabled": "resolveOrValue('adhesion_type') == 'raft'",
+                            "settable_per_mesh": false,
+                            "settable_per_extruder": false
+                        },
+                        "raft_base_wall_count":
```
casperlamboo, line 6366: Can we move this setting to the top, I think it makes sense that these settings are ordered in the same order in which they are printed.

wawanbreton: There are other settings for "Print Speed", "Acceleration", "Jerk" and "Fan Speed" and they are all on the same orgnization, which I matched, so I think it makes sense to keep it consistent (altough I would have preferred it the other way indeed)

casperlamboo: Let's flip the order of those aswell, I remember implementing those. If they are in this order I would consider that a mistake on my part.

wawanbreton: Ok, that's just visual so it can be done

PR #17954, opened 2024-01-09
CURA-7812-decrease-Qml-singletons

saumyaj3:
#Description
CURA-7812

### DIFF
--- a/resources/qml/Settings/SettingTextField.qml
+++ b/resources/qml/Settings/SettingTextField.qml
@@ -223,7 +223,7 @@ SettingItem
 
                 cursorShape: Qt.IBeamCursor
 
-                onPressed: {
+                onPressed:(mouse)=> {
                     if (!input.activeFocus)
                     {
                         base.focusGainedByClick = true
--- a/cura/CuraActions.py
+++ b/cura/CuraActions.py
@@ -3,10 +3,11 @@
 
 from typing import List, cast
 
-from PyQt6.QtCore import QObject, QUrl, QMimeData
+from PyQt6.QtCore import QObject, QUrl, pyqtSignal, pyqtProperty
 from PyQt6.QtGui import QDesktopServices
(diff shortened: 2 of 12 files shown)


resources/qml/Settings/SettingTextField.qml
```diff
@@ -223,7 +223,7 @@ SettingItem
 
                 cursorShape: Qt.IBeamCursor
 
-                onPressed: {
+                onPressed:(mouse)=> {
```
wawanbreton, line 226: I never saw this QML syntax :smile: what does it do ?

wawanbreton (commented):
I'm a bit disturbed by the fact of adding deprecated functions, this is quite a non-sense :smile: However I understand that deprecating a singletonization is not really possible.
On the other hand, I see no warning pop up when I run your code. What is the actual effect of the `deprecated` decorator ? Is it just a reminder for ourselves ?

saumyaj3:
> I'm a bit disturbed by the fact of adding deprecated functions, this is quite a non-sense 😄 However I understand that deprecating a singletonization is not really possible. On the other hand, I see no warning pop up when I run your code. What is the actual effect of the `deprecated` decorator ? Is it just a reminder for ourselves ?

If you go to qml and use one of the registersingletons, the warnings appear. At the moment I removed all the registerSingletons used in Cura. Replaced it with what we desire. They are not used in any part of our code. but you can use them with the @deprecated warning.

resources/qml/Settings/SettingTextField.qml
wawanbreton: Oh yes indeed, because `mouse` is used implicitely inside the function, so that makes it explicit. Good to know :smiley: :+1:

wawanbreton:
Ok, so the warning don't show up for us, but they will if someone is using code calling them, that's perfect :+1:

PR #17955, opened 2024-01-09
Create AddCuraSettings.py

GregValiant:
Adds most of the Cura project settings to the end of the gcode file.

# Description

This adds about 400 categorized settings to the end of the gcode file.  Options include "All" or "By section".  Additional options are the calculated Min and Max print speeds and some file information added to the opening paragraph of the gcode.

## Type of change

# How Has This Been Tested?

- [ X] Debugging in current and previous Cura versions.

**Test Configuration**:
* Operating System:  Windows 10 Pro

# Checklist:

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/AddCuraSettings.py
+++ b/plugins/PostProcessingPlugin/scripts/AddCuraSettings.py
@@ -1,0 +1,844 @@
+# Copyright (c) 2023 GregValiant (Greg Foresi)
+##   This post processor adds most of the Cura settings (~400) to the end of the Gcode file.  Which settings are added depends on things like the Extruder Count, Cura setup, etc.
+## My thanks to Aldo Hoeben who pointed out how to get the 'currency symbol' from Cura.  It was the icing on the cake.
+
+from UM.Application import Application
+import UM.Util
+from ..Script import Script
+import time
+import re
+from UM.Qt.Duration import DurationFormat
+import configparser
+from UM.Preferences import Preferences
+from UM.Message import Message
+
+class AddCuraSettings(Script):
+    """Add the Cura settings as a post-script to the g-code.
+    """
+
+    def getSettingDataString(self):
+        return """{
+            "name": "Add Cura Settings",
+            "key": "AddCuraSettings",
+            "metadata": {},
+            "version": 2,
+            "settings":
+            {
+                "all_or_some":
+                {
+                    "label": "All or Some...",
+                    "description": "Include all categories or you can pick which categories to include.  Selecting 'All' will add about 360 lines for a single extruder print and about 440 lines for a dual extruder print.",
+                    "type": "enum",
+                    "options": {
+                        "all_settings": "All Categories",
+                        "pick_settings": "Select Categories"},
+                    "default_value": "all_settings"
+                },
+                "general_set":
+                {
+                    "label": "General",
+                    "description": "The General settings.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": "all_or_some == 'pick_settings'"
+                },
+                "file_info":
+                {
+                    "label": "     Add File Info",
+                    "description": "Add some of the 'General information' (sliced file name, gcode file size, etc.) to the opening paragraph of the Gcode file .",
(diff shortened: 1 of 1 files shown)


fieldOfView:
Instead of hardcoding all the settings, I think this should instead "discover" the settings that are in the ContainerStacks (global stack, extruder stacks). If you hardcode all the settings, you have to carefully update this script whenever settings are changed or added.

GregValiant:
A maintenance issue.  I see that.  I had to add a couple before posting this ("Group Outer Walls" comes to mind).

The ";SETTING_3" section of gcode only contains the "changes", and it's not really readable.  I am trying to accomplish a couple of things with this:

- Group the settings in the order that they appear in Cura.
- Allow the user to only include those sections they are interested in.
- Only display the settings that are relevant to the project.  If Bridge Settings aren't enabled then there is no reason to show all the bridge settings.
- It should be readable.

GregValiant:
I've been thinking about the "maintenance" issue and I don't see a way out.  Of the ~651 settings on @fieldOfView 's keyword page I've included about 400.  Those settings that are included required a decision as to whether (in my opinion) the user would find them useful.  Settings like "material_break_preparation_temperature" weren't included.

If settings are added to Cura between releases I don't believe there would be many.  However, each new setting would need a decision made as to whether to include it in this script, and a decision as to what category to place it in.  After that it would be adding one line of code into the appropriate category so the setting would be included.

My opinion here is that "Some" is better than "None" and adding new settings would be an excellent job for the "new kid" in the office, or maybe a way to punish someone who was late for work.

casperlamboo:
Hi @GregValiant, Thanks once again for your contribution, I think it is very useful for alot of people to have all their settings visible in the generated gcode. However, I'd have to agree with @fieldOfView on this that the maintainability of this code isn't desirable. With the current workload of the team I don't see a possibility where we can structurally add work for the team. Incidental work is always possible, adding structural work not.

What I would suggest is to add something like the following. A dictionary containing all categories you defined, and for each category a list of settings we'd like to include (maybe some additional data (like an enum) should be added here inorder to display various types of data correctly).
```python
    add_settings_list = {
        "postprocess_set": [

        ],
        "general_set": [],
        "machine_set": ["machine_name", "material_diameter", ...],
        ...
    }
```
Then to add the settings we would have to loop over the dictionary, and settings and write universal code that is able to correctly output the various settings. This would greatly improve maintainability as you would only have to change the dictionary. 

Additionally, what I would also like to add is a fallback for if cura fails to retrieve that setting. This would prevent the plugin to fail in the situation where we ever to decide to remove/rename a setting but don't update the postprocessing plugin. With the current setup such universal code would have to be copied for all settings. If we were to implement the dictionary approach we would only have to write this code once, and it would be implemented for all settings.

Let me know if you agree with these remarks, and if we can help setting up such a system!

Thanks,
Casper

GregValiant:
Thanks @casperlamboo .  I'll try and school myself on the "dictionary" thing as I haven't used it previously.
I'll close this until I'm able to come up with a more acceptable alternative.  I think the concept has value.

What might be an insurmountable obstacle is the "decision" process.  I looked at each setting and made a decision as to whether the particular setting was something that a user would want/need.  That whittled down the total number of settings to something manageable.

_"...a fallback for if cura fails to retrieve that setting"_ 
I noticed that if I had the plugin installed in a 4.x Cura version that no errors were generated if a setting didn't exist.  Since the script was written using Cura 5.6 there are a lot of settings missing when it runs in 4.x.  There were no errors generated.  I thought that was odd, but that's how it worked.

I put together a script to pull settings out of fdmprinter.def.json and compare it against a list pulled from fdmprinter.def.json in a different Cura version.  Between 5.6 and 5.7a there was a single setting obsoleted, and 19 settings added.  That's not from the dictionary of course, but I wanted an idea of what kind of numbers I'm looking at.  Of the 19 some are specific to the nature of the nightly build, others would need a decision made.
I'm guessing that "bv_temp_warn_limit" and "print_temp_warn_limit" are settings that a user wouldn't care about.  It would be my decision to not include them in "Add Cura Settings".

Back to the drawing board.

PR #17971, opened 2024-01-10
Prime/purge tower optimisation

wawanbreton:
See https://github.com/Ultimaker/CuraEngine/pull/1875


wawanbreton:
I made some polishing on the new settings, so it is now ready for undrafting !

It will require an upgrade script though, because I changed the setting `prime_tower_enabled` to `prime_tower_mode`, @casperlamboo can you tell me if this should be part of the PR ?

@MariMakes would you have a few minutes to read the names and descriptions of the new settings and give me a feedback ?

casperlamboo (approved):

PR #17983, opened 2024-01-11
Setting QT_QUICK_FLICKABLE_WHEEL_DECELERATION

saumyaj3:
for qt scroll related issues

CURA-11468

### DIFF
--- a/cura_app.py
+++ b/cura_app.py
@@ -16,7 +16,7 @@ import argparse
 import faulthandler
 import os
 
-# try converting to integer
+# set the environment variable QT_QUICK_FLICKABLE_WHEEL_DECELERATION to 5000 as mentioned in qt6.6 update log to overcome scroll related issues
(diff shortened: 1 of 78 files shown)


cura_app.py
```diff
@@ -15,6 +15,10 @@
 import argparse
 import faulthandler
 import os
+
+# set the environment variable QT_QUICK_FLICKABLE_WHEEL_DECELERATION to 5000 as mentioned in qt6.6 update log to overcome scroll related issues
+os.environ["QT_QUICK_FLICKABLE_WHEEL_DECELERATION"] = str(int(os.environ.get("QT_QUICK_FLICKABLE_WHEEL_DECELERATION", "5000")))
```
casperlamboo, line 20: Won't this produce a value error when `QT_QUICK_FLICKABLE_WHEEL_DECELERATION` is set to a non number string.

casperlamboo (approved):

PR #18042, opened 2024-01-16
Fix method printers fan range

wawanbreton:
Fix Method X/XL fan range

CURA-11441


wawanbreton:
This is superseded by https://github.com/Ultimaker/Cura/pull/17880

PR #18055, opened 2024-01-17
Reduce initial layer speed

pkuiper-ultimaker:
Reduce initial layer speed to improve reliability for S3 and S5 machines when no glue is used.

PP-421


wawanbreton (approved):
Changes as expected !

PR #18064, opened 2024-01-18
PP-423 Increase small skin width and wipe distance

pkuiper-ultimaker:
Increase small skin width from 0.8mm to 4mm to prevent jerky moves in small skin areas. 
Increase wipe distance to 0.8mm to improve seam quality.
For Method machines.

PP-423


wawanbreton (approved):

wawanbreton (changes requested):
This value is over the standard warning limit which is `2 * machine_nozzle_size`. Please consider changing the `maximum_value_warning` so that the new value fits

pkuiper-ultimaker:
> This value is over the standard warning limit which is `2 * machine_nozzle_size`. Please consider changing the `maximum_value_warning` so that the new value fits

I changed the warning value in the fdmprinter.def.json from 3 line widths to 10 line widths.

wawanbreton (approved):

PR #18128, opened 2024-01-25
Cura 11542 optimized prime tower

wawanbreton:
Wrapper PR to enable building a full Cura application from an external repository
Don't forget to delete it once done


wawanbreton:
Your destiny was to be deleted, sorry

PR #18188, opened 2024-01-31
Fix mac Builds

casperlamboo:
Since the pyqt 6.6 upgrade the mac builds were failing. Main fix was to upgrade pyinstaller (which allowed us to remove a whole lot of ugly ductyping). 

In the new packaged app the file structure is a bit different for mac, so had to change some paths.

Also see Uranum PR: https://github.com/Ultimaker/Uranium/pull/936/files

CURA-11483


UltiMaker-Cura.spec.jinja
```diff
@@ -70,188 +70,7 @@ coll = COLLECT(
...
-        try:
-            osxutils.sign_binary(self.name, self.codesign_identity, self.entitlements_file, deep = True)
-        except Exception as e:
-            logger.warning(f"Error while signing the bundle: {e}")
-            logger.warning("You will need to sign the bundle manually!")
-
-        logger.info(f"Building BUNDLE {self.tocbasename} completed successfully.")
-
-app = UMBUNDLE(
+app = BUNDLE(
```
jellespijker, line 73: If we don't need the duck typing anymore, then we probably can skip this whole conditional jinja block and go for the out-of-the-box solution

conanfile.py
```diff
@@ -163,8 +163,10 @@ def _conan_installs(self):
     def _python_installs(self):
         python_installs = {}
 
+        python_executable = "python3" if self.settings.os == "Macos" else "python"
+
         # list of python installs
-        python_ins_cmd = f"python -c \"import pkg_resources; print(';'.join([(s.key+','+ s.version) for s in pkg_resources.working_set]))\""
+        python_ins_cmd = f"{python_executable} -c \"import pkg_resources; print(';'.join([(s.key+','+ s.version) for s in pkg_resources.working_set]))\""
```
jellespijker, line 169: This shouldn't be needed anymore with my recent changes. If it is still needed then apply the environment context to this, see:

- https://github.com/Ultimaker/conan-config/pull/10
- https://github.com/Ultimaker/conan-config/pull/9
- https://github.com/Ultimaker/Cura/pull/18138

cura/CuraApplication.py
```diff
@@ -389,6 +391,7 @@ def __addExpectedResourceDirsAndSearchPaths(self):
 
             # venv site-packages
             Resources.addSearchPath(os.path.join(app_root, "..", "share", "cura", "resources"))
+            Resources.addSearchPath(os.path.join(app_root, "..", "Resources", "share", "cura", "resources"))
```
jellespijker, line 394: Yeah sure, just add all those extra paths ;-)

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 pytest
-pyinstaller==5.8.0
+pyinstaller==6.3.0
```
jellespijker, line 2: If we're using the latest version then I would suggest to unpin the version

jellespijker (commented):
Just some adhoc remarks, I know it is not ready yet

UltiMaker-Cura.spec.jinja
casperlamboo: 🦆

cura/CuraApplication.py
```diff
@@ -373,9 +373,17 @@ def __addExpectedResourceDirsAndSearchPaths(self):
...
-        Resources.addSecureSearchPath(os.path.join(self._app_install_dir, "share", "cura", "resources"))
+        if platform.system() == "Darwin":
+            Resources.addSecureSearchPath(os.path.join(app_root, "Resources", "share", "cura", "resources"))
+            Resources.addSecureSearchPath(os.path.join(app_root, "..", "Resources", "share", "cura", "resources"))
+            Resources.addSecureSearchPath(
+                os.path.join(self._app_install_dir, "..", "Resources", "share", "cura", "resources"))
+        else:
+            Resources.addSecureSearchPath(os.path.join(app_root, "share", "cura", "resources"))
+            Resources.addSecureSearchPath(os.path.join(self._app_install_dir, "share", "cura", "resources"))
+            Resources.addSecureSearchPath(os.path.join(app_root, "Resources", "share", "cura", "resources"))
```
wawanbreton, line 385: This one is duplicated in the `if` and in the `else`, so it could be outside (or removed somewhere)

wawanbreton (commented):
Looks good, less code = less bugs !

cura/CuraApplication.py
casperlamboo: oh yea, you're right. Let me change this

casperlamboo: https://github.com/Ultimaker/Cura/pull/18188/commits/d8cc99bf0deb951d32059e47b7690a0c76babfcb

PR #18199, opened 2024-02-01
crash while opening printer setting

saumyaj3:
It happened because of creating MachineActionManager as a singleton. Code is refractored to make it work.

CURA-11558


cura/CuraApplication.py
```diff
@@ -186,7 +187,7 @@ def __init__(self, *args, **kwargs):
 
         self._cura_formula_functions = None  # type: Optional[CuraFormulaFunctions]
 
-        self._machine_action_manager = None  # type: Optional[MachineActionManager.MachineActionManager]
+        self._machine_action_manager = None  # type: Optional[MachineActionManager]
```
casperlamboo, line 190: Use modern python typing
```suggestion
        self._machine_action_manager: Optional[MachineActionManager] = None
```
We didn't do this throughout the codebase as it wasn't always available, but for newly added code we should always use modern typing

PR #18246, opened 2024-02-05
Cura 11561 mockup pap

wawanbreton:
Implements a mockup for testing the Printer Agnostic Project new feature. This is really a preview and would probably require some technical+feature polishing to be fully usable.

Don't merge to main !

CURA-11561


plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -228,11 +234,14 @@ def preRead(self, file_name, show_dialog=True, *args, **kwargs):
         self._resolve_strategies = {k: None for k in resolve_strategy_keys}
         containers_found_dict = {k: False for k in resolve_strategy_keys}
 
+        # Check whether the file is a UCP, which changes some import options
+        is_ucp = file_name.endswith('.ucp')
```
nallath, line 238: It's fine for now, but for the future i would just look for the extra settings file. That way it's already compatible if we ever decide to merge it.

plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) 2022 Ultimaker B.V.
 // Cura is released under the terms of the LGPLv3 or higher.
 
-import QtQuick 2.10
+import QtQuick 2.14
```
nallath, line 4: Why update this? You aren't using antything from a newer QtQuick version. This import only defines what functions are available. Increasing it doesn't mean you suddenly get bugfixes you wouldn't otherwise.

plugins/3MFWriter/ThreeMFWorkspaceWriter.py
```diff
@@ -13,15 +15,50 @@
...
+        self._success = False
+        self._export_model = None
+        self._stream = None
+        self._nodes = None
+        self._mode = None
+        self._config_dialog = None
 
-    @call_on_qt_thread
-    def write(self, stream, nodes, mode=WorkspaceWriter.OutputMode.BinaryMode):
+    def _preWrite(self):
```
nallath, line 37: Not the greatest solution, but it's fine for the mockup. When integrating this, i do think we should split off the preWrite.

plugins/3MFReader/ThreeMFWorkspaceReader.py
wawanbreton: This would be cleaner indeed :+1:

plugins/3MFReader/WorkspaceDialog.qml
wawanbreton: Right, I needed it at some point when playing with the table view/model items but didn't change it back.

plugins/3MFWriter/ThreeMFWorkspaceWriter.py
wawanbreton: The reader classes have explicit `preRead` and `read` methods, so we should probably do the same for the writers. I didn't want to go there for the mockup, but for integration yes.

For the extra param, I don't really understand what you mean. What I think is that we should add a param to the `write` method which is the mimetype of the file to be written. Or allow the mode param (which is currently useful only for the STLs) to be any type, and then have a specific enum for the 3MF/UCP.

PR #18310, opened 2024-02-13
Cura 10542 drop to buildplate

saumyaj3:
# Description

Drop to buildplate preference setting shown while opening a project, a user can select if they want to drop or not. By default it is set to the user preference
Capture

### DIFF
--- a/cura/PlatformPhysics.py
+++ b/cura/PlatformPhysics.py
@@ -80,9 +80,11 @@ class PlatformPhysics:
 
             # Move it downwards if bottom is above platform
             move_vector = Vector()
+
             if node.getSetting(SceneNodeSettings.AutoDropDown, app_automatic_drop_down) and not (node.getParent() and node.getParent().callDecoration("isGroup") or node.getParent() != root) and node.isEnabled(): #If an object is grouped, don't move it down
                 z_offset = node.callDecoration("getZOffset") if node.getDecorator(ZOffsetDecorator.ZOffsetDecorator) else 0
                 move_vector = move_vector.set(y = -bbox.bottom + z_offset)
+
             # If there is no convex hull for the node, start calculating it and continue.
             if not node.getDecorator(ConvexHullDecorator) and not node.callDecoration("isNonPrintingMesh") and node.callDecoration("getLayerData") is None:
                 node.addDecorator(ConvexHullDecorator())
--- a/cura/PlatformPhysics.py
+++ b/cura/PlatformPhysics.py
@@ -38,7 +38,7 @@ class PlatformPhysics:
         self._minimum_gap = 2  # It is a minimum distance (in mm) between two models, applicable for small models
 
         Application.getInstance().getPreferences().addPreference("physics/automatic_push_free", False)
-        Application.getInstance().getPreferences().addPreference("physics/automatic_drop_down_per_model", "never")
+        Application.getInstance().getPreferences().addPreference("physics/automatic_drop_down", False)
 
     def _onSceneChanged(self, source):
         if not source.callDecoration("isSliceable"):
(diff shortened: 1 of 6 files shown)


resources/qml/Toolbar.qml
```diff
@@ -222,7 +222,7 @@ Item
         UM.Label
         {
             id: toolHint
-            text: UM.Controller.properties.getValue("ToolHint") != undefined ? UM.ActiveTool.properties.getValue("ToolHint") : ""
+            text: UM.Controller.properties.getValue("ToolHint") != undefined ? UM.Controller.properties.getValue("ToolHint") : ""
```
saumyaj3, line 225: This change is for the deprecation warning, because of singleton removal.

plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -300,6 +300,20 @@ UM.Dialog
...
+                {
+                    id: dropToBuildPlate
+                    width: parent.width
+                    height: childrenRect.height
+                    spacing: UM.Theme.getSize("default_margin").width
+                    UM.CheckBox
+                    {
+                        text: catalog.i18nc("@text:window", "Drop models to buildplate")
+                        checked: UM.Preferences.getValue("physics/automatic_drop_down") == True
+                        onCheckedChanged: UM.Preferences.setValue("physics/automatic_drop_down", checked)
```
rburema, line 313: While I think the general approach of this ticket is good, the setting does more than just take effect when loading a model into Cura: It also handles what happens if you make a model float regardless of opening.

Also; what happens if you open a model, change the checkbox to be different from your preference, then load a (floating) STL? (Or 'non-workspace' 3MF, or other...)

So here should probably be a dropdown operation attached for the just opened models somehow (in case of a workspace, that'll remove all others from the buildplate, so it can just be all models loaded), instead of changing the entire preference each time we load a workspace.

rburema (changes requested):
Good idea!

I just think we should not change the _overall_ preference when opening a model. I think that if changed in the open-workspace-dialog, the user intends to change it just for that instance, not in general. (See also below.)

plugins/3MFReader/WorkspaceDialog.qml
saumyaj3: 32d9e6b45c56f65beea413248849b8599e62c30a
I made a separate preference variable that silently mimics drop_to_buildplate value until a user selects otherwise for that model. Solves this issue 💯

cura/PlatformPhysics.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, controller, volume):
 
         Application.getInstance().getPreferences().addPreference("physics/automatic_push_free", False)
         Application.getInstance().getPreferences().addPreference("physics/automatic_drop_down", False)
+        Application.getInstance().getPreferences().addPreference("physics/per_model_drop", False)
```
rburema, line 42: I don't think we should make this temporary variable a preference -- those are basically only used if it's a user setting the program needs to remember between sessions (so when the program closes). This seems like both overkill and hacking the preference system.

Since we're in the same session, there must be another way to properly communicate this to the platform physics object. Maybe the platform physics could pull it from somewhere, but since they're separate systems, perhaps a signal (like 'suppress drop to buildplate only for the next operation') would be better.

rburema (changes requested):
Sorry, I still have something to complain about ;-)

rburema (approved):

Zogar89:
Hi, I have the issue of open a stored project with objects below the bed, and cura put its over the bed. This solve that?

rburema:
@Zogar89 That sounds like a separate issue, can you please report it as a new issue here on github, perhaps with attached screenshots or a project, so that we can see what's going on?

PR #18374, opened 2024-02-19
Add definitions and profiles to ZYYX+, ZYYX Pro and ZYYX Pro ii

theodorhansson:
# Description

This adds three printers that were previously missing from Cura. 

* ZYYX+
* ZYYX Pro
* ZYYX Pro ii

This also changes the manufacturer of the following printer. Which was previously made by Magicfirm Europe, but the firm was rebranded to ZYYX Labs AB.
* ZYYX Agile

## Type of change

Implements 

# How Has This Been Tested?

Testprints have been performed on all ZYYX printers except ZYYX Agile.

**Test Configuration**:
- [x] Windows
- [ ] Linux (the profiles work but compiling to .x3g does not work on all machines)
- [x] MacOS 

# Checklist:


theodorhansson:
Note: this PR references some materials not yet implemented in Cura. These materials are currently awating approval in [PR#281 in Ultimaker/fdm_materials](https://github.com/Ultimaker/fdm_materials/pull/281).

DavidPrytzArcombe:
I tested one of these profilers on a YYX+, it worked well and sped up the print. Would recommend.

theodorhansson:
The profiles in this PR have been beta-tested by some different users of ZYYX printers and I feel that the profiles are now ready for review. The profiles were distributed to the testers from ([theodorhansson/ZYYX-Cura-Printer-Profiles](https://github.com/theodorhansson/ZYYX-Cura-Printer-Profiles/releases).

It would be appreciated if someone could review the changes proposed.

theodorhansson:
@saumyaj3 @nallath @HellAholic @wawanbreton

wawanbreton:
Hi @theodorhansson, this is quite a big PR, but it looks good at first glance. I will take some time to review it. What about the "Agile" printer then ? If this is no official product, I would indeed recommend to remove it, otherwise your users may get confused or think that this is an incoming printer.

theodorhansson:
@wawanbreton I do not work for ZYYX, but I have spoken with their CEO over the phone. He claimed that the "Agile" printer was a concept that would evolve into the "Pro". My recommendation is that the "Agile" printer can be removed from Cura.

theodorhansson:
@wawanbreton It's been removed, but the bot complains...

wawanbreton:
Indeed :sweat_smile: I'll see with the team how to handle this properly

wawanbreton:
So, the consequences of removing the printer definition are the following:
* Users having an instance of this printer won't be able to use it anymore
* Project files using this printer won't be usable anymore

The very proper way to handle this situation would have been to create an upgrade script, and convert the printers and project files using Agile to now using Pro, which is what actually happened. However, this is much more complex, so if we assume nobody in the world (except maybe a few engineers at ZYYX) have used this printer, we can safely ignore this warning. Do you think it would be worth contacting them again to check, or can we safely remove it, knowing the consequences ?

theodorhansson:
My view is that very few people use the printer and it would therefore not be worth writing an upgrade script. The ZYYX Pro was released in 2017, so I doubt that even the engineers at ZYYX are still using the "Agile". Furthermore, up until this PR I believe that most users used Simplyfy3D instead of Cura.

wawanbreton:
Ok, looks like they just made a test with Cura, that ended up in the mainstream for weak reasons. I'm good with removing it then. If people somehow get to complain about this, they can still restore the files on their installation, and we shall put it back, but that is extremely unlikely.

resources/quality/zyyx_plus/ABS/zyyx_plus_abs_fast.inst.cfg
```diff
@@ -0,0 +1,23 @@
+[general]
+definition = zyyx_plus
+name = Fast ABS
+version = 4
+
+[metadata]
+material = generic_abs
+quality_type = fast
+setting_version = 3
```
wawanbreton, line 9: `setting_version` should be 23. If you are actually using old interpretation/naming of settings, they should be manually upgraded. Same for all other files.

resources/variants/zyyx/zyyx_pro_carbon0.6.inst.cfg
```diff
@@ -0,0 +1,18 @@
...
+
+[metadata]
+setting_version = 3
+type = variant
+hardware_type = nozzle
+
+[values]
+acceleration_enabled = True # No effect on print, only for better print time prediction
+acceleration_print = 1000
+line_width = 0.6
```
wawanbreton, line 14: This is not necessary. `line_width` is by default equal to `machine_nozzle_size`, which you already properly define. Same for other variants.

wawanbreton (changes requested):
I spotted one issue with setting_version, and a possible optimization for variants. Otherwise looks very good !

theodorhansson:
@wawanbreton Versions updated and line_width removed!

wawanbreton:
Nice, thanks ! I just saw that some variants defined a line_width different from the machine_nozzle_size tough, those should probably be restored ?

Also there are some conflicts with main, because we just upgraded the settings_version to 24 (didn't think we would merge it that fast :grinning:) so you can ignore the conflicts and force your files removals.

wawanbreton:
> Should I bump the settings version to 24?

As you are on it, you can. Otherwise we plan to make an other pass for it.

wawanbreton:
changes between 23 and 24 are really self-contained and don't concern your settings at all, so you can safely put either 23 or 24

theodorhansson:
Then 23 is ok by me.

wawanbreton:
Hi @theodorhansson, can you just fix the git conflicts ? After that I guess we can merge this PR

wawanbreton:
Very nice, thanks ! I can now approve and it will be merged after further testing.

wawanbreton (approved):

wawanbreton:
Hi @theodorhansson, we just realized that the profiles contain a flow ratio of 420, which is totally outside usual specs. Is this on purpose, or possibly a mistake ?

theodorhansson:
Hi @wawanbreton! It is on purpose. The extruder was changed by the manufacturer, but the firmware was not. I believe it was a geared Bondtech, but I'm not 100%. Unfortunately I have not personally printed with a ZYYX Pro II.

theodorhansson:
I looked again at the Simplify3D profiles from the manufacturer and they also contain an abnormally large extrusion multiplier.

wawanbreton:
Ok, thanks for your quick answer :slightly_smiling_face:
We have decided to exclude this printer for the moment: the very high flow value may have side-effects on the slicing engine that nobody will see until they start a print. Since we are unable to confirm that this is not the case, we prefer not to risk it for now.
Maybe you can contact ZYYX and ask them for some testing, if they have an interest to it ?

PR #18403, opened 2024-02-22
Re-introduce `prime_tower_enable`

casperlamboo:
# Description

Previously we introduced new settings related to the prime tower. Specifically the setting `prime_tower_enable` was changed to `prime_tower_mode` where `prime_tower_enable=False` is reinterpreted to `prime_tower_mode=None` and `prime_tower_enable=True` is can be either `prime_tower_mode=Normal` or `prime_tower_mode=Interleaved`.

This reinterpretation poses some issues related to upgrade script, value resolving etc. Additionally it isn't consistent with how settings are defined elsewhere. For example, for the support setting we have two settings, namely: `(bool)support_enabled` and `(enum)support_mode` where `support_mode` can be either `tree` or `normal`. Following this same logic we should have introduced `(bool)prime_tower_enable` and `(enum)prime_tower_mode`.

also see enginge PR https://github.com/Ultimaker/CuraEngine/pull/2033

CURA-11645

## Type of change

- [ ] Translations

# How Has This Been Tested?

Yas

**Test Configuration**:
* Operating System: macos13

# Checklist:


resources/definitions/fdmprinter.def.json
```diff
@@ -6794,20 +6794,30 @@
...
                     "options":
                     {
-                        "none": "None",
                         "normal": "Normal",
                         "interleaved": "Interleaved"
                     },
-                    "default_value": "none",
-                    "enabled": "extruders_enabled_count > 1",
+                    "default_value": "normal",
+                    "enabled": "prime_tower_enabled",
```
pkuiper-ultimaker, line 6820: prime_tower_enable not enable**d**

pkuiper-ultimaker (changes requested):

resources/definitions/fdmprinter.def.json
casperlamboo:
```suggestion
                    "enabled": "prime_tower_enable",
```

pkuiper-ultimaker (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -6794,20 +6794,30 @@
...
+                    "enabled": "extruders_enabled_count > 1",
+                    "default_value": false,
+                    "resolve": "(extruders_enabled_count > 1) and any(extruderValues('prime_tower_enable'))",
+                    "settable_per_mesh": false,
+                    "settable_per_extruder": false
+                },
                 "prime_tower_mode":
                 {
                     "label": "Prime Tower",
-                    "description": "<html>Print a tower next to the model, which serves to prime the material after each nozzle switch.<ul><li><b>None:</b> do not generate a prime tower</li><li><b>Normal:</b> generate a bucket in which secondary materials are primed</li><li><b>Interleaved:</b> generate a prime tower as sparse as possible. This will save time and filament, but is only possible if the used materials adhere to each other.</li></ul></html>",
```
wawanbreton, line 6800: I think the first sentence should be changed, because the mode now only describes how the prime tower will be printed.

casperlamboo: Sure, can you post a suggestion for the updated description?

wawanbreton:
I see a lot of settings in fdmprinter.def.json that still rely on `prime_tower_mode` to see if the prime tower is enabled. They should go back to using `prime_tower_enable`
It is also probably missin in `SliceInfo.py` and in settings visibility profiles.

resources/definitions/fdmprinter.def.json
```diff
@@ -6794,20 +6794,30 @@
...
+                    "description": "<html>Print a tower next to the model, which serves to prime the material after each nozzle switch.<ul><li><b>Normal:</b> generate a bucket in which secondary materials are primed</li><li><b>Interleaved:</b> generate a prime tower as sparse as possible. This will save time and filament, but is only possible if the used materials adhere to each other.</li></ul></html>",
```
wawanbreton, line 6811:
```suggestion
                    "description": "<html>How to generate the prime tower:<ul><li><b>Normal:</b> create a bucket in which secondary materials are primed</li><li><b>Interleaved:</b> create a prime tower as sparse as possible. This will save time and filament, but is only possible if the used materials adhere to each other.</li></ul></html>",
```

PR #18466, opened 2024-02-29
Cura 9838 brim location

wawanbreton:
* Replaced setting `brim_outside_only` by `brim_location` to give more brim option.
* Extended the use of `brim_inside_margin` which is now more widely used. The description, documentation and default value have changed but I kept the name which still makes sense, and to make retro-compatibility easier.

Comes along with https://github.com/Ultimaker/CuraEngine/pull/2040


saumyaj3 (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -6069,26 +6069,33 @@
...
                     "limit_to_extruder": "skirt_brim_extruder_nr",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
                 "brim_inside_margin":
                 {
-                    "label": "Brim Inside Avoid Margin",
-                    "description": "A part fully enclosed inside another part can generate an outer brim that touches the inside of the other part. This removes all brim within this distance from internal holes.",
+                    "label": "Brim Avoid Margin",
+                    "description": "When growing the brim around a surface, it may touch an other surface where brim is not expected. This removes all brim within this distance from clean surfaces.",
```
casperlamboo, line 6092: I think the part "when growing the brim around a surface" is a bit too technical and only makes sense when you know the implementation of a brim, I would simplify the text to
```suggestion
                    "description": "A brim around a surface may touch an other surface where brim is not expected. This removes all brim within this distance from brimless surfaces.",
```

casperlamboo (approved):

PR #18523, opened 2024-03-06
CURA-11657 Fix crashes with USB printing

wawanbreton:
There is apparently a bug in Qt 6.6.0 that causes a segmentation fault when deleting a Python object that has been instanciated in QML, if a custom QML property has been set on this object. This bug has been [reported to Riverbank](https://www.riverbankcomputing.com/pipermail/pyqt/2024-March/045751.html).

In the meantime, I could still fix the crash because the properties set in QML are apparently useless now. They concern a resolve function on the "material_bed_temperature" which is a global setting. I am pretty sure this code has been copy-pasted from an other similar case, but in this case this is not needed. Please confirm this :)

However, the PyQt bug is not fixed yet and we may encounter other crashes. I just think/hope this is the only QML screen we ever destroy so this bug happens, and on other cases we don't destroy the screen so it doesn't trigger.

CURA-11657

### DIFF
--- a/resources/qml/PrintMonitor.qml
+++ b/resources/qml/PrintMonitor.qml
@@ -132,8 +132,6 @@ ScrollView
             key: "material_bed_temperature"
             watchedProperties: ["value", "minimum_value", "maximum_value", "resolve"]
             storeIndex: 0
-
-            property var resolve: Cura.MachineManager.activeStack != Cura.MachineManager.activeMachine ? properties.resolve : "None"
         }
 
         UM.SettingPropertyProvider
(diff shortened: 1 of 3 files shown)


jellespijker:
Good that you report the issue. I'm curious if you get a response. Riverbank has proven to be unresponsive in reported bugs or questions from us in the past. Lets hope that this has been improved.

wawanbreton:
@jellespijker The mailing list looks pretty active, even for side-questions, so I have good hope to get an answer someday. Also being able to provide a minimalistic project to reproduce the bug is usually appreciated, so that may help in this case :crossed_fingers:

saumyaj3 (approved):

saumyaj3:
Looks like the code removed was trying to make it just faster to access properties in the USB printing (talk about nanoseconds). Doesn't hold any true purpose and causes crash. And also not used widely.

PR #18533, opened 2024-03-07
Add Post Processing script to set initial extruder offset

ansonl:
# Description

This post processing script sets multi-extruder printers' initial extruder positions to the correct value after retracting the filament during start G-code.

1. Before priming a nozzle at the beginning of the print, the position of the filament in the nozzle is unknown but assumed to be between `switch_extruder_retraction_amount` and `0` because the filament may have been retracted the extended `switch_extruder_retraction_amount` length at the end of the last print (or not). When a print is started and the filament position is assumed to be 0 (at the nozzle already) no/underextrusion occurs for `switch_extruder_retraction_amount` distance of filament because the slicer assumes the printer is extruding filament when the filament is still moving down the heatbreak towards the nozzle. 

2. Normally multiple nozzles on a multi-extruder printer are primed by extruding some filament during the start g-code to ensure the filament is at a known position in the extrusion train. *E.g. Prime by `switch_extruder_retraction_amount` + a bit more.* 
 
3. When the printer has more than a single nozzle, the extruders that are NOT the `initial_extruder_nr` need to be retracted by `switch_extruder_retraction_amount` after priming to avoid oozing. 

4. After start G-code ends and the printing starts, the slicer assumes that each extruder starts from `0` position rather than -`switch_extruder_retraction_amount`. When switching to each extruder for the first time, the extruder printing movements start with E position of `0` leading to no extrusion for a while until the filament moves down the heatbreak into the nozzle.

5. The slicer remembers the extruder position when it generates code to retract the extruder by `switch_extruder_retraction_amount` when the extruder is switched away from in the toolchange and restores the extruder position by moving the filament forward by `switch_extruder_retraction_amount` distance when the extruder is switched to in a toolchange. So this no/under extrusion problem only happens on the first toolchange until the extruder has extruded at least `switch_extruder_retraction_amount`mm of filament.

I fixed this issue by creating the post processing script to set the correct position for the extruder for the first time each extruder is used. `G92 E0` is corrected to `G92 E-16` (if the printer had a `switch_extruder_retraction_amount` of 16). 

I drew the below diagram to visually show the extruder position problem. 

extrusion issue

Another solution for this problem would be if Cura assumes that each extruder has been retracted to `switch_extruder_retraction_amount` distance at the end of the start G-code and starts each extruder's initial position at the start of the print G-code to -`switch_extruder_retraction_amount` instead of 0. 

## Type of change

# How Has This Been Tested?

I have run this post processing script on multiple prints and extrusion is improved due to setting extruder position to the correct value. No oozing occurs after priming because the extruder retracts filament by `switch_extruder_retraction_amount` in start G-code and extrusion begins immediately when the slicer switches to that extruder.

**Test Configuration**:
* Operating System: Win10

# Checklist:


plugins/PostProcessingPlugin/scripts/SetInitialSwitchedExtruderOffset.py
```diff
@@ -0,0 +1,202 @@
...
+                    "default_value": false
+                },
+                "setT1ExtruderOffset":
+                {
+                    "label": "Set Extruder 2 Offset",
+                    "description": "Set the initial offset of this extruder by Nozzle Switch Retraction Distance.",
+                    "type": "bool",
+                    "default_value": "CuraApplication.getInstance().getExtruderManager().getInitialExtruderNr() != 1"
+                },
+                "setT2ExtruderOffset":
```
saumyaj3, line 55: how about adding a parameter "enabled" to see if the extruder exist in the printer and then show the parameter to set the offset of extruder?

ansonl: Which Cura variable should I use to get the "enabled" state of an extruder index? I had tried to get the extruder enabled state previously but got an error trying to access the value.

saumyaj3: try "machine_extruder_count"

ansonl: I tried reading the value during getSettingDataString() in the "enabled" json field like below but it always evaluates to False

```json
"setT1ExtruderOffset":
                {
                    "label": "Set Extruder 2 Offset",
                    "description": "Set the initial offset of this extruder by Nozzle Switch Retraction Distance.",
                    "type": "bool",
                    "default_value": "CuraApplication.getInstance().getExtruderManager().getInitialExtruderNr() != 1",
                    "enabled": "CuraApplication.getInstance().getGlobalContainerStack().getProperty(\\\"machine_extruder_count\\\", \\\"value\\\") >= 2"
                },
```

When I read the value normally right before the json string in getSettingDataString() it logs the correct values in cura.log

```python
logthis = f'SetInitialSwitchedExtruderOffset machine_extruder_count {CuraApplication.getInstance().getGlobalContainerStack().getProperty("machine_extruder_count", "value")}'
log2= f'machine_extruder_count>=2 {CuraApplication.getInstance().getGlobalContainerStack().getProperty("machine_extruder_count", "value") >= 2}'
Logger.log("d", logthis)
Logger.log("d", log2)
```

I looked at the other plugins to find any "enabled" examples that call python and they only reference other settings fields. Is there a restriction on running Python in "enabled"?

ansonl: From the Cura logs, it seems like "enabled" code is unable to call blacklisted methods that are only blacklisted for the json?
```
2024-04-24 23:41:39,953 - ERROR - [MainThread] UM.Settings.SettingFunction._safeCompile [77]: Use of illegal method extruderIds in function ("2" in CuraApplication.getInstance().getExtruderManager().extruderIds) for setting
2024-04-24 23:41:39,953 - ERROR - [MainThread] UM.Settings.SettingFunction._safeCompile [77]: Use of illegal method getProperty in function (CuraApplication.getInstance().getGlobalContainerStack().getProperty("machine_extruder_count", "value") >= 4) for setting
```

PR #18534, opened 2024-03-07
Use proper gradual flow plugin version

wawanbreton:
Bumping gradual flow plugin version to 0.1.0-beta.3, not 0.2.0 because we don't have a proper settings upgrade workflow for engine plugins yet.


wawanbreton:
I can't test a full build yet because the branch doesn't have the standard name...

jellespijker (approved):

PR #18552, opened 2024-03-08
Pin package and requirements versions

wawanbreton:


conandata.yml
```diff
@@ -1,17 +1,17 @@
...
-  - "cura_binary_data/(latest)@ultimaker/testing"
-  - "fdm_materials/(latest)@ultimaker/testing"
-  - "curaengine_plugin_gradual_flow/0.1.0-beta.3"
-  - "dulcificum/latest@ultimaker/testing"
+  - "uranium/(latest)@ultimaker/stable"
+  - "curaengine/(latest)@ultimaker/stable"
+  - "cura_binary_data/(latest)@ultimaker/stable"
+  - "fdm_materials/(latest)@ultimaker/stable"
+  - "curaengine_plugin_gradual_flow/0.1.0-beta.3@ultimaker/stable"
+  - "dulcificum/latest@ultimaker/stable"
```
casperlamboo, line 8: we should only pin to stable for repos that have release branches, i believe `dulcificum` doesn't have a release branch?

wawanbreton: No indeed, but I thought we would manually build a package on the stable channel ?

jellespijker: Creating a Conan package in the cura-workflows is only used for a specific package. Such as a public release. We are now simply redirecting traffic from the `testing` branches to the `stable` branches.

jellespijker (changes requested):
Only point to stable for repo's with an actual release branch

casperlamboo (approved):

PR #18554, opened 2024-03-08
Add missing nylon material to material map

casperlamboo:
CURA-11709

### DIFF
--- a/cura/PrinterOutput/Models/MaterialOutputModel.py
+++ b/cura/PrinterOutput/Models/MaterialOutputModel.py
@@ -28,7 +28,7 @@ class MaterialOutputModel(QObject):
                             "abs-wss1"	:{"name" :"absr_175"      ,"guid": "88c8919c-6a09-471a-b7b6-e801263d862d"},
                             "asa" 		:{"name" :"asa_175"       ,"guid": "416eead4-0d8e-4f0b-8bfc-a91a519befa5"},
                             "nylon-cf" 	:{"name" :"cffpa_175"     ,"guid": "85bbae0e-938d-46fb-989f-c9b3689dc4f0"},
-                             "nylon12-cf": {"name": "nylon12-cf", "guid": "3c6f2877-71cc-4760-84e6-4b89ab243e3b"},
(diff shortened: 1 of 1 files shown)


cura/PrinterOutput/Models/MaterialOutputModel.py
```diff
@@ -28,6 +28,7 @@ def getMaterialFromDefinition(guid, type, brand, name):
                             "abs-wss1"	:{"name" :"absr_175"      ,"guid": "88c8919c-6a09-471a-b7b6-e801263d862d"},
                             "asa" 		:{"name" :"asa_175"       ,"guid": "416eead4-0d8e-4f0b-8bfc-a91a519befa5"},
                             "nylon-cf" 	:{"name" :"cffpa_175"     ,"guid": "85bbae0e-938d-46fb-989f-c9b3689dc4f0"},
+                             "nylon12-cf": {"name": "nylon12-cf", "guid": "3c6f2877-71cc-4760-84e6-4b89ab243e3b"},
```
HellAholic, line 31: `nylon12-cf_175`
following the same pattern

casperlamboo:
```suggestion
                             "nylon12-cf": {"name": "nylon12-cf_175", "guid": "3c6f2877-71cc-4760-84e6-4b89ab243e3b"},
```

jellespijker:
It still feels wrong to have such a mapping. But sometimes there is no way around such a thing

casperlamboo:
> It still feels wrong to have such a mapping. But sometimes there is no way around such a thing
There is a good way around this actually. We can create this map dynamically on start up. We just have to loop through all materials, then if the [`reference_material_id`](https://github.com/Ultimaker/fdm_materials/blob/97219c0cc5274a544217927c160d9b332c494ea8/ultimaker_nylon12-cf_175.xml.fdm_material#L10) property is present in that material that material is inserted in the map using `reference_material_id` as key.

saumyaj3 (approved):

PR #18567, opened 2024-03-11
Add native CAD plugin to Enterprise edition

jellespijker:
The native CAD plugin has been added to the requirements and reference copy operations in the Conan data file and Conan file script. Now the native CAD plugin will only be included if the enterprise build flag is set. This ensures that unnecessary dependencies are not brought into non-enterprise builds.

> [!CAUTION]
> Untested just a quick port you also might want to validate that native_cad_plugin version 2.0.0 is indeed the latest released version

# Description

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:


wawanbreton:
Is this change required for 5.7 ?

jellespijker:
If we want a 5.7 Enterprise with the CAD plugin bundled yes

jellespijker:
But this isn't needed for the beta ;-)

wawanbreton:
Ok, so maybe merge it after 5.7 release ? I'm a bit scared of side-effects, so if we don't need it yet, I'd prefer to wait.

jellespijker:
I would just merge this for the stable release,  otherwise you have the same weird version issue we had with 5.6.0 and 5.6.0+Enterprise versions. Which was a hack.

By not merging it for the creation of 5.7.0 conan release package, you're basically adding a new feature later.

imo the risk is limited

wawanbreton (approved):
Well, let's do it then

PR #18568, opened 2024-03-11
Changed names of materials as per materials list

saumyaj3:
Changed names of materials as per materials list
Removed materials with GUID not found in materials

CURA-11710


casperlamboo (approved):

casperlamboo:
Why removal of `im-pla` and `nylon-cf`?

saumyaj3:
> Why removal of `im-pla` and `nylon-cf`?

I couldn't find the GUID in materials, so they were redundant. I remember previously when I created them I copied them from makerbot guid maps

PR #18587, opened 2024-03-13
[CURA-11521] Fix last line when previewing active simulation.

rburema:
Warning: While this does properly work and can go into prod... Currently, this does add an _entire_ extra batch to the rendering for just the last layer. So this also may be something of a starter-PR than my colleagues who wanted a fun ticket might riff off of ;-)

In any case, the original problem was that since we switched to making the simulation time-based (at least from a users perspective), longer lines would wait, then just show up after the head has gone completely over them.

In this PR the last line is drawn up until the head position instead (as opposed to just not showing it).


plugins/SimulationView/SimulationPass.py
```diff
@@ -143,6 +144,10 @@ def render(self):
                 if self._layer_view.getCurrentLayer() > -1 and ((not self._layer_view._only_show_top_layers) or (not self._layer_view.getCompatibilityMode())):
                     start = 0
                     end = 0
+                    vertex_before_head = not_a_vector
+                    vertex_after_head = not_a_vector
+                    last_to_next_ratio = 0.0
```
saumyaj3, line 149:
```suggestion
                    vertex_distance_ratio = 0.0
```

plugins/SimulationView/SimulationPass.py
```diff
@@ -168,6 +175,8 @@ def render(self):
                                                    polygon.data[index + 1][2])
                                     vec = pos_a * (1.0 - ratio) + pos_b * ratio
                                     head_position = vec + node.getWorldPosition()
+                                    vertex_after_head = pos_b
+                                    towards_next_vertex = 2
```
saumyaj3, line 179: maybe a small comment here why we add 2 as towards next vertex.

plugins/SimulationView/SimulationPass.py
```diff
@@ -159,6 +164,8 @@ def render(self):
                                 ratio = self._layer_view.getCurrentPath() - math.floor(self._layer_view.getCurrentPath())
                                 pos_a = Vector(polygon.data[index][0], polygon.data[index][1],
                                                polygon.data[index][2])
+                                vertex_before_head = pos_a
+                                last_to_next_ratio = ratio
```
saumyaj3, line 168:
```suggestion
                                vertex_distance_ratio = ratio
```

plugins/SimulationView/SimulationPass.py
```diff
@@ -203,6 +217,17 @@ def render(self):
                     current_layer_batch.addItem(node.getWorldTransformation(), layer_data)
                     current_layer_batch.render(self._scene.getActiveCamera())
 
+                    # Last line may be partial
+                    if vertex_after_head != not_a_vector and vertex_after_head != not_a_vector:
+                        self._layer_shader.setUniformValue("u_last_vertex", vertex_before_head)
+                        self._layer_shader.setUniformValue("u_next_vertex", vertex_after_head)
+                        self._layer_shader.setUniformValue("u_last_line_ratio", last_to_next_ratio)
```
saumyaj3, line 224:
```suggestion
                        self._layer_shader.setUniformValue("u_last_line_ratio", vertex_distance_ratio)
```

saumyaj3 (approved):

plugins/SimulationView/SimulationPass.py
rburema, line 179:
```suggestion
                                    towards_next_vertex = 2  # Add two to the index to print the current and next vertices as an 'unfinished' line (to the nozzle).
```

saumyaj3 (approved):

PR #18603, opened 2024-03-14
Added Cura.MachineSelector in place of combobox

saumyaj3:
# Description

UCP | Trigger area for mouse over printers list drop down does not work consistently


plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -103,12 +103,17 @@ UM.Dialog
                     comboboxTitle: catalog.i18nc("@action:label", "Open With")
                     comboboxTooltipText: catalog.i18nc("@info:tooltip", "Printer settings will be updated to match the settings saved with the project.")
                     comboboxVisible: workspaceDialog.visible && manager.updatableMachinesModel.count > 1
```
casperlamboo, line 105: Now that the `combobox` property isn't being used anymore any logic that was previously handled through the `WorkspaceSection` component doesn't work anymore. The solution implemented here doesn't work for me
```
2024-03-18 11:30:56,354 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [448]: file:///Users/c.lamboo/ultimaker/Cura/plugins/3MFReader/WorkspaceDialog.qml:122: ReferenceError: comboboxVisible is not defined
```
Do we know why the behavior is different when the component is passed via the `combobox` property? I would solve a lot of duplicate code, things like positioning and visibility are already handled in that component which is now implemented twice.

saumyaj3: I believe the two mouse area of combobox and machineselector creates a conflict of interest thats why we get this issue. We still use combobox below in the code for materialResolveComboBox

plugins/3MFReader/WorkspaceSection.qml
```diff
@@ -15,7 +15,11 @@ Item
     property Component content: Item { visible: false  }
 
     property alias comboboxTitle: comboboxLabel.text
-    property Component combobox: Item { visible: false }
+    property Component combobox: Item
+        {
+            z: -1
```
casperlamboo, line 20: this isn't really needed. The placeholder is an invisible `Item` conaining no style, and no children. If it's not rendered anyway no need to place it on top

saumyaj3: [eac2d3b](https://github.com/Ultimaker/Cura/pull/18603/commits/eac2d3b173e0819bddd8fd68e132932e9ad5617e)

casperlamboo (approved):

PR #18606, opened 2024-03-14
[PP-452][CURA-11786] Clean up the CloudAPIClient

Frederic98:
## Cleaned up CloudAPIClient.py
- The `method_x` machine id to DF cluster name dictionary in the `getClustersByMachineType` method was hard-coded in the python module. This is now moved to a separate json file.
- str.format() converted to f-string
- Properly formatted some function calls, and a function definition

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
@@ -5,6 +5,7 @@ import urllib.parse
 from json import JSONDecodeError
 from time import time
 from typing import Callable, List, Type, TypeVar, Union, Optional, Tuple, Dict, Any, cast
+from pathlib import Path
 
 from PyQt6.QtCore import QUrl
 from PyQt6.QtNetwork import QNetworkRequest, QNetworkReply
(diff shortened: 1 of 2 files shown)


plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
```diff
@@ -30,6 +31,14 @@
 """The generic type variable used to document the methods below."""
 
 
+try:
```
nallath, line 34: I think it would be slightly better to do this in the same way as it's done for the xml_material_profile

Right now this will load the file when the cloudapi.py is imported, instead of first time that the data is needed.

plugins/UM3NetworkPrinting/src/Cloud/machine_id_to_name.json
```diff
@@ -0,0 +1,5 @@
+{
+    "ultimaker_method": "MakerBot Method",
+    "ultimaker_methodx": "MakerBot Method X",
+    "ultimaker_methodxl": "MakerBot Method XL"
+}
```
rburema, line 5: Maybe make use of the `product_to_id.json` mapping that already exists?

rburema: Never mind, you need it separately here.

rburema (approved):

rburema:
I still have to test it and cherry-pick it to the 5.7 branch, but otherwise the code looks fine.

rburema:
Closed in favour of the cherry-picked version, here: #18741

PR #18618, opened 2024-03-15
Added learn more button while saving and opening UCP

saumyaj3:
CURA-11734

Added learn more button in UCP


wawanbreton (approved):
I made a small commit to add a blank line on the QML code.
Also I'm not a very big fan of the position of the buttons, but that's personal preference. I'll let QA decide whether this is good enough for the release.

PR #18614, opened 2024-03-15
Updated translations

wawanbreton:
These translations have been made using Lokalise for the first time. The process is still very experimental, and as this is the first time, the translations files have probably been entirely been rewritten. Please make sure that:
* Languages have not been mixed
* Translations show in the UI

CURA-11732

PR #18659, opened 2024-03-18
Update Italian Language

GMagician:
Added some missing messages

# Description

This translate some Italian messages

## Type of change

- [ ] Translations

### DIFF
--- a/resources/i18n/it_IT/cura.po
+++ b/resources/i18n/it_IT/cura.po
@@ -3792,11 +3792,11 @@ msgstr "Salva progetto Cura"
 
 msgctxt "@option"
 msgid "Save Cura project and .makerbot print file"
-msgstr "Salva progetto progetto Cura e file di stampa .makerbot"
+msgstr "Salva progetto Cura e file di stampa .makerbot"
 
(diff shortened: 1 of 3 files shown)


wawanbreton:
Hi @GMagician, and thanks for the work you provided !
However, I'm a bit embarrassed, because in the meantime we also got our own Italian translations processed by professional translators :grimacing: They are available on [this PR](https://github.com/Ultimaker/Cura/pull/18614) which should be merged very soon.
On the other hand, you probably have a better technical understanding of the software than the translators. Would you like to make a review of the changes they provided, or proceed an other way ?

GMagician:
> Hi @GMagician, and thanks for the work you provided ! However, I'm a bit embarrassed, because in the meantime we also got our own Italian translations processed by professional translators 😬 They are available on [this PR](https://github.com/Ultimaker/Cura/pull/18614) which should be merged very soon. On the other hand, you probably have a better technical understanding of the software than the translators. Would you like to make a review of the changes they provided, or proceed an other way ?

Yep, I'll read the PR and post my opinion. Of course professional translators may have care that I may lack

PR #18737, opened 2024-03-26
Add visibility control for settings in 3MFWriter plugin

saumyaj3:
Introduced a new property 'show_in_menu' to control the visibility of individual settings in 3MFWriter plugin. Modified 'SettingExport' class constructor to store this visibility status. Also, updated the 'SettingsExportGroup' model to filter the settings based on visibility before presenting them in the GUI.

CURA-11723

### DIFF
--- a/plugins/3MFWriter/SettingsExportGroup.py
+++ b/plugins/3MFWriter/SettingsExportGroup.py
@@ -36,9 +36,7 @@ class SettingsExportGroup(QObject):
     @pyqtProperty(list, constant=True)
     def visibleSettings(self):
         if self._visible_settings == []:
-            for item in self._settings:
-                if item.isVisible:
-                    self._visible_settings.append(item)
+            self._visible_settings = list(filter(lambda item : item.isVisible, self._settings))
         return self._visible_settings
 
     @pyqtProperty(int, constant=True)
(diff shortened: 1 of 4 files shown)


plugins/3MFWriter/SettingsExportModel.py
```diff
@@ -117,6 +117,9 @@ def _exportSettings(settings_stack):
         is_exportable = any(key in SettingsExportModel.PER_MODEL_EXPORTABLE_SETTINGS_KEYS for key in user_keys)
 
         for setting_to_export in user_keys:
+            show_in_menu = True
+            if setting_to_export  in SettingsExportModel.PER_MODEL_EXPORTABLE_SETTINGS_KEYS:
+                show_in_menu = False
```
casperlamboo, line 122: expression can be simplified to
```suggestion
            show_in_menu = setting_to_export not in SettingsExportModel.PER_MODEL_EXPORTABLE_SETTINGS_KEYS
```

plugins/3MFWriter/SettingsExportGroup.py
```diff
@@ -32,6 +33,14 @@ def name(self):
     def settings(self):
         return self._settings
 
+    @pyqtProperty(list, constant=True)
+    def visibleSettings(self):
+        if self._visible_settings == []:
+            for item in self._settings:
+                if item.isVisible:
+                    self._visible_settings.append(item)
```
casperlamboo, line 41: I have a slight preference for writing this as the following as it is more concise
```suggestion
            self._visible_settings = list(filter(lambda item : item.isVisible, self._settings))
```
however, I get that that is personal preference.

PR #18743, opened 2024-03-27
Update support extruder settings behavior

saumyaj3:
Changed the settings in RecommendedSupportSelector.qml so that now all extruders will not be updated by default. However, when a user changes the selected index, a temporary update to all extruders will be triggered. This allows for specific user-directed changes to propagate correctly across all extruders.

CURA-11776

### DIFF
--- a/resources/qml/PrintSetupSelector/Recommended/RecommendedSupportSelector.qml
+++ b/resources/qml/PrintSetupSelector/Recommended/RecommendedSupportSelector.qml
@@ -56,7 +56,7 @@ RecommendedSettingSection
                 width: parent.width
                 settingName: "support_structure"
                 propertyRemoveUnusedValue: false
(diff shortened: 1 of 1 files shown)


wawanbreton:
Any idea why this is required for `support_structure` and not for `support_type` ? I was afraid it would suffer the same bug, but apparently not.

saumyaj3:
> Any idea why this is required for `support_structure` and not for `support_type` ? I was afraid it would suffer the same bug, but apparently not.

I believe this is the case because support_structure depend on the material you use(so extruder) but same is not true in case of support type. We have to take this into consideration that while changing support extruder(not active extruder) correct support structure is updated. But updates global stack while changing it directly.

wawanbreton:
Ok, makes sense indeed. It is a quite complex situation :grimacing:

PR #18744, opened 2024-03-27
[CURA-8897] Clear the last project name when the platform is empty.

rburema:
This should fix the reverting of project name to the previous one when the current project was not saved as a project (yet).

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1459,9 +1459,9 @@ class CuraApplication(QtApplication):
             self.sceneBoundingBoxChanged.emit()
 
         if count > 0:
-            ProjectOutputDevice.setLastOutputName(None)
             self._platform_activity = True
         else:
+            ProjectOutputDevice.setLastOutputName(None)
             self._platform_activity = False
         self.activityChanged.emit()
 
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -33,6 +33,7 @@ from UM.Message import Message
 from UM.Operations.AddSceneNodeOperation import AddSceneNodeOperation
 from UM.Operations.GroupedOperation import GroupedOperation
 from UM.Operations.SetTransformOperation import SetTransformOperation
+from UM.OutputDevice.ProjectOutputDevice import ProjectOutputDevice
 from UM.Platform import Platform
 from UM.PluginError import PluginNotFoundError
 from UM.Preferences import Preferences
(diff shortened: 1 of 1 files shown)


cura/CuraApplication.py
```diff
@@ -1457,7 +1458,11 @@ def updatePlatformActivity(self, node = None):
             self._scene_bounding_box = scene_bounding_box
             self.sceneBoundingBoxChanged.emit()
 
-        self._platform_activity = True if count > 0 else False
+        if count > 0:
+            ProjectOutputDevice.setLastOutputName(None)
```
wawanbreton, line 1462: I'm a bit surprised to find it here. Wouldn't it be better in `deleteAll` ?

rburema: I had it in delete all first, but that doesn't cover the ctrl-A, delete (select all, delete) use-case. Since that only indirectly deletes all (doesn't go via that function) since you only happen to have all selected.

wawanbreton: Ok, I could argue that making a `ctrl-A + delete` doesn't have the same meaning than making a "clear build plate" but I agree that in the current workflow, most users would expect that.

wawanbreton (approved):

cura/CuraApplication.py
rburema: If you where to argue that I can could counter that with that what we _actually_ want here is not checking for 'delete all', but 'are there any meshes on the buildplate that might be a reference to the previously opened project/thing', which would be true when there is at least on mesh still there. 😉 

(Also 'activity' might be a bit of a misnomer here, as far as I can see, the 'activity' is just there being anything on the buildplate ... but we can't really rename that function.)

wawanbreton: It took me some time to figure out the meaning of this function indeed. But as you can, we can't really rename it.

And your other argument is also very true, so we can definitely keep this behavior.

smartavionics, line 1462: I'm not familiar with this part of Cura but I wonder if the call to setLastOutputName(None) should be done when count is 0 rather than when count > 0 ?

rburema: I think you're right -- I'm not _entirely_ sure why it also seems to work when the current code is run then.

rburema: Fortunately, the fix still works the other way around. I changed it to the code that makes more sense.

PR #18745, opened 2024-03-27
Implement syncing feature for Ultimaker account

saumyaj3:
CURA-11465

# Description
Updated account syncing functionality for logged-in and non-logged-in sessions in Ultimaker. More specifically, syncing is now engaged when the UI becomes visible and stopped when adding a machine action-> addmachinedialog is triggered. The modifications also include fallback to login if the user is not logged in.

### DIFF
--- a/resources/qml/WelcomePages/AddUltimakerPrinter.qml
+++ b/resources/qml/WelcomePages/AddUltimakerPrinter.qml
@@ -133,7 +133,7 @@ Control
                         text = catalog.i18nc("@button", "Waiting for new printers")
                         busy = true;
                         enabled = false;
-                        Cura.API.account.login();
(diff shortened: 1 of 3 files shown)


casperlamboo (approved):

resources/qml/WelcomePages/AddUltimakerPrinter.qml
```diff
@@ -133,7 +133,7 @@ Control
                         text = catalog.i18nc("@button", "Waiting for new printers")
                         busy = true;
                         enabled = false;
-                        Cura.API.account.login();
+                        Cura.API.account.isLoggedIn? Cura.API.account.sync():Cura.API.account.login();
```
fieldOfView, line 136: Wouldn't an `if`/`else` statement have been much more readable?

PR #18756, opened 2024-03-28
[CURA-11703] Reset entire scene around center for UCP.

rburema:
Since we can't rely on the build-volume --because we can't know on which printer we open-- the safest bet (and the thing allowed by 3mf if I recall) is to just center the entire scene on the buildplate for Universal Cura Project files.

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -10,6 +10,8 @@ from typing import cast, Dict, List, Optional, Tuple, Any, Set
 
 import xml.etree.ElementTree as ET
 
+from UM.Math.AxisAlignedBox import AxisAlignedBox
+from UM.Math.Vector import Vector
 from UM.Util import parseBool
 from UM.Workspace.WorkspaceReader import WorkspaceReader
 from UM.Application import Application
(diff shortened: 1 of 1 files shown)


plugins/3MFReader/ThreeMFWorkspaceReader.py
```diff
@@ -936,6 +938,24 @@ def read(self, file_name):
         if nodes is None:
             nodes = []
 
+        if self._is_ucp:
+            # We might be on a different printer than the one this project was made on.
+            # The offset to the printers' center isn't saved; instead, try to just fit everything on the buildplate.
+            full_extents = None  # Don't initialize to 'Null'!
```
casperlamboo, line 944: I don't quite get this comment, python has no `null` value, and it's closest equivalent is `None`. and the value is initialized to `None`
```suggestion
            full_extents = None
```

casperlamboo (approved):

plugins/3MFReader/ThreeMFWorkspaceReader.py
rburema: I'm fine with removing it, but for posterities sake:

The code didn't work when initializing with an 'empty' AABB, which _should_ have been set up so that this _does_ work, at least in my opinion. _Inside_ the AABB the vectors where set to a constant called 'Null' (yes, not 'None'), which is 0,0,0 (should have been 'infinite in the other direction' for min/max).

Anyway, I get why my comment was insufficient to communicate that 😅

PR #18790, opened 2024-04-01
Make `Experimental` sub-settings indented to improve visual usability

Sophist-UK:
# Description

This PR simply makes the `Experimental` sub-settings indented so that it is more readable and the location of these settings is more easily identified.

Bridging and Adaptive Layers were done in separate earlier PRs.

This is the twelfth of several PRs to do this for other groups of settings.

## Type of change

# How Has This Been Tested?

I have made this same update in my installation of Cura 5.6.0 and tested that the changes work.

**Test Configuration**:

* Windows 10
* Cura 5.6.0
* File: C:\Program Files\UltiMaker Cura 5.6.0\share\cura\resources\definitions\fdmprinter.def.json

# Checklist:

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -7603,36 +7603,39 @@
                     "enabled": "(support_enable  or support_meshes_present) and support_pattern == 'zigzag'",
                     "limit_to_extruder": "support_infill_extruder_nr",
                     "settable_per_mesh": false,
-                    "settable_per_extruder": true
-                },
-                "support_skip_zag_per_mm":
-                {
-                    "label": "Support Chunk Size",
-                    "description": "Leave out a connection between support lines once every N millimeter to make the support structure easier to break away.",
-                    "type": "float",
-                    "unit": "mm",
-                    "default_value": 20,
-                    "minimum_value": "0",
-                    "minimum_value_warning": "support_line_distance",
-                    "enabled": "(support_enable or support_meshes_present) and support_pattern == 'zigzag' and support_skip_some_zags",
-                    "limit_to_extruder": "support_infill_extruder_nr",
-                    "settable_per_mesh": false,
                     "settable_per_extruder": true,
                     "children":
                     {
-                        "support_zag_skip_count":
+                        "support_skip_zag_per_mm":
                         {
-                            "label": "Support Chunk Line Count",
-                            "description": "Skip one in every N connection lines to make the support structure easier to break away.",
-                            "type": "int",
-                            "default_value": 5,
-                            "value": "0 if support_line_distance == 0 else round(support_skip_zag_per_mm / support_line_distance)",
-                            "minimum_value": "1",
-                            "minimum_value_warning": "3",
+                            "label": "Support Chunk Size",
+                            "description": "Leave out a connection between support lines once every N millimeter to make the support structure easier to break away.",
+                            "type": "float",
+                            "unit": "mm",
+                            "default_value": 20,
+                            "minimum_value": "0",
+                            "minimum_value_warning": "support_line_distance",
                             "enabled": "(support_enable or support_meshes_present) and support_pattern == 'zigzag' and support_skip_some_zags",
                             "limit_to_extruder": "support_infill_extruder_nr",
                             "settable_per_mesh": false,
-                            "settable_per_extruder": true
+                            "settable_per_extruder": true,
+                            "children":
+                            {
+                                "support_zag_skip_count":
+                                {
+                                    "label": "Support Chunk Line Count",
+                                    "description": "Skip one in every N connection lines to make the support structure easier to break away.",
+                                    "type": "int",
+                                    "default_value": 5,
+                                    "value": "0 if support_line_distance == 0 else round(support_skip_zag_per_mm / support_line_distance)",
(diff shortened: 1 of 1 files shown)


wawanbreton:
Hi @Sophist-UK, and thank you for all these PRs :sweat_smile: 
I think they really make sense together, because it kinds of introduces a new way of globally organizing the settings in Cura.
Would it be easy for you to make a single PR for all this changes ? If not, I can merge them all in a single branch to test this as a whole, then accept them individually. But i'm also a bit afraid of git conflicts because this all changes the same file (although at different places).

wawanbreton:
Well, I'm a bit embarrassed now. I was told that having "children" settings is not just a visual thing but will also change the way the settings are resolved. So, I do like your idea very much, but I can't accept those PRs as they would probably create a lot of weird bugs :smiling_face_with_tear:
On the other hand, we do acknowledge for the need to reorganize the settings to make them clearer, so we will try to work and make changes in that direction.

Sophist-UK:
I was careful to avoid making multiple changes on the same parts of the file.

I am unclear why what is only a visual change should have any impact on the code. I did review the code where Cura generates the settings it passes to CuraEngine and couldn't find anything that filters the settings by their hierarchical position, nor could I see anything in CuraEngine that would be impacted by only hierarchical changes in Cura.

In the end this file is interpreted by Qt and AFAIK Qt creates the hierarchy - how much Cura even knows about the hierarchy I am unclear about. The settings themselves are associated with the name and not the position.

I did test each of these PRs visually on my 5.6 installation - I can probably generate Gcode against this file both before and after using a reasonably complex file to see whether the gcode changes which would give some general confidence that the changes won't have any impact, but I cannot check each and every variations of such a large number of settings.

Sophist-UK:
I have now found where and how `children` nodes are processed at https://github.com/Ultimaker/Cura/blob/549d107ab12da6a613c96bd49b50457d9429d37a/printer-linter/src/printerlinter/linters/defintion.py#L161-L165

All that seems to happen here is that the children are processed as if they were root nodes and added to the settings definitions which itself has no parent/children setting hierarchy.

Obviously the same `fdmprinters.def.json` file is also used to drive the UI, but that seems to be entirely separate code to that which manages the settings themselves.

nallath:
It's even a bit more complicated than that in the sense that the "parent" settings are _never_ used by the engine. The engine simply _does not_ have a setting for line_width. It only has all the child settings that are specializations for it. 

An easy example (as it only has two child settings) is "Support Interface Line Width". The engine doesn't use this setting, as it's only there as an convenience to the user, as users often want to set both "support roof line with" and "support floor line width" in a single go. 

We manage to do that by giving each child setting a value that depends on it's children. In some cases it's a direct inheritance, sometimes it's a formula. But if the user overrides the value, the formula (and thus the link) gets lost. Back in the early days, users then got frustrated that changing the "top" setting (eg "Support Interface Line Width") would have no effect (as both child settings no longer have a formula but a fixed value). In order to make that a tad bit clearer, we automatically disable the setting.

I don't know from the top of my mind where this code exactly lives (it's been years since we've made this and had to touch it), but if i remember correctly, we also call them "descendants".

Sophist-UK:
Actually it is not any more complicated at all. 

Firstly AFAIK all settings are passed to the engine and in fact I checked the code and the engine actually DOES use parent some settings. In the cases you speak of, the Engine code just doesn't need the parent settings because it uses the child settings that are derived from the parent (unless overridden) by the value property reference to the parent name. That does not AFAIK mean that the parent settings are not available to the Engine - AFAIK the real situation is that "The engine simply does ~not~ have a setting for line_width. It ~only has all~ *chooses to use only* the child settings that are specializations for it."

wawanbreton:
> the real situation is that "The engine simply does ~not~ have a setting for line_width. It ~only has all~ _chooses to use only_ the child settings that are specializations for it."

That is probably true indeed. But whether the engine uses the parent setting or not is not really what matters. I'm just afraid that in some situations the **value** that is resolved for the setting will be different. I'll make some checks.

nallath:
It's indeed by convention that the engine should only use the "leaf" settings.

Sophist-UK:
> It's indeed by convention that the engine should only use the "leaf" settings. But i'm not quite sure why that matters? In the current logic the engine should never use a leaf setting and in the few times it did, it caused huge issues (as it lead to settings not having the expected effect). If there are currently cases where it uses a parent setting, this is a bug and should be fixed.

As I previously stated this is **NOT** the case in the way you phrased it. The convention is **not to use parent settings unless they are boolean** - boolean parents are used quite often by the engine.

> Of course we could consider changing this logic, but I don't think that we should do so lightly.

I do NOT believe that there is any logic that needs changing, but if there is then I agree it needs to be done carefully.

nallath:
Huh? Boolean statements? I don't understand what you mean.

Sophist-UK:
I never mentioned "boolean ** statements**". I mentioned boolean parents by which I mean parent values along the lines of  "enable bridging" or similar (I haven't checked the code again to find a valid example, but there are parent settings which are checked by the engine).

It is only when parents have numeric values and these propagate to children through the `value` property that the Engine should use the propagated values and not refer to the parent and this is for code logic reasons and AFAICT not due to any constraint on how settings are passed from Cura to the Engine.

PR #18798, opened 2024-04-02
Values now shown while saving UCP

saumyaj3:
For all the values even numericals were passed through the translations. so nothing came out for numbers. Now we filter it for enums and then translate it.

CURA-11801


wawanbreton (approved):
Ok, looks better :)

PR #18824, opened 2024-04-05
Add detection for deleted files in printer linter

saumyaj3:
This update adds a new check for deleted files in the printer linter. This will alert the user when a file has been deleted that could potentially disrupt upgrade scripts. An argument "--deleted" is also added to terminal.py to facilitate this new check. Additionally, the printer-linter version has been incremented to 0.1.2.

CURA-10903


wawanbreton (approved):
This looks good to me, although I have a very limited understanding of how the printer-linter is organized...

PR #18883, opened 2024-04-12
Clearing operation stack when a new project is loaded

saumyaj3:
CURA-11658

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -2281,8 +2281,8 @@ class CuraApplication(QtApplication):
         return 0
 
     @pyqtSlot()
-    def deleteAll(self, only_selectable: bool = True) -> None:
-        super().deleteAll(only_selectable = only_selectable)
+    def deleteAll(self, only_selectable: bool = True, clear_all:bool = False) -> None:
+        super().deleteAll(only_selectable = only_selectable, clear_all = clear_all)
 
         # Also remove nodes with LayerData
         self._removeNodesWithLayerData(only_selectable = only_selectable)


cura/Settings/SettingOverrideDecorator.py
```diff
@@ -60,6 +60,17 @@ def __init__(self, *, force_update = True):
         if force_update:
             self._updateNextStack()
 
+    def clear(self) -> None:
```
wawanbreton, line 63: Is this the most common naming for such a method in Python ? I know other languages use e.g. `release` or `dispose`.

wawanbreton (changes requested):
I'm a bit afraid that this change can create crashes in Cura. Those variables may be expected to be non-Null when calling some methods on the objects, and the objects are actually still referenced and callable. I would go for only clearing the references that actually hold a lot of memory (mesh data) and don't touch the rest.

cura/Settings/SettingOverrideDecorator.py
saumyaj3: I changed the name of clear to clearDecoratorData. I believe this gives a better naming for it.

wawanbreton (approved):
I'm still afraid of potentional crashes, e.g. when setting `self._global_stack` to `None`, this object is being called in every method without a check :grimacing: but I let you be the final judge of this. Or maybe have a second opinion from @nallath ?

cura/CuraApplication.py
```diff
@@ -2281,8 +2281,8 @@ def appHeight(self) -> int:
         return 0
 
     @pyqtSlot()
-    def deleteAll(self, only_selectable: bool = True) -> None:
-        super().deleteAll(only_selectable = only_selectable)
+    def deleteAll(self, only_selectable: bool = True, clear_all:bool = False) -> None:
```
nallath, line 2284:
```suggestion
    def deleteAll(self, only_selectable: bool = True, clear_all:bool = False) -> None:
```

We should guard against bool traps here. You can do that by by adding a param that eats up all unnamed arguments.

nallath:
I'm not quite sure how this would even prevent the memory that much?

PR #18916, opened 2024-04-19
Add support z seam settings

saumyaj3:
This update includes two new settings in the expert visibility profile: 'support_z_seam_away_from_model' and 'support_z_seam_min_distance'. These allow users more control over the spatial relationship between the z seam of the support structure and the actual 3D model, ensuring a smoother post-printing process

CURA-11227


plugins/SimulationView/layers3d.shader
```diff
@@ -356,7 +356,10 @@ geometry41core =
             EndPrimitive();
         }
 
-        if ((u_show_starts == 1) && (v_prev_line_type[0] != 1) && (v_line_type[0] == 1)) {
+        if ((u_show_starts == 1) && (
+            ((v_prev_line_type[0] != 1) && (v_line_type[0] == 1)) ||
+            ((v_prev_line_type[0] != 4) && (v_line_type[0] == 4))
+            )) {
```
casperlamboo, line 362: Why is this shader code added to the PR, seems unrelated?

resources/definitions/fdmprinter.def.json
```diff
@@ -4775,6 +4775,35 @@
...
+                    {
+                        "support_z_seam_min_distance":
+                        {
+                            "label": "Min Z Seam Distance from Model",
+                            "description": "The distance between the model and its support structure at the z-axis seam.",
+                            "unit": "mm",
+                            "minimum_value": "0.001",
+                            "value": "line_width * 2",
+                            "default_value": 0.8,
+                            "maximum_value": "line_width * 4",
```
casperlamboo, line 4798: I wouldn't add a max value here, we block the user from slicing if the value is set too high. But there is nothing dangerous with actually slicing with such a value, it won't damage the printer or anything. I think just a value warning suffices here.

plugins/SimulationView/layers3d.shader
saumyaj3: This shader code is for showing the start of support. basically showing seams in support

resources/definitions/fdmprinter.def.json
casperlamboo: If it doesn't work for "large" values it will also break for "small" values for some weird support structures. The code should be robust against that.

saumyaj3: It just doesn't find the seam location in that case, with not working thats what I meant.

saumyaj3: However I understand removing the max value, as the slice can still happen (y)

casperlamboo: > It just doesn't find the seam location in that case, with not working thats what I meant. It just place the seam at original location. The shortest distance

Hense the value warning

plugins/SimulationView/layers3d.shader
casperlamboo: I'm a bit puzzled to why this is needed all of a sudden?

saumyaj3: Since we added support seam, it is only fair that we show them in front end.

resources/definitions/fdmprinter.def.json
saumyaj3: Done here: https://github.com/Ultimaker/Cura/pull/18916/commits/f3172cb10ba55f025f1f3bd9b2de5b35c19b0169

PR #18945, opened 2024-04-24
Add new linter for long profile names

saumyaj3:
new linter for long profile names
CURA-11153


casperlamboo (approved):

printer-linter/src/printerlinter/linters/profile.py
```diff
@@ -1,9 +1,42 @@
+import re
 from typing import Iterator
```
HellAholic, line 2:
```suggestion
from typing import Iterator, Tuple
```

PR #19025, opened 2024-05-06
Fix for slow file opening

### LINKED ISSUE #16249: Unable to open any .stl files at all.
### Cura Version

5.4.0

### Operating System

Mac Catalina

### Printer

Cr-10S

### Reproduction steps

Cura 5.4.0 beta and fresh installed 5.4.0 will not open any .stl files.  I've tried at least 15 of them, and nothing ever appears on the virtual build-late.  I've tried uninstalls, new installs, restarts, new .stl's from tinkercad, opening .stl's from inside cura, changing default app to Cura... all of the .stl's will still open on prusaslicer.

### Actual results

No Stl will appear on the virtual print bed.

### Expected results

stl should appear on print bed

### Add your .zip and screenshots here ⬇️

na

gtollini:
# Description

This fixes #16249 .

After looking at my logs (pretty much identical to [this](https://github.com/Ultimaker/Cura/issues/16249#issuecomment-2045990821)), I decided to take a deeper look into the code. As it turns out, 32 tty ports were being tested, 4 at a time, with 15s timeout, twice each. This would give me an expected wait of 32 * 15 * 2 / 4 = 240s of wait, or about 4 minutes. My logs actually match this expected time, so it took Cura 4 minutes between me selecting a file to open and it actually opening on Cura.

Ideally, this wouldn't be a hard-coded ```True```. But I'm not that familiar with Qt - or the Cura project as a whole - so I wouldn't know the best way to handle this. Some suggestions are:

- Add a boolean to the configurations, something like "Auto Detect Serial Ports" or "Auto Detect Baud Rates". And please, make the default ```False```.
- Instead of scanning for tty ports and trying every baud rate, have the user supply this data. 

## Type of change

# How Has This Been Tested?

I had the same problem as listed on #16249. After building with the new source code, it opened files quickly - as expected. 

# Checklist:

### DIFF
--- a/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
+++ b/plugins/USBPrinting/USBPrinterOutputDeviceManager.py
@@ -84,8 +84,8 @@ class USBPrinterOutputDeviceManager(QObject, OutputDevicePlugin):
             if container_stack.getMetaDataEntry("supports_usb_connection"):
                 machine_file_formats = [file_type.strip() for file_type in container_stack.getMetaDataEntry("file_formats").split(";")]
                 if "text/x-gcode" in machine_file_formats:
-                    # We only limit listing usb on windows is a fix for connecting tty/cu printers on MacOS and Linux
-                    port_list = self.getSerialPortList(only_list_usb=Platform.isWindows())
+                    # We only list usb, otherwise scanning all serial ports might take too long
+                    port_list = self.getSerialPortList(only_list_usb=True)
             self._addRemovePorts(port_list)
             time.sleep(5)
 


gtollini:
> Could this have had any relation with the following? #19186

Thanks for your comment, Kafufafu. I've added the two related issues to the list issues this PR fixes.

sneakythr0ws:
merge it pls!!!

IT-Wizard:
I've tested the new 5.8 version, it is still not implemented.
The load of an STL file is still 4 minutes long. :o/

wawanbreton:
Hi, we do have some update on this: we have finally decided to disable the USB printing plugin by default. However, in order not to break the workflow of users that actually use it, this will be done in two steps:
* Cura 5.10 will save a (hidden) preference that will be enabled only if you do use USB printing at some point
* In the next version, this preference will be used to activate the USB printing plugin or not

This way, the USB printing should be automatically enabled for those who need it, and disabled for those who don't. We also intend to make the preference visible when adding a compatible printer, so that new users should also find it.

So this will not be fixed in 5.10, but will be in the version after. In the meantime, I invite you to disable the USB printing plugin if you don't use it:

image

PR #19029, opened 2024-05-06
Cura 11777 yeet sketch to makerbot

saumyaj3:
# Description

At the moment the Method Sketch printer(not sketch large) is added in Cura in this branch and corresponding changes are done to add the abstract printer from the cloud as well. Makerbot plugin is extended to add the "print.gcode" file instead of "print.jsontoolpath" for the mimetype : **application/x-makerbot-sketch**
The flavor for Griffin works with the Sketch printers.


resources/definitions/ultimaker_sketch.def.json
```diff
@@ -0,0 +1,259 @@
...
+        "machine_max_feedrate_x": { "default_value": 300 },
+        "machine_max_feedrate_y": { "default_value": 300 },
+        "machine_max_feedrate_z": { "default_value": 40 },
+        "machine_min_cool_heat_time_window": { "value": "15" },
+        "machine_name": { "default_value": "Makerbot Sketch" },
+        "machine_nozzle_cool_down_speed": { "default_value": 0.8 },
+        "machine_nozzle_heat_up_speed": { "default_value": 1.4 },
+        "machine_start_gcode": { "default_value": "M140 S50 T0; Set Platform Temp\nM104 S220 T0; Set Extruder Temp\nG90; Use Absolute Positioning\nG28; Home\nM132 X Y Z A B; Set Current Position to Home\nG161 X Y F3300; Move to min axes positions\nM7 T0; Wait For Platform to Heat\nM6 T0; Wait For Extruders to Heat\nM651; Turn on back fan\nM907 X100 Y100 Z40 A80 B20; Set Stepper Currents\nM106; Enable Cooling Fan\n; Purge Line\nG92 E0; Reset Extruder Axis Position\nG1 X-26.18 Y-75.90 Z0.200 F420\nG1 X26.18 Y-75.90 E10\nG92 E0; Reset Extruder Axis Position\n; Start GCode\n" },
+        "machine_width": { "default_value": 150 },
+        "material_bed_temperature": { "maximum_value": "115" },
```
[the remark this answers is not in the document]
saumyaj3: e33264625a6ea94cc563a4d0c804d9242352d30a

resources/definitions/ultimaker_sketch.def.json
```diff
@@ -0,0 +1,259 @@
...
+        "variants_name": "Extruder",
+        "weight": -1
+    },
+    "overrides":
+    {
+        "acceleration_enabled": { "value": false },
+        "adhesion_type": { "value": "'raft'" },
+        "brim_width": { "value": "3" },
+        "cool_fan_speed": { "value": "100" },
+        "default_material_print_temperature": { "value": "215" },
```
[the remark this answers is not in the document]
saumyaj3: e33264625a6ea94cc563a4d0c804d9242352d30a

plugins/MakerbotWriter/MakerbotWriter.py
```diff
@@ -92,14 +104,18 @@ def write(self, stream: BufferedIOBase, nodes: List[SceneNode], mode=MeshWriter.
...
             return False
 
-        json_toolpaths = du.gcode_2_miracle_jtp(gcode_text_io.getvalue())
-        metadata = self._getMeta(nodes)
+        if file_format == "application/x-makerbot-sketch":
+            filename, filedata = "print.gcode", gcode_text_io.getvalue()
+            self._PNG_FORMATS = self._PNG_FORMAT
+        else:
+            filename, filedata = "print.jsontoolpath", du.gcode_2_miracle_jtp(gcode_text_io.getvalue())
+            self._PNG_FORMATS = self._PNG_FORMAT + self._PNG_FORMAT_METHOD
```
casperlamboo, line 118:
```suggestion
        elif file_format == "application/x-makerbot-sketch":
            filename, filedata = "print.jsontoolpath", du.gcode_2_miracle_jtp(gcode_text_io.getvalue())
            self._PNG_FORMATS = self._PNG_FORMAT + self._PNG_FORMAT_METHOD
        else:
             raise Exception("Unsupported mime type") 
```
Or even better, use a [match](https://docs.python.org/3/tutorial/controlflow.html#match-statements) statement

saumyaj3: https://github.com/Ultimaker/Cura/pull/19029/commits/fdfc5dcc67f8efac60404421e86b357b5ef3a5af here 👯

plugins/MakerbotWriter/MakerbotWriter.py
```diff
@@ -143,7 +161,7 @@ def _getMeta(self, root_nodes: List[SceneNode]) -> Dict[str, any]:
                         nodes.append(node)
 
         meta = dict()
-
+        file_format = global_stack.definition.getMetaDataEntry("file_formats")
```
casperlamboo, line 164: I get why you did it like this. I had a different solution in mind where we would basically duplicate the whole plugin and each plugin would listen to a different mimetype. This solution is clearly "easier" to implement/understand but it does go a bit against the intended use of plugins. @nallath do you maybe have an idea how to implement this in a better way?

saumyaj3: I believe the other way would be to create a separate plugin for sketch printers. Or if we can somehow use limit mimetypes from the printer definitions? and then have this information to use in plugin?

nallath: Duplicating is perhaps a bit word? I'd say that it's mostly inheritance with a specialization that overrides the mimetype.

casperlamboo: https://github.com/Ultimaker/Cura/pull/19029/commits/21b8694289a89bb5319a9b3ebc0246d80b88076a

PR #19075, opened 2024-05-14
CURA-11902 fix loaded support mesh position

wawanbreton:
Due to commit cdb2a5789d8ffe9ba4e5669d7bbb575b13d30fd6 all the meshes are re-centered when saving to 3MF so that they can be loaded centered, even when only models are imported. However, this should not be applied to objects that are a child of an other object, because it changes their relative position. This PR fixes the issues by only applying the mesh centering to top-level objects, so that we keep the best of both behaviors.

Contributes to CURA-11902
Fixes https://github.com/5axes/SpoonAntiWarping/issues/8

### DIFF
--- a/plugins/3MFWriter/ThreeMFWriter.py
+++ b/plugins/3MFWriter/ThreeMFWriter.py
@@ -112,7 +112,6 @@ class ThreeMFWriter(MeshWriter):
         savitar_node = Savitar.SceneNode()
         savitar_node.setName(um_node.getName())
 
-        node_matrix = um_node.getLocalTransformation()
         mesh_data = um_node.getMeshData()
 
         if center_mesh:
--- a/plugins/3MFWriter/ThreeMFWriter.py
+++ b/plugins/3MFWriter/ThreeMFWriter.py
@@ -96,7 +96,8 @@ class ThreeMFWriter(MeshWriter):
     @staticmethod
     def _convertUMNodeToSavitarNode(um_node,
                                     transformation = Matrix(),
-                                    exported_settings: Optional[Dict[str, Set[str]]] = None):
+                                    exported_settings: Optional[Dict[str, Set[str]]] = None,
+                                    center_mesh = False):
         """Convenience function that converts an Uranium SceneNode object to a SavitarSceneNode
 
         :returns: Uranium Scene node.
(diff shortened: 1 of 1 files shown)


plugins/3MFWriter/ThreeMFWriter.py
```diff
@@ -111,16 +112,21 @@ def _convertUMNodeToSavitarNode(um_node,
...
+            # compensate for original center position, if object(s) is/are not around its zero position
+            if mesh_data is not None:
+                extents = mesh_data.getExtents()
+                if extents is not None:
+                    # We use a different coordinate space while writing, so flip Z and Y
+                    center_vector = Vector(extents.center.x, extents.center.y, extents.center.z)
+                    node_matrix.setByTranslation(center_vector)
+            node_matrix.multiply(um_node.getLocalTransformation())
+        else:
+            node_matrix = um_node.getLocalTransformation()
```
casperlamboo, line 129: Do we. need this else? `node_matrix` is already initialised to `um_node.getLocalTransformation()`

wawanbreton: I ended up removing the first declaration, which removes a useless call in case `center_mesh` is true, and I think it makes the code a bit more readable.

fieldOfView:
Did you mean to merge this into 5.7? Will there be a 5.7.2 for this issue?

wawanbreton:
> Did you mean to merge this into 5.7? Will there be a 5.7.2 for this issue?

Not for this issue specifically, but we have a few bugs that are critical enough to deserve a 5.7.2

5axes:
Not totaly Solved with Cura 5.7.2

Ok additional support or tab elements are at the right place .. but if you save and load the project the part is not at the right place.
image

wawanbreton:
Oh no :astonished:  That's weird, because this is exactly the case I worked on. I must have been missing something. We will probably have to re-open it then.
At least the elements keep their position relative to each other, so it must be possible to re-place the whole group on the build plate, making it almost useable.

5axes:
The problem occurs if the "Automatically drop models to the build plate" option is not active.

HellAholic:
Drop to build plate disabled, Automatically drop models is disabled.
image
image
image
Project saved, opening it again:
https://github.com/Ultimaker/Cura/assets/28710690/5f7453a7-47c9-4b1b-9c3e-3705fa50d29a

5axes:
ok @HellAholic .. not on issue with the Spoon or Tab Anti Warping or Support but with the "Part For Calibration" Plugin.  

No support / or any Tab. Just Add Calibration cube Save and Load.

image

HellAholic:
@5axes I'll add a ticket for investigation on this one 👍

PR #19090, opened 2024-05-17
Update linux.yml

HellAholic:
add self-hosted-Linux-X64 to options

# Description

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -38,6 +38,7 @@ on:
         type: choice
         options:
           - ubuntu-22.04
+          - self-hosted-Linux-X64
 
 jobs:
   linux-installer:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -34,7 +34,7 @@ on:
       operating_system:
         description: 'OS'
         required: true
-        default: 'ubuntu-22.04'
+        default: 'self-hosted-Ubuntu24-X64'
         type: choice
         options:
           - ubuntu-22.04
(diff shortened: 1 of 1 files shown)


wawanbreton (approved):

wawanbreton:
Don't we want to use this runner for other jobs ? We have a lot of good candidates...

HellAholic:
Might get a bit tricky like waiting for another job to finish before you can make a build for example. But it is possible to set up two runners on the same machine... so present your candidates 💯

jellespijker:
> Don't we want to use this runner for other jobs ? We have a lot of good candidates...

as long as those other jobs are all job, that trusted users start, so that malicious code isn't automatically executed

wawanbreton:
I would say at least:
* `Cura-workflows/conan-package-release.yml`: This one we actively wait for it when doing a release, and it could be much faster on our runner.

Other possible candidates:
* `Cura-workflows/conan-package-create-linux.yml`
* `Cura-workflows/unit-test.yml`
* `Cura-workflows/benchmark.yml`

But not sure that can be done safely for them.. The GH documentation says "self-hosted runners should almost never be used for public repositories on GitHub" so there must be a way to do it securely, but I don't want to go into it yet :sweat_smile:

jellespijker:
my advice: only run for the installers, e.q. user trigger, never in an automatic trigger such as Unit Tests, package creations on push etc.

wawanbreton:
Ok, that makes sense. So we can at least use it for `Cura-workflows/conan-package-release.yml`, which is called either manually or by the new release candidate script (which is run manually).

PR #19124, opened 2024-05-23
Reduce permissions on printer-linter GitHub actions

wawanbreton:
Previously they had no permission set, so by default all write access. We now specify only the required permissions for each of them.


jellespijker (approved):
lgtm

PR #19128, opened 2024-05-24
Preventing PVA and BAM to alter temperatures

pkuiper-ultimaker:
PVA and BAM should not change the build plate and build volume temperature settings of the model material.

PP-371


THeijmans (commented):
PVA and BAM will soften in the printhead at build volume temperatures above 40C, I think 40C should be the hard limit. This is a great way of defining the maximum and optimal value for a setting though.

pkuiper-ultimaker:
> PVA and BAM will soften in the printhead at build volume temperatures above 40C, I think 40C should be the hard limit. This is a great way of defining the maximum and optimal value for a setting though.

OK, I will put the maximum build volume temperature of 40C for PVA. I'll keep it 70C for BAM because it needs to work together with PC.

wawanbreton (approved):
Looks good to me, however I had to resolve some git conflicts, so @pkuiper-ultimaker please make a quick check and tell me if something is missing

PR #19150, opened 2024-05-28
Add the new setting for antimatter polarity

wawanbreton:
As advised by a very smart woman who claimed she came from the future and adding this would make the world better within the next 5000+ years.


wawanbreton:
I'll find an other way to save the world, doesn't matter.

PR #19165, opened 2024-05-30
Sketch and Sketch Large profiles

pkuiper-ultimaker:
Print profiles for PLA and T-PLA for the Sketch and Sketch Large.

PP-333


resources/definitions/ultimaker_sketch.def.json
```diff
@@ -1,6 +1,6 @@
 {
     "version": 2,
-    "name": "UltiMaker Sketch",
+    "name": "MakerBot Sketch",
```
wawanbreton, line 3: Please consider renaming the file, this is quite confusing (same for the others)

wawanbreton (commented):
Is the product name really "MakerBot Sketch" ? If so, please ensure the file namings are consistent.
Also note that it will appear in the "UltiMaker B.V." category in Cura, where all the current printers are prefixed with UltiMaker.

resources/definitions/ultimaker_sketch.def.json
pkuiper-ultimaker: Hi Erwan,

Yes, it should be MakerBot Sketch. 
It was changed recently, because the PM & Marketing for education wanted it this way..

wawanbreton: Ok, then the file should be called `makerbot_sketch.def.json`, or do you prefer to keep it this way ?

pkuiper-ultimaker: I prefer it this way. I like to keep the Ultimaker machines as a group. These files are not user facing so it is OK to keep the name ultimaker_sketch.def.json

wawanbreton (approved):

PR #19167, opened 2024-05-31
0.15mm modes for the factor 4 for PVA and BAM.

pkuiper-ultimaker:
Needed by 3rd party material profiles that require support PVA or BAM support.

PP-485


rburema (approved):
PR itself looks fine to me & works (ran into something else while testing, but that doesn't seem to be part of this) -- I'm just wondering what the other 0.15mm materials are on the F4?

PR #19187, opened 2024-06-05
Improve overhang quality T-PLA 0.2mm engineering mode

pkuiper-ultimaker:
Reduce the build plate, build volume and print temperature of the T-PLA 0.2mm engineering mode with 5C to improve the overhang quality. Strength and dimensional accuracy were not affected by this change.

PP-488


rburema (changes requested):
See the comment in slack or in the ticket.

rburema (approved):

PR #19194, opened 2024-06-06
CURA-10769 Automate release action

wawanbreton:
The purpose of this change is to provide new actions to avoid manual operations at some points of the release process. The implemented actions are:
* Feature freeze:
 1. Input the short version number that will be the branch name
 2. Create branches on all appropriate repositories
 3. Set versions of conandata.yml files in the branch on all appropriate repositories
* Release candidate:
 1. Input the complete version number of the release candidate
 2. Set versions of conandata.yml files in the branch on all appropriate repositories
 3. Create a proper tag for the relase candidate (-RC1, -RC2, ...)
 4. Build packages for all dependencies, then for Cura
 5. Build installers for all platforms
 6. Create the release draft containing the formatted changelog and the installer as assets

Technical note:
* The packages building step should in theory be called on the Cura-workflows repository using a workflow_dispatch event, however you can't wait for a remote action execution reliably, because the start action doesn't return any ID. I tried different options/actions and all of them failed to provide a reliable solutions, thus the actions are called directly from here, and the required secret CURA_CONAN_PACKAGE_RELEASE_PAT has been copied to the Cura repository.

See associated PR https://github.com/Ultimaker/cura-workflows/pull/20

CURA-10769


wawanbreton:
Script processing will now obviously fail because it calls other scripts from the `main` branch, where they won't be updated until the branch is actually merged.

resources/conandata.yml
```diff
@@ -0,0 +1 @@
+version: "5.8.0-alpha.0"
```
nallath, line 1: Should this be committed?

nallath (changes requested):

resources/conandata.yml
wawanbreton: Yes, previously the `cura_resources` package used the `conandata.yml` from Cura, which is easier for testing but not for integration. Now it has its own file.

nallath (approved):

PR #19211, opened 2024-06-10
Only request permissions when token changed

nallath:
Since the permissions can only change with a token, we only need to check what the permissions at that moment (instead of every 60 seconds) Fixes CURA-11870


craig-king (approved):

wawanbreton (approved):
Awesome :smile:

PR #19224, opened 2024-06-12
Get last 5.7 fixes

wawanbreton:


wawanbreton:
I actually reverted the version bump of twisted on main, because this is a mistake I made :disappointed: but it doesn't hurt...

rburema:
> I actually reverted the version bump of twisted on main, because this is a mistake I made 😞 but it doesn't hurt...

Ah sorry, I didn't see that 😓 --- like you said, it doesn't hurt at the moment, so I'll leave it in for now.

PR #19227, opened 2024-06-13
CURA-11947 Improve multimaterial raft

wawanbreton:


resources/definitions/fdmprinter.def.json
```diff
@@ -5127,6 +5127,18 @@
                         }
                     }
                 },
+                "support_infill_density_multiplier_layer_0":
+                {
+                    "label": "Support Infill Density Multiplier First Layer",
```
rburema, line 5132: We're calling it 'Initial' layer everywhere else in the English names. Although this is also meant for the raft... but then it's not a single layer anymore.

resources/definitions/fdmprinter.def.json
```diff
@@ -5127,6 +5127,18 @@
...
+                    "description": "Multiplier for the infill on the first layers of the support. Increasing this may help for bed adhesion.",
```
rburema, line 5133: In the same vein: It's probably a bit unclear to the user what is meant by 'first'.

rburema (approved):

resources/definitions/fdmprinter.def.json
wawanbreton: Yeah it's sometimes called "initial" and sometimes "first", so I'm never sure which one to use :smiley: but if you're sure Initial is the official word, I'll change it.

rburema: In the _setting names_ at least, it's initial. In documentation etc. it's sometimes first I think.

wawanbreton: Oops, I though the "official" setting name for things related to first layers was "layer_0", because there are many occurences of this. I'll change it.

rburema (approved):

PR #19283, opened 2024-06-25
CURA-11937 Default layer start position to machine_extruder start position

wawanbreton:
We have 2 settings that are used to setup initial positions in the engine: `machine_extruder_start_pos_x/y` and `layer_start_x/y`. They are used in different contexts, however it would make sense that they are setup accordingly. Thus I added a default value for the `layer_start` position, which is usually not defined, to match `machine_extruder_start` position.

:warning: The settings `layer_start_x/y` are defined in `fdmprinter.def.json`, but the `machine_extruder_start_pos_x/y` are defined in `fdmextruder.def.json`. The implementation currently works, but I'm not sure this is actually a valid setup. An other option would be to use an average position of all the extruders start position (which is already done for UM printers and an other 3rd party).

CURA-11937


rburema (commented):
I think the whole averaging is indeed done to make sure we have a sensible value. I don't think we can rely on the 'raw' value being anything in general.

I'd prefer we go with that approach since we already know that it works.

wawanbreton:
Ok, did that, it is probably safer indeed. And this way we can remove the overrides that had this formula as a definition.

PR #19315, opened 2024-07-02
[CURA-11999] Disable 'temp should be defined in material' warning for now.

rburema:
There was an entire ticket for this specific warning previously (CURA-10904) so I'm disabling it rather than removing it outright.


wawanbreton (approved):
That makes sense. Can you add a comment in the original ticket to note that we did disable it on purpose ?

PR #19323, opened 2024-07-03
Significantly improve the speed of the version upgrader (and boot time!)

nallath:
This knocked about 10 seconds from my boot time. Huzzah :D

The issue was basically that the old style getCFG version was being used. These still used the full cfg config parser which isn't fast. So by removing these functions it becomes a lot faster!


wawanbreton (approved):
Less code, less bugs :laughing: 
Good catch !!

PR #19324, opened 2024-07-04
[CURA-11887] Fuzzy Skin Outside Only: Settings for what's considered 'Outside'.

rburema:
See [the matching backend PR](https://github.com/Ultimaker/CuraEngine/pull/2113) for the full story.


rburema:
> The names and descriptions are a bit hard to read, but I think that's the best we can have 😅

Yeah, there's only two big problems in programming, and they're the halting problem, naming things and off by one errors :-p

> Also shouldn't we set `settable_pre_extruder` to `True`, or is it implied by `settable_per_mesh` ?

I'm not sure if that's implied, can't hurt to add it. - I'll make a commit later.

rburema:
Ah no actually, I remember why I did that; this is the way enabling fuzzy skin also works. So I'm not going to mess with this further (and just focus on the bugs we still have for beta).

wawanbreton:
Oh yes, makes sense. You don't really want to enable it by extruder

PR #19326, opened 2024-07-05
Change default_value to value

wawanbreton:
This is necessary because the parent definition of the settings has changed


rburema (approved):
lgtm!

Issue #7224, opened 2020-03-04
Enlarge Machine Settings Dialog


thesec:
With version 4.5 the start and end gcode's text fields became veeeeery small - see attached screenshot.
Please enlarge / allow scrolling / allow to enlarge the dialog itself ...

![Screenshot 2020-03-04 at 12 14 47](https://user-images.githubusercontent.com/91770/75874542-3633ad00-5e12-11ea-8b06-a479fdb8b4a2.jpg)

tech2077:
This is also the case in 4.5 on MacOS. The window is too small and cannot be resized.


<img width="853" alt="image" src="https://user-images.githubusercontent.com/528634/75908788-7003b400-5e10-11ea-99dc-d24c652f238e.png">

mahtDFR:
Thanks for the report.
@thesec which OS are you running here? MacOS as well?

smartavionics:
I agree the initial size of those gcode boxes is way too small. However, on Linux, it is possible to resize the whole dialog like any other window.

thesec:
@mahtDFR - yes, am running OS X, too.

fieldOfView:
I propose that instead of enlarging the window (which is a sliding slope), the gcode snippets are moved to a different tab, or they are edited using a widget that pops up a dialog.

thesec:
@fieldOfView - please keep the extruder's gcodes on mind when you decide to take it that way to keep it consistent.

nallath:
The issue was that the dialog still had hardcoded values in them (instead of using the dynamic sizing used by the rest of the code).
It also had it's min & max size set to be the same, which linux and windows interpreted as "Resize it to whatever you want" and osx as "No resize allowed". 
I've fixed both, so it shouldn't be needed to move the gcode to a different tab (anyway, I hope it won't be needed, because that's a lot more work)

PR #19333, opened 2024-07-06
Add 3 New UI Settings

TheSin-:
# Description

- Add UI to set `machine_start_gcode_first` so that no gcode is adding infront of the start_gcode.
- Add float input box to set `machine_extruder_change_duration` which added proper extruder change times to the estimated print time.
- Add textbox to set `machine_extruder_prestart_code` to add gcodes that will run pre extruder change, things like heat up so we are at change before with pick up the next extruder.

**Required for https://github.com/Ultimaker/CuraEngine/pull/2114**

## Type of change

- [ ] Translations

# How Has This Been Tested?

UI has been tested, Engine has not yet.

machine_settings
Extruder_settings

**Test Configuration**:
* Operating System: MacOS

# Checklist:


resources/definitions/fdmprinter.def.json
```diff
@@ -511,6 +511,16 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_start_gcode_first":
+                {
+                    "label": "Start GCode must be first",
+                    "description": "Do not add any gcodes before the start gcode block.",
```
nallath, line 517: Try to also include when and why one should use this. 

So i'd personally word it a bit more like

"This setting controls if the start-gcode is forced to always be the first g-code that is done. Without this it can happen that other g-code, such as a T0 can be inserted before the start g-code"

resources/definitions/fdmextruder.def.json
```diff
@@ -120,6 +131,18 @@
                     "settable_per_meshgroup": false,
                     "type": "str"
                 },
+                "machine_extruder_change_duration":
+                {
+                    "default_value": 0,
+                    "description": "The time it'll take for the extruder change to execute.",
```
nallath, line 137: Same as the other one. The description should be a bit longer to explain why and when one would use this.

nallath (changes requested):

resources/definitions/fdmprinter.def.json
TheSin-: Thank you for the example I believe I have this fixed now

TheSin-:
Added a small fix to apply variable expansion to the extruder prestart code

TheSin-:
Just having a hard time applying defaults to my printer definition

```
    "overrides":
    {
        "machine_start_gcode_first": { "value": true },
        "machine_extruder_prestart_code": { "default_value": "M104 S{material_print_temperature} T{extruder_nr}" }
    }
```

in the `base.def` file isn't working, neither setting gets applied on create new, did I miss an other spot to be able to apply these by default on new printer creation?

TheSin-:
Updated for the latest code and tested, still I can not set these values when adding a new printer, but the options themselves to work.  I can not figure out why they won't preset with a new printer though.

TheSin-:
Update, the first extruder gets the proper defaults values, all extruders after that do need get the defaults from base, not able to figure out why it's not propagating to each extruder :\

Also the default value for start_gcode_first isn't auto checked on adding a new printer.  Am I suppose to version up the definitions so they auto apply?

but everything else is tested and working now.

rburema:
For future reference -- we've had problems with the size of these fields on the minimum required screen resolution, so we could fix that by making them a little higher -- I fear these changes will make that problem appear again for the leftmost two fields.

TheSin-:
oh what is the min I'll resize to that in the future, I didn't find any documentation on that so I just used the default size it shows up in.

rburema:
Sorry for the late reply -- I'm not sure if it's in the _developer_ docs indeed. We'd previously decided that the min resolution for a screen is [1024x768](https://support.makerbot.com/s/article/1667410778209), but I'm not too sure we're adhering too closely to that anymore.

In any case, that cascades down until you have the size of the popup at that resolution, the min size of which was determined sort-of experimentally. -- In other words, we just tweaked it so it works on smaller screens as well (though maybe not actually down to that minimum stated). -- Sorry for the vagueness, I wasn't always aware of the details.

In any case, it's merged now, and not that big of a deal; let's see if there are any complaints, and if so, we'll pick it up from there.

TheSin-:
sounds good I'll make sure I test at 1024x768 moving forward though, for myself do you have a screen shot of the issue?  I was sure I tested lower res than that.

Eitherway thank you for the assist I'll be more cautious in the future, and since this is finally in I'll start on phase 2 of this mod this weekend.

rburema:
@TheSin- The issue is basically this: https://github.com/Ultimaker/Cura/issues/7224

PR #19355, opened 2024-07-10
CURA-12005 Include slicemetadata in makerbot file

wawanbreton:
Include the very same file as generted into UFP files. The existing export method has been moved from UFPWriter top `API/Settings` in order to make it easily callable by the two plugins.

:warning: In the makerbot archive, the file is placed at the root with the other files, because there is no folder structure. In the UFP this file is placed in the `Cura` subfolder.


rburema (approved):
Good work -- `-w --color-moved` made it really clear that the change is basically just a move 😁

PR #19392, opened 2024-07-17
Add "instance0" entry to gaggles

casperlamboo:
With the commits https://github.com/Ultimaker/Cura/commit/fdfc5dcc67f8efac60404421e86b357b5ef3a5af and https://github.com/Ultimaker/Cura/commit/2f8fc2cabef19b23f2d024105b21ad7ccd5b09a4 the instance list was changed from containing all model instances to a single instance called "instance0". Digital factory expects the same instances to be present in the gaggles object.


sergii-shevchenko-ultimaker (approved):

wawanbreton (approved):
Code is fine, I trust you that it does good things :smiley:

PR #19393, opened 2024-07-17
[CURA-11958] fix auto-select `.makerbot` format

rburema:
Fix a bug where the auto-selection of the right format didn't happen, because the old (hacky) condition on which this was decided is no longer valid.

### DIFF
--- a/cura/Settings/GlobalStack.py
+++ b/cura/Settings/GlobalStack.py
@@ -83,6 +83,15 @@ class GlobalStack(CuraContainerStack):
         """
         return self.getMetaDataEntry("supports_material_export", False)
 
+    @pyqtProperty("QVariantList", constant = True)
+    def getOutputFileFormats(self) -> List[str]:
+        """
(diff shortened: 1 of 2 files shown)


cura/Settings/GlobalStack.py
```diff
@@ -83,6 +83,15 @@ def supportsMaterialExport(self):
         """
         return self.getMetaDataEntry("supports_material_export", False)
 
+    @pyqtProperty("QVariantList", constant = True)
```
wawanbreton, line 86: Isn't there a specific tag for methods ? This is not really a property

wawanbreton (approved):
One minor remarks, looks good

cura/Settings/GlobalStack.py
rburema: I don't think you can set `constant` on a method. This optimizes it a bit, and I also think that since it is constant, that from the perspective of QML, you _can_ actually see it as a property.

PR #19396, opened 2024-07-18
[CURA-11940] XmlMaterial: Don't 'translate' to yes when it's a cura-setting.

rburema:


plugins/XmlMaterialProfile/XmlMaterialProfile.py
```diff
@@ -1083,10 +1083,9 @@ def _addSettingElement(self, builder, instance):
             # Skip material properties (eg diameter) or metadata (eg GUID)
             return
 
-        if instance.value is True:
-            data = "yes"
-        elif instance.value is False:
-            data = "no"
+        truth_map = { True: "yes", False: "no" }
```
wawanbreton, line 1086: I'm a big fan of this variable name :rofl:

wawanbreton (approved):

plugins/XmlMaterialProfile/XmlMaterialProfile.py
rburema: I wasn't sure what else to call it 😅

PR #19419, opened 2024-07-22
Add some options to CreateThumbnail to allow for some customisation o…

geekykayaker-anon:
# Description

Currently the CreateThumbnail plugin writes out the form:
`thumbnail begin <x>x<y>`
My printer (Creality V3 KE) has a display which shows thumbnails, unfortunately it expects the headers to be formed:
`png begin <x>*<y>`

In addition to this, Moonraker ONLY seems to accept the first format, and not the second.

In order for thumbnails to appear both on my printer screen AND in Mainsail (which used Moonraker) I have to output each resolution twice, once for each format.

This MR adds support for toggling between:

-  'thumbnail' and 'png
- '*' and 'x'

## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Test 1
I created a project with the old plugin, and set some settings.
My changes didn't impact those settings, and produced the same GCODE output when re-generated

Test 2:
I set inverted the first flag, and then got the expected behaviour in the GCODE.

Test 3:
I set inverted the second flag, and then got the expected behaviour in the GCODE.

Test 4:
With the each of the four combinations of the two new flags as steps, I can get all four combinations of output shown in a CGODE file.

**Test Configuration**:
MacOs Sequoia
Cura5

# Checklist:


dvdmrtnz:
Current version of CreateThumbnail in Cura 5.10.0 actually seems to be working just fine with my Ender 3 V3 KE firmware version V1.1.0.15. I have simply added two thumbnails: One 96x96px thumbnail and another 300x300px and it's displayed just fine on the web interface and on the printer's display.







Gcode is using this format:

```
; thumbnail begin 300x300 38196

; thumbnail begin 96x96 7944
```

Here is full gcode:

[CE3V3KE_xyzCalibration_cube.zip](https://github.com/user-attachments/files/20753724/CE3V3KE_xyzCalibration_cube.zip)

Issue #19388, opened 2024-07-16
Smarter printing above sparse infill


Hello1024:
### Is your feature request related to a problem?

When printing above sparse infill (ie. 5%), there are quite a lot of shapes that fail to print well.

Specifically, the 'skin edge support layers' feature often draws shapes into the infill which are fully unsupported (A), because they're lines that by unlucky chance miss all the infill, or draws paths where part of the path clearly is unsupported (B).

![image](https://github.com/user-attachments/assets/d263431c-5ba3-472c-a729-817b3ed86a74)

The end result is that frequently these bits of unsupported printing stick to the print head and make blobs of plastic elsewhere, or even catch on bits of the extruder/fan and knock the whole print off the bed.

### Describe the solution you'd like

These paths need to be made printable in some way.     Consider a line to be printable if both endpoints of a line are within `line_width` of a line on the layer below. 

If a line is unprintable, it should be made to be printable by adding extra support lines on the layer below.  These could form a little like lightning infill - starting with a few basic straight lines a few layers down, to make the area that needs to be supported fully supported where it needs to be.

### Describe alternatives you've considered

n/a

### Affected users and/or printers

all users who print sparsely

GregValiant:
I've had this problem myself so these are just some thoughts.  Without increasing the support density (dramatically in some cases) I don't know that this can be fixed.
![infill fall through](https://github.com/user-attachments/assets/3d29e878-c427-4e9e-bbef-d4fb4c690d13)
The features in the black marks are going to just fall through the grid.  Those might be able to be handled with some sort of fix.
The features in the red shape will fail because you can't print a curve over air, and you can't change directions of straight extrusions over air.  That's a slightly different problem.
I usually address this problem by adding another "floor" (so the plastic has more chance to build up) or going with "Gradual Infill Steps".  Even then some of that is going to be over air.
Without the full structure building up from the bottom, something is going to print over air.  Even the "Stepped Infill" has to print over air each time the density increases.

smartavionics:
How about a setting that would force an infill layer to be printed as, say, 50% density zigzag pattern when the first layer of infill support is detected? Alternatively, you could have a setting that specifies that a particular layer (or layers) were infilled using a 50% zigzag?

GregValiant:
This is with @smartavionics idea but using an "Infill Mesh" configured for zig-zag at 50%.  This looks like it would work for the problem I was having.
[...]

PR #19446, opened 2024-07-29
Add setting for extra_infill_lines_to_support_skins

Hello1024:
# Description

This is the config necessary for the fix of  #19388

## Type of change

- [X] Printer definition file(s)

# How Has This Been Tested?

Tested with https://github.com/Ultimaker/CuraEngine/pull/2122

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2326,7 +2326,7 @@
                     "label": "Skin Edge Support Thickness",
                     "description": "The thickness of the extra infill that supports skin edges.",
                     "unit": "mm",
-                    "default_value": 0.8,
+                    "default_value": 0,
                     "minimum_value": "0",
                     "maximum_value": "machine_height",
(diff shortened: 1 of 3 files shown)


rburema (changes requested):
Hi!

Again, thanks for the extensive contribution!

(Please also see my review of the engine part.)

Here, I'd like to disable it entirely for lightning infill. In my tests, it didn't change anything, and this feature doesn't make a lot of sense in that case, since that type of infill is already meant to address these issues on its own.

(I'm less inclined to want to change it for gradual infills steps, as it works, and makes a little more sense.)

For the rest, it seems to work quite nicely! I _did_ run into something I think is at the very least a bit counterintuitive if I multiply the # of infill lines (see the pictures). I don't think this can do much harm, but OTOH, I don't think those lines are needed there.

three_infill_mult_effct

(how it is with this feature off, for comparison):

three_infill_mult_no_effct

)

Hello1024 (commented):
Done.    Please take another look

rburema (approved):

PR #19450, opened 2024-07-30
[CURA-12050] Start using the 'force_depends_on_settings' property.

rburema:
This is useful if the relations between settings are cross-extruder, which can happen when a setting that is currently limited to one extruder, depends on another setting (in a value-formula of the like) which is currently limited to _another_ extruder.

Note that I renamed the 'depends_on_settings' to 'force_depends_on_settings' as the former may imply that you should do it for _all_ dependant settings, instead of it being the exception (as we expect that this will occur rarely -- for example in multi-material rafts).

See backend PR: https://github.com/Ultimaker/Uranium/pull/959

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -414,6 +414,9 @@ class CuraApplication(QtApplication):
         SettingDefinition.addSupportedProperty("settable_globally", DefinitionPropertyType.Any, default=True,
                                                read_only=True)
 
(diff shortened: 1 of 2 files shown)


nallath (changes requested):
Force_depends_on_settings should not be used in fdmprinter.def.json as we don't need it for all printers just for method

nallath:
Closing this one as these changes are not needed

PR #19451, opened 2024-07-30
CURA-12050 add force depends on setting to method

nallath:


rburema (approved):
I see some things that could be left out (since they would live in the same container-stack), but better to be safe, and they can do little harm.

PR #19477, opened 2024-08-04
Add printer definitions for Sovol SV08.

sesse:
The SV08 (or SV-08; nomenclature is not completely consistent) is a relatively new printed based on Voron 2.4, running Klipper. This adds printer, extruder and material definitions for it, based on the voron2_base definitions plus Sovol's published profiles for Orca Slicer:

  https://drive.google.com/drive/folders/1KWjLxwpO_9_Xqi_f6qu84HRxZi26a_GN

Unfortunately, the included STL model for the platform does not have texture coordinates, so we cannot use the platform texture (unless someone goes to add them manually or otherwise adjusts the model).

The following settings were not carried over, mostly because I could not find any obvious equivalent in Cura:

  - Machine: "retract_before_wipe": [ "0%" ], "machine_max_acceleration_extruding": [ "20000" ], "machine_max_acceleration_retracting": [ "5000" ], "retract_length_toolchange": [ "2" ], "wipe_distance": [ "2" ], "retract_lift_below": [ "343" ], "thumbnails_format": "PNG", "before_layer_change_gcode": "TIMELAPSE_TAKE_FRAME\nG92 E0",

  - Filament (using ABS as an example; the exact values differ between the four material profiles): "nozzle_temperature_range_low": [ "190" ], "nozzle_temperature_range_high": [ "250" ], "full_fan_speed_layer": [ "0" ], "slow_down_layer_time": [ "4" ], "slow_down_min_speed": [ "10" ], "overhang_fan_speed": [ "30" ], "overhang_fan_threshold": [ "25%" ], "temperature_vitrification": [ "60" ], "activate_air_filtration": [ "1" ], "close_fan_the_first_x_layers": [ "3" ], "complete_print_exhaust_fan_speed": [ "60" ], "during_print_exhaust_fan_speed": [ "100" ],

  - Process: A bunch (e.g. bridge_flow, elefant_foot_compensation, overhang_1_4_speed, etc. etc.), but it's unclear how many are printer-specific and how many are just Orca defaults where Cura wants to do things differently.

The start and end G-code are copied over verbatim, except that it leaves the printer in relative coordinate mode and Cura does not set this explicitly back to absolute, so we need an explicit G90 at the end. (Also, there seems to be a Klipper issue where G90 does not reset extrusion to absolute as well, so we need to send an explicit M82.)

Unfortunately, Cura chooses SS_ as prefix instead of SV08_. I don't know if there is a way to override this; the other Sovol printers seem to have the same issue.

I've tested this with the standard 0.4mm nozzle and ABS/PLA, using the Moonraker plugin. PETG and TPU are untested, in part because the current nozzle is said to be unsafe for PETG. The time estimates from Cura are not all that good, but klipper_estimator helps.

Future work would include supporting the 0.2mm, 0.6mm and 0.8mm nozzles. There are separate profiles for them, with different layer height, support settings, print speeds, etc. -- and then there is a specific PLA/0.2mm profile with lower printing speed and higher fan settings.

# Description

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:


sesse:
Update in v3: Rebased, some more fan settings have been ported over, filaments have been put in the correct directories,and the PRINT_START macro is given initial bed and nozzle temperatures for the benefit of better mainline Klipper integration.

resources/definitions/sovol_sv08.def.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": 2,
+    "name": "Sovol SV08",
+    "inherits": "voron2_base",
```
HellAholic:
```suggestion
    "inherits": "voron2_base",
```
This type of inheritance (cross-manufacturer) is not recommended. This would mean that if anyone makes modification to voron2 printer series, this printer will also get affected without being in the scope of the developer.

sesse: Hm, but it _is_ a Voron2. Or at least a Voron2 derivative, so if someone made a change to voron2_base, I would assume that we _want_ the change to propagate to SV08. Would it be better to copy the entire voron2_base somewhere?

HellAholic: I'm not familiar with the Sovol machines to judge the effects accurately, but when a change is done to a machine series / base definition, the quality profiles and variants are also adjusted to accommodate the changes. Otherwise your print settings will have two sources that don't align.
There is a PR open for changing the voron2_base https://github.com/Ultimaker/Cura/pull/20023
You can see the changes also propagate into the quality files.

You can create a separate base for this machine specifically if it's the only one that is a voron2 derivative in the Sovol series, alternatively, you can just inherit from fdm printer and apply the same overrides as voron2 base to the sovol definition.

The main discussion here is that the changes to voron2_base will not be seen as changes that affect Sovol SV08 and therefore the print results might degrade.

sesse: SV08 is a Voron 2.4, Sovol Zero is some other Voron derivative (I don't know much about it), the rest are different designs.

I can move the voron2_base overrides into this printer and inherit directly from fdm, I think maybe that's the safest.

HellAholic: That seems to be the safest choice given the context of the printer, that way you're sure the profiles and test prints remain valid.

sesse: I hope to get this done this weekend; there are some completely unrelated projects that take priority. I guess I also need a rebase at some point.

sesse: I've made the changes locally, and tried to rebase, but evidently, whatever build instructions don't work anymore; following the instructions on https://github.com/Ultimaker/Cura/wiki/Running-Cura-from-Source gives me simply:

```
kos:~/nmu/Cura> source venv/bin/activate
kos:~/nmu/Cura> conan install . --build=missing --update -g VirtualPythonEnv

  File "/home/sesse/.local/bin/conan", line 5, in <module>
    from conans.conan import run
ModuleNotFoundError: No module named 'conans'
```

IIRC building Cura was a huge pain last time (and I say this as someone who compiles C++ software pretty much all day :-) ); can I just send the rebased PR and hope it works?

HellAholic: You don't really need to compile or run from source for checking the definitions.
Assuming you're on linux/mac, just copy the files into `~/.config/cura/<version>` there are folders for definition, extruders, variants, ... and it adds the printer when you launch that version of cura into the list, so you can check things out using an appImage or a package.

Tests/Linters also will catch if any of the overrides are not correct on the repository.

sesse: Evidently using the precompiled versions now hits https://github.com/Ultimaker/Cura/issues/19837 for me, so that's a no-go, too.

I uploaded the patch, so we'll see what the linters do.

HellAholic: I think you forgot to push the changes :P

sesse: Gah, I pushed but to wrong branch. Better now?

HellAholic:
One "issue" remaining is the override for the default_value of the retraction_min_travel which is set to 1, the thing is that the formula for value in fdmprinter will override this default_value, so if it is important for it to have the value of 1, then overriding value instead would be best. On the other hand, you can just remove the override and let Cura handle it through the fmdprinter definition.

sesse:
> One "issue" remaining is the override for the default_value of the retraction_min_travel which is set to 1, the thing is that the formula for value in fdmprinter will override this default_value, so if it is important for it to have the value of 1, then overriding value instead would be best. On the other hand, you can just remove the override and let Cura handle it through the fmdprinter definition.

I've never thought much about it, or experimented with something else (I'm fairly sure it's just copied from the Orca profiles). This setting is for stringing, right? I can't recall having much stringing problems at all on the SV08.

I can remove it if you wish, or just remove it on your side. Thank you so much for shepherding this through, I haven't had the chance to be as proactive on this PR as I'd hoped.

HellAholic:
Made the commit removing the override.
I'll check things out later to see if everything works in a build.
If you could, test things out during the public beta period once the printer is merged to see if it prints as expected and quality is good.

sesse:
> If you could, test things out during the public beta period once the printer is merged to see if it prints as expected and quality is good.

Yes, I'll be happy to, assuming I can get the prebuilts to run again. :-) (I think maybe there are some workarounds with downgrading zlib that are available.)

HellAholic (approved):

sesse:
Thank you! I'll try to keep track of when the betas are out.

sesse:
I eventually managed to compile latest git, and it seems to work fine. I only made a couple of simple test prints because that's what I happened to need right now, but there were no issues and I’ve done more complicated prints with this definition before.

The printer really wants klipper_preprocessor as other Klipper printers do (to get adaptive bed meshing, better ETAs, etc.), but this test was with completely plain Cura.

PR #19492, opened 2024-08-06
[CURA-11957] Makerbot file formats don't just have 1 mime-type.

rburema:
There are machines with other mime-types. Prepare for that one.

(otherwise solved by previous efforts)

### DIFF
--- a/plugins/DigitalLibrary/resources/qml/SaveProjectFilesPage.qml
+++ b/plugins/DigitalLibrary/resources/qml/SaveProjectFilesPage.qml
@@ -208,7 +208,7 @@ Item
         anchors.rightMargin: UM.Theme.getSize("thin_margin").height
 
         enabled: UM.Backend.state == UM.Backend.Done
-        currentIndex: UM.Backend.state == UM.Backend.Done ? (Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot") ? 1 : 0) : 2
+        currentIndex: UM.Backend.state == UM.Backend.Done ? ((Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot") || Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot-sketch")) ? 1 : 0) : 2
 
         textRole: "text"
         valueRole: "value"


rburema:
❌'s are probably because this is based on a main where that also happens 😅

fieldOfView:
This code would benefit from not being two ternaries, and from some documentation about what is going on here.

rburema:
@fieldOfView Fair. I'll have look to see if I can make it less confusing (and more documented).

rburema:
So I kept the double ternary, but given how short that line now is, I don't think that's a problem.

plugins/DigitalLibrary/resources/qml/SaveProjectFilesPage.qml
```diff
@@ -208,7 +208,14 @@ Item
         anchors.rightMargin: UM.Theme.getSize("thin_margin").height
 
         enabled: UM.Backend.state == UM.Backend.Done
-        currentIndex: UM.Backend.state == UM.Backend.Done ? (Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot") ? 1 : 0) : 2
+
+        // Pre-select the correct index, depending on the situation (see the model-property below):
+        // - Don't select any post-slice-file-format when the engine isn't done.
+        // - Choose either the S-series or the Makerbot-series of printers' format otherwise, depending on the active printer.
+        // This way, the user can just click 'save' without having to worry about wether or not the format is right.
+        property int isMakerbotFormat: Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot") || Cura.MachineManager.activeMachine.getOutputFileFormats.includes("application/x-makerbot-sketch")
```
wawanbreton, line 216: Is there a place we could get the makerbot mime types from ? If we add a third one in some months, it is very likely that we will do the very same mistake...

wawanbreton (commented):
Minor remark, but code is much more readable :smiley:

plugins/DigitalLibrary/resources/qml/SaveProjectFilesPage.qml
rburema: I suppose we could make it into a method somewhere and mark that as a pyqtproperty.

I'm not sure how likely it is that we even continue with any sort of `.makerbot` filetype beyond what the Method and Sketch lines have now.

wawanbreton: Indeed, we can hope very hard that we don't have a 3rd Makerbot file format :sweat_smile: let's don't change this then.

wawanbreton (approved):

PR #19499, opened 2024-08-07
propose spiralize version of Eazao profiles

rburema:
So there are a few limitations when you go the 'spiralize' route, such as that it can only be single-walled (other than on the bottom).

Also, this mode assumes that the input is a 'solid' model, that it then makes into a vase. If the model going into it is _already_ a vase, it may give strange output.

For vase-mode, you can also set it to gradually increase the height, instead of going up only at the end (/beginning) of each layer -- so I've done that.

Lastly, I've taken the liberty to set the fill for top/bottom to 'concentric' instead of lines.


Eazao:
Hi [rburema](https://github.com/rburema),

Thank you very much for your help. We fully agree with your submission. You can merge your proposal into cura. We hope to see updates for Eazao devices in the latest cura.

Thank you again for your help.

PR #19503, opened 2024-08-07
[CURA-11956] The easy way to fix a slow response is to up the timeout...

rburema:
This might on one hand have (other) consequences, and given the slowness we encounter not even be enough in some cases (may be related to the size of the messages?) May need to be fixed on server side.

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
@@ -42,7 +42,7 @@ class CloudApiClient:
     CLUSTER_API_ROOT = f"{ROOT_PATH}/connect/v1"
     CURA_API_ROOT = f"{ROOT_PATH}/cura/v1"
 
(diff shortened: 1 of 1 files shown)


wawanbreton:
Should this PR be undrafted ?

Could we, instead of using a timeout, monitor the sent bytes, and cancel the request if they don't grow up ? This way we could set a longer timeout, which would be the ultimate security only.

rburema:
> Should this PR be undrafted ?

It felt like a bit of a hack, so I made it a draft.

> Could we, instead of using a timeout, monitor the sent bytes, and cancel the request if they don't grow up ? This way we could set a longer timeout, which would be the ultimate security only.

Maybe? I haven't tried it, so I'm not sure if _anything_ actually comes in before the deadline.

wawanbreton:
> Maybe? I haven't tried it, so I'm not sure if _anything_ actually comes in before the deadline.

This probably uses a `QNetworkReply`, which has an `uploadProgress` signal. So technically it should be doable, but it is a much larger change :smiley:

wawanbreton (approved):

PR #19539, opened 2024-08-15
CURA-11501 Fix excluded materials being sometimes visible

wawanbreton:
Apparently the `exclude_materials` property of `MachineNode` has changed over time, and one of its use has not been updated, leading to some cases where materials could be non excluded.

### DIFF
--- a/cura/Machines/MachineNode.py
+++ b/cura/Machines/MachineNode.py
@@ -14,7 +14,6 @@ from cura.Machines.QualityChangesGroup import QualityChangesGroup  # To construc
 from cura.Machines.QualityGroup import QualityGroup  # To construct groups of quality profiles that belong together.
 from cura.Machines.QualityNode import QualityNode
 from cura.Machines.VariantNode import VariantNode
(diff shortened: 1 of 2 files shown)


cura/Machines/MachineNode.py
```diff
@@ -168,13 +168,17 @@ def preferredGlobalQuality(self) -> "QualityNode":
...
-    def isExcludedMaterial(self, material: MaterialNode) -> bool:
+    def isExcludedMaterialBaseFile(self, material_base_file: str) -> bool:
         """Returns whether the material should be excluded from the list of materials."""
         for exclude_material in self.exclude_materials:
-            if exclude_material in material["id"]:
+            if exclude_material in material_base_file:
                 return True
         return False
 
+    def isExcludedMaterial(self, material: MaterialNode) -> bool:
```
wawanbreton, line 178: Well, now we can set this function deprecated, as we are no more using it, right ?

rburema: Done! :-)

PR #19547, opened 2024-08-16
 CURA-12096 move gradual flow plugin back into the engine core

wawanbreton:
Include the extra settings defined by the gradual flow plugin, in the main core.
Also remove all calls to the plugin as a dependancy, as it now part of the engine.

CURA-12096

Comes with https://github.com/Ultimaker/CuraEngine/pull/2133


casperlamboo (commented):
I think we also need an upgrade script converting all `_plugin__curaenginegradualflow__0_1_0__xxx` settings to regular settings.

wawanbreton:
> I think we also need an upgrade script converting all `_plugin__curaenginegradualflow__0_1_0__xxx` settings to regular settings.

Yes we do, I added it to the list of "tickets that require an upgrade script for 5.9"

casperlamboo (approved):

PR #19555, opened 2024-08-19
add settable_globally property

HellAholic:
CURA-12024
Clean up of the definition.
Definition not in use in the engine so there are no effects at the moment to test and validate.
Using the godmode plugin or exporting the slice setting sent to engine using custom scripts now show the values are updating correctly based on the extruder values.
The global setting is set to extruder.-1


wawanbreton (approved):
I don't understand all the implications of this change, but it seems to make sense because some similar settings have this.

Issue #18530, opened 2024-03-06
Slicing - Unexpected error


gregreenwood:
### Cura Version

5.6.0 & 5.7.0-alpha.0

### Operating System

Linux Mint 21 Cinnamon

### Printer

Longer LK5 Pro

### Describe model location

No

### Describe your model

Not in Cura.  It slices OK in Prusa Version 2.7.1+linux-x64-GTK3

### Add your .zip here ⬇️

[SDI_Power_Shell-Body.3mf.zip](https://github.com/Ultimaker/Cura/files/14516919/SDI_Power_Shell-Body.3mf.zip)

GregValiant:
Thanks for the report.
On Windows 10 it was one of the fastest slices I've seen in a long time.  The model is free from errors.

I'll mark this as a duplicate and the Cura team will take a look.
![image](https://github.com/Ultimaker/Cura/assets/64202104/cab749f4-8dc1-450d-bcf1-9f15acad4abc)

gregreenwood:
I was using LMint V21 Cinnamon and the applications were appimages.

gregreenwood:
I fixed the issue by deleting the /home/[my user
name]/.local/share/cura/5.6 folder and redownloaded the appimage for 5.6
and ran it.  Now it slices.

Regards
GG

gregreenwood:
Correction.  The bug returned when I loaded CuraEngine Tiled Infill
Generation 0.2.0 from the Marketplace!
I started with Honeycomb and it failed to slice.  Choosing anything after
that, it fails no matter what I choose.
Uninstalling the Tiled Generation 0.2.0 fixes the failure to slice.

Hope find the bug.
Regards
GG

GregValiant:
That will help.  Thanks for adding it.

Issue #17739, opened 2023-12-22
Cura fails with every model


hd42:
### Cura Version

5.6.0

### Operating System

Ubuntu Linux

### Printer

Elegoo Neptune 4 plus

### Describe model location

default

### Describe your model

simple pyramid

### Add your .zip here ⬇️

[EN3P_Pyramide.zip](https://github.com/Ultimaker/Cura/files/13754513/EN3P_Pyramide.zip)

5axes:
Sory no issue for me : 
![image](https://github.com/Ultimaker/Cura/assets/11015345/e12bc55c-4a00-466d-8bd7-d9ed80d71dd4)

but in your project the printer is detected as **Anycubic Kobra Youtube** ????
![image](https://github.com/Ultimaker/Cura/assets/11015345/d172dcbe-3b69-4b32-92b8-810bbf85ab20)

hd42:
[CFFFP_TOP2.zip](https://github.com/Ultimaker/Cura/files/13829282/CFFFP_TOP2.zip)
I tried creating a new printer profile, but still get the same error.

5axes:
Still no problem for me :  
![image](https://github.com/Ultimaker/Cura/assets/11015345/d6196216-d7fc-46d6-ba43-9e4e86bb0d5c)

hd42:
I disabled all plugins and now it works. I'll try to enable them one after the other to find out which one caused the problem.

5axes:
similar to https://github.com/Ultimaker/Cura/issues/17551 ?

hd42:
The culprit appears to be the CuraEngineTiledInfill 0.2.0 plugin.

hd42:
Thank you for your help, @5axes.

Issue #17551, opened 2023-12-07
Cura 5.6.0 - Slicing fails with CuraEngine Tiled Infill Generation 0.2.0 Plugin Enabled


rsimpson1297:
### Cura Version

5.6.0

### Operating System

Linux (UltiMaker-Cura-5.6.0-linux-X64_cbc4dfea87e29f7635e925d1ba0769c0.AppImage)

### Printer

Creality Ender 3 v3 SE

### Reproduction steps

Downloaded a fresh AppImage, launched and configured for my printer, performed a slice on a known good STL. Slice completed normally.
Installed the CuraEngine Tiled Infill Generation 0.2.0 Plugin, restarted the application. 
Opened the same STL that sliced correctly, performed a slice, slice fails.

In troubleshooting, I disabled the plugin, restarted Cura, performed the same test and the slice was successful.
![Screenshot at 2023-12-07 12-14-46](https://github.com/Ultimaker/Cura/assets/145204621/c200555e-d4d7-4df0-a1e2-cbd23e7ca1fe)

### Actual results

Error is generated

### Expected results

successful slice

### Add your .zip and screenshots here ⬇️

[cura.log](https://github.com/Ultimaker/Cura/files/13603098/cura.log)
[CuraEngineTiledInfill.log](https://github.com/Ultimaker/Cura/files/13603099/CuraEngineTiledInfill.log)

Issue #17441, opened 2023-11-27
Slicing failed with an unexpected error


edscil:
### Cura Version

5.5.0

### Operating System

Linux Mint

### Printer

Vertex delta K8800

### Name abnormal settings

no changes

### Describe model location

centered

### Describe your model

I tried to slice several different models and each one had the same error.

### Add your .zip here ⬇️

I checked the cura.log and found the following error warning repeated many times:
- WARNING - [MainThread] UM.Settings.SettingDefinition._deserialize_dict [668]: Unrecognised property comment in setting _plugin__curaenginegradualflow__0_1_0__max_flow_acceleration
I had installed the extension 'CuraEngine Tiled Infill Generation' about a week before the problems started, but I had sliced a number of models between the time of installation and when the problem began.  I uninstalled the extension and the problem ended.  I am now able to slice models without difficulty.  Just thought you might like to know.

Nuc1eoN:
Same waring here. Occasionally I do get a slicing error.
I do not have the "CuraEngine Tiled Infill Generation" extension.

johnneerdael:
I see the same error on 5.6 and 5.7

PR #19599, opened 2024-08-30
CURA-12091 fix infill plugin start crash

wawanbreton:
When starting an engine plugin executable from within an AppImage, the AppImage hooks make some black magic to make sure the called executable is started from the OS environment. In our case this is not what we want, as the plugin is supposed to use the environment provided with the AppImage, which is compatible with more platforms.

This PR adds an environment variable before starting the engine plugin executable, to tell AppImage to consider the executable as if it was part of its environment.

This can also be achieved by starting cura with e.g. `APPDIR_MODULE_DIR=$HOME/.local/share/cura/5.8/plugins/CuraEngineTiledInfill/CuraEngineTiledInfill/x86_64/Linux ./UltiMaker-Cura-5.8.0-linux-X64.AppImage`. Thus, one can use the Infil plugin version 2.0.1 with Cura 5.8.0 on platforms where it would not run (e.g. Mint 20). In 5.9.0 it will work out of the box.

Requires https://github.com/Ultimaker/CuraEngine_plugin_infill_generate/pull/19 (Or Infill plugin package 2.0.1)

May fix https://github.com/Ultimaker/Cura/issues/17441
Fixes https://github.com/Ultimaker/Cura/issues/17551
Fixes https://github.com/Ultimaker/Cura/issues/17739
Fixes https://github.com/Ultimaker/Cura/issues/18530

CURA-12091

### DIFF
--- a/cura/BackendPlugin.py
+++ b/cura/BackendPlugin.py
@@ -3,6 +3,7 @@
 import socket
 import os
 import subprocess
+import pathlib
 from typing import Optional, List
 
 from UM.Logger import Logger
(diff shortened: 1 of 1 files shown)


rburema (approved):
lgtm

PR #19644, opened 2024-09-12
Cura 12138 export for support option

HellAholic:
CURA-12138
Requires https://github.com/Ultimaker/Uranium/pull/970

- [X] Add option under Help: "Export project for support"
- [X] Make it easy to find: Place it either at the top with a line separator under it .. or.. under sponsor cura and put the line separator above and bellow it 
- [X] Add a naming scheme that prevents users and support personnel from overwriting projects by accident
"exported-(package/project)-for-support-timestamp/random-number).3mf"
ex: exported-project-for-support-20240911-13-13.3mf
- [X] Should not impact the normal project saving flow or naming
- [X] Include log file in archive


plugins/3MFWriter/ThreeMFWorkspaceWriter.py
```diff
@@ -33,7 +35,7 @@ def setExportModel(self, model: SettingsExportModel) -> None:
         if self._ucp_model != model:
             self._ucp_model = model
 
-    def _write(self, stream, nodes, mode=WorkspaceWriter.OutputMode.BinaryMode):
+    def _write(self, stream, nodes, mode, include_log):
```
nallath, line 38: Include log is a bool trap. I'd also argue that instead of removing the default for the mode, we should just set the include_log to be False by default

plugins/3MFWriter/ThreeMFWorkspaceWriter.py
```diff
@@ -191,6 +198,17 @@ def _writeContainerToArchive(container, archive):
             Logger.error("File became inaccessible while writing to it: {archive_filename}".format(archive_filename = archive.fp.name))
             return
 
+    @staticmethod
+    def _writeLogFile(archive):
```
nallath, line 202: Missing typing.

nallath (changes requested):

plugins/3MFWriter/ThreeMFWorkspaceWriter.py
wawanbreton: That's right, I have removed the `mode` default value not because of the `include_log`, but because it is a "private" method and only called at one place, with a value for the `mode`. Restoring the default value and setting `include_log` to `False` is also fine if you prefer.

PR #19666, opened 2024-09-20
Pp 528 unused extruder settings

Frederic98:
# Description
Ever since the dawn of time Cura, having a material in Extruder 2 active influences the settings in Extruder 1. This is even when you only use the material in Extruder 1.

For example, PLA needs a buildplate temperature of 60°C on S-line, while ABS needs 80°C. Because there is only 1 bed to set the temperature of, the highest of the two is used to make sure the object sticks well to the bed. This, however, results in the PLA being quite droopy because it stays quite soft at that temperature - a compromise needed when printing the two materials together.

If, however, you only use the PLA in Extruder 1, you have to remember to manually disable Extruder 2, as otherwise the PLA uses a bed temperature of 80°C for no good reason.

This problem also applies to various other settings that cannot be set per extruder, like the Build Volume Temperature, Adhesion Type, Materials Shrinkage Compensation, Support Structure, …

# Solution
Turns out, Cura already has a function `ExtruderManager.getUsedExtruderStacks()`, which goes over a bunch of settings to check which extruders are _actually_ used, not just the enabled ones. For example, it checks:
- `support_..._extruder_nr` if support is enabled
- `skirt_brim_extruder_nr` if `adhesion_type` is `skirt` or `brim`
- `raft_..._extruder_nr` if `adhesion_type` is `raft`

The result of this function is now saved in new setting (`extruder_used`). This setting lives in the machine section, so is not settable by the user and is driven by Cura.  
The function `extruderValues` that is available in `SettingFunction`s then only gives the values of the actually used extruders. If there are no used extruders (yet), like when there is no mesh loaded, it falls back to the old behaviour of using all the enabled extruders.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2494,7 +2494,7 @@
                     "enabled": "machine_heated_build_volume",
                     "settable_per_mesh": false,
                     "settable_per_extruder": false,
-                    "depends_on_settings": ["extruder_used"]
+                    "depends_on_settings": [ "extruder_used" ]
                 },
                 "material_print_temperature":
                 {
@@ -2608,7 +2608,7 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false,
-                    "depends_on_settings": ["extruder_used"]
+                    "depends_on_settings": [ "extruder_used" ]
                 },
                 "material_bed_temperature_layer_0":
                 {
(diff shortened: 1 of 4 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -2587,7 +2610,7 @@
                     "type": "float",
                     "default_value": 60,
                     "value": "default_material_bed_temperature",
-                    "resolve": "max(extruderValues('material_bed_temperature'))",
+                    "resolve": "max(extruderValues('material_bed_temperature', where='extruder_used'))",
```
casperlamboo, line 2613: Keyword arguments are not supported in Curator. We will need to extend the parser to support this.

PR #19691, opened 2024-09-26
Method(s) updates for 5.9

alanbjorklund:
Method printers: Adding new materials (Tough PLA, PETG, NylonCF, ABS) and updated profiles for 5.9
PP-511

Added lots of updates for method printers including:

-ABS 175
-Nylon CF 175
-PETG 175
-Tough PLA 175
-preliminary settings for Labs materials (stored in generic files for now)
-updates to default infill strategy (matching CloudPrint)

updates to roof layer order and thickness
standardized soluble support values
standardized soluble raft values
added many max value / min value warnings to method def json (less yellow values in UI)
updated Method(s) build volume to match CloudPrint
added logic to extend X volume when the right extruder is disabled
adjust first layer settings when raft is not selected
enable prime tower when 2 extruders are active
use soluble raft when support is disabled and 2 extruders are active
hide parameters to select first layer extruder (toolpath always starts with the left extruder)
never use interleaved prime tower for Method printers
hide raft jerk & accel parameters

# How Has This Been Tested?

Copied new files into an internal Cura build (sep19) then sliced and / or printed dozens of tests.


wawanbreton (approved):

cura/PrinterOutput/FormatMaps.py
```diff
@@ -44,7 +45,7 @@ class FormatMaps:
          "cpe": {"name": "CPE", "guid": "da1872c1-b991-4795-80ad-bdac0f131726"},
          "hips": {"name": "HIPS", "guid": "a468d86a-220c-47eb-99a5-bbb47e514eb0"},
          "tpu": {"name": "TPU 95A", "guid": "19baa6a9-94ff-478b-b4a1-8157b74358d2"},
-         "im-pla": {"name": "Tough", "guid": "de031137-a8ca-4a72-bd1b-17bb964033ad"}
+         "im-pla": {"name": "Tough", "guid": "96fca5d9-0371-4516-9e96-8e8182677f3c"}
```
wawanbreton, line 48: Why are the GUIDs for abs, petg and im-pla changed ? This has to be consistent with DF.

wawanbreton (approved):

pkuiper-ultimaker (approved):

PR #19703, opened 2024-10-01
[CURA-12097] Upgrader for 5.8 -> 5.9 version.

rburema:


wawanbreton (approved):
Code is fine, but IMHO the upgrade script would deserve some documentation regarding the purpose of the conversion, indicating that this is following the integration of the gradual flow plugin. On the other hand, this is just an upgrade script, se we don't really care.

Side-note: I only reviewed the Python scripts, I trust you on the 5000+ changed files :sweat_smile:

Issue #19716, opened 2024-10-06
[5.8.1] Infill Line Distance dosen't change


Alexb92360:
### Cura Version

5.8.1

### Operating System

Windows 11

### Printer

Ankermake M5C

### Reproduction steps

1. Select the Infill Density

### Actual results

The infill line distance dosen't change, i have to insert an absolute value

### Expected results

The Infill line distance have to be calculated automatically, take note that there isn't the "Fx" button next the value to be auto calculated

### Add your .zip and screenshots here ⬇️

![image](https://github.com/user-attachments/assets/a01ae415-bdb3-43ab-81c6-f978c5f8519d)

Alexb92360:
So how can be fixed?

Inviato da Posta per Windows

Da: HellAholic
Inviato: lunedì 7 ottobre 2024 22:11
A: Ultimaker/Cura
Cc: Alexb92360; Author
Oggetto: Re: [Ultimaker/Cura] [5.8.1] Infill Line Distance dosen't change(Issue #19716)

The definition file for the printer is overriding the formula defined in fdmprinter, so it will not try to use the formula to calculate the value and defaults to either the value provided in printer definition, quality profile, or user defined.
image.png (view on web)
Formula in fdmprinter
image.png (view on web)

PR #19723, opened 2024-10-07
Remove unnecessary override

HellAholic:
Bug fix, value should not be overridden since it cancels out the formula calculating the distance based on the sparse  
Update, the override is setting the value to the value that the parent definition would calculate.
Resolves https://github.com/Ultimaker/Cura/issues/19716

image
image

### DIFF
--- a/resources/definitions/ankermake_m5c.def.json
+++ b/resources/definitions/ankermake_m5c.def.json
@@ -67,7 +67,6 @@
             ]
         },
         "infill_extruder_nr": { "value": -1 },
-        "infill_line_distance": { "default_value": 8 },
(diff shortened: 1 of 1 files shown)


wawanbreton:
What is the point of setting this hard-coded 8 value then ? If I'm correct, it will never be used in normal conditions within Cura. I'm fine with it, but at some point, somebody probably put this value for a good reason ?

wawanbreton:
So could you adjust the formula for the printer, so that it gives an 8 at 10%, and something appropriate for other values ?

wawanbreton:
Do you want to redirect to 5.9 ?

wawanbreton (approved):

PR #19733, opened 2024-10-10
Using the new cool_during_extruder_switch feature to switch the Metho…

pkuiper-ultimaker:
The Method machines actively cool the nozzle at the end of a layer switch. Use the new cool_during_extruder_switch feature to enable this behavior and remove it from the start and end GCode template.

PP-530

### DIFF
--- a/resources/definitions/ultimaker_method_base.def.json
+++ b/resources/definitions/ultimaker_method_base.def.json
@@ -158,6 +158,7 @@
         "bridge_wall_material_flow": { "value": "material_flow" },
         "bridge_wall_speed": { "value": "speed_wall" },
         "brim_width": { "value": 5 },
+        "cool_during_extruder_switch": { "value": "'only_last_extruder'" },
         "cool_fan_enabled":
         {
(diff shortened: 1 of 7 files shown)


alanbjorklund (approved):
I printed a Method X slice, the fans behaved as expected.

wawanbreton (approved):
Looks good !
In the future, could you consider making a base extruder definition for all Method printers, and inherit from it ? Their definition looks extremely similar. I don't see any inheriting extruder definition though, so maybe there is a reason for not doing it ?

wawanbreton:
@pkuiper-ultimaker I rebased the changes on 5.9 branch. If you have future PRs, please don't forget to base them on 5.9

pkuiper-ultimaker:
> Looks good ! In the future, could you consider making a base extruder definition for all Method printers, and inherit from it ? Their definition looks extremely similar. I don't see any inheriting extruder definition though, so maybe there is a reason for not doing it ?

I assumed that inheriting extruder ptofiles was not possible, because I have not seen examples. But perhaps it works. I will give it a try.

PR #19755, opened 2024-10-15
[CURA-11815] Fix 'early' auto-slice.

rburema:
This could cause slicing to hang forever until settings where changed. So stop too-early slicing, but keep a memo of whether we've already been requested to slice in the initialization interval, so auto-slice doesn't fail either.


wawanbreton (approved):
That makes sense !

PR #19762, opened 2024-10-16
[CURA-12155] Was too small to fit in the start/end-gcode in new printers.

rburema:
This might have some impact on the min window size you can use Cura with, but it was getting a bit silly.


wawanbreton (approved):
I completely approve !!

PR #19763, opened 2024-10-16
Small corrections for 5.9 Stable

pkuiper-ultimaker:
PP-533

### DIFF
--- a/resources/definitions/ultimaker_method_base.def.json
+++ b/resources/definitions/ultimaker_method_base.def.json
@@ -342,6 +342,11 @@
         "multiple_mesh_overlap": { "value": 0 },
         "optimize_wall_printing_order": { "value": true },
         "prime_blob_enable": { "enabled": false },
+        "prime_tower_enable":
+        {
+            "default_value": "extruders_enabled_count > 1",
+            "enabled": "extruders_enabled_count > 1"
+        },
         "prime_tower_flow": { "value": "material_flow" },
         "prime_tower_line_width":
         {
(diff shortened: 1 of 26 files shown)


resources/definitions/ultimaker_method_base.def.json
```diff
@@ -342,6 +342,11 @@
         "multiple_mesh_overlap": { "value": 0 },
         "optimize_wall_printing_order": { "value": true },
         "prime_blob_enable": { "enabled": false },
+        "prime_tower_enable":
+        {
+            "default_value": "extruders_enabled_count > 1",
+            "enabled": "extruders_enabled_count > 1"
+        },
```
wawanbreton, line 349: This should not be necessary. This setting is already disabled based on the same criterion, and also has a proper resolve function. Do you have cases where it still makes a prime tower with only 1 extruder ?

resources/definitions/ultimaker_sketch.def.json
```diff
@@ -148,6 +163,7 @@
             "enabled": true,
             "value": "resolveOrValue('print_sequence') != 'one_at_a_time'"
         },
+        "print_sequence": { "enabled": false },
```
wawanbreton, line 166: Is there anything wrong with ona-at-a-time for sketch ?

alanbjorklund: Its untested and we are missing some gantry measurements.  I'll make sure we test it before the next Cura release.

resources/definitions/ultimaker_method_base.def.json
alanbjorklund: I was getting no prime tower for Method X PLA + PLA, but maybe there was an issue with my Cura resources files.

resources/definitions/ultimaker_sketch.def.json
wawanbreton: Ok, makes sense to disable it for now then

resources/definitions/ultimaker_method_base.def.json
pkuiper-ultimaker: I removed the enable and set the default value to true. This has the desired behavior (Method X PLA + PLA shows a prime tower)

wawanbreton (approved):

wawanbreton:
For some reason, on Sketch printers the `Build Plate Temperature Initial Layer` doesn't follow the `Build Plate Temperature`, is that on purpose ?

PR #19770, opened 2024-10-17
CURA-12185 add sketch thumbnail sizes

wawanbreton:
Sketch printer deserves the same thumbnails


jellespijker (approved):
LGTM;

Added a ticket [NP-462](https://ultimaker.atlassian.net/browse/NP-462) on our backlog to make sure we use the same number of png's

PR #19787, opened 2024-10-21
Make Metallic PLA 175 color option

alanbjorklund:
* Remove Metallic intent files
* hide metallic PLA from Method bots

PP-534


wawanbreton:
@alanbjorklund It looks like this branch was not initiated fro 5.9

alanbjorklund:
> @alanbjorklund It looks like this branch was not initiated fro 5.9

Yes, thanks.  I did not expect it to get into the Beta release.  Can we get this into stable?

wawanbreton:
Well right now I'm having trouble seeing what the actual changes are because of the wrong origin branch. But according to your description it doesn't look risky, so maybe.

alanbjorklund:
Ok, thanks.  I was just advised that there might be a complication with implementing this PLA color change in NeoPrep, so lets hold off on merging until we sort that out. @wawanbreton  cc @pkuiper-ultimaker

alanbjorklund:
I'll break this up into smaller PRs for Cura 5.10

PR #19796, opened 2024-10-23
Re-write of Search and Replace

GregValiant:
Added some options:
- Replace only the 'First Instance' of the Search term.
- Limit the Search to a layer range.
- 'Ignore StartUp G-code'
- 'Ignore Ending G-code'

# Description

Added flexibility to the script.  The options still allow a full Search and Replace as well as a limited Search and Replace.

## Type of change

# How Has This Been Tested?

I have been using this version for a year with multiple Cura versions.

**Test Configuration**:
Win10 Pro on a laptop.
UM Cura 4.13.1 to 5.9beta.

# Checklist:

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
+++ b/plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
@@ -1,16 +1,16 @@
 # Copyright (c) 2017 Ghostkeeper
 # The PostProcessingPlugin is released under the terms of the LGPLv3 or higher.
+# Altered by GregValiant (Greg Foresi) February, 2023.
+#    Added an option to replace the 'First Instance Only'
+#    Added option to limit the Search to a layer range.
+#    Added 'Ignore StartUp G-code' and 'Ignore Ending G-code' options
 
-import re # To perform the search and replace.
-
+import re
 from ..Script import Script
-
+from UM.Application import Application
 
 class SearchAndReplace(Script):
-    """Performs a search-and-replace on all g-code.
-
-    Due to technical limitations, the search can't cross the border between
-    layers.
+    """Performs a search-and-replace on the g-code.
     """
 
     def getSettingDataString(self):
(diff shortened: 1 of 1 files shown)


wawanbreton:
Thanks, we will take a look when we are not overbusy :blush: 
For internal reference: CURA-12325

fieldOfView:
There seems to be a lot of unnecessary code duplication after line 170

GregValiant:
Thanks @fieldOfView.
I must have missed the first few suggestions.  The latest commit addresses the issues from line 170.  I'll look at the others here in a minute.

A quick squash and all the changes are in that last commit.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +22,175 @@ def getSettingDataString(self):
...
+            if not is_regex:
+                search_string = re.escape(search_string)
+            search_regex = re.compile(search_string)
+            for num in range(start_index, end_index, 1):
+                layer = data[num]
+                if re.search(search_regex, layer) and replaceone == False:
+                    data[num] = re.sub(search_regex, replace_string, data[num], 1)
+                    replaceone = True
+                    break
+                if replaceone: break
```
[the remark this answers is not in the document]
GregValiant: I think that was vestigial code.  There may have been a second "for" in there and I used the second 'break' to back all the way out.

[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +22,175 @@ def getSettingDataString(self):
...
+                    "description": "When enabled the Ending Gcode is unaffected.",
+                    "type": "bool",
+                    "default_value": true,
+                    "enabled": "not enable_layer_search"
                 }
             }
         }"""
 
     def execute(self, data):
+        curaApp = Application.getInstance().getGlobalContainerStack()
```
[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +22,175 @@ def getSettingDataString(self):
...
+        ignore_start = self.getSettingValueByKey("ignore_start")
+        ignore_end = self.getSettingValueByKey("ignore_end")
+        if enable_layer_search:
+            ignore_start = True
+            ignore_end = True
+        first_instance_only = bool(self.getSettingValueByKey("first_instance_only"))
 
-        for layer_number, layer in enumerate(data):
-            data[layer_number] = re.sub(search_regex, replace_string, layer) #Replace all.
+    #Find the raft and layer:0 indexes--------------------------------------------------------------------------
```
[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+                    "label": "Replace with:",
+                    "description": "The 'Search For' text will get replaced by this text.  For MultiLine insertions use the newline character 'backslash plus n' as the delimiter. If your Search term ends with a 'newline' remember to add 'newline' to the end of this Replace term.",
                     "type": "str",
                     "default_value": ""
                 },
                 "is_regex":
                 {
                     "label": "Use Regular Expressions",
-                    "description": "When enabled, the search text will be interpreted as a regular expression.",
+                    "description": "When disabled the search string is treated as a simple text string.  When enabled, the search text will be recompiled as a 'regular' python expression.",
```
wawanbreton, line 41: I don't understand what you mean with `'regular' python expression`. The text is actually interpreted as what is truely called a 'regular expression', so I find the previous text more explicit (with the enabled/disable addition)

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
 
-        for layer_number, layer in enumerate(data):
-            data[layer_number] = re.sub(search_regex, replace_string, layer) #Replace all.
+        #Find the raft and layer:0 indexes
+        raft_start_index = 0
+        layer_0_index = 0
+        start_index = 1
+        end_index = len(data)
+        try:
+            for l_num in range(2,12,1):
```
wawanbreton, line 127: I do agree that it is very unlikely to have more that ~12 raft layers, but I think this magic number could be avoided quite easily

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+                    layer_0_index = l_num
+                    break
+            if raft_start_index == 0:
+                raft_start_index = layer_0_index
+                raft_layers = 0
+            elif raft_start_index < layer_0_index:
+                raft_layers = layer_0_index - raft_start_index
+            else:
+                raft_layers = 0
+        except:
```
wawanbreton, line 141: If this `except` is supposed to catch a specific case, please define the type of caught exception (`KeyError` probably ?) and add a small comment

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+            if not enable_layer_search:
+                if ignore_start:
+                    start_index = 2
+                else:
+                    start_index = 1
+                if ignore_end:
+                    end_index = len(data) - top_layer
+                else:
+                    end_index = len(data)
+            elif enable_layer_search:
```
wawanbreton, line 155: If I'm correct, `enable_layer_search` will always be a valid bool, right ? In this case you dont need the `elif` but just an `else`

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+        except:
+            pass
+            
+        #Determine the actual start and end indexes of the data
+        try:
+            if not enable_layer_search:
+                if ignore_start:
+                    start_index = 2
+                else:
+                    start_index = 1
```
wawanbreton, line 150: Please add an empty line between if/else blocks to help separate them visually

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+                if ignore_end:
+                    end_index = len(data) - top_layer
+                else:
+                    end_index = len(data)
+            elif enable_layer_search:
+                if start_layer < 1 and start_layer != -6:
```
wawanbreton, line 156: Ok, why -6 ? :smile:

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
+                    start_index = 2
+                else:
+                    start_index = raft_start_index + start_layer - 1
+                if end_layer == -1:
+                    end_index = len(data) - top_layer
+                else:
+                    end_index = raft_start_index + int(end_layer)
+                if end_index > len(data) - 1: end_index = len(data) - 1 #For possible user input error
+                if int(end_index) < int(start_index): end_index = start_index #For possible user input error
+        except:
```
wawanbreton, line 168: I do not see anything that could generate an exception in this piece of code, is the `try/except` somehow necessary ?

wawanbreton (changes requested):
Hi @GregValiant, the code looks overall good and I do like the new settings very much, this is a real improvement :slightly_smiling_face: 
I do have some questions, and a few nitpicking code-style remarks but nothing serious

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
fieldOfView: If anything other than `regular expression`, it should be `Python regular expression`; there are minute differences in how Python, PERL, Javascript interpret regular expressions.

wawanbreton: Yeah true, that would be the most accurate naming

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
             {
                 "search":
                 {
-                    "label": "Search",
-                    "description": "All occurrences of this text will get replaced by the replacement text.",
+                    "label": "Search for:",
+                    "description": "CASE SPECIFIC.  'LAYER' is not the same as 'Layer'.  All occurrences of this text (within the search range) will be replaced by the 'Replace with' string.",
```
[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
```diff
@@ -23,37 +23,170 @@ def getSettingDataString(self):
...
                     "type": "str",
                     "default_value": ""
                 },
                 "replace":
                 {
-                    "label": "Replace",
-                    "description": "The search text will get replaced by this text.",
+                    "label": "Replace with:",
+                    "description": "The 'Search For' text will get replaced by this text.  For MultiLine insertions use the newline character 'backslash plus n' as the delimiter. If your Search term ends with a 'newline' remember to add 'newline' to the end of this Replace term.",
```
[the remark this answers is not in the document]
GregValiant: I couldn't figure out a way to put \n into the line without it throwing an error and refusing to load.
That's how I ended up going with the text descriptions.  If you can manage it I'd appreciate it.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
GregValiant: You get a chuckle here.  That is part of my "stumble around until I find something that works" coding style.
I changed it to an Enumerate which does eliminate the magic number.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
GregValiant: It is my experience that rafts _always_ start on LAYER: -6 regardless of the number of raft layers.  That's why there can be gaps in the numbers between the top raft layer and LAYER:0.  The layer numbering often goes -6, -5, -4, -3, 0, 1, 2.

You tried to catch me with another magic number there didn't you.  In this case it belongs to Cura.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
GregValiant: It had to do with the gaps that existed between the top raft layer (which might have been -2) and layer:0.  Since that no longer seems to occur maybe the "try / except" can go away.
This is part of the "-6" thing.  It has to do with the preview numbering and converting that to the gcode numbering.
I need to think about that some more.

GregValiant: The raft numbering is still an issue.  The -6 won't work if there are more than 6 raft layers, or if there is a large airgap.  It also seems to be dependent on whether supports are being build on top of the raft prior to the model starting.
The truly annoying part is that I have never used a raft.

GregValiant:
Much simpler.
One list for layers and one list for data and a quick compare to figure out the start_index and end_index.
As an added bonus - no magic numbers and no 'try/except'.

plugins/PostProcessingPlugin/scripts/SearchAndReplace.py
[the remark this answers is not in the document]
wawanbreton: Ok, that was tricky indeed. I had to use a raw string `r"..."` in combination with double backslash, one should have been enough but the text is maybe resolved multiple times. 4 backslashes also work but is less elegant :smile: 
FYI I also had to set the layer search regex as a raw string, as it contains a backslash.

GregValiant:
Looks good.
As I final thing, I think you should change the year in my "by line" to 2025.

"\\n" eh?
That's making me feel sheepish.  I tried several different ways to get '\n' displayed rather than acted on and that obviously wasn't one that occurred to me.

Thanks.

wawanbreton:
> That's making me feel sheepish. I tried several different ways to get '\n' displayed rather than acted on and that obviously wasn't one that occurred to me.

That's ok, I have faced similar issues A LOT of times, within different languages/encodings/file types/... so I know how annoying it can be :slightly_smiling_face:

PR #19798, opened 2024-10-23
Refactor binaries filtering and update dependencies

jellespijker:
Introduce a mechanism to filter blacklisted binaries from the build configuration and update the conandata.yml with a new blacklist section. Additionally, remove unnecessary dependencies from requirements.txt, particularly entries for PyQt6-NetworkAuth and related packages.

Contribute to CURA-12229

---

_EDIT(by rburema)_: I'm not a great fan of some of the (read: my) commits -- it's very messy and should be refactored, but it works just fine and could be used for the beta-builds.


conanfile.py
```diff
@@ -217,6 +217,62 @@ def _generate_cura_version(self, location):
...
+            "qtdatavis3d",
+            "qtlottie",
+            "qtmqtt",
+            "qtnetworkauth",
+            "qtquick3d",
+            "qtquick3dphysics",
+            "qtquicktimeline",
+            "qtvirtualkeyboard",
+            "qtwayland"
+        ]
```
wawanbreton, line 235: Shouldn't this list also be in the conandata.yml ?

conanfile.py
```diff
@@ -217,6 +217,62 @@ def _generate_cura_version(self, location):
...
+                still_exist = True
+                for forbidden in prohibited:
+                    if forbidden.lower() in str(pathname).lower():
+                        to_remove_files.append(pathname)
+                        still_exist = False
+                        break
+                if not still_exist:
+                    continue
+                with open(pathname, "rb") as file:
+                    bytez = file.read().lower()
```
wawanbreton, line 253: Are you looking for library dependancies by opening the binary file and searching for specific names inside them ? This is a bit cavalier :sweat_smile:

wawanbreton (approved):
I'm still no fan of having this in the conanfile, but those changes are consistent.

wawanbreton:
Merging anyway because we need this now. We will discuss refining it later. if ever.

PR #19799, opened 2024-10-24
[CURA-12188] Fix build plate z-fighting

rburema:
The build-plate grid and the printer model where on top of each other, and the additional sorting didn't apply (properly) since both of them where a different render-type.

(With bonus commit that solves rendering outer-wall line-starts when doing seam-scarfs.)


wawanbreton (approved):
Code is fine, but can you not get rid of the `has_textured_buildplate` attribute ? The presence of a non-empty `platform_texture` should be enough, right ?

wawanbreton:
> Code is fine, but can you not get rid of the `has_textured_buildplate` attribute ? The presence of a non-empty `platform_texture` should be enough, right ?

Oh no, forget that. There are meshes with textures that are not for the build plate. It is required then.

Issue #19438, opened 2024-07-27
Improve Japanese localization


h1data:
### Cura Version

5.7.2

### Operating System

not specific

### Printer

not specific

### Reproduction steps

Hello, I've recently considered improving Cura's Japanese localization.
(and discovered issue #19401)

Based on the [Translating Cura](https://github.com/Ultimaker/Cura/wiki/Translating-Cura) wiki page, I have been preparing PR for 5.7.2, but other translation words were added in the 5.8 branch.
Also, it is said your team uses Lokalise translation service in #18614.

How do I join the translation work?

### Actual results

There are problems in Japanese localization.

#### Example
```
msgctxt "@action:inmenu"
msgid "Toggle Full Screen"
msgstr "留め金 フルスクリーン"
```

"留め金" might be from "Toggle", but rather a metal hook that holds doors/gates.<br> 
It seems due to bad machine translation and not checked in the actual program.<br>
(it would be "全画面表示切替")

That is just an example and there are other mistranslations.

### 2. Inconsistent terms

The common terms in the program should be consistent but there are some inconsistent translations.

For example, "material" is a common term but there are different translations; "フィラメント" (filament), "材料" (translated straightly), or "マテリアル" (katakana for "material").

### 3. misuses of `&` operators for the menu shortcut keys.

`&x` indicates the shortcut key `x` in menus, however, there are misuses for `&` operators.

#### Example
```
msgctxt "@title:menu menubar:toplevel"
msgid "&File"
msgstr "&ファイル"
```
`&フ` is not recognized as a shortcut key, so just displayed as "&ファイル".

In most Japanese software, the menu shortcut key is displayed in parenthesis like "ファイル(<ins>F</ins>)", thus it would be "ファイル(&F)". 

#### Example
```
msgctxt "@text"
msgid ""
"- Add material profiles and plug-ins from the Marketplace\n"
"- Back-up and sync your material profiles and plug-ins\n"
"- Share ideas and get help from 48,000+ users in the UltiMaker community"
msgstr "- マーケットプレースから材料プロファイルとプラグインを追加- 材料プロファイルとプラグインのバックアップと同期- UltiMakerコミュニティで48,000人以上のユーザーとアイデアを共有してアドバイスをもらう"
```
Those lines are not broken into lines. (see additional information)

There should be `\n` at each end and split into lines like below.

```
msgctxt "@text"
msgid ""
"- Add material profiles and plug-ins from the Marketplace\n"
[...]

PR #19473, opened 2024-08-04
improve Japanese localization

h1data:
# Description

Improving Japanese localization 
Resolves: #19438

## Type of change

- [x] Translations

# How Has This Been Tested?

- [x] Compiled "po" files to "mo"  and checked the view of the application.

**Test Configuration**:
* Operating System: Windows 10 

# Checklist:

- [x] My code follows the style guidelines of this project as described in [UltiMaker Meta](https://github.com/Ultimaker/Meta) and [Cura QML best practices](https://github.com/Ultimaker/Cura/wiki/QML-Best-Practices)
- [x] I have read the [Contribution guide](https://github.com/Ultimaker/Cura/blob/main/CONTRIBUTING.md) 
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have uploaded any files required to test this change

PR #19817, opened 2024-10-27
improve Japanese translation for 5.9

h1data:
# Description

Improving Japanese localization for 5.9
Resolves: #19438

## Type of change

- [x] Translations

# How Has This Been Tested?

- [x] Compiled *.mo file by gettext and checked the appearance of the built application.

**Test Configuration**:

# Checklist:

### DIFF
--- a/resources/i18n/ja_JP/cura.po
+++ b/resources/i18n/ja_JP/cura.po
@@ -3,9 +3,9 @@ msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2024-03-13 09:02+0100\n"
-"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
-"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
-"Language-Team: LANGUAGE <LL@li.org>\n"
+"PO-Revision-Date: 2024-10-27 12:31+0000\n"
+"Last-Translator: h1data <h1data@users.noreply.github.com>\n"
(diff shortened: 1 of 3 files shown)


h1data:
@wawanbreton Here is the Japanese translation updated for 5.9.

Our changes in #19473 were reverted in 546df30,
so I updated the changes again with additional corrections since #19473.
Lokalize folks might have only applied template *.pot files except revised *.po files.
Could you inform them about the translation changes?

wawanbreton:
Thanks ! I'll take a look at the PR quickly.

> Could you inform them about the translation changes?

Not really, I am not in direct contact with the translators. However I am the one running the procedure to send the labels to be translated. This procedure is still being debugged, so maybe there is still something wrong with it. Or maybe I just got confused because your PR arrived at the same time as the translations and I did something wrong.

resources/i18n/ja_JP/cura.po
```diff
@@ -44,113 +44,114 @@ msgstr "%1m"
 
 msgctxt "@action:inmenu menubar:printer"
 msgid "&Add Printer..."
-msgstr "＆プリンターを追加する..."
+msgstr "プリンターを追加(&A)..."
```
wawanbreton, line 47: I love the idea of restoring the `&A`, but just for my curiosity, is it useable at all with a Japanese keyboard ?

h1data: Yes, it's the traditional style for shortcuts in Japanese applications.
image

Of course, I have confirmed it works in 5.9.1 like `Alt+a`.

wawanbreton: Ok nice, thanks for the explanation :slightly_smiling_face:

PR #19818, opened 2024-10-27
Updated Eazao printer settings to allow material selection.

Eazao:
Some time ago, we submitted the settings for M500, M600, M700, Potter and Zero printers. 

https://github.com/Ultimaker/Cura/pull/16294

We downloaded the latest Cura 5.9.0_beta.1 version and found that we could not select materials. 

According to the suggestions, we changed the retract settings to "true" and changed "has material" to "true".

https://github.com/Ultimaker/fdm_materials/pull/308

We would like to update these settings for the M500, M600, M700, Potter and Zero printers.

PR #19819, opened 2024-10-27
Remove obsolete overrides

HellAholic:
Replacing https://github.com/Ultimaker/Cura/pull/19818

- Removed overrides for has_materials
- Removed override for retraction_enable
- Removed override for cool_fan_enabled


wawanbreton:
I fafil to see why these variables can be removed, they do have different values in fdmprinter definition

wawanbreton (approved):

PR #19827, opened 2024-10-29
NavLib integration test

pskowronskiTDx:
This is a draft PR which introduces the integration with 3Dconnexion devices.


fieldOfView:
Shouldn't this be implemented as a plugin andhook into the controller instead of directly inside CuraApplication?

PR #19836, opened 2024-10-30
Add exclude material rule for eazao_clay

HellAholic:
CURA-12220

Exclude clay 175 material for the method and sketch


wawanbreton (approved):
Shouldn't we also do this for UM printers ?

wawanbreton:
Oh yes indeed, they should be fine :laughing:

PR #19846, opened 2024-11-01
Cura 12255 string fix translation

HellAholic:
CURA-12255

image

Comes with https://github.com/Ultimaker/conan-ultimaker-index/pull/15

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -1,16 +1,16 @@
-version: "5.9.0-beta.1"
+version: "5.9.0-beta.2"
 requirements:
-  - "cura_resources/5.9.0-beta.1"
-  - "uranium/5.9.0-beta.1"
-  - "curaengine/5.9.0-beta.1"
-  - "cura_binary_data/5.9.0-beta.1"
-  - "fdm_materials/5.9.0-beta.1"
+  - "cura_resources/5.9.0-beta.2"
+  - "uranium/5.9.0-beta.2"
+  - "curaengine/5.9.0-beta.2"
+  - "cura_binary_data/5.9.0-beta.2"
+  - "fdm_materials/5.9.0-beta.2"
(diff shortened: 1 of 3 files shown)


resources/i18n/cs_CZ/fdmprinter.def.json.po
```diff
@@ -1883,6 +1903,26 @@ msgctxt "magic_fuzzy_skin_outside_only description"
...
+msgstr ""
+
+msgctxt "variant_name"
+msgid "KLEMA 250 Twin Size"
+msgstr ""
+
+msgctxt "variant_name"
+msgid "KLEMA 500 Size"
+msgstr ""
+
```
rburema, line 1925: Discussion results: We don't really want these in here, alter the 'variant-name' script to listen to a new 'should translate' (default = false) in fdmprinter.

resources/i18n/cs_CZ/fdmprinter.def.json.po
```diff
@@ -3909,6 +3989,14 @@ msgctxt "brim_smart_ordering description"
...
 msgstr "Prohodí pořadí tisku vnitřní a druhé nejvnitřnější čáry límce. Toto usnadňuje odstraňování límce."
 
+msgctxt "variant_name"
+msgid "SwiftTool"
+msgstr ""
+
+msgctxt "variant_name"
+msgid "SwiftTool24"
+msgstr ""
+
```
rburema, line 3999: Same (see above).

rburema (changes requested):

wawanbreton:
We now have only 3 translated variants names:
* "Nozzle Size", used by default
* "Print Core", used by UM printers
* "Extruder", used by MB printers

rburema (approved):

PR #19858, opened 2024-11-05
Update Changelog for Second Beta

MariMakes:
# Description

Update the changelog for the second beta

### DIFF
--- a/resources/texts/change_log.txt
+++ b/resources/texts/change_log.txt
@@ -29,6 +29,16 @@
 - Fixed a bug that prevented changes to the initial layer diameter from being applied properly, contributed by @ThomasRahm
 - Fixed a rounding issue in the engine math, @0x5844
 
+* Bugs resolved during the first Beta
+- Spiralize outer contour no longer adds unnecessary seams to the model
+- When coasting is enabled the printhead no longer drops down 
+- Updated the strategy for sharpest corner - smart hiding seam location combination
+- Fixed a bug where long scarf seams and short scarf seam steps would result in the seam being moved outward
+- Fixed a bug where seam acceleration and deceleration were not working 
+- Triple-clicking values and searches that you entered now has you select everything so you can edit it in the field similar to other desktop applications
+- Updated dependencies and improved how unwanted and unused binaries are excluded from the building process. 
+- Added Eazao material for the new Eazao printers
+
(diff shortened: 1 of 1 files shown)


resources/texts/change_log.txt
```diff
@@ -29,6 +29,16 @@
 - Fixed a bug that prevented changes to the initial layer diameter from being applied properly, contributed by @ThomasRahm
 - Fixed a rounding issue in the engine math, @0x5844
 
+* Bugs resolved during the first Beta
```
rburema, line 32:
```suggestion
* Bugs resolved after the first Beta
```

resources/texts/change_log.txt
```diff
@@ -29,6 +29,16 @@
...
+- Spiralize outer contour no longer adds unnecessary seams to the model
+- When coasting is enabled the printhead no longer drops down 
+- Updated the strategy for sharpest corner - smart hiding seam location combination
+- Fixed a bug where long scarf seams and short scarf seam steps would result in the seam being moved outward
+- Fixed a bug where seam acceleration and deceleration were not working
```
rburema, line 37:
```suggestion
- Fixed a bug where seam acceleration and deceleration were not working 
- Fixed a bug where infill was wrongly removed from narrow spaces 
```

CURA-12068

resources/texts/change_log.txt
```diff
@@ -46,6 +56,12 @@
 - Updated Japanese translations by @h1data
 - Brazilian Portuguese by @Patola
 
+* Known issues
+- This second beta introduces a change that affects the UI so please use buttons, dropdowns, menus, and flows. We hope we did this correctly but it is a tricky change. If you see the crash please report it to us, and we will also see it show up in our crash tracking tool. 
+- There is a slight regression in terms of seam alignment in different strategies we hope to resolve before stable
+- We see an increase in micro-segments in the outer wall, which we hope to resolve before stable (an increase in microsegments in the infill is expected)
```
rburema, line 62:
```suggestion
- We see an increase in micro-segments in the outer wall, which we hope to resolve before stable 
```

That's because there is no infill otherwise -- the previous behaviour was just wrong, so we don't really have to mention it.

rburema (changes requested):
Looks great! I'd like a few changes though :-)

PR #19902, opened 2024-11-13
Fix layer height (0.1mm iso 0.15mm) for the normal mode of the Ultimaker 2

pkuiper-ultimaker:
PP-539


wawanbreton (approved):
Any idea why this got lost ?

pkuiper-ultimaker:
> Any idea why this got lost ?

I think we updated the default layer height in the past in ultimaker.def.json to move from 0.1mm to 0.2mm. This global quality file seemed to rely on this.

PR #19912, opened 2024-11-15
Updated Changelog for stable

MariMakes:
Proposal Updated Changelog for Stable, contributes to CURA-12285

### DIFF
--- a/resources/texts/change_log.txt
+++ b/resources/texts/change_log.txt
@@ -24,21 +24,30 @@
 - Fixed a bug that prevented the "Sync materials with printers" page from displaying when accessed immediately after starting Cura or signing in
 - Fixed a bug where reloading an updated model would result in not all the duplicates of the model being updated
 - Fixed a bug that showed incompatible materials when switching between printers with a different filament diameter. 
-- Fixed a bug where tree support in enclosed holes have missing layers
+- Fixed a bug where tree support in enclosed holes has missing layers
 - Made improvements in the G-codePathModify slot for Engine Plugins, contributed by @ThomasRahm
 - Fixed a bug that prevented changes to the initial layer diameter from being applied properly, contributed by @ThomasRahm
 - Fixed a rounding issue in the engine math, @0x5844
 
-* Bugs resolved after the first Beta
+* Bugs resolved since the Beta release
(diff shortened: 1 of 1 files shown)


resources/texts/change_log.txt
```diff
@@ -24,21 +24,30 @@
...
-* Bugs resolved after the first Beta
+* Bugs resolved since the Beta release
 - Spiralize outer contour no longer adds unnecessary seams to the model
 - When coasting is enabled the printhead no longer drops down 
-- Updated the strategy for sharpest corner - smart hiding seam location combination
+- Fixed coasting strategy so it's only applied to outer/inner walls, and not to skin, top-bottom, or infill
+- Fixed wiping direction when printing a single wall
+- Improved wiping in combination with the scarf seam
+- Significantly improved the strategy for sharpest corner - smart hiding seam location combination
 - Fixed a bug where long scarf seams and short scarf seam steps would result in the seam being moved outward
```
wawanbreton, line 39:
```suggestion
- Fixed a bug where long scarf seam length and short scarf seam steps would result in the seam being moved outward
```

resources/texts/change_log.txt
```diff
@@ -24,21 +24,30 @@
...
 - Fixed a bug where seam acceleration and deceleration were not working
```
wawanbreton, line 40:
```suggestion
- Fixed a bug where seam acceleration and deceleration around seam were not working
```

resources/texts/change_log.txt
```diff
@@ -48,6 +57,7 @@
 - Various setting improvements for ABS and Nylon for UltiMaker Method series
 - Various setting improvements for dual extrusion support for the UltiMaker Method series
 - Implemented a hard limit so the bed temperature cannot be more than 120C on all UltiMaker printers.
```
wawanbreton, line 59:
```suggestion
- Implemented a hard limit so the bed temperature cannot be more than 120°C on all UltiMaker printers.
```

wawanbreton (changes requested):
Small suggestions, fell free to decline.
Also I would like to add a know issue for the 3D performance degradation on Wayland. It does work now, didn't before, but users will see significant change if they are switching from X11 to Wayland.

Issue #14375, opened 2023-01-24
[macOS] UltiMaker Cura 5.2.2 has a damaged "bundle-identifier" (CFBundleIdentifier)


core-code:
### Application Version

5.2.2

### Platform

macOS

### Printer

None

### Reproduction steps

all previous versions of UltiMaker had a proper CFBundleIdentifier (`nl.ultimaker.cura`) but the newest download has a broken one (`nl.ultimaker.cura_UltiMaker_Cura_5.2.2`).

the version number is definitely not supposed to be inside your bundle identifier as per Apple's documentation.

also, due to the change from previous versions you'll loose all native macOS preferences and caches.

### Actual results

broken bundle identifier `nl.ultimaker.cura_UltiMaker_Cura_5.2.2`

### Expected results

bundle identifer should stay at `nl.ultimaker.cura`

### Checklist of files to include

- [ ] Log file
- [ ] Project file

### Additional information & file uploads

see above

MariMakes:
Hey @core-code,

Thanks for the report 👍 
This is outside my field of expertise but it seems to be related to building the .dmgs. 
I'll bring it up to the team to see what they can do to improve it. Fingers crossed 🤞 

Can you help me understand how this damaged bundle-identifier is affecting your use of Cura?

MariMakes:
Hey @core-code,

I discussed your issue with the team. 
We've added a ticket to the backlog with the intent of improving this. 

After digging into this it seems to effect permissions, and window locations and might require you to enter your password more often. They seem to be minor conveniences. Do you have other examples? 

For internal reference  CURA-10269

Thanks for the report! 👍

core-code:
yeah i'd file it as a "minor convenience"

core-code:
still damaged in 5.3.0 - bundle identifier is ( but should not be ):

`nl.ultimaker.cura_UltiMaker_Cura_5.3.0`

martinvandiemen:
Please change this back to `nl.ultimaker.cura`.

martinvandiemen:
When can we expect a fix for this?

MariMakes:
Hey @martinvandiemen,

We have a ticket on our backlog with the intent to find a new workaround for installing multiple versions using pkg, that should resolve this. It doesn't have a high priority so I don't expect it to be resolved very soon.

Can you share how this is an inconvenience for you? That will help us to reprioritize it if needed.

martinvandiemen:
[...]

PR #19932, opened 2024-11-20
Resolve issue 14375 ( https://github.com/Ultimaker/Cura/issues/14375 )

core-code:
Fixes https://github.com/Ultimaker/Cura/issues/14375


wawanbreton (approved):
Thanks for the change, it does make sense indeed that the bundle identifier remains the same. The `CFBundleVersion` should also be set properly, making updating properly working.

PR #20006, opened 2024-12-07
Update description Add (restart required)

HellAholic:
- Update the text to include (restart required)
- Update the description to indicate it only affects the model's Y axis values in the UI
- All other settings are unchanged such as printhead properties
Comes with a UM PR 
https://github.com/Ultimaker/Uranium/pull/983


HellAholic:
@GregValiant added your proposal here in the same PR ☝️

GregValiant:
Thanks for the help with both Rotate and Translate.
I've been using Cura for going on 5 years and I'd never noticed that the "Machine head with fans polygon" description has always been wrong.  That was certainly adding to the confusion.

rburema (approved):

PR #20013, opened 2024-12-09
Create PurgeLinesAndUnload.py

GregValiant:
# Description
This script has 4 options.
**Add Purge Lines:**
Will draw lines left, right, top, or bottom of the build plate and either full length of half length.  If a print takes up the entire width then the purge lines could be moved to the bottom.
**Circle Around To Layer Start:**
Creates an orthogonal tool path that moves the nozzle around the periphery of the build plate before moving in to the Layer Start location.  This keeps strings from dragging across the print location on the build plate.
**Adjust Starting E Location:**
Cura adds a retraction after the StartUp Gcode.  If there is also a retraction within the StartUp Gcode (or when using Add Purge Lines) it causes a double retraction.  This option will change that "G1 E-" line to a "G92 E" so the filament location within the nozzle can be adjusted to provide for an exact start to the skirt/brim/raft.
**Unload FIlament at Print End:**
Adds an unload sequence of G1 E commands to back the filament out of the hot end and out of the extruder.  Works well with long-tube bowden printers.

**The script supports:**
Absolute and Relative extrusion
Rectangular and Elliptic beds
Origin-At-Center whether True or False.
Disallowed Areas

# Settings
image

## Type of change

# How Has This Been Tested?

I've been using this script for a year.

**Test Configuration**:
Dell Laptop 16mb of RAM
* Operating System:
Windows 10 Pro
Cura versions 4.13.1 and up.

# Checklist:

- [ X] I apologize ahead of time for my coding style.


GregValiant:
@HellAholic the "Move around to the print start" function is something that came up in a post a while ago.  I did account for cases where Cura adds the "Move to prime tower" line at the end of the startup.  The script will add an orthogonal move to an edge before the actual move to the prime tower location.
I have high confidence in the code for "Rectangular" beds.  The G2/G3 coding in the "Elliptic" section seems OK but needs another set of eyeballs.
I'll leave this as a draft for now.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,745 @@
...
+        if self.getSettingValueByKey("move_to_start"):
+            self._move_to_start(data)
+        if self.getSettingValueByKey("adjust_starting_e"):
+            self._adjust_starting_e(data)
+        if self.getSettingValueByKey("enable_unload"):
+            self._unload_filament(data)
+        # Format the startup and ending gcodes
+        data[1] = self._format_string(data[1])
+        data[len(data) - 1] = self._format_string(data[len(data) - 1])
+        return data
```
HellAholic, line 145:
```suggestion
    def execute(self, data):
         # Mapping settings to corresponding methods
        procedures = {
            "add_purge_lines": self._add_purge_lines,
            "move_to_start": self._move_to_start,
            "adjust_starting_e": self._adjust_starting_e,
            "enable_unload": self._unload_filament
        }

        # Run selected procedures
        for setting, method in procedures.items():
            if self.getSettingValueByKey(setting):
                method(data)
        # Format the startup and ending gcodes
        data[1] = self._format_string(data[1])
        data[-1] = self._format_string(data[-1])
        return data
```
Just a different way of writing the same code suggestion. The benefit is that you can add/modify easier later.

HellAholic:
I'll take a look to see if I can update the code style. Will also try to play around with the current iteration to first get a grasp on how things work and what the output looks like. Might be a slow burner but it's on my todo list 👍

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
GregValiant: I made the suggested changes for Commit 2.
Also added support to "Adjust Starting E" to account for firmware retraction.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,752 @@
...
+        for setting, method in procedures.items():
+            if self.getSettingValueByKey(setting):
+                method(data)
+        # Format the startup and ending gcodes
+        data[1] = self._format_string(data[1])
+        data[-1] = self._format_string(data[-1])
+        return data
+
+    # Add Purge Lines to the user defined position on the build plate
+    def _add_purge_lines(self, data: str):
```
HellAholic, line 151: based on the code I assume the data is a list of gcode commands, so in that case, the type should be `data: list[str]` and you would return the data as `list[str]`

In few places I saw that you just do a return without passing the data back, I'll mark them, but I haven't gone through the code completely to know if it's working or not, that comes later :P

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,752 @@
...
+            if "G92 E0" in startup_section[num]:
+                insert_index = num
+                break
+            # In Relative Extrusion mode - insert above the M83 line
+            elif "M83" in startup_section[num]:
+                insert_index = num
+                break
+        startup_section.insert(insert_index, purge_str)
+        data[1] = "\n".join(startup_section)
+        return
```
HellAholic, line 385: here I think you would want to return the data

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,752 @@
...
+                break
+        move_str = self._format_string(move_str)
+        if move_str.startswith("\n"):
+            move_str = move_str[1:]
+        startup.append(move_str)
+        startup.append(lay_count)
+        data[1] = "\n".join(startup)
+        # Remove any double spaced lines
+        data[1] = data[1].replace("\n\n", "\n")
+        return
```
HellAholic, line 578: same here (return data)

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,752 @@
...
+            min_y = -abs((machine_depth / 2) - 7)
+        if prime_tower_x < midpoint_x and prime_tower_y < midpoint_y:
+            prime_tower_location = "LF"
+        elif prime_tower_x > midpoint_x and prime_tower_y < midpoint_y:
+            prime_tower_location = "RF"
+        elif prime_tower_x > midpoint_x and prime_tower_y > midpoint_y:
+            prime_tower_location = "RR"
+        elif prime_tower_x < midpoint_x and prime_tower_y > midpoint_y:
+            prime_tower_location = "LR"
+        return prime_tower_location
```
HellAholic, line 729:
```suggestion
        # Determine coordinate limits based on origin position
        if origin_at_center:
            midpoint_x, midpoint_y = 0, 0
            max_x, min_x = (machine_width / 2) - 7, -((machine_width / 2) - 7)
            max_y, min_y = (machine_depth / 2) - 7, -((machine_depth / 2) - 7)
        else:
            midpoint_x, midpoint_y = machine_width / 2, machine_depth / 2
            max_x, min_x = machine_width - 7, 7
            max_y, min_y = machine_depth - 7, 7

        # Determine quadrant of the prime tower
        if prime_tower_x < midpoint_x and prime_tower_y < midpoint_y:
            return "LF"  # Lower-Front
        elif prime_tower_x > midpoint_x and prime_tower_y < midpoint_y:
            return "RF"  # Right-Front
        elif prime_tower_x > midpoint_x and prime_tower_y > midpoint_y:
            return "RR"  # Right-Rear
        elif prime_tower_x < midpoint_x and prime_tower_y > midpoint_y:
            return "LR"  # Lower-Rear
```

you can do a bit of a trick and a reverse on the conditions

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,752 @@
...
+        elif purge_end_loc == "RR":
+            if prime_tower_loc == "LF":
+                adj_lines = f"G0 F{travel_speed} X{min_x} ; Move to edge\nG0 F600 Z0 ; nail down the string\nG0 F600 Z2 ; move up\n"
+            if prime_tower_loc == "RF":
+                adj_lines = f"G0 F{travel_speed} Y{min_y} ; Move to edge\nG0 F600 Z0 ; nail down the string\nG0 F600 Z2 ; move up\n"
+            if prime_tower_loc == "RR":
+                adj_lines = ""
+            if prime_tower_loc == "LR":
+                adj_lines = f"G0 F{travel_speed} X{min_x} ; Move to edge\nG0 F600 Z0 ; nail down the string\nG0 F600 Z2 ; move up\n"
+        return adj_lines
```
HellAholic, line 703:
```suggestion
        # Determine coordinate limits based on origin position
        if origin_at_center:
            midpoint_x, midpoint_y = 0, 0
            max_x, min_x = (machine_width / 2) - 7, -((machine_width / 2) - 7)
            max_y, min_y = (machine_depth / 2) - 7, -((machine_depth / 2) - 7)
        else:
            midpoint_x, midpoint_y = machine_width / 2, machine_depth / 2
            max_x, min_x = machine_width - 7, 7
            max_y, min_y = machine_depth - 7, 7

        # Adjustment templates
        move_x_edge = f"G0 F{travel_speed} X{{x}} ; Move to edge\nG0 F600 Z0 ; Nail down the string\nG0 F600 Z2 ; Move up\n"
        move_y_edge = f"G0 F{travel_speed} Y{{y}} ; Move to edge\nG0 F600 Z0 ; Nail down the string\nG0 F600 Z2 ; Move up\n"

        # Determine adjustment lines based on locations
        if purge_end_loc == "LF":
            if prime_tower_loc == "LF":
                return ""
            elif prime_tower_loc in {"RF", "RR"}:
                return move_x_edge.format(x=max_x)
            elif prime_tower_loc == "LR":
                return move_y_edge.format(y=max_y)

        elif purge_end_loc == "RR":
            if prime_tower_loc == "RR":
                return ""
            elif prime_tower_loc == "LF":
                return move_x_edge.format(x=min_x)
            elif prime_tower_loc == "RF":
                return move_y_edge.format(y=min_y)
            elif prime_tower_loc == "LR":
                return move_x_edge.format(x=min_x)

        return ""
```
similar to the other one, you can do couple of coding tricks.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
HellAholic: I think for the most part python doesn't really care about the type declaration (don't ask me how I figured out xD), but it helps during the code writing to recommend functions/operations that can be performed on it.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
HellAholic: side note, here you're not using the midpoint_x, midpoint_y & the bed_shape

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
GregValiant: It might be a post-processor thing but sometimes (maybe all the time) a "return data" will send the "data" list back to Cura assuming that all the work is done.  The script ends when that happens even though there might be more to do.

My understanding is that "data" is global so when the value is changed as in "data[1] = data[1].replace("\n\n", "\n")" then that is the value of data throughout the script and "return data" would be redundant.  That's why the only "return data" in the script was in the Execute function.

I'm self taught so my coding is decidedly un-pythonic.  It's more of a mish-mash of Python, Visual Basic, the Excel 4.0 macro language, and Fortran I.  I mention Fortran I because it was the only actual programming class I've ever taken.  That was in 1969.  I know where I want to go with the code, but the way I get there is rarely a straight line.

If you ever look at "Advanced Cooling Fan Control" you will be horrified.  Remco certainly was when he looked at it.  I think at some point he threw up his hands in disgust and just ran with it since the code worked.  It's just ugly.  There have been exactly zero bug reports about it.

Make whatever changes you feel are necessary because most of the time I don't really understand what you are asking.  This script is one I keep active for every print so I thought I'd pass it along.  It's handy.  If you want to throw it out be my guest.  It won't hurt my feelings.

HellAholic: I learned coding with Pascal so I have the same approach of first make it work and then worry about it not being pretty. Then many years later I had to make a 4D scanner and needed to process videos from 100 cameras into frames and sync them together through an audio trigger, so I started with python. So I see coding as a tool, if it gets the job done, that's all that matters :P (but it's personal opinion, it gets a bit difficult with more people in a team and that's why you get a coding style and such).

The return data thing was a question cause I attempted to run the script but it didn't generate any gcode commands for me, so that seemed like a usual suspect. But if you have it running then it might be something else I'm missing.

In some cases, I can't write anything better either, it's just a lot of checks and a lot of conditions, so the suggestions you see are mostly tips and tricks, they might help for something in future or they might not (making it harder to understand) that's why I add them as suggestions :) . I'm sure the devs will write helper functions and do some fancy stuff. But my questions or suggestions are mainly me trying to understand the problem or issues that you had and the process that you followed to resolve it which resulted in the code. 

TLDR:
I'll poke around some more, see if I can get it working on my machine.

GregValiant:
Good morning.
I missed the "machine_disallowed_areas" thing.  After looking at the MethodX disallowed areas I think simply excluding 'Purge Lines' and 'Move to Start' is the best way to go.  The polygons in a definition file can be written differently by different authors.  They are easy to draw, but tough to parse.
The same thing might be true for any multi-extruder machine that does not "Share Heater" and "Share Nozzle since both the Purge Lines and the Unload will be for the active nozzle.
I'll spend some time thinking on those.

Regarding not showing the post-processed gcode in the initial preview, I recall seeing complaints about that here in the bug reports.  The Initial Preview is done before any post-processing so the file would need to be saved, and then the gcode read back in.  That'll take a while.  Maybe a "Load Gcode file for Preview" button (that popped up after saving, or became yet another function of the "Slice" button) would be better so the user would have an option.  (If Arc-Welder is enabled then it won't show correctly anyway.)

On a different subject...
It would be nice if there was a ToolTip attached to each name in the "Post Processing Scripts" list.  Something like:
image
You would still need to load the PP to see the description, but there would be an indication of what it actually does.

HellAholic:
Aloha,
Let me know what you come up with, you can always add "single extruder only" to the script or something.
I checked the offset, it works somewhat, but also needs a secondary field for start and end of the purge line. There are some general patterns in how the gcode is made, so if I can make a template out of it, then adjusting the values would be easier. I'll see if I can make something that's more generic but simple and robust.

The slice result preview "should" in theory be simple enough, just have to figure out where I need to poke in the code to make it move. I think reloading the changed gcode is the doable part, but loading back to the original does not sound like a feasible option.

> On a different subject...
> It would be nice if there was a ToolTip attached to each name in the "Post Processing Scripts" list.
That should also be doable, but it requires some changes to the qml, adding a getter to the python with a pyqtsignal and such, and reading the description in a try/except to handle scripts without a description.
The main issue is that the qml for that part is fairly old, so it will probably need some updating:
- List is not properly spaced (line height / label height) -> this is mostly my OCD and other things being triggered
- The selected highlights is not great -> same as above xD
- The text (script name) is added to the list makes it so that the tooltip only appears if you hover over the left part (first 50 pixels or something)
- There is also always the risk of the unexpected bug popping out of nowhere

GregValiant:
I hate unexpected bugs.  The expected ones are much better.
I think I've got the "disallowed areas" figured out without having to exclude those printers.

GregValiant:
Wow.  I have been integrating the "Disallowed Areas" into the code.  There wasn't a lot of breakage, but there are more variable names to deal with (and there was already a lot).
Straightening that out is going to take a minute.  There are a lot of combinations to set up for to debug.
Fortunately, not many (if any) delta's have disallowed areas so I'm ignoring it for the G2/G3's.

I'll go through your suggestions and make the changes and then push another commit.

HellAholic:
It might also be an idea to use enum strings, doesn't even need to be linked to the string value it can be an int, but to keep it as close to the current implementation as possible:
```
from enum import Enum

# class syntax
class Location(Enum):
    LEFT_FRONT = "LF"
    LEFT_REAR = "LR"
    RIGHT_FRONT = "RF"
    RIGHT_REAR =  "RR"
    
 # usage
 if purge_end_loc == Location.LEFT_FRONT
```
makes it easier to work with strings in the comparisons and also reduces the change of making a typo, also helps with autocomplete xD

You can also do the calculations for some values and store them in the self, since it's being passed along to the different functions, that will reduce the number of variables you need to pass with each function call.
curaApp is one example, but probably you'll find other things that can be done the same way.
so if in your initialize() function instead of curaApp = ... you say self.curaApp = ...
then in all the following functions that have the self as the first argument, you can just reference it by saying:
self.curaApp.getProperty("something", "something")
the only thing you need to keep in mind is not adding the same name as the super class. (make unique names)

TLDR:
Add the bits you're comfortable with and I'll do a refactor afterwards, make sure functionality is there, I'll make it pretty :P

GregValiant:
I have to append this to my "List of things I didn't know".
image
**IS A LIE.**
Any gcode file (or UFP file) opened in Cura can be _post-post-processed_ if the "Save to..." button is clicked.  That extra post-processing generally makes a mess.

Plugins have a block so that doesn't happen.  I've blocked that from happening with this script.

HellAholic:
Although it's a bit of a rare usecase to reload a gcode and run a post processing script on it, since the information regarding the post processing is already in at the end of header of the gcode file it should be possible to allow for running additional post processing without re-running the same scripts that already were executed on the gcode.
So reading the gcode, looking if there is a POSTPROCESSED comment in it, if so, filtering out the scripts already executed, and executing the other ones.
```
;END_OF_HEADER
;POSTPROCESSED
;  [PurgeLinesAndUnload]
```

GregValiant:
In the next week or two I'm going to tear this down and put it back together.  It's working, but with the changes I've hacked in for the Prime Tower move and the Disallowed Areas it's turned into spaghetti code.  It's pretty bad when even I can't stand it.

So take a break.  I'll add a new commit when I get done changing it.

GregValiant:
@HellAholic 
I have made alterations, added a procedure, moved things around, found a couple of typos, and altered comments.
There are a lot of changes.
Do you want me to add a commit or go about updating this some other way?

HellAholic:
Hey Greg,
Add the commit and I'll work based on that. Took a break and didn't do much, mostly end of year work.

GregValiant:
There it is.  New and improved.

One of the printers I used for the Disallowed Areas was a UM3.  It has 6 areas defined and because I used a simple method of determining the "useable space" as a rectangle (rather than the actual shape), it ends up pushing the purge lines towards the print. There aren't many printers with the problem though, so the warning message I added should (hopefully) be sufficient.

HellAholic:
Adding a _v2 of the script to here mostly for safe-keeping, I'm just playing around with the code to see if I can simplify it a bit. Might break things so I'm doing it on a separate version to compare the output at the end.
I'll remove it from the PR once it's either at a good condition, or beyond saving :)

HellAholic:
@GregValiant I'm a bit confused by the `_move_to_start`
example below:
```
            if self.start_location == "LF":
                if goto_str == "LtFrt":
                    move_str += f"G0 F{self.speed_travel} X{self.machine_right - 5} ; Ortho move\n"
                    move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                    move_str += f"G0 F{self.speed_travel} Y{self.machine_front + 5} ; Ortho move\n"
                    move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
```
we want to move to the start location, and if I understand the code, we're already there, but the moves are to right front.
When you got some time, could you take another look at that section?

GregValiant:
There are a couple of possible options for where the "Move to Start" will originate.  Since it is the last alteration of the gcode, we know where it's going TO, but it might be starting out from the Right Rear if "Add Purge Lines" is selected, or it might be starting from the Prime Tower location if that move is in the Gcode.  So we don't really know the start point until other things are decided.

The design of the script was mostly done using 5.7 and 5.8.  The "Layer start X" and "Layer start Y" seem to be different in 5.9.0.  In earlier versions they would also dictate the start of the bed adhesion, but now the bed adhesion start seems to be related to the Z-seam location, or possibly the support Z seam location.

I'll take another look at it this evening.  I did remove one of the 'If "LF" ' sections because the nozzle was definitely going to be in the left front corner.  Maybe the one you noted can be taken out as well.

Since I brought up "going to..." I might have to abandon using the Layer Start X and Y in favor of reading the start of the first extrusion of the print.  That way if something is affecting the skirt/brim starting point it will be taken into account.

GregValiant:
I got side-tracked by the Start X and Y thing.
In that snippet it does appear that the first move should be to the "machine_left + 5".

if self.start_location == "LF":
                if goto_str == "LtFrt":
                    move_str += f"G0 F{self.speed_travel} X{**_self.machine_left + 5_**} ; Ortho move\n"
                    move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                    move_str += f"G0 F{self.speed_travel} Y{self.machine_front + 5} ; Ortho move\n"
                    move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"

Through the entire design stage I assumed (Hah!) that using the midpoint of the build plate to determine the "quadrant" would be correct.  Last night I loaded a small model and put it in the left-front.  It would seem that I should have used the model midpoint to determine the "starting quadrant".
With the model in the "LF", and the Z-seam at the "RR", if the model fits into the LF Quadrant, then the move still goes across the footprint of the model.
Now I'm thinking that the MIN and MAX of the model should be read, and the "midpoint" determined from those, rather than from the build plate size.

This is really making my head hurt.  It appears that "_get_print_extents" was abandoned-in-place.  I don't see it being used anywhere.  The only reference is "max_print_size = self._get_build_plate_extents()" and then max_print_size isn't used.

Merry Christmas by the way.

HellAholic:
Merry Christmas :)
I have a ticket planned to make the initial move an optional setting which is disabled by default and always listen to the user facing setting for Start X and Start Y, so that should make things more consistent.

So in theory, the start of the bed adhesion / skirt / support on the first layer should be the closest point to the Start X / Y, but the main issue is the printer definitions are community contributed over 7-ish years and they do not all follow the best practice. So that's why you see references to prime tower on a single extruder printer that cannot even enable prime tower. That's another challenge to standardize the definitions, and a separate subject.

For the move to start, if I boil it down until only the core remains we have this part:

```
        if self.bed_shape == "rectangular":
            if goto_str == "LtFrt" and self.start_location is not "LF":
                move_str += f"G0 F{self.speed_travel} X{self.machine_left + 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                move_str += f"G0 F{self.speed_travel} Y{self.machine_front + 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
            elif goto_str == "RtFrt" and self.start_location is not "RF":
                move_str += f"G0 F{self.speed_travel} X{self.machine_right - 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                move_str += f"G0 F{self.speed_travel} Y{self.machine_front + 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
            elif goto_str == "LtBk" and self.start_location is not "LR":
                move_str += f"G0 F{self.speed_travel} X{self.machine_left + 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                move_str += f"G0 F{self.speed_travel} Y{self.machine_back - 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
            elif goto_str == "RtBk" and self.start_location is not "RR":
                move_str += f"G0 F{self.speed_travel} X{self.machine_right - 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
                move_str += f"G0 F{self.speed_travel} Y{self.machine_back - 5} Z2 ; Ortho move\n"
                move_str += f"G0 F600 Z0 ; Nail down the string\nG0 Z2 ; Move up\n"
        elif self.bed_shape == "elliptic" and self.origin_at_center:
            radius = self.machine_width / 2
            offset_sin = round(2 ** .5 / 2 * radius, 2)
            if goto_str == "LtFrt":
                move_str += f"G0 F{self.speed_travel} X-{offset_sin} Z2 ; Move\nG0 Y-{offset_sin} Z2 ; Move to start\n"
            elif goto_str == "LtBk":
                if self.start_location == "RR":
                    move_str += f"G2 X0 Y{offset_sin} I{offset_sin} J{offset_sin} ; Move around to start\n"
                else:
                    move_str += f"G0 F{self.speed_travel} X-{offset_sin} Z2 ; Ortho move\nG0 Y{offset_sin} Z2 ; Ortho move\n"
            elif goto_str == "RtFrt":
                move_str += f"G0 F{self.speed_travel} X{offset_sin} Z2 ; Ortho move\nG0 Y-{offset_sin} Z2 ; Ortho move\n"
            elif goto_str == "RtBk":
                move_str += f"G0 F{self.speed_travel} X{offset_sin} Z2 ; Ortho move\nG0 Y{offset_sin} Z2 ; Ortho move\n"
        move_str += ";---------------------[End of layer start travels]"
```

I believe you're trying to over engineer things a bit too much. So if end of purge (start_location -> should rename to end_purge_location for clarity xD) and the Start of Layer are the same, we don't do anything. Otherwise regardless of where the print head is, we just simply tell it to go to that location and don't filter the extra X/Y moves, if the print head is at X or Y coordinates already then nothing should happen, if not, then it moves there.

Also since you're calculating machine_left/right/top/bottom based on the shape of the build plate and origin, then you can reuse the same format (I double checked, should be fine). So it makes the code much simpler and boiled down.

Of course you can add more to it, but if you have a generic base, you can add functions that trim the move based on the first extrusion location.

HellAholic:
I checked the output of V1 (Original Greg's production) vs V2
They are mostly the same, in V2 I changed the travel to start (now full move rather than only half) in the few scenarios I checked it looked good.
- Could not get my head around the rounded buildplates, at least no gcode viewer that I could find would display the arcs correctly
- Disallowed area will need adjustment, I propose just adding a manual offset by user if we want to keep it simple, otherwise I think using the functions of cura that calculates the allowed/disallowed area is an easier option than going at it from scratch.
- Also if you encounter weird front is back and back is front on the method printers that is correct/ish - it gets adjusted during change from gcode to toolpath

Give the V2 version a look, test a few things out, and let me know if it behaves as expected or not.

GregValiant:
You've been busy.  I took a look.  There was just one thing and it's what I've been thinking on the last day or so:

If the model is small and in the left front corner, then there won't be an initial move and the nozzle can cross the print area.
- I propose to make a "Y" move just in case of that scenario.  "start_x" and "start_y" could become "self.start_x" and "self.start_y".
- There was a typo (extra space character) in line 332.

Here is the file I altered.  The file is "...V2" and the script has been renamed "...V2" to avoid any more confusion than already exists.  I'll let you make any commit.
[PurgeLinesAndUnload_V2.zip](https://github.com/user-attachments/files/18253229/PurgeLinesAndUnload_V2.zip)

I have a macro that reads gcode files into AutoCAD and can translate the G2/G3 moves.  I'll go over them again to make sure they are correct.  (You will have noticed that I got lazy and didn't calculate any "half" purges for elliptic beds.)

GregValiant:
The G2's and G3's look good.

GregValiant:
I think that might be it.  Is it "Ready for Review"?

HellAholic:
The idea was just cleaning up and seeing if things still work. Also a mix of code review and code refactor, cause the functionality was there already. Pretty much when you're happy with the proposed V2 version you can add the code to your commits and remove the V2 version, I was just solving a puzzle here xD

GregValiant:
Ready for review.

HellAholic:
I've played around with the script a bit more, couple of things that might be nice to add:
- Currently only extruder[0] values are in effect -> if we can make it to purge for active extruder [0,1,2,3,4,...] then it would be more generic
- I tested on my S3 (I print PLA on extruder 0 and TPLA on extruder 1), so if I try the script on extruder 1, the offset of the nozzle means that the file is not accepted (out of bounds error). I assume it would be the case for other multi-extruders with a nozzle offset.
image

GregValiant:
Any line with an "E" in it that does not specifically state which tool will be for the active extruder.  That's why I didn't go any further into it.  The "purge lines" don't have a retraction so the retract_amount doesn't come up.  The purge lines would seem to be tool independent.

The one thing I do see is that if there is a "Nozzle switch retraction distance" then it might not get accounted for at the start.  In that case I don't know how the printer would know where the end of the filament is without the prime blob or something.
But it isn't really "changing tools" at the beginning, but if your last print ended with T0 and this one starts with T1, then the location of the end of the filament is either an unknown, or maybe the firmware knows.  If the end is 16mm back from the nozzle then it's going to take a very long purge line before it starts actually extruding.  The total purge distance for your printer looks to be about 6.6mm so it won't purge at all if the filament is in a "Nozzle switch" retracted location.
If "Prime Blob" is enabled, that should take into account a print start with the filament retracted and the purge lines should be fine without worrying about the active tool.

We start to get into a lot more combinations of things.  Since much of the start of a print on a UM machine is within the firmware, I can't check it out.

HellAholic:
So all of that is handled in FW for the UM printers, priming and such, no need to go that deep.
The only case I think we should pay attention to is the nozzle offset. My proposal is as follows:
- Adding a setting to the script for "Purge Extruder": "Default: 0, options: 0,1,2,..."
- Followed by nozzle offset X/Y that can either be retrieved from the extruder setting or manually typed by the user -> used as offset to the machine_(X->left/right)/(Y->front/back) && self.start_x, self.start_y
- Default behavior is extruder[0] so nothing changes for majority of single extruder printers
- If anything other than the extruder[0] is selected, the only affected changes would be the nozzle offset X/Y applied in calculations && the following settings:
```
        retract_dist = self.extruder[0].getProperty("retraction_amount", "value")
        retract_enable = self.extruder[0].getProperty("retraction_enable", "value")
        retract_speed = self.extruder[0].getProperty("retraction_retract_speed", "value") * 60
        material_diameter = self.extruder[0].getProperty("material_diameter", "value")
        mm3_per_mm = (material_diameter / 2) ** 2 * 3.14159
        init_line_width = self.extruder[0].getProperty("skirt_brim_line_width", "value")
        print_speed = round(self.extruder[0].getProperty("speed_print", "value") * 60 * .75)
```
Outside of this, I don't think it's necessary to worry too much about the FW settings in UM machines, I used it as an example to showcase why we might need to consider the nozzle offset X/Y and allowing for extruder selection by user.

GregValiant:
I understand.  If we are starting with T1 then the offset needs to be applied and the possibility of some other different parameters is there.
I'll take a look.  This shouldn't take long to add.  Debugging won't have to include a bunch of different printer definitions so that should go easier.

GregValiant:
Alright, I give...Where is "initial_extruder_nr" stored?

HellAholic:
I'm not 100% sure if there is a better way for it or not, based on where it lives in `CuraApplication.getInstance().getExtruderManager().getInitialExtruderNr()`

So need to:
```
```
then add the `self.initial_extruder_nr = None` to the __init__(self)
and when you're executing the script, you need to get the reference, if you attempt to get it in the init, it will crash or at least it did on my attempt.
Here it worked and I think it's the correct place to get the value since you're about to execute the script on the generated gcode, so all setting changes have been applied by the user.
```
    def execute(self, data):
        self.initial_extruder_nr = CuraApplication.getInstance().getExtruderManager().getInitialExtruderNr()
```
I moved the message for the startup warning in the execute to see if I get the correct extruder and it seems to be working,
image

GregValiant:
Thanks for that.  

Are these statements true?
_"Extruder 1 is always T0, and is always the primary extruder"
"Extruder 1 will never have offsets."_
Some machine definitions refer to "Right Extruder" and "Left Extruder".  I've always found that confusing.  I really didn't want to have to go through a bunch of definitions again.

I'm currently dealing with "Y" offsets and it is making my head hurt.  "X" offsets seems to be working fine BUT I have made the assumption that the above statements are in fact true (at least for UM machines).  Another assumption I've made is that any X offset will be a positive number, but the text boxes for offsets will allow negative numbers.

HellAholic:
>  "Extruder 1 is always T0, and is always the primary extruder"
That's how Cura is set up (you can look at support blockers for example, they get assigned to T0 as a mesh)

>  "Extruder 1 will never have offsets."
It might have an offset, it's machine dependent. For UM printers that is the correct assumption as far as I remember.

"Right Extruder" and "Left Extruder" in case of the UM printers is used as position, since you can use different extruder variants in left or right position on the head, so you need the reference to where the Extruder of Type A is placed (left position or right position). That allows you to make the extruder variant definitions themselves position agnostic.

For the negative/positive value, it will sort itself out, meaning:
- If you define the travel allowed area in the x axis as `0 - nozzle_offset_x < allowed < machine_width - nozzle_offset_x` then if nozzle offset is negative it would still shift in correct direction.
Same for the Y I think, so if you can work it for the positive, the negative is also gonna work without touching it.

GregValiant:
I didn't change that warning message.  Regardless of the extruder number, if there are purge lines in the Startup they are going to be commented out.

Here is the script with the extruder offsets considered.
Assumptions:
Delta printers will not have "disallowed areas".
T0 is the primary extruder and will not have Offsets.
If T1 has offsets then they will be positive "X" values and "Y" can be negative or positive.

I added a function at the bottom that determines the values for the "initial extruder".  Whether a print starts with T0 or T1 makes no difference and the proper values are applied.
I'll post it here and you can take a look.  If you request minor changes then I'll do a commit.
[PurgeLinesAndUnload.zip](https://github.com/user-attachments/files/18266654/PurgeLinesAndUnload.zip)

I'm getting on a motorcycle and getting out of here.  Have fun.

HellAholic:
Not gonna poke at it too much xD You're pretty much on point.
Just one comment:
```
    def _get_initial_tool(self) -> int:
        # Find the Initial Extruder
        self.init_ext_nr = Application.getInstance().getExtruderManager().getInitialExtruderNr()
        # If there is an extruder offset X then it will be used to adjust the "machine_right" and a Y offset will adjust the "machine_back"
        if bool(self.curaApp.getProperty("machine_use_extruder_offset_to_offset_coords", "value")):
            self.nozzle_offset_x = self.extruder[1].getProperty("machine_nozzle_offset_x", "value")
            self.nozzle_offset_y = self.extruder[1].getProperty("machine_nozzle_offset_y", "value")
        else:
            self.nozzle_offset_x = 0.0
            self.nozzle_offset_y = 0.0
        num = self.init_ext_nr
```
to
```
    def _get_initial_tool(self) -> int:
        # Find the Initial Extruder
        num = Application.getInstance().getExtruderManager().getInitialExtruderNr()
        # If there is an extruder offset X then it will be used to adjust the "machine_right" and a Y offset will adjust the "machine_back"
        if bool(self.curaApp.getProperty("machine_use_extruder_offset_to_offset_coords", "value")):
            self.nozzle_offset_x = self.extruder[num].getProperty("machine_nozzle_offset_x", "value")
            self.nozzle_offset_y = self.extruder[num].getProperty("machine_nozzle_offset_y", "value")
        else:
            self.nozzle_offset_x = 0.0
            self.nozzle_offset_y = 0.0
```
Since you're returning the num it will be assigned to self.init_extruder_nr in line 203 so no need for assigning it here.
I would use the initial extruder number for getting the nozzle offset still rather than defaulting to 1, since you might be printing with the first extruder T0 and applying the offset of T1.

Have fun and ride safe.

GregValiant:
_"I would use the initial extruder number for getting the nozzle offset still rather than defaulting to 1"_

On a dual-head printer with "side-by-side" nozzles - one of the nozzles must always have an offset.
One of the assumptions in the code is that T0 will always be the primary and that T1 will always be the one that is offset.  That will be true for very close to 100% of the supported printers.

If the "primary" nozzle is T1, then the code (as written) should probably exit with a message.  It might be possible to evaluate "T0 offset vs. T1 offset" and make a decision as to which one to use (and I'll think about that) but that seems a very small selection set of printers that would be defined that way.  I suppose it would be possible to have "T0 Xoffset = -12.5" and "T1 Xoffset = 12.5" but I've never seen that.

Changing from the hard coded "extruder[1]" to "extruder[num]" would keep the offsets from loading when T0 starts the print but that would allow "T0" to make it all the way to the right extents of the bed which would not be possible with hotends on a common printhead.

My thinking is that:  If there are offsets, they will be listed under T1, and those offsets are always in play regardless of which nozzle is in use.

HellAholic:
I get your point, without going into a multitude of checks it's not possible to do. Some machines can move with both extruders in the entire build plate area and others have physical limitations (S3 vs Method for example).
Then you can ignore that part. We're getting into the realm of expecting too much from a post processing script xD
I think it would be just easier to add it to the engine at this point :P

GregValiant:
I recall a post you made a couple of months ago where you indicated you were in favor of adding it directly to the engine rather than as a post-processor.  An alternative would be to change this to an Extension/Plugin.  The settings wouldn't be out of place in "Bed Adhesion".

With the additions of the disallowed areas and tool offsets, it has grown in scope.  As a post-processor (or plugin) it could be a stop-gap until changes are made to the engine.

I'll continue the debugging today but I haven't seen any issues and certainly no show-stoppers.

I did add "kick-outs" to both "Add Purge Lines" and "Circle around to..." if the particular printer has offsets assigned to T0.

GregValiant:
One change you made that I don't care for very much involves the end location of the "wipe" move at the end of the purge.
An example would be when the purge location is "Purge Left.   The nozzle ends up at Y=35 after the wipe, and the next move added might be to machine_right.  That X move is made at Y=35.  The original design had a move off the purge line first, then a move to Y=1, and then the move to machine_right.  That insures that the nozzle is moving around the periphery rather than 35mm into the build surface.  
The same thing happens if the purge lines are at "Purge Front" and the first move is to machine_depth.  That move is made at X=35 rather than X=1.0.

It requires a close examination of the gcode preview to see that travel move at Y=35 and it would be hard for most people to notice it if it was in fact a problem and ran across the footprint of their print.  You just know that someone will complain.

HellAholic:
I don't think you mean the ones with the purge_str += bit.
The other one is easy to adjust:
Copy over the code below for the move_to_location function.
```
    def _move_to_location(self, location_name: str, location: tuple) -> str:
        """
        Compare the input tuple (B) with the end purge location (A) and describe the move from A to B.
        Parameters:
            location_name (str): A descriptive name for the target location.
            location (tuple): The target tuple (e.g., ("right", "front")).
        Returns:
            str: G-code for the move from A to B or an empty string if no move is required.
        """
        # Validate input
        if len(self.end_purge_location) != 2 or len(location) != 2:
            raise ValueError("Both locations must be tuples of length 2.")

        # Extract components
        start_side, start_depth = self.end_purge_location
        target_side, target_depth = location

        moves = [f";MESH:NONMESH---------[Move to {location_name}]\nG0 F600 Z2 ; Move up\n"]

        # Helper function to add G-code for moves
        def add_move(axis: str, position: float) -> None:
            moves.append(
                f"G0 F{self.speed_travel} {axis}{position} ; Start move\n"
                f"G0 F600 Z0 ; Nail down the string\n"
                f"G0 F600 Z2 ; Move up\n"
            )

        # Compare sides
        if start_side != target_side:
            if target_side == Location.RIGHT:
                add_move("X", self.machine_left + 5)
                if start_depth == Location.FRONT:
                    add_move("Y", self.machine_front + 5)
                else:
                    add_move("Y", self.machine_back - 5)
                add_move("X", self.machine_right)
            else:
                add_move("X", self.machine_right - 5)
                if start_depth == Location.FRONT:
                    add_move("Y", self.machine_front + 5)
                else:
                    add_move("Y", self.machine_back - 5)
                add_move("X", self.machine_left)

        # Compare positions
        if start_depth != target_depth:
            if target_depth == Location.REAR:
                add_move("Y", self.machine_front + 5)
                if start_side == Location.LEFT:
                    add_move("X", self.machine_left + 5)
                else:
                    add_move("X", self.machine_right - 5)
                add_move("Y", self.machine_back)
            else:
                add_move("Y", self.machine_back - 5)
                if start_side == Location.LEFT:
                    add_move("X", self.machine_left + 5)
                else:
                    add_move("X", self.machine_right - 5)
                add_move("Y", self.machine_front)
        if len(moves) == 1 and self.start_y:
            moves.append(f"G0 F{self.speed_travel} Y{self.start_y} ; Move to start Y\n")
        # Combine moves into a single G-code string or return a comment if no movement is needed
        return "".join(moves) if len(moves) > 1 else f";----------[Already at {location_name}, No Moves necessary]\n"
```

HellAholic:
You can adjust the distance from the edge of the build plate, currently set to 5 based on the original code.
You can add a setting to the script which is "Travel distance from build plate edge" or something like that and just set that one :P 
Something along the lines of: `self.travel_margin = self.getSettingValueByKey("travel_distance_from_edge")` in the init, and then use the self.travel_margin instead of 5.

GregValiant:
I didn't get that fancy.
This doesn't seem to create any double moves, and puts the nozzle in the corner before going to a far edge.
I put it in right before "Compare Sides".  It does a double move and then dips to the plate.  Without the added Z moves it's a little quicker.
```
        # Move to a corner
        if start_side == Location.LEFT:
            moves.append(f"G0 F{self.speed_travel} X{self.machine_left + 6} ; Init move\n")
        elif start_side == Location.RIGHT:
            moves.append(f"G0 F{self.speed_travel} X{self.machine_right - 6} ; Init move\n")
        if start_depth == Location.FRONT:
            add_move("Y", self.machine_front + 6)
        elif start_depth == Location.REAR:
            add_move("Y", self.machine_back - 6)
        # Compare sides
```
That creates this:
```
;MESH:NONMESH---------[Move to Layer Start]
G0 F600 Z2                         ; Move up
G0 F7200 X223.0                    ; Init move
G0 F7200 Y223.0                    ; Start move
G0 F600 Z0                         ; Nail down the string
G0 F600 Z2                         ; Move up
G0 F7200 X1                        ; Start move
G0 F600 Z0                         ; Nail down the string
G0 F600 Z2                         ; Move up
;---------------------[End of layer start travels]
```

I just pushed a commit on this.

GregValiant:
Timing is everything...[Reddit](https://www.reddit.com/r/Cura/comments/1hp56ll/how_to_make_purge_line_on_the_front_of_bed/)

PartySausage:
> I just pushed a commit on this.

Loaded the new commit & sliced on my E3 S1 Pro. G-code looks good & seems to be functioning as expected in the PrusaSlicer G-code viewer with the Purge Lines on the Front Edge & Circle around to layer start enabled. I'll be doing a print tomorrow so can test it in the real world.

PartySausage:
> > I just pushed a commit on this.
> 
> Loaded the new commit & sliced on my E3 S1 Pro. G-code looks good & seems to be functioning as expected in the PrusaSlicer G-code viewer with the Purge Lines on the Front Edge & Circle around to layer start enabled. I'll be doing a print tomorrow so can test it in the real world.
I can't comment on the coding side & best practice but I've done several small prints with this commit on my simple setup & didn't experience any functional issues

GregValiant:
Looks good.
I would suggest a small revision to "moves" in line 331:
```
moves = [f";MESH:NONMESH---------[Circle around to {location_name}]    Start from: {str(start_side)} {str(start_depth)}    Go to: {target_side} {target_depth}\nG0 F600 Z2 ; Move up\n"]
```
and line 358 is deleted.

That moves the comment from the end of data[1]) to the beginning of the travel section.  Fits better I think.
```
;MESH:NONMESH---------[Circle around to Layer Start]    Start from: right rear    Go to: right rear
```
I'll add a commit with that change.

On another slightly different topic:
When I was playing around in 5.9.0 and trying different things with the Purge Lines script - I noticed that I cannot change the "Layer Start".  It appears to always be "right rear".
Prior to 5.9.0 it wasn't like that.  The "Layer Start" was always positioned according to the "Layer Start X" and "Layer Start Y" in the Travel Settings.  That no longer appears to be true in 5.9.0 and those two settings appear to not have any effect.  Not being able to adjust the layer start point can mean a lot of extra travel in models that are "C" shaped, or evolve into a "C" shape as the print progresses.
I'm thinking about opening a bug report in regards to that.  If you get a chance, take a quick look and let me know what you think.

HellAholic:
I have a ticket open for that in the backlog, should be in before the next release. That includes making the travel move an optional move so it can be active / inactive and should always listen to the Layer Start X and Layer Start Y regardless of all other elements (prime tower position).

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,881 @@
...
+                    "description": "The amount of filament to unload.  Bowden printers usually require a significant amount and direct drives not as much.",
+                    "type": "int",
+                    "default_value": 440,
+                    "unit": "mm  ",
+                    "enabled": "enable_unload"
+                },
+                "unload_quick_purge":
+                {
+                    "label": "    Quick purge before unload",
+                    "description": "When printing something fine that has a lot of retractions in a short space (like lettering or spires) right before the unload, the filament can get hung up in the hot end and unload can fail.  A quick purge will soften the end of the filament so it will retract correctly.  This 'quick puge' will take place at the last position of the nozzle.",
```
wawanbreton, line 163: I do see the need of doing this, however purging at the last position of the nozzle will probably be over the model and ruin it, right ?

wawanbreton, line 163: And you also miss an `R` in `'quick puRge'` :slightly_smiling_face:

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,881 @@
...
+                    "options": {
+                        "purge_full": "Full",
+                        "purge_half": "Half"},
+                    "default_value": "purge_full",
+                    "enabled": "add_purge_lines and is_rectangular"
+                },
+                "move_to_start":
+                {
+                    "label": "Circle around to layer start",
+                    "description": "Depending on where the 'Layer Start X' and 'Layer Start Y' are for the print, the opening travel move can pass across the print area and leave a string there.  This option will generate an orthogonal path that moves the nozzle around the edges of the build plate and then comes in to the Start Point.  The nozzle will drop and touch the build plate at each stop in order to nail down the string so it doesn't follow in a straight line.",
```
wawanbreton, line 120: Touching the build plate with the hot nozzle can be destructive in some configurations. Can you make the warning a bit more visible ? Maybe add a :warning: emoji ? Not sure this is possible...

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,881 @@
...
+from enum import Enum
+
+
+class Location(str, Enum):
+    LEFT = "left"
+    RIGHT = "right"
+    TOP = "top"
+    BOTTOM = "bottom"
+    REAR = "rear"
+    FRONT = "front"
```
wawanbreton, line 28: Why do you have top+rear and bottom+front ?

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,881 @@
...
+                purge_str += f"G1 F{self.print_speed} X{x_stop} Y{self.machine_back} E{purge_volume} ; First line\n"
+                purge_str += f"G0 X{x_stop} Y{self.machine_back - 3} ; Move over\n"
+                purge_str += f"G1 F{self.print_speed} X{self.machine_right - 10} Y{self.machine_back - 3} E{purge_volume * 2} ; Second line\n"
+                # Retract if enabled
+                purge_str += f"G1 F{int(self.retract_speed)} E{round(purge_volume * 2 - self.retract_dist, 5)} ; Retract\n" if self.retraction_enable else ""
+                purge_str += "G0 F600 Z8 ; Move Up\nG4 S1 ; Wait 1 second\n"
+                # Wipe
+                purge_str += f"G0 F{self.print_speed} X{self.machine_right - 20} Y{self.machine_back - 3} Z0.3 ; Slide over and down\n"
+                purge_str += f"G0 X{self.machine_right - 35} Y{self.machine_back - 3} ; Wipe\n"
+                self.end_purge_location = Position.RIGHT_REAR
```
wawanbreton, line 516: There is some code duplication here, the (Python) code is very similar for the 4 locations. Could you make a function that takes some argument ? Same for the other types below.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -0,0 +1,881 @@
...
+
+        data[1] = "\n".join(startup)
+        # Remove any double-spaced lines
+        data[1] = data[1].replace("\n\n", "\n")
+        return data
+
+    # Unloading a large amount of filament in a single command can trip the 'Overlong Extrusion' warning in some firmware.  Unloads longer than 150mm are split into individual 150mm segments.
+    def _unload_filament(self, data: str) -> str:
+        extrude_speed = 3000
+        quick_purge_speed = 240
```
wawanbreton, line 781: Would it make sense to use `retraction_prime_speed` instead, or maybe `switch_extruder_prime_speed` ? I'm afraid those fixed values will not work on some printers.

wawanbreton (commented):
Hi, this is a nice improvement ! I do have a printer that I always manually prime while it is starting, so I could totally use that instead :smile:

I have a few remarks, but this is almost approveable in its current state :+1: 

And for a future version, I would love to have settings for:
* Distance to the border
* Number of purge lines

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
GregValiant: The need for a final purge came about when there were a lot of retractions in a short span prior to the print ending.  The "shape" at the end of the filament could get hard and prevent the extruder from being able to back the filament out of the hot end.  In most cases that doesn't happen, but I had it happen a couple of times.  The "final purge" was the fix.

The insertion point is just before the M104 S0 line in the ending gcode.  If there is no M104 S0, then there is no insertion.  If it is there, and if there is no movement of the nozzle away from the print (for example a UM S5 doesn't have parking in the Ending Gcode), then yes, that final purge could be over the print.

A couple of points.
1. It's an option.
2. Most printers seem to have a "parking" move in their Ending Gcode so the purge would be "over there".
3. It is covered in the description.  (Of course people would actually need to read the description). 

I suppose there could be a "Present Print" code added but that would seen to be user preference and it could not be added if there was already one present in the Ending Gcode.  I think it would be better (and a lot simpler) to allow the user to just ignore it and leave it turned off.

PS:  There is a typo in the description.  "This 'quick puge' will"

GregValiant: That's the New Jersey spelling as nobody who lives there seems to pronounce their "R's" anyway.  Take the word "sure".  It is always pronounced "shu-a" in the northeast US.
I was just trying to be geographically and phonetically sensitive.
Change it if you must.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
GregValiant: Yes, it does appear that way.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
GregValiant: I checked a lot of definitions.  When they bother to note it, they have the "machine_max_feedrate_e" at 50mm/sec.  If it is higher; I don't care as 50 is sufficient.
If it is a direct drive, users probably don't need the Unload function.
If it is lower (in the printer M203 registry) then whatever that value is will limit it anyway.
"Switch extruder prime speed" would be OK for multi-extruder machines only.

I did have it set to 'machine_max_feedrate_e' but since not all printer definitions bother to set that it is the generic "speed of light" from fdmprinter.def.json.  That's a little high for my cheap-ass Ender.
At the end of the day I decided 50 was "close enough".

"quick_purge_speed " could be handled differently.  Something like:
quick_purge_speed = nozzle_size * 500
The "500" is not a magic number but is a scientifically calculated (on my slide rule) fudge factor based on:  the area of the nozzle inside diameter, the flow rate, and the conversion to mm/min.  It's close enough to ignore the other math and takes the nozzle size (of extruder[0]) into account.
I've made that change.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
HellAholic: It's mostly to separate the references for Nozzle Position and Purge location.
Top and bottom refer to the edges (the entire edge line)
"bottom": "On front edge (Ymin)",
"top": "On back edge (Ymax)"},
while rear & front are used for the nozzle position.
Could be merged into two but having them separated makes the check for purge location vs position easier in the code.
Beyond that, they are the same.

HellAholic:
@GregValiant Changed the top/bottom to Rear/front. Also found an extra R 🤷‍♂️

GregValiant:
Please check my last commit.  I was getting "conflict's" and I think I resolved them.

"Distance to Border"
I'll look at adding a setting for that.

"Number of purge lines".
It would always need to be positive and would involve adding an index distance.  Right now the script moves the nozzle 3mm so the lines don't interfere with each other and maybe pick up boogers.  An even number of lines insures that the nozzle will only end up in the "left front", or the "right rear".  Odd numbers would put them into corners that are not taken into account in "Circle around...".  If 2 purge lines provide 400mm of purge, and you need to add two more, is 800 + the skirt/brim going to be enough?

GregValiant:
I removed the default border of 1mm and added a setting for "Border Distance".
I got another "conflicted" message.  If you would be so kind as to see if I resolved it correctly it would be appreciated.

HellAholic:
To get rid of the conflict bit, before you make a commit, stash your changes and do a pull, then restore your changes and make the commit. 

I see one issue that I'll do a fix using github's editor.
After my commit, do a pull from the branch :P

HellAholic:
I have couple of changes on my local version, I'm not sure about 🤔 [Don't know/remember how made these 🤷‍♂️ ]
image
See if you remember anything about these. ☝️

HellAholic:
AAAAh, you added the Move to a corner bit instead of the compare side logic.
The bit at the top for the start_x, start_y and the warning was overwritten.
Should I restore that bit?

GregValiant:
I don't remember any of that.
Is it only the warning message that is affected?  As long as the code is correct and any tool offsets are being utilized it should be OK either way.
I seem to recall adding the blocker for when T0 has offsets.  Maybe it occurred as part of that.

HellAholic:
The output of the two is close enough, I'm looking into making the wipe moves for rectangular a bit consistent and turn it into a helper function.

Found a potential "issue"in line 565 that sets the speed to travel speed for the wipe move, this is the only place, so I don't think this was intended.

If the F{self.print_speed} remains consistent in all wipe moves then we can do it as:

`
def generate_wipe_gcode(x_1=None, y_1=None, x_2=None, y_2=None, z=0.3):
            first_line = f"G0 F{self.print_speed} X{x_1} Y{y_1} Z0.3 ; Slide over and down\n"
            second_line = f"G0 X{x_2} Y{y_2} ; Wipe\n"
            
            return first_line + second_line
`

HellAholic:
Ok, so adding a helper function would generate twice the amount of code since we need 2 coordinates for the first move and 2 for the second one, I could not make it pretty or more understandable than the current format. So instead I made the wipe moves for the rectangular build plates more consistent. I also fixed some indentation bits.

GregValiant:
I think I know how I deleted the 'start_x' 'start_y' lines.

I originally added the "border distance" below the start_x and start_y in that INIT section.  When I deleted it to go another way I think I inadvertently deleted those two lines as well.  It may have been a Senior Moment.  Sorry about that.
The script has so many convolutions now it's getting to take a long time to go through the various possibilities to debug.

The wipe function was really the only one I figured could be split off per Erwan's comment.  Looking at it I figured there wouldn't be much saving (if any).

This has turned out to be more work than the Cooling Fan script was.  It's only 80 lines shorter as well.

HellAholic:
Also when you have a typo in the commit comment about fixing a typo
I hate the autocorrect...
image

Give this one a look, let's see if Erwan is pleased with the changes tomorrow 🤞

GregValiant:
I was a checker for a long time in machine design studios.  I would give any 'new kids' projects that I had previously checked as references so they would know how I liked things laid out.  It saved a lot of rework down the road.
It's the same here.  I won't mention names but one reviewer didn't like spaces around math symbols.  I had to take them all out.  So the next project I left them all out and a different reviewer had me put them in.
Now I keep a box of spares next to the keyboard.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
wawanbreton: I would say either change it or add a small comment to make it more explicit what they designate

wawanbreton:
> "Number of purge lines".
> It would always need to be positive and would involve adding an index distance. Right now the script moves the nozzle 3mm so the lines don't interfere with each other and maybe pick up boogers. An even number of lines insures that the nozzle will only end up in the "left front", or the "right rear". Odd numbers would put them into corners that are not taken into account in "Circle around...". If 2 purge lines provide 400mm of purge, and you need to add two more, is 800 + the skirt/brim going to be enough?

Could also be "Number of lines pairs" so that the final number would always be even :slightly_smiling_face: 

So I have a strange way of working, I usually have a skirt but in order to remove it properly, I don't really want it to be used as a prime and I end up manually insert the filament. But maybe I would also do the same with the purge lines because they would not be easy to remove either :sweat_smile: 

> The wipe function was really the only one I figured could be split off per Erwan's comment. Looking at it I figured there wouldn't be much saving (if any).

Yes sorry, I know I am a bit fanatical about code duplication, I have been traumatized in the past by people copy-pasting the same piece of ~20 lines of code over and over, and sometimes they slightly changed it to fix something, but not in all places :slightly_smiling_face: but I also know how easy it is to make this mistake of changing a line of code, and forgetting (or not knowing) that there is a similar one somewhere else. The purpose is not really to make the file smaller, but to make sure that whoever will changes it in the future is less likely to create mistakes/inconsistencies.

> Give this one a look, let's see if Erwan is pleased with the changes tomorrow 🤞

I am indeed :smile: 
In this case that was just suggestions for improvements, the initial code looked good already.

wawanbreton (approved):

GregValiant:
My printer came with a light duty 25mm wide putty knife.  It's good at removing purge lines and skirts.

I ended up putting an override for "Skirt Height" into my printer definition file.  I never understood how one person (obviously a fingernail biter) got to have the default skirt height set to "3".

When I design these things (all things really) I go for what I call the "Solid Gold Edition".  That means all the bells and whistles I can think of.  It generally adds flexibility but at the cost of making things complicated.
I had thought that "once out and once back" of purge would be sufficient and in many cases a lot more than sufficient.  That's why there is a setting to cut that in half.
I mentioned that debugging for the "12 purge options" is both tedious and mind numbingly boring.  Unlike doubling the purge lines - adding a prime blob would (off the top-of-my-head) be easier as it wouldn't effect where the print head ends up.
"blob/retract/move to purge start/unretract/purge using existing code".

That would have the advantage that users would not damage their fingernail polish when removing the blob as seems to be the case when removing purge lines and skirts.  Just sayin'.

GregValiant:
Three cups of coffee later...
image

```
    def _get_blob_code(self) -> str:
        if not self.prime_blob_enable or self.prime_blob_distance == 0:
            return ""
        # Set extruder speed for 1.75 filament
        speed_blob = round(float(self.nozzle_size) * 500)
        # Adjust speed if 2.85 filament
        if self.material_diameter > 2: speed_blob *= .4
        blob_string = "G0 F1200 Z20 ; Move up\n"
        blob_string += f"G1 F{speed_blob} E{self.prime_blob_distance} ; Blob\n"
        blob_string += f"G1 F{self.retract_speed} E-{self.retract_dist} ; Retract\n"
        blob_string += "G92 E0 ; Reset extruder\n"
        blob_string += "M300 P500 ; Beep\n"
        blob_string += "G4 S2 ; Wait\n"
        return blob_string
```
results in:
```
;TYPE:CUSTOM----------[Purge Lines at MinX]
G0 F600 Z2                      ; Move up
G92 E0                             ; Reset extruder
G0 F1200 Z20                  ; Move up
G1 F200 E35                     ; Blob
G1 F2100 E-7                    ; Retract
G92 E0                              ; Reset extruder
M300 P500 S600              ; Beep
G4 S2                                ; Wait
G0 F10500 X0 Y10            ; Move to start
blah, blah, blah...
```

It's just like adding more purge lines only different.

@HellAholic should we do this?

wawanbreton:
> My printer came with a light duty 25mm wide putty knife. It's good at removing purge lines and skirts

I have a build plate that is quite soft, so I tried that once and made a scratch in it :sweat_smile: Adhesion is very good though, and also removal is very easy, so I do like it, but it is just annoying for prime lines :slightly_smiling_face: 

> I had thought that "once out and once back" of purge would be sufficient and in many cases a lot more than sufficient. That's why there is a setting to cut that in half.

Maybe that is just me doing weird things, indeed ! I could also try the prime blob... We do have this feature in the engine btw, but I think it is only for printers that have a firmware implementation of it. In this case we should just make it work on all printers instead.

GregValiant:
Gotta be getting close here.
The code allows the purge blob when purge location is either left or front.
Adjustment to the quick_purge_speed to account for 2.85 filament.

plugins/PostProcessingPlugin/scripts/PurgeLinesAndUnload.py
```diff
@@ -562,7 +585,7 @@ def calculate_purge_volume(line_width, purge_length, volume_per_mm):
                 purge_str += "G0 F600 Z8 ; Move Up\nG4 S1 ; Wait for 1 second\n"
                 # Wipe
                 purge_str += f"G0 F{self.print_speed} X{self.machine_right - 3 - self.border_distance} Y{self.machine_back - 20} Z0.3 ; Slide over and down\n"
-                purge_str += f"G0 X{self.machine_right - 3 - self.border_distance} Y{self.machine_back - 35} ; Wipe\n"
+                purge_str += f"G0 F{self.speed_travel} X{self.machine_right - 3 - self.border_distance} Y{self.machine_back - 35} ; Wipe\n"
```
HellAholic, line 588: This one seems suspicious, I removed it since I don't think it should be the travel speed based on the consistency.
Also for the following two wipe moves since the other 5 wipes didn't have this F line and I do not think it is necessary unless it's a different F value.
Could be a code mixup but just want to make sure before taking it out :P

GregValiant:
Anything would be better than blobbing into the middle of a print.
I tried an easy fix but it was too difficult.  Easier to add more settings and let the user decide where the WC should be.

When you enable the prime blob "Blob Location X" and "Blob Location Y" will now be available.
Careful though, if you put the blob in the exact location of the purge line start - the nozzle was sit down into the blob.  Not good.  That's a user thing though.  Blob on the right and purge on the left or blob in the back and purge on the front will work.

I'll do the commit here in a minute.

GregValiant:
This last commit (after three tries) adds two settings so the user can dictate where the prime blob will occur.

PartySausage:
> This last commit (after three tries) adds two settings so the user can dictate where the prime blob will occur. @PartySausage see if this fix works for your situation.

@GregValiant  That seems to do the trick in the simulator, Well done for the quick turnaround. 

I'd done an amateur fix by moving the call to the blob procedure to after the start move in the sections that generate the purge code for the Front & Left Purge lines, which seemed to work, but like you stated this also runs the risk of the purge line starting in the prime blob.

When using purge lines on the front of the plate the purge start is X10 Y0 so I thought I'd try with the purge blob at X0 Y0 & see how/if that works

GregValiant:
I looked at simply moving the insertion point but right away I saw the potential interference with the purge lines.  The user still has to be careful, but a working combination of settings should quickly become apparent.

GregValiant:
Just thinking about this...
Should there be a setting for a user to input "Minimum Z Height"?
Having a fancy emoji is all well and good, but doesn't really help.

I've written it in and it's sitting here.

PR #20023, opened 2024-12-11
Voron volumetric speed

NerdyGriffin:
# Description

This updates the Voron base definition files and material profiles to make better use of the latest features and defaults inherited from `fdmprinter.def.json`. The goal is to simplify the printer definition, reduce the number of redundant lines of code, and replace hard-coded values with parametric definitions so that the base printer definition can more effectively adapt to custom Voron printer configurations. I have also designed the behavior of these printer definitions to be more consistent with the community-made Voron definitions found in other open-source slicers, to keep these definitions up-to-date with the changes and improvements that have been made to both the Voron printer hardware and Voron support/compatibility seen on other slicers.

I have chosen the parametric print speeds so that they result in the exact same print speeds as the previous hard-coded values. This ensures that print speeds continue to be calculated appropriately for other combinations of nozzle sizes and layer heights that are not yet standardized in this printer definition, while keeping this new printer definition fully backwards-compatible with existing Voron users' custom print/material profiles.

## Type of change

# How Has This Been Tested?

- [x] I manually created various print profiles (e.g. each combination of nozzle size and layer height) and reviewed the results of the parametric print speeds to ensure they produce the same results as the previously hard-coded `print_speed` values for all the supported material types.
- [x] For each of the print settings I have modified (e.g. `cool_min_speed`, `travel_speed`, etc.), I manually checked the values displayed in the slicer when using different combinations of printers, nozzle sizes, layer heights, and material types to ensure that they were calculated to the results I intended.

**Test Configuration**:

# Checklist:


resources/definitions/voron2_base.def.json
```diff
@@ -21,36 +21,37 @@
...
-        "alternate_extra_perimeter": { "default_value": true },
-        "bridge_fan_speed": { "default_value": 100 },
+        "acceleration_enabled": { "default_value": true },
+        "acceleration_layer_0": { "value": "math.ceil(acceleration_print / 10)" },
+        "acceleration_print": { "default_value": 5000 },
+        "acceleration_roofing": { "value": "math.ceil(acceleration_topbottom * 3000 / 5000) " },
+        "acceleration_support": { "value": "math.ceil(acceleration_print / 2)" },
+        "acceleration_travel": { "value": "acceleration_print if magic_spiralize else min(math.ceil(acceleration_print * 7000 / 5000), round((machine_max_acceleration_x + machine_max_acceleration_y) / 2, -2))" },
+        "acceleration_wall_0": { "value": "math.ceil(acceleration_wall * 3000 / 5000)" },
+        "adhesion_type":
```
magnetoxgarage, line 31: Maybe I'm doing something wrong, but it appears that there is an issue with `adhesion_type` not recognizing the values when written this way. Seen in the cura log as `NameError: name 'skirt' is not defined`. Tested on Cura 5.9 and 5.10-alpha in Windows.
I was able to fix it and adhesion works properly by deleting default_value and adding quotes around each string value:
`"adhesion_type": { "value": "'brim' if draft_shield_enabled else 'skirt'" },`

Example of other printers doing it this way: https://github.com/Ultimaker/Cura/blob/ac0fd927dbbff570b3c264dc70ad525170a754a4/resources/definitions/lotmaxx_sc10.def.json#L18
https://github.com/Ultimaker/Cura/blob/ac0fd927dbbff570b3c264dc70ad525170a754a4/resources/definitions/lotmaxx_sc20.def.json#L18

If this is an error on my parts somewhere, I really apologize for wasting your time

WCEngineer: You are correct about the single quotes, I have fixed that error before, so I am surprised it is missing from this branch. Thank you for pointing it out

magnetoxgarage: Ah, after some more testing, I was able to verify that `default_value` does work correctly. It seems the only issue was the quotes. Logs are clear now. Sorry for that part.
You're welcome, glad it helped! Your pull request has helped me as a reference for better default settings for Cura profile for the Peopoly Magneto X printer. Thank you for your work!

PR #20039, opened 2024-12-16
CURA-12352 Different minimum layer time for layers that contain overhangs

wawanbreton:
CURA-12352

:warning:  Depends on CURA-11966, so don't review until merged

Comes with https://github.com/Ultimaker/CuraEngine/pull/2187


resources/definitions/fdmprinter.def.json
```diff
@@ -8401,15 +8426,13 @@
                     "value": "support_angle",
                     "settable_per_mesh": true
                 },
-                "wall_overhang_speed_factor":
+                "wall_overhang_speed_factors":
```
rburema, line 8429: We can't really change settings like this without doing an upgrader.

People will have values like '2.5' for a setting that no longer exists, that _should_ be altered to `[3]` for this 'new' setting. (In other words: Older files should still work as expected.)

wawanbreton: The usual process (as far as I understood) is to note the required setting update in the dedicated ticket (in this case, CURA-12359) and write the whole upgrade script at once. But as you implemented it, it's also good :)

wawanbreton:
I do approve this PR, but I can't because I created it :smile:

HellAholic (approved):
Seems legit :P

PR #20041, opened 2024-12-16
CURA-12236 expose preferred travel area in combing

wawanbreton:
Add a setting which was previously hard-coded, and allows to fine-tune the distance to walls when doing an inside combing travel move.

CURA-12236

Comes with https://github.com/Ultimaker/CuraEngine/pull/2188


resources/definitions/fdmprinter.def.json
```diff
@@ -4412,6 +4412,21 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "retraction_combing_avoid_distance":
+                {
+                    "label": "Distance to Walls",
```
nallath, line 4417: Not a huge fan of this name. I'd probably call it something like "Distance to walls when combing", as "Distance to walls" is a bit too generic and can be confusing. 

Perhaps we could also use "Combing avoid distance" or something like that?

wawanbreton: I totally agree with changing it. We already have `Travel Avoid Distance` though, which is the avoidance distance around the models. Maybe `Inside Travel Avoid Distance` ?

nallath (approved):

PR #20124, opened 2025-01-13
[PP-537] Introduce nominal_layer_height

Frederic98:
# PP-537 - CURA-12020: Introduce `nominal_layer_height`

## Description
This PR introduces the `nominal_layer_height` setting. This is the layer height that the final part should have. The real `layer_height` and `layer_height_0` become children of this setting.

## Rationale
For the Factor 4, we multiply the layer height with the material shrinkage factor. This is done so that an object of 20mm high printed with a 0.2mm layer height, will still have exactly 100 layers. This improves the dimensional accuracy, because otherwise the object might not have an exact whole number of layers.

## Type of change

- [ ] Translations

# How Has This Been Tested?
Ran Cura from source with these changes

**Test Configuration**:
* Operating System: Ubuntu

# Checklist:

### DIFF
--- a/cura/Machines/Models/MachineModelUtils.py
+++ b/cura/Machines/Models/MachineModelUtils.py
@@ -15,7 +15,7 @@ def fetchLayerHeight(quality_group: "QualityGroup") -> float:
     from cura.CuraApplication import CuraApplication
     global_stack = CuraApplication.getInstance().getMachineManager().activeMachine
 
-    default_layer_height = global_stack.definition.getProperty("layer_height", "value")
+    default_layer_height = global_stack.definition.getProperty("nominal_layer_height", "value")
 
     # Get layer_height from the quality profile for the GlobalStack
     if quality_group.node_for_global is None:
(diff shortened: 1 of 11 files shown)


rburema (changes requested):
For all other printers, the default value is taken for nominal layer height.

How quality profile names appear in 5.9.x:
profile_names_ce3p_in_59

What it looks like with this ticket:
profile_names_ce3p_in_ticket

The 'actual' values for that printer:
values_for_layer_height_settings

One way this could _maybe_ be fixed is if we rename it to nominal layer height for each printer, and then have an upgrader that takes old saved files, checks if the layer height and initial layer height are the same and then sets them to the initial layer height. -- But that's a bit hacky and unreliable.

rburema:
@Frederic98 We think it's a better idea to just round the value for display purposes (at least on the top) than to introduce a new setting; since it's for shrinkage anyway, it's likely to fall within rounding errors.

Frederic98:
Hmm yeah for most cases that's probably enough. It won't work as expected for e.g. Ultrafuse 316L, which has a shrinkage factor of 124%, but it's a good enough fix for now.

rburema:
@Frederic98 I've implemented the rounding [here](https://github.com/Ultimaker/Cura/pull/20210), so I'm going to close this for now.

PR #20129, opened 2025-01-14
Add intent profiles for Method High Speed

alanbjorklund:
These intent modes were tuned by Eric MacNeil.  The jerk & accel values will be converted into a accel_config by the Makerbot writer plugin, and the Method firmware will interpret these values for each path.

The self support speed settings were unchanged from the previous defaults, but the model speed was increased for ABSR, PLA & TPLA.

PP-544

New intent modes include:

ABSR High Speed (MX, MXL)
ABSR High Speed Solid (MX, MXL)
PLA High Speed (M, MX, MXL)
PLA High Speed Solid (M, MX, MXL)
TPLA High Speed (M, MX, MXL)
TPLA High Speed Solid (M, MX, MXL)

These changes were tested extensively in the New York printer lab, as well as by Alan in New Jersey.  The most commonly observed failure mode was self support under-extrusion. This failure mode was eliminated by reverting the self support settings to the previous defaults.

### DIFF
--- a/resources/definitions/ultimaker_method_base.def.json
+++ b/resources/definitions/ultimaker_method_base.def.json
@@ -33,12 +33,14 @@
     {
         "acceleration_enabled":
         {
-            "enabled": false,
+            "enabled": true,
             "value": true
         },
         "acceleration_infill":
         {
-            "enabled": false,
+            "maximum_value": 3500,
+            "minimum_value": 200,
+            "minimum_value_warning": 750,
             "value": "acceleration_print"
         },
         "acceleration_layer_0":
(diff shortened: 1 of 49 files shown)


rburema (approved):
Looks good to me, as long as you're sure that the alterations made to the `..._base_...` file really are OK for the entire Method-series (also when not using the high-speed profiles).

Another(?) really small nitpick is that in the interface, some profiles now shows up as 'Fast - Highspeedsolid' (funnily enough it would be totally valid if written in Dutch, since we have compound words like that). I'm OK with keeping it like this at least for now, but I thought you might like to know.

alanbjorklund:
The settings in the base profile are the current defaults assigned via the firmware.  The current defaults get overridden only when the high speed / high speed solid intents are selected.  The lack of a space in the names "highspeed" & "highspeedsolid" is something we noticed, but there was no obvious solution so we didn't consider this a blocker for release.

PR #20155, opened 2025-01-21
Initial add of a definition, extruder, quality and variant resources for BIQU B2

bjuraga:
# Description

Initial add of a definition, extruder, quality and variant resources for BIQU B2

## Type of change

# How Has This Been Tested?
- [ ] Tested on my BIQU B2 Printer, not yet ready

### DIFF
--- a/resources/definitions/biqu_b2.def.json
+++ b/resources/definitions/biqu_b2.def.json
@@ -1,0 +1,32 @@
+{
+    "version": 2,
+    "name": "Biqu B2",
+    "inherits": "biqu_b1",
+    "metadata":
+    {
+        "visible": true,
+        "author": "Boris Juraga",
+        "machine_extruder_trains": {
+            "0": "biqu_base_extruder_0",
+            "1": "biqu_b2_extruder_1"
+        },
+        "quality_definition": "biqu_b2"
+    },
+    "overrides":
+    {
+        "gantry_height": { "value": 27.5 },
+        "machine_extruder_count": { "default_value": 2 },
+        "machine_head_with_fans_polygon":
+        {
+            "default_value": [
+                [-33, 35],
+                [-33, -23],
+                [33, -23],
+                [33, 35]
+            ]
+        },
+        "machine_end_gcode": { "default_value": "M104 S0\nM140 S0\n;Retract the filament\nG91\nG1 E-30 F300\nG1 Z5\nG90\nG28 X0 Y0" },
+        "machine_name": { "default_value": "BIQU B2" },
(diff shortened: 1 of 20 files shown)


bjuraga:
For anyone that can help guide me, my main issue is that it seems when the switch to the second extruder happens, the nozzle temperature goes to zero. It is not noticeable on the first few layers that i print for the first extruder, although the second extruder is used to print some parts of the layer, mainly parts of the skirt. Once the second extruder needs to print parts of the tower and then on the actual model, within seconds of printing the tower i get warnings that cold extrusion is prevented. 

Here's the sliced model:
afbeelding
And this is the position where those errors start to spam (LAYER 6):
afbeelding
The gcode, but i am not able to read it (yet):
[BB2_UCP_flip.gcode.txt](https://github.com/user-attachments/files/18487535/BB2_UCP_flip.gcode.txt)

GregValiant:
I see those lines in your gcode.  It acted like a normal dual extruder printer.  It seems like "extruders share heater and nozzle" were ignored.

I downloaded enough files to install the printer with a 0.4 nozzle using PLA.
The settings look good.  I set T1 to be the support extruder and the gcode I generated is fine.  The temperature stays the same until the ending gcode where it properly shuts off.

If the printing temperature of each nozzle is different, an M109 gets inserted for the hotter nozzle.  That's a Cura thing and may be a bug as I had "Wait for Nozzle Temperature" un-ticked.  I wouldn't want the M109 lines in there because there is no movement away from the print associated with them.  That could leave a divot or a blob.

bjuraga:
I learned about shared nozzle and heater options and had them set in a second commit, after the initial file i posted. It did reduce the temp control commands, so i am going in a good direction. 

I am porting over commands from the official cura profile by BIQU for the B2, but i see a lot of settings that i am still to decide if they should be set. 

I did run few tests, but i scaled the model to 1cm x 1cm. I do see a color change on almost every layer, and that makes no sense to me since i want 6 layers in one color and then on top another 4 layers with the second color. What ends up happening is, the printer changes the filament (color) and prints half the skirt in the next color, again initiates a color change and then goes to the model, but is in reality still extruding the wrong color (filament). 


here is the gcode for attempt 7:
[777.gcode.txt](https://github.com/user-attachments/files/18570449/777.gcode.txt)

This is using the files from as in my last commit

bjuraga:
2 things left:
1. this profile (B2) is slower than the original i am extending (B1), need to figure out why
2. if i use this profile, and i print a single model, but i want to print it with the second filament/extruder, the printer still defaults to the first. This is seen in the Gcode few lines before `;START` there is a `T0` and later there is no `T1` command. I deed to figure out why Cura does not generate the T1 command at the start of a print

HeySteven:
Hi Bjuraga,
I just stumbled upon your github page.

I have been trying to get my BIQU B2 (B1 with dual extruder upgrade) working correctly.
I keep getting closer but have yet to successfully complete a 2 color print.

The closest I've come is using the firmware compiled by Jusssoz.
https://github.com/Juusoz/Biqu-B1-Dual-extrusion-BLTouch-firmware
last updated around 4 years ago.

You might want to check it out to see what he has done to resolve some of the issues you're hitting.

One issue with Juusoz's build is that he built it using 
Marlin 2.9.2 (Nov 7, 2021) which with my BIQU printer 7 and v1.4 board has a number of minor issues such as canceling a print job doesn't stop it properly (printer keeps printing) or the Marlin screen totally locking up and requiring you to reset/reboot the machine to regain control.

I didn't have these issues using
Marlin bugfix-2.0.9.3.x (Mar 27 2022) 
https://github.com/bigtreetech/SKR-3/tree/master/Firmware/Marlin-bugfix-2.0.9.3.x
(sorry forget where I downloaded the compiled firmware using this)
but that firmware wasn't properly compiled to support the 2nd extruder.

Hope this info helps.

HeySteven:
I was able today to do a somewhat successful print with Juusoz's build – the 2nd extruder is doing a crap job compared to the first but it’s working.
Possibly I could improve things by tweaking the settings on the Ultimaker Cura 5.9 slicer (for example by increasing the Flow, but then again maybe not.

Does your build include support for the BL Touch?

 

*	For example, if i don't want to change filaments, but 
want to print with the second extruder, in case supports get generated, 
they will use the first extruder. So we need to choose the support 
extruder, and not forget.

That’s what I’ve observed – whatever extruder I used for generating supports is the default extruder for the next print job if I’m print non-extruder specific gcode (for example code generated for the B1).
If I reboot the printer the 1st extruder once again becomes the default.

 

*	The quality and speed is getting closer to the B1 original configuration as 
well.

If I use the 1st extruder the quality matches the B1 original configuration (or at least comes close – I haven’t done a thorough test comparing different filaments, etc.)

2nd extruder is doing a crappy job, possibly I can compensate by tweaking the slicer. I could be something a simple as needing to increase the extrusion rate (Flow).

 

Another issue I’ve had to learn how to handle is – if the filament for the 2nd extruder is too low in the chamber it’ll block the filament from the 1st from loading and nothing will extrude.

So I’ve learned to partially retrack (manually) the 2nd extruder filament at the end of a print job to prevent issues.  
Not sure if the issue happens the other way around (2nd extruder filament blocking the 1st) because I only just got the 2nd extruder working correctly.

 

I spent a long time trying to get the firmware issues ironed out too.  I got frustrated to the point where I actually just shelved the printer for over a year and walked away from it until my son talked me into trying to get going again.

BIQU website, instructions, etc. are lame.  It should not be this painful to get what you paid for to work properly.

bjuraga:
It seems like the B2 is not officially supported by BIQU, but was only sold
viabtheir store. I bought it as soon as i was available, but had no time to
assemble it for years.

The second extruder did the same for me on so many firmwares, but at,the
end was caused by me not changing the jumpers before i added the new servo
board. If you get some 'No Printer attached' error or something about
'Z0.000' error, it will be fixed by the jumpers plus firmware.

On mine i see no difference now between the two extruders, so you should
expect no difference as well on yours. Without special Cura settings that
is.

I don't  have the BL touch so i wouldn't know. I read that there are
firmwares that support it but need some settings changed, can't help there.

In my profile i made sure that i set the extruder after the startup script
again since Cura sets it weirdly before the startup script, not after. I
plan to move extruder 'warm up/reset' code to each extruder's startup
script. I amnlooking for conditional code execution in startup scripts. I
think i read something in some release notes for Cura, need to track it
down. So i don't need to reset the printer.

The biqu startup end the end codes make sure to reset the filament to -40mm
for every extruder, so i have not needed to clean the nozzle for over a
month.

One last hurdle is that it seems like it takes super long before the
filament is properly switched. Way more than i think the 135mm or so set up
by the default profile. If i manually do the swap like this:
- active extruder 1 extruding fine
- retract 40mm (unload)
- switch extruders to #2 (was already loaded then unloaded 40mm)
- load 50mm, i see some filament #1  purely extruded
- set the extrusion to 100mm at a time, extrude 100m (one press on the UI)
- imat the end it is mostly filament #2 but still impure. (i use black on
#1 and orange on #2)
- i extrude 40mm ( 4 times 10mm step)
- still not 100% #2 filament
- i continue in 10mm steps and i get to between 200mm and 240mm needed
before i get only filament #2.
This seems like a physycal limitation that i just need to accept.

Other things:
i swapped the nozzle from to the original that came with B1. Not sure that
changed anything, so i might return the new one. I just need to re level
the printer and i have it so well now that i don't want to mess with it.
The original one behavior was the same as the new one, and i think the fix
was purely FW and jumpers.

bjuraga:
All my open issues listed above are mow either fixed or not in scope

HellAholic (approved):

PR #20162, opened 2025-01-21
PC-ABS & PC-ABS-FR 175 Updates

alanbjorklund:
Added PC-ABS & PC-ABS-FR intent and quality files
for Method X & Method XL.  These materials pair with
1C, 1XA and Labs extruders.

Exclude PC-ABS and PC-ABS-FR from sketch printers
and Method.

PP-550


alanbjorklund:
I need to update the self support settings again, I will create a new PR this week.

Issue #20168, opened 2025-01-22
Initial Printing Temperature Not Working


MisterLuMox:
### Cura Version

5.9.0

### Operating System

windows 11

### Printer

cr-10 v2

### Reproduction steps

Initial print temperature is not working most of the time. Sometimes when I reboot it works, but usually it does not and I have to manually edit the G code to include it (it's really important!!). 

### Actual results

The temperature always defaults to the Printing Temperature

### Expected results

The first layer print temperature should be at the higher initial print temperature setting.  M104/M109 searches only show 240

[CCR10_lower A arm.zip](https://github.com/user-attachments/files/18510605/CCR10_lower.A.arm.zip)

### Add your .zip and screenshots here ⬇️

[Screenshot 2025-01-22 111337.zip](https://github.com/user-attachments/files/18510609/Screenshot.2025-01-22.111337.zip)

GregValiant:
Thanks for the report.
It sounds like the setting you want is "Printing Temperature Initial Layer".

"Initial Print Temperature" is the _"OK you can start now"_ temperature and "Final Print Temperature" is the "Cool down to this before the print ends" temperature.  (My experience is that both of those only work for multi-extruder printers.)

Try just using the "Printing Temperature Initial Layer" setting.  The print will start at that temperature and then just before the start of the second layer you will see the "M104 S" line with the main "Print Temperature" setting.

Let me know if that works for you.

MisterLuMox:
Switched to Printing Temperature Initial Layer and that fixed it. The language isn't that clear with 'Initial print temperature'. Adding 'before print starts' might make that more clear. 

Thanks for the speedy response.

GregValiant:
Yeah, it isn't good.  I'll see if they are willing to change the tooltips for those.
I've requested a change to the tooltip.
Part of the request is to hide the settings for single extruder machines.  Everything I did pointed to the settings only being used by multi-extruder printers but that may not be true.  See #20172 for the suggested wording.

Issue #19190, opened 2024-06-05
Parameter "Initial Printing Temperature"has no effect


XXLRay:
### Cura Version

5.7.0

### Operating System

Windows 10 Pro

### Printer

Snapmaker original

### Reproduction steps

1. I set Printing temperature to 260
2. I set Initial Printing Temperature to 240
3. I slice

### Actual results

 Setting Initial Printing Temperature has no effect. Cura creates the following GCODE.

...
;Generated with Cura_SteamEngine 5.7.0
M140 S65
M105
M190 S65
M104 S260
M105
M109 S260
...

### Expected results

I would rather expect the following (M104 and M109 set to 240 followed bei M104 set to 260).

...
;Generated with Cura_SteamEngine 5.7.0 but manually adjusted
M140 S65
M105
M190 S65
M104 S240
M105
M109 S240
M104 S260
...

This is an issue because my printer (Snapmaker original) doesn't keep a temperature of 260 very constant. In the beginning it waits for a constant temperature but because this is not the case it doesn't start printing. The results are fine if I alter the GCODE namually accordingly

### Add your .zip and screenshots here ⬇️

[Bohrschablone Kantenschutz u+or.zip](https://github.com/user-attachments/files/15593256/Bohrschablone.Kantenschutz.u%2Bor.zip)

GregValiant:
Thanks for the report.
Both "Initial Printing Temperature" and "Final Printing Temperature" are used by multi-extruder printers.  You are correct that they have no effect on your single extruder machine.

Cura adds that "prepend gcode" with the temperatures whenever there are no temperature callouts in the StartUp gcode.  
If within your StartUp you were to have something like...

M140 S{material_bed_temperature_layer_0}             ;Start to heat the bed
M109 S{material_print_temperature_layer_0 - 20}    ; Heat the hot end to 20° below the Initial Layer Print Temperature
M190 S{material_bed_temperature_layer_0}            ; Wait for the bed
M104 S{material_print_temperature_layer_0}           ; Heat the hot end the rest of the way but don't wait for it.

Those would show up in your gcode as:
M140 S65
M109 S240
M190 S65
M104 S260

If you are having trouble maintaining a fixed 260 you might need to run Auto-Tune (if your firmware allows it).  You can adjust the P.I.D. settings of the hot end.  A properly tuned hot end doesn't bounce around.

I think both of those settings should be hidden for single extruder machines to avoid this confusion.
"material_initial_print_temperature":                
[...]

Issue #14019, opened 2022-12-11
Initial/final layer print temperature doesn't match desired print temperature 


Reynolds-Beta:
### Application Version

5.2.1

### Platform

Windows 11

### Printer

Ultimaker S5

### Reproduction steps

1. Add a new material. In this case I duplicated generic PLA to create a MatterHackers MHBuild PLA material which (from experience) operates better from a bit lower temp than the generic settings
2. Set Default Printing Temperature in the material to 185C - this is the lower range of MHBuild PLA recommended temps.
3. Set Default Build Plate Temperature in the material to 40C. Again, as recommended
4. Active the new material profile, choosing to discard any settings from previous settings
5. Choose the Default Fine (0.1) profile - you can actually choose almost any "profile" and get a bunch of different temperatures _other_ than what you set.

See the material printing temperatures highlighted.

### Actual results

Printing and build-plate temperatures do not reflect what was entered - or at least there isn't an explanation as to why they are different and how they are controlled. See picture below.

Some of these values are now below the range of recommended settings from the manufacturer (see Initial/Final Layer Temperature) - in this case 180 is the low point of the range.

![image](https://user-images.githubusercontent.com/44441515/206917794-f3bed75d-8ccf-4091-8e57-6810c92689b4.png)

### Expected results

Entered values are reflected in settings. If such is the case, then the material settings (when making the material profile) should have the additional fields to set the calculated fields.

**and**

Explanations on why they are different to expected values.

### Checklist of files to include

- [ ] Log file
- [ ] Project file

### Additional information & file uploads

Choosing other profiles such as Default Extra Fast change the values to even different than Default Fine, again without explanation or documentation. See below.

![image](https://user-images.githubusercontent.com/44441515/206918103-466e7cce-4a09-4677-bb08-39b7380fe85e.png)

Reynolds-Beta:
Appreciate the response, and I understand the situation - it's complex - isn't it always?

In this particular case - I'm not using a custom "profile", just the canned ones for the Ultimaker S5 (Visual, Engineering, Draft, etc.). This is always my starting point, after which if I need to tweak a setting, I'll do so _then_ save it as a custom profile. 

Let's break it down by the bugs in just this case:
- Ignoring initial/final layer temp, the **Printing Temperature** setting is 195, not 185 that I set
- Build plate is set to 60, not the 40 that I set
- Irrespective of whether they are included in gcode or not, the initial/final layer temp are incorrect,

[...]

PR #20172, opened 2025-01-23
Update 'fdmprinter.def.json'

GregValiant:
This will be a draft.  The changes to the tooltips should be OK, but the others I'm not 100% sure on.

Alter two settings:
'material_initial_print_temperature' and 'material_final_print_temperature'
Change the tooltips to better describe the settings (the current wording is not clear), and change the "enabled" logic to include " and machine_extruder_count > 1".  (That will need to be checked but it doesn't appear that the settings are used for single extruder prints.)
Tooltip changes will of course require translation changes.
Changed the "type" from float to int as most machines aren't really set up for 1/100 of a degree.
Changed the minimum_value from "-273.15" to "0" because it was just silly.
Changed the maximum_value 
Changed the default value from 200 to material_print_temperature 

This is in response to a couple of bug reports including #20168 and #19190 and maybe #14019 

## Type of change

- [ X] Improvement / clarification

# How Has This Been Tested?

I have had both settings hidden for years.

**Test Configuration**:
* Operating System:
Win10Pro and Cura versions since 4.8 (although the settings were disabled using definition file overrides).

# Checklist:

- [X ] My code follows the style guidelines of this project as described in [UltiMaker Meta](https://github.com/Ultimaker/Meta) and [Cura QML best practices](https://github.com/Ultimaker/Cura/wiki/QML-Best-Practices)
- [ not applicable ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -515,16 +515,6 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
-                "machine_start_gcode_first":
-                {
-                    "label": "Start GCode must be first",
-                    "description": "This setting controls if the start-gcode is forced to always be the first g-code. Without this option other g-code, such as a T0 can be inserted before the start g-code.",
-                    "type": "bool",
-                    "default_value": false,
-                    "settable_per_mesh": false,
-                    "settable_per_extruder": false,
-                    "settable_per_meshgroup": false
-                },
                 "extruder_prime_pos_z":
                 {
                     "label": "Extruder Prime Z Position",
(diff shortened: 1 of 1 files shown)


HellAholic:
Tooltip change and number of extruders check is good. The setting is for multi-extruder printers mainly for when a switch happens.

I would not change the other settings as there could be side effects.
So I've heard so many stories about people printing all kinds of stuff on FDM printers, ranging from metal to chocolate to ceramics to ice-cream :) [Why are you against ice-cream printers :)] So while the range for average user might not be useful it allows for creativity and experimentation. Same with the float vs int.
That's just my personal 2 cents about this.

GregValiant:
I reverted everything except for the tooltips and the "enabled" logic.

HellAholic:
I think -190 is the sweet spot xD with liquid nitrogen but I think the original implementation was like: "If you wanna do something wrong, do it right" and went for the easy sure way. (these are just speculations, any similarities between this and any character, real or fictional is completely accidental).

GregValiant:
This last commit is in response to a bug report ( #20466 ) that was caused because there was no "minimum value" called out for "Minimum Even Wall Line Width" and "Minimum Odd Wall Line Width".  I added:
"minimum_value": "0.4 * machine_nozzle_size" to both settings.

"Z-seam position" dropdown:  Re-ordered the settings so they go clockwise from the left front. 
Added a "Middle" setting and adjusted the Zseam X and Y boxes to show the bed midpoint when "Middle" is selected.

GregValiant:
Commit 6 is a bug fix for #20499 .  Changing the "coasting_volume" {"minimum_value"} for "0.0" to "0.001" fixed the problem although the real problem is likely somewhere else in the code.

GregValiant:
I'm going to close this. Some ideas been addressed in other PR's and this isn't a good way to deal with "fdmprinter.def.json" as it affects too many things.

PR #20193, opened 2025-01-27
CURA-12150 Add local preference to monitor use of USB writing

wawanbreton:
This is the first step of smartly enabling the USB printing: we add an internal preference, that is `False` by default, and will be set to `True` whenever the user actually prints something over USB. This will remain as is for Cura 5.10, in order to "log" users that are actually using it. In a future version, we will use this preference to enable USB printing or not. This way, all the users that have been actively using it on Cura 5.10 should have it still enabled without any action. Others will be prompted with a specific UI when adding a printer. This is not perfect, but should allow us to catch a large majority of users of USB printing.

CURA-12150


rburema (approved):
lgtm

PR #20194, opened 2025-01-27
CURA-12224 handle loooong setting names

wawanbreton:
This PR has 2 changes:
* Give some less space to the settings input fields, so that we have more for the text. It also makes the values closer to their units, and in the vast majority of cases the input fields are mostly empty
* Make the field higher when the text doesn't fit in, so that it properly resizes with multi-line names

image

### DIFF
--- a/resources/qml/Settings/SettingItem.qml
+++ b/resources/qml/Settings/SettingItem.qml
@@ -13,7 +13,7 @@ import "."
 Item
 {
     id: base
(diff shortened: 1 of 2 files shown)


rburema (approved):
The only downside is this, but we can't guarantee that for each language anyway.
maybe_not_ideal

HellAholic:
Proposal for universal icon for not overridden:
¯\_(ツ)_/¯

PR #20198, opened 2025-01-27
PP-550 PC-ABS & PC-ABS-FR 175

alanbjorklund:
Add PC-ABS & PC-ABS-FR configurations for Method X
and Method XL.  These materials are allowed on 1C,
1XA and Labs extruders.  Included is a solid intent mode.

The first PR which I closed had good model performance,
however the self support was printing inconsistently.
Larger support regions had under-extruded portions. 
This PR includes changes that address the under-extrusion,
including fan, line width, print speed and flow adjustments.

I ran prints on Method X and Method XL with both materials
to confirm the changes are a net positive.

I added PC-ABS & PC-ABS-FR to the sketch, sketch sprint, and
method exclude materials lists.

### DIFF
--- a/resources/definitions/ultimaker_method.def.json
+++ b/resources/definitions/ultimaker_method.def.json
@@ -59,6 +59,8 @@
             "zyyx_pro_",
             "octofiber_",
             "fiberlogy_",
+            "ultimaker_pc-abs",
+            "ultimaker_pc-abs-fr",
             "ultimaker_metallic_pla_175"
         ],
         "has_machine_materials": true,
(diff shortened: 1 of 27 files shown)


wawanbreton:
@alanbjorklund I had to fix some conflicts to update to main, please check that the printer definitions files are correct (excluded materials)

alanbjorklund:
Screenshot 2025-02-06 at 11 31 22 AM
These excluded materials look correct- PC-ABS and PC-ABS-FR should be excluded for Method, Sketch & Sprint, and included for only Method X and Method XL.

wawanbreton (approved):

PR #20200, opened 2025-01-27
PP-538 Initial Profiles for Replicator+

alanbjorklund:
Add definition for Replicator+, along
with three extruders: SE+, SE+ Tough,
and SE+ Experimental.

Add Replicator+ settings for PLA & TPLA.
These are similar to Method settings,
with some first layer adjustments.

PP-538


resources/definitions/ultimaker_replicator_plus.def.json
```diff
@@ -0,0 +1,257 @@
...
+            "ultimaker_metallic_pla",
+            "ultimaker_pc-abs"
+        ],
+        "has_machine_quality": true,
+        "has_materials": true,
+        "has_variants": true,
+        "machine_extruder_trains": { "0": "ultimaker_replicator_extruder" },
+        "preferred_material": "ultimaker_pla_175",
+        "preferred_quality_type": "draft",
+        "preferred_variant_name": "replicator_smart_extruder_plus",
```
HellAholic, line 67:
```suggestion
        "preferred_variant_name": "ultimaker_replicator_smart_extruder_plus",
```

resources/definitions/ultimaker_replicator_plus.def.json
```diff
@@ -0,0 +1,257 @@
...
+        "file_formats": "application/x-makerbot-replicator_plus",
+        "platform": "ultimaker_replicator_plus_platform.3MF",
+        "exclude_materials": [
+            "dsm_",
+            "Essentium_",
+            "imade3d_",
+            "chromatik_",
+            "3D-Fuel_",
+            "bestfilament_",
+            "emotiontech_",
```
HellAholic, line 19:
```suggestion
            "eazao_",
            "emotiontech_",
```

HellAholic:
image

Draft intent for the PLA has multiple errors with speeds set to 0 for:
`Top Surface Outer Wall Speed, Outer Wall Speed, Initial Layer Print Speed, Minimum Speed, Initial Layer Speed`

If cloud (DF) functionality is also part of the expected outcome, the name mapping should also be added for printer and extruders.

alanbjorklund:
Thanks for the feedback, I'll push a new commit today.  Those zero values are very strange, when I tested this locally there were no missing values or errors.

alanbjorklund:
@wawanbreton I corrected the zero values you mentioned.  However, when testing this I noticed the print temperatures were incorrect.  I was able to adjust 3 of the 4 temperatures, but the material_print_temperature value is still incorrect (200C, when according to ultimaker_pla_175 profile it should be 215C).

I thought this may have been due to the MakerBot vs UltiMaker name we assigned to the Replicator+, but I checked it against the Sketch printers which also use both UltiMaker and MakerBot printer names. The Replicator names are consistent with what we previously released for Sketch printers.

@Frederic98 I remember you mentioned the "+" in the extruder name (Smart Extruder+) might cause issues.  Could the incorrect temperature be due to the plus sign?

Frederic98:
Might need to add the machine to a translation dictionary somewhere in the fdm_material reader plugin

wawanbreton (approved):

Frederic98:
Cura seems to struggle with assembling the stack when first adding the printer. When I changed material to Tough PLA, and then back to PLA, it picked up the machine and nozzle settings from the fdm_material file

Frederic98:
Tried to reproduce it ~5 times but it didn't happen again. I did notice that Cura doesn't respect the `preferred_variant_name`. Maybe this is a related symptom?

PR #20201, opened 2025-01-27
PP-561 0.27 layer height options for Sprint

alanbjorklund:
Enable PLA and TPLA at 0.27 layer height via
Balanced & Draft profile options.

Draft mode has higher acceleration values, and
adjusted overhang settings.  Also additional minor settings changes
such as first layer height, and initial cooling fan.

PP-561

### DIFF
--- a/resources/definitions/ultimaker_sketch_sprint.def.json
+++ b/resources/definitions/ultimaker_sketch_sprint.def.json
@@ -130,7 +130,7 @@
             ]
         },
         "infill_before_walls": { "value": false },
-        "infill_line_width": { "value": 0.45 },
+        "infill_line_width": { "value": "0.45 if layer_height == 0.2 else 0.4" },
         "infill_overlap": { "value": 10 },
         "infill_pattern": { "value": "'lines'" },
         "infill_sparse_density": { "value": 15 },
--- a/resources/definitions/ultimaker_sketch_sprint.def.json
+++ b/resources/definitions/ultimaker_sketch_sprint.def.json
@@ -53,8 +53,7 @@
             "ultimaker_pva",
             "ultimaker_rapidrinse",
             "ultimaker_sr30",
-            "ultimaker_petg",
-            "ultimaker_metallic_pla"
+            "ultimaker_petg"
         ],
         "has_machine_quality": true,
         "has_materials": true,
@@ -380,10 +379,12 @@
         "wall_0_wipe_dist": { "value": 0.2 },
         "wall_line_width_x": { "value": 0.58 },
         "wall_overhang_angle": { "value": 35 },
-        "wall_overhang_speed_factor":
+        "wall_overhang_speed_factors":
         {
             "minimum_value_warning": 15,
-            "value": 17.5
+            "value": [
+                18
+            ]
         },
         "wall_thickness": { "value": 1 },
         "wall_x_material_flow_layer_0": { "value": "material_flow" },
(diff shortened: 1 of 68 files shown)


resources/definitions/ultimaker_sketch_sprint.def.json
```diff
@@ -130,7 +130,7 @@
             ]
         },
         "infill_before_walls": { "value": false },
-        "infill_line_width": { "value": 0.45 },
+        "infill_line_width": { "value": "0.45 if layer_height == 0.2 else 0.4" },
```
rburema, line 133: Are you sure you want this change only when the `layer_height` is _exactly_ `0.2`, or is it more a 'greater than', 'less than' situation?

alanbjorklund: I agree this isn't the best logic for infill line width, however for now I think this works well for the tuned profiles we'll be offering in 5.10.  We are going to tune more layer heights in the future, so I'll likely be updating this logic again to better fit with a larger set of profiles.

HellAholic:
The layer height in use for the intents is 0.27mm but the files are named as 0.3mm which create a disconnect between what the file name and what it is referring to.

alanbjorklund:
> The layer height in use for the intents is 0.27mm but the files are named as 0.3mm which create a disconnect between what the file name and what it is referring to.

@Frederic98 the profile generation script added "0.3" to the filename even though the layer height is 0.27.  Should I simply rename these after the script generates all of the profiles, or should we alter the script to name the files by the layer height?

Frederic98:
> > The layer height in use for the intents is 0.27mm but the files are named as 0.3mm which create a disconnect between what the file name and what it is referring to.
> 
> @Frederic98 the profile generation script added "0.3" to the filename even though the layer height is 0.27. Should I simply rename these after the script generates all of the profiles, or should we alter the script to name the files by the layer height?

That's because you've selected the profile to be `verydraft`. This [corresponds to a 0.3mm layer height](https://github.com/Ultimaker/autogenerate_print_profiles/blob/b1331c11fcbfe7cdec2e77b014f10d5622c71237/resources/all_qualities.csv#L5C1-L5C62). We could define a new quality type there, but I don't know if that's the best course of action...

Frederic98:
So we could:
- Define a 0.27mm layer height profile (call it "imperial" or something like that, the user doesn't see this name so it doesn't really matter what we call it, hence "verydraft" and "slightlycourse" 😅).
- Define a 0.25mm layer height, and change the profiles to that. Don't know how much of an impact that 0.02mm is going to have. This would require re-testing all the profiles and that doesn't seem feasible so close to release.
- Ignore it. Not very pretty to have 0.3mm in the filename when the profile is 0.27mm, but it doesn't influence anything. But if we ever do create a 0.3mm profile, this will give some big issues with conflicting names.

alanbjorklund:
I will test option 1 for now (Define a 0.27mm layer height profile) since there isn't enough time to fully test option 2.  I think we should avoid option 3 in case we eventually release a 0.3 profile for Sprint.

wawanbreton:
FYI @alanbjorklund I can't make a build for this branch because it is not properly named, the PP-XXX should always be first :slightly_smiling_face:

alanbjorklund:
> FYI @alanbjorklund I can't make a build for this branch because it is not properly named, the PP-XXX should always be first 🙂

Oh, I'm sorry about that.  Is there a way we can rename this branch?

wawanbreton:
> Oh, I'm sorry about that. Is there a way we can rename this branch?

The only way would be to create a new branch over the existing one, but you would have to re-create the PR and all the references to it would be broken, so maybe it's not worth it.

PR #20203, opened 2025-01-28
StartSliceJob.py regex - tweak to extruder_nr_expr capture group

DragRedSim:
# Description

Previously, any expression in a G-code block that was to be evaluated at slicing time could not contain a comma, as the regex would treat everything before the first comma found as the "expression", and anything after it as the "extruder_nr_expr". This causes a problem when trying to evaluate an expression that takes multiple values; for example, one such as `{min(machine_max_feedrate_x, machine_max_feedrate_y)*60, -1}`, which calculates a feedrate for moving the toolhead as part of the preparation steps. This change alters the selection of the "extruder_nr_expr" value to only find a string composed of numbers and the minus sign, in order to ensure this field only contains the extruder number in use, and thus allows the capture of the "expression" value to extend past the first comma, to take in the entire expression.

This fixes any expressions being interpreted by the parser which involve the use of a function that takes more than one argument.

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A - attempted slicing using a stock install of Cura 5.9.0, with the following line in the machine start G-code: `G0 X-2.1 Y20 F{min(machine_max_feedrate_x, machine_max_feedrate_y)*60, -1} ;Move to start position`. This line was output in the sliced gcode as `G0 X-2.1 Y20 FNone ;Move to start position`

- [ ] Test B - made change in the installed file StartSliceJob.py as documented in this pull request and restarted Cura, re-sliced the same file and checked the line in the start G-code, which now outputs `G0 X-2.1 Y20 F12000 ;Move to start position`

**Test Configuration**:

# Checklist:


DragRedSim:
OK, so turns out when I was developing the regex I didn't realise the `re` package used by Cura in this segment of code does not have support for the recursive function call I was using to ensure that all brackets were positively matched. The alternative `regex` library, however, does have support for this feature, however it is a third-party dependency, rather than the second-party one currently in use (as `re` is a core module of Python, whereas `regex` requires installation). I don't think there's much more I can do with this at the current point, not without manually parsing the string (basically counting brackets and looking for commas) to determine where a complete expression is able to be used.

wawanbreton:
Hi @DragRedSim, thanks for the proposal ! I took the liberty to add a unit test case that matches your example, but unfortunately it doesn't pass, the resulting extruder_nr seems to be incorrect. It works for you case because the settings you are using are global, not extruder-specific, so specifying the `-1` extruder is actually useless and can be removed.

But I do see value in this improvement, and although I am not a big fan of the braces-counting, it seems that you have already explored all the other possibilites, so this is the best way in the end :slightly_smiling_face:

PR #20240, opened 2025-02-05
CURA-11978 Retract and unretract in a travel

wawanbreton:
Adds 2 settings to setup the amount of retraction/priming that should be processed while travelling

Requires https://github.com/Ultimaker/CuraEngine/pull/2204


jellespijker (approved):
LGTM

PR #20260, opened 2025-02-11
Fix "Open in Cura" button on Thingiverse

hadess:
Make the "Open in Cura" button on Thingiverse work by informing
the system that Cura is the application to handle URLs with the
cura scheme, and that it can handle URLs, not just local files.

This fixes this error message visible in Firefox' console when clicking
the button on the website:
Prevented navigation to “cura://open?file=https%3A%2F%2Fwww.thingiverse.com%2Fdownload%3A14134689” due to an unknown protocol.

See https://github.com/bambulab/BambuStudio/commit/62006bba30f39


wawanbreton:
Hi, thanks for the improvement !

I do already have a setup that works though, on KUbuntu, so I don't really see why this is required, can you tell me a bit more ? I'm not very familiar with this and I had to struggle a bit to have it working actually :smile:

hadess:
I'm not sure what information you need. This is what's needed for the app to advertise it supports this URI scheme. See:
https://specifications.freedesktop.org/shared-mime-info-spec/latest/ar01s02.html#id-1.3.18

No configuration should be necessary.

wawanbreton:
Well, as we provide an AppImage, users don't really "install" Cura, they just download it and start the AppImage. In order to have it working with Firefox, I have to manually create a new Application entry that links to the AppImage, and has the proper arguments. I don't really know what the role of the embedded `.desktop` file is, but maybe I'm just doing it wrong.

hadess:
Cura is also shipped as a Flatpak on Flathub:
https://flathub.org/apps/com.ultimaker.cura

wawanbreton:
Well, I had never used Flatpak so far, but this is very convincing. It even extracts the contents of the AppImage so starting Cura is much faster with it.
And the PR makes more sense, thanks a lot :slightly_smiling_face:

wawanbreton (approved):

PR #20261, opened 2025-02-11
CURA-12047 3DConnexion devices support

wawanbreton:
Adds support to the 3DConnexion mouses natively to Cura, on Windows and Mac

CURA-7435
CURA-12047

:warning: Before merging, don't forget to change the mac build workflow and the cura_resources channel


fieldOfView:
I've said this before, but I strongly feel this should be a (builtin) plugin instead of being hard-coded into Cura itself. This is why Cura has plugins.

wawanbreton:
> I've said this before, but I strongly feel this should be a (builtin) plugin instead of being hard-coded into Cura itself. This is why Cura has plugins.

Hi @fieldOfView, well I totally agree with you :slightly_smiling_face: this has been developped by 3DConnexion itself, and in its current state they have integrated it directly, which was easier. But this PR is still a draft and we intend to move the code into a plugin before it reaches next release, hopefully that will fit into our planning.

plugins/3DConnexion/NavlibClient.py
```diff
@@ -0,0 +1,267 @@
...
+        self._renderer = renderer
+        self._pointer_pick = None
+        self._was_pick = False
+        self._hit_selection_only = False
+        self._picking_pass = None
+        self._pivot_node = OverlayNode(node=SceneNode(), image_path=Resources.getPath(Resources.Images, "cor.png"), size=2.5)
+        self.put_profile_hint("UltiMaker Cura")
+        self.enable_navigation(True)
+
+    def pick(self, x, y, check_selection = False, radius = 0.):
```
nallath, line 28: This could do with a bit more typing.

plugins/3DConnexion/NavlibClient.py
```diff
@@ -0,0 +1,267 @@
...
+                    elif not valid_hit and grid_resolution == 0.:
+                        return None
+
+                    if picked_depth < min_depth:
+                        min_depth = picked_depth
+                        result_position = self._picking_pass.getPickedPosition(coord_x, coord_y)
+
+        return result_position
+
+    def get_pointer_position(self) -> "pynav.NavlibVector":
```
nallath, line 74: Not sure if we should change this, but the rest of the code uses lower camel case instead of snakecase

plugins/3DConnexion/OverlayNode.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2016 Ultimaker B.V.
```
nallath, line 1: Don't think that copyright is correct

plugins/3DConnexion/OverlayNode.py
```diff
@@ -0,0 +1,68 @@
...
+# Cura is released under the terms of the AGPLv3 or higher.
```
nallath, line 2: Same goes for that one

requirements.txt
nallath, line 1: Why this empty file?

resources/themes/cura-light/theme.json
```diff
@@ -646,7 +646,10 @@
         "wizard_progress": [10.0, 0.0],
 
         "message": [30.0, 5.0],
-        "message_close": [2, 2],
+      "message_close": [
+        2,
+        2
+      ],
```
nallath, line 652: Why change this?

cura/CuraApplication.py
```diff
@@ -843,6 +845,7 @@ def _loadPlugins(self) -> None:
         self._plugin_registry.addType("profile_reader", self._addProfileReader)
         self._plugin_registry.addType("profile_writer", self._addProfileWriter)
         self._plugin_registry.addType("backend_plugin", self._addBackendPlugin)
+        self._plugin_registry.addType("view_manipulator", self._addViewManipulator)
```
nallath, line 848: I can't seem to find any documentation what a view manipulator is supposed to do. So for future reference, we should describe that somewhere.

cura/CuraApplication.py
```diff
@@ -1945,6 +1948,9 @@ def _addBackendPlugin(self, backend_plugin: "BackendPlugin") -> None:
     def getBackendPlugins(self) -> List["BackendPlugin"]:
         return self._backend_plugins
 
+    def _addViewManipulator(self, view_manipulator: "PluginObject"):
```
nallath, line 1951:
```suggestion
    def _addViewManipulator(self, view_manipulator: "PluginObject") -> None:
```

If you don't define what is returned, it doesn't automatically assume it's none (it assumes it can be anything)

nallath (changes requested):

plugins/3DConnexion/NavlibClient.py
wawanbreton: Most of those methds are overrides from the parent class, so we can't change them. And for consistency within the file, I suggest that we keep the same casing for all the methods.

resources/themes/cura-light/theme.json
wawanbreton: Probably a zealous IDE changing the indentation, there are other occurences.

PR #20270, opened 2025-02-12
Update version to 5.10.0 and adjust workflow paths

saumyaj3:
Upgraded the version in `conandata.yml` to 5.10.0. Updated the GitHub Actions workflow to include changes to `conandata.yml` in the trigger paths. This ensures builds run correctly when the version file is modified.

NP-732


HellAholic:
Changes are already on Main

PR #20278, opened 2025-02-14
Adjust settings ultimaker_s8.def.json

HellAholic:
CURA-12416

### DIFF
--- a/resources/definitions/ultimaker_s8.def.json
+++ b/resources/definitions/ultimaker_s8.def.json
@@ -382,7 +382,6 @@
             "value": "speed_wall"
         },
         "support_brim_line_count": { "value": 5 },
-        "support_density": { "value": "15 if support_structure == 'tree' else 20" },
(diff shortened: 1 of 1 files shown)


resources/definitions/ultimaker_s8.def.json
```diff
@@ -382,7 +382,6 @@
             "value": "speed_wall"
         },
         "support_brim_line_count": { "value": 5 },
-        "support_density": { "value": "15 if support_structure == 'tree' else 20" },
         "support_infill_rate": { "value": "80 if gradual_support_infill_steps != 0 else 15" },
```
wawanbreton, line 385: Where is the value gone ?

HellAholic: support_infill_rate is "Support Density", it was a case of double declaration once with the setting label and the other with the setting name.

wawanbreton: Ah ok :smile:

wawanbreton (approved):

PR #20291, opened 2025-02-19
CURA-12400 Improve open source referencing

wawanbreton:
Try to gather as much relevant information as possible about third-party libraries to display them in the about dialog

CURA-12400

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -607,3 +607,10 @@ python_translation_source_folders:
 qml_translation_source_folders:
   - resources/qml
   - plugins
+
+extra_dependencies:
+  conan:
+    version: "2.7.1"
+    sources_url: https://github.com/conan-io/conan
+    license: MIT
+    summary: Conan C/C++ package manager
--- a/conanfile.py
+++ b/conanfile.py
@@ -43,6 +43,7 @@ class CuraConan(ConanFile):
         "cura_debug_mode": [True, False],  # FIXME: Use profiles
         "internal": [True, False],
         "i18n_extract": [True, False],
+        "skip_licenses_download": [True, False],
     }
     default_options = {
         "enterprise": False,
@@ -52,6 +53,7 @@ class CuraConan(ConanFile):
         "cura_debug_mode": False,  # Not yet implemented
         "internal": False,
         "i18n_extract": False,
+        "skip_licenses_download": False,
     }
 
     def set_version(self):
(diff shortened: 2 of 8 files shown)


conandata.yml
```diff
@@ -607,3 +607,10 @@ python_translation_source_folders:
 qml_translation_source_folders:
   - resources/qml
   - plugins
+
+extra_dependencies:
+  conan:
+    version: "2.7.1"
```
rburema, line 613: I saw you put a comment in the Uranium repo for Qt to change the version in the requirements -- maybe it's a good idea to do something similar here, even if only in one place.

conanfile.py
```diff
@@ -135,6 +143,176 @@ def _python_installs(self):
 
         return python_installs
 
+    @staticmethod
+    def _is_repository_url(url):
+        # That will not work for ALL open-source projects, but should already get a large majority of them
+        return (url.startswith("https://github.com/") or url.startswith("https://gitlab.com/")) and "conan-center-index" not in url
```
rburema, line 149: Maybe also add sourceforge? Not quite as popular as it once was (for reasons), but still a few. It's fine if not though. -- I'm OK with this approach, but it doesn't seem as robust as you'd want (urllib.parse might be overkill and not help much though...)

conanfile.py
```diff
@@ -135,6 +143,176 @@ def _python_installs(self):
...
+                        with tarfile.open(sources_path, 'r:gz') as sources_archive:
+                            license_file = "LICENSE"
+
+                            for source_file in sources_archive.getnames():
+                                if Path(source_file).name == license_file:
+                                    sources_archive.extract(source_file, temp_dir)
+
+                                    license_file_path = os.path.join(temp_dir, source_file)
+                                    with open(license_file_path, 'r') as file:
+                                        dependency_description["license_full"] = file.read()
```
rburema, line 190: At this point we're like 9 indentations in, or so? 😅

conanfile.py
```diff
@@ -135,6 +143,176 @@ def _python_installs(self):
...
+                                         ("Source", False),
+                                         ("Source Code", False),
+                                         ("Repository", False),
+                                         ("Code", False),
+                                         ("homepage", True),
+                                         ("Homepage", True)]:
+            try:
+                url = data["info"]["project_urls"][source_url]
+                if check_source and not self._is_repository_url(url):
+                    # That will not work for ALL open-source projects, but should already get a large majority of them
```
rburema, line 202: Maybe do a log statement here.

conanfile.py
```diff
@@ -135,6 +143,176 @@ def _python_installs(self):
...
+    def _get_license_from_repository(sources_url, version, license_file_name = None):
+        git_url = sources_url
+        if git_url.endswith('/'):
+            git_url = git_url[:-1]
+        if not git_url.endswith(".git"):
+            git_url = f"{git_url}.git"
+        git_url = git_url.replace("/cgit/", "/")
+
+        tags = [f"v{version}", version]
+        files = ["LICENSE", "LICENSE.txt", "LICENSE.md", "COPYRIGHT", "COPYING", "COPYING.LIB"] if license_file_name is None else [license_file_name]
```
rburema, line 226: Could this also be (in) a folder sometimes?

(Also spelling, since we saw that LICENCE is technically correct 😛)

cura/UI/OpenSourceDependency.py
```diff
@@ -0,0 +1,40 @@
+#  Copyright (c) 2025 Ultimaker B.V.
```
rburema, line 1: Nitpick: Should be UltiMaker these days (drop the B.V., capitalize the M).

resources/qml/Dialogs/AboutDialog.qml
```diff
@@ -95,23 +92,45 @@ UM.Dialog
...
             Layout.fillWidth: true
             Layout.preferredWidth: 1
+
+            Component
+            {
+                id: componentLicenseDialog
+
+                LicenseDialog
+                {
+                }
```
rburema, line 121: Did you mean to add anything in here? -- I'm used to those being on the same line if not 😅

rburema (approved):
I only have very small remarks. (Though some of the code is quite build-script-y, but that's to be expected...)

conandata.yml
wawanbreton: True, the conan version is in every recipe, but I can add it at least to the Cura one

conanfile.py
wawanbreton: For all our dependencies, I had only GitHub and Gitlab. And this is actually very convenient because they have similar "URL APIs", not sure this is the case with sourceforge

conanfile.py
wawanbreton: Yes indeed :disappointed: I'll make a separate function

conanfile.py
wawanbreton: Again, this is based on what I found on the repositories of our actual dependencies. When they do a folder, that usually means they have multiple licences so I don't know which one to pick. This is a best-effort.

resources/qml/Dialogs/AboutDialog.qml
wawanbreton: Yes indeed, missed that :smile:

conanfile.py
wawanbreton: Yes indeed :disappointed: I'll make a separate function

rburema (approved):
lgtm

PR #20315, opened 2025-02-26
CURA-12359 write upgrade script for cura 5.10

wawanbreton:
The upgrade script was already written, so this PR just polishes the new/changed settings visibility, applies the new SDK version, and converts a forgotten setting in some Method profiles.

CURA-12359


rburema:
Note to future self(/selves): Don't forget the `"setting_version"` in the `fdmprinter` and `fdmextruder` files next time.

wawanbreton:
> Note to future self(/selves): Don't forget the `"setting_version"` in the `fdmprinter` and `fdmextruder` files next time.

Maybe that would deserve a documentation with the different steps to look over ? There are a few indeed, and it is the same for every release.

rburema:
> Maybe that would deserve a documentation with the different steps to look over ? There are a few indeed, and it is the same for every release.

Agreed, though, if we automate (most of) it, like discussed, we also get a sort of 'documentation in action'.

PR #20316, opened 2025-02-26
Update PurgeLinesAndUnload.py

GregValiant:
This change adds a setting for "minimum Z height" for the "nail down the string" moves.
There was an error in the "Prime Blob" code that didn't properly account for the retraction and there was no corresponding prime.  This commit fixes both of those.

# Description

Added a user setting for "move_to_start_min_z" so the user has an option of keeping the nozzle from actually touching down on the bed.  The description for "move to start" was updated.

Added flexibility

## Type of change

- [X ] Bug fix

**Test Configuration**:  Win 10 Pro and various Cura versions

# Checklist:


GregValiant:
@HellAholic, @wawanbreton  - this commit encompasses the addition of a setting for the "minimum Z height" for 'nailing down the string' as well as the bug fix for the Prime Blob code.

GregValiant:
@HellAholic you put the bug back in.
G1 F{retract_speed} E{**retract_distance**}
doesn't work because the printer is in absolute extrusion mode.  If the prime blob is 20mm then the retraction will be back to the "retract_distance" location.  For a direct drive that might be back to E=1 which would be a retraction of 19mm.  That leaves a dry start on the purge lines.
G1 F300 E20                     ; Blob
G1 F2100 E1.0                  ; Retract
That's why I changed it to
G1 F{retract_speed} E{**self.prime_blob_amount - retract_distance**}
I could have switched to relative and then back but I went for less lines to add.
G1 F300 E20                     ; Blob
G1 F2100 E19.0                ; Retract

You have to fix it because today is my birthday!!

HellAholic:
The code you added was:
```
purge_str += f"G1 F{self.retract_speed} E{self.retract_dist} ; Unretract\n"
purge_str += "G92 E0 ; Reset extruder\n"
```
This exact code is repeated 8 times, all the change I added does is uses a helper function to add the exact same code in the same place but allows to change the gcode pattern in one place and apply it to all 8 cases.
It does not change the output result at all so I'm confused by your comment here.

The code for
```
blob_string += f"G1 F{self.retract_speed} E{self.prime_blob_distance - self.retract_dist} ; Retract\n"
```
is still in the same place, that is not changed.

GregValiant:
Sorry, that's my bad.
It does work as expected and avoids the overly long retraction so I'll retract my statement.

HellAholic:
@GregValiant I made cherry-pick PR to 5.10 branch for this as the changes are bug fixes and non-breaking.
https://github.com/Ultimaker/Cura/pull/20368

GregValiant:
@HellAholic it doesn't appear that this has been implemented yet.  Will it make it into 5.10 stable?

HellAholic:
The PR you made was against main (so it wouldn't get there). I did some cherry-pick from commits and made a PR to 5.10
Should be in :P

GregValiant:
Should I close this one or should it remain open?

Issue #20183, opened 2025-01-24
Settings values replacements not working when in a single line (for Klipper Macro) when using extruder indexing


seth995:
### Cura Version

5.9

### Operating System

MacOS Sierra

### Printer

custom w/ klipper

### Reproduction steps

File has 2 parts, temp for extruder0 is 200, temp for extruder1 is 240, Bed temp is 70

add start gcode in single line:
START_PRINT EXTRUDER={extruders_enabled_count} EXTRUDER1=1 EXTRUDER_TEMP={material_print_temperature, 0} EXTRUDER1_TEMP={material_print_temperature, 1} BED_TEMP={material_bed_temperature}

When separated as such each value comes through as expected:
START_PRINT 
EXTRUDER={extruders_enabled_count} 
EXTRUDER1=1 
EXTRUDER_TEMP={material_print_temperature, 0} 
EXTRUDER1_TEMP={material_print_temperature, 1} 
BED_TEMP={material_bed_temperature}

### Actual results

Values in produced Gcode from single line is :
START_PRINT EXTRUDER=2 EXTRUDER1=1 EXTRUDER_TEMP=210

values in produced Gcode from separated line is:
START_PRINT 
EXTRUDER=2 
EXTRUDER1=1 
EXTRUDER_TEMP=200 
EXTRUDER1_TEMP=240 
BED_TEMP=70

### Expected results

Values should be 
START_PRINT EXTRUDER=2 EXTRUDER1=1 EXTRUDER_TEMP=200 EXTRUDER1_TEMP=240 BED_TEMP=70

### Add your .zip and screenshots here ⬇️

[CURA5_9_Bug.zip](https://github.com/user-attachments/files/18537689/CURA5_9_Bug.zip)

GregValiant:
Thanks for the report.
I agree there is a bug in there somewhere.
It appears that {material_print_temperature, 0} is causing it to fail.  If I leave out the extruder number it works.
Usually, when there is a failure in a replacement the entire code fails and the gcode will show the replacement pattern text unchanged.
I have no idea where that "210" number is coming from.
This seems to be OK:
START_PRINT EXTRUDER={extruders_enabled_count} EXTRUDER1=1 BED_TEMP={material_bed_temperature}
EXTRUDER_TEMP={material_print_temperature,0}
EXTRUDER1_TEMP={material_print_temperature,1}

I think the next release of Cura is coming up.  @HellAholic or @wawanbreton this would appear to warrant some attention.

[...]

PR #20318, opened 2025-02-26
CURA-12386 G-Code replacement fails with single line code

wawanbreton:
The regex was not specific enough and would catch the rest of the line, now we force stopping as soon as we see a `{` or `}`

CURA-12386
Fixes [20183](https://github.com/Ultimaker/Cura/issues/20183)

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -265,6 +265,10 @@ pycharm_targets:
     module_name: Cura
     name: pytest in TestSettingVisibilityPresets.py
     script_name: tests/Settings/TestSettingVisibilityPresets.py
+  - jinja_path: .run_templates/pycharm_cura_test.run.xml.jinja
+    module_name: Cura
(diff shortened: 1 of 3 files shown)


rburema (approved):
Looks good to me.
- Unit-tests! Yes! 🎈 
- I briefly thought about `[^}]` instead of `[^{}]`, but that wouldn't allow us to nest either (future features ...) and this does indeed make just as much sense (it's just wether you have a malformed bracket _inside_ the replacement versus one _outside_ -- but in the end it's the same syntax error).

rburema:
Interestingly, it produces `None` in the GCode when you do `{}`, and just ignores any errors (so there's not error shown to the user) -- but I think it's quite fine for now!

wawanbreton:
Yeah, the idea is just to have this working as best as possible within the current limitations, until we replace it by something more powerful (Jinja !!!)

PR #20321, opened 2025-02-26
[CURA-12254] Don't lose text-field focus when scrolling.

rburema:
When scrolling to settings, we might want to collapse the opened pop-ups (from extruder choses and the like) ... but that shouldn't come at the expense of losing focus on text-field settings occasionally (most notably of course if the scroll-bar changes length due to what you just changed).


wawanbreton (approved):
Nice simple solution :slightly_smiling_face:

PR #20325, opened 2025-02-27
Improved self-support settings for the S8

pkuiper-ultimaker:
Improved self-support settings for the S8. Better release due to the horizontal expansion. More robustness due to brims, and larger build plate trunks and avoid travel over support. Faster printing due to higher support speeds and less tree branches.Improved part strength due to higher infill flow rates. Added 1 flooring layer to all Ultimaker machines to improve bottom part quality. Small bug fixes to the sketch/replicator profiles. Updated settings version def files

PP-580 and PP-581


wawanbreton (approved):
Looks good !

PR #20344, opened 2025-03-04
[CURA-12441] Margin wasn't wide enough to see that it's a tree.

rburema:
Reduce apparent visual clutter by making the supposed meaningful margin actually meaningful visually.


wawanbreton (approved):
Awesome :smile:

PR #20345, opened 2025-03-04
[CURA-12442] Same cache as other setting-lists to reduce slowness on scroll.

rburema:
This does incur a cost on the first load of the list though.

### DIFF
--- a/resources/qml/Preferences/SettingVisibilityPage.qml
+++ b/resources/qml/Preferences/SettingVisibilityPage.qml
@@ -164,6 +164,7 @@ UM.PreferencesPage
                 expanded: ["*"]
                 visibilityHandler: UM.SettingPreferenceVisibilityHandler {}
             }
+            cacheBuffer: 1000000   // Set a large cache to effectively just cache every list item.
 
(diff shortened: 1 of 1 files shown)


wawanbreton (approved):

wawanbreton:
The delay at loading is a bit long indeed, but acceptable. I assume you have tested intermediate values ?

rburema:
> The delay at loading is a bit long indeed, but acceptable. I assume you have tested intermediate values ?

@wawanbreton I just used this value because the other caching lines (in the normal settings view and one other place, also loading settings) have the same to be honest 😅

wawanbreton:
> @wawanbreton I just used this value because the other caching lines (in the normal settings view and one other place, also loading settings) have the same to be honest 😅

Ok, I did a quick test and intermediate values don't help at all for the scrolling, so :+1:

rburema:
> Ok, I did a quick test and intermediate values don't help at all for the scrolling, so 👍

Yeah I didn't think they would, because the only way it would end up slower due to increased cache size, is if it somehow reserves all the cache space beforehand, which I don't think is a good idea in a caching mechanism that's supposed to have general use.

PR #20346, opened 2025-03-04
[CURA-12443] Fix value-units can now change visually/immediately on printer switch.

rburema:


resources/qml/Settings/SettingTextField.qml
```diff
@@ -92,6 +92,8 @@ SettingItem
 
         UM.Label
         {
+            id: unitlabel
```
wawanbreton, line 95:
```suggestion
            id: unitLabel
```

wawanbreton (approved):
Just a CaSinG suggestion, otherwise good

resources/qml/Settings/SettingTextField.qml
rburema: That's OK! -- I can't directly commit this, since I end up using it in the binding below and that needs to be changed too then.

Hm, maybe I can try suggesting the other needed change myself and then bundling those changes 🤔

resources/qml/Settings/SettingTextField.qml
```diff
@@ -106,6 +108,16 @@ SettingItem
             horizontalAlignment: (input.effectiveHorizontalAlignment == Text.AlignLeft) ? Text.AlignRight : Text.AlignLeft
             textFormat: Text.PlainText
             color: UM.Theme.getColor("setting_unit")
+
+            Binding
+            {
+                target: unitlabel
```
rburema, line 114:
```suggestion
                target: unitLabel
```

resources/qml/Settings/SettingTextField.qml
wawanbreton: Yeah I basically used the suggestion feature to make it clear, but you don't HAVE to use it :slightly_smiling_face:

rburema: I figured, but it _is_ very easy when you can use them 😁.

PR #20352, opened 2025-03-06
CURA-12169 fix autoarrange for disallowed areas

wawanbreton:
1. Initial bug was caused by trying multiple placement strategies but not reconstructing the items between iterations, which caused inconsistencies in data given to libnest2d.
2. Changed the UI behavior to not display a partial re-arrange which is often worse than the current situation
3. Add more strategies to test to increase the likeliness of finding an arrangement

### DIFF
--- a/cura/Arranging/Nest2DArrange.py
+++ b/cura/Arranging/Nest2DArrange.py
@@ -54,63 +54,73 @@ class Nest2DArrange(Arranger):
         machine_depth = self._build_volume.getDepth() - (edge_disallowed_size * 2)
         build_plate_bounding_box = Box(int(machine_width * self._factor), int(machine_depth * self._factor))
 
+        # Use a tiny margin for the build_plate_polygon (the nesting doesn't like overlapping disallowed areas)
+        half_machine_width = 0.5 * machine_width - 1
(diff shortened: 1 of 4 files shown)


cura/Arranging/Nest2DArrange.py
```diff
@@ -80,37 +64,64 @@ def findNodePlacement(self) -> Tuple[bool, List[Item]]:
...
 
-            if clipped_area.getPoints() is not None and len(
-                    clipped_area.getPoints()) > 2:  # numpy array has to be explicitly checked against None
-                for point in clipped_area.getPoints():
-                    converted_points.append(Point(int(point[0] * self._factor), int(point[1] * self._factor)))
-
-                disallowed_area = Item(converted_points)
-                disallowed_area.markAsDisallowedAreaInBin(0)
-                node_items.append(disallowed_area)
+            polygons_disallowed_areas += _convert_points(clipped_area.getPoints())
```
rumboon, line 90: Is the `None` check on `clipped_area.getPoints()` no longer required?

wawanbreton: Yes, it is now done in the `_convert_points` function

rumboon (approved):

PR #20371, opened 2025-03-10
CURA-12453 fix auto arrange overlapping models

wawanbreton:
* Use DONT_ALIGN alignment, otherwise all the transforms get re-centered, including the ones of the fixed objects
* Remove DONT_ALIGN strategy for starting point, which is a non-handled case in the inner library

CURA-12453

:warning: To be merged after https://github.com/Ultimaker/Cura/pull/20352


HellAholic (approved):
image

PR #20384, opened 2025-03-13
CURA-12439 Improve TableView manual column resizing

wawanbreton:
Rewrite the interaction to allow resizing the columns relative to each other, which is much more natural

CURA-12439

### DIFF
--- a/resources/qml/TableView.qml
+++ b/resources/qml/TableView.qml
@@ -67,8 +67,8 @@ Item
                         right: parent.right
                         top: parent.top
                         bottom: parent.bottom
+                        rightMargin: -width / 2
                     }
-                    anchors.rightMargin: -width / 2
                     width: UM.Theme.getSize("wide_lining").width * 2
                     enabled: index < headerRepeater.count - 1
                     acceptedButtons: Qt.LeftButton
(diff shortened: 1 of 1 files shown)


resources/qml/TableView.qml
```diff
@@ -50,57 +52,106 @@ Item
...
-                    //Resize handle.
+                    // Resize handle
                     anchors
                     {
                         right: parent.right
                         top: parent.top
                         bottom: parent.bottom
                     }
-                    width: UM.Theme.getSize("default_lining").width
+                    anchors.rightMargin: -width / 2
```
rburema, line 71: This line be in the 'anchors' section above, as just `rightMargin: -width / 2`.

resources/qml/TableView.qml
```diff
@@ -50,57 +52,106 @@ Item
...
+                    enabled: index < headerRepeater.count - 1
+                    acceptedButtons: Qt.LeftButton
+                    cursorShape: enabled ? Qt.SizeHorCursor : Qt.ArrowCursor
+
+                    property var dragLastPos
 
-                    MouseArea
+                    onPressed: (mouse) => { dragLastPos = mapToItem(parent, mouse.x, mouse.y) }
+
+                    onPositionChanged: (mouse) =>
```
rburema, line 81: I didn't know this notation works! We usually just put `function(arg) { blah = bloeh(bleh) }` instead.

rburema (approved):
Looks good, let's test it.

resources/qml/TableView.qml
wawanbreton: It is quite uncommon indeed, but I discovered it somewhere in the Cura code :smile: not sure the other notation has the same meaning for callbacks, I'll try

rburema: It's fine if this works! I'm not even sure we have a coherent standard for QML tbh 😅

PR #20386, opened 2025-03-13
Cura 12466 experimental setting visibility

HellAholic:
CURA-12466


wawanbreton (approved):
A few remarks, but good to go, this is much faster !

PR #20389, opened 2025-03-14
CURA-12458 security fixes that came in via the inbox

wawanbreton:
Implements 2 fixes that have been reported as security issues. Although they are both not technically required, they are part of security good practices and will silent warnings for automatic reporting tools.

CURA-12458

### DIFF
--- a/cura/OAuth2/AuthorizationRequestHandler.py
+++ b/cura/OAuth2/AuthorizationRequestHandler.py
@@ -127,6 +127,7 @@ class AuthorizationRequestHandler(BaseHTTPRequestHandler):
     def _sendHeaders(self, status: "ResponseStatus", content_type: str, redirect_uri: str = None) -> None:
         self.send_response(status.code, status.message)
         self.send_header("Content-type", content_type)
+        self.send_header("Strict-Transport-Security", "max-age=900")
         if redirect_uri:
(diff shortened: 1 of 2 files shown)


cura/OAuth2/AuthorizationRequestHandler.py
```diff
@@ -127,6 +127,7 @@ def _handleNotFound() -> ResponseData:
     def _sendHeaders(self, status: "ResponseStatus", content_type: str, redirect_uri: str = None) -> None:
         self.send_response(status.code, status.message)
         self.send_header("Content-type", content_type)
+        self.send_header("Strict-Transport-Security", "max-age=900")
```
rburema, line 130: Do we need to have `includeSubdomains` as well? We might not need to, since it's a single endpoint.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security

rburema (approved):
Seems fine! (I don't think we need my remark.)

cura/OAuth2/AuthorizationRequestHandler.py
wawanbreton: Yeah the scope is extremely limited, if I'm correct we don't even use domains in this case because the callback usually goes directly to localhost, so let's keep it simple

Issue #20441, opened 2025-03-24
Post processing script At layer change GCO to insert all in uppercase not working with Klipper


bobRoss99999:
### Cura Version

5.1.0-Beta 1

### Operating System

Windows 11

### Printer

Ender 3 v3 ke Using Klipper Btt pad 7

### Reproduction steps

1. Added post processing script: Insert at layer change SET_VARIABLE VARIABLE=layer VALUE=Layer+1
2. Sliced and print

### Actual results

Didn't noticed the **Insert at layer change** notice/tool tips saying that **everything will be chaged to upper case**, so:

Klipper config displaying error **VARIABLE must not contain upper case**.

I understand that this error is by design, but is not working with Klipper GCODE command format.

I suggest to add a warning in the Tooltips and changing code to not change everything to Upper case.

### Expected results

Gcode to insert field not to be converted to upper case

### Add your .zip and screenshots here ⬇️

![Image](https://github.com/user-attachments/assets/e6121005-1ffa-4cda-9fe9-b8a8848a9234)

![Image](https://github.com/user-attachments/assets/32c7ed1d-e81f-40da-b4d3-64e6d6f823ae)

bobRoss99999:
Adding this  From klipper documentation:

![Image](https://github.com/user-attachments/assets/cf0fffcb-37ec-4d45-9e29-4f256330419b)

HellAholic:
![Image](https://github.com/user-attachments/assets/0f3b4be0-45ab-4e11-9915-24502f7c1c62)
You can use the script version below, made the conversion to upper case a checkbox, if you don't select it, the code will be inserted exactly as you type it in.

[InsertAtLayerChange.zip](https://github.com/user-attachments/files/19455334/InsertAtLayerChange.zip)

![Image](https://github.com/user-attachments/assets/2fa65dbc-9f13-4fa5-8abc-c9c0d1283d4d)

https://github.com/Ultimaker/Cura/pull/20444

PR #20444, opened 2025-03-25
InsertAtLayerChange: Make the conversion to uppercase optional

HellAholic:
Add an option to allow users to not convert to uppercase.

- Add the option type bool
- Set to be enabled by default
- Add condition to only convert if bool value is true

Solves: https://github.com/Ultimaker/Cura/issues/20441


GregValiant:
For all the myriad settings I stick into things it never occurred to me to add this.  I thought about no uppercase, but Marlin doesn't read lowercase commands.  The only reason I put the "upper" in there was to catch typos.
Nice fix.  Simple and even I can understand it.

plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
```diff
@@ -91,7 +100,7 @@ def execute(self, data):
         if not bool(self.getSettingValueByKey("enabled")):
             return data
         #Initialize variables
-        mycode = self.getSettingValueByKey("gcode_to_add").upper()
+        mycode = self.getSettingValueByKey("gcode_to_add").upper() if bool(self.getSettingValueByKey("convert_to_upper")) else self.getSettingValueByKey("gcode_to_add")
```
wawanbreton: Do you really need the `bool()` cast ?

wawanbreton (approved):
Looks good, just a question

plugins/PostProcessingPlugin/scripts/InsertAtLayerChange.py
HellAholic: You are correct, not needed as it already returns a bool.
-> 5db75acffbdc593bbb15efe0f1eca2b502631fce

PR #20473, opened 2025-04-03
[PP-609] Update travel speed settings for Ultimaker S8 

pkuiper-ultimaker:
Back to 500mm/s now that the edge case bug in the motion planner has been fixed in the latest firmware.

PP-609


pkuiper-ultimaker:
Still testing. Will reenable later.

PR #20476, opened 2025-04-03
[PP-611]-Hide-Method-Accel

alanbjorklund:
The Method gantry acceleration parameters do not allow a user to increase the print speed.  We have to expose and map another parameter (extruder acceleration) in order to increase the gantry acceleration values past the default value of 800.

Its confusing to the user if the exposed gantry acceleration parameters can be increased, yet do not produce a faster print.  Hiding these parameters for now resolves this confusion.

# How Has This Been Tested?

Eric and Alan looked at the UI to ensure it shows the expected parameters.  None of the changes produce different acceleration values, so no changes to the toolpath are expected.

### DIFF
--- a/resources/definitions/ultimaker_method_base.def.json
+++ b/resources/definitions/ultimaker_method_base.def.json
@@ -39,6 +39,9 @@
         "acceleration_infill":
         {
             "enabled": false,
+            "maximum_value": 3500,
+            "minimum_value": 200,
+            "minimum_value_warning": 750,
             "value": "acceleration_print"
         },
         "acceleration_layer_0":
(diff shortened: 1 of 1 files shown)


HellAholic (changes requested):
PR contains changes from main branch. Setting the version to 5.11 and applying other changes like the different header color of the UI.
Please create a new PR with only the intended changes.

rburema:
This was made against 5.10.x -- is this still relevant for 5.11.0?

alanbjorklund:
> This was made against 5.10.x -- is this still relevant for 5.11.0?

No, this is only for 5.10.x.  5.11 is planned to have these parameter exposed in the Cura UI.

PR #20480, opened 2025-04-04
[PP-602] New configurations for the S line printers

pkuiper-ultimaker:
PP-602


resources/definitions/ultimaker_s6.def.json
```diff
@@ -0,0 +1,53 @@
+{
+    "version": 2,
+    "name": "UltiMaker S6",
+    "inherits": "ultimaker_s8",
```
wawanbreton, line 4: Shouldn't it inherit S5 ?

resources/definitions/ultimaker_s6.def.json
```diff
@@ -0,0 +1,53 @@
...
+    "metadata":
+    {
+        "visible": true,
+        "author": "UltiMaker",
+        "manufacturer": "Ultimaker B.V.",
+        "file_formats": "application/x-ufp;text/x-gcode",
+        "platform": "ultimaker_s7_platform.obj",
```
wawanbreton, line 11: Is it really the exact same buildplate ?

resources/definitions/ultimaker_s6.def.json
```diff
@@ -0,0 +1,53 @@
...
+        "supports_network_connection": true,
+        "supports_usb_connection": false,
+        "variants_name": "Print Core",
+        "variants_name_has_translation": true,
+        "weight": -2
+    },
+    "overrides":
+    {
+        "adhesion_type": { "value": "'brim'" },
+        "machine_name": { "default_value": "Ultimaker S6" }
```
wawanbreton, line 51:
```suggestion
        "machine_name": { "default_value": "UltiMaker S6" }
```

resources/intent/ultimaker_s8/um_s8_cc_plus_0.6_petcf_0.2mm_engineering.inst.cfg
wawanbreton, line 1: This file has been moved, the previous version doesn't exist, is that on purpose ? Isn't it going to cause some trouble for people that did use it ?

resources/definitions/ultimaker_s6.def.json
wawanbreton: Note: this is consistent with S8, are they both wrong ?

resources/definitions/ultimaker_s6.def.json
pkuiper-ultimaker: No. We need to inherit the settings from the high speed printer.

resources/definitions/ultimaker_s6.def.json
pkuiper-ultimaker: You are correct, should have been the S5 platform. I will adjust. I also noticed that the backpanel is not showing correctly and needs to be adjusted.

resources/definitions/ultimaker_s6.def.json
pkuiper-ultimaker: No should be UltiMaker. I will adjust.

wawanbreton (approved):

wawanbreton (approved):

PR #20489, opened 2025-04-06
Create ZHopOnTravel.py

GregValiant:
I've always found Z-hops as a necessary evil, but rarely for an entire file.
This script allows the user to customize Z-hops in the Gcode for either particular layers, or a range of layers.  The user can add hops to get over feather edges that might curl up, or infill that the nozzle might strike.

The main options are:

- 'Layer List' which is a comma delimited list of individual layers ( Ex: "12,45,127")
- 'Range of Layers' which would be a contiguous range of layers (Ex: Start at the beginning of layer 12 and continue to the end of layer 127) .

Other options are as you see here:
image

The script is compatible with:
- Absolute and Relative Extrusion
- Firmware Retraction
- Extra Prime Amount >= 0
- Adaptive Layers
- G2/G3 arc moves are supported but are treated as straight line moves.

Cura travel settings:
- If Z-hops are enabled in Cura there will sometimes be second higher hop.  The "return to" height will be correct.
- Compatible if retractions are enabled or not.

The setting "Add Z-hops to Infill Only" is limited as the script can only tell when the nozzle is printing infill.  If the nozzle is traveling from a "Wall over there" to a "Wall over here" and crosses infill during the travel there would not be a Z-hop added.

The script is **_NOT_** compatible with:
- Print Sequence = "One at a Time"

Example gcode for an added retraction and Z-hop (Extra prime amt = 0.5mm³ and absolute extrusion, z-hop = 0.50)
```
G1 X163.72 Y109.548 E558.95488
G1 F2100 E551.95488                     ; Retract
G0 F600 Z3.25                           ; Hop Up
G0 F10500 X88.4 Y115
G0 X88 Y115
G0 F600 Z2.75                           ; Hop Down
G92 E551.73969                          ; Extra prime adjustment
G1 F2100 E558.95488                     ; UnRetract
G1 F4500 X87.6 Y115 E558.9721
```
I don't know how many people use Cura for pellet feeders, but they can't retract and so this could be used to add Z-hops when required.

## Type of change

**Test Configuration**:
Various versions of Cura

* Operating System:
Windows 10 Pro

# Checklist:


GregValiant:
My original thought was to alter the Z-hops that Cura puts in.  If the user was to set the "New Hop Height" to 0.00 over a certain layer range then that would negate the Cura hops for those layers.  They could enter layers "13,25,66" and on those specific layers the Z-hop height would be their new value which could be 0.00 as easy as anything else.  That idea goes back awhile to a script I called "AlterZHops".

Then a feature request here mentioned only hopping across Infill.  I can't do that in post process (no model) but I could add hopping when printing the infill.  That was easier to do when controlling all the hops, rather than simply altering the existing hops.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+        retract_speed = int(extruder[0].getProperty("retraction_retract_speed", "value")) * 60
+        prime_speed = int(extruder[0].getProperty("retraction_prime_speed", "value")) * 60
+        firmware_retract = global_stack.getProperty("machine_firmware_retract", "value")
+        relative_extrusion = global_stack.getProperty("relative_extrusion", "value")
+        self._cur_z = float(global_stack.getProperty("layer_height_0", "value"))
+        filament_dia = extruder[0].getProperty("material_diameter", "value")
+        extra_prime_vol = extruder[0].getProperty("retraction_extra_prime_amount", "value")
+        extra_prime_dist = extra_prime_vol / (math.pi * (filament_dia / 2)**2)
+        self._add_retract = self.getSettingValueByKey("add_retract")
+        min_travel_dist = self.getSettingValueByKey("min_travel_dist")
```
rburema, line 173: Unused variable (the `min_travel_dist` is later retrieved by another `self.getSettingValue...`).

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+        if list_or_range == "list_of_layers":
+            layer_string = self.getSettingValueByKey("layers_of_interest")
+            layer_list = layer_string.split(",")
+            layer_list.sort()
+            for layer in layer_list:
+                for num in range(2, len(data) - 1):
+                    if ";LAYER:" + str(int(layer) - 1) + "\n" in data[num]:
+                        index_list.append(num)
+            start_index = index_list[0]
+            end_index = index_list[len(index_list) - 1]
```
rburema, line 190: It seems `end_index` is only actually used in the `elif` block below, which is of course skipped when inside of this `if` block.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                lines[index] = front_txt + str(" " * (40 - len(front_txt))) +";" +  back_txt
+            dn_lines = "\n".join(lines) + "\n"
+        else: # for lines with no prime
+            front_txt = dn_lines.split(";")[0]
+            back_txt = dn_lines.split(";")[1]
+            dn_lines = front_txt + str(" " * (40 - len(front_txt))) +";" +  back_txt + "\n"
+        self._is_retracted = False
+        return dn_lines
+
+    def _track_all_axes(self, data: str, cmd_list: int, start_index: int, relative_extrusion: bool) -> str:
```
rburema, line 505: I think maybe the type for `cmd_list` isn't meant to be `int`.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                        self._is_retracted = False
+                elif relative_extrusion:
+                    if self._cur_e < 0 or "G10" in line:
+                        self._is_retracted = True
+                        self._cur_e = 0
+                    if line.startswith("G11"):
+                        self._is_retracted = False
+                        self._cur_e = 0
+        self._prev_e = self._cur_e
+        return
```
rburema, line 544: This method should either return a string, or the signature should be changed to reflect a return type of `None`.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                if hop_end > 0 and index >= hop_end:
+                    # If there is no 'F' in the next line then add one to reinstate the Travel Speed (so the z-hop speed doesn't carry over through the travel moves)
+                    if not " F" in lines[index] and lines[index].startswith("G0"):
+                        lines[index] = lines[index].replace("G0", f"G0 F{speed_travel}")
+                    hop_down_lines = self.get_hop_down_lines(retraction_amount, speed_zhop, retract_speed, prime_speed, extra_prime_dist, firmware_retract, relative_extrusion, hop_height, lines[index])
+                    lines[index] = hop_down_lines + lines[index]
+                    self._is_retracted = False
+                    hop_end = 0
+                    hop_start = 0
+                    hop_down = ""
```
rburema, line 323: This variable doesn't seem to be used anywhere else?

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                retract_cnt += data[index_nr].count("; Retract")
+            msg_txt = str(hop_cnt) + " Z-Hops were added to the file\n"
+            if self._add_retract:
+                msg_txt += str(retract_cnt) + " Retracts and unretracts were added to the file"
+            Message(title = "[Z-hop On Travel]", text = msg_txt).show()
+        except:
+            pass
+        return data
+
+    def _total_travel_length(self, l_index: int, lines: str) -> float:
```
rburema, line 344: This signature claims the method is returning a `float`, but I think it's meant to return `Tuple[int,int]` (you'll have to import `Tuple` from `typing` ... or write `(int, int)`).

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        This function gets the distance from the previous location to the current location.
+        """
+        try:
+            dist = math.sqrt((self._prev_x - self._cur_x)**2 + (self._prev_y - self._cur_y)**2)
+        except ValueError:
+            return 0
+        return dist
+
+    def get_hop_up_lines(self, retraction_amount: float, speed_zhop: str, retract_speed: str, prime_speed: str, extra_prime_dist: float, firmware_retract: bool, relative_extrusion: bool, hop_height: str) -> str:
```
rburema, line 380: The `prime_speed` parameter is unused in this function (though I suppose that may be to keep the signature congruent to the other one).

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+            self._cur_e = round(self._cur_e - retraction_amount, 5)
+        if reset_type in [3, 7] and hop_retraction: # add retract and firmware retract
+            up_lines = "G10 ; Retract\n" + up_lines
+        if reset_type in [5, 13] and hop_retraction: # add retract and relative extrusion
+            up_lines = f"G1 F{retract_speed} E-{retraction_amount} ; Retract\n" + up_lines
+            self._cur_e = 0
+        if reset_type in [11, 15] and hop_retraction: # add retract, firmware retraction, extra prime
+            up_lines = "G10 ; Retract\n" + up_lines
+
+        # Format the added lines for readability
```
rburema, line 412: Maybe consider making this fragment a function, since it's repeated verbatim in the other method.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                lines[index] = front_txt + str(" " * (40 - len(front_txt))) +";" +  back_txt
+            up_lines = "\n".join(lines) + "\n"
+        else: # for lines without a retraction
+            front_txt = up_lines.split(";")[0]
+            back_txt = up_lines.split(";")[1]
+            up_lines = front_txt + str(" " * (40 - len(front_txt))) +";" +  back_txt + "\n"
+        return up_lines
+
+    # The Zhop down may require different kinds of primes depending on the Cura settings.
+    def get_hop_down_lines(self, retraction_amount: float, speed_zhop: str, retract_speed: str, prime_speed: str, extra_prime_dist: str, firmware_retract: bool, relative_extrusion: bool, hop_height: str, next_line: str) -> str:
```
rburema, line 427: (Same story here, but with `retract_speed` being the unused one instead -- though again that may be because of the signature.)

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,544 @@
...
+                dn_lines += "\nM82 ; Absolute extrusion"
+        # Add retract 1 + relative extrusion 4 + extra prime 8
+        if reset_type == 13 and hop_retraction:
+            dn_lines += f"\nG1 F{prime_speed} E{round(retraction_amount + extra_prime_dist, 5)} ; Unretract with extra prime"
+        # Add retract 1 + firmware retraction 2 + relative extrusion 4 + extra prime 8
+        if reset_type == 15 and hop_retraction:
+            dn_lines += "\nG11 ; UnRetract"
+            dn_lines += f"\nG1 F{prime_speed} E{round(extra_prime_dist, 5)} ; Extra prime"
+
+        # Format the added lines for readability
```
rburema, line 490: (Here's the other instance of the repeated lines.)

rburema (commented):
I usually don't do these anymore, but as Erwan is busy-ish, I made some review-comments to hopefully help this along.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Removed the line.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Vestigial.  That line isn't required anymore and I removed it.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Yeah, should be 'str'.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: The function has no value to return and is just re-setting global variables.  I'll change it to "Return None".

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: That should be hop_down_lines = "" to re-initialize the variable.
Changed.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Changed to  -> int: and the return to a list as "return [l_index, g_num]" or [0, 0].
That seems to work fine.  It is treated as a list when it is called.
If you prefer it to be a Tuple please handle that.  I always avoid Tuples and Quaternions.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Not required so I removed it.  Probably vestigial.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: I changed the previous line to "if reset_type in [3, 7, 11, 15] and hop_retraction:" which seems to cover it.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Combined type 7 with type 3:  "if reset_type in [3, 7] and hop_retraction:"

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
GregValiant: Not quite the same thing.  Combining the Type 11 with the Type 15 would result in unnecessary M83 lines in the gcode.
I didn't make this change at this time.

GregValiant:
What is "Resolve Conversation"?

I'm going to do some more debugging on this just to make sure there are no "unexpected consequences" or a fat finger caused an inadvertent delete.
I'll make the new commit later today or tomorrow.
Thanks for the help.

GregValiant:
I haven't been able to break it so I've made a new commit with the changes.

GregValiant:
@HellAholic I made a commit based on feedback on the UM forum.  The inserted Z-hop lines used "G0 F Z" and I have changed those to "G1 F Z" to match the Cura syntax.

GregValiant:
I found a bug in line 505.  It resulted in the first Z hop-up being wrong and effected the entire layer.

GregValiant:
I found and fixed a bug that occurred when the "End Layer" was greater than the "Total Layer Count" of the print.

GregValiant:
Commit 7 is a bug fix of an initial design error.

GregValiant:
@HellAholic has there been any internal discussion regarding including or dropping this?  There have been questions on a couple of forums.

HellAholic:
@GregValiant Nothing in terms of dropping. We do have a ticket for this to include it: CURA-12523

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+                    "description": "Enter '-1' to indicate the top layer, or enter a specific Layer number from the Cura preview.  The changes will end at the end of this layer.",
+                    "unit": "Lay# ",
+                    "type": "int",
+                    "default_value": -1,
+                    "minimum_value": "-1",
+                    "enabled": "zhop_travel_enabled and list_or_range == 'range_of_layers'"
+                },
+                "hop_height": {
+                    "label": "Z-Hop Height",
+                    "description": "I refuse to provide a description for this.",
```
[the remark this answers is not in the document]
HellAholic: Apparently "Resistance is futile" 🤣

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+        except:
+            pass
+        return data
+
+    def _total_travel_length(self, l_index: int, lines: str) -> int:
+        """
+        This function gets the cummulative total travel distance of each individual travel move.
+        :parameters:
+            g_num: is the line index as passed from the calling function and when returned indicates the end of travel
+            travel_total: is the cummulative travel distance
```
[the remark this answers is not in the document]
GregValiant: Thank you Mrs. Trudell.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+                    end_index = len(data) - 3
+                else:
+                    end_index = len(data) - 2
+            elif end_layer != -1:
+                for num in range(1, len(data) - 1):
+                    if ";LAYER:" + str(end_layer) + "\n" in data[num]:
+                        end_layer = data[num].splitlines()[0].split(":")[1]
+                        end_index = num
+                        break
+            if end_index == None:
```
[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+                            self._prev_y = self._cur_y
+                            self._cur_y = self.getValue(line, "Y")
+                        if " Z" in line and self.getValue(line, "Z"):
+                            self._cur_z = self.getValue(line, "Z")
+
+                    # Check whether retractions have occured
+                    if line[0:3] in ["G1 ", "G2 ", "G3 "] and "X" in line and "Y" in line and "E" in line:
+                        self._is_retracted = False
+                        self._cur_e = self.getValue(line, "E")
+                    elif (line.startswith("G1") and "F" in line and "E" in line and not "X" in line or not "Y" in line) or "G10" in line:
```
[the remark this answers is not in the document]
GregValiant: Done.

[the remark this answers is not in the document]
GregValiant: Done but needs additional debugging.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+        Determine if the hop will require a retraction
+        :parameters:
+            reset_type:  An indicator to handle differences when Firmware Retraction, and Relative Extrusion, and Extra Prime are enabled
+            up_lines:  The inserted line(s) for the Z-hop Up
+            front_text and back_text:  Are the line splits to account for existing gcode lines that have comments in them
+        """
+        hop_retraction = not self._is_retracted
+        if not self._add_retract:
+            hop_retraction = False
+        reset_type = 0 # no other options
```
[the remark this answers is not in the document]
GregValiant: Done.

plugins/PostProcessingPlugin/scripts/ZHopOnTravel.py
```diff
@@ -0,0 +1,553 @@
...
+                        self._cur_y = self.getValue(line, "Y")
+                    if " Z" in line and self.getValue(line, "Z"):
+                        self._cur_z = self.getValue(line, "Z")
+
+                # Check whether retractions have occured and track the E location
+                if not relative_extrusion:
+                    if line.startswith("G1 ") and " X" in line and " Y" in line and " E" in line:
+                        self._is_retracted = False
+                        self._cur_e = self.getValue(line, "E")
+                    elif line.startswith("G1 ") and " F" in line and " E" in line and not " X" in line and not " Y" in line:
```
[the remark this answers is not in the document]
GregValiant: "Similar" is not "Exactly alike".
Refused as this one does not have mention of G10.

GregValiant:
@HellAholic I've been known to make messes when I make these changes with a new commit.
Here is the altered file.  If you would be so kind as to do what needs to be done I would appreciate it.

[ZHopOnTravel Commit.zip](https://github.com/user-attachments/files/22311433/ZHopOnTravel.Commit.zip)

HellAholic:
@GregValiant I have a suggestion to add a limiter to the z-hop so people don't accidentally set a large value

```
                "hop_height": {
                    "label": "Z-Hop Height",
                    "description": "The relative 'Height' that the nozzle will 'Hop' in the 'Z'.",
                    "unit": "mm  ",
                    "type": "float",
                    "default_value": 0.5,
                    "minimum_value": 0,
                    "maximum_value_warning": 5,
                    "maximum_value": "10 if max_hop_height_protection else 999",
                    "enabled": "zhop_travel_enabled"
                },
                "max_hop_height_protection": {
                    "label": "Z-Hop Height limiter",
                    "description": "Sets a maximum height limit of 10mm for the Z-Hop. When disabled, allows values up to 999mm. Make sure there is enough clearance for the Z-hop height selected if you decide to disable this.",
                    "type": "bool",
                    "default_value": true
                },
```

GregValiant:
Instead of that, just set the maximum to 10.  If they need more then they must have a big custom printer.  They can alter the code themselves.

I would add this in "get_hop_up_lines" at around line 426:
```
        if self._cur_z + hop_height < self.machine_height:
            up_lines = f"G1 F{speed_zhop} Z{round(self._cur_z + hop_height,2)} ; Hop Up"
        else:
            up_lines = f"G1 F{speed_zhop} Z{round(self.machine_height,2)} ; Hop Up"
```
That will keep the upper height below the max of the machine to avoid crashing.

HellAholic (approved):

PR #20496, opened 2025-04-08
[PP-607] Improve CPE and PC surface quality

pkuiper-ultimaker:
PC: Prevent holes in outside wall of PC prints due to air capture during unretracts:
Reduce retract length, inner wall before outer wall, inner wall speed equal to outer wall. We also lowered the flow rate to 95% for PC because we noticed that it was over extruding.

CPE: CPE is VERY fragile to printer over itself or bumps. Lower all speeds to 40mm/s, goto 'lines' infill type to prevent crossing over lines, 3 walls to prevent issues with infill to reach the outer wall.

PP-607

### DIFF
--- a/resources/quality/ultimaker_s8/um_s8_cc_plus_0.4_pc_0.2mm.inst.cfg
+++ b/resources/quality/ultimaker_s8/um_s8_cc_plus_0.4_pc_0.2mm.inst.cfg
@@ -14,6 +14,8 @@ weight = -2
 [values]
 cool_min_layer_time = 6
 cool_min_layer_time_fan_speed_max = 12
-retraction_amount = 8
+inset_direction = inside_out
+material_flow = 95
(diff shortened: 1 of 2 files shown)


resources/quality/ultimaker_s8/um_s8_cc_plus_0.4_pc_0.2mm.inst.cfg
```diff
@@ -14,6 +14,8 @@ weight = -2
 [values]
 cool_min_layer_time = 6
 cool_min_layer_time_fan_speed_max = 12
-retraction_amount = 8
+inset_direction = inside_out
+material_flow = 95
```
wawanbreton, line 18: Is it specific to this printer ? Shouldn't it be in the material definition ?

wawanbreton (approved):

PR #20518, opened 2025-04-17
CURA-12099 export and import to bambu 3mf

wawanbreton:
BambuLab printers use a custom 3MF file format, which is 3MF plus some extra data inside, e.g. GCode file and other necessary information.
The 3MF exporter has then been extended to allow exporting those files, using a custom mime type.

CURA-12099
Requires https://github.com/Ultimaker/Uranium/pull/995


rburema (approved):
There's a refactor now that splits out the bambu specific code into it's own little space. Otherwise the code looked good to go!

wawanbreton (commented):
Nice refactoring @rburema, I unoficially approve this PR.
Only small suggestion: maybe the `DefaultVariant` should be called `CuraVariant` ?

rburema:
> the write function broke all other exporters 😅

Whoops, I should've caught that! Glad you realized it yourself.

> Only small suggestion: maybe the DefaultVariant should be called CuraVariant ?

Ah, since it's our project format, not 'default' 3MF as such -- yes, I think you're right. (I kept confusing myself which was which between ThreeMfVariant and DefaultVariant when 'writing'/evaluating this as well.) -- I'll fix that once I've finished going over my email/startup tasks.

--

P.S. I really want to pull main into this as it'll probably fix that unit-test thing, but didn't do so yet because this will not be merged _to_ main directly.

PR #20561, opened 2025-04-29
Fix loading creality ender3 v3ke

casperlamboo:
Use `default_value` for string fields. When using `value` for a string field the string is interpreted as a variable instead of a string. The fields should either be string-quoted (which is a bit ugly since there will be double quotes) or be a `default_value`.

NP-935


wawanbreton (approved):
Should be a good fix indeed

PR #20562, opened 2025-04-29
[CURA-12512] Add speed settings for auxilary ('build volume') fan.

rburema:
Previously you could only set on/off starting on a certain layer.

Backend PR: https://github.com/Ultimaker/CuraEngine/pull/2226

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -4887,13 +4887,12 @@
                 "build_volume_fan_speed_0":
                 {
                     "label": "Initial Layers Build Volume Fan Speed",
-                    "description": "Build-volume fan speed until 'Build Volume Fan Speed at Layer'.",
+                    "description": "The fan speed (as a percentage) for the auxiliary or build-volume fan, that is set until the layer specified at 'Build Volume Fan Speed at Layer' is reached. After that, the speed is set by 'Build Volume Fan Speed' instead (so not this 'Initial Layers' one).",
                     "unit": "%",
                     "type": "float",
                     "minimum_value": "0",
                     "maximum_value": "100",
                     "default_value": 0,
-                    "value": "0.0",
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -4884,6 +4884,36 @@
...
+                "build_volume_fan_speed_0":
+                {
+                    "label": "Initial Layers Build Volume Fan Speed",
+                    "description": "Build-volume fan speed until 'Build Volume Fan Speed at Layer'.",
+                    "unit": "%",
+                    "type": "float",
+                    "minimum_value": "0",
+                    "maximum_value": "100",
+                    "default_value": 0,
+                    "value": "0.0",
```
wawanbreton, line 4896: The `value` is no required if there is a `default_value` with the same number

resources/definitions/fdmprinter.def.json
```diff
@@ -4884,6 +4884,36 @@
...
+                "build_volume_fan_speed":
+                {
+                    "label": "Build Volume Fan Speed",
+                    "description": "Build-volume fan speed starting from 'Build Volume Fan Speed at Layer'.",
+                    "unit": "%",
+                    "type": "float",
+                    "minimum_value": "0",
+                    "maximum_value": "100",
+                    "default_value": 100,
+                    "value": "100.0",
```
wawanbreton, line 4911: The `value` is no required if there is a `default_value` with the same number

wawanbreton (commented):
Looks good, but I'm having difficulties understanding what the settings do. Maybe a rewording would be helpful.

resources/definitions/fdmprinter.def.json
rburema: done!

wawanbreton (approved):
Great, new descriptions are much more understandable :smile:

PR #20565, opened 2025-04-30
[CURA-12446] Added settings for a different number of walls in top-/bottom-most skin.

rburema:
See also engine PR: https://github.com/Ultimaker/CuraEngine/pull/2227


rburema:
- (label) name may not be OK yet, '... Skin' could confuse people into thinking its multiple layers instead of top/bottom most (exposed to air)

wawanbreton:
> * (label) name may not be OK yet, '... Skin' could confuse people into thinking its multiple layers instead of top/bottom most (exposed to air)

You can have a look at `flooring_extruder_nr` and `roofing_extruder_nr` for the wording, which is slightly different, not sure it is better but at least it would be consistent.

resources/definitions/fdmprinter.def.json
```diff
@@ -1843,6 +1843,51 @@
...
+                    "minimum_value": "0",
+                    "maximum_value_warning": "10",
+                    "maximum_value": "999999",
+                    "type": "int",
+                    "enabled": "(top_layers > 0 or bottom_layers > 0)",
+                    "limit_to_extruder": "top_bottom_extruder_nr",
+                    "settable_per_mesh": true,
+                    "children":
+                    {
+                        "wall_line_count_0":
```
wawanbreton, line 1861: Is this for the initial layer (on the build plate), or the roofing layers ? I think it it the latter, then I suggest:
```suggestion
                       "wall_line_count_flooring":
```

resources/definitions/fdmprinter.def.json
```diff
@@ -1843,6 +1843,51 @@
...
+                            "value": "wall_line_count_skin",
+                            "minimum_value": "0",
+                            "maximum_value_warning": "10",
+                            "maximum_value": "999999",
+                            "type": "int",
+                            "enabled": "bottom_layers > 0",
+                            "limit_to_extruder": "top_bottom_extruder_nr",
+                            "settable_per_mesh": true
+                        },
+                        "wall_line_count_top":
```
wawanbreton, line 1875: Same here, if I understand correctly:
```suggestion
                       "wall_line_count_roofing":
```

resources/definitions/fdmprinter.def.json
rburema: Oh right, the `0` postfix is mostly for the outer-wall, so this would work better.

resources/definitions/fdmprinter.def.json
rburema: I think this makes sense as well.

resources/definitions/fdmprinter.def.json
rburema: ... except the flooring doesn't necessarily have to be on the bottom, I think _initial might cover it more closely (which also communicates that it's really only 1 layer).

wawanbreton: Hmm, calling it `layer_0` does look confusing to me, it seems like it is only for the `initial` layer, the one on the build plate. `flooring` is definitely not perfect, but this is what we used for other settings, and also for consistency with the `roofing` settings. I let it to your judgment.

rburema: Yeah it _is_ only used for the initial layer, applying it to flooring in general would cause bridging to fail (and not be much better even if bridging is off -- the new 'floating' walls would still fall off the sides).

rburema: (See line 140 in layerPart.cpp in the engine : `layer_nr == 0 ? getTopOrBottom(-1, "wall_line_count_layer_0", layer_nr, slicer->layers, mesh.settings) : Shape(),` )

wawanbreton (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -1543,6 +1543,19 @@
...
+                        },
+                        "roofing_extension":
+                        {
+                            "label": "Top Surface Extension",
+                            "description": "Determines how much the top surfaces are extended beneath overlapping surfaces. By adjusting this value, you can ensure that the outer edges of the top surfaces are concealed by the layers above, resulting in a better visual quality, particularly for models with curved surfaces.",
+                            "type": "float",
+                            "default_value": "0.4",
+                            "value": "roofing_line_width * (wall_line_count_roofing + 1)",
+                            "minimum_value": "0",
+                            "maximum_value_warning": "roofing_line_width * 10",
```
rburema: Do we even want a max. value warning here?

resources/definitions/fdmprinter.def.json
```diff
@@ -1543,6 +1543,19 @@
...
                             "settable_per_mesh": true
+                        },
+                        "roofing_extension":
+                        {
+                            "label": "Top Surface Extension",
+                            "description": "Determines how much the top surfaces are extended beneath overlapping surfaces. By adjusting this value, you can ensure that the outer edges of the top surfaces are concealed by the layers above, resulting in a better visual quality, particularly for models with curved surfaces.",
+                            "type": "float",
+                            "default_value": "0.4",
+                            "value": "roofing_line_width * (wall_line_count_roofing + 1)",
+                            "minimum_value": "0",
```
rburema, line 1581: (might want to make this a wall-line width or half of that -- I'm not sure if we actually want this to be zero -- or even as a warning?)

resources/definitions/fdmprinter.def.json
wawanbreton: If I understand correctly, the warning is to say "your value will work but doesn't really make sense", which I think is the case case... But I would also be ok with removing it.

resources/definitions/fdmprinter.def.json
wawanbreton: Well this technically works and in some situations it could make sense, so I would just allow it.

rburema:
superseded by #21634

PR #20592, opened 2025-05-13
Add missing error-handling.

rburema:
There was error handling for the url-request, but not for handling the subsequent resolution.

(Comes from crash reports.)

CURA-12592


wawanbreton (approved):
Looks good, but do you know what the actual generated exception was ? The code looks good, so it should quite never occur...

PR #20612, opened 2025-05-21
Create PrintSkewCompensation.py

GregValiant:
This version creates a "machine_name.skew.log" file to record the skew settings.  The settings are machine specific and load every time the script loads and update every time the script runs.
This does not account for the 3 models that are necessary but there is a link at the top of the py file.

# Description
Provides 3 means of adjusting for skew:
Cura (post process the gcode)
Marlin M852
Klipper SET_SKEW

**Test Configuration**:
Tested with Cura 5.10.0 but it is non-version specific.
* Operating System:
* Windows 10 Pro

# Checklist:

- [ X] My coding style still sucks

### DIFF
--- a/plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
+++ b/plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
@@ -449,4 +449,7 @@ class PrintSkewCompensation(Script):
                 skew_log_file.write(the_line)
         except IOError as e:
             Logger.log("e", f"Failed to write to log file {log_file_name}: {e}")
+            Message(
+                title = "[Print Skew Compensation]",
+                text = "The log file could not be created.  Your settings cannot be saved and so cannot be recalled later.").show()
         return None


GregValiant:
@HellAholic 
If there is no "scripts" folder then this script will create it and that is where the log file would be placed.

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
```diff
@@ -0,0 +1,405 @@
...
+    def initialize(self) -> None:
+        super().initialize()
+
+        # Get the "Printer Name.skew.log" file and parse it for the settings.  If there isn't a log file, then create it.  The log file gets updated to the current settings every time the script runs.
+        set_lines = None
+        active_printer = Application.getInstance().getGlobalContainerStack().getProperty("machine_name", "value")
+        try:
+            config_path = Resources.getConfigStoragePath()
+            
+            pp_path = config_path + "\\scripts\\"
```
HellAholic, line 32:
```suggestion
            pp_path = os.path.join(config_path, "scripts")
```

Different operating systems handle the path joining differently so it's easier to use os.path.join to merge paths

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
```diff
@@ -0,0 +1,405 @@
...
+        script_list = scripts.split("\n")
+        for s_index, script in enumerate(script_list):
+            script_list[s_index] = script.split("]")[0]
+            script_list[s_index] = script_list[s_index].replace("[", "")
+        for s_index, script in enumerate(script_list):
+            if "PrintSkewCompensation" in script and s_index != 0:
+                Message(
+                    title = "[Print Skew Compensation]", 
+                    text = "Should be first in the Post-Processor list.  It will run if it isn't first, but any following post-processors should act on the changes made by 'Print Skew Compensation'.").show()
+                break
```
HellAholic, line 217: This part seems a bit complicated for what it's doing.
Wouldn't it be easier to just check if "[PrintSkewCompensation]" is in the list at index 0?

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
GregValiant: Later in execute I use:
log_file_name = config_path + "\\scripts\\" + active_printer + ".skew.log"
If I do the same trick as you suggest - how do I concatenate the file name?

HellAholic: That would be ```log_file_name = os.path.join(config_path, "scripts", f"{active_printer}.skew.log")```

HellAholic: Alternatively if you store the path in self for the pp_path, self.pp_path = os.path.join(config_path, "scripts")
then later you can just use that:
```log_file_name = os.path.join(self.pp_path, f"{active_printer}.skew.log")```

**Comment by GregValiant** [comment]
I made the changes and didn't appear to break anything.

I'm using the machine_name for the log file name.  When I pull the "machine_name" it always returns the generic "printer model" name (ex: UltiMaker S3) but I need the explicit name (ex: Ultimaker S3 #2).  If a user has more than one printer of the same model, the skew will be the same for all of them.

So my question is:  Where do I get the actual name of the printer?
image

**Comment by GregValiant** [comment]
Well that will work.
Young Greg provided the drum roll.  He's also a drummer in a heavy metal band.

**Comment by GregValiant** [comment]
Commit with the suggested changes and the bug fix.

File: plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py [hunk 3]
```diff
@@ -0,0 +1,394 @@
...
+        if self.compensation_method == "method_cura":
+            data = self.cura_compensation(data)
+        elif self.compensation_method == "method_marlin":
+            data = self.marlin_compensation(data)
+        elif self.compensation_method == "method_klipper":
+            data = self.klipper_compensation(data)
+
+        # Add the script settings to the gcode if requested
+        if self.getSettingValueByKey("add_settings_to_gcode"):
+            setting_string = f";  Print Skew Compensation Settings:\n;    xy_ac_measurement: {self.xy_ac_dist}\n;    xy_bd_measurement: {self.xy_bd_dist}\n;    xy_ad_measurement: {self.xy_ad_dist}\n;       XY skew factor: {round(self.xy_skew_factor,8)}\n;    xz_ac_measurement: {self.xz_ac_dist}\n;    xz_bd_measurement: {self.xz_bd_dist}\n;    xz_ad_measurement: {self.xz_ad_dist}\n;       XZ skew factor: {round(self.xz_skew_factor,8)}\n;    yz_ac_measurement: {self.yz_ac_dist}\n;    yz_bd_measurement: {self.yz_bd_dist}\n;    yz_ad_measurement: {self.yz_ad_dist}\n;       YZ skew factor: {round(self.yz_skew_factor,8)}\n"
```
HellAholic, line 246:
```suggestion
           setting_string = (
               f";  Print Skew Compensation Settings:\n"
               f";    xy_ac_measurement: {self.xy_ac_dist}\n"
               f";    xy_bd_measurement: {self.xy_bd_dist}\n"
               f";    xy_ad_measurement: {self.xy_ad_dist}\n"
               f";        XY skew factor:    {round(self.xy_skew_factor, 8)}\n"
               f";    xz_ac_measurement: {self.xz_ac_dist}\n"
               f";    xz_bd_measurement: {self.xz_bd_dist}\n"
               f";    xz_ad_measurement: {self.xz_ad_dist}\n"
               f";        XZ skew factor:    {round(self.xz_skew_factor, 8)}\n"
               f";    yz_ac_measurement: {self.yz_ac_dist}\n"
               f";    yz_bd_measurement: {self.yz_bd_dist}\n"
               f";    yz_ad_measurement: {self.yz_ad_dist}\n"
               f";        YZ skew factor:    {round(self.yz_skew_factor, 8)}\n"
           )
```
sorry my initial comment was not clear, this is what I meant with the all at once

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
```diff
@@ -0,0 +1,394 @@
...
+            self.xy_skew_factor = 0.00
+            self.xz_skew_factor = 0.00
+            self.yz_skew_factor = 0.00
+            self.add_settings_to_gcode = False
+        return None
+
+    def _write_log_file(self, log_file_name: str):
+        # Write the log file
+        skew_log_file = open(log_file_name, "w")
+        the_line = f"active_printer:{self.active_printer}\ncompensation_method:{self.compensation_method}\nxy_ac_dist:{self.xy_ac_dist}\nxy_bd_dist:{self.xy_bd_dist}\nxy_ad_dist:{self.xy_ad_dist}\nxz_ac_dist:{self.xz_ac_dist}\nxz_bd_dist:{self.xz_bd_dist}\nxz_ad_dist:{self.xz_ad_dist}\nyz_ac_dist:{self.yz_ac_dist}\nyz_bd_dist:{self.yz_bd_dist}\nyz_ad_dist:{self.yz_ad_dist}\nxy_skew_factor:{round(self.xy_skew_factor,8)}\nxz_skew_factor:{round(self.xz_skew_factor,8)}\nyz_skew_factor:{round(self.yz_skew_factor,8)}\nadd_settings_to_gcode:{str(self.add_settings_to_gcode)}"
```
HellAholic, line 391:
```suggestion
        the_line = (
            f"active_printer:{self.active_printer}\n"
            f"compensation_method:{self.compensation_method}\n"
            f"xy_ac_dist:{self.xy_ac_dist}\n"
            f"xy_bd_dist:{self.xy_bd_dist}\n"
            f"xy_ad_dist:{self.xy_ad_dist}\n"
            f"xz_ac_dist:{self.xz_ac_dist}\n"
            f"xz_bd_dist:{self.xz_bd_dist}\n"
            f"xz_ad_dist:{self.xz_ad_dist}\n"
            f"yz_ac_dist:{self.yz_ac_dist}\n"
            f"yz_bd_dist:{self.yz_bd_dist}\n"
            f"yz_ad_dist:{self.yz_ad_dist}\n"
            f"xy_skew_factor:{round(self.xy_skew_factor, 8)}\n"
            f"xz_skew_factor:{round(self.xz_skew_factor, 8)}\n"
            f"yz_skew_factor:{round(self.yz_skew_factor, 8)}\n"
            f"add_settings_to_gcode:{str(self.add_settings_to_gcode)}"
        )"
```
same here.

HellAholic: I can batch them both and commit directly from the review.

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
```diff
@@ -0,0 +1,424 @@
...
+import os.path
+from UM.Resources import Resources
+import os
+
+class PrintSkewCompensation(Script):
+    def initialize(self) -> None:
+        super().initialize()
+        """ Get the "Printer Name.skew.log" file and parse it for the settings.  If there isn't a log file, then create it for the active printer and use the default values.  The log file gets updated to the current settings every time the script runs.  When the user activates a different printer the script will re-initialize"""
+        set_lines = None
+        active_printer = Application.getInstance().getGlobalContainerStack().getName
```
HellAholic, line 27:
```suggestion
        active_printer = Application.getInstance().getGlobalContainerStack().getName()
```

plugins/PostProcessingPlugin/scripts/PrintSkewCompensation.py
```diff
@@ -0,0 +1,424 @@
...
+            self.yz_ac_temp = float(lines[8].split(":")[1])
+            self.yz_bd_temp = float(lines[9].split(":")[1])
+            self.yz_ad_temp = float(lines[10].split(":")[1])
+            self.xy_skew_factor = float(lines[11].split(":")[1])
+            self.xz_skew_factor = float(lines[12].split(":")[1])
+            self.yz_skew_factor = float(lines[13].split(":")[1])
+            self.add_settings_to_gcode = bool(lines[14].split(":")[1])
+        else:
+            # Set the defaults
+            self.active_printer = Application.getInstance().getGlobalContainerStack().getName
```
HellAholic, line 381:
```suggestion
            self.active_printer = Application.getInstance().getGlobalContainerStack().getName()
```

HellAholic:
Did some cleaning, also made a mess and reverted it.
- Checked the skew.log creation
- Checked the scripts folder creation (if non-existing, also works if it's already there)
Seems good.

Some additional things can be improved but I leave it to a dev reviewer to comment if they're necessary, just mentioning a few things that I would do:
- Separate the logic for getSettings from setDefaultSettings: having two separate smaller functions with explicit intent is better, you can also add some error handling in case the log file has been manually modified and the script cannot read it.
- The methods can be updated to constants rather than just literal strings 
- Probably can assign certain values like the path to the configuration, active_printer,... once and then re-use the same instance self.config_path, self.active_printer. Also ensures you're passing less variables between helper functions

GregValiant:
I hate git desktop.  I made some changes, didn't pull the file from here with your changes, and now there are numerous "conflicts".

Here.  I made some changes and split "setDefaults" out of "getSettings".  I manually added all your changes (I think).

[PrintSkewCompensation.zip](https://github.com/user-attachments/files/20378339/PrintSkewCompensation.zip)

I didn't change some things because I'm concerned about a user changing printers mid-stream.  Right now it seems to load the printer settings correctly for the printer instance that is active.

GregValiant:
I was more concerned about users getting in and changing things and the log becoming unreadable. 

I'm printing "unskewed" test models now and they have 2 hours to run.  If they turn out square this should be good to go.

(My XZ skew factor is 7.071e-05.  That's probably close enough for government work.)

HellAholic:
From what I can see, there is enough coverage in terms of exception handling for that in place.

fieldOfView:
I don't think a postprocessing _script_ should be in charge of creating the `scripts` folder. This functionality should be in the postprocessing _plugin_. I distinctly remember creating code that did that "long ago", this may be broken now if @GregValiant thinks he needs to do it from his script.

It rubs me in the wrong way that _logs_ are added into a _scripts_ folder. I find it much more logical to have the logs in the same folder as the main `cura.log` folder: `Resources.getStoragePath(Resources.Resources, log_filename)`  

Personally I think this whole postprocessing script should be a Marketplace plugin. It requires way too much explanation on how to use it and what to use it for to be plonked into the postprocessing script dropdown without explanation.

GregValiant:
We went round and round [HERE](https://github.com/HellAholic/PrintSkewCompensation) and [HERE](https://github.com/HellAholic/PrintSkewCompensation/pull/3) with "plugin vs postprocess".
Remco provided [this link](https://github.com/Ultimaker/Uranium/blob/main/UM/Settings/AdditionalSettingDefinitionsAppender.py) on the "right" way to do it.
At that point the subject of Muscle Cars became much more interesting.

HellAholic:
@GregValiant 
The Remco's comment was about the Support interface material change plugin. The other one was just a clean up (done already xD) but it might need changes to work with the new pp script you have (or at least take some ideas from it).

GregValiant:
I knew I had butchered one of the PR's.  That's better than my normal "butcher all of them".

The skew log could certainly go in the base Config folder with the Cura.log.

HellAholic:
Placing them in the root is fine but that would mean that they will not be copied over on upgrades (so if you run the script in 5.9, then upgrade to 5.10, the log files will not carry over to the 5.10 config folder).

GregValiant:
As Rosanna Rosannadana used to say ..."It's always something".

HellAholic:
Not sure why every normal discussion needs to be in the spelled out format but okay.
**root as the top-level directory on a file system, so if we're talking about the resources folder / configuration folder as its own file system, root is the same as where cura.log file lives.**

HellAholic:
You can remove the two warnings, info log should be enough, It's better to have the logs and take them out once everything is tested than to not have them and try and figure out why something is not working.

Side note, instead of a printer.skew.log, you set and get values in the user container stack of the active printer
to set the values as an example:
```
machine_name = Application.getInstance().getGlobalContainerStack().getName()
try:
        # Get the user container to set the property
        user_container = global_stack.findContainer(criteria = ["id", f"{machine_name}_user"])
        if user_container and isinstance(user_container, InstanceContainer):
                user_container.setProperty("xy_ac_measurement", "value", self.xy_ac_temp)
                user_container.setProperty("xy_bd_measurement", "value", self.xy_bd_temp)
                user_container.setProperty("xy_ad_measurement", "value", self.xy_ad_temp)
                user_container.setProperty("xz_ac_measurement", "value", self.xz_ac_temp)
                user_container.setProperty("xz_bd_measurement", "value", self.xz_bd_temp)
                user_container.setProperty("xz_ad_measurement", "value", self.xz_ad_temp)
                user_container.setProperty("yz_ac_measurement", "value", self.yz_ac_temp)
                user_container.setProperty("yz_bd_measurement", "value", self.yz_bd_temp)
                user_container.setProperty("yz_ad_measurement", "value", self.yz_ad_temp)
        else:
                # Something has gone wrong
except Exception as e:
       # Something has gone very wrong
```
and same story to get the values out of the user stack:
```
self.xy_ac_temp = user_container.getProperty("xy_ac_measurement", "value")
```
No need to store things in logs

GregValiant:
The simple "except:" works.  I tried with FileOpenError and IOError and neither worked as expected.

Putting the settings into the stack seems like it would work.
If the settings don't exist - create them and set the values to the script defaults.
If the settings do exist then grab the values to use in post process.

I'll see what I can do with it.  In the meantime, I'll drop this back to draft status.

It's funny - all this work and the actual "working section" of the script hasn't been touched.  It's been all about the record keeping.

HellAholic:
Sounds good. Let me know if you need help with something.

It's the same as making a car, you get the frame, engine, wheels, steering and you can drive it, but then comes things like the suspensions, windshield, breaks, lights, seat, and all the safety features.

GregValiant:
This doesn't work for me:
```
        user_container = None
        machine_name = global_stack.getName()
        try:
            user_container = global_stack.findContainer(criteria = ["id", f"{machine_name}_user"])
        except Exception as e:
            Logger.log("i", f"-----------The container is not in the stack.  Exception: {e}")
            # Get the user container to set the property
        Logger.log("i", f"-----------user_container = {user_container}")
```
2025-05-24 10:55:54,509 - INFO - [MainThread] PostProcessingPlugin.PostProcessingPlugin.PrintSkewCompensation.initialize [114]: -----------The container is not in the stack.  Exception: list indices must be integers or slices, not str
2025-05-24 10:55:54,524 - INFO - [MainThread] PostProcessingPlugin.PostProcessingPlugin.PrintSkewCompensation.initialize [116]: -----------user_container = None

GregValiant:
Thank you sir.  Much better.
(Things like that are why many people insist on a high degree of specificity.)

GregValiant:
Next show-stopper:
```
if user_container and isinstance(user_container, InstanceContainer):
   user_container.setProperty("xy_ac_measurement", "value", 141.42)
test_AC = user_container.getProperty("xy_ac_measurement", "value")
```

There is no error but "test_AC" comes back == None.
It appears that the getProperty / setProperty do not "createProperty"

HellAholic:
Could you share the script as a zip?

HellAholic:
[PrintSkewCompensation.zip](https://github.com/user-attachments/files/20427793/PrintSkewCompensation.zip)
Give this one a try, it can save to the user container but the issue is that the existing gcode gets overwritten and that is a no go. Struggle is real.

Did you have to pull out half the engine out to change the battery? "Modern car design"!

HellAholic:
[PrintSkewCompensation_settings_container_version.zip](https://github.com/user-attachments/files/20427842/PrintSkewCompensation_settings_container_version.zip)

GregValiant:
Chrysler made a couple of models that needed to be accessed through the wheelhouse so the wheel needed to be pulled to get the battery out.

This Acadia has:
1. A big junction box mounted on top of the battery.
2. 8 large conductors had to be removed.
3. 6 ridiculous plastic hold downs that required special tool B17 (or four screwdrivers used all-at-once to release the clips).
4. A simple but nearly inaccessible hold-down clamp on the lower edge.
5. A big hold-down bar across the top that required 15 minutes of staring at it before figuring out that another gizmo had to be unclipped before the hold-down bolt could be accessed.  It was still tough to get off.
6. A step ladder so the old man could lift the battery out of the tall vehicle.

But wait, there's more...
A trip to O'Reilly auto parts.
Carry the heavy battery into the store.
They confirmed that the battery had a dead cell.  "You bought this at Advance Auto Parts and it's less than four years old.  It should still be under warranty."
Carry the heavy battery back to the car.
A trip to Advance Auto Parts.
Carry the heavy battery into the store.
"That ain't one of ours.  You go it at AutoZone."
Carry the heavy battery back to the car.
A trip to AutoZone.
Carry the heavy battery into the store.
"Yep, you bought that here but that one had a 3 year warranty.  You want another one?"
FU
Carry the heavy battery back to the car.
A trip back to Advance Auto Parts.
Carry the heavy battery into the store.
$265 for a Die Hard battery.
Carry the new heavy battery back to the car.
Installation took about 15 minutes since I miraculously didn't break any of the cheesy plastic clips.
It runs.
All of this was done on a bright 94° 90% humidity Florida day.  I've just now dried out.

Oh yeah, code.
I anticipated the over-write thing.  I'll figure it out.
So far your script isn't working for me like it did for you.  It will be a bit before I dive into it.

Did you give up trying to create the settings in the container?

HellAholic:
Jesus....

So have two versions in the comments, both use containers, one is writing it to the _settings the other one to the _user
The _settings is where the normal start gcode lives so user can modify it.

GregValiant:
I guess it's working as designed.
I'll look at it more tomorrow, but I think I like it.  The settings are attached to the printer and won't be lost.

HellAholic:
Probably would need to double check the save to the gcode and reading back from gcode (method, measurements, and the rest) but general logic of it should be fine. Added a lot of error handling again to debug the code so they might need some trimming as well.
One major part is that I used the current version of the script that does not contain your changes to how the calculations are made, but that function should be isolated enough, didn't touch that part so your code should still work (the new version).

Check things out when you feel like it. I'll have to check things with the devs as well, I don't think the start gcode that the user changes on a machine should be in the settings stack but that's another discussion separate to this.

fieldOfView:
The more I read in to this, the more I think this should not be a postprocessing script. At least not one that ships with Cura.

Postprocessing scripts have no explanation, no documentation, or even a readme. Apparently this script requires models to be downloaded, and a user is going to have to come to that conclusion by reading the Python code of the actual script. Then it creates log files somewhere, which are not really log files but configuration files?

Make it a plugin. It is really not that hard; see the Z Offset plugin which is also a postporcessing script disguised as a plugin.  At least a plugin can have a readme, a github repository of its own, with an issue queue and a place to discuss how the hell to use this functionality.

fieldOfView:
> I agree that a plugin is a better way to go.

So, why isn't it then? Other than "waffling", you give no reason why it cannot be a plugin.

> Because it is running as a post-processor, the script adds the default settings (or existing settings) to the PP UI settings, and it alters the "machine_start_gcode" in the Machine Settings dialog

Wait, what?? It is a postprcoessing script. Why does it not just postprocess the existing start gcode on the fly?

This is a totally unintuitive mess. Postprocessing scripts should postprocess the gcode, not make settings changes! Removing a postprocessing script should remove any trace of the script on the next slice.

GregValiant:
There are currently 0 options for UM folks to pick from in regards to Print Skew Compensation.
1. "Plugin" is sitting on the sidelines ([here](https://github.com/HellAholic/PrintSkewCompensation))
2. "Post-process" is sitting on the sidelines here as a draft.

When the work on those is done (and HellAHolic and I agree that we can't do any more damage), then I anticipate that UM will have three options:
1. Go with the Plugin.
2. Go with the Post Processor
3. Throw them both in the friggin' toilet.

I'm leaning towards "Plugin" but I haven't fallen off the fence yet.  I am in favor of Cura "keeping up with the Joneses" and providing something similar in function to what the Slic3R clones have.
I'll put Aldo down as a definite "Plugin" with no waffling (BTW we're not talking Belgian Waffles here).
HellAHolic can cast his own vote.
I think Remco took a peek but I have no idea where he's standing on this.

HellAholic:
For reference, the plugin version on github:
https://github.com/HellAholic/CuraPrintSkewCompensation

GregValiant:
@HellAholic Can this be closed now?

PR #20618, opened 2025-05-22
Remove unused hack

ratijas:
# Description

Introduced back in 798c1f198c48c0f3f590bca64f254c776befc0da and unused since 5992e66ed670c038f3a05a58840321686076fdee, this is no longer an appropriate way to perform type checks in QML by modern Qt standards. Just use instanceof operator instead, if needed; it works since Qt 5.10

Source: https://doc.qt.io/qt-6/qtqml-javascript-hostenvironment.html

## Type of change

- [x] Chore/cleanup

# How Has This Been Tested?

- [ ] Search for the function name in all project files, find none.

# Checklist:

### DIFF
--- a/resources/qml/Cura.qml
+++ b/resources/qml/Cura.qml
@@ -942,21 +942,4 @@ UM.MainWindow
         okButtonText: catalog.i18nc("@button", "Save new profile")
         onAccepted: CuraApplication.getQualityManagementModel().createQualityChanges(newName, true);
     }
-
-    /**
-     * Function to check whether a QML object has a certain type.
-     * Taken from StackOverflow: https://stackoverflow.com/a/28384228 and
-     * adapted to our code style.
-     * Licensed under CC BY-SA 3.0.
-     * \param obj The QtObject to get the name of.
-     * \param class_name (str) The name of the class to check against. Has to be
-     * the QtObject class name, not the QML entity name.
-     */
-    function qmlTypeOf(obj, class_name)
-    {
-        //className plus "(" is the class instance without modification.
-        //className plus "_QML" is the class instance with user-defined properties.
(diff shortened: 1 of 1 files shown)


fieldOfView:
Strictly speaking, removing this function constitutes a major SDK version change, which is a pain because it means that all Marketplace plugins (whether they use the function or not) need to be changed and resubmitted.

Please see if there is a way to keep the function, but make it use the `instanceof` operator under the hood. Bonus points: have it emit a deprecation warning, though I don't remember if there's a generic way to do that from QML.

ratijas:
> Strictly speaking, removing this function constitutes a major SDK version change, which is a pain because it means that all Marketplace plugins (whether they use the function or not) need to be changed and resubmitted.

Oh, I was under an impression that it was some kind of entry point for the whole application like what would be `main.qml` in some other projects. Is it a piece of library with external API? (Irrelevant, but it was also literally just the first file I opened in the workspace while searching for .qml components.)

> Please see if there is a way to keep the function, but make it use the `instanceof` operator under the hood. Bonus points: have it emit a deprecation warning, though I don't remember if there's a generic way to do that from QML.

No can do. The function expects a string, but you can't just resolve a string back to a type that it represents. Unfortunately, as of Qt 6.10 QML still lacks any sort of dedicated deprecation syntax/API either, you'd just have to console.warn() or something.

fieldOfView:
Points taken. Then it is up to the Cura team to decide if this is minor enough to skip the SDK version update. To be honest, I don't know if any plugin is using the call from their QML.

ratijas:
Sooo… now what?

If it's gonna stuck forever, I'd rather just close and delete the branch honestly.

ratijas:
Polite ping?

PR #20640, opened 2025-05-28
[CURA-12543] Painting UI

rburema:
Tool (and view) to be used for paint-on-anything. Quite rudimentary for the moment, but should allow for smooth painting. Note that this doesn't do UV-unwrapping or assigning the textures properly yet, so for now you can only tests it on models that load properly with textures (so just OBJ with attached UV-coords for now). Needless to say, the paint doesn't _do_ anything yet. -- For all of those things, see the other branches that may now be floating around.

- Select brush-color, brush-size, brush-shape (paint-type is left for when we actually implement the different things you'd be able to do with this).
- Paint by selecting the model, selecting the paint tool (which will show the paint-view) then dragging over the model (where they have a texture).
- Undo/redo with (shift-)control-L (Z is taken for the overall undo).

Backend PR (needed for all of this): https://github.com/Ultimaker/Uranium/pull/999


plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+
+        self._last_text_coords: Optional[Tuple[int, int]] = None
+
+    def _createBrushPen(self) -> QPen:
+        pen = QPen()
+        pen.setWidth(self._brush_size)
+        color = self._color_str_to_rgba[self._brush_color]
+        pen.setColor(QColor(color[0], color[1], color[2], color[3]))
+        match self._brush_shape:
+            case "A":
```
wawanbreton: Could you use an enum for the brush shapes instead of letters ?

plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+
+        # compute unit vectors of directions of lines A and B
+        udir_a = a - pt
+        udir_a /= numpy.linalg.norm(udir_a)
+        udir_b = b - c
+        udir_b /= numpy.linalg.norm(udir_b)
+
+        # find unit direction vector for line C, which is perpendicular to lines A and B
+        udir_res = numpy.cross(udir_b, udir_a)
+        udir_res /= numpy.linalg.norm(udir_res)
```
wawanbreton: if `udir_a` and `udir_b` are colinear, `udir_res` is a null vector, so you get a division by 0, may it happen in real conditions ?

plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+                return True
+            return False
+
+        if event.type == Event.KeyReleaseEvent:
+            evt = cast(KeyEvent, event)
+            if evt.key == KeyEvent.ControlKey:
+                self._ctrl_held = False
+                return True
+            if evt.key == KeyEvent.ShiftKey:
+                self._shift_held = False
```
wawanbreton: You can also access the keyboard modifiers using `QGuiApplication::keyboardModifiers`, so maybe you don't have to handle `shift` and `ctrl` manually ?

plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+                return True
+            if evt.key == Qt.Key.Key_L and self._ctrl_held:
```
wawanbreton: You can also create a `QAction` and assign it a shortcut, so that you don't have to manually handle the pressed keys

plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+                self._last_text_coords[0] * w,
+                self._last_text_coords[1] * h,
+                texcoords[0] * w,
+                texcoords[1] * h
+            )
+            paintview.addStroke(sub_image, start_x, start_y)
+
+            self._last_text_coords = texcoords
+            Application.getInstance().getController().getScene().sceneChanged.emit(node)
+            return True
```
wawanbreton, line 338: For more clarity within this function, could you rename the various sub-casts of the event to more specific names, like `mouse_event` ?

plugins/PaintTool/PaintTool.py
```diff
@@ -0,0 +1,286 @@
...
+
+        return False
```
wawanbreton, line 340: Technically you should call `super().event(event)` here and return its result, but this may be inconsistent with the rest of the application.

plugins/SolidView/SolidView.py
```diff
@@ -36,11 +35,13 @@ class SolidView(View):
     """Standard view for mesh models."""
 
     _show_xray_warning_preference = "view/show_xray_warning"
+    _show_overhang_preference = "view/show_overhang"
+    _paint_active_preference = "view/paint_active"
```
wawanbreton: Seems to be unused, is it an aborted attempt ?

plugins/PaintTool/PaintTool.py
rburema: > may it happen in real conditions ?

That would mean that a parallel line-segment exists to a line-segment BC, starting in a point A and ending in a point within the triangle formed by ABC (the other cases are equivalent) -- I don't think that can happen without either;
- the triangle being degenerate -- which is something that wouldn't happen on a properly UV-mapped, manifold mesh (it'd also be hard to click, but I wouldn't count it our due to floating-point shenanigans) OR
- the point being clicked and A being one and the same -- which could happen I suppose

It may be a good idea to do something just in case.

plugins/SolidView/SolidView.py
rburema: Yeah, well, close enough to that -- it was sort of like the current attempt kept evolving to the point where I didn't need `SolidView` _at all_ anymore (which I think would eventually happen, maybe just not in this ticket already).

plugins/PaintTool/PaintTool.py
rburema: I don't dare change that at this juncture -- because that'd mean I'd have to do that for all those other instances as well 😅

plugins/PaintTool/PaintTool.py
rburema: done!

plugins/PaintTool/PaintTool.py
rburema: I did this, but then deleted it in the next commit because I resolved it with an Action, like you said :-p

PR #20680, opened 2025-06-11
CURA-12557 handle deactivated things

wawanbreton:
This PR makes the disabled printers and projects in Digital Factory more user-friendly in Cura:
* Display a specific warning icon when selecting an inactive printer (updated in real-time)
* Display a specific error message when sending a print job to an inactive printer
* Display inactive DL projects as disabled and show a tooltip to indicate the reason

:information_source: I am not very proud of the changes in `MachineManager`. The idea was to add specific attributes to `CloudOutputDevice` only, so that other classes are not disturbed, but as this class is in a plugin, I cannot call it directly from the `MachineManager` (or it would be even uglier). Suggestions are welcome !

CURA-12557


rburema:
>  The idea was to add specific attributes to CloudOutputDevice only, so that other classes are not disturbed, but as this class is in a plugin, I cannot call it directly from the MachineManager (or it would be even uglier). Suggestions are welcome !

What I'd probably do is think about how this notion of 'inactivity' can be generalized so that it'd be applicable to more situations (even if those situations don't exist at the moment) -- so that it's not specific to the whole 'cloud' situation, but can still be used by it.

In this case you might not even have to do much more than to erase the word 'cloud' from the variables of the non-plugin parts of this PR. -- Or have a 'reason' field (inactive = true, reason = 'cloud') if you want to make it 'fancy'.

PR #20693, opened 2025-06-15
Workflow fixes

HellAholic:
# Description

Adjustment to the workflow files:
- path -> paths: The correct keyword for filtering pull request triggers by file paths is paths (plural), not path (singular). 
- indentation fix: with keyword should be on the same indent level as uses.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/printer-linter-pr-diagnose.yml
+++ b/.github/workflows/printer-linter-pr-diagnose.yml
@@ -2,7 +2,7 @@ name: printer-linter-pr-diagnose
 
 on:
   pull_request:
-    path:
+    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/printer-linter-pr-diagnose.yml
```diff
@@ -2,7 +2,7 @@ name: printer-linter-pr-diagnose
 
 on:
   pull_request:
-    path:
+    paths:
```
wawanbreton, line 5: Good catch :smile:

PR #20694, opened 2025-06-15
Create slicing-error-check.yml

HellAholic:
# Description

Workflow that is triggered when an issue is opened or edited, if the issue has the label slicing error, it checks to see if a .zip file has been attached.
If it does not contain a .zip file, it will post a message and add the label needs info automatically.

## Type of change

### DIFF
--- a/.github/workflows/slicing-error-check.yml
+++ b/.github/workflows/slicing-error-check.yml
@@ -19,6 +19,7 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
+              const core = require('@actions/core');
               const issue = context.payload.issue;
               const issueNumber = issue.number;
 
@@ -45,7 +46,7 @@ jobs:
               } else {
                 console.log(`Issue #${issueNumber} does not match slicing error criteria. No action needed.`);
               }
-              console.log(`echo "needs_info=${setNeedsInfoOutput.toString()}" >> $GITHUB_OUTPUT`);
+              core.setOutput('needs_info', setNeedsInfoOutput.toString());
 
       - name: Add comment if project file is missing
         if: ${{ steps.check_issue_details.outputs.needs_info == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/slicing-error-check.yml
```diff
@@ -0,0 +1,69 @@
...
+  processSlicingError:
+    runs-on: ubuntu-latest
+    outputs:
+      needs_info: ${{ steps.check_issue_details.outputs.needs_info }}
+    steps:
+      - name: Check for project file and set output
+        id: check_issue_details
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
wawanbreton: If I'm correct, this line is not necessary (default behavior)

.github/workflows/slicing-error-check.yml
```diff
@@ -0,0 +1,69 @@
...
+    types: [opened, edited]
+
+permissions:
+    issues: write
+
+jobs:
+  processSlicingError:
+    runs-on: ubuntu-latest
+    outputs:
+      needs_info: ${{ steps.check_issue_details.outputs.needs_info }}
```
wawanbreton: Do you need this as a global output ? It is used between steps, but if not used by an external script, this can be removed.

wawanbreton (approved):
Cool addition :slightly_smiling_face: and looks good, just some suggestions

.github/workflows/slicing-error-check.yml
HellAholic: Should be fine without it, but was using an outdated method (echo) to set the output and was not sure if it functioned without it.
Using core.setOutput should be fine.

.github/workflows/slicing-error-check.yml
HellAholic: Removed. Can add back if necessary. :P

PR #20703, opened 2025-06-17
add Anycubic kobra 3 v2 (and ace pro) profiles

SamBkamp:
# Description

Added printer, extruder and build plate profiles/models for the anycubic Kobra 3 v2 (and ace pro)

## Type of change

# How Has This Been Tested?

- [ ] Printed single colour print successfully 
- [ ] Printed multi colour print (with ace pro) successfully 

## Important note
The printer will complain about missing gcode (error 10133) however this notification can safely be ignored. This notification also pops up when slicing using the Kobra 2 profile. Anycubic has not made their firmware entirely open-source so it is very hard to reverse engineer what exactly is missing. Comparing Cura-sliced gcode (using this profile) and anycubic slicer sliced gcode yields very similar results. 

When using the multicoloured printing mode (with the ace pro), it is important to keep the extruder standby temperature to be the same as your printing temperature, otherwise it will get to cold (as there is only one physical print head/nozzle). I have added definitions of this to the profile, However I think material profiles override this directive. 

**Test Configuration**:
* Operating System: Debian


resources/definitions/anycubic_kobra3v2.def.json
```diff
@@ -0,0 +1,60 @@
...
+    "version": 2,
+    "name": "Anycubic Kobra 3 v2",
+    "inherits": "fdmprinter",
+    "metadata":
+    {
+        "visible": true,
+        "author": "Sam Bonnekamp",
+        "manufacturer": "Anycubic",
+        "file_formats": "text/x-gcode",
+        "platform": "anycubic_kobra3v2_buildplate.STL",
```
HellAholic: Missing the platform file error (printer linter) due to `.STL` reference here vs `.stl` file that has been uploaded.

General point for all the files:
Avoid mixed capitalization for names or file formats, use lowercase as much as possible unless it's unavoidable. It's a cleaner code style and ensures issues like this don't happen.

resources/definitions/anycubic_kobra3v2.def.json
```diff
@@ -0,0 +1,60 @@
...
+            "default_value": "Anycubic Kobra 3 v2"
+        },	
+	"machine_buildplate_type": { "default_value": "PEI Spring Steel" },
+	"machine_heated_bed": { "default_value": true},
+	
+	"material_bed_temperature":
+	{	    "maximum_value": "110",
+	    "maximum_value_warning": "90"	    
+	},	
+        "default_material_print_temperature":
```
HellAholic: 'Overriding default_material_print_temperature with the same value (default_value: 210) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2.def.json
```diff
@@ -0,0 +1,60 @@
...
+	{
+	    "maximum_value": 300,
+	    "default_value": 210
+	},
+        "material_print_temperature":
+	{
+	    "maximum_value": 300,
+	    "default_value": 210
+	},
+        "material_print_temp_wait": { "value": true },
```
HellAholic: 'Overriding material_print_temp_wait with the same value (value: True) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2.def.json
```diff
@@ -0,0 +1,60 @@
...
+	"machine_center_is_zero": { "default_value": false },
+	"layer_height": { "default_value": 0.2 },
+        "speed_slowdown_layers": { "value": 2 },
```
HellAholic: 'Overriding speed_slowdown_layers with the same value (value: 2) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2_ACE_PRO.def.json
```diff
@@ -0,0 +1,67 @@
...
+        },
+        "machine_extruder_count": { "default_value": 4 },
+	"machine_buildplate_type": { "default_value": "PEI Spring Steel" },
+	"machine_heated_bed": { "default_value": true},
+	"material_bed_temperature":
+	{
+	    "maximum_value": "110",
+	    "maximum_value_warning": "90"	    
+	},	
+        "material_print_temp_wait": { "value": true },
```
HellAholic, line 38: 'Overriding material_print_temp_wait with the same value (value: True) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2_ACE_PRO.def.json
```diff
@@ -0,0 +1,67 @@
...
+	"machine_center_is_zero": { "default_value": false },
+	"layer_height": { "default_value": 0.2 },
+        "speed_slowdown_layers": { "value": 2 },
```
HellAholic: 'Overriding speed_slowdown_layers with the same value (value: 2) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2_ACE_PRO.def.json
```diff
@@ -0,0 +1,67 @@
...
+	"material_diameter": { "default_value": 1.75 },
+	"material_initial_print_temperature":
+        {
+            "maximum_value_warning": 295,
+            "value": "material_print_temperature + 5"
+        },	
+        "default_material_print_temperature":
```
HellAholic: 'Overriding default_material_print_temperature with the same value (default_value: 210) as defined in parent definition: fdmprinter'

resources/definitions/anycubic_kobra3v2_ACE_PRO.def.json
```diff
@@ -0,0 +1,67 @@
...
+	{
+	    "maximum_value": 300,
+	    "default_value": 210
+	},
+        "material_print_temperature":
```
HellAholic: 'Overriding material_print_temperature with the same value (default_value: 210) as defined in parent definition: fdmprinter'

resources/extruders/anycubic_kobra3v2_ACEPRO_extruder_0.def.json
```diff
@@ -0,0 +1,16 @@
+{
+    "version": 2,
+    "name": "ACE Pro Colour 1",
```
HellAholic: "name": "ACE Pro Colour 1",

Color vs Colour, not my battle to fight but Cura is based on the American English, so Color would be the preferred option for consistency.
Same point for the other extruders

HellAholic (changes requested):
Thanks for the PR.
Some adjustments/updates needed, listed as below. Nothing major or critical.
Please take a look.

HellAholic:
Regarding your notes, 

> The printer will complain about missing gcode (error 10133) however this notification can safely be ignored. This notification also pops up when slicing using the Kobra 2 profile. Anycubic has not made their firmware entirely open-source so it is very hard to reverse engineer what exactly is missing. Comparing Cura-sliced gcode (using this profile) and anycubic slicer sliced gcode yields very similar results.

the important notes section cannot be included with the printer definition for the users inside the application, so after slice they will encounter the error and refer to the website that indicates a faulty G-code generated. If the two generated Gcodes are similar, use a small cube (10mm) and slice it in both applications, then compare the results and see what the differences are. If that does not clarify the source, you could start removing the differences from the file that does not generate an error until the error appears.

>When using the multicoloured printing mode (with the ace pro), it is important to keep the extruder standby temperature to be the same as your printing temperature, otherwise it will get to cold (as there is only one physical print head/nozzle). I have added definitions of this to the profile, However I think material profiles override this directive.

This might indicate that you need to set `machine_nozzle_cool_down_speed` and `machine_nozzle_heat_up_speed` property. 
https://github.com/Ultimaker/Cura/blob/e627d09b1d7a055c8f0592a49a06d658c756bf82/resources/definitions/fdmprinter.def.json#L342C18-L362
You can use the fdmprinter definition as reference. There are also similar printer definitions for machines that only have a single physical extruder but have multiple virtual ones (latest additions are the bambu series https://github.com/Ultimaker/Cura/blob/main/resources/definitions/bambulab_base.def.json, https://github.com/Ultimaker/Cura/blob/main/resources/definitions/bambulab_x1.def.json)

Even with what is called a virtual extruder, you would go down to the standby temperature and heat back up to ensure there is sufficient time for the material to also reach print temperature to extrude properly and not cause under-extrusion.

SamBkamp:
Thanks so much for the concise suggested changes to the conf, clearly I'm still new to the Cura source code and have some stuff to learn. As for the virtual extruder, I assumed that the hotend would reheat to normal temperatures after the filament swap, however the display on my printer read 175c (the standby temperature). This printer also has some weird proprietary temperature setting gcode (G9111) and I wonder if those two are related. I will change the directives you've suggested and do some more testing.

I will push the simple adjustments now, and I will continue working on the printer error and multi-coloured print functionality and push those later. Hopefully by tonight.

SamBkamp:
a quick question; the unit test is failing because it says "relative_extrustion" being set to true is an unnecessary default value, but when I check the [fdmprinter file](https://github.com/Ultimaker/Cura/blob/e627d09b1d7a055c8f0592a49a06d658c756bf82/resources/definitions/fdmprinter.def.json#8278) it says its false by default. Am I missing something?

HellAholic:
The fail message is a bit unclear in this case, so what it means is because the parent definition has a "value" field, the "default_value" is ignored.

The parent (fdmprinter) has this:                     "value": "machine_gcode_flavor==\"RepRap (RepRap)\"",
https://github.com/Ultimaker/Cura/blob/e627d09b1d7a055c8f0592a49a06d658c756bf82/resources/definitions/fdmprinter.def.json#L8279C21-L8279C74

So if you just want to set it to true then you would need to override the value rather than default_value

SamBkamp:
Interesting. Im a little confused why "value" is a string when the "type" field says bool, is this because a string automatically evaluates to true?

HellAholic:
The content of the value is resolved, so looks like a string but it's a statement that results in a return (true / false).
So it's a resolveOrValue that either takes the value if it's a literal value like '50' or 'true' but also takes things like the one you see there (if statements, conditional statements, etc).

HellAholic:
It doesn't automatically resolve to true, the reason why the error is show is because you're setting the default_value to true, but then it's overridden with the formula of the value from the parent definition.
So in this case, it will be false regardless of what you set the default_value to, because the machine_gcode_flavor is not RepRap (RepRap).

if you need it to be enabled, then overriding the value would be the step to take rather than overriding the default_value.
default_value can be seen as a fallback, if no value is provided by any of the definitions or materials.

SamBkamp:
I figured out what was causing the error on the printer, the missing code is just the thumbnail. Considering Cura doesn't export screenshots yet. My question is: considering this will (hopefully) go in the main Cura release, what image should I use as the default thumbnail? Is there an "official" cura thumbnail or something?

HellAholic:
There is a post processing script in Cura (Create Thumbnail), could you check and see how it behaves if you use that post processing script to add a thumbnail to the gcode? you would have to go to extensions > post processing > and add the script manually.

- Just want to see if that resolves the issue with the error code (without the 32px cura thumbnail in start of gcode)
- If it does, how will it interact if you already have the 32px thumbnail there (can we leave the 32px cura thumbnail as a fallback if the user doesn't do the post processing?)

SamBkamp:
Just tested it, if the user doesn't use the post processing script it will fallback on the thumbnail in the start gcode section. If the user does include the post processing script it will use the thumbnail generated by that. Seems like everything is working now (including the weird problem with the multi-coloured printing standby temperature)

HellAholic (approved):
Looks good.
Thanks for the PR and the detective work :)
I hope you follow it up with some nice print profiles for the printer :P

PR #20712, opened 2025-06-19
Update normal.inst.cfg

HellAholic:
Quality does not have any values
The layer height defaults to 0.2 which makes it identical to other global quality profile for draft resources/quality/draft.inst.cfg

CURA-12627


wawanbreton:
Hmm, it is a very small change but has a large impact ! This is the default quality, right ? Maybe the draft quality should be changed instead ? Just asking questions :stuck_out_tongue:

HellAholic:
Comparing the list of available quality files, the missing layer height is the 0.1 (which also lines up with the definition of fine in UM printers).
The draft/Fast is the 0.2mm layer height also based on the UM profiles.
So I think the 0.1 is the correct value that fits the gap as we do have 0.06, xx , 0.15, 0.2, 0.3, ..., 0.6

wawanbreton:
It seems consistent indeed... But that means all the "normal" profile that are now in 0.2 will change to 0.1, right ? This has a big impact as all the print times will be 2 times longer...

HellAholic:
I'll do an assessment of impact, so we can leave this open for now until we have the data.

Frederic98:
Hmm... These qualities are for fdmprinter.def.json. There, the default layer height is 0.1mm. So adding it shouldn't have any impact (?) 🤔 
https://github.com/Ultimaker/Cura/blob/95f35275be464edf537a4abb76bd16c04f95cd60/resources/definitions/fdmprinter.def.json#L809

HellAholic:
It has an impact for printers that don't have quality profiles, but need to check the diff between the 0.2 fine and 0.2 draft. Technically it's a bug that the fine was resolving to 0.2mm but better double check.

HellAholic:
From the data we have, there is a long list of printers that are affected with this change, however, when we take a look at per printer usage of the fine 0.2, there is only 8 printers that have sliced with that profile.
From that, the top printer with the most number of slices for Fine 0.2 only has 1.7% of slices done in that profile in the past year.
I think this should be included as a highlight during the release notes but the overall impact at the moment is pretty contained.

wawanbreton:
> From the data we have, there is a long list of printers that are affected with this change, however, when we take a look at per printer usage of the fine 0.2, there is only 8 printers that have sliced with that profile. From that, the top printer with the most number of slices for Fine 0.2 only has 1.7% of slices done in that profile in the past year. I think this should be included as a highlight during the release notes but the overall impact at the moment is pretty contained.

Cool, thanks for checking ! Should we get an approval from @MariMakes, or maybe that was already discussed ?

wawanbreton (approved):

PR #20733, opened 2025-06-27
CC+ core

pkuiper-ultimaker:
Add CC+0.6 profiles
Add profiles for PC, CPE+ and Breakaway on AA+0.4.

PP-641


pkuiper-ultimaker:
Merged with https://github.com/Ultimaker/Cura/pull/20774

PR #20751, opened 2025-07-07
CURA-12408 separate alpha config folders

wawanbreton:
When using an alpha version, create a config folder specific to this version, so that non compatible configs won't mix

CURA-12408
Requires https://github.com/Ultimaker/Uranium/pull/1002


rburema (approved):
lgtm!

PR #20765, opened 2025-07-16
[CURA-12622] Warning on unused extruder.

rburema:
Since the bed-temp (for example) might be influenced by unused but (probably accidentally) enabled extruders, we want to be able to warn the user of such.

Note that it isn't necessary to take skirt/brim into account, since unused extruders (in the rest of the print) are already dropped from the those, even when the prime-tower is enabled.

Engine: https://github.com/Ultimaker/CuraEngine/pull/2240

### DIFF
--- a/plugins/CuraEngineBackend/CuraEngineBackend.py
+++ b/plugins/CuraEngineBackend/CuraEngineBackend.py
@@ -2,6 +2,7 @@
 #  Cura is released under the terms of the LGPLv3 or higher.
 
 import argparse #To run the engine in debug mode if the front-end is in debug mode.
+from cmath import isnan
 from collections import defaultdict
 import os
 from PyQt6.QtCore import QObject, QTimer, QUrl, pyqtSlot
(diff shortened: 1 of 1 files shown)


plugins/CuraEngineBackend/CuraEngineBackend.py
```diff
@@ -960,8 +961,26 @@ def _onPrintTimeMaterialEstimates(self, message: Arcus.PythonMessage) -> None:
...
+            else:
+                material_amounts.append(material_use_for_tool)
+
+        if no_use_warnings:
+            extruder_names = [self._global_container_stack.extruderList[int(idx)].definition.getName() for idx in no_use_warnings]
+            Message(
+                text=catalog.i18nc("@message", "At least one extruder that wasn't disabled, remains unused in this print (preview). "
+                                               f"Unused extruders: '<b>{", ".join(extruder_names)}</b>'. This can sometimes become a problem "
+                                               "(for example when the bed-temperature is adjusted by the material-profile present in the unused extruder). "
+                                               "It therefore might be desirable to disable these unused extruders manually, depending on the situation. "),
```
RKox91: [nitpick] may I suggest slightly altered version? I believe that this is a bit more clear about the potential impact on quality and reliability

```suggestion
                text=catalog.i18nc("@message", ""**Warning**: not all extruders that were left enabled are actually extruding in this printjob."  
f"Unused extruders: '<b>{", ".join(extruder_names)}</b>'." 
"This may lead to sub-optimal settings! To enjoy optimal printed part quality and reliability, please consider to disable the unused extruder(s)!"),
```

rburema: Thanks! I think that might indeed be more clear -- excepting the bit about 'sub-optimal settings', the sort of users that need this warning are just going to ignore that I feel, so a concrete example might help.

What do you think @HellAholic ?

HellAholic: Depending on the material in the unused extruder, this might actually need to be an error, not just a warning, since it can lead to print failures. There are a few scenarios we still need to consider, and the exact wording of the message isn’t final yet.

The main goal is to help users avoid failed prints, so the message should be clear and firm, not just a gentle suggestion. If a user chooses to ignore the warning or error, then they’re accepting the risk themselves.

One candidate message could be:

```
The following extruder(s) are enabled but not used in this print: {", ".join(extruder_names)}.

Unused extruders will affect the print, as their material profiles can still modify global settings such as bed temperature.
The impact may range from minor inconsistencies to serious print failures,
depending on the materials and configuration involved.

To avoid these issues, it is strongly recommended to disable any extruders not used in this print.
Continuing with them enabled means accepting responsibility for any resulting problems.
```

RKox91: I'm ok with making the wording more firm to convey the message better. The last sentence feels a bit too aggressive though. I would suggest to involve PM or UX to decide on the final wording.

I'm taking into account that the user might already be a bit frustrated that the warning message pops up only after slicing and not before, so I would like to avoid any aggressive tone in this message.

wawanbreton (approved):
Looks good to me, provided the wording has been sorted out :smile:

PR #20774, opened 2025-07-18
CC+0.6 profile improvements

pkuiper-ultimaker:
- Improve surface quality of PET-CF and PA-CF profiles
- Added back the 0.3mm profiles for CC+0.6
- Small bug fix: remove self reference to s8 quality profiles (enables the return of BASF metal profiles)

### DIFF
--- a/resources/definitions/ultimaker_s5.def.json
+++ b/resources/definitions/ultimaker_s5.def.json
@@ -72,7 +72,11 @@
         "brim_width": { "value": "3" },
         "build_volume_temperature": { "maximum_value": 50 },
         "cool_fan_speed": { "value": "50" },
-        "default_material_print_temperature": { "value": "200" },
+        "default_material_print_temperature":
+        {
+            "maximum_value_warning": "320",
+            "value": "200"
+        },
         "extruder_prime_pos_abs": { "default_value": true },
         "gantry_height": { "value": "55" },
         "infill_pattern": { "value": "'zigzag' if infill_sparse_density > 80 else 'triangles'" },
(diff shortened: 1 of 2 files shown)


pkuiper-ultimaker:
> On `main`, @Frederic98 changed the inheritance to s5, but this came along with some settings changes to ensure a consistent behavior, should I cherry-pick this commit: [a8ab9ec](https://github.com/Ultimaker/Cura/commit/a8ab9ecb75c957419a5d3025fbb8bf0f57dbed43)
> 
> Also, doing a machine inheritance change is quite critical. Are we sure that we don't have other settings that end up having different values ?

If you look at the S7 definition file (were it used to point to) you will see it inherits from the S5 and thet no (critical) overrides are present. So the risk is low.

pkuiper-ultimaker:
> On `main`, @Frederic98 changed the inheritance to s5, but this came along with some settings changes to ensure a consistent behavior, should I cherry-pick this commit: [a8ab9ec](https://github.com/Ultimaker/Cura/commit/a8ab9ecb75c957419a5d3025fbb8bf0f57dbed43)
> 
> Also, doing a machine inheritance change is quite critical. Are we sure that we don't have other settings that end up having different values ?

Yes please

wawanbreton:
> If you look at the S7 definition file (were it used to point to) you will see it inherits from the S5 and thet no (critical) overrides are present. So the risk is low.

Ok, didn't see that. So the list of potential changes is very limited and exhaustive indeed.

wawanbreton (approved):

PR #20791, opened 2025-07-23
[CURA-12624] New authentication workflow for new firmware.

rburema:
You first have to say ALLOW/DISSALOW on the printer now before you can view data with user-names in it. So deal with that in the Cura code here.

Should still work for printers on older firmware too.

P.S. There's two TODO's in the code which I should've picked up -- but the proposed class to move them to is _outside_ of the SDK (bundled plugin), while the class that I should cut them from is _inside_ (public methods in not-plugin).

P.P.S. I added a new commit with some defensive coding so the response to server-failure is less brittle -- but I couldn't test that one properly because I did it from home.

CURA-12624

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
+++ b/plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
@@ -125,7 +125,7 @@ class ClusterApiClient:
         """Delete a print job from the queue."""
 
         url = "{}/print_jobs/{}".format(self.CLUSTER_API_PREFIX, print_job_uuid)
-        self._manager.deleteResource(self.createEmptyRequest(url))
+        self._manager.deleteResource(self.createEmptyRequest(url, method=HttpRequestMethod.DELETE))
 
     def setPrintJobState(self, print_job_uuid: str, state: str) -> None:
         """Set the state of a print job."""


plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
```diff
@@ -74,52 +89,59 @@ def getPrintJobs(self, on_finished: Callable[[List[ClusterPrintJobStatus]], Any]
...
     def getPrintJobPreviewImage(self, print_job_uuid: str, on_finished: Callable) -> None:
         """Get the preview image data of a print job."""
 
         url = "{}/print_jobs/{}/preview_image".format(self.CLUSTER_API_PREFIX, print_job_uuid)
-        reply = self._manager.get(self._createEmptyRequest(url))
+        reply = self._manager.get(self.createEmptyRequest(url))
         self._addCallback(reply, on_finished)
 
-    def _createEmptyRequest(self, path: str, content_type: Optional[str] = "application/json") -> QNetworkRequest:
+    def createEmptyRequest(self, path: str, content_type: Optional[str] = "application/json", method: str = "GET", skip_auth: bool = False) -> QNetworkRequest:
```
wawanbreton: Maybe make the method an enum, since it can only have 3 values ? You do use it as a string in the end though, so I would understand that this is annoying.

wawanbreton (approved):
Looks good to me, the TODOs could have been done at the same time indeed, but I don't see a strong need for them.

plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
rburema: It did indeed turn out that Python just autoboxes `StrEnums`, so; fixed! 😄

wawanbreton: Cool, love it :smile:

plugins/UM3NetworkPrinting/src/Network/SendMaterialJob.py
```diff
@@ -147,7 +147,8 @@ def _sendMaterialFile(self, file_path: str, file_name: str, material_id: str) ->
 
         # FIXME: move form posting to API client
         self.device.postFormWithParts(target = "/cluster-api/v1/materials/", parts = parts,
-                                      on_finished = self._sendingFinished)
+                                      on_finished = self._sendingFinished,
+                                      request=self._cluster_api.createEmptyRequest("/cluster-api/v1/materials/", content_type=None, method="POST"))
```
Frederic98: `AttributeError: 'SendMaterialJob' object has no attribute '_cluster_api'`

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
```diff
@@ -204,7 +204,8 @@ def _startPrintJobUpload(self, unique_name: str = None) -> None:
             parts.append(self._createFormPart("name=require_printer_name", bytes(unique_name, "utf-8"), "text/plain"))
         # FIXME: move form posting to API client
         self.postFormWithParts("/cluster-api/v1/print_jobs/", parts, on_finished=self._onPrintUploadCompleted,
-                               on_progress=self._onPrintJobUploadProgress)
+                               on_progress=self._onPrintJobUploadProgress,
+                               request=self._cluster_api.createEmptyRequest("/cluster-api/v1/print_jobs/", content_type=None, method="POST"))
```
Frederic98: Should use `self._getApiClient()` instead of `self._cluster_api` just to be safe

Frederic98:
I've made `LocalClusterOutputDevice._getApiClient()` public, so that `SendMaterialJob` can access it.

plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
```diff
@@ -158,6 +193,67 @@ def _parseModels(self, response: Dict[str, Any], on_finished: Union[Callable[[Cl
...
+
+    def _setupAuth(self) -> None:
+        """ Handles the setup process for authentication by making a temporary digest-token request to the printer API.
+        """
+
+        if self._auth_tries >= ClusterApiClient.AUTH_MAX_TRIES:
+            Logger.warning("Maximum authorization temporary digest-token request tries exceeded. Is printer-firmware up to date?")
+            return
+
+        username = CuraApplication.getInstance().getCuraAPI().account.userName
```
Frederic98: Shouldn't rely on a cloud connection (just to get a username string) to connect to a local printer over LAN. Maybe get the username from the OS instead (like `os.getlogin()` or `getpass.getuser()`)

rburema: You'll get a 'blank' username when not logged in. I'm a bit hesitant to pull in more information, even though it's probably alright.

plugins/UM3NetworkPrinting/src/Network/SendMaterialJob.py
rburema: Thanks, will change.

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
rburema: Agreed, will change.

plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
rburema: ... or it used to in the prototype? I think I changed that.

rburema:
Superseded by #20940

PR #20821, opened 2025-07-30
[CURA-12630] Use newly exposed 'reloadNodes' so we can use Uraniums version of that.

rburema:
This (previously) caused issues where the code deleted in this ticket was almost the same as in Uranium, except it was slightly buggy (but it also did slightly more, hence the new `on_done` parameter).

See also: https://github.com/Ultimaker/Uranium/pull/1006


wawanbreton (approved):
Less code, less bugs 😄

PR #20840, opened 2025-08-04
Cura 12660 painting UI improvements

wawanbreton:
This PR provides many changes in the user interactions of the paint mode, to make it more user-friendly. The changes are quite different from what we discussed, but I think it achieves the same goal.

CURA-12660
Requires https://github.com/Ultimaker/Uranium/pull/1007

### DIFF
--- a/plugins/PaintTool/PaintTool.py
+++ b/plugins/PaintTool/PaintTool.py
@@ -12,6 +12,7 @@ from numpy import ndarray
 
 from UM.Application import Application
 from UM.Event import Event, MouseEvent, KeyEvent
+from UM.Job import Job
 from UM.Logger import Logger
 from UM.Scene.SceneNode import SceneNode
 from UM.Scene.Selection import Selection
@@ -22,6 +23,7 @@ from cura.CuraApplication import CuraApplication
 from cura.PickingPass import PickingPass
 from UM.View.SelectionPass import SelectionPass
 from .PaintView import PaintView
+from .PrepareTextureJob import PrepareTextureJob
 
 
 class PaintTool(Tool):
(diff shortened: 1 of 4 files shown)


plugins/PaintTool/paint.shader
```diff
@@ -55,10 +54,10 @@ fragment =
         color_index = (color_index << (32 - 1 - u_bitsRangesEnd)) >> 32 - 1 - (u_bitsRangesEnd - u_bitsRangesStart);
 
         vec4 diffuse_color = vec4(u_renderColors[color_index] / 255.0, 1.0);
-        highp float n_dot_l = clamp(dot(normal, light_dir), 0.0, 1.0);
+        highp float n_dot_l = mix(0.3, 0.7, dot(normal, light_dir));
```
rburema, line 57: I've noticed mirrored objects become quite dark -- it's not that important right now but I'll probably start looking here.

plugins/PaintTool/paint.shader
```diff
@@ -122,17 +120,16 @@ fragment41core =
...
```
rburema, line 123: ... and here :-p

plugins/PaintTool/PaintTool.py
```diff
@@ -343,9 +391,46 @@ def event(self, event: Event) -> bool:
...
\ No newline at end of file
+            Application.getInstance().getController().getScene().sceneChanged.emit(node)
+
+    def _onSelectionChanged(self):
+        super()._onSelectionChanged()
+
+        self.setActiveView("PaintTool" if len(Selection.getAllSelectedObjects()) == 1 else None)
+        self._updateState()
+
+    def _updateState(self):
```
rburema, line 410: I think it'd be better to add a word or two on _what_ state this is. Either in documentation or through renaming the method.

rburema (approved):
LGTM -- I'm altering some things in the Uranium counterpart of this though.

PR #20842, opened 2025-08-04
CURA-12661 make more global paint undo

wawanbreton:
Implements a full-stroke undo-redo instead of each stroke atomic operation, which is more user-expected.

This uses a QUndoStack instead of the previous custom-made mechanism, so that we can use common undo-redo mechanisms.

:warning:  Based on branch CURA-12660_painting-UI-improvements
CURA-12661

### DIFF
--- a/plugins/PaintTool/PaintUndoCommand.py
+++ b/plugins/PaintTool/PaintUndoCommand.py
@@ -40,7 +40,7 @@ class PaintUndoCommand(QUndoCommand):
 
         bit_range_start, bit_range_end = self._bit_range
         full_int32 = 0xffffffff
-        clear_mask = full_int32 ^ (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
+        clear_texture_bit_mask = full_int32 ^ (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
                     32 - 1 - bit_range_end))
         image_rect = QRect(0, 0, self._stroke_mask.width(), self._stroke_mask.height())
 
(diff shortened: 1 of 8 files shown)


plugins/PaintTool/PaintUndoCommand.py
```diff
@@ -0,0 +1,104 @@
...
+    def id(self) -> int:
+        # Since the undo stack will contain only commands of this type, we can use a fixed ID
+        return 0
+
+    def redo(self) -> None:
+        actual_image = self._texture.getImage()
+
+        bit_range_start, bit_range_end = self._bit_range
+        full_int32 = 0xffffffff
+        clear_mask = full_int32 ^ (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
```
casperlamboo: I find the variable name a bit misleading. When I read `clear_mask` I'd expect a bit-mask for the clear color. However, when reading the code this is the the bit-mask for the textures (if I read the file correctly). I would instead call this variable something like `clear_texture_bit_mask`.

rburema (commented):
The code is OK and it probably works fine (yet to check) -- I just don't quite understand why you can't just bunch up (or set a bookmark into) the already exiting undo/redo stacks for each 'real' action the user makes.

It would be way easier to implement and you'd keep the underlying 'detailed' undo for when that's necessary. Maybe adding the QUndoStack prohibited this?

Anyway, if this works I'm mostly keeping this PR as-is regardless (minus fixing the variable-name Casper pointed out) -- it's easier for us to go with the flow and merge this now (and there's nothing really wrong with it).

plugins/PaintTool/PaintUndoCommand.py
```diff
@@ -40,15 +40,15 @@ def redo(self) -> None:
 
         bit_range_start, bit_range_end = self._bit_range
         full_int32 = 0xffffffff
-        clear_mask = full_int32 ^ (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
+        clear_texture_bit_mask = full_int32 ^ (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
```
rburema, line 43: Whoops, forgot to change that -- thanks! 😅

PR #20851, opened 2025-08-08
CURA-12449 handling painted models map

wawanbreton:
This PR adds:
* Some UI elements to properly display and interact with multi-material painting
* Settings for the multi-material splitting

Requires https://github.com/Ultimaker/CuraEngine/pull/2243
CURA-12528


resources/definitions/fdmprinter.def.json
```diff
@@ -7915,6 +7915,35 @@
...
                     "settable_per_extruder": false
+                },
+                "multi_material_paint_resolution":
+                {
+                    "label": "Multi-material Precision",
+                    "description": "The precision of the details when generating multi-material shapes based on painting data. A lower precision will provide more details, but increase the slicing time and memory.",
+                    "unit": "mm",
+                    "type": "float",
+                    "enabled": "extruders_enabled_count > 1",
+                    "default_value": "0.2",
```
casperlamboo: Default value should be without quotes, otherwise it's interpreted as a string
```suggestion
                    "default_value": 0.2,
```

resources/definitions/fdmprinter.def.json
```diff
@@ -7915,6 +7915,35 @@
...
+                    "settable_per_extruder": false
+                },
+                "multi_material_paint_deepness":
+                {
+                    "label": "Multi-material Deepness",
+                    "description": "The deepness of the painted details inside the model. A higher deepness will provide a better interlocking, but increase slicing time and memory. Set a very high value to go as deep as possible. The actually calculated deepness may vary.",
+                    "unit": "mm",
+                    "type": "float",
+                    "enabled": "extruders_enabled_count > 1",
+                    "default_value": "4",
```
casperlamboo: Default value should be without quotes, otherwise it's interpreted as a string
```suggestion
                    "default_value": 4,
```

plugins/PaintTool/PaintTool.py
```diff
@@ -176,7 +185,7 @@ def clear(self) -> None:
         width, height = self._view.getUvTexDimensions()
         clear_image = QImage(width, height, QImage.Format.Format_RGB32)
         clear_image.fill(Qt.GlobalColor.white)
-        self._view.addStroke(clear_image, 0, 0, "none", False)
+        self._view.addStroke(clear_image, 0, 0, "none" if self.getPaintType() != "extruder" else "0", False)
```
rburema, line 188: ... hang on, so we have both `"none"` and `"0"` here. -- Ok it took me a while to get this; one is the 'none' of 'none', 'allow', 'disallow' and the other is the '0' of the 1st extruder.

I think maybe a paint-type should have an associated default 'null-ish' value, so it can just be set here.

rburema (approved):
I'm not 100% sold on the split between brush-extruder and brush-color (2 variables for what should be the same thing, one of which is only used in a specific circumstance), but should we ever want to refactor that it's not that hard.

plugins/PaintTool/PaintTool.py
wawanbreton: Well, "none" and "0" really have different meanings, so I wouldn't be comfortable with merging them into a single thing. Maybe we could find a better name instead of "none" though ? Would that help ?

wawanbreton:
> I'm not 100% sold on the split between brush-extruder and brush-color (2 variables for what should be the same thing, one of which is only used in a specific circumstance), but should we ever want to refactor that it's not that hard.

Maybe when we get more painting types. As long as we have only 2, it keeps manageable IMHO, but starting with 3 I agree that we would have to do something cleaner.

plugins/PaintTool/PaintTool.py
rburema: Mostly it just feels a bit odd to me to check the paint-type here as well -- and then _only_ this one. It's fine as is, but it'd be better if a paint-type could just define an 'empty'/'clear' type, so we can say `self.getPaintClearValue()` here or something like that.

Issue #20878, opened 2025-08-16
due to slicing error occured


sanketmomte2003-collab:
### Cura Version

5.10.2

### Operating System

windows 11

### Printer

ender 3 

### Describe your problem and add the package for technical support as a .zip here ⬇️

just small part of .stl file not responding for slicing

sanketmomte2003-collab:
[export-20250818-152207 (3).zip](https://github.com/user-attachments/files/21832769/export-20250818-152207.3.zip)

sanketmomte2003-collab:
give me the solution fast

GregValiant:
Setting the "Inside Travel Avoid Distance" to 0.60 (the default) allows it to slice.
The model does have errors and needs repair.

<img width="804" height="630" alt="Image" src="https://github.com/user-attachments/assets/4d5c4c9e-c32e-46da-a953-4a9d95beabcd" />

@HellAholic this does appear to be a setting bug.  It doesn't seem to like a "0" value.  I'm removing the "slicing error" tag.

HellAholic:
Thanks @GregValiant 
I can set the minimum value to "0.001" or something similar which blocks slicing for when it's set to 0.
I'll discuss the change internally.

PR #20880, opened 2025-08-18
Set minimum value for Inside Travel Avoid distance to nozzile size * 0.5

HellAholic:
Minimum value should be the same as minimum value warning since any lower value will cause collision with the printed part

Prevents: https://github.com/Ultimaker/Cura/issues/20878

CURA-12687


wawanbreton (approved):
Makes sense to me, did you do a sanity check with PPM ?

PR #20882, opened 2025-08-19
[CURA-12662] fix paint-splatter issue

rburema:
Previous solutions weren't taking the disconnected nature of UV-maps into account nearly enough. This meant that thick brush-strokes could (in UV-space) overlap with areas that where nowhere near the actual stroke in 3D.

Instead (after reworking the backend so we can look at face-connectivity; see [**here**](https://github.com/Ultimaker/Uranium/pull/1008)), intersect each triangle with the brush stroke (in camera-related space), then render that to the texture.

This unfortunately does make it a bit slower, so we have to work in some optimizations.

CURA-12662


plugins/PaintTool/PaintTool.py
```diff
@@ -182,104 +209,116 @@ def clear(self) -> None:
...
+            v /= total
+            w /= total
+
+        return u * a2 + v * b2 + w * c2
+
+    def _getStrokePolygon(self, size_adjust: float, stroke_a: numpy.ndarray, stroke_b: numpy.ndarray) -> Polygon:
+        shape = None
+        match self._brush_shape:
+            case PaintTool.Brush.Shape.SQUARE:
+                shape = Polygon.approximatedCircle(self._brush_size * size_adjust, 4)
```
wawanbreton, line 252: This might be wrong 😃

plugins/PaintTool/PaintTool.py
```diff
@@ -182,104 +209,116 @@ def clear(self) -> None:
...
-        if (abs(click_a[0] - click_b[0]) < 0.0001 and abs(click_a[1] - click_b[1]) < 0.0001) or (face_a < 0 and face_b < 0):
-            return
-        if face_b < 0 or face_a == face_b:
-            substrokes.append((self._last_text_coords, texcoords_a))
-            return
-        if face_a < 0:
-            substrokes.append((self._last_text_coords, texcoords_b))
-            return
+    @staticmethod
+    def _remapBarycentric(triangle_a: Polygon, pt: numpy.ndarray, triangle_b: Polygon) -> numpy.ndarray:
```
wawanbreton: I do not understand what this function does, it seems to return a point that is equal to pt 🤔

plugins/PaintTool/PaintTool.py
```diff
@@ -341,8 +379,9 @@ def event(self, event: Event) -> bool:
                 if not self._picking_pass:
                     return False
 
-            camera = self._controller.getScene().getActiveCamera()
-            if not camera:
+            if self._camera is None:
+                self._updateCamera()
```
wawanbreton, line 405: Maybe this could be done only when the camera position changes ?

wawanbreton (commented):
I have mostly questions about what some functions do, but the code looks good.
We can talk about the possible optimizations when we actually meet.
On the other hand, I can't get it to work properly with a cube on Linux, the preview disc works pretty well, but the painting barely works 🤔

plugins/PaintTool/PaintTool.py
rburema: We already spoke in-person, but for future reference and anyone else reading this; the comment was more about the single remaining _use_ of the function rather than an unfamiliarity with Barycentric coordinates -- I'm currently using this to remap screen-space coordinates (`triangle_a`, `pt`) to UV-space coordinates (`triangle_b`, output of the function).

plugins/PaintTool/PaintTool.py
rburema: Read the 'if' again; this is _only_ done when the camera is null 😉.

wawanbreton: Oh right :sweat_smile:

Issue #20898, opened 2025-08-27
Printbed goes down too far


Eliteronin319:
### Cura Version

5.9.1

### Operating System

windows 10

### Printer

Tronxy x5sa pro

### Reproduction steps

1) I canceled any print (happens every time atm, wasn't an issue some weeks ago. Someone on a forum mentioned some probable failure in the gcode. So im sending the actual g-code from cura itself and the file after it was sliced using an Editor.)

![Image](https://github.com/user-attachments/assets/f6dc288e-375d-4183-81b7-9e0b2d71420d)
![Image](https://github.com/user-attachments/assets/bf893779-85bb-4b93-9fde-420fab5fe013)

### Actual results

The print bed went down to the bottom of the printer and then tried to go further, resulting in a screeching sound. 
I couldn't move the bed until it happened.

### Expected results

The printer should just cancel my print

### Add your .zip and screenshots here ⬇️

![Image](https://github.com/user-attachments/assets/d0961f42-db15-4f62-8b49-0e425f6ee881)
![Image](https://github.com/user-attachments/assets/8546fdb8-22f8-43fd-bf81-edf07cd01c10)

GregValiant:
Thanks for the report.
I saw your post on the Forum.  I don't see a problem in the Gcode.  The "Ending Gcode" might be a problem but that comes from the printer definition rather than from Cura.  The Cura portion of the Gcode ends with the Z at whatever the Z height of the final layer was.

The author of the TronXY definition files appears to be ""KV/AdderMk2" who was either contracted with by TronXY or works for TronXY.  UM doesn't (can't) get involved in other manufacturers printer definition files.

As PartySausage and Slashee pointed out on the Forum, "two commands on one line" is a problem.  There may also be something going on when the M84 line tells the steppers to shut off.  Some printers have been known to travel down all by themselves if the Z motor is disabled.

I've gone over the stock Ending Gcode and made some changes.
Copy and paste this into the "Ending G-Code" box in the Machine Settings in Cura.

```
;----------Ending Gcode
G92 E0 ; Reset Extruder
G91 ; Set Positioning to Relative
M83 ; Set Extruder to Relative
G1 E-{retraction_amount} F{retraction_retract_speed*60} ; Retract
G1 F{speed_z_hop*60} Z0.2 ; Raise nozzle .2mm
G1 F{speed_travel*60} X5 Y5 ; Wipe out
G1 F{speed_z_hop*60} Z10 ; Raise nozzle 10mm
[...]

PR #20903, opened 2025-08-29
Added missing \n before the G1 Z10

HellAholic:
# Description

Fix the typo in the end gcode;
missing \n that would make the command that would set the Z to a comment

## Type of change

- [X] Printer definition file(s)

https://github.com/Ultimaker/Cura/issues/20898


saumyaj3:
@HellAholic should this be based on 5.11?

PR #20908, opened 2025-09-02
PP-655: S8 legacy cores

Frederic98:
# Description
Merge together with Ultimaker/fdm_materials#366

Except for some explainable differences, a flatpack of the S5 results in the same settings as a flatpack of the S8 legacy core profiles. The differences are:
- `machine_name`, `machine_start_gcode`, `machine_gcode_flavor`, ...
- 2 settings that are not settable per extruder (and no resolve): `cool_during_extruder_switch`, `prime_tower_mode`
- Some small differences that are caused by the layer height being scaled with the material Z shrinkage factor on the S8, but not on the S5.
- Differences of <= 0.1% are ignored

See also Ultimaker/autogenerate_print_profiles#296

## Type of change

- [ ] Translations

# How Has This Been Tested?

- Ran test print for PLA 0.2mm Balanced on AA 0.4
- Compared flatpacks between S5 and S8

# Checklist:

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -8808,7 +8808,7 @@
                     "value": "line_width + support_xy_distance + 1.0",
                     "enabled": "bridge_settings_enabled",
                     "settable_per_mesh": true,
-                    "settable_per_extruder": false
+                    "settable_per_extruder": true
                 },
                 "bridge_skin_support_threshold":
                 {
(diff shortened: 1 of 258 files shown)


wawanbreton:
Hmm, actually it seems that all the other settings that define `"settable_per_mesh": true` don't define `settable_per_extruder` at all, so maybe we should just do the same :thinking: 
@bremco can we have your opinion on this ?

rburema:
The `settable_per_mesh` and `settable_per_extruder` (mostly) denote engine limitations (and in the cases where it does not, other, similar, limitations), and should probably only be set in fdmprinter or fdmextruder -- unless you make them _more_ strict and set them to `false`, it's not going to magically be able to make the engine do those things.

So if the engine is indeed capable of setting these things per mesh or extruder, go ahead and do so, at a high level (or set to false if you want to restrict them -- but then someone should test if that actually works).

As for settings defining them for both values; I think I've seen that in the past. It can't hurt to do that, so if we want to be certain, then do so.

PR #20909, opened 2025-09-02
Move the bridging settings from "Experimental" to "Support"

pkuiper-ultimaker:
Move the bridging settings from "Experimental" to "Support" and enabled the visibility of the settings in advance and expert visibility settings mode. Also fixed some visibility settings that were missing in the advanced visibility settings (pressure advance and min. layer time on overhangs)

PP-150


wawanbreton:
I would also suggest to rename the main setting "Enable Bridging" instead of "Enable Bridge Settings"

pkuiper-ultimaker:
> I would also suggest to rename the main setting "Enable Bridging" instead of "Enable Bridge Settings"

Done

PR #20924, opened 2025-09-05
Improve time estimate replicator plus

pkuiper-ultimaker:
PP-674


wawanbreton:
@pkuiper-ultimaker I did see some settings that were overridden with the same values as their parents, and then I realised that the printer-linter was disabled, and also had a bug... So now I fixed it, and there are lots of comments, sorry :sweat:

HellAholic:
It seems due to an issue in the time estimation, these unnecessary overrides are necessary for accurate time estimation.
So we decided to include the PR as is for now until we've addressed the core issue on the estimation logic.

PR #20928, opened 2025-09-06
Cura 12701 user estimation factor

HellAholic:
# Description

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:

# Checklist:

### DIFF
--- a/cura/UI/PrintInformation.py
+++ b/cura/UI/PrintInformation.py
@@ -166,6 +166,15 @@ class PrintInformation(QObject):
     def printTimes(self) -> Dict[str, Duration]:
         return self._print_times_per_feature[self._active_build_plate]
 
+    def _getTimeEstimationFactor(self) -> Optional[float]:
+        """Called when the global container stack changes to emit timeEstimationFactorChanged signal"""
+        global_stack = self._application.getGlobalContainerStack()
+        
+        if global_stack is None:
+            return None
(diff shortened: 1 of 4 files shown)


rburema:
I think I'd personally just set it up per-printer in the settings, that way, we can communicate it to the engine & that can use the 'proper' time-estimates, which would then also work with c. cloud.

fieldOfView:
Looks like the field is in the wrong column; I don't think this has anything to do with the printhead (right column) and should instead be in the left column (printer)

GregValiant:
I thought this would have been merged.  Is it just sitting on a shelf someplace (next to the Ark of the Covenant)?

rburema:
Read through the code again, and I'd be OK with this @HellAholic how do we go about merging this, since you're meant to do that now :-p

(No, I think the Ark is next to the missing tablet with commands 11 through 15, somewhere in sector plural Z alpha.)

HellAholic:
Switching personality matrix. Initiate up is down protocol. Keeping an eye out for:
- The Cherubim with the bad attitude
- The hovering “KEEP OUT” hologram
- And the temporal distortion that makes lunch take four hours

Since it has a conan package, I can follow the same flow we have for other tickets. It's been quite a while since I poked at this so technically:
"I have no memory of this place"

PR #20935, opened 2025-09-08
Z seam middle option

HellAholic:
# Description

- Adds an option for Z Seam at middle
- Adds an option to set the layer start X and Y to the z seam position (only available on user defined z seam position)
- I blame Greg

## Type of change

# How Has This Been Tested?

- [x] Slice and preview on machine with origin at center with different options
- [x] Slice and preview on machine with origin at front-left with different options
Machine center is zero (origin at center)
A) Z seam relative
B) Z seam absolute 
Machine center is not zero (origin at front left)
A) Z seam relative
B) Z seam absolute

**Test Configuration**:

# Checklist:

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -1392,7 +1392,7 @@
                 "z_seam_position":
                 {
                     "label": "Z Seam Position",
-                    "description": "The position near where to start printing each part in a layer.",
+                    "description": "The position near where to start printing each part in a layer. Note: For the 'Middle' option, the reference point differs critically based on Z Seam Relative setting - it will be the middle of the buildplate (absolute) or the middle of each object (relative), resulting in completely different seam placements.",
                     "type": "enum",
                     "options":
                     {
@@ -1415,7 +1415,7 @@
                         "z_seam_x":
                         {
                             "label": "Z Seam X",
-                            "description": "The X coordinate of the position near where to start printing each part in a layer.",
+                            "description": "The X coordinate of the position near where to start printing each part in a layer. When Z Seam Relative is disabled, this is an absolute position on the buildplate. When enabled, this is relative to each object's center.",
                             "unit": "mm",
                             "type": "float",
                             "default_value": 100.0,
@@ -1427,7 +1427,7 @@
                         "z_seam_y":
                         {
                             "label": "Z Seam Y",
-                            "description": "The Y coordinate of the position near where to start printing each part in a layer.",
+                            "description": "The Y coordinate of the position near where to start printing each part in a layer. When Z Seam Relative is disabled, this is an absolute position on the buildplate. When enabled, this is relative to each object's center.",
                             "unit": "mm",
                             "type": "float",
                             "default_value": 100.0,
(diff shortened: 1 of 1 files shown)


GregValiant:
**"I blame Greg"**
I have large shoulders that can carry the weight.
I have bad knees that cannot.

Now horseshoe shapes can come out with the Z-seam on the inside.
image
(They always could if the user knew to change the numbers.)

resources/definitions/fdmprinter.def.json
```diff
@@ -4766,14 +4767,25 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "layer_start_at_z_seam":
+                {
+                    "label": "Layer Start at Z Seam",
+                    "description": "Start printing at the Z seam position.",
+                    "type": "bool",
+                    "default_value": false,
+                    "enabled": "z_seam_type == 'back'",
```
rburema, line 4881: What about machines that don't start from the back?

HellAholic: it's the z_seam_type which I was also confused by the naming, this is actually `"Z Seam Alignment": User specified` seam type. 😕 
https://github.com/Ultimaker/Cura/blob/1508c6160dd6520cc5cbc9441786aa72ca9060fb/resources/definitions/fdmprinter.def.json#L1374

The idea being, if the user updates the z seam position, the layer start also updates based on the change in the user specified section.
We can't know for sure where that would be from the front end in the other options (sharpest corner / Random / ...)

rburema: I think I saw that before at one point. I must've forced myself to forget it, given how much sense it makes 😆

rburema:
Another thing; for the other options it doesn't really matter whether it's 'from the object' versus 'from the build-plate', but for 'middle' it does matter; placing the object on another spot can result in a completely different seam placement. (I think there are ways that can still happen with the other ones as well, but not as prominently.)

I don't think that should hold up this PR at all though, just something to think about.

GregValiant:
In regards to "Middle" and _"...placing the object on another spot can result in a completely different seam placement."_

I had considered that.  I decided it was a good feature (rather than an annoyance) as I had to use that a couple of times to get what I wanted from the model.

rburema:
@GregValiant I think I agree, but I was thinking more w.r.t. communication to the user to prevent confusion.

HellAholic:
Added a commit with extra explanation to the description of the settings to clarify hopefully 😝

PartySausage:
I've been taking a look at this tonight & something a bit weird is happening with the Middle X Y calculation. As on my E3 S1 PRO 220 x 220 bed I'm getting the following.. When I was expecting 110,110 for absolute coordinates & 0, 0 for relative coordinates

image

&

image

HellAholic:
image
image

PartySausage:
I think I've got to the bottom of it. 

I found that the formula used in the fdmprinter.def.json for calculating the z_seam_x & z_seam_y is registered in all my quality profiles and wasn't being updated with the new formula to account for the new 'middle' option. Once I deleted those lines the z_seam_x & z_seam_y are now calculated correctly

I don't know why these were in the quality profiles & there are many other [values] in them, which I can only presume are overrides on the base profile so I guess anything that contains a formula that gets changed as part of a new Cura version might not be reflected in the quality profile when used.

PR #20950, opened 2025-09-11
[CURA-12333] Disable USB-cable-printing by default (except for people who used it already).

rburema:
We set the preference value for people who used USB-printing in the previous version, from the next version onwards, it'll be disabled by default -- as not many people still print by USB-cable, and more importantly, it can mess up some systems. It's not going away at the moment though.


resources/qml/Preferences/GeneralPage.qml
```diff
@@ -367,6 +370,36 @@ UM.PreferencesPage
...
+
+                text: catalog.i18nc("@info:tooltip", "Printing via USB-cable isn't automatically enabled for new installs anymore, as the scanning of ports to find connected printers can give trouble on some systems.")
+
+                UM.CheckBox
+                {
+                    id: usbPrintCheckbox
+                    checked: boolCheck(UM.Preferences.getValue("usb_printing/enabled"))
+                    onClicked: UM.Preferences.setValue("usb_printing/enabled", checked)
+
+                    text: catalog.i18nc("@option:check", "Enable printing via USB-cable (restart required)")
```
wawanbreton: Other preferences have a `*` to indicate that the restart is required, I suggest to do the same here
```suggestion
                    text: catalog.i18nc("@option:check", "Enable printing via USB-cable *")
```

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -367,6 +370,36 @@ UM.PreferencesPage
                 width: UM.Theme.getSize("default_margin").width
             }
 
+            UM.Label
+            {
+                font: UM.Theme.getFont("medium_bold")
+                text: catalog.i18nc("@label", "Slice-Output")
```
wawanbreton: Not a big fan of this title, but I can't find anything better 😅

GregValiant:
This sort of thing can bump up against the language barrier.  This is my 2 cents worth.  I think being more direct here is a good thing.

image

The label:
_"USB Printing (Enable/Disable)"_
The tooltip:
_"Printing via USB-cable does not work with all printers and scanning for ports can interfere with other connected serial devices (ex: earbuds).  It is no longer 'Automatically Enabled' for new Cura installations.  If you wish to use USB Printing then enable it by checking the box and then restarting Cura.  Please Note:  USB Printing is no longer maintained.  It will either work with your computer/printer combination, or it won't."_

wawanbreton:
> This sort of thing can bump up against the language barrier. This is my 2 cents worth. I think being more direct here is a good thing.
> image
> 
> The label: _"USB Printing (Enable/Disable)"_ The tooltip: _"Printing via USB-cable does not work with all printers and scanning for ports can interfere with other connected serial devices (ex: earbuds). It is no longer 'Automatically Enabled' for new Cura installations. If you wish to use USB Printing then enable it by checking the box and then restarting Cura. Please Note: USB Printing is no longer maintained. It will either work with your computer/printer combination, or it won't."_

Thanks a lot for the suggestion :slightly_smiling_face:

resources/qml/Preferences/GeneralPage.qml
rburema: I did notice, but a bunch of them have 'restart required' instead. I think those outnumber the ones with a star even.

resources/qml/Preferences/GeneralPage.qml
rburema: That's what I thought when I wrote it as well -- I really couldn't fit it into any of the other categories.

Technically, since it's a plugin, it shouldn't even be here, but in it's own entry in the 'Extensions' menu -- but I thought that'd be way harder to find (and a bit of overkill for 1 option).

rburema:
@GregValiant I think the tooltip you wrote is an improvement indeed, but I specifically mention the cable in the option as well, so that people won't think printing via USB _disk_ is disabled (yes, that's a strange thing to assume if you know how anything works, but users don't always have that insight).

wawanbreton (approved):

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -356,7 +359,37 @@ UM.PreferencesPage
...
+            {
+                //: Spacer
+                height: UM.Theme.getSize("default_margin").height
+                width: UM.Theme.getSize("default_margin").width
+            }
+
+            UM.Label
+            {
+                font: UM.Theme.getFont("medium_bold")
+                text: catalog.i18nc("@label", "Output To Device")
```
HellAholic:
```suggestion
                text: catalog.i18nc("@label", "Connection and Control")
```

The Output To Device seems a bit too internal as category name, had a bit of a discussion and I believe Connection and Control is a generic enough category title to replace it.

resources/qml/Preferences/GeneralPage.qml
```diff
@@ -356,7 +359,37 @@ UM.PreferencesPage
...
+
+                text: catalog.i18nc("@info:tooltip", "Printing via USB-cable does not work with all printers and scanning for ports can interfere with other connected serial devices (ex: earbuds). It is no longer 'Automatically Enabled' for new Cura installations. If you wish to use USB Printing then enable it by checking the box and then restarting Cura. Please Note: USB Printing is no longer maintained. It will either work with your computer/printer combination, or it won't.")
+
+                UM.CheckBox
+                {
+                    id: usbPrintCheckbox
+                    checked: boolCheck(UM.Preferences.getValue("usb_printing/enabled"))
+                    onClicked: UM.Preferences.setValue("usb_printing/enabled", checked)
+
+                    text: catalog.i18nc("@option:check", "Enable/disable USB-cable printing (* restart required)")
```
HellAholic:
```suggestion
                    text: catalog.i18nc("@option:check", "Enable USB-cable printing (* restart required)")
```
Having disable in the label could confuse people to check the setting to disable the feature. The checkbox status should be sufficient, it also follows with the rest of the options' behavior.

PR #20970, opened 2025-09-19
Cura 12731 fix crash when loading a second paint model

wawanbreton:
The actual fix is just adding a defensve condition, but I sneaked in other changes for optimizations of the painting UI performance. I can make a separate PR if required.

CURA-12731

### DIFF
--- a/plugins/PaintTool/PaintTool.py
+++ b/plugins/PaintTool/PaintTool.py
@@ -5,7 +5,7 @@ import math
 from enum import IntEnum
 import numpy
 from PyQt6.QtCore import Qt, QObject, pyqtEnum, QPointF
-from PyQt6.QtGui import QImage, QPainter, QPen, QBrush, QPolygonF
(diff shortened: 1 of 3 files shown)


plugins/PaintTool/PaintCommand.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self._texture.setSubImage(cleared_image, self._bounding_rect.left(), self._bounding_rect.top())
+
+    def _makeClearedTexture(self) -> Tuple[QImage, QPainter]:
+        dest_image, painter = self._preparePainting()
+        self._clearTextureBits(painter)
+        return dest_image, painter
+
+    def _clearTextureBits(self, painter: QPainter):
+        raise NotImplementedError()
```
rburema, line 54: Given that people would need to implement this method for every `PaintCommand`, I expect a bit of documentation here.

wawanbreton: Yes indeed, this one at least would require documentation. I'll try to sneak that in an other PR :smile:

PR #20974, opened 2025-09-23
[CURA-12708] Store printer-API creds in prefs

rburema:
It's OK to do this in this manner and not via for example the keyring, since we there's no actual user-name involved, only the machine-name -- anyone on the same machine is _already_ capable of sending the basically the exact same request that originated the access in the first place.

(This is to further support the work for RED 3.3 compliance on the printer/firmware.)

_Also, this supersedes the other branch where I was trying to do this via the keyring -- that can be deleted once this is in._

### DIFF
--- a/cura/Backups/Backup.py
+++ b/cura/Backups/Backup.py
@@ -36,7 +36,7 @@ class Backup:
     IGNORED_FOLDERS = []  # type: List[str]
     """These folders should be ignored when making a backup."""
 
-    SECRETS_SETTINGS = ["general/ultimaker_auth_data"]
+    SECRETS_SETTINGS = ["general/ultimaker_auth_data", "cluster_api/auth_ids", "cluster_api/auth_keys", "cluster_api/nonce_counts", "cluster_api/nonces"]
     """Secret preferences that need to obfuscated when making a backup of Cura"""
 
     catalog = i18nCatalog("cura")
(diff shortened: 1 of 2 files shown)


casperlamboo (approved):

wawanbreton (approved):

fieldOfView:
May still be a good idea to not store this in projects too?
https://github.com/Ultimaker/Cura/blob/master/plugins/3MFWriter/ThreeMFWorkspaceWriter.py#L174

wawanbreton:
> May still be a good idea to not store this in projects too? https://github.com/Ultimaker/Cura/blob/master/plugins/3MFWriter/ThreeMFWorkspaceWriter.py#L174

Yes indeed, good catch, thanks !

HellAholic:
I jumped the commit action,
there is a whitelist for the preferences https://github.com/Ultimaker/Cura/blob/7a5c773e90ae97c7037d19f3077137a4c8aeb1dd/plugins/3MFWriter/ThreeMFWorkspaceWriter.py#L90 
So we don't need to specifically mention these items for the preferences.
However, the backups of cura should be considered. These keys hold same level of authentication information as "general/ultimaker_auth_data", so adding them to that list is the correct path.
https://github.com/Ultimaker/Cura/blob/7a5c773e90ae97c7037d19f3077137a4c8aeb1dd/cura/Backups/Backup.py#L34
The warning was valid, we just had to place them in a different list.
Thanks.

PR #20976, opened 2025-09-24
CURA-12730 fix paint switch 2 to 4 extruders

wawanbreton:
This fixes switching between printers with different numbers of extruders, that previously confused the painting. We now allocate a data range in the texture that can always contain up to 16 extruders, so that it will work in all cases.

CURA-12730


plugins/PaintTool/PaintView.py
```diff
@@ -114,14 +116,17 @@ def _makePaintModes(self):
...
         extruders_colors: Dict[str, "PaintView.PaintType"] = {}
 
-        for extruder_item in self._extruders_model.items:
-            if "color" in extruder_item:
+        for extruder_index in range(PaintView.MAX_EXTRUDER_COUNT):
+            extruder_item = self._extruders_model.getExtruderItem(extruder_index)
+            if extruder_item is None:
+                extruder_item = self._extruders_model.getExtruderItem(0)
+
+            if extruder_item is not None and "color" in extruder_item:
```
casperlamboo, line 124: I personally find this a bit cleaner since I read the second if-statement as a continuation of the first (and it saves one comparison)
```suggestion
            elif "color" in extruder_item:
```

casperlamboo (approved):

plugins/PaintTool/PaintView.py
wawanbreton: In this case, there is a first `if` that can set the `extruder_item` to a "fallback" value in case it is `None`, then there is the "normal" processing that re-checks for `None` because it may still happen, so I cannot use an `elsif` here, they are really supposed to be processed one after each other.

PR #20980, opened 2025-09-25
[CURA-12732] fix slice paint hover

rburema:
Don't cancel slice when hovering over the model in paint mode.


rburema:
@wawanbreton Removed the method, a new one wasn't needed, we already had that functionality it turns out :-)

wawanbreton:
> @wawanbreton Removed the method, a new one wasn't needed, we already had that functionality it turns out :-)

Ah great :slightly_smiling_face:

PR #20989, opened 2025-09-29
CURA-12740 properly handle material painting when object is set to extruder != 1

wawanbreton:
This requires watching which extruders are affected to which objects, and re-write the whole painting texture when the extruder of an object changes.

Requires https://github.com/Ultimaker/CuraEngine/pull/2249
CURA-12740


plugins/PaintTool/PaintView.py
```diff
@@ -293,18 +347,15 @@ def beginRendering(self) -> None:
...
-            if paint_texture != self._current_paint_texture and paint_texture is not None:
-                self._current_paint_texture = paint_texture
-
-                self._cursor_texture = OpenGL.getInstance().createTexture(paint_texture.getWidth(), paint_texture.getHeight())
-                self._paint_shader.setTexture(0, self._current_paint_texture)
-                image = QImage(paint_texture.getWidth(), paint_texture.getHeight(), QImage.Format.Format_ARGB32)
-                image.fill(0)
-                self._cursor_texture.setImage(image)
-                self._paint_shader.setTexture(1, self._cursor_texture)
+        # for node in Selection.getAllSelectedObjects():
```
casperlamboo:
```suggestion
```
Left over comment?

plugins/PaintTool/PaintView.py
```diff
@@ -243,32 +284,45 @@ def setPaintType(self, paint_type: str) -> None:
...
+        if self._painted_object is None:
+            return None
 
-        if stack_key not in self._paint_undo_stacks:
+        try:
+            return self._paint_undo_stacks[self._painted_object][self._current_paint_type]
+        except KeyError:
             stack: QUndoStack = QUndoStack()
-            stack.setUndoLimit(32)  # Set a quite low amount since some commands copy the full texture
+            stack.setUndoLimit(16)  # Set a quite low amount since some commands copy the full texture
```
casperlamboo, line 292: Why do we limit the undo stack in this PR, logic seems unrelated and 32 already feels like quite a low number?

plugins/PaintTool/PaintCommand.py
```diff
@@ -53,11 +53,15 @@ def _makeClearedTexture(self) -> Tuple[QImage, QPainter]:
...
-    def _getBitRangeMask(self) -> int:
-        bit_range_start, bit_range_end = self._bit_range
+    @staticmethod
+    def getBitRangeMask(bit_range: tuple[int, int]) -> int:
+        bit_range_start, bit_range_end = bit_range
         return (((PaintCommand.FULL_INT32 << (32 - 1 - (bit_range_end - bit_range_start))) & PaintCommand.FULL_INT32) >>
                 (32 - 1 - bit_range_end))
 
+    def _getBitRangeMask(self) -> int:
+        return PaintCommand.getBitRangeMask(self._bit_range)
```
casperlamboo, line 62: Why keep this method, it's private it should not be used by any other function nor api. We can just use the static function instead?

plugins/PaintTool/PaintView.py
wawanbreton: In the meantime, undo/redo logic has been changed to have a stack per object+feature, so now you can have a lot of stacks in memory. But yes this change should have been done in the proper PR.

plugins/PaintTool/PaintCommand.py
wawanbreton: This is because of my OO-mind: I see `self._bit_range` as a private variable, and `_getBitsRangeMask` as protected, so children classes should only access the method and not the variable (basic OO-principle, all variables should be private). In Python it does not really makes sense indeed, but at least child classes can call it without having to get the bit range.

PR #21001, opened 2025-10-01
[CURA-12752] Fix prime-tower for multi-material painting

rburema:
The prime-tower (w.r.t. the front-end, it was already working in-engine) didn't 'listen' to painted-on materials. Now it does.

I'm not _quite_ happy with the code, but this should be a good general direction.

I also managed to fix the 'if you first paint seams and _then_ materials something breaks' issue ... _unless_ there's some little vs. big-endian thing going on.

### DIFF
--- a/cura/BuildVolume.py
+++ b/cura/BuildVolume.py
@@ -117,6 +117,7 @@ class BuildVolume(SceneNode):
 
         self._has_errors = False
         self._application.getController().getScene().sceneChanged.connect(self._onSceneChanged)
+        self._application.getController().getScene().sceneChanged.connect(self._onStackChanged)
 
         # Objects loaded at the moment. We are connected to the property changed events of these objects.
         self._scene_objects = set()  # type: Set[SceneNode]
@@ -655,7 +656,7 @@ class BuildVolume(SceneNode):
                     extra_z = retraction_hop
         return extra_z
 
-    def _onStackChanged(self):
+    def _onStackChanged(self, *args) -> None:
         self._stack_change_timer.start()
 
     def _onStackChangeTimerFinished(self) -> None:
--- a/plugins/CuraEngineBackend/StartSliceJob.py
+++ b/plugins/CuraEngineBackend/StartSliceJob.py
@@ -764,21 +764,7 @@ class StartSliceJob(Job):
 
     @staticmethod
     def _getUsedExtruders(node: SceneNode) -> List[int]:
-        used_extruders = []
-
-        # Look at extruders used in painted texture
-        node_texture = node.callDecoration("getPaintTexture")
-        texture_data_mapping = node.callDecoration("getTextureDataMapping")
-        if node_texture is not None and texture_data_mapping is not None and "extruder" in texture_data_mapping:
-            texture_image = node_texture.getImage().copy()
-            image_ptr = texture_image.bits()
-            image_ptr.setsize(texture_image.sizeInBytes())
-            image_size = texture_image.height(), texture_image.width()
-            image_array = numpy.frombuffer(image_ptr, dtype=numpy.uint32).reshape(image_size)
-
-            bit_range_start, bit_range_end = texture_data_mapping["extruder"]
-            image_array = (image_array << (32 - 1 - (bit_range_end - bit_range_start))) >> (32 - 1 - bit_range_end)
-            used_extruders = numpy.unique(image_array).tolist()
+        used_extruders = ExtruderManager.getInstance().getPaintedExtruders(node)
 
         # There is no relevant painting data, just take the extruder associated to the model
         if not used_extruders:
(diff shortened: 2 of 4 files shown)


plugins/PaintTool/PaintTool.py
```diff
@@ -408,7 +408,7 @@ def event(self, event: Event) -> bool:
                 Logger.logException("e", "Error when adding paint stroke")
 
             self._last_world_coords = world_coords
-            self._updateScene(node)
+            self._updateScene(node if event_caught else None)
```
wawanbreton: I am afraid this is useless, because in `_updateScene` we get the selected object if the given node is `None`, which will in practice retrieve the same node 😄

cura/BuildVolume.py
```diff
@@ -117,6 +117,7 @@ def __init__(self, application: "CuraApplication", parent: Optional[SceneNode] =
 
         self._has_errors = False
         self._application.getController().getScene().sceneChanged.connect(self._onSceneChanged)
+        self._application.getController().getScene().sceneChanged.connect(self._onStackChanged)
```
casperlamboo:
```suggestion
        scene = self._application.getController().getScene()
        scene.sceneChanged.connect(self._onSceneChanged)
        scene.sceneChanged.connect(self._onStackChanged)
```
At first I thought both lines were duplicated. Maybe if we make the line less cognitive complex the slight difference becomes a bit more apparent.

plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -764,27 +764,10 @@ def _addRelations(self, relations_set: Set[str], relations: List[SettingRelation
 
     @staticmethod
     def _getUsedExtruders(node: SceneNode) -> List[int]:
```
casperlamboo: If I read this function name then I assume the function would return all used extruders in the print. As far as I know an extruder can be used due to one of three reasons
- a model is assigned to the extruder
- settings dictate a certain extruder should be used
- an extruder is used by painting
The function body only implements reason three.

To me it feels like that either the function `_getUsedExtruders` is a misnomer, or `node.callDecoration("getPaintedExtruders")` is.

cura/Scene/SliceableObjectDecorator.py
```diff
@@ -29,6 +39,29 @@ def getPaintTexture(self) -> Optional[Texture]:
     def getPaintTextureChangedSignal(self) -> Signal:
         return self.paintTextureChanged
 
+    def setPaintedExtrudersCountDirty(self) -> None:
+        self._texture_change_timer.start()
```
casperlamboo, line 43: What happens if this function is called multiple times in fast succession? Lets say the function is called at time interval $t + 0$, $t + 50$ and $t + 100$, will `_onTextureChangeTimerFinished` be called at  $t + 500$, $t + 550$ and $t + 600$ or only at $t + 600$?

rburema: We use the same technique all throughout the front-end to make sure that if you do a lot of operations in quick succession it'll only fire once, so it'll only fire at `t + 600` in your scenario.

Just looked up the documentation for `QTimer.start(...)` just to be sure; and indeed it does say it either starts _or restarts_ the timer.

plugins/CuraEngineBackend/StartSliceJob.py
rburema: The first option -- I do think it should be `_getPaintedExtruders`.

cura/BuildVolume.py
rburema: It's already less complex than before the change 😉 -- but I see what you mean.

I think a comment will do the trick though.

plugins/CuraEngineBackend/StartSliceJob.py
rburema: ... hm I'll change it to `_getMainExtruders` instead, as it's supposed to give the main model extruder 'color' if no painting is involved (that is, the method is still relied on even if nothing is painted).

cura/BuildVolume.py
rburema: Oops, I somehow thought the 'suggested change' was what the comment was about 😓 -- yeah I think it's better, will be adding that.

plugins/PaintTool/PaintClearCommand.py
```diff
@@ -5,6 +5,7 @@
 
 from PyQt6.QtGui import QUndoCommand, QImage, QPainter, QBrush
 
+from Scene.SliceableObjectDecorator import SliceableObjectDecorator
```
HellAholic:
```suggestion
from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
```

plugins/PaintTool/PaintStrokeCommand.py
```diff
@@ -7,6 +7,7 @@
 from PyQt6.QtCore import QRect, QRectF, QPoint
 from PyQt6.QtGui import QUndoCommand, QImage, QPainter, QPainterPath, QPen, QBrush
 
+from Scene.SliceableObjectDecorator import SliceableObjectDecorator
```
HellAholic:
```suggestion
from cura.Scene.SliceableObjectDecorator import SliceableObjectDecorator
```

saumyaj3 (approved):

PR #21002, opened 2025-10-02
CURA-12743 optimize painting performance

wawanbreton:
This PR optimizes the performance of real-time painting by doing some behavior changes:
* The heavy calculation of projecting the paint stroke to the texture is now done in the C++ instead of Python code
* The number of buffer copies and internal paintings has been drastically reduced

Requires https://github.com/Ultimaker/Uranium/pull/1012
Requires https://github.com/Ultimaker/libUvula/pull/2
CURA-12743


plugins/PaintTool/PaintTool.py
```diff
@@ -269,90 +271,42 @@ def _getStrokePolygon(self, stroke_a: numpy.ndarray, stroke_b: numpy.ndarray) ->
...
-        return res
+        stroke_poly.toType(numpy.float32)
+
+        mesh_indices = self._mesh_transformed_cache.getIndices()
+        if mesh_indices is None:
+            mesh_indices = numpy.array([], dtype=numpy.int32)
+
+        res = uvula.project(stroke_poly.getPoints(),
+                            self._mesh_transformed_cache.getVertices(),
+                            mesh_indices,
```
casperlamboo, line 298: Do you know if we're copying meshdata each time we invoke the `uvula.project` function, or are the bindings good enough to be able to work solely by points/references?

wawanbreton: From everything I could read, this should really not do any data copy (hence also using std::span), but I can only trust pybind11 documentation :smile:

casperlamboo: good!

Issue #21012, opened 2025-10-05
Unable To Group


insertinterestingnamehere:
### Cura Version

5.11.0 Beta 1

### Operating System

Linux x86-64 (Fedora 42)

### Printer

Custom

### Reproduction steps

Group two objects then try to ungroup them.

### Actual results

The ungroup operation doesn't show up as available. If you ungroup using an undo action you're no longer able to group the objects either because that option also no longer shows as available. Using the keyboard shortcuts also does not work.

### Expected results

Grouping and ungrouping should be available as options.

### Add your .zip and screenshots here ⬇️

Note: this is a regression between the 5.11.0 alpha and the 5.11.0 beta. The alpha version works fine in this regard.

PartySausage:
Yes I can confirm this. CTRL + Z  (Undo) seems to work for ungrouping though. 
I'm also seeing that when I group a model all my tools become greyed out (not just paint) & remain greyed out even if I ungroup, revert to the original model with CTRL + Z or clear the build late and load a new model. I have to reopen Cura to enable them again

<img width="964" height="761" alt="Image" src="https://github.com/user-attachments/assets/be0525ea-d591-478e-a11f-7e45d115faca" />

HellAholic:
Thanks for the report,
Known issue mentioned in the release notes for the beta. Paint tool does not work with grouping models or grouped models.

PartySausage:
> Thanks for the report, Known issue mentioned in the release notes for the beta. Paint tool does not work with grouping models or grouped models.

Hi @HellAholic That i knew I read the release that's why I mentioned the (not just paste) part. what I was trying to show in my original screenshot is that none of the tools work they are all greyed out grouped then ungroupd & you have to restart Cura to get them back

Here is the same model in 5.10.2 after I multiplied x 2 grouped, ungrouped & reverted to the original single instance of the model

<img width="961" height="676" alt="Image" src="https://github.com/user-attachments/assets/90724356-2f5b-492d-8269-141f3e1a1664" />

GregValiant:
Me too.
After "Grouping" models:
- The entire left toolbar disables.
- The three Group/Merge commands in the right-click menu become disabled.
- crtl-z" can be used to ungroup, but the disabilities persist.
- Models can be selected...but that's it.
- Closing and restarting Cura resolve the problem.
- I don't see anything helpful in the log.

[...]

PR #21019, opened 2025-10-06
CURA-12761 Handle group selection as multi-selection w.r.t painting

wawanbreton:
Fixes group handling, and disable painting for grouped objects. This is not really nice because it forces the user to ungroup objects to be able to paint them. However, the painting is currently designed to allow painting a single object at once, and selecting a group does select multiple objects. This should be improved later.

CURA-12761
Fixes https://github.com/Ultimaker/Cura/issues/21012

### DIFF
--- a/plugins/PaintTool/PaintTool.qml
+++ b/plugins/PaintTool/PaintTool.qml
@@ -327,7 +327,7 @@ Item
         UM.Label
         {
             anchors.fill: parent
-            text: catalog.i18nc("@label", "Select a single model to start painting")
+            text: catalog.i18nc("@label", "Select a single ungrouped model to start painting")
             verticalAlignment: Text.AlignVCenter
             horizontalAlignment: Text.AlignHCenter
         }
--- a/plugins/PaintTool/PaintTool.qml
+++ b/plugins/PaintTool/PaintTool.qml
@@ -327,7 +327,7 @@ Item
         UM.Label
         {
             anchors.fill: parent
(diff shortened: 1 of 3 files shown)


plugins/PaintTool/PaintTool.qml
```diff
@@ -327,7 +327,7 @@ Item
         UM.Label
         {
             anchors.fill: parent
-            text: catalog.i18nc("@label", "Select a single model to start painting")
+            text: catalog.i18nc("@label", "Select a single ungrouped model to start painting")
```
casperlamboo: Don't we have string freeze? 🥶

casperlamboo (approved):

plugins/PaintTool/PaintTool.qml
wawanbreton: Yes we do... Maybe we should apply this change only on main ?

casperlamboo: I think we should, otherwise we'll get untranslated string which is worse then slightly wrong strings imo. Also, I think we should adhere to our string freeze.

wawanbreton: Makes sense, I will make a separate PR

wawanbreton: https://github.com/Ultimaker/Cura/pull/21025

casperlamboo (approved):

PR #21021, opened 2025-10-06
CURA-12747 Fix unable to slice when using paint-on-seam first

wawanbreton:
Use the more robust method to get used extruders in texture, so that we don't get inexisting extruders when used in combination with paint-on-seam
CURA-12747

### DIFF
--- a/resources/i18n/ja_JP/cura.po
+++ b/resources/i18n/ja_JP/cura.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2025-09-22 08:45+0200\n"
-"PO-Revision-Date: 2025-03-02 04:30+0000\n"
+"PO-Revision-Date: 2025-10-06 13:55+0000\n"
 "Last-Translator: h1data <h1data@users.noreply.github.com>\n"
 "Language-Team: Japanese <https://github.com/h1data/Cura/wiki>\n"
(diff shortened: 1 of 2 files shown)


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -770,15 +770,18 @@ def _getUsedExtruders(node: SceneNode) -> List[int]:
...
+            image_ptr = texture_image.constBits()
             image_ptr.setsize(texture_image.sizeInBytes())
             image_size = texture_image.height(), texture_image.width()
             image_array = numpy.frombuffer(image_ptr, dtype=numpy.uint32).reshape(image_size)
 
             bit_range_start, bit_range_end = texture_data_mapping["extruder"]
-            image_array = (image_array << (32 - 1 - (bit_range_end - bit_range_start))) >> (32 - 1 - bit_range_end)
-            used_extruders = numpy.unique(image_array).tolist()
+            full_int32 = 0xffffffff
+            bit_mask = (((full_int32 << (32 - 1 - (bit_range_end - bit_range_start))) & full_int32) >> (
```
casperlamboo, line 781: What is the benit of the `(... & full_int32)` part? An `&` operation on a set of full 1's results in the same input?

wawanbreton: If I remember correctly, this is because Python uses 64-bits ints by default, so at some point I had to add this to filter out the MSB part. But we could also try to remove it, it may be a remnant of some code change...

casperlamboo (approved):

PR #21069, opened 2025-10-21
Safeguard against decrease in number of extruders

HellAholic:
When the painted model goes from a machine with higher number of extruders to a lower number of extruders with painted data, adding the safeguard to prevent crash with a try-except

CURA-12800

### DIFF
--- a/cura/Settings/ExtruderManager.py
+++ b/cura/Settings/ExtruderManager.py
@@ -257,9 +257,10 @@ class ExtruderManager(QObject):
             painted_extruders = node.callDecoration("getPaintedExtruders")
             if painted_extruders is not None:
                 for extruder_nr in painted_extruders:
-                    if str(extruder_nr) not in self.extruderIds:
-                        extruder_nr = int(self._application.getMachineManager().defaultExtruderPosition)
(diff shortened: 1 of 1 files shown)


cura/Settings/ExtruderManager.py
```diff
@@ -257,6 +257,8 @@ def getUsedExtruderStacks(self) -> List["ExtruderStack"]:
             painted_extruders = node.callDecoration("getPaintedExtruders")
             if painted_extruders is not None:
                 for extruder_nr in painted_extruders:
+                    if str(extruder_nr) not in self.extruderIds:
+                        extruder_nr = int(self._application.getMachineManager().defaultExtruderPosition)
                     used_extruder_stack_ids.add(self.extruderIds[str(extruder_nr)])
```
wawanbreton:
```suggestion
                    try:
                        used_extruder_stack_ids.add(self.extruderIds[str(extruder_nr)])
                    except KeyError:
                        pass
```
Adding the `defaultExtruderPosition` could be wrongly adding an unused extruder. In this case I do prefer a try/except to indicate that this is kind of an edge-case, and it also avoids some string/int conversions.

wawanbreton (changes requested):

PR #21083, opened 2025-10-24
CURA-12811 settings with errors can be ignored when slicing

wawanbreton:
This PR fixes a few minor issues regarding checking for settings on errors. The commits may be reviewed separately for better understanding.

The actual fix for the mentioned bug is more of a workaround though. I did find 2 signals in the `ContainerStack` class, namely `propertyChanged` and `propertiesChanged`, which theoretically should always be emitted together. However, in practice there are a lot of places where one of them is emitted from outside the class (which should be forbidden) and the other one is not. I tried to add a method to emit them together, but ran into an infinite loop of self-calling signals. Given the number of users and emitters of those signals, it is very difficult to pinpoint if someone specific is doing something wrong, so giving up for now.
Is there an exaplanation for having those 2 separate signals ? The documentation says nothing specific about them.

CURA-12811

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -1,16 +1,16 @@
 version: "5.12.0-alpha.0"
 commit: "unknown"
 requirements:
-  - "cura_resources/5.12.0-alpha.0"
-  - "uranium/5.12.0-alpha.0"
-  - "curaengine/5.12.0-alpha.0"
-  - "cura_binary_data/5.12.0-alpha.0"
-  - "fdm_materials/5.12.0-alpha.0"
+  - "cura_resources/5.12.0-alpha.0@ultimaker/testing"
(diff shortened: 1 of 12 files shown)


cura/Settings/MachineManager.py
```diff
@@ -515,13 +489,10 @@ def clearUserSettingAllCurrentStacks(self, key: str) -> None:
             container.sendPostponedEmits()
 
     @pyqtProperty(bool, notify = stacksValidationChanged)
+    @deprecated("This property was already inactive and will now be removed, use MachineErrorChecker.hasError instead.", since="5.11.0")
```
rburema: Surely 'since 5.12.0'? ;-)

wawanbreton: Oh yes, it has changed in the meantime :smile:

PR #21158, opened 2025-11-24
Add changes for windows arm64 and update dependencies

HellAholic:
This pull request introduces several updates to enable Windows ARM build compatibility, dependency management, and package versioning. The main changes include adding support for Windows ARM builds, refining how platform-specific dependencies are handled (especially for `pynavlib`), updating the `cpython` version, and cleaning up dependency definitions for `shapely`.

**Windows build and dependency management:**

* Added support for self-hosted Windows ARM runners in the workflow by updating `.github/workflows/windows.yml`.
* Ensured `pynavlib` is only included for Windows x64 builds by updating `conandata.yml` and modifying the PyInstaller spec generation in `conanfile.py` to exclude `pynavlib` on Windows ARM64. [[1]](diffhunk://#diff-545553e7b8144a900437cd74842021b8640d473390f8a836db86ad55d85dd069R555-R557) [[2]](diffhunk://#diff-63f09721ade419d29886e8190ffa3c9c756baa9bef8802a7f2aeabf3b545164fR552-R555)

**Dependency updates:**

* Upgraded the `cpython` dependency from version 3.12.2 to 3.12.7 in `conanfile.py`.
* Added `shapely/2.0.6` to the conan requirements in `conandata.yml` and removed its pip hashes, consolidating its management under Conan. [[1]](diffhunk://#diff-545553e7b8144a900437cd74842021b8640d473390f8a836db86ad55d85dd069R12) [[2]](diffhunk://#diff-545553e7b8144a900437cd74842021b8640d473390f8a836db86ad55d85dd069L413-L457)

**Add test Workflow management to be able to validate the changes on branch**

* Updated the workflow to use a specific branch/tag (`CURA-12814`) for the Windows installer job for testing windows ARM.


arm64_cpp20_profile
wawanbreton, line 1: This should be in a profile in `conan-config`

conandata.yml
```diff
@@ -127,7 +128,6 @@ pyinstaller:
         - "PyQt6.sip"
         - "stl"
         - "keyrings.alt"
-        - "pynavlib"
```
wawanbreton, line 125: Are we really ok with doing this ?

wawanbreton (changes requested):

conandata.yml
HellAholic: Tried an alternative approach to filter out the dependency on arm windows but keep it for x64.
Requires testing still.

wawanbreton (approved):

wawanbreton (approved):

.github/workflows/find-packages.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     name: Create installers
     needs: find-packages
     if: ${{ inputs.start_builds == true && needs.find-packages.outputs.discovered_packages != '' }}
-    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@main
+    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@CURA-12814
```
HellAholic:
```suggestion
    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@main
```
Apply suggestion after the cura-workflows branch is merged to main

.github/workflows/installers.yml
```diff
@@ -29,7 +29,7 @@ on:
 jobs:
   installers:
     name: Create installers
-    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@main
+    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@CURA-12814
```
HellAholic:
```suggestion
    uses: ultimaker/cura-workflows/.github/workflows/cura-installers.yml@main
```
Apply suggestion after the cura-workflows branch is merged to main

.github/workflows/windows.yml
```diff
@@ -34,11 +34,12 @@ on:
         type: choice
         options:
           - self-hosted-Windows-X64
+          - self-hosted-Windows-ARM64
           - windows-2022
 
 jobs:
   windows-installer:
-    uses: ultimaker/cura-workflows/.github/workflows/cura-installer-windows.yml@main
+    uses: ultimaker/cura-workflows/.github/workflows/cura-installer-windows.yml@CURA-12814
```
HellAholic:
```suggestion
    uses: ultimaker/cura-workflows/.github/workflows/cura-installer-windows.yml@main
```
Apply suggestion after the cura-workflows branch is merged to main

.github/workflows/windows-arm.yml
```diff
@@ -0,0 +1,48 @@
...
+        description: 'OS'
+        required: true
+        default: 'self-hosted-Windows-ARM64'
+        type: choice
+        options:
+          - self-hosted-Windows-ARM64
+
+jobs:
+  windows-installer:
+    uses: ultimaker/cura-workflows/.github/workflows/cura-installer-windows-arm.yml@CURA-12814
```
HellAholic:
```suggestion
    uses: ultimaker/cura-workflows/.github/workflows/cura-installer-windows-arm.yml@main
```

resources/qml/Menus/ExtensionMenu.qml
```diff
@@ -39,7 +39,7 @@ Cura.Menu
         {
             id: sub_menu
             title: model.name
-            shouldBeVisible: actions !== undefined
+            shouldBeVisible: actions != null
```
casperlamboo:
```suggestion
            shouldBeVisible: !!actions
```

conandata.yml
```diff
@@ -591,6 +548,9 @@ pip_requirements_core:
       hashes:
         - sha256:8a1513379d709975552d202d942d9837758905c8d01eb82b8bcc30918929e7b8
         - sha256:d162dc04946d704503b2edc4d55f3dba5c1d539ead017afa00142c38b9885755
+
+  Windows_x64:
+    # pynavlib is only available for Windows x64, not ARM64
```
rburema, line 630: We need to make sure customers know they can't use the 3D mouse on Win ARM64 builds then.

resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
```diff
@@ -61,7 +61,12 @@ Cura.ExpandablePopup
                     Layout.alignment: Qt.AlignCenter
 
                     property var extruderStack: activeMachine ? activeMachine.extruderList[model.index]: null
-                    property bool valueWarning: !Cura.ExtruderManager.getExtruderHasQualityForMaterial(extruderStack)
+                    property bool valueWarning: {
+                        // Explicitly read material and variant to register them as QML binding dependencies.
+                        var _material = extruderStack ? extruderStack.material : null
+                        var _variant = extruderStack ? extruderStack.variant : null
```
rburema, line 67: Is that also true if they're in a scope? It could be, but I thought that `{...}` would destroy bindings.

resources/qml/Menus/ExtensionMenu.qml
```diff
@@ -58,7 +58,7 @@ Cura.Menu
             }
         }
 
-        onObjectAdded: function(index, object) { extensionMenu.insertMenu(index, object) }
-        onObjectRemoved: function(index, object) { extensionMenu.removeMenu(object)}
+        onObjectAdded: function(index, object) { if (object.shouldBeVisible) { extensionMenu.insertMenu(index, object) } }
+        onObjectRemoved: function(index, object) { extensionMenu.removeMenu(object) }
```
rburema, line 62: I assume this will also work if the object wasn't visible when added? In that hopefully, the menu ignores irrelevant 'remove' calls.

rburema (approved):
LGTM

conandata.yml
HellAholic: I think we can ask the 3dconnexion to provide the package for arm, but it was a bit premature to ask them if we don't ship a build.

resources/qml/Menus/ConfigurationMenu/ConfigurationMenu.qml
HellAholic: So due to the Qt changes, we need to explicitly register the bindings, otherwise it is not triggering an update on the change.
The `{ ... }` would be breaking when using with a static value I believe.

resources/qml/Menus/ExtensionMenu.qml
HellAholic: Yes, tried with the visibility, but this section is for the macos native menu, and potential follow up work might change this in: CURA-13240 (as either revert/exclusion, or implementation changes to work as expected with the visible/should be visible properties, similar issue in the File menu)

PR #21163, opened 2025-11-25
CURA-12833 Add setting to enable interlaced bridging lines

wawanbreton:
Add option to print bridging lines interlaced, i.e. print them in 2 monotonic passes so that adjacent lines won't be printed just after each other.

CURA-12833
Requires https://github.com/Ultimaker/CuraEngine/pull/2266

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -486,6 +486,19 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "machine_time_estimation_factor":
+                {
+                    "label": "Print Time Estimation Factor",
+                    "description": "Multiplier for adjusting print time estimates. 100% shows original estimate. Higher values indicate longer actual print times.",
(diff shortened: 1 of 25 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -9011,6 +9011,15 @@
                     "enabled": "bridge_settings_enabled",
                     "settable_per_mesh": true
                 },
+                "bridge_interlace_lines":
+                {
+                    "label": "Interlace Bridge Lines",
+                    "description": "blablabla.",
```
casperlamboo: Think this is a placeholder

wawanbreton: Yes indeed, at least it made its job as being recognized as such :grin:

casperlamboo:
We should still add this setting to the various visibility files

wawanbreton:
> We should still add this setting to the various visibility files

I do agree, however all the briding settings are not present in those files, probably because bridging is still experimental. So leaving as is.

PR #21166, opened 2025-11-25
Correct calculation of initial bottom layers & clean up

NOVAXIM:
# Explanation
These changes are intended to calculate the correct number of layers when slicing a model before printing. This only works in specific scenario where the `Initial Layer Height` is different than the `Layer Height`. In addition to the correct calculation, we obtain the correct layers result on the slider.

> [!NOTE]
> Most users don't notice the issue because `Initial Bottom Layers` is not shown by default (even in Expert settings) and even fewer users check these values.

> [!IMPORTANT]
> Please note that the calculation of the number of layers is always based on the layer thickness value.

# Description of changes

Line 1724: Improved method of calculating layers (does not create a layer with 0.001 mm)
Line 1739: The formula for calculating the number of `Initial Bottom Layers` shows the correct values finally
Line 1754: Improved method of calculating layers (does not create a layer with 0.001 mm)
Line 3766: the formula "* 30 / 60" is easier to write as "/ 2"
Line 7269: the formula "/ 60 * 30" is easier to write as "/ 2"

## Type of change

- [x] Clean up

# How Has This Been Tested?

- [x] Tested with Cura version 5.11.0 and previous. Working well.


rburema (approved):
looks good to me; I'll pass it onto QA
thanks for the contrib

HellAholic:
Did a round of testing, found couple of edge cases that based on experience and the way profiles are set, the first one is considered as a blocker.
So the formula can return either a 0 or a negative number which will causes scenarios such as the following
- Initial layers is calculated as 0 [Blocker as users might not even notice the change and end up without the bottom layer]
image
- Negative value is returned [unable to slice message prevents a print fail, as the field falls below the minimum value, so a critical rather than a blocker]
image

If you could take a look and adjust the formula to also consider those situations that would be great.

NOVAXIM:
Interesting observations. I'll look into this and respond soon.

MariMakes:
Hey @NOVAXIM did you get a chance to look at the suggestions?

NOVAXIM:
Thank you for your patience and pointing out this imperfection. I really appreciate it.

Of course, I have a solution to this problem. However, I noticed that you want to print layer 0.2 when the first layer is 0.4 – something is wrong.
I really wanted to fix it so that there was a red background for "Bottom Thickness" when lower than "Initial Layer Height", but that's not possible.

### Here's the corrected version of my formulas:
- [x] Initial Bottom Layers negative value is not possible
- [x] I have standardized the formulas for all types of layers. 
- [x] I took into account the `"minimum_value_warning": "2"` used in `top_layers`, `bottom_layer` and `initial_bottom_layers` (that's why I removed the strange value: 0.2 and multiplied the layer height by 2). From now on, you need the full size of the layer to print. Additionally, the color field will warn you (yellow), if you break the two-layer minimum rule (which was previously inconsistent).

I tested it thoroughly, but look at it critically to make sure everything is perfect. Thank you.

NOVAXIM:
> [!TIP]
> If you want to quickly test this solution, all changes are in the printer definition that is on my repository:
> [SW-X4-series-definition-for-Cura](https://github.com/NOVAXIM/SW-X4-series-definition-for-Cura)
> Select version: 5.10.2-latest, 5.12.0 or 5.13.0 (depending on the installed version of Cura).
> Of course, choose the **X4 series** printer. Detailed instructions are available on the main page (README.md).

NOVAXIM:
Hi @HellAholic, @rburema and @MariMakes! 😊

Just wanted to kindly bump this PR. I have thoroughly addressed the edge cases and the blocker you mentioned, @HellAholic. Initial bottom layers can no longer result in negative or zero values. I have also standardized the formulas across all layer types and fixed the inconsistent two-layer minimum warning rules.

The corrected implementation has been tested and is ready for your critical review. Could you please take a look when you have a free moment? 

Thank you for your time and help in making this contribution better! 🙏

resources/definitions/fdmprinter.def.json
```diff
@@ -1721,7 +1721,7 @@
                                     "maximum_value": "999999",
                                     "type": "int",
                                     "minimum_value_warning": "2",
-                                    "value": "math.ceil(round(top_thickness / resolveOrValue('layer_height'), 4))",
+                                    "value": "int(round(top_thickness / resolveOrValue('layer_height'), 4))",
```
HellAholic: `math.ceil` is used intentionally here, so changing it to int affects the behavior for a lot of printers (including more than 100 profiles for the UM machines).

resources/definitions/fdmprinter.def.json
```diff
@@ -1751,7 +1751,7 @@
                                     "default_value": 6,
                                     "maximum_value": "999999",
                                     "type": "int",
-                                    "value": "math.ceil(round(bottom_thickness / resolveOrValue('layer_height'), 4))",
+                                    "value": "int(round(bottom_thickness / resolveOrValue('layer_height'), 4))",
```
HellAholic: Same as above, for `math.ceil`

HellAholic (changes requested):
The switch to `int` from `math.ceil` has a major impact on a large number of printers and profiles. It is also not required/mentioned for the changes stated in the PR.

resources/definitions/fdmprinter.def.json
NOVAXIM: int has no negative impact on printers - it only improves counting accuracy.
Improved method of calculating layers (prevents counting layers from 0.001 mm).

resources/definitions/fdmprinter.def.json
NOVAXIM: int has no negative impact on printers - it only improves counting accuracy
Improved method of calculating layers (prevents counting layers from 0.001 mm).

NOVAXIM:
> The switch to `int` from `math.ceil` has a major impact on a large number of printers and profiles. It is also not required/mentioned for the changes stated in the PR.

Yes, it does have an impact, not so much on printers, but on counting the number of layers.
Look at the screenshot below:
obraz
In the new method of counting, the number of layers must be covered in the layer thickness. In this example, you will only reach layer 7 when you specify 1.12mm of thickness (7 x 0.16mm). Compare to what you have in Cura now.
In the new counting system, you need to have full thickness to count a layer. In the old system, the printer created a layer from a void. If you had 0.001 mm, the printer counted it as a full layer. This wasn't good.
This won't change how you print. It will affect all printer profiles, but it won't break them. It's just a way of counting layers.

HellAholic:
The concern is not whether the new counting method is better, but that changing how `layer counts` are derived changes the effective values used by existing printer profiles.

From UltiMaker's perspective, this would affect the majority of our print profiles. Every profile has been tuned and validated with the current behavior. If a profile is configured for 5 bottom layers and, after this change, the same settings result in 3 bottom layers, that changes the printed part. 

That means every affected profile would need to be re-validated to ensure print quality is still acceptable. The same is likely true for many third-party printer profiles of the community members. Users are unlikely to notice until they start a print.

So while the physical layer height does not change, the behavior of existing profiles does. That compatibility risk and the associated validation effort make this change unsuitable.

NOVAXIM:
> If a profile is configured for 5 bottom layers and, after this change, the same settings result in 3 bottom layers, that changes the printed part.

Not. My new mathematical rule for counting layers is rounding down, but currently it is rounding up. Therefore, in some (not all) cases the number of layers may be only 1 less.

I've analyzed almost all the printer profiles, and I have to say that most are very good and wouldn't affect my proposed new layer count system (it provide the number of layers, not the value in mm). However, that's a topic for a separate thread. I'll return in another thread, where I'll discuss all aspects of the issue.

Thank you for your attention and input. I'm reverting to the original formula for calculating layers (top/bottom). Please merge this.

HellAholic (approved):

Issue #20347, opened 2025-03-04
improve layer visualization like prusa slicer


raffa1799:
### Is your feature request related to a problem?

I am frustrated every time I use the layers slider to see at what height a certain part is printed, but only the number of layers is indicated and not the height, I also noticed there is no option to choose whether to print the supports synchronized with the layers of the piece or not, perhaps there is also an error when I print a piece with a layer height of 0.2mm and I have a support z distance of 0.25mm the distance remains 0.2mm from the piece.
If the media interface thickness is set to 0.2 and the layer is set to 0.2, I don't know why but 2 layers are printed instead of one

### Describe the solution you'd like

insert next to the layer number, the layer height and printing time reached at that point, as on Prusa slicer.

### Describe alternatives you've considered

set to 0.05 layer heigth

### Affected users and/or printers

all users!

### Additional information & file uploads

Layer heigth set to 0.2mm , the support interface layer will be one

![Image](https://github.com/user-attachments/assets/d18784f1-881c-4fa3-bc90-4ca7dac895df)
![Image](https://github.com/user-attachments/assets/cd533c93-3b5d-48fa-8cfa-7e458dabff7e)

add time and Z heigth to bed in mm

damian238:
Those would be great additions. Also nice would be an ability to type in a specific layer number (or height) to view the situation at that specific layer.

Issue #7683, opened 2020-05-05
Estimated Time remaining


RowdyRocket:
I usually load up Cura to see about where I am on my print to get an idea of how much time is remaining. 

It would be wonderful if the slider in preview mode would give you an estimated time remaining from the point selected in the preview slider bar.

![Print Progress](https://user-images.githubusercontent.com/16073365/81072283-ecb81a80-8eb3-11ea-8575-ff1572156b4f.jpg)

fieldOfView:
Next time please fill in the issue template. I'm guessing you use the OctoPrint plugin?

The biggest problem here is that for performance reasons, Cura normally does not display gcode files, but specially created layer data. That layer data is no longer available when a new slice is made, let alone when Cura is restarted. There is also no direct correlation between gcode line numbers and layer data, which makes it hard to sync up the information about print progress with the layer data.

RowdyRocket:
Sorry I didn't use the template. And No, no OcyoPrint plugin, that was just something I threw together in a paint program to show what I was looking for. Does Octoprint offer this? I'm not asking that it track in real-time just an estimated time as I scroll through the preview. Simple math time divided by slices times slice I'm currently viewing should tell elapsed time then subtract that from overall time to get remaining.

Thanks for the quick response. I'll use the template next time.

Ghostkeeper:
CuraEngine could send over the print time estimates along with the layer data. It already calculates this (and puts it in the g-code). I don't really see much of a problem with that. This is different from the g-code-line-number information in that other feature request, because there is just a single point where the layer gets sent as a whole to the front-end, so it's not sensitive to programmer mistakes or hard on maintenance.

The only problem that I see is where to fit it in the interface. For most people there is a big panel of settings there, which is why we moved the layer counter to be displayed above the handle. It would also be a bit ugly/overwhelming to have panels coming out from multiple directions, in my opinion. I'm no expert in GUI design though, so maybe my colleagues have a better idea.

mahtDFR:
We think this is a valid feature but we have decided to defer it as we can't give it priority right now. We would be very open to someone contributing a solution as a pull request.

louiswouters:
@RowdyRocket You may use [this](https://github.com/louiswouters/3D-Printing/blob/master/scripts/ShowProgress.py) PostProcessing script to write the current layer and estimated time to the display of your printer.

5axes:
> 
[...]

Issue #6179, opened 2019-08-08
Display Z height instead/in addition to layer number on print preview


AbeFM:
**Is your feature request related to a problem? Please describe.**
Being able to find the location  of a critical feature is one of the best uses of the slicing preview. Often, this input can be critical in determining where to make changes. 

This is also of major concern when layer sizes change for any reason (first layer thickness, variable layer heights).

**Describe the solution you'd like**
![MillimetersPlease](https://user-images.githubusercontent.com/42385/62744790-07c92d00-b9fd-11e9-944b-f9842a2c8d8e.png)
I kept it all the same size.

**Describe alternatives you've considered**
In theory the same information could be gathered by importing the model into an external cad program, then offset by the physical placement of the object in the CURA space. Then do the layer heights math.....

One could create a primitive (support blocker) and scale it until it matches the object... But it still can be hard to get the information.

**Affected users and/or printers**
I think it hits a pretty broad swath of people. I'm kinda afraid it's right there in front of me and I'm missing it. :-)

**Additional context**

![image](https://user-images.githubusercontent.com/42385/62743969-62f92080-b9f9-11e9-9c5b-9c7500022469.png)

Brand X:
![image](https://user-images.githubusercontent.com/42385/62744179-4ad5d100-b9fa-11e9-9c19-f1fcef9a7b00.png)

Brand Y:
![image](https://user-images.githubusercontent.com/42385/62744257-9daf8880-b9fa-11e9-9c11-8baa5693775c.png)

Vandrasc:
Devs see CURA-6714, for our internal reference.

DivingDuck:
+1 
Would be nice having both information available.

serial8killer:
Please do this!!!.....

Ghostkeeper:
Duplicate of https://github.com/Ultimaker/Cura/issues/5744

Issue #5744, opened 2019-05-15
[FQ] More insformation beside the vertical layer slider


yet-another-average-joe:
**Application Version**
All

**Platform**
All

**Printer**
not relevant

**Steps to Reproduce**
not relevant

**Actual Results**
Only the current layer # is displayed on the left side of the Z slider.

**Expected results**
Two more informations 

- the height (thickness) of the current layer
- the Z position from the buildplate ; usefull to find glitches or holes in the original model

**Additional Information**
After slicing some models, holes can appear in some walls, due to thin walls that can't be sliced properly, but look fine in the CAD software. Having the Z position of the current layer would help a lot when fixing the original model.

Ghostkeeper:
See also https://github.com/Ultimaker/Cura/issues/5206 (not a duplicate though).

Ghostkeeper:
We discussed this and agree that it would be useful to optionally put the Z coordinate of the current layer there.

Layer thickness is not very useful though as usually all layers will just be the same, and it's not the information that you'd be adjusting with the layer view slider. We might as well show infill speed instead.

yet-another-average-joe:
Thanks !

Adapative layers is one of the greatest features ! Don't most users turn it on ? Results are amazing. Reading the chart and the model colors can be hard, and even impossible for people suffering of color blindness (dyschromatopsia)

Ok, speed and thickness are not adjusted with the slider. Maybe these informations could be displayed with an "analogic" cursor in the color charts ? Just an idea...

Off topic : I searched for the dev env you use, and there is no installation tutorial for Windows. Only for Mac and Ubuntu. And Cura can't be executed in a Virtual Box machine (due to supported GL version). Is this tutorial abandonned ?

Ghostkeeper:
The colour scale in layer view was designed to work well for people with red-green colour blindness (deuteranomaly). Other types of colour blindness are very rare, but a separate theme could be made for them that changes the colours if necessary.

> Adapative layers is one of the greatest features ! Don't most users turn it on ? Results are amazing.

About 0.2% of prints have this setting enabled. It tends to have some print quality issues like banding and pillowing. The skin thickness and support Z distance are inaccurate. I'd advise only using it if you don't use support in any case.

> And Cura can't be executed in a Virtual Box machine (due to supported GL version).

Half of the Cura team is on Windows right now. Also, one developer has VirtualBox set up to be able to test Linux-specific issues (or Windows-specific issues by seeing that they don't happen on Linux).

[...]

PR #21201, opened 2025-12-09
Show layer height in Simulation View slider labels

HellAholic:
# Description

Adds display of the current and minimum layer heights (in mm) to the Simulation View slider labels. Implements new properties and logic in SimulationView and SimulationViewProxy to retrieve accurate layer heights for both sliced and gcode data. Bumps plugin version to 1.1.0.
image
image
image
image
Adaptive layer height + Raft
image

## Type of change

# How Has This Been Tested?

- [X] Load model, Slice and inspect the heights (with extra settings such as Raft / Adaptive layer)
- [X] Load Gcode, inspect the heights
- [X] Initiate @GregValiant protocol

**Test Configuration**:
* Operating System:
* Windows 11

# Checklist:

### DIFF
--- a/plugins/SimulationView/SimulationView.py
+++ b/plugins/SimulationView/SimulationView.py
@@ -309,14 +309,15 @@ class SimulationView(CuraView):
             if not layer_data:
                 continue
             
-            # Check if we already have cached data for this node
-            current_node_id = id(node)
-            if self._layer_heights_cache_node_id == current_node_id and self._layer_heights_cache:
+            # Check if we already have cached data for this layer_data object
+            # Use id of the layer_data itself, not the node, since node might be reused
+            current_layer_data_id = id(layer_data)
+            if self._layer_heights_cache_node_id == current_layer_data_id and self._layer_heights_cache:
                 # Cache is still valid, no need to recalculate
                 return
             # Cache is invalid or empty, recalculate
             self._layer_heights_cache.clear()
-            self._layer_heights_cache_node_id = current_node_id
+            self._layer_heights_cache_node_id = current_layer_data_id
             
             has_gcode_decorator = node.getDecorator(GCodeListDecorator) is not None
             


GregValiant:
This is outstanding.  The term "Totally bad-ass" comes to mind.
I found 3 Feature Requests that relate to this:
#20347
[#5744](https://github.com/Ultimaker/Cura/issues/5744)
[#6179](https://github.com/Ultimaker/Cura/issues/6179)

plugins/SimulationView/SimulationViewProxy.py
```diff
@@ -46,6 +46,14 @@ def currentLayer(self):
     def minimumLayer(self):
         return self._simulation_view.getMinimumLayer()
 
+    @pyqtProperty(float, notify=currentLayerChanged)
+    def currentLayerHeight(self):
+        return self._simulation_view.getCurrentLayerHeight()
+
+    @pyqtProperty(float, notify=currentLayerChanged)
+    def minimumLayerHeight(self):
+        return self._simulation_view.getMinimumLayerHeight()
```
rburema, line 55: These will recalculate (go through the entire scene, for every layer) _each_ time, _every_ time you request an update.

Please cache this at the very least, but even better, just go through the layers once on receiving them, and then save all of that; then you can query that structure.

GregValiant:
"Feature Request" #7683 is slightly different.  They wanted the "Time Remaining" added to the slider as well.

HellAholic:
For displaying Time we need to return time per layer from the engine side, the layer message only has id, height, thickness.
The data we use for the print time estimation on the front end also doesn't contain per layer info, it only has the per feature times.
Parsing the gcode does not seem like a good approach so I don't think the time part can be done neatly within this feature from front-end only.

GregValiant:
Everything was working fine until I started playing with Aldo's plugins (trying to get them to work in 5.11).
I have not re-installed Cura, but I think I need to.
I did re-install the files for the Simulation View, and deleted "Sidebar GUI" and "Tabbed Settings" but I still get a label showing a height of "0.00mm" at all layers.

image

PartySausage:
it works for me with aldos tabbed & sidebar plugins apart from a niche case where if I'm in the preview tab, change a setting, hit slice, wait for the model to grey out then switch to the prepare tab while it's slicing. When the slice completes & I switch back to the preview tab wherever the slider is reports 0.00mm until I move the slider where it starts to display the height again. I didn't see this scenario as worthy of a bug post given the minimal impact

GregValiant:
I added the last commit and it looks good now.

PartySausage:
> I added the last commit and it looks good now.

Still working here too. The observation I made above still exists but I could live with that

I did tinker with my install of the SimulationSliderLabel.py so the layer height box matches the height of layer no. box. I don't know if it would work for everyone or if you think that is worthy of a change to the official version

image

[SimulationSliderLabel.zip](https://github.com/user-attachments/files/24159989/SimulationSliderLabel.zip)

GregValiant:
Another long-time feature request bites the dust.  Congratulations!!

(No one will ever know that it was all because you didn't want to write a calculator into "HellaFusion")

HellAholic:
I'm too lazy to write it in the plugin, I'll do many times the work just to avoid the work 🤣

PR #21282, opened 2026-01-07
Remove authentication flow when communicating with locally connected printer

casperlamboo:
# Description

Make local printer API up-to-date with printer by removing the authentication flow.

## Type of change

- [ ] Translations

# How Has This Been Tested?

**Test Configuration**:
* Operating System: MacOS ventura 13.3

# Checklist:


cura/Backups/Backup.py
```diff
@@ -36,9 +36,6 @@ class Backup:
     IGNORED_FOLDERS = []  # type: List[str]
     """These folders should be ignored when making a backup."""
 
-    SECRETS_SETTINGS = ["general/ultimaker_auth_data", "cluster_api/auth_ids", "cluster_api/auth_keys", "cluster_api/nonce_counts", "cluster_api/nonces"]
```
rburema, line 39: `SECRET_SETTINGS` should probably still exist, and be `["general/ultimaker_auth_data"]`. Only the `"cluster_api/<...>"` ones where added.

Unless we don't need that anymore either? I didn't look too deep into that, since I was _adding_ at the time, not _removing_.

later: Looked it up, and the variable is still used in the `Backup` class below. So even if it _shouldn't_ exist (which I doubt TBCH), that should still be taken care of (as in, also deleted).

plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
```diff
@@ -1,63 +1,39 @@
...
-class HttpRequestMethod(StrEnum):
-    GET = "GET",
-    HEAD = "HEAD",
-    POST = "POST",
-    PUT = "PUT",
-    DELETE = "DELETE",
-    CONNECT = "CONNECT",
-    OPTIONS = "OPTIONS",
-    TRACE = "TRACE",
-    PATCH = "PATCH",
```
rburema, line 40: This was part of a refactor, for my sanity. I'm ok with deleting it if it's easier though (or in any case). Just saying that we could keep this (& related) if we wanted to.

later: Hm, I see now that this wasn't completely done, since the manager still has those separate post/put/.../etc. methods. Then it makes more sense to remove it outright.

plugins/UM3NetworkPrinting/src/Network/ClusterApiClient.py
```diff
@@ -303,25 +190,9 @@ def parse() -> None:
...
-                    nonce_match = re.search(r'nonce="([^"]+)', str(reply.rawHeader(b"WWW-Authenticate")))
-                    if nonce_match:
-                        self._nonce = nonce_match.group(1)
-                        self._nonce_count = 1
-                        self._setLocalValueToPrefDict("cluster_api/nonce_counts", self._nonce_count)
-                        self._setLocalValueToPrefDict("cluster_api/nonces", self._nonce)
-                        CuraApplication.getInstance().savePreferences()
-                else:
-                    self._on_error(reply.errorString())
+                Logger.warning("Cluster API request error: %s", reply.errorString())
```
rburema: When writing new code, prefer using string interpolation (a.k.a. 'f-strings', at least that's what they're called in Python) instead:

```suggestion
                Logger.warning(f"Cluster API request error: {reply.errorString()}")
```

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
```diff
@@ -196,18 +196,26 @@ def _startPrintJobUpload(self, unique_name: str = None) -> None:
             return
         self._progress.show()
         parts = [
-            self._createFormPart("name=owner", bytes(f"user@{platform.node()}", "utf-8"), "text/plain"),
-            self._createFormPart("name=\"file\"; filename=\"%s\"" % self._active_exported_job.getFileName(),
-                                 self._active_exported_job.getOutput())
+            self._createFormPart("name=owner", bytes(self._getUserName(), "utf-8"), "text/plain"),
```
rburema: Please don't use the actual user-name here! That's how a large part of the whole thing got started in the first place! 😆

rburema (changes requested):
I've to send this back for now, sorry.

I think at least the secret setting list should be restored, and, maybe more importantly, we shouldn't re-introduce the sending of the username! (See comments for more details.)

plugins/UM3NetworkPrinting/src/Network/LocalClusterOutputDevice.py
casperlamboo: oops, this came from a merge conflict. Thought I picked the non-user sensitive one

PR #21286, opened 2026-01-09
CURA-12894 Add honeycomb and octagonal infill types

wawanbreton:
This PR adds 2 new infill types:
* Honeycomb
* Octagonal

As a bonus, it also adds the ability to set an angle on gyroid infill, since the engine now supports it.

By default, those infills have specific angles:
* Honeycomb: `[0, 60, 120]` => This is not strictly necessary but makes sure that the doubled-lines of each hexagon don't add up at the same position
* Octagonal: `[0, 90]` => This is more required, since `[0]` just gives a pattern with single non-joined lines

Requires https://github.com/Ultimaker/CuraEngine/pull/2272
CURA-12894

### DIFF
--- a/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
+++ b/plugins/UM3NetworkPrinting/src/Cloud/CloudApiClient.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Ultimaker B.V.
+# Copyright (c) 2026 UltiMaker
 # Cura is released under the terms of the LGPLv3 or higher.
 import json
 import urllib.parse
@@ -11,6 +11,7 @@ from PyQt6.QtCore import QUrl
 from PyQt6.QtNetwork import QNetworkRequest, QNetworkReply
 
 from UM.Logger import Logger
+from UM.TaskManagement.HttpRequestData import HttpRequestData
 from UM.TaskManagement.HttpRequestManager import HttpRequestManager
 from UM.TaskManagement.HttpRequestScope import JsonDecoratorScope
 from cura.API import Account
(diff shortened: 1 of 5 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -2197,7 +2197,7 @@
                 "infill_pattern":
                 {
                     "label": "Infill Pattern",
-                    "description": "The pattern of the infill material of the print. The line and zig zag infill swap direction on alternate layers, reducing material cost. The grid, triangle, tri-hexagon, cubic, octet, quarter cubic, cross and concentric patterns are fully printed every layer. Gyroid, cubic, quarter cubic and octet infill change with every layer to provide a more equal distribution of strength over each direction. Lightning infill tries to minimize the infill, by only supporting the ceiling of the object.",
+                    "description": "The pattern of the infill material of the print. The line and zig zag infill swap direction on alternate layers, reducing material cost. The grid, triangle, tri-hexagon, cubic, octet, quarter cubic, cross and concentric patterns are fully printed every layer. Gyroid, Honeycomb, Octagon, cubic, quarter cubic and octet infill change with every layer to provide a more equal distribution of strength over each direction. Lightning infill tries to minimize the infill, by only supporting the ceiling of the object.",
```
casperlamboo, line 2201: Can we do something smart with this when it comes to translations? seems a bit wasteful to pay for this string for all languages if we only change a list

wawanbreton: We could change the translations manually, but we would have to know the proper translation for each new word in every language. And the cost of translating a single label is really not much...

wawanbreton:
> <placeholder for PP&M requested changes>

<strike>There is actually no change on the Cura part of the PR :slightly_smiling_face: </strike>

Ok, that was a mistake :grin:

PR #21289, opened 2026-01-10
Add multiline support for custom G-code in FilamentChange script of PostProcessing plugin

hobbe:
# Description

This improves the Filament Change script of the Post Processing plugin to support G-Code Before and G-Code After multiline input using the text area introduced by #21298.

It also fixes the insertion of the before/after G-code at the layer level.

Requires:
- #21298 

## Type of change

- [ ] Translations

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B
- [x] Manual testing 

**Test Configuration**:

# Checklist:

### DIFF
--- a/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
+++ b/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
@@ -39,6 +39,42 @@ UM.Dialog
 
         anchors.fill: parent
 
+        // Helper function to check if a setting should use multiline text area
+        // Supports "multiline" or "@[multiline]" or "@[multiline, other] comment"
+        function isMultilineSetting(definition)
+        {
+            if (!definition || !definition.comments)
+            {
+                return false;
+            }
+            
+            var commentsLower = definition.comments.toLowerCase();
+            
+            // Simple format: exact match
+            if (commentsLower === "multiline")
+            {
+                return true;
+            }
+            
+            // Directive format: parse @[...] and check if multiline is in the list
+            var directiveStart = commentsLower.indexOf("@[");
+            var directiveEnd = commentsLower.indexOf("]", directiveStart);
+            if (directiveStart >= 0 && directiveEnd > directiveStart)
+            {
+                var directivesText = commentsLower.substring(directiveStart + 2, directiveEnd);
+                var directives = directivesText.split(",");
+                for (var i = 0; i < directives.length; i++)
+                {
+                    if (directives[i].trim() === "multiline")
+                    {
+                        return true;
+                    }
+                }
+            }
+            
+            return false;
+        }
+
         ButtonGroup
         {
             id: selectedScriptGroup
@@ -301,6 +337,10 @@ UM.Dialog
                     {
                         if (provider.properties.enabled == "True" && model.type != undefined)
                         {
+                            if (definition && definition.comments && definition.comments.toLowerCase() === "multiline")
+                            {
+                                return UM.Theme.getSize("standard_list_lineheight").height + UM.Theme.getSize("narrow_margin").height + (UM.Theme.getSize("setting_control").height * 3);
+                            }
                             return UM.Theme.getSize("section").height;
                         }
                         else
@@ -331,6 +371,19 @@ UM.Dialog
                         settingLoader.item.showLinkedSettingIcon = false
                         settingLoader.item.doDepthIndentation = false
                         settingLoader.item.doQualityUserSettingEmphasis = false
+                        // Pass properties explicitly to custom components that don't extend SettingItem
+                        if (settingLoader.item.hasOwnProperty("definition")) {
+                            settingLoader.item.definition = settingLoader.definition
+                        }
+                        if (settingLoader.item.hasOwnProperty("settingDefinitionsModel")) {
+                            settingLoader.item.settingDefinitionsModel = settingLoader.settingDefinitionsModel
+                        }
+                        if (settingLoader.item.hasOwnProperty("propertyProvider")) {
+                            settingLoader.item.propertyProvider = settingLoader.propertyProvider
+                        }
+                        if (settingLoader.item.hasOwnProperty("globalPropertyProvider")) {
+                            settingLoader.item.globalPropertyProvider = settingLoader.globalPropertyProvider
+                        }
                     }
 
                     sourceComponent:
@@ -348,7 +401,7 @@ UM.Dialog
                             case "bool":
                                 return settingCheckBox
                             case "str":
-                                return settingTextField
+                                return base.isMultilineSetting(definition) ? settingTextArea : settingTextField
                             case "category":
                                 return settingCategory
                             default:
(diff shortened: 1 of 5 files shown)


hobbe:
> Thanks for the PR.
> 
> Unfortunately, this change doesn’t add multi-line support as described. Introducing `|` as a line separator creates an additional, G-code–specific convention that users need to learn and that we then have to normalize internally as this separator would only work for this input field.
> 
> G-code already supports newlines, and users can currently enter `\n` to achieve multi-line input without any code changes. Using a non-standard separator therefore adds inconsistency without providing new functionality.

Thank you, @HellAholic, for the feedback.
It's true that I introduced a custom separator, but I was thinking more of a Cura FilamentChange specific convention, and documented in the tooltip.

Is '\n' a G-code convention or just a python convention?

Today, when I use '\n' in the current script, eg. `M300 S1000 P10000\nM300 S440 P500`, the G-code output is:
```
;BEGIN FilamentChange plugin:
M300 S1000 P10000\nM300 S440 P500
M600
;END FilamentChange plugin
```
Is this understood by the printer as?
```
;BEGIN FilamentChange plugin
M300 S1000 P10000
M300 S440 P500
M600
;END FilamentChange plugin
```

HellAholic:
Without adding a similar textArea as the start/end gcode in printer settings, you would need the `replace` regardless to add multi-line from the field currently available.
So you can add the textArea to the plugin (which is considerably more work) or instead of introducing a new delimiter `|` you can use the same known delimiter that the gcode supports and users are familiar with `\n` and add the support with that.
```
color_change = color_change + after_macro.replace('\\n', '\n') + "\n"
```
Which turns the `;FirstLine\n;SecondLine\n;ThirdLine` into:
```
;FirstLine
;SecondLine
;ThirdLine
```

hobbe:
> Without adding a similar textArea as the start/end gcode in printer settings, you would need the `replace` regardless to add multi-line from the field currently available. So you can add the textArea to the plugin (which is considerably more work) or instead of introducing a new delimiter `|` you can use the same known delimiter that the gcode supports and users are familiar with `\n` and add the support with that.
> 
> ```
> color_change = color_change + after_macro.replace('\\n', '\n') + "\n"
> ```
> 
> Which turns the `;FirstLine\n;SecondLine\n;ThirdLine` into:
> 
> ```
> ;FirstLine
> ;SecondLine
> ;ThirdLine
> ```

Thank you for the guidance. I went for the easy way, which should be sufficient for most use cases for now.

GregValiant:
I added "multi-line" support to a couple of other Post Processors.  I used a "comma" as the delimiter.

Within "Filament Change" the "multi-layer" support also uses commas as the delimiters between layer numbers.  For consistency, I would suggest using commas and then swapping them for "\n" in the script.

One thing about "Filament Change" that has always been incorrect is that it inserts the "color_change code" at the beginning of "data[item]".  That means that the ";LAYER:" line gets pushed down and is no longer the first line of the data[item].  Not good if there are other post-processers that are enabled below "Filament Change" and that are keying on ";LAYER:".

I would suggest something like this (comma delimiters and "insert" the new code rather than simply "appending" it).
(I changed the filename of this one for debugging purposes.)
A side effect is that this will work with One-at-a-Time print sequence.

[FilamentChangeGV.zip](https://github.com/user-attachments/files/24567939/FilamentChangeGV.zip)

```
;BEGIN FilamentChange
GVF
VAL
BBB
M600 E30.00 U300.00 X0.00 Y0.00
M300
M999
M888
;END FilamentChange
```

HellAholic:
https://github.com/Ultimaker/Cura/pull/21298
This adds the support for multiline text area field to the post processing plugin.
You would need to check what else is required in terms of the data item and layer info (@GregValiant please do the thing 😝, I did the other thing).

So if this goes through (main change is only the QML file that can be dropped into the postprocessing plugin in cura), then you would need to consider the data/layer and for each field that wants to be multiline just add to the settings item:
"comments": "multiline",

Works only for settings of type "str".

GregValiant:
"...I did the other thing..."
Yes you did and it managed to wipe out my wonderful "refresh" button that I had fallen in love with.

Now I need to re-add the button _AND_ add a PR for the real thing.

@hobbe Since @HellAholic was kind enough to roll "multi-line" into Cura proper, I'll fix the insertion method for the code in "Filament Change".  I'll let you know when it is ready.  I might be able to add a commit to this PR.  We'll see.

After that, _someone_ has made a change so that I have to go through all the other post processors and check for compatibility.

hobbe:
Thank you, this is great!
Since multiline is introduced in #21298, I will close this PR.
@GregValiant , you should probably add the "insert" fix in a dedicated PR.

hobbe:
OK, I misunderstood the purpose of #21298. It introduces the tools to add text area to script options.
I will update this PR to use it.

hobbe:
Back to draft until #21298 is merged.

hobbe:
> I added "multi-line" support to a couple of other Post Processors. I used a "comma" as the delimiter.
> 
> Within "Filament Change" the "multi-layer" support also uses commas as the delimiters between layer numbers. For consistency, I would suggest using commas and then swapping them for "\n" in the script.
> 
> One thing about "Filament Change" that has always been incorrect is that it inserts the "color_change code" at the beginning of "data[item]". That means that the ";LAYER:" line gets pushed down and is no longer the first line of the data[item]. Not good if there are other post-processers that are enabled below "Filament Change" and that are keying on ";LAYER:".
> 
> I would suggest something like this (comma delimiters and "insert" the new code rather than simply "appending" it). (I changed the filename of this one for debugging purposes.) A side effect is that this will work with One-at-a-Time print sequence.
> 
> [FilamentChangeGV.zip](https://github.com/user-attachments/files/24567939/FilamentChangeGV.zip)
> 
> ```
> ;BEGIN FilamentChange
> GVF
> VAL
> BBB
> M600 E30.00 U300.00 X0.00 Y0.00
> M300
> M999
> M888
> ;END FilamentChange
> ```

Hi @GregValiant 
I tested your "insert" fix in [FilamentChangeGV.zip](https://github.com/user-attachments/files/24567939/FilamentChangeGV.zip), but there is an issue on layer mapping.
If I insert FilamentChange for layer "10", G-code is inserted for layers 9 and layers 90 to 99.

GregValiant:
I missed a trick there.  The search has to be before splitting the layer so it includes the newline character.
Here is the new version.

[FilamentChange.zip](https://github.com/user-attachments/files/24689177/FilamentChange.zip)

@HellAholic - should @hobbe add this as a commit here or should there be a new PR?

HellAholic:
Don't see a limitations for what changes can be in the PR here. It's updating the same script, so should be fine here.

hobbe:
> I missed a trick there. The search has to be before splitting the layer so it includes the newline character. Here is the new version.
> 
> [FilamentChange.zip](https://github.com/user-attachments/files/24689177/FilamentChange.zip)
> 
> @HellAholic - should @hobbe add this as a commit here or should there be a new PR?

Thank you @GregValiant , it works.
I appended your fix to this PR in 6e4d46d.

HellAholic:
@hobbe The PR that updates the plugin is now merged in main, let me know if you're ready with this one.

hobbe:
> @hobbe The PR that updates the plugin is now merged in main, let me know if you're ready with this one.

All good !
Thanks

HellAholic (approved):

PR #21298, opened 2026-01-12
Add multiline text area for gcode post processing plugin

HellAholic:
# Description
image
image

Example modified script:
[FilamentChange.py](https://github.com/user-attachments/files/24571645/FilamentChange.py)

For the settings of type "str", adding the `comments: "multiline" ` or `comments: "@[multiline]"`  field will transform the textField into a textArea with multiline support.

Change scope is limited to the qml file in the plugin. There was an attempt to add the qml file to the resources but that would have required an API/SDK major version change

## Type of change

# How Has This Been Tested?

- [x] @GregValiant protocol initiated (probably the person most affected)
- [x] Used the changed script on updated plugin, entered values, sliced, examined gcode

**Test Configuration**:

# Checklist:

### DIFF
--- a/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
+++ b/plugins/PostProcessingPlugin/PostProcessingPlugin.qml
@@ -39,6 +39,42 @@ UM.Dialog
 
         anchors.fill: parent
 
+        // Helper function to check if a setting should use multiline text area
+        // Supports "multiline" or "@[multiline]" or "@[multiline, other] comment"
+        function isMultilineSetting(definition)
+        {
+            if (!definition || !definition.comments)
+            {
+                return false;
+            }
+            
+            var commentsLower = definition.comments.toLowerCase();
+            
+            // Simple format: exact match
+            if (commentsLower === "multiline")
+            {
+                return true;
+            }
+            
+            // Directive format: parse @[...] and check if multiline is in the list
+            var directiveStart = commentsLower.indexOf("@[");
+            var directiveEnd = commentsLower.indexOf("]", directiveStart);
+            if (directiveStart >= 0 && directiveEnd > directiveStart)
+            {
+                var directivesText = commentsLower.substring(directiveStart + 2, directiveEnd);
+                var directives = directivesText.split(",");
+                for (var i = 0; i < directives.length; i++)
+                {
+                    if (directives[i].trim() === "multiline")
+                    {
+                        return true;
+                    }
+                }
+            }
+            
+            return false;
+        }
+
         ButtonGroup
         {
             id: selectedScriptGroup
@@ -365,12 +401,7 @@ UM.Dialog
                             case "bool":
                                 return settingCheckBox
                             case "str":
-                                // Check if comments field indicates multiline
-                                if (definition && definition.comments && definition.comments.toLowerCase() === "multiline")
-                                {
-                                    return settingTextArea;
-                                }
-                                return settingTextField
+                                return base.isMultilineSetting(definition) ? settingTextArea : settingTextField
                             case "category":
                                 return settingCategory
                             default:


HellAholic:
Made a refactor and moved the qml component outside of the plugin.
The functionality and behavior remains the same before and after the change. The initial version was self contained (for testing and collaboration), the new version fits with the code guidelines.

hobbe:
Works as expected 👍 

One remark, the **tooltips** are not showing for the new text areas.
But there are no tooltips either for the component used in Start/End G-code in printer settings.

HellAholic:
The tooltips are omitted intentionally, this is based on the previous experience with the tooltips of start/end gcode in machine setting that was causing trouble when modifying the values.
Added a comment to clarify in the code for future reference. Signals are declared to satisfy the requirements of the loader but they are never emitted.

plugins/PostProcessingPlugin/PostProcessingPlugin.qml
```diff
@@ -348,6 +365,11 @@ UM.Dialog
                             case "bool":
                                 return settingCheckBox
                             case "str":
+                                // Check if comments field indicates multiline
```
rburema: So I assume this is `comments` so you don't have to add another property definition to the settings?

We don't really use the comments usually, so I think this might be OK, but maybe it'd be a good idea to make it so that you could still have a comment _after_ the multiline thing. Or maybe if we want to sneak in more of these things, have a system like:

```
...
"comment": "[[[multiline, futuretech, addedfeature]]] Actual comment with human words!"
...
```

Might be overkill though, just a thought!

plugins/PostProcessingPlugin/SettingTextArea.qml
```diff
@@ -0,0 +1,108 @@
...
+    // Standard setting item properties (required by loader but unused in this component)
+    property bool showRevertButton: false
+    property bool showInheritButton: false
+    property bool showLinkedSettingIcon: false
+    property bool doDepthIndentation: false
+    property bool doQualityUserSettingEmphasis: false
+
+    // Internal state tracking (unused but kept for potential future use)
+    property string textBeforeEdit
+    property bool textHasChanged
```
rburema, line 30: Since this part is in the 'main' resources, instead of the plugin, we'd be stuck with it until the next major version since it'd be in the API/SDK.

This does in fact go for the entire QML file.

I can imagine a SettingTextArea be useful in the general sense, so it's not wrong to place it here, but if you/we don't want to deal with the hassle it'd also be good to place it with the plugin.

HellAholic: I think a good middle ground could be adding the component, as is (so separate file), to the plugin, later on if needed we can move it to the application qml folder.
That would have the flexibility for future and the functionality for the current.
I'll make some changes :)
Thanks for the feedback.

HellAholic: It's placed separately with the plugin now, if later we decide that this is a nice to have we can just move the file into resources 👍

plugins/PostProcessingPlugin/PostProcessingPlugin.qml
HellAholic: Yes, mainly went with comment to avoid adding another property to settings.  

I went with the overkill way 😝 
Used "@[...]" as the marker for parsing, seemed more recognizable to me, open to suggestions if the `[[[` is more favorable, I can see that it would be more deliberate. So let me know if you have an opinion for one way or the other.

rburema: I think `@[...]` is totally fine. I can't really see people typing that by accident into a comment (at the very least not at the beginning), and `[[[...]]]` was just the thing that popped into my head as an example when I wrote the comment, so no loss there.

PR #21306, opened 2026-01-16
CURA-12556 formula parser in cura engine

wawanbreton:
This PR moves the GCode-template replacement algorithm from the Cura front-end to the engine. This allows for more flexibility because the engine has more information when resolving the templates. Especially for the extruders start/end GCodes, which can now be different between each layer.

Requires https://github.com/Ultimaker/CuraEngine/pull/2274
CURA-12556

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -6,7 +6,7 @@ requirements:
   - "curaengine/5.14.0-alpha.0@ultimaker/testing"
   - "cura_binary_data/5.14.0-alpha.0@ultimaker/testing"
   - "fdm_materials/5.14.0-alpha.0@ultimaker/testing"
-  - "dulcificum/5.10.0"
+  - "dulcificum/5.10.2"
   - "pysavitar/5.11.0-alpha.0"
   - "pynest2d/5.10.0"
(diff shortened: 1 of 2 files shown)


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -49,181 +45,6 @@ class StartJobResult(IntEnum):
...
-        expression: Optional[str] = code.group("expression")
-        extruder_nr_expr: Optional[str] = code.group("extruder_nr_expr")
-        end_of_line: Optional[str] = code.group("end_of_line")
-
-        # The following variables are not settings, but only become available after slicing.
-        # when these variables are encountered, we return them as-is. They are replaced later
-        # when the actual values are known.
-        post_slice_data_variables = ["filament_cost", "print_time", "filament_amount", "filament_weight", "jobname"]
-        if expression in post_slice_data_variables:
-            return f"{{{expression}}}"
```
casperlamboo, line 129: I don't see this logic in the changed engine code, don't we need it there aswell?

PR #21327, opened 2026-01-22
Rework PauseAtHeight.py

GregValiant:
A re-write of the classic Pause script.

# Description

Changes and additions:
1. Added more 'Pause Command' options for alternate firmware flavors.
2. Added 'Unload' option (prior to pause) and 'Reload' and 'Purge' options (after the pause).
3. Added 'Reason for Pause' options.
a. If 'Reason_for_pause' == 'Filament Change' then Unload, Reload, and Purge become available.
b. If 'Reason for Pause' == 'All Others' then the filament change options are hidden as they are not required for insertions or other pause events.
4. Added 'Multiple Pause Layers' option.
a. This works well when all pauses will have the same settings.  Delimit the layer numbers with commas.-  If pauses require different settings then add another instance of the script.
b. Multiple messages can be added (also delimited by commas).  Each message will be assigned to the equivalent pause layer.
c.  Example:  For pause layers 15,23,48 the messages could be:  White,Red,Blue
5. Added support for "Firmware Retraction"
6. Added support for 'One at a Time' print sequence.
a. Pauses can be at different layers in different parts.
b. All pause layers must be listed (use the Cura Preview layer numbers).
c. A pause at 'Layer:5' will only result in a pause at the first layer:5 encountered whereas pauses at '15,23,67' might be in different models.
d. Models can be skipped, or have pauses at different layers than other models, and some models could be different colors or a different material.
7. Added 'Flow Rate' option for 'Redo Layer'.
a. (Note that "Redo Layer" is disabled when in One-at-a-Time mode.)
8. Multi-extruder printers now use the Cura settings of the tool that is active when the pause occurs (retraction distance, retract and prime speeds, etc.)
9. The 'Stepper Timeout' has been de-confused (hopefully).

1. Obsolete:
        The "Retraction option" is removed.  Retractions are now handled in the script.

## Type of change

# How Has This Been Tested?

I have been using it for about 2 years now.

**Test Configuration**:
* Operating System:
* Windows 10 Pro
* Cura versions 4.13.1 through 5.12.1

# Checklist:

NOTE:  This version has some overlap with the current "PauseAtHeight".  That one could be made obsolete as we did with 'DisplayInfoOnLCD'.  The original remains but it just shows a message informing the user that they should use the new one.


GregValiant:
@HellAholic,
Do you think there is a chance of this being added(?) or as a replacement.

I changed the name of this one to try to avoid confusion but there would be the question of what to do with the original "PauseAtHeight".

HellAholic:
This one I would go with replacement since it's extending the `Pause At...` in my opinion. As what to do with regards to the behavior to support previous workflows, we can have the default values set in such a way that the behavior would be the same as the previous script.

GregValiant:
Crap.
I'll go over it and see what needs to be done.  Things like "Disarm Timeout" was in seconds but I changed that to minutes.  There might need to be some "conversions" needed.

GregValiant:
I appears to just have needed a couple of variables renamed so they match from old to new.
I have to wring it out some more, but I don't anticipate a lot of work there.

PartySausage:
I've performed some testing on this as I use pausing for filament swaps quite frequently. So far everything seems to work as it should, however I have a couple of implementation queries, which I was wondering if you could help me with for my understanding

The first is that I think there is a piece of obsolete code as I don't think the condition could ever be achieved

                    elif self.redo_layer and self.reason_for_pause == "reason_other":
                        self.pause_lines_list.append(f"M{extrusion_mode_numeric} ; Switch back to {extrusion_mode_string} E values")`

The second query surrounds the switching between Relative/Absolute Extrusion, where I was wondering if there was a reason why the script switches to Relative Extrusion even if Relative Extrusion is already being used then back to Relative Extrusion at the end of the pause. 
I was thinking that an easier solution would be to do the check whether Relative Extrusion was being used here` 
 
               # Don't change the extrusion mode if pause method is UM 'griffin'
                if pause_method != "griffin":
                    self.pause_lines_list.append("M83 ; Relative extrusion")

Then only switch to Relative Extrusion & back to Absolute Extrusion at the end if Absolute Extrusion was in use before the pause   

This would enable this code snipets to be simplified so that the append lines, which use _**extrusion_mode_numeric**_ would contain an explicit M82 command, which is only appended if Absolute Extrusion was in use before the pause                    

                    extrusion_mode_string = "absolute"
                    extrusion_mode_numeric = 82

                    relative_extrusion = Application.getInstance().getGlobalContainerStack().getProperty("relative_extrusion", "value")
                    if relative_extrusion:
                        extrusion_mode_string = "relative"
                        extrusion_mode_numeric = 83

                    if not self.redo_layer:
                        self.pause_lines_list.append(f"M{extrusion_mode_numeric} ; Switch back to {extrusion_mode_string} E values")

GregValiant:
1. When I started on the revision my incliniation was to simply remove "redo layer".  Later I saw a way to keep it in when "All Others" was true.  I went back and made the change.  As a result of that change, the "elif" statement can never be true ("redo_layer" is not available when "reason_other" is true).  I will remove that "If" statement.

2. That also looks like vestigial code BUT it insures that the extrusion mode is set correctly to "relative" when the print resumes so I'm inclined to leave it.  There may have been a condition I came across during debugging (with Firmware Retraction enabled?) where it was needed.

PartySausage:
Thanks for clearing that up. I only have the one flavour of printer to test this on so never no if there are other printer configurations where the extrusion switching or lack of could cause problems but what you've done is good defensive programming in case something throws a curveball in the future

HellAholic:
This one requires some work. The overall tone would be:
1. Keeping the old history, adding to it rather than replacing it at the top.
2. The __init__ function if I remember correctly is for when opening Cura, and the plugin loads, the script pops up, so it's a bit of a structural thing, otherwise you get an exception when you open cura with no printers for example because you initialize the script and the first thing you do is checking for global stack 😝 
3. `:param` addition is a good choice in terms of direction but it should be used for indicating the variables/parameters that you need to pass to the function, most of the functions get the self or data, so it's not a valid choice to add them to initialize for example. Some of the functions that could use the `:param` don't have it, for example the `_track_tool_nr()`
4. keeping the setting names the same as much as possible (backwards compatibility), labels and descriptions can be updated but the setting name itself should remain the same
5. Biggest part is the `_find_pause`; for a single operation of adding pause at height or layer, it's fine to have a for loop and then apply the logic, but as soon as you start extending the functionality, it becomes challenging to check every line of the gcode gainst 300 lines of code. At the very least I would extract each specific section into its own helper function

I'll do some clean up and post a version here as comment.

HellAholic:
[PauseAtHeight.py](https://github.com/user-attachments/files/29656832/PauseAtHeight.py)

Still need to test things out yourself and compare, but this is more of a structural guide.

GregValiant:
I'm getting this in a crash report.

```
  File "C:\Program Files\UltiMaker Cura 5.13.0\share\cura\plugins\PostProcessingPlugin\PostProcessingPlugin.py", line 288, in addScriptToList
    new_script.initialize()
  File "C:\Users\grego\AppData\Roaming\cura\5.13\scripts\PauseAtHeightHH.py", line 59, in initialize
    super().initialize()
  File "C:\Program Files\UltiMaker Cura 5.13.0\share\cura\plugins\PostProcessingPlugin\Script.py", line 50, in initialize
    self._definition.deserialize(json.dumps(setting_data))
  File "UM\Settings\DefinitionContainer.py", line 325, in deserialize
    parsed, is_valid = self.readAndValidateSerialized(serialized)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "UM\Settings\DefinitionContainer.py", line 275, in readAndValidateSerialized
    self._verifyJson(parsed)
  File "UM\Settings\DefinitionContainer.py", line 501, in _verifyJson
    raise IncorrectDefinitionVersionError("Definition uses version {0} but expected version {1}".format(json_dict["version"], self.Version))
UM.Settings.DefinitionContainer.IncorrectDefinitionVersionError: Definition uses version 3 but expected version 2
```

HellAholic:
It's the setting version, I made a booboo updating it to 3, it should be 2.
```
    def getSettingDataString(self) -> str:
        return """{
            "name": "Pause at Height",
            "key": "PauseAtHeight",
            "metadata": {},
            "version": 2,
            "settings":
            {
```
[PauseAtHeight.py](https://github.com/user-attachments/files/29657577/PauseAtHeight.py)

GregValiant:
In way of an explanation...
I have had projects that required several pauses (my personal record is 9).  It was irksome that 9 separate instances of "Pause" were required and that each instance had to be configured separately.  When the pause actually occurred, there was the manual filament handling for each change.  Hand "Purging" was required but for some printers was not really an option.  The script was also clumsy when used for "Insertions" of nuts or magnets.  

So I set out to include the _unload / reload / purge_ option to eliminate most of the manual filament handling, and allow for a slightly different configuration between insertions and filament changes.  I also wanted to have the ability to utilize a single instance of the PP even for pauses at multiple layers when they were all of the same configuration.
The "re-do layer" options was raw so now the user can adjust the flow and speed rather than trying to reprint the layer at 100% flow and 100% feedrate.  I was thinking that more people might take advantage of it if it didn't cause problems with the printer and/or an over-extruded layer in the print.

So far it seems to run the same with your revisions as it did in my effort.  I'll continue to wring it out.

@PartySausage if you would care to download and test HellaHolic's latest version it would be much appreciated.

PartySausage:
@GregValiant I've tested HellAholic's revised script with a several different theoretical scenarios using By Layer & By Height for All Others & Filament Change pauses with multiple pauses for each and haven't seen any hiccups. 👍👍👍

I've been using my own fork of your script with a couple of extra settings so I merged the new script in with that version, which given the number of changes took some doing. Anyway, the resulting G-Code files seem to contain everything they should

```
;
;TYPE:CUSTOM---------------; Pause at End of Preview Layer: 25 (End of G-Code LAYER: 24)
M83                        ; Relative Extrusion
G0 F300 Z6.2               ; Move Up to Clear Print
G0 F9000 X0 Y110           ; Move to Park Location
G0 F300 Z15                ; Move Up to Minimum Clearance Height
M104 S210                  ; Standby Temperature
M84 S3600                  ; Keep Motors Engaged for 60.0 Minutes
M412 S0                    ; Disable Filament Run-Out Sensor
M300                       ; G-Code Before Pause
M25                        ; PAUSE
M300                       ; G-Code After Pause
M104 S210                  ; Resume Temperature
G0 F9000 X87.701 Y85.097   ; Move to Resume Location
G0 F300 Z5.2               ; Move Down to Resume Height
G1 E0.1 F2400              ; Extra Prime
M83                        ; Switch Back to Relative Extrusion
G92 E0                     ; Reset Extruder Location
;--------------------------; Pause at Layer/Height: End of the Pause G-Code
;
```

GregValiant:
Th this afternoon's commit (on 7/5/2026) includes HellaHolic's suggestions and a couple of minor changes by myself.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -1,189 +1,515 @@
...
-
-from typing import List, Tuple
+from typing import Tuple
+from UM.Message import Message
 
 class PauseAtHeight(Script):
-    def __init__(self) -> None:
-        super().__init__()
+    #def __init__(self) -> None:
+    #    super().__init__()
```
HellAholic:
```suggestion
```

This part is technically not necessary and mostly for following the structure. So I would either include it or remove it completely 😝

Python automatically triggers the parent's __init__()

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -1,189 +1,515 @@
...
-                    "default_value": 0,
-                    "enabled": "pause_method not in [\\\"griffin\\\", \\\"repetier\\\"]"
+                    "default_value": 200,
+                    "enabled": "enable_pause_at_height and pause_method not in [\\\"griffin\\\", \\\"repetier\\\"]"
+                },
+                "tool_temp_overide_enable":
+                {
+                    "label": "Hidden setting Temp Overide Enable",
+                    "description": "Enable tool changes to overide the print temperature.",
+                    "type": "bool",
```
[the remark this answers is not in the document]
HellAholic: Case of the missing R. `overide` -> `override`

GregValiant:
@HellAholic 
Please run the last commit through the copilot again.  I think I caught all the requested changes.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +542,717 @@ def getSettingDataString(self) -> str:
...
+                if not layers_started:
+                    continue
+                if use_tool_temperature:
+                    if line.startswith("M109 S") or line.startswith("M104 S"):
+                        resume_print_temperature = self.getValue(line, "S")
+                if self.getValue(line, "Z") is not None:
+                    current_z = self.getValue(line, "Z")
+            if index >= pause_index - 1:
+                break
+        return layers_started, current_z, resume_print_temperature
```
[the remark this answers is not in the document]
HellAholic: This is intentional, no reason to track the inactive nozzle temperature which should be already at standby temperature before the pause.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +542,717 @@ def getSettingDataString(self) -> str:
...
+                self.pause_lines_list.append(f"G0 F{self.speed_z_hop} Z{working_z} {working_z_txt}")
+            # Unretract when necessary
+            if purge_amount != 0 and self.retraction_enabled:
+                self.pause_lines_list.append(f"G1 F{self.retraction_prime_speed} E{self.retraction_amount} ; Unretract")
+
+            # Change the extrusion mode as required
+            self.pause_lines_list.append(f"M{extrusion_mode_numeric} ; switch back to {extrusion_mode_string} ; E values")
+
+            # Reset extruder value to pre pause value
+            self.pause_lines_list.append(f"G92 E{current_e} ;Reset extruder")
```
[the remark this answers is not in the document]
HellAholic: @GregValiant 
technically valid but is a very low-impact edge case. Whether it's worth fixing is a judgment call — the fix is a one-liner, but the scenario it covers is nearly hypothetical.

GregValiant:
I'm sorry I got lost on this one.

I had a fiasco involving my water well and I've been busy making the repair.  PITA.  The original pressure tank lasted 32 years and that was considered the norm.  Now it's 3-5 years and they cost 10 times more than they did.  Junk.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +542,717 @@ def getSettingDataString(self) -> str:
...
+
+            if purge_amount != 0:
+                if firmware_retract and not is_retracted and self.retraction_enabled:
+                    retraction_count = 1 if control_temperatures else 3 # Retract more if we don't control the temperature.
+                    for i in range(retraction_count):
+                        self.pause_lines_list.append("G11 ; Unretract")
+                else:
+                    if not is_retracted and self.retraction_enabled:
+                        self.pause_lines_list.append(f"G1 F{self.retraction_prime_speed} E{self.retraction_amount} ; Unretract")
+
```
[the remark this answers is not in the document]
HellAholic: the unretract being inside if purge_amount != 0 seems like a real gap.
G11 count concern is overstated

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +542,717 @@ def getSettingDataString(self) -> str:
...
+        for prev_line in reversed(prev_lines):
+            current_e = self.getValue(prev_line, "E")
+            if re.search("G1 F(\d+\.\d+|\d+) E(-?\d+\.\d+|-?\d+)", prev_line) or "G10" in prev_line:
+                if is_retracted == None:
+                    is_retracted = True
+            if current_e is not None:
+                if is_retracted is None:
+                    is_retracted = False
+                break
+        return current_e, is_retracted
```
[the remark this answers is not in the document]
HellAholic: Not a practical concern given Cura's gcode structure

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +542,717 @@ def getSettingDataString(self) -> str:
...
+        else:
+            # When 'By Height' the string of heights must be converted to layers
+            pause_layer_list = self._pause_layer_from_height(data)
+
+        # Convert the pause layer numbers into their respective data[indexes] from the preview_layer_list.  This is a necessity for pausing in One at a Time mode.
+        pause_index_list = []
+        for p_layer in pause_layer_list:
+            if str(p_layer) in self.preview_layer_list:
+                pause_index_list.append(self.preview_layer_list.index(p_layer))
+        return pause_layer_list, pause_index_list
```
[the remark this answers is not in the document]
HellAholic: Valid but requires a specific triggering condition.
The length mismatch only causes wrong pairing when an invalid/missing layer is not at the tail of the list.

For typical user input like "5,10,999" (invalid layer at the end), the loop iterates over the shorter pause_index_list and indexes pause_layer_list[0]/[1] — both correct.

Mispairing only happens if the invalid layer is in the middle, e.g. "5,999,10" — then pause_layer_list[1] = "999" gets paired with the index for layer 10.

That's a real bug, but it requires both an invalid layer number AND non-sorted input from the user. Worth fixing (it's a one-liner), but it's not a common path.


So TLDR @GregValiant :
Can be fixed by sorting the user's input list.

```suggestion
        try:
            pause_layer_list.sort(key=lambda x: int(x.strip()))
        except ValueError:
            pass
        # Convert the pause layer numbers into their respective data[indexes] from the preview_layer_list.  This is a necessity for pausing in One at a Time mode.
        pause_index_list = []
        for p_layer in pause_layer_list:
            if str(p_layer) in self.preview_layer_list:
                pause_index_list.append(self.preview_layer_list.index(p_layer))
        return pause_layer_list, pause_index_list
```

HellAholic:
No rush, just went through the comments co-pilot made to close the ones out of context.

Nothing is being built to last anymore. Best of luck on the repairs. Hope both you and your wallet survive the fiasco.

GregValiant:
I haven't seen any comments here about the core of the re-design which was to allow a single instance to control multiple pauses, and to add more pause commands, and split off "Insertions" which don't require filament changes.  I also like the unload and reload/purge sequence as that was always manual and could be difficult on some machines.

I'll keep my hands out of it and let you work.  I have t sit here and provide consolation for my wallet.  It's pissed.

HellAholic:
Did some clean up and separated the big block of `if else` into separate methods. Take a look when things calmed down a bit.

[PauseAtHeight.py](https://github.com/user-attachments/files/30150972/PauseAtHeight.py)

GregValiant:
I'm playing with it.  So far I haven't been able to break it.

GregValiant:
It works for me.
#21458 is related to this PR.  I believe this version addresses the concerns of that PR.

PartySausage:
Same here no hiccups so far 👍

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -1,189 +1,517 @@
...
+                    "minimum_value_warning": 0.2,
+                    "maximum_value_warning": 15,
+                    "maximum_value": 20,
+                    "enabled": "enable_pause_at_height and head_park_enabled and pause_method not in ['griffin', 'repetier']"
                 },
                 "retraction_amount":
                 {
                     "label": "Retraction",
-                    "description": "How much filament must be retracted at pause.",
+                    "description": "Obsolete - Disabled. Kept for compatibility with older versions of the script.  The retraction amount is now determined by the Cura settings for the active tool at the pause.",
```
wawanbreton: What actually happens if we remove the obsolete settings ? On the principle, they should just be ignored.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +540,788 @@ def getSettingDataString(self) -> str:
...
+        """
+        If the 'Redo Layer' setting is active, prepends a copy of the previous layer to the
+        current pause layer so the filament is flowing again on resume.  Updates return_to_x/y
+        so the head moves back to the start of the redo layer instead of the end.
 
-        :param data: List of layers.
-        :return: New list of layers.
+        :return: (data, return_to_x, return_to_y, current_e, is_retracted)
+        """
+        if not (self.redo_layer and self.reason_for_pause == "reason_filament" and not self.one_at_a_time):
```
wawanbreton: Maybe this should also be disabled with multiple extruders ? If the last layer contained a tool change, are you not going to extrude with the wrong extruder ?

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +540,788 @@ def getSettingDataString(self) -> str:
...
+        for prev_line in reversed(prev_lines):
+            current_e = self.getValue(prev_line, "E")
+            if re.search("G1 F(\d+\.\d+|\d+) E(-?\d+\.\d+|-?\d+)", prev_line) or "G10" in prev_line:
+                if is_retracted == None:
+                    is_retracted = True
+            if current_e is not None:
+                if is_retracted is None:
+                    is_retracted = False
+                break
+        return current_e, is_retracted
```
wawanbreton, line 1102: Is this function working properly with relative extrusion ? Shouldn't it always return 0 in this case ?

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +540,788 @@ def getSettingDataString(self) -> str:
...
+        prev_lines = prev_layer.split("\n")
+        for current_line in prev_lines:
+            new_e = self.getValue(current_line, "E", current_e)
+            if new_e != current_e:
+                if re.search("G1 F(\d+\.\d+|\d+) E(-?\d+\.\d+|-?\d+)", current_line) or "G10" in current_line:
+                    if is_retracted is None:
+                        is_retracted = True
+                elif current_e is not None and is_retracted is None:
+                    is_retracted = False
+                current_e = new_e
```
wawanbreton: Isn't that the same as `getPreviousE` ?

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +540,788 @@ def getSettingDataString(self) -> str:
...
+        control_temperatures = self.global_stack.getProperty("machine_nozzle_temp_enabled", "value")
+        purge_speed = round(self.nozzle_size * self.PURGE_SPEED_MULTIPLIER)
+
+        if pause_method == "repetier":
+            if not is_retracted and self.retraction_enabled:
+                self.pause_lines_list.append(f"G1 F{self.retraction_retract_speed} E-{self.retraction_amount} ; Retract")
+            if park_enabled:
+                # Move the head to the park location
+                if current_z + move_z > self._machine_height:
+                    move_z = 0
```
wawanbreton:
```suggestion
                move_z = min(move_z, self._machine_height - current_z)
```
There are 2 other occurrences below, if you decide to apply it. They could also be moved out the `if` by the way, so that you would have it only once.

plugins/PostProcessingPlugin/scripts/PauseAtHeight.py
```diff
@@ -212,396 +540,788 @@ def getSettingDataString(self) -> str:
...
+        disarm_timeout = self.getSettingValueByKey("disarm_timeout") * 60
+        beep_at_pause = self.getSettingValueByKey("beep_at_pause")
+        beep_length = self.getSettingValueByKey("beep_length")
+        gcode_before, gcode_after = self.getGcodeBeforeAndAfter()
 
-                # Look for the feed rate of an extrusion instruction
-                if self.getValue(line, "F") is not None and self.getValue(line, "E") is not None:
-                    current_extrusion_f = self.getValue(line, "F")
+        if txt_msg:
+            self.pause_lines_list.append(f"M117 {txt_msg} ; Message to LCD")
```
wawanbreton: I know this is very paranoid and not a regression, but this allows gcode injection. Can we keep in mind to add some security over it ?

wawanbreton (approved):
Looks good to me ! I do have some remarks/questions, but the global quality is good, and from what I understand it already went through a lot of testing, so feel free to reject the suggestions.

wawanbreton:
Also, can someone explain the "redo layer" principle ? I like the idea of priming the nozzle, but if you replay the entire layer, that is definitely going to over-extrude and make it ugly :thinking:

HellAholic:
I made some tweaks but I am also not sure if it's functioning as expected after the changes 😝 So here is the changed version with:
- Z-clamp logic
- Potential Duplicate/incorrect E-tracking in _apply_redo_layer
- Multi-extruder + redo layer (guard)
[PauseAtHeight_V2.py](https://github.com/user-attachments/files/32002529/PauseAtHeight_V2.py)

The idea for keeping the obsolete settings was to prevent any breaking for old 3mf files, but they can be filtered out/removed if needed or if their removal doesn't cause a problem. They are disabled so they should not affect the script itself.

For the redo layer poking @GregValiant

GregValiant:
The "Redo layer" function is vestigial.  I have no idea what scenario would be required for it to be necessary.
99.9% of users are changing filament or making insertions.  If those work right, this should be good to go.

A full debugging with all options combinations is going to take a while.  I'll ignore "Redo" for now as it is an edge use case anyway.  It was working fine in previous testing.  I have never had occasion to use it in an actual project.

GregValiant:
The "E Reset" is now broken for the Redo Layer.  The "current_e is incorrect.  I'm looking.

HellAholic:
Oke, then we can keep the current implementation that is tested and clean things up in a separate later PR.

HellAholic:
Removed the redo layer and obsolete settings.
If there is a complaint (feature is used) during beta we can add the functionality back but at the moment we can't find a usecase for it and maintaining the code with potential issues is less than ideal.

PR #21341, opened 2026-01-28
[CURA-12917] Add a setting to continue bridging over support.

rburema:
This would help with keeping underside surfaces consistent, no matter where the support is generated.

See back-end PR: https://github.com/Ultimaker/CuraEngine/pull/2276
CURA-12917


casperlamboo (approved):

resources/definitions/fdmprinter.def.json
```diff
@@ -8948,6 +8948,16 @@
...
                     "settable_per_meshgroup": false
                 },
+                "bridge_over_support":
+                {
+                    "label": "Bridging over Support",
+                    "description": "Treat support as open air as far as bridging is concerned. Turning this off treats any supported areas as something that can be bridged to, and stops bridign over supported areas.",
+                    "type": "bool",
+                    "default_value": true,
+                    "resolve": "any(extruderValues('support_enable')) and any(extruderValues('bridge_settings_enabled'))",
+                    "settable_per_mesh": true,
```
HellAholic:
```suggestion
                    "resolve": "any(extruderValues('support_enable')) and any(extruderValues('bridge_settings_enabled'))",
                    "enabled": "any(extruderValues('support_enable')) and any(extruderValues('bridge_settings_enabled'))",
                    "settable_per_mesh": true,
```

Given the setting resolve value is linked to bridging and support, should it also be only visible when the conditions are met? Otherwise the user would enable the bridging over support while bridging is disabled.

casperlamboo: agreed!

rburema: Sounds reasonable, let's do it.

resources/definitions/fdmprinter.def.json
```diff
@@ -8948,6 +8948,17 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "bridge_over_support":
+                {
+                    "label": "Bridging over Support",
+                    "description": "Treat support as open air as far as bridging is concerned. Turning this off treats any supported areas as something that can be bridged to, and stops bridign over supported areas.",
```
HellAholic:
```suggestion
                    "description": "Treat support as open air as far as bridging is concerned. Turning this off treats any supported areas as something that can be bridged to, and stops bridging over supported areas.",
```
😛

wawanbreton (approved):
Approved on behalf of @rburema

PR #21345, opened 2026-01-29
Fix material diameter for method line

casperlamboo:
CURA-12937

### DIFF
--- a/resources/definitions/ultimaker_method_base.def.json
+++ b/resources/definitions/ultimaker_method_base.def.json
@@ -403,6 +403,7 @@
         "machine_start_gcode": { "default_value": "G0 Z20" },
         "machine_width": { "default_value": 283.3 },
         "material_bed_temperature": { "enabled": "machine_heated_bed" },
+        "material_diameter": { "default_value": 1.75 },
         "material_final_print_temperature":
         {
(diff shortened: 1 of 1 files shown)


HellAholic (approved):
Yes

saumyaj3:
A question. Why branch through main? This will be going in neoprep right? shouldn't we take the neoprep branch we already have: [NP-Branch-to-sync-neoprep-painting](https://github.com/Ultimaker/Cura/tree/NP-Branch-to-sync-neoprep-painting) as well to use this package incurator

casperlamboo:
> A question. Why branch through main? This will be going in neoprep right? shouldn't we take the neoprep branch we already have: [NP-Branch-to-sync-neoprep-painting](https://github.com/Ultimaker/Cura/tree/NP-Branch-to-sync-neoprep-painting) as well to use this package incurator

Because that branch can't be merged to main, and we need this change in main aswell.

I basically had two options
- merge to main, changes will be in neoprep with the next release
- merge to neoprep branch, cherry pick commit to main, create curator package, update neoprep
I choose for option 1 because I made a judgement call that the issue it fixes was not serious enough to warrant this whole flow. But it is mostly out of _laziness_.

PR #21349, opened 2026-01-30
CURA-10381 Add setting to enable infill splitting close to seam

wawanbreton:
This PR adds a setting to selectively enable the "Split Infill Close to Seam" feature, which allows reducing the travel move between the infill and outer wall by splitting the infill line close to wher it should end, providing a proper end position.

Requires https://github.com/Ultimaker/CuraEngine/pull/2279
CURA-10381

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -5597,11 +5597,12 @@
                         "zigzag": "Zig Zag",
                         "cross": "Cross",
                         "gyroid": "Gyroid",
+                        "lightning": "Lightning",
                         "honeycomb": "Honeycomb",
                         "octagon": "Octagon"
                     },
                     "default_value": "zigzag",
-                    "enabled": "support_enable or support_meshes_present",
+                    "enabled": "(support_enable or support_meshes_present) and support_infill_rate > 0",
                     "limit_to_extruder": "support_infill_extruder_nr",
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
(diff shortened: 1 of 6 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -2666,6 +2655,21 @@
...
+                    "label": "Infill Start Position",
+                    "description": "<ul><li><b>None:</b> No specific position will be used.</li><li><b>Random:</b> Use a random position for each infill area. This prevents one segment becoming the strongest, but it does so at the cost of an additional travel move.</li><li><b>End close to seam:</b> the infill lines will be ordered so that they end as close as possible to the next wall seam. If required, a line will be split to provide an appropriate ending position. This helps shortening the travel move between the infill and the wall start position, which should provide a better wall seam, at the cost of adding travel moves during infill print.</li></ul>",
+                    "type": "enum",
+                    "options":
+                    {
+                        "none": "None",
+                        "random": "Random",
+                        "close_to_wall_seam": "End close to seam"
+                    },
+                    "default_value": "shortest",
```
casperlamboo: This default value isn't one of the option, shoule be `none`?

```suggestion
                    "default_value": "none",
```

resources/definitions/ultimaker_s8.def.json
```diff
@@ -578,6 +578,7 @@
             "maximum_value_warning": 300,
             "value": "speed_wall"
         },
+        "split_infill_close_to_seam": { "default_value": true },
```
casperlamboo: This setting was defined while developing the setting right?

```suggestion
        "infill_start_position": { "default_value": "close_to_wall_seam" },
```

resources/definitions/fdmprinter.def.json
wawanbreton: Oh yes indeed, I changed the value but forgot to change the default 🤦

resources/definitions/ultimaker_s8.def.json
wawanbreton: That was the first version, this one is obsolete. I will remove this value definition since it doesn't look useful specifically for this printer anyway.

PR #21378, opened 2026-02-05
Paint by face concept

HellAholic:
Proof of concept for paint by face

paint_face_concept

Requires https://github.com/Ultimaker/libUvula/pull/9 & https://github.com/Ultimaker/Uranium/pull/1051


plugins/PaintTool/PaintTool.py
```diff
@@ -280,6 +305,39 @@ def get_projected_on_plane(pt: numpy.ndarray) -> numpy.ndarray:
...
+        mesh_indices = self._mesh_transformed_cache.getIndices()
+        if mesh_indices is None:
+            mesh_indices = numpy.array([], dtype=numpy.int32)
+
+        coplanar_faces = uvula.getConnectedFaces(self._mesh_transformed_cache.getVertices(),
+                                                 mesh_indices,
+                                                 self._node_cache.getMeshData().getUVCoordinates(),
+                                                 self._node_cache.getMeshData().getFacesConnections(),
+                                                 self._view.getUvTexDimensions()[0],
+                                                 self._view.getUvTexDimensions()[1],
```
rburema: This is such a small thing, but, you could also get this before the call like:
```
tex_w, tex_h = self._view.getUvTexDimentions()
```
(I know it's the same as in `.project` and it won't make much difference, but still...)

plugins/PaintTool/PaintTool.qml
```diff
@@ -193,28 +193,52 @@ Item
...
+
+            BrushShapeButton
+            {
+                id: buttonBrushFace
+                shape: Cura.PaintToolBrush.FACE
+
+                text: catalog.i18nc("@action:button", "Face")
+                toolItem: UM.ColorImage
+                {
+                    source: UM.Theme.getIcon("LayFlatOnFace")
```
rburema, line 205: Hahaha 😆

plugins/PaintTool/PaintTool.qml
```diff
@@ -193,28 +193,52 @@ Item
...
+            }
         }
 
-        UM.Label
+        UM.TooltipArea
         {
-            text: catalog.i18nc("@label", "Brush Size")
+            width: childrenRect.width
+            height: childrenRect.height
+            text: shapeSizeSlider.isFaceAngle ? catalog.i18nc("@tooltip", "Select neighboring faces whose angle differs by no more than this value.") : ""
```
rburema: Will this cause a small tooltip with no text to show up when `isFaceAngle` is off? If not then that's fine, but if so, we can also just write down what the 'Brush Size' does -- in fact, maybe we should in any case, because if a popup doesn't show up for that (and ppl try it), why would they (then also) try to hover over for the angle-version?

rburema (approved):
LGTM!

plugins/PaintTool/PaintTool.qml
HellAholic: image
image

It does not show an empty pop up. We could add the text for the brush size indeed 🤔

plugins/PaintTool/PaintTool.py
HellAholic: Done in https://github.com/Ultimaker/Cura/pull/21378/changes/927f744bb04d81fa8b04238efb6ba0694b84bfc0

plugins/PaintTool/PaintTool.qml
HellAholic: Added the tooltip for brush size for consistency in https://github.com/Ultimaker/Cura/pull/21378/changes/7cd0e8dfbceb9f47befe3567505d3dbb21db4e67

Issue #21377, opened 2026-02-05
Paint material using face selection


Thomas-QCD:
### Is your feature request related to a problem?

Im frustrated with the painting tool. I paint with my cursor, typically for the purpose of assigning separate material. The paint tool is difficult to use, and there does not seem to exist an undo function for this

### Describe the solution you'd like

Since there exists a feature to select face to align to build plate, I can assume code already exists to select a single face of the model, and if thats the case, it would seem easier to select faces of the model to "paint" material on

### Describe alternatives you've considered

No alternatives exist

### Affected users and/or printers

All users

GregValiant:
The Paint tool is a new feature and will have some growing pains.
I don't use the Material capability myself so this is just me looking in from the outside.

- Per this feature request - Ability to pick a face to paint.  Large irregular areas can take a long time with the brush.  A "bucket" tool for filling larger areas?  
  - This could make painting material more accurate, and quicker.
  - What happens when the painted surface goes all the way to the edge?  Would it stop short to create an "inset" of the 2nd material, or should it include the walls around the top surface?  That would cause the material on the sides to change.  It would be changing materials every layer.
- Ability to pick a depth.
  - The "depth of material" seems to be hard coded at 4mm(?)  If it could be user a controlled depth (maybe with a minimum of 1 layer height??) it would add flexibility for the user.  
- Ability to press and hold the "Shift" key to get orthogonal moves from the brush (ala AutoCAD and others) .  Trying to free-hand straight lines with the mouse is tough.

@wawanbreton , Any thoughts?

HellAholic:
For the Depth and precision, they are available to users on the front-end. in the settings tab for the material painting:
<img width="515" height="333" alt="Image" src="https://github.com/user-attachments/assets/92f230af-daca-472f-ac57-4d463512e297" />

Paint by face proof of concept based on the current structure, 
https://github.com/Ultimaker/Cura/pull/21378

wawanbreton:
Hi, I'm very glad that you made some feedback on this feature 🙂 

* The per-face painting feature is a must-have indeed. @HellAholic made a POC that is super nice, however it needs some rework because the heavy calculation is done in the Python front-end and could be very slow on large models.
* Depth picking: already available indeed 🙂 however the fact that you didn't find it makes me think that the naming of the setting may be non-intuitive. Would you have a better suggestion ? Maybe display it directly on the material painting panel ?
[...]

PR #21393, opened 2026-02-11
Allow painting straight lines by holding the shift key

wawanbreton:
[Kooha-2026-02-11-09-05-30.webm](https://github.com/user-attachments/assets/e9157d2f-f0a1-4d61-80eb-d0568d74cbe4)

You can now draw straight lines by holding the shift key, and mix free drawing with straight lines.

Requires https://github.com/Ultimaker/Uranium/pull/1024
Suggested in https://github.com/Ultimaker/Cura/issues/21377


PartySausage:
I tested this out on my 5.11.0 build with the updated controller.py but finding that when I hold the shift key and select the second point to apply the straight line the view is reverting to solid view & exiting paint mode

```
2026-02-11 13:22:13,226 - DEBUG - [MainThread] UM.Controller._updateActiveView [150]: Setting active view to PaintTool
2026-02-11 13:22:13,248 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [62]: Loading [C:\Program Files\UltiMaker Cura 5.11.0\share\cura\plugins\PaintTool\paint.shader]...
2026-02-11 13:22:13,291 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [62]: Loading [C:\Program Files\UltiMaker Cura 5.11.0\share\cura\resources\shaders\camera_distance.shader]...
2026-02-11 13:22:33,835 - DEBUG - [MainThread] UM.Controller._updateActiveView [150]: Setting active view to SolidView
2026-02-11 13:22:33,856 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [62]: Loading [C:\Program Files\UltiMaker Cura 5.11.0\share\cura\resources\shaders\xray.shader]...
2026-02-11 13:26:16,784 - WARNING - [MainThread] UM.Qt.QtApplication.__onQmlWarning [449]: file:///C:/Program Files/UltiMaker Cura 5.11.0/share/cura/plugins/PaintTool/PaintTool.qml:191:17: Unable to assign [undefined] to int
2026-02-11 13:26:16,786 - DEBUG - [MainThread] UM.Controller._updateActiveView [150]: Setting active view to PaintTool
2026-02-11 13:26:21,797 - DEBUG - [MainThread] UM.Controller._updateActiveView [150]: Setting active view to SolidView
2026-02-11 13:26:21,814 - DEBUG - [MainThread] UM.View.GL.ShaderProgram.load [62]: Loading [C:\Program Files\UltiMaker Cura 5.11.0\share\cura\resources\shaders\xray.shader]...
```

wawanbreton:
> I tested this out on my 5.11.0 build with the updated controller.py but finding that when I hold the shift key and select the second point to apply the straight line the view is reverting to solid view & exiting paint mode

Then you must have updated only the code in the Cura part, you also need to apply this change: https://github.com/Ultimaker/Uranium/pull/1024. I just found out that it causes some behavior changes in the object selection though, so will provide a fix later.

PartySausage:
> Then you must have updated only the code in the Cura part, you also need to apply this change: [Ultimaker/Uranium#1024](https://github.com/Ultimaker/Uranium/pull/1024). I just found out that it causes some behavior changes in the object selection though, so will provide a fix later.

Yes I applied that, unless the previous version is cached somewhere and it's not being updated

wawanbreton:
> Yes I applied that, unless the previous version is cached somewhere and it's not being updated

Indeed, if I remember correctly, on a Windows installation part of the application runs from a frozen version (even if the Python source is also provided). The `PaintTool` can be updated because it is an "embedded plugin", so not part of the frozen core. But the `Controller` is part of the core, so there is no way to update it, at least none that I know of. Maybe @rburema or @HellAholic know a way ?

PartySausage:
> > Yes I applied that, unless the previous version is cached somewhere and it's not being updated
> 
> Indeed, if I remember correctly, on a Windows installation part of the application runs from a frozen version (even if the Python source is also provided). The `PaintTool` can be updated because it is an "embedded plugin", so not part of the frozen core. But the `Controller` is part of the core, so there is no way to update it, at least none that I know of. Maybe @rburema or @HellAholic know a way ?

Ahh That'll be the culprit then... If it's not updatable I'll just have to be patient & wait for 5.12

rburema:
I think it might be possible, but not without a whole lot of hassle, and at the risk of breaking the current install. It's better to wait for 5.12 I think, as these changes have been made on top of the in-development 5.12 as well.

PartySausage:
> I think it might be possible, but not without a whole lot of hassle, and at the risk of breaking the current install. It's better to wait for 5.12 I think, as these changes have been made on top of the in-development 5.12 as well.

I think that's a very good idea. Looking forward to trying some of the new features I've seen popping up :-)

wawanbreton:
> I think it might be possible, but not without a whole lot of hassle, and at the risk of breaking the current install. It's better to wait for 5.12 I think, as these changes have been made on top of the in-development 5.12 as well.

I don't will it will make it into 5.12 though. We already have passed feature freeze and this is definitely a new feature :grimacing: so I am afraid you will have to wait until 5.13...

PartySausage:
> > I think it might be possible, but not without a whole lot of hassle, and at the risk of breaking the current install. It's better to wait for 5.12 I think, as these changes have been made on top of the in-development 5.12 as well.
> 
> I don't will it will make it into 5.12 though. We already have passed feature freeze and this is definitely a new feature 😬 so I am afraid you will have to wait until 5.13...
😭

HellAholic:
I have a separate installation of Cura that I use for development, the changes that are in UM require the `conan create . --build=missing` then I copy over the package that is built manually to that installation. It's just a lazy way of running from source.
So....

I just replaced the key 😝 instead of shift as the modifier, I used Q 🤷 
```
        # Handle Q key for drawing straight lines
        if event.type == Event.KeyPressEvent:
            from UM.Event import KeyEvent
            if cast(KeyEvent, event).key == Qt.Key.Key_Q:
                self._q_key_pressed = True
                return True
        elif event.type == Event.KeyReleaseEvent:
            from UM.Event import KeyEvent
            if cast(KeyEvent, event).key == Qt.Key.Key_Q:
                self._q_key_pressed = False
                return True
```
and the shift pressed was replaced with q_pressed 

```
q_pressed = self._q_key_pressed
```

So if you are not hellbent on using shift for testing, you can use the file below, and hold Q
Changes are then contained to Cura plugin only.
[PaintTool_Q_Key.zip](https://github.com/user-attachments/files/25243063/PaintTool_Q_Key.zip)

image

PartySausage:
> So if you are not hellbent on using shift for testing, you can use the file below, and hold Q
> Changes are then contained to Cura plugin only.
> [PaintTool.zip](https://github.com/user-attachments/files/25239063/PaintTool.zip)

Awesome. Thank You.  😊

From the brief testing I've done so far I do have a couple of observations, which I don't know how relevant they are to the pukka version 
> After clicking the toolbar button, The paint tool doesn't fully activate until the Q key at least momentarily pressed & you can't paint either freehand or straight lines until you've done that
> Similarly, You can't manipulate the view with the mouse or keyboard until the Q button is momentarily pressed. Although menus, icons, buttons, right click, dropdowns & fields still respond the mouse click events

Once the Q button has been pressed the paint tool activates & works as normal in the active session, even when loading new models as long as it's in the same session

Maybe these are the 'behaviour changes' @wawanbreton was referring to earlier

HellAholic:
@PartySausage That one is on me.
The value is not initialized for the key_press bool
open the file and add `self._q_key_pressed: bool = False` to line 70
Or 👇 
[PaintTool_Q_Key.zip](https://github.com/user-attachments/files/25242866/PaintTool_Q_Key.zip)

Alternative version with Ctrl instead of Q
[PaintTool_Ctrl_key.zip](https://github.com/user-attachments/files/25243786/PaintTool_Ctrl_key.zip)

PartySausage:
That's great work thanks for the quick turn around on the fix. This is going to save a load of trial & error painting seams. 

If I was to ever suggest additional improvements to this feature it would be to have a single point click option which then automatically paints a full vertical seam for the height of the model

HellAholic:
In scope of the change, I have the following feedback
- The last coordinate is shared between Seam and Material
  Could we use a dict to store the last coords per paint type allowing for future expansion: `dict[str, Optional[numpy.ndarray]] = {}`
  
- Shift vs Ctrl discussion (or both 😝)
  Shift for drawing lines at  `(π/4)*Int` angle intervals, Ctrl for Flat(ish) line Vs Free-Form

- The line connects from the last clicked coordinates to the current mouse pointer coordinates. On curved/sloped models we might need a different approach 👇 
drawing_line_curved_surface

wawanbreton:
>     * The last coordinate is shared between Seam and Material
>       Could we use a dict to store the last coords per paint type allowing for future expansion: `dict[str, Optional[numpy.ndarray]] = {}`
Makes sense, I overlooked that

>     * Shift vs Ctrl discussion (or both 😝)
>       Shift for drawing lines at  `(π/4)*Int` angle intervals, Ctrl for Flat(ish) line Vs Free-Form
Drawing π/4-aligned lines could be interesting indeed, that comes together with the idea of drawing a full vertical line

>     * The line connects from the last clicked coordinates to the current mouse pointer coordinates. On curved/sloped models we might need a different approach 👇
That would be nice indeed, but painting a consistent line over a curved 3D mesh is much more complex than our currrent approach :stuck_out_tongue:

wawanbreton:
> Once the Q button has been pressed the paint tool activates & works as normal in the active session, even when loading new models as long as it's in the same session
> 
> Maybe these are the 'behaviour changes' @wawanbreton was referring to earlier

Nope, but the Q key is not handled as the Shift key, so with Shift you should not have this behavior.

plugins/PaintTool/PaintTool.py
```diff
@@ -73,6 +72,7 @@ def __init__(self, view: PaintView) -> None:
         self._mouse_held: bool = False
 
         self._last_world_coords: Optional[numpy.ndarray] = None
+        self._last_clicked_coords: Dict[numpy.ndarray] = {}
```
rburema: Typing: A `Dict` type mapping has a from and a to, but here only one of the values is present.

rburema (approved):
LGTM! (Except a small typing thing, see below.)

plugins/PaintTool/PaintTool.py
wawanbreton: Oh yes indeed, I totally missed that :smiley: Fixed in https://github.com/Ultimaker/Cura/pull/21393/changes/e9c233cd9088c017e244928a728f34797a95d582

PR #21399, opened 2026-02-11
CURA-12853 add message painting in compatibility mode

wawanbreton:
Display a message in the painting panel when trying to use it in compatibility mode, to indicate that this is not compatible. Also display the normal model view instead of invisible models.

CURA-12853

Note: there was an attempt to make it compatible in this branch: https://github.com/Ultimaker/Cura/tree/CURA-12853-painting-doesnt-work-in-opengl-compatibility-mode but we gave up because the compatibility shader was missing some critical functions. It may be possible to make it work with a deeper work on the shader.

### DIFF
--- a/plugins/PaintTool/PaintTool.qml
+++ b/plugins/PaintTool/PaintTool.qml
@@ -343,7 +343,7 @@ Item
         UM.Label
         {
             anchors.fill: parent
-            text: catalog.i18nc("@label", "Painting is not available in OpenGL compatibility mode. Please use a different system.")
+            text: catalog.i18nc("@label", "Painting is not available on this device. Your graphics card or drivers do not fully support it. Updating your graphics drivers may enable this feature.")
             verticalAlignment: Text.AlignVCenter
             horizontalAlignment: Text.AlignHCenter
         }


fieldOfView:
For consistency with other things that are not available in compatibility mode, shouldn't the tool just be unavailable? Otherwise I would say the rotate tool panel would need similar messaging.

wawanbreton:
> For consistency with other things that are not available in compatibility mode, shouldn't the tool just be unavailable? Otherwise I would say the rotate tool panel would need similar messaging.

Hmm, thanks for the remark, that makes some sense indeed. In the case of the rotation, it is not the full feature that is unavailable though, just the selection by face. For the painting, you really have nothing. And completely disabling the tool would make it less explicit why it is missing. The same argument goes for the rotation, so I will discuss it with the team tomorrow.

fieldOfView:
I agree with what you say. I am all for the message, but even more for consistency ;-)

HellAholic:
Template of the message for future reference:
`f”{feature} is not available on this device. Your graphics card or drivers do not fully support it. Updating your graphics drivers may enable this feature.”`

If we later decide to add a message to the scene to inform the user when they are in compatibility mode with read more option we can utilize something such as:
`f"Cura is running in compatibility mode on this device because your graphics card or drivers do not meet the full requirements. As a result, the following {[features]} are not available. Updating your graphics drivers may enable these features."`

PR #21446, opened 2026-03-02
pt_BR translations for Cura 5.12

Patola:
# Description

This updates the Brazilian Portuguese strings for Cura 5.12

## Type of change

- [X] Translations

# How Has This Been Tested?

- [X] Compiled and ran locally on my system

**Test Configuration**:
* Operating System: Archlinux

# Checklist:

### DIFF
--- a/conandata.yml
+++ b/conandata.yml
@@ -275,7 +275,7 @@ pycharm_targets:
 pip_requirements_core:
   any_os:
     charon:
-      url: "git+https://github.com/ultimaker/libcharon@master/s-line#egg=charon"
+      url: "git+https://github.com/ultimaker/libcharon@cura_main#egg=charon"
     certifi:
       version: "2023.5.7"
       hashes:
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -12,6 +12,7 @@ import xml.etree.ElementTree as ET
 
 from UM.Math.AxisAlignedBox import AxisAlignedBox
 from UM.Math.Vector import Vector
+from UM.Operations.AddSceneNodeOperation import AddSceneNodeOperation
 from UM.Util import parseBool
 from UM.Workspace.WorkspaceReader import WorkspaceReader
 from UM.Application import Application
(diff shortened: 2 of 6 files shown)


wawanbreton (approved):
Looks good !
Review is not super easy though, many strings were split in multiple lines and accounted as changes. although the text is the same. Could you possibly setup your editor to avoid this ?

wawanbreton:
Also, not critical, but some new settings names are a bit long and don't fit the default UI on a single line. We usually make the settings names short enough that they can fit. If you can think of a shorter name, that would be perfect. Otherwise it is also fine, the full name is still visible.
image

Patola:
> Looks good ! Review is not super easy though, many strings were split in multiple lines and accounted as changes. although the text is the same. Could you possibly setup your editor to avoid this ?

I didn't notice that. I have been using `lokalize` to do the translation. Will try and find such setting.

Patola:
> Also, not critical, but some new settings names are a bit long and don't fit the default UI on a single line. We usually make the settings names short enough that they can fit. If you can think of a shorter name, that would be perfect. Otherwise it is also fine, the full name is still visible. image

Yeah, Portuguese has this issue of long words. At the top of my head I can't see an easy way to shorten that particular expression without hurting comprehension, but I will try to think of a way.

Patola:
> Looks good ! Review is not super easy though, many strings were split in multiple lines and accounted as changes. although the text is the same. Could you possibly setup your editor to avoid this ?

I saw that now and it is a pain, it broke many source strings into multiline strings. I am very sorry for this, I've been using lokalize for a long time and this might be new behavior.

I found out that if I create a project for it and set line wrap to a big value (I've used 8000), this issue does not happen anymore. Should I try and recreate the translations from the previous version? I had no idea this was happening. However the release date of Cura 5.12 is tomorrow, no? Very tight timeframe.

wawanbreton:
> I saw that now and it is a pain, it broke many source strings into multiline strings. I am very sorry for this, I've been using lokalize for a long time and this might be new behavior.
> 
> I found out that if I create a project for it and set line wrap to a big value (I've used 8000), this issue does not happen anymore. Should I try and recreate the translations from the previous version? I had no idea this was happening. However the release date of Cura 5.12 is tomorrow, no? Very tight timeframe.

Don't worry, this is fine. We also use lokalise and I have a Python script to sanitize the output PO files because gettext and Lokalise have a tendency to format them in a very different way. But next time I will extract the translations, it will be fixed. It is annoying only for the review, which is done now :slightly_smiling_face: 

> Yeah, Portuguese has this issue of long words. At the top of my head I can't see an easy way to shorten that particular expression without hurting comprehension, but I will try to think of a way.

Same with German :sweat_smile: I will let the PR open until the very last moment, in case you find something.

PR #21448, opened 2026-03-02
Defer 3MF node loading until machine activation

HellAholic:
Postpone reading/adding 3MF scene nodes until after the target machine/UCP is activated to ensure correct build volume and settings are applied. Adds a deferred filename field and stores the file during initial load, then reads and adds nodes in new _loadAndAddNodesToScene using AddSceneNodeOperation after activation. Fixes _updateActiveMachine calls to use the reader's is_ucp state, resets is_ucp only after nodes are added, and clears the deferred filename in _clearState. Also updates WorkspaceDialog and QML logic so machine selection/combobox visibility behave correctly for UCP files.

The old path got scene-adding "for free" by returning nodes from read(); the deferred path has to do it explicitly because the framework's return-based mechanism is bypassed.

CURA-13013

### DIFF
--- a/plugins/3MFReader/ThreeMFWorkspaceReader.py
+++ b/plugins/3MFReader/ThreeMFWorkspaceReader.py
@@ -12,6 +12,7 @@ import xml.etree.ElementTree as ET
 
 from UM.Math.AxisAlignedBox import AxisAlignedBox
 from UM.Math.Vector import Vector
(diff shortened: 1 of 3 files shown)


wawanbreton (approved):
Band-aid-y code, but does properly fix the loading. Not much better we can do before release.

PR #21456, opened 2026-03-06
CURA-13005 Properly display empty layers

wawanbreton:
Since the associated engine changes now allow fully empty layers being generated, we now handle this case properly in the UI and display a disabled slider when the layer is empty.

Comes with https://github.com/Ultimaker/CuraEngine/pull/2297
CURA-13005


rburema (approved):
LGTM. Do we want this in main or in the 5.13? (I think main is sensible in this case...)

plugins/SimulationView/LayerSlider.qml
```diff
@@ -219,6 +219,11 @@ Item
             // Normalize values between range, since using arrow keys will create out-of-the-range values
             value = sliderRoot.normalizeValue(value)
 
+            if(value == getValue())
```
casperlamboo:
```suggestion
            if (value == getValue())
```
Missing space

casperlamboo (approved):

PR #21461, opened 2026-03-08
Update fdmprinter.def.json Initial and Final print temperatures

GregValiant:
# Description

The "Initial Print Temperature" and "Final Print Temperature" are not used by single extruder machines but both settings remain "enabled" and they can cause some confusion. This PR disables the settings when "machine_extruder_count == 1" so that they are not visible when a machine has a single extruder.

This PR also includes minor changes to the descriptions.

## Type of change

- [ X] Printer definition file(s)

## Note
I'll leave this as a Draft as it involves "fdmprinter.def.json".

**Test Configuration**:
* Operating System:
Win 10 Pro

# Checklist:

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2728,7 +2728,7 @@
                 "material_initial_print_temperature":
                 {
                     "label": "Initial Printing Temperature",
-                    "description": "The minimal temperature while heating up to the Printing Temperature at which printing can already start.",
+                    "description": "A multi-extruder setting. The minimal temperature while heating up to the Printing Temperature at which printing can start.",
                     "unit": "\u00b0C",
                     "type": "float",
                     "default_value": 200,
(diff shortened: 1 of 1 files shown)


HellAholic:
I added an explanation as to why this is not a valid approach in the issue report. The modified settings don't apply only to multi extruder machines.

GregValiant:
When I looked at Ghostkeeper's "Settings Guide" plugin, both "Initial Print Temperature" and "Final Print Temperature" were described as being related to "extruder switching".  That is why I assumed (silly me) that they were only used by multi-extruder machines.

[material_initial_print_temperature](https://github.com/Ghostkeeper/SettingsGuide/blob/master/resources/articles/material/material_initial_print_temperature.md)

[material_final_print_temperature](https://github.com/Ghostkeeper/SettingsGuide/blob/master/resources/articles/material/material_final_print_temperature.md)

HellAholic:
So slightly longer version:


The principle behind these settings is to reduce oozing at the start and end of printing. Although the descriptions often refer to extruder switches in multi-extruder setups, the same idea applies to any printer.

Oozing mostly occurs when the nozzle is at its highest temperature but not actively printing. By starting or ending slightly below the normal printing temperature, the nozzle spends less time sitting idle at the full printing temperature, which reduces unwanted material leakage.

For example, if the printing temperature is 250 °C and the printer reaches that temperature while the nozzle is parked at X0 Y0, but the actual print begins at X150 Y150, the nozzle would already be at full temperature during the travel move. During that time, it may start to ooze before printing begins. Using a slightly lower Initial Printing Temperature delays reaching the full temperature until the nozzle is already printing, so any oozing happens during extrusion where it is expected.

Similarly, near the end of printing, the Final Printing Temperature allows the nozzle to begin cooling slightly before it finishes. This reduces the time the nozzle remains at full temperature while idle, which also helps minimize oozing.

In multi-extruder printers this behavior helps during extruder switches, but the underlying principle—avoiding idle time at maximum temperature—applies to all machines.

GregValiant:
It's your call.
I'll go ahead and close this (but I'm leaving the settings disabled in my install).

PR #21476, opened 2026-03-16
CURA-13036 Ensure build volume initialized before arrange

HellAholic:
Defensive programming:
Fix a potential race where BuildVolume dimensions could remain zero because globalContainerStackChanged/activeQualityChanged fire before BuildVolume is constructed. Call _onStackChanged() during initialization and re-trigger it from _onEngineCreated() (with a warning) if the volume AABB is still None so dimensions are populated and rebuild can succeed. In CuraApplication._arrangeAll, handle a None build-volume bounding box by logging a warning, requesting a rebuild, and aborting arrange to avoid comparing against an uninitialized volume.

CURA-13036


wawanbreton (changes requested):
Code looks good, but there are way too many comments in my opinion. The full explanation could be present in the commit details, but the in-code comments should remain short.

PR #21479, opened 2026-03-18
CURA-13038 Windows Installers functionality extension

HellAholic:
1. Windows EXE (NSIS) installer to add file type association.
This change will/should not override the user change values for the file type association.

image

image

image

2. Windows MSI (Wix) installer to add file type association.
It does not offer the same flexibility, can only include it as part of the features.
This change will/should not override the user change values for the file type association.

image

image

image

3. Add option to the installers to create a desktop shortcut. It's set to unchecked by default.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2257,7 +2257,7 @@
                     "type": "bool",
                     "default_value": true,
                     "value": "(infill_pattern == 'cross' or infill_pattern == 'cross_3d' or infill_multiplier % 2 == 0) and infill_wall_line_count > 0",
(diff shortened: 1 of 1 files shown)


HellAholic:
Hola 😃 
- UFP level 4 was mainly personal choice, can be set to level 4 (and removing the /o for nsis)
- `.model` was used previously for all items, keeping it for STL as it is the most common file type used (w.r.t. older versions indeed) but there are arguments to be made for alternatives.
- I did not specifically look into using loops to see if it's possible or not 🤔. So could indeed be something at least for the NSIS version

HellAholic:
Made couple of adjustments (needs to be tested, purely code update):
- Using for loops (list for file type support added in the installer python files)
- *STL, *.3mf, *.UFP, and cura://* set by default
- URL association not changed

https://github.com/Ultimaker/Cura/actions/runs/23485856008

rburema:
I'll start a new build then 😁  _edit:_ I see you started one as well.

PR #21480, opened 2026-03-19
Cura 13039 lightning tree support

wawanbreton:
This PR allows the lightning infill to be generated for support structures, either tree or normal. It also adds a few settings to more fine-tune the behavior of the support infill, and refines some existing settings so that they are visible only when appropriate.

CURA-13039
Requires https://github.com/Ultimaker/CuraEngine/pull/2310

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -2257,7 +2257,7 @@
                     "type": "bool",
                     "default_value": true,
                     "value": "(infill_pattern == 'cross' or infill_pattern == 'cross_3d' or infill_multiplier % 2 == 0) and infill_wall_line_count > 0",
-                    "enabled": "infill_pattern != 'lightning' and infill_pattern == 'cross' or infill_pattern == 'cross_3d' or infill_pattern == 'concentric' or infill_multiplier % 2 == 0 or infill_wall_line_count > 1",
+                    "enabled": "infill_pattern != 'lightning' and (infill_pattern == 'cross' or infill_pattern == 'cross_3d' or infill_pattern == 'concentric' or infill_multiplier % 2 == 0 or infill_wall_line_count > 1)",
                     "limit_to_extruder": "infill_extruder_nr",
                     "settable_per_mesh": true
                 },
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -5806,6 +5886,19 @@
                         }
                     }
                 },
+                "support_brim_minimum_hole_area":
+                {
+                    "label": "Minimum hole area in brim support",
+                    "description": "Holes in the support brim that have an area lesser this value will be filtered out, resulting in a more consistent brim. Set to 0 to have a fully generated support brim",
```
casperlamboo:
```suggestion
                    "description": "Holes in the support brim that have an area less than this value will be filtered out, resulting in a more consistent brim. Set to 0 to have a fully generated support brim",
```

resources/definitions/fdmprinter.def.json
```diff
@@ -5618,6 +5619,19 @@
                     "settable_per_mesh": false,
                     "settable_per_extruder": true
                 },
+                "support_infill_multiplier":
+                {
+                    "label": "Support Infill Line Multiplier",
+                    "description": "Convert each support infill line to this many lines. The extra lines do not cross over each other, but avoid each other. This makes the infill stiffer, but increases print time and material usage.",
```
casperlamboo, line 5625: (suggestion)
```suggestion
                    "description": "Convert each support infill line to the set amount of lines. The extra lines do not cross over each other, but avoid each other. This makes the infill stiffer, but increases print time and material usage.",
```

casperlamboo: > The extra lines do not cross over each other, but avoid each other.

I would maybe omit this part, I think I know what you mean (its generating one continuous line without intersections without self-intersections). But it's a bit vague and in my opinion offers very little information to the user.

resources/definitions/fdmprinter.def.json
```diff
@@ -5806,6 +5886,19 @@
                         }
                     }
                 },
+                "support_brim_minimum_hole_area":
```
casperlamboo, line 5889: Just for my understanding, this additional setting is not needed for the lightning infill per-se. But it's to resolve the todo that was in the engine code?

resources/definitions/fdmprinter.def.json
wawanbreton: This description is taken from the "Infill Line Multiplier", adapted to support. And I do think the comment is useful because otherwise you really can miss the benefit of this multiplier. It could probably be more concise though.

resources/definitions/fdmprinter.def.json
wawanbreton: It was not only for the todo itself, but when activating the lightning with tree support I had a few small branches that ended up being printed over air, because of this filter. When using this configuration it is quite important to set this value to 0.

resources/definitions/fdmprinter.def.json
casperlamboo: Not sure if conciseness is the issue here. I think the concept is very hard to communicate in such a short amount of text. I think it's better not to share this information then it is to share a message a user is unable to understand. But that is not the hill I'm gonna die on.

The other point still stands though, the code suggestion to change "this many lines" into "the set amount of lines". Agree this is an improvement?

wawanbreton: > The other point still stands though, the code suggestion to change "this many lines" into "the set amount of lines". Agree this is an improvement?

Yes indeed. Shall I also apply it to the other setting, at the cost of a new translation ?

casperlamboo: Nah, don't think it's worth it if it comes at the cost of changing other strings. And then I value consistency over slight improvement.

PR #21499, opened 2026-03-27
Skip extruder check for anti-overhang meshes

HellAholic:
# Description

Replace the previous isNonPrintingMesh check with isAntiOverhangMesh and update the comment. Anti-overhang meshes don't use any extruder, so the extruder-enabled check is now skipped for those nodes, preventing groups from being incorrectly skipped when only anti-overhang meshes are present.

Both infill mesh and cutting mesh can apply the values from the disabled extruder and cause the engine to pick up the disabled extruder as active tool for the overlap, but support eraser mesh (isAntiOverhangMesh) is extruder agnostic.

This was previously evaluating support eraser meshes when slicing on the front end.
image

CURA-11827

## Type of change

# How Has This Been Tested?

- [x] check if the normal flow still works: support eraser prevents support generation
- [x] check if infill/cutting/support meshes are block slicing when assigned to disabled extruder

**Test Configuration**:

# Checklist:


plugins/CuraEngineBackend/StartSliceJob.py
```diff
@@ -409,9 +409,9 @@ def run(self) -> None:
             stack = global_stack
             skip_group = False
             for node in group:
-                # Only check if the printing extruder is enabled for printing meshes
-                is_non_printing_mesh = node.callDecoration("evaluateIsNonPrintingMesh")
-                if not is_non_printing_mesh:
+                # Anti-overhang meshes don't use any extruder, so skip the extruder check for them
+                is_anti_overhang_mesh = node.callDecoration("isAntiOverhangMesh")
+                if not is_anti_overhang_mesh:
```
casperlamboo, line 414: Don't we need both conditions?
```suggestion
                is_anti_overhang_mesh = node.callDecoration("isAntiOverhangMesh")
                is_non_printing_mesh = node.callDecoration("evaluateIsNonPrintingMesh")
                if not is_anti_overhang_mesh and not is_non_printing_mesh:
```

HellAholic: `node.callDecoration("evaluateIsNonPrintingMesh")` always returns `None`. -> always block slicing

https://github.com/Ultimaker/Cura/blob/2cb536439c84ec5f4169008c5682a75d7d6db11a/cura/Settings/SettingOverrideDecorator.py#L144

The method `_evaluateIsNonPrintingMesh` works as expected but it's a private method.
We want to block the slicing for any mesh that is not anti_overhang as they will affect the results based on the extruder assigned (`infill_mesh`, `cutting_mesh`, and `support_mesh`) as they take the settings of the assigned extruder (as expected).

So don't believe that condition is needed in fixed/unfixed state. the is_anti_overhang_mesh should be sufficient, as that is the only mesh we want to skip checking if the assigned extruder is active or not.

plugins/CuraEngineBackend/plugin.json
```diff
@@ -3,6 +3,6 @@
     "author": "Ultimaker B.V.",
     "description": "Provides the link to the CuraEngine slicing backend.",
     "api": 8,
-    "version": "1.0.1",
+    "version": "1.1.0",
```
wawanbreton: This is just a bugfix, so increasing the patch is fine
```suggestion
    "version": "1.0.2",
```

plugins/CuraEngineBackend/StartSliceJob.py
wawanbreton: I think only the `isAntiOverhangMesh` check is enough at this point indeed. There may be other meshes that are non-printing but for which we still need to check the extruder.
The other approach would be to not add the `SliceableObjectDecorator` to the support blockers, but currently that would exclude them form being sent to the engine. So not doable in a small fix.

wawanbreton (approved):
Not 100% sure without running tests, but I seem to agree with the original approach.

PR #21505, opened 2026-03-30
CURA-13064 Expand the Retract Before Outer Wall setting

wawanbreton:
Change the "Retract Before Outer Wall" to be an enumeration so that we can handle more detailed behaviors.

Requires https://github.com/Ultimaker/CuraEngine/pull/2314
CURA-13064


resources/definitions/fdmprinter.def.json
```diff
@@ -4851,9 +4851,16 @@
...
-                    "default_value": false,
+                    "description": "<html>Retraction behavior when moving to start an outer wall:<ul><li><b>Automatic:</b> retraction will be calculated according to other settings</li><li><b>Retract:</b> all the travel moves to outer walls will be retracted</li><li><b>No retract:</b> all the travel moves to outer walls will not be retracted</li><li><b>No retract first:</b> the first travel move to an outer wall will not be retracted, others will use automatic</li></ul></html>",
+                    "type": "enum",
+                    "options":
+                    {
+                        "automatic": "Automatic",
+                        "force_retracted": "Retract",
+                        "force_not_retracted": "No retract",
+                        "force_not_retracted_first": "No retract first"
+                    },
```
HellAholic:
```suggestion
                    "description": "<html>Retraction behavior when moving to start an outer wall:<ul><li><b>Automatic:</b> Retraction will be calculated according to other settings</li><li><b>Always:</b> All the travel moves to outer walls will be retracted</li><li><b>Never:</b> All the travel moves to outer walls will <b>not</b> be retracted</li><li><b>First unretracted:</b> The first travel move to an outer wall will not be retracted, others will use automatic</li></ul></html>",
                    "type": "enum",
                    "options":
                    {
                        "automatic": "Automatic",
                        "force_retracted": "Always",
                        "force_not_retracted": "Never",
                        "force_not_retracted_first": "First unretracted"
                    },
```

resources/definitions/fdmprinter.def.json
```diff
@@ -4851,9 +4851,16 @@
                 "travel_retract_before_outer_wall":
                 {
                     "label": "Retract Before Outer Wall",
-                    "description": "Always retract when moving to start an outer wall.",
-                    "type": "bool",
-                    "default_value": false,
+                    "description": "<html>Retraction behavior when moving to start an outer wall:<ul><li><b>Automatic:</b> Retraction will be calculated according to other settings</li><li><b>Always:</b> All the travel moves to outer walls will be retracted</li><li><b>Never:</b> All the travel moves to outer walls will not be retracted</li><li><b>First unretracted":</b> The first travel move to an outer wall will not be retracted, others will use automatic</li></ul></html>",
```
casperlamboo:
```suggestion
                    "description": "<html>Retraction behavior when moving to the start an outer wall:<ul><li><b>Automatic:</b> Retraction will be calculated according to other settings</li><li><b>Always:</b> All the travel moves to outer walls will be retracted</li><li><b>Never:</b> All the travel moves to outer walls will not be retracted</li><li><b>First unretracted":</b> The first travel move to an outer wall will not be retracted, others will use automatic</li></ul></html>",
```

casperlamboo (approved):

resources/definitions/fdmprinter.def.json
HellAholic: `to start` as verb works as well.
can be changed to `to the start of an outer wall`

wawanbreton: This was already there before :stuck_out_tongue:

casperlamboo: oh yes, i ment "moving to the start of an outerwall". I think it reads a bit more cleanly compared to "moving to start an outerwall". It think the latter anthropomorphizing the printer a bit; it appears like the printer itself is deciding to start moving.

wawanbreton: Sure.. I don't really have a preference, so I will apply your suggestion.

PR #21507, opened 2026-03-30
CURA-13060 5.13 version upgrade

wawanbreton:
* Setting "Randomize Infill Start" has been integrated into the larger behavior "Infill Start/End Preference"
* Setting "Retract Before Outer Wall" was a boolean and is now an enum to handle more subtle behaviors

CURA-13060


HellAholic:
LGTM ✅

PR #21515, opened 2026-04-01
[CURA-13024] Spike: merge thin parts of floor/roof/skin

rburema:
So [this PR](https://github.com/Ultimaker/CuraEngine/pull/2316) makes the engine capable of merging roof/floor into skin or vice-versa, given a distance, where anything under this is considered 'too thin' and will be merged to the other areas if next to each other -- this is that distance.


resources/definitions/fdmprinter.def.json
```diff
@@ -1791,6 +1791,19 @@
...
                 },
+                "top_bottom_skin_merge_distance":
+                {
+                    "label": "Top/Bottom/Skin Merge Distance",
+                    "description": "Merge top/bottom into skin, each other, or vice-versa when areas become too thin.",
+                    "type": "float",
+                    "unit": "mm",
+                    "default_value": 0.4,
+                    "minimum_value": "0.0",
+                    "value": "line_width",
```
wawanbreton: I would make the default value much larger, like `line_width*5` or maybe a fixed value in millimeters (in case of large nozzles). Any area thinner than this will not really be eye-visible anyway, and printing it may end up uglier that when printed as part of a bigger area.

resources/definitions/fdmprinter.def.json
```diff
@@ -1791,6 +1791,19 @@
...
+                    "maximum_value": "min(machine_depth, machine_width)/2.0",
```
wawanbreton: Maybe also add a warning value to indicate that a very-large-but-authorized-value doesn't really make sense ?

wawanbreton: Also, why a full `machine_depth` but a semi-`machine_width` ? 🤔

resources/definitions/fdmprinter.def.json
```diff
@@ -1791,6 +1791,19 @@
...
+                    "settable_per_mesh": true,
+                    "settable_per_extruder": false
```
wawanbreton, line 1805: Why would it be `settable_per_mesh` but not `settable_per_extruder` ?

resources/definitions/fdmprinter.def.json
```diff
@@ -1791,6 +1791,19 @@
                     "limit_to_extruder": "top_bottom_extruder_nr",
                     "settable_per_mesh": true
                 },
+                "top_bottom_skin_merge_distance":
+                {
+                    "label": "Top/Bottom/Skin Merge Distance",
+                    "description": "Merge top/bottom into skin, each other, or vice-versa when areas become too thin.",
```
wawanbreton: Maybe the description could be more explicit, not sure it is understandable out of context.

wawanbreton (commented):
Sorry, small change but a lot of comments 😅

resources/definitions/fdmprinter.def.json
rburema: Technically, I'm already cheating by putting it above 0 ;-)

But yeah, I tried 1.2mm during my tests and that seems to work a lot better.

PPM will need to take a look at this anyway.

resources/definitions/fdmprinter.def.json
rburema: I could make a warning from the current maximum value (I think maybe I intended that), as it's not that it won't work at arbitrarily large values, just that it doesn't make much sense.

rburema: Take a closer look at where the `()` are placed ;-)

resources/definitions/fdmprinter.def.json
rburema: I thought that setting it per extruder wouldn't work, since I'm decently sure there may be situations in which the top/bottom extruder is different from the skin extruder.

resources/definitions/fdmprinter.def.json
rburema: Yes, that needs some work.

resources/definitions/fdmprinter.def.json
wawanbreton: Oh yeah that is right, since it interacts between skin, roofing and flooring and they can have different extruders, that is quite a mess :sweat_smile:

resources/definitions/fdmprinter.def.json
wawanbreton: Oh, my bad...

wawanbreton (approved):

wawanbreton (approved):

PR #21531, opened 2026-04-08
[CURA-12961] Spike to add auto-brim option; DRAFT!

rburema:
See the main work in the 'backend'/engine: https://github.com/Ultimaker/CuraEngine/pull/2320

Very W.I.P. part of a spike to investigate whether we can actually do this / basis for discussion.


resources/definitions/fdmprinter.def.json
```diff
@@ -6525,17 +6525,18 @@
                 "adhesion_type":
                 {
                     "label": "Build Plate Adhesion Type",
-                    "description": "Different options that help to improve both priming your extrusion and adhesion to the build plate. Brim adds a single layer flat area around the base of your model to prevent warping. Raft adds a thick grid with a roof below the model. Skirt is a line printed around the model, but not connected to the model.",
+                    "description": "Different options that help to improve both priming your extrusion and adhesion to the build plate. Brim adds a single layer flat area around the base of your model to prevent warping (Select Auto-Brim to detect the need for a Brim automatically; if not, Skin will be printed instead). Raft adds a thick grid with a roof below the model. Skirt is a line printed around the model, but not connected to the model.",
```
wawanbreton, line 6528: I think you meant Skirt ?
```suggestion
                    "description": "Different options that help to improve both priming your extrusion and adhesion to the build plate. Brim adds a single layer flat area around the base of your model to prevent warping (Select Auto-Brim to detect the need for a Brim automatically; if not, Skirt will be printed instead). Raft adds a thick grid with a roof below the model. Skirt is a line printed around the model, but not connected to the model.",
```

PR #21537, opened 2026-04-13
Preserve original filename when opening cura:// URLs

nzinov:
When a model is opened via a `cura://open?file=…` URL (e.g. Thingiverse's "Send to Cura" button), the downloaded file is shown in Cura with a name like `tmpAbC123model.stl` instead of `model.stl`.

The handler in `CuraApplication._openUrl` already does the work to recover the real filename — it parses the `Content-Disposition` header, and falls back to the last URL path segment. But it then passes that name as the `suffix` argument to `tempfile.NamedTemporaryFile`, which produces `tmpXXXXXX<suffix>`. So the careful filename extraction is effectively wasted.

This PR writes the download to a fresh temp directory under its real filename instead. The temp file is still disposable, but the name Cura displays (and suggests on save) is now the original model name.

## Before / after

- **Before:** `tmpAbC123thingiverse_123456_model.stl`
- **After:** `thingiverse_123456_model.stl`

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -1875,11 +1875,16 @@ class CuraApplication(QtApplication):
                             if content_disposition_match is not None:
                                 filename = content_disposition_match.group("filename").strip("\"")
 
-                        tmp = tempfile.NamedTemporaryFile(suffix=filename, delete=False)
-                        with open(tmp.name, "wb") as f:
+                        # Write the download to a fresh temp directory under
+                        # its real filename, so the file shown in Cura keeps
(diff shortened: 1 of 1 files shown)


casperlamboo:
Hi @nzinov thank you for the PR, seems like useful addition. Could you share how this feature has been tested? I do recall our team trying something like this, but running into the issue where tempfolders would be cleared before we could load them where tempfiles would not.

nzinov:
@casperlamboo I use it locally on my machine without any problems for a while now, but I have not tested it otherwise.

PR #21538, opened 2026-04-13
CURA-13108 Anchor warning text between icon and button

HellAholic:
Replace the hard-coded width calculation for the warning text with left/right anchors to tie it between the warning icon and the reset button. Add a right anchor and symmetric thin margins, and ensure Text.WordWrap is set. This makes the warning label layout responsive and avoids manual width arithmetic.

CURA-13108


wawanbreton (approved):
Much better 👍

PR #21570, opened 2026-04-28
CURA-12754 dynamic dialogs

wawanbreton:
This PR attemps to reduce the number of dialogs/objects loaded in the background that are not actually used by the application. In QML it is very easy to declare a Dialog and just display it when you need it, however this instanciates QML/Python resources that will possibly never be used. This is OK as long as it concerns a small number of elements, but when having tens of those, it is starting to consume a lot of unnecessary resources.
* Dialogs improvements: place the dialogs inside `Component` items, instanciate them only when necessary, and destroy them as soon as they are hidden.
* Models improvements: do not load all the models at startup, because they keep updating in the background even if they are never displayed. Instead, the QML file creates a local model when it is displayed, and it gets released when hidden.

Requires https://github.com/Ultimaker/Uranium/pull/1038
CURA-12754

### DIFF
--- a/cura/UI/WhatsNewSubPagesModel.py
+++ b/cura/UI/WhatsNewSubPagesModel.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2026 Ultimaker B.V.
+# Cura is released under the terms of the LGPLv3 or higher.
+
+import os
+from typing import Optional, Dict, List, Tuple
+
+from PyQt6.QtCore import QObject, pyqtProperty, pyqtSlot
+
+from UM.Logger import Logger
+from UM.Resources import Resources
+
+
+class WhatsNewSubPagesModel(QObject):
+    """
+    This model loads the sub-pages of the what's new screen according to the text/html pages available
+    in the resources folder.
+    """
+
+    image_formats = [".png", ".jpg", ".jpeg", ".gif", ".svg"]
+    text_formats = [".txt", ".htm", ".html"]
+    image_key = "image"
+    text_key = "text"
+
+    def __init__(self, parent: Optional["QObject"] = None) -> None:
+        super().__init__(parent)
+
+        images, max_image = WhatsNewSubPagesModel._collectOrdinalFiles(Resources.Images, WhatsNewSubPagesModel.image_formats)
+        texts, max_text = WhatsNewSubPagesModel._collectOrdinalFiles(Resources.Texts, WhatsNewSubPagesModel.text_formats)
+        highest = max(max_image, max_text)
+
+        self._subpages: List[Dict[str, Optional[str]]] = []
+        for n in range(0, highest + 1):
+            self._subpages.append({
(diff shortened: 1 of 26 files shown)


cura/UI/WhatsNewSubPagesModel.py
```diff
@@ -0,0 +1,87 @@
...
+            self._subpages.append({
+                WhatsNewSubPagesModel.image_key: None if n not in images else images[n],
+                WhatsNewSubPagesModel.text_key: None if n not in texts else self._loadText(texts[n])
+            })
+        if len(self._subpages) == 0:
+            self._subpages.append({WhatsNewSubPagesModel.text_key: "~ There Is Nothing New Under The Sun ~"})
+
+    @staticmethod
+    def _collectOrdinalFiles(resource_type: int, include: List[str]) -> Tuple[Dict[int, str], int]:
+        result = {}  # type: Dict[int, str]
```
casperlamboo: We can use modern type hints here
```suggestion
        result: Dict[int, str] = {}
```

plugins/3MFReader/WorkspaceDialog.qml
```diff
@@ -75,13 +77,19 @@ UM.Dialog
...
                 id: resolveStrategiesModel
+
+                signal initialized()
+
                 // Instead of directly adding the list elements, we add them afterwards.
                 // This is because it's impossible to use setting function results to be bound to listElement properties directly.
                 // See http://stackoverflow.com/questions/7659442/listelement-fields-as-properties
                 Component.onCompleted:
                 {
+                    console.log("init list")
```
casperlamboo: left over debug statement?

```suggestion
```

resources/qml/Cura.qml
```diff
@@ -608,163 +604,204 @@ UM.MainWindow
...
+            modality: Qt.WindowModal
+            fileMode: FileDialog.FileMode.OpenFiles
+            nameFilters: UM.MeshFileHandler.supportedReadFileTypes;
+            currentFolder: CuraApplication.getDefaultPath("dialog_load_path")
+            onAccepted:
+            {
+                // Because several implementations of the file dialog only update the folder
+                // when it is explicitly set.
+                var f = currentFolder;
+                currentFolder = f;
```
casperlamboo, line 624: I see that this is just a remanence of the old code, but I really don't see how this code would change anything 🤔

wawanbreton: Well, given the comment, it seems to be a hack to fix some bug on some platform, so I would not remove it. It can indeed trigger an event because the `currentFolder` is set.

plugins/3MFReader/WorkspaceDialog.qml
wawanbreton: Totally :smile:

casperlamboo (approved):

PR #21609, opened 2026-05-20
[CURA-13167] Set the transient parent for the open project dialog in 3MFReader

HellAholic:
Ensure the dialog is properly modal and opens within the application window

CURA-13167


rburema (approved):
LGTM.

PR #21630, opened 2026-05-27
typos and enhance clarity in change log

HellAholic:
CURA-13187

### DIFF
--- a/resources/texts/change_log.txt
+++ b/resources/texts/change_log.txt
@@ -27,7 +27,7 @@
 - Fixed an issue where overhang speed would be applied inconsistently when both Split Infill and Overhang Speed settings were enabled
 - Fixed an issue where the profile has changes notification would overlap with UI elements
 - Fixed an issue where certain setting configurations would cause a large amount of small line segments to be generated for the prime tower
-- Fixed an issue where clicking the "Login to Digital Factory" button twice would fail the authentication
+- Fixed an issue where clicking the "Sign in to Digital Factory" button twice would fail to log in
 
 * Printer definitions, profiles and materials:
 - Updated high-speed S-Line profiles to improve tree support strength and improved seam behavior by ending infill close to the seam and traveling to the seam unretracted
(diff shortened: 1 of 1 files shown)


rburema (approved):
A very minor point:

Counterintuitively, I wouldn't use 'the authentication' there (it's definitely not incorrect though), as the point being made is that it fails 'authentication' in general, not a specific authentication (even though that's what is happening in practice -- it's a 'how language is used in practice' quirk).

Hm actually, the sentence feels off both ways. Maybe '... would fail to authenticate [the user].' would scan better?

(Ironically, in the same sentence, there's the opposite thing as well, where 'Login' would just have been a nonsense word a century ago, but saying 'Log in', while correct, makes the sentence seem incredibly clunky.)

resources/texts/change_log.txt
```diff
@@ -24,13 +24,13 @@
 - Fixed an issue where certain 3MF files would incorrectly load with an invalid setting configuration
 - Fixed an issue where certain setting combinations would skip parts of the skin and walls sections
 - Resolved an issue where slicing would fail when both Raft build plate adhesion and Support Lightning Infill were enabled
-- Fixed an issue where overhang speed would be applied inconsistently when both Split Infill and Overhang Speed settings are enabled
+- Fixed an issue where overhang speed would be applied inconsistently when both Split Infill and Overhang Speed settings were enabled
 - Fixed an issue where the profile has changes notification would overlap with UI elements
 - Fixed an issue where certain setting configurations would cause a large amount of small line segments to be generated for the prime tower
-- Fixed an issue where clicking the "Log in to Digital Factory" button twice would fail authentication
+- Fixed an issue where clicking the "Login to Digital Factory" button twice would fail the authentication
```
HellAholic: @rburema 👇 ❓ 
```suggestion
- Fixed an issue where clicking the "Sign in to Digital Factory" button twice would fail to log in
```

HellAholic: Side note:
"login vs log in"
It was a noun vs verb for me, since here it was used as a noun for describing the button the login (noun) came to my mind.
Second update here uses the verb version so `to log in`

PR #21633, opened 2026-05-28
Revert profiles to back to "correct" pre-value error

casperlamboo:
CURA-13195


wawanbreton (approved):
Same as F4, so I assume this is OK. If you can provide the commit where the initial change was made, that would be perfect for traceability.

HellAholic (approved):

casperlamboo:
> Same as F4, so I assume this is OK. If you can provide the commit where the initial change was made, that would be perfect for traceability.

Added a comment to the jira ticket since it's in the autogenerate repo

PR #21634, opened 2026-05-29
CURA-12446 top bottom wall count v3

wawanbreton:
This PR changes a few existing settings and adds a bunch of new settings, regarding the walls counts:
* No more "Extra wall count" for skin and infill, but a full count instead. This allows having actually less walls for thoses areas
* Add specific wall counts for flooring, roofing and initial layer

CURA-12446
Requires https://github.com/Ultimaker/CuraEngine/pull/2334
Supersedes https://github.com/Ultimaker/Cura/pull/20565


wawanbreton:
Unit tests are genuinely failing, have to fix some profiles

resources/definitions/fdmprinter.def.json
```diff
@@ -1883,20 +1972,6 @@
                     "limit_to_extruder": "top_bottom_extruder_nr",
                     "settable_per_mesh": true
                 },
-                "skin_outline_count":
```
rburema, line 1886: We will need an upgrade(r) for this.

resources/definitions/fdmprinter.def.json
```diff
@@ -2320,18 +2395,6 @@
                     "limit_to_extruder": "infill_extruder_nr",
                     "settable_per_mesh": true
                 },
-                "infill_wall_line_count":
```
rburema, line 2323: We will need an upgrade(r) for this.

rburema (approved):
Do we do the upgrader part as a separate bit, or does it belong with this ticket?

I'm fine with either, but if this is meant to be in 5.14, we should at least make a ticket.

wawanbreton:
> Do we do the upgrader part as a separate bit, or does it belong with this ticket?
> 
> I'm fine with either, but if this is meant to be in 5.14, we should at least make a ticket.

Usually we do it separately, so I created the upgrade script ticket for 5.14 and linked this ticket

PR #21639, opened 2026-06-02
[CURA-13056, CURA-12580] Support painting

rburema:
Backend PR 'needed' for this PR: https://github.com/Ultimaker/Uranium/pull/1041
Engine: https://github.com/Ultimaker/CuraEngine/pull/2337


plugins/PaintTool/PaintTool.qml
```diff
@@ -12,8 +12,9 @@ Item
 {
     id: base
 
-    width: childrenRect.width
-    height: childrenRect.height
+    // NOTE: Uses the main 'kid's dimensions directly, since childrenRect is only updated on _growing_, not _shrinking_.
```
wawanbreton: Not super important, but this word feels a bit out of place 😄 
```suggestion
    // NOTE: Uses the main child's dimensions directly, since childrenRect is only updated on _growing_, not _shrinking_.
```

plugins/PaintTool/PaintTool.qml
```diff
@@ -32,6 +33,15 @@ Item
         onTriggered: UM.Controller.triggerAction("redoStackAction")
     }
 
+    property UM.SettingPropertyProvider supportEnabled: UM.SettingPropertyProvider
+    {
+        id: supportEnabled
+        containerStack: Cura.MachineManager.activeMachine
+        key: "support_enable"
+        watchedProperties: [ "value", "enabled", "description" ]
```
wawanbreton: If am I correct, only the `value` is used in this context
```suggestion
        watchedProperties: [ "value" ]
```

plugins/PaintTool/PaintView.py
```diff
@@ -368,6 +373,20 @@ def _add_mapping(actual_mapping: Dict[str, tuple[int, int]], nb_storable_values:
 
         return start_index, end_index
 
+    def _getSupportAngleShaderValue(self) -> float:
+        if self._current_paint_type != "support":
+            return 1.0
```
wawanbreton: Maybe use a larger value to disable the overhang display, to avoid rounding issues ? In case a vertical surface ends up being calculated with a 1.00001 angle

plugins/PaintTool/PaintView.py
```diff
@@ -368,6 +373,20 @@ def _add_mapping(actual_mapping: Dict[str, tuple[int, int]], nb_storable_values:
...
+        global_container_stack: GlobalStack = Application.getInstance().getGlobalContainerStack()
+        if not global_container_stack:
+            return 1.0
+        if not global_container_stack.getValue("support_enable"):
+            return 1.0
```
wawanbreton:
```suggestion
        if not global_container_stack or not global_container_stack.getValue("support_enable"):
            return 1.0
```

wawanbreton (approved):
More tiny suggestions, but looks overall good !

PR #21646, opened 2026-06-04
Add the update all button to the manage packages tab

HellAholic:
# Description

Allows the users to update all the packages that have the Update available state automatically in sequence.
It delays the Restart prompt until all the packages have been updated successfully through the update all flow.

Fixed a bug where the update button would not be visible for a plugin with a new version available until the user navigated to the manages packages and back to the plugins tab.

MarketplaceUpdate

Sort by installed: Moving the packages that are installed to the top of the list

image

## Type of change

# How Has This Been Tested?

- [x] Updating multiple plugins from plugins tab and package manager tab

**Test Configuration**:

# Checklist:

### DIFF
--- a/plugins/Marketplace/PackageList.py
+++ b/plugins/Marketplace/PackageList.py
@@ -232,7 +232,7 @@ class PackageList(ListModel):
         This acts only on the package models present in this specific list instance, which keeps
         the action scoped to the currently visible tab.
         """
-        if self._active_bulk_update_id:
+        if self._bulk_update_in_progress:
             return
 
         self._pending_bulk_update_ids = []
--- a/plugins/Marketplace/PackageList.py
+++ b/plugins/Marketplace/PackageList.py
@@ -175,12 +175,7 @@ class PackageList(ListModel):
     @pyqtProperty(bool, notify = hasUpdatablePackagesChanged)
     def hasUpdatablePackages(self) -> bool:
         """True when at least one package in this list has a pending update available."""
-        for index in range(self.rowCount()):
-            data = self.getItem(index)
-            package = data.get("package") if data else None
-            if package and package.canUpdate and not package.isMissingPackageInformation:
-                return True
-        return False
+        return self.updatablePackagesCount() >= 1
 
     @pyqtProperty(int, notify = hasUpdatablePackagesChanged)
     def updatablePackagesCount(self) -> int:
--- a/plugins/Marketplace/PackageList.py
+++ b/plugins/Marketplace/PackageList.py
@@ -52,7 +52,6 @@ class PackageList(ListModel):
         self._license_dialogs: Dict[str, QObject] = {}
 
         self._sort_by_install_status = False
-        self._pre_sort_order: List[str] = []  # Package IDs in order before install-status sort was applied.
 
         # Queue for sequential bulk updates triggered from the tab-level "Update all" button.
         self._pending_bulk_update_ids: List[str] = []
@@ -175,7 +174,7 @@ class PackageList(ListModel):
     @pyqtProperty(bool, notify = hasUpdatablePackagesChanged)
     def hasUpdatablePackages(self) -> bool:
         """True when at least one package in this list has a pending update available."""
-        return self.updatablePackagesCount() >= 1
+        return self.updatablePackagesCount >= 1
 
     @pyqtProperty(int, notify = hasUpdatablePackagesChanged)
     def updatablePackagesCount(self) -> int:
@@ -195,7 +194,6 @@ class PackageList(ListModel):
             self._sort_by_install_status = value
             self.sortByInstallStatusChanged.emit()
             if value:
-                self._saveOrder()
                 self._applyInstallStatusSort()
             else:
                 self._restoreOrder()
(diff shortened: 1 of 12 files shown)


HellAholic:
@GregValiant 😉 useful change?

GregValiant:
I don't think I would use it a lot.

I noticed a couple of times when I responded to the "updates available" button of the MarketPlace, I had to search all around to determine which particular plugins had updates available.
Automatically updating them would be nice, or just a "sort" button to bring the _"plugins-of-interest"_ to the top of the list.

It usually takes a few days after a new Cura release before some folks update their plugins (Aldo has the version blocker in his).

HellAholic:
The current list is sorted based on the last updated, so when a new plugin is published it comes automatically to the top of the list, could be adding a secondary condition for the sort to also consider installed packages during the sort...
🤔 🤔 🤔 🤔

GregValiant:
But the new plugin will have an "Install" or "Uninstall" button text rather than an "Update" button text?
If you could sort by button text then a second condition could be the date.
I don't think it would be necessary though.  Just having the updates at the top of the list should be sufficient.  
(I typed "sort" but I'm thinking it's more of a "filter".)

HellAholic:
I'm thinking it could be a checkbox `filter installed plugins/packages` or something more palatable, when checked it filters the list to only show the packages that are installed with the ones that have updates on the top 🤔 
I'll poke around a bit to see what are the other options.

plugins/Marketplace/PackageList.py
```diff
@@ -138,6 +158,136 @@ def getPackageModel(self, package_id: str) -> Optional[PackageModel]:
...
+    @pyqtProperty(int, notify = hasUpdatablePackagesChanged)
+    def updatablePackagesCount(self) -> int:
+        """Returns the number of packages in this list that have a pending update available."""
+        count = 0
+        for index in range(self.rowCount()):
+            data = self.getItem(index)
+            package = data.get("package") if data else None
+            if package and package.canUpdate and not package.isMissingPackageInformation:
+                count += 1
+        return count
```
rburema, line 188: Code duplication (well, almost), can be rewritten like so -- although you do lose the early-out then;

(Also note that the 2nd function is unchanged -- it's more clear that it's about this batch of code this way.)

```suggestion
    @pyqtProperty(bool, notify = hasUpdatablePackagesChanged)
    def hasUpdatablePackages(self) -> bool:
        """True when at least one package in this list has a pending update available."""
        return self.updatablePackagesCount() >= 1

    @pyqtProperty(int, notify = hasUpdatablePackagesChanged)
    def updatablePackagesCount(self) -> int:
        """Returns the number of packages in this list that have a pending update available."""
        count = 0
        for index in range(self.rowCount()):
            data = self.getItem(index)
            package = data.get("package") if data else None
            if package and package.canUpdate and not package.isMissingPackageInformation:
                count += 1
        return count
```

plugins/Marketplace/PackageList.py
```diff
@@ -138,6 +158,136 @@ def getPackageModel(self, package_id: str) -> Optional[PackageModel]:
...
+            if package:
+                self._pre_sort_order.append(package.packageId)
+
+    def _restoreOrder(self) -> None:
+        """Restore the item order that was saved before the install-status sort was applied."""
+        if not self._pre_sort_order:
+            return
+        order_map = {pkg_id: i for i, pkg_id in enumerate(self._pre_sort_order)}
+        self.sort(lambda model: order_map.get(model.packageId, len(self._pre_sort_order)), key = "package")
+        self._pre_sort_order = []
```
rburema: Why not just re-sort the packages like we originally sorted them in this case?

It's not like this is faster, since the 'order restorer' also calls `sort`.

plugins/Marketplace/PackageList.py
```diff
@@ -138,6 +158,136 @@ def getPackageModel(self, package_id: str) -> Optional[PackageModel]:
...
+
+    @pyqtSlot()
+    def updateAllPackages(self) -> None:
+        """Queue updates for all currently listed packages that can be updated.
+
+        This acts only on the package models present in this specific list instance, which keeps
+        the action scoped to the currently visible tab.
+        """
+        if self._active_bulk_update_id:
+            return
```
rburema: Why not use the `self._bulk_update_in_progress` variable here? (Especially since you set this to `None` _before_ attempting to start the potential next bulk update step -- that could start to fail the moment we try out 'real' multi-threading.)

rburema (changes requested):
Also, I know it's not your idea, and you're not likely to do it that way, but I'd caution against sorting by text on buttons -- not just because it sorts on a secondary effect (text on the button, which we could change at some point and then break something that we didn't see coming) instead of a ground truth (the actual status), but also because of translations.

plugins/Marketplace/PackageList.py
HellAholic: The API sorts by last_updated server-side and packages are simply returned in that order, so there's no last_updated or date property on PackageModel.

I have a different approach using an `insertion_index`. The idea being we add an index to the packages when we are reading them back from the API, and then just re-sort based on that index when we want to restore the order. This should at least be faster than the implementation above 😝
See: https://github.com/Ultimaker/Cura/pull/21646/changes/d0f218ac807d663f22133a28932f6da21c053700

Let me know if this is oke 😸

plugins/Marketplace/PackageList.py
HellAholic: Done. see: https://github.com/Ultimaker/Cura/pull/21646/commits/b4bf594193e71882a2e00111a030d530473ce752

plugins/Marketplace/PackageList.py
rburema: Yeah that's fine. It's just strange we don't have a canonical ordering on our side.

HellAholic: Marketplace is the entire series of Stranger things wrapped in one 😝 Probably there were some ideas, but if out of box it lines up with MVP, then we have the tendency of leaving working things alone.

rburema: If it ain't broke... yes it is, you just haven't noticed it yet :-p
(But that's true of nearly all software of course.)

rburema (approved):
LGTM!

PR #21664, opened 2026-06-17
[CURA-13191] support warning

rburema:
Show a warning when a model that needs support is going to be sliced.

Needs https://github.com/Ultimaker/Uranium/pull/1044

### DIFF
--- a/cura/Scene/SliceableObjectDecorator.py
+++ b/cura/Scene/SliceableObjectDecorator.py
@@ -35,7 +35,7 @@ class SliceableObjectDecorator(SceneNodeDecorator):
         self.paintTextureChanged = Signal()
 
         self._texture_change_timer: Optional[QTimer] = None
-        self._texture_bitrange_maybe_dirty: Optional[tuple[int, int]] = None
+        self._texture_bitflags_maybe_dirty: int = 0b0
 
     def setNode(self, node: "SceneNode") -> None:
         if self._node is not None:
@@ -71,13 +71,9 @@ class SliceableObjectDecorator(SceneNodeDecorator):
             self._texture_change_timer.setSingleShot(True)
             self._texture_change_timer.timeout.connect(self._onTextureChangeTimerFinished)
 
-        if self._texture_bitrange_maybe_dirty is None:
-            self._texture_bitrange_maybe_dirty = bitrange
-        else:
-            self._texture_bitrange_maybe_dirty = (
-                min(self._texture_bitrange_maybe_dirty[0], bitrange[0]),
-                max(self._texture_bitrange_maybe_dirty[1], bitrange[1])
-            )
+        for i in range(bitrange[0], bitrange[1] + 1):
+           self._texture_bitflags_maybe_dirty |= 0b1 << i
+
         self._texture_change_timer.start()
 
     def _onTextureChangeTimerFinished(self) -> None:
@@ -94,18 +90,19 @@ class SliceableObjectDecorator(SceneNodeDecorator):
 
         def bitrange_dirty(name: str) -> bool:
             named_range = self._texture_data_mapping.get(name, None)
-            return (
-                named_range is not None and self._texture_bitrange_maybe_dirty is not None and
-                named_range[0] <= self._texture_bitrange_maybe_dirty[1] and
-                self._texture_bitrange_maybe_dirty[0] <= named_range[1]
-            )
+            if named_range is None:
+                return False
+            named_bitflags = 0b0
+            for i in range(named_range[0], named_range[1] + 1):
+                named_bitflags |= 0b1 << i
+            return (named_bitflags & self._texture_bitflags_maybe_dirty) != 0b0
 
         if "extruder" in self._texture_data_mapping and bitrange_dirty("extruder"):
             self._updatePaintedExtruders(image_array)
         if "support" in self._texture_data_mapping and bitrange_dirty("support"):
             self._updatePaintedSupport(image_array)
 
-        self._texture_bitrange_maybe_dirty = SliceableObjectDecorator.EMPTY_BITRANGE
+        self._texture_bitflags_maybe_dirty = 0b0
 
         from cura.CuraApplication import CuraApplication
         CuraApplication.getInstance().globalContainerStackChanged.emit()
@@ -122,7 +119,7 @@ class SliceableObjectDecorator(SceneNodeDecorator):
     def _updatePaintedSupport(self, image_array) -> None:
         bit_range_start, bit_range_end = self._texture_data_mapping["support"]
         # We only need the 'allow' bit; 'dissallow' or 'no value' don't change wether or not support will be generated.
-        bit_mask = 0x1 << bit_range_start
+        bit_mask = 0b1 << bit_range_start
         self._painted_support_texels = numpy.any(image_array & bit_mask)
 
     def setPaintTexture(self, texture: Texture) -> None:
--- a/cura/Scene/SliceableObjectDecorator.py
+++ b/cura/Scene/SliceableObjectDecorator.py
@@ -1,14 +1,16 @@
-# Copyright (c) 2025 UltiMaker
+# Copyright (c) 2026 UltiMaker
 # Cura is released under the terms of the LGPLv3 or higher.
+
 import copy
 import json
 import numpy
 
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, Set
 
 from PyQt6.QtCore import QBuffer, QTimer
 from PyQt6.QtGui import QImage, QImageWriter
 
+from UM.Decorators import deprecated
 from UM.Scene.SceneNodeDecorator import SceneNodeDecorator
 from UM.View.GL.OpenGL import OpenGL
 from UM.View.GL.Texture import Texture
@@ -28,6 +30,7 @@ class SliceableObjectDecorator(SceneNodeDecorator):
         if application is not None:
             application.getMachineManager().extruderChanged.connect(self._updateIsAssignedToDisabledExtruder)
         self._painted_extruders: Optional[List[int]] = None
+        self._painted_support_texels: int = 0
 
         self.paintTextureChanged = Signal()
 
@@ -54,7 +57,11 @@ class SliceableObjectDecorator(SceneNodeDecorator):
     def getPaintTextureChangedSignal(self) -> Signal:
         return self.paintTextureChanged
 
+    @deprecated("Replaced by 'setPaintedCountsDirty', since that now also counts the painted 'support' texels.", since="5.14.0")
     def setPaintedExtrudersCountDirty(self) -> None:
+        self.setPaintedCountsDirty()
+
+    def setPaintedCountsDirty(self) -> None:
         if self._texture_change_timer is None:
             # Lazy initialize the timer because constructor can be called from non-Qt thread
             self._texture_change_timer = QTimer()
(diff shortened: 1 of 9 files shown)


cura/Scene/SliceableObjectDecorator.py
```diff
@@ -84,8 +100,10 @@ def _onTextureChangeTimerFinished(self) -> None:
         texel_counts = numpy.bincount((image_array & bit_mask) >> bit_range_start)
         self._painted_extruders = [extruder_nr for extruder_nr, count in enumerate(texel_counts) if count > 0]
 
-        from cura.CuraApplication import CuraApplication
-        CuraApplication.getInstance().globalContainerStackChanged.emit()
+    def _updatePaintedSupport(self, image_array) -> None:
+        bit_range_start, bit_range_end = self._texture_data_mapping["support"]
+        bit_mask = 0x1 << bit_range_start
```
wawanbreton: Although I would be very much in favor or reducing the bit mask calculation, I am not sure this new one works.
There is a similar formula in the method above, which I think should be placed in a method to be reused on both.

cura/Scene/SliceableObjectDecorator.py
```diff
@@ -84,8 +100,10 @@ def _onTextureChangeTimerFinished(self) -> None:
...
+        self._painted_support_texels = numpy.count_nonzero(image_array & bit_mask)
```
wawanbreton: Since we only check the number of texels against 0, maybe this can be optimized by only finding if there is at least one existing !=0 value here ?

plugins/ModelChecker/ModelChecker.py
```diff
@@ -92,24 +103,57 @@ def checkObjectsForShrinkage(self):
...
+        if len(warning_nodes) <= 0:
+            return False
+
+        self._caution_message.setText(catalog.i18nc(
+            "@info:status",
+            "<p>One or more 3D models may not print optimally due to the model shape and missing support:</p>"
+            "<p>{model_names}</p>"
+            "<p>Enable auto-support or paint on support for better print quality and reliability.</p>"
+            "<p><a href=\"https://support.ultimaker.com/s/article/1667417606331\">View support settings guide</a></p>"
+            ).format(model_names = ", ".join([n.getName() for n in warning_nodes])))
```
wawanbreton: Can we add an action button here, like we have for the unused extruders popup ?

plugins/ModelChecker/OverhangChecker.py
```diff
@@ -0,0 +1,116 @@
...
+from cura.Settings.GlobalStack import GlobalStack
+
+
+def getSupportAngle() -> float:
+    global_container_stack: GlobalStack = Application.getInstance().getGlobalContainerStack()
+    if not global_container_stack or not global_container_stack.getValue("support_enable"):
+        return 1.0
+    extruder_nr = int(global_container_stack.getExtruderPositionValueWithDefault("support_extruder_nr"))
+    if extruder_nr < 0 or extruder_nr >= len(global_container_stack.extruderList):
+        return 1.0
```
wawanbreton: Why return 1.0 in error cases ? It seems to be a value in radians, so 1.0 is actually valid, I would tend to return 0.0 or pi/2 instead. On the other hand, that should really not happen...

plugins/ModelChecker/OverhangChecker.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Set, Tuple
+
+import numpy
+
+from UM.Application import Application
+from UM.Scene.SceneNode import SceneNode
+from cura.Settings.GlobalStack import GlobalStack
+
+
+def getSupportAngle() -> float:
```
wawanbreton, line 14: Could you write a basic documentation to at least indicate that the returned value is in radians ?

plugins/ModelChecker/OverhangChecker.py
```diff
@@ -0,0 +1,116 @@
...
+    global_container_stack: GlobalStack = Application.getInstance().getGlobalContainerStack()
+    if not global_container_stack or not global_container_stack.getValue("support_enable"):
+        return 1.0
+    extruder_nr = int(global_container_stack.getExtruderPositionValueWithDefault("support_extruder_nr"))
+    if extruder_nr < 0 or extruder_nr >= len(global_container_stack.extruderList):
+        return 1.0
+    return (global_container_stack.extruderList[extruder_nr].getValue("support_angle") or 90.0) * math.tau / 360.0
+
+
+def getCloseToBuildplate() -> float:
```
wawanbreton: The naming is a bit confusing here, I did expect it to return the "close to buildplate" status, but it is actually a distance.
```suggestion
def getCloseToBuildplateDistance() -> float:
```

plugins/PaintTool/PaintCommand.py
```diff
@@ -48,12 +48,12 @@ def undo(self) -> None:
...
 
-        self._setPaintedExtrudersCountDirty()
+        self._setPaintedCountsDirty()
         self._texture.updateImagePart(self._bounding_rect)
 
-    def _setPaintedExtrudersCountDirty(self) -> None:
+    def _setPaintedCountsDirty(self) -> None:
         if self._sliceable_object_decorator is not None:
-            self._sliceable_object_decorator.setPaintedExtrudersCountDirty()
+            self._sliceable_object_decorator.setPaintedCountsDirty()
```
wawanbreton: Maybe we could optimize this by actually using a separate dirty state for the exruder and support counts ? If that is required at all...

wawanbreton (changes requested):
Mostly remarks, I just have strong doubts about the bit mask calculation

cura/Scene/SliceableObjectDecorator.py
rburema: I made this one more simple, since we only need the bit for 'prefer'; the other two values don't 'add' to the question 'will there be support generated?'

I don't think you can do that with the other one; you need all of them.

cura/Scene/SliceableObjectDecorator.py
rburema: I thought about that, but I didn't think numpy would have an `any` ... but it turns out it has, so I think we can!

plugins/ModelChecker/ModelChecker.py
rburema: Yes, good idea -- let me try.

plugins/ModelChecker/OverhangChecker.py
rburema: Oops! I think this happened because the related value we use in the shader does go from 0 to 1 -- but that's completely irrelevant here of course.

plugins/ModelChecker/OverhangChecker.py
rburema: Will do!

plugins/PaintTool/PaintCommand.py
rburema: I'm not sure if that'll help much in practice.

cura/Scene/SliceableObjectDecorator.py
wawanbreton: Oooooh right, then this is a single-bit bitmask, which makes it simpler indeed :smile: maybe a comment would be appropriate here then !

plugins/PaintTool/PaintCommand.py
wawanbreton: Well, a tiny bit, so feel free to ignore of course :slightly_smiling_face:

rburema: https://github.com/Ultimaker/Cura/pull/21664/commits/7af4fbd5ed06792e55bf5b8fef5abe4c4051dcee

cura/Scene/SliceableObjectDecorator.py
rburema: Comment also in: https://github.com/Ultimaker/Cura/pull/21664/commits/7af4fbd5ed06792e55bf5b8fef5abe4c4051dcee

cura/Scene/SliceableObjectDecorator.py
```diff
@@ -59,16 +60,24 @@ def getPaintTextureChangedSignal(self) -> Signal:
...
             self._texture_change_timer.setInterval(500)  # Long interval to avoid triggering during painting
             self._texture_change_timer.setSingleShot(True)
             self._texture_change_timer.timeout.connect(self._onTextureChangeTimerFinished)
 
+        if self._texture_bitrange_maybe_dirty is None:
+            self._texture_bitrange_maybe_dirty = bitrange
+        else:
+            self._texture_bitrange_maybe_dirty = (
+                min(self._texture_bitrange_maybe_dirty[0], bitrange[0]),
+                max(self._texture_bitrange_maybe_dirty[1], bitrange[1])
```
wawanbreton: Technically this is not 100% correct, the new dirty range could include bits that are not dirty:
* self._texture_bitrange_maybe_dirty = (0, 2)
* bitrange = (6, 8)

=> new self._texture_bitrange_maybe_dirty = (0, 8) so the bits (3,5) are now also considered as dirty even though not explicitely marked.

However, I am perfectly fine with keeping it as is, since the fix would probably overcomplexify the thing.

cura/Scene/SliceableObjectDecorator.py
```diff
@@ -83,11 +92,21 @@ def _onTextureChangeTimerFinished(self) -> None:
...
+                self._texture_bitrange_maybe_dirty[0] <= named_range[1]
+            )
+
+        if "extruder" in self._texture_data_mapping and bitrange_dirty("extruder"):
             self._updatePaintedExtruders(image_array)
-        if "support" in self._texture_data_mapping:
+        if "support" in self._texture_data_mapping and bitrange_dirty("support"):
             self._updatePaintedSupport(image_array)
 
+        self._texture_bitrange_maybe_dirty = SliceableObjectDecorator.EMPTY_BITRANGE
```
wawanbreton: Is that different as initializing it to `None` like in the constructor ?

wawanbreton (approved):
Looks good, just one remark but only to be adresses if you strongly feel about it.

cura/Scene/SliceableObjectDecorator.py
rburema: Oops, that was left over from an earlier attempt when I tried to avoid the 'maybe'/None, but that just overcomplicated things. So it should be `None` there instead of the nonexistant contant.

cura/Scene/SliceableObjectDecorator.py
rburema: Turning it into bitflags (like I started out doing originally) turned out to be not as complicated as I thought last week. Have a look: https://github.com/Ultimaker/Cura/pull/21664/commits/c2ac99f1f78010a404878c95222b1cdc86232afc

wawanbreton: Ah nice :slightly_smiling_face:

plugins/ModelChecker/ModelChecker.py
```diff
@@ -126,17 +96,27 @@ def _createView(self):
...
+        self._last_checker_job = None
+        self.onUpdated.emit()
+
+    def _startJob(self) -> None:
+        self._caution_message.getActions().clear()
+
+        self._last_checker_job = ModelCheckerJob(self._caution_message, self._unhandled_slice)
+        self._last_checker_job.needsRetry.connect(lambda: self.onChanged.emit())
+        self._last_checker_job.finished.connect(self._onCheckerJobFinished)
+        self._last_checker_job.run()
```
HellAholic:
```suggestion
        self._last_checker_job.start()
```

We would need to change to start  to queue the job 😝

rburema: Well, that's embarassing 😅

HellAholic (approved):
LGTM

PR #21686, opened 2026-06-27
Cura 13245

HellAholic:
Addresses:
https://github.com/Ultimaker/Cura/discussions/21681#discussioncomment-17445068

Save value for the TextArea and TextField in the machine settings before closing (destroying) the window.

CURA-13245

### DIFF
--- a/resources/qml/MachineSettings/GcodeTextArea.qml
+++ b/resources/qml/MachineSettings/GcodeTextArea.qml
@@ -77,6 +77,11 @@ Item
                 }
             }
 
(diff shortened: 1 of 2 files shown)


rburema (approved):
Nicely solved -- this is probably because we now destroy the windows (because constructing them just-in-time) right? If so, I wonder if there are any other instances where this will be a problem.

HellAholic:
I did a quick look around in the code and could not find any similar cases. Special thanks to @GregValiant for finding the issue.

PR #21688, opened 2026-06-29
[CURA-13247] Adjust the "support_brim_minimum_hole_area" enable condition

HellAholic:
The setting should be visible when there is support brim enabled in addition to either support being enabled or a support mesh being present. With the previous support_brim_enable condition in isolation, the setting would be visible when no support is generated or no support mesh is present.

CURA-13247


rburema (approved):
LGTM

PR #21689, opened 2026-06-30
[CURA-13125] Add setting dependency for cooling enabled for feature cooling settings

HellAholic:
Decision related to larger setting structure:
- Multiple settings that would set fan speed → all of these should not be set when print cooling is disabled
- If you want to disable cooling for certain features → set fan speed to 0%
- If you want fan speed unchanged from current fan speed → use setting formula

Comes with:
https://github.com/Ultimaker/CuraEngine/pull/2344

CURA-13125


rburema (approved):
LGTM

PR #21693, opened 2026-07-02
[CURA-12685] Fix group model position not respected when opening 3mf

HellAholic:
This pull request makes a small change to the `_read` method in `ThreeMFReader.py`. The change ensures that for group nodes (nodes without mesh data but with children), the combined bounding box of all children is used to calculate extents, improving accuracy when handling hierarchical 3MF files.

CURA-12685


rburema (approved):
Makes a lot of sense. What makes less sense is that we didn't have this previously :-p ... Must've gotten lost somewhere in previous refactors?

HellAholic:
Potentially it was also covered up by the drop to build plate issue we previously resolved. The original report was from before the recent refactors (5.10.x).

PR #21694, opened 2026-07-02
CURA-13226 Frontend adjustments for benchmark tooling

wawanbreton:
This PR provides two minor improvements that were necessary for the new benchmarking tools:
* Remove the specific GCodePrefix message. The header is now treated as a regular GCode part, so this message is no more required.
* Append the environment variables for the cura resources, instead of defining them. This was we can more easily combine them with the variables defined by e.g. `fdm_materials`

CURA-13226
Requires https://github.com/Ultimaker/CuraEngine/pull/2345

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -57,7 +57,7 @@ jobs:
       always() &&
       needs.check_actor.outputs.proceed == 'true' &&
       needs.find-ticket-packages.result == 'success'
-    uses: ultimaker/cura-workflows/.github/workflows/unit-test.yml@CURA-13226_slice-speed-benchmark-tooling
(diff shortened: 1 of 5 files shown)


casperlamboo (approved):

.github/workflows/unit-test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       always() &&
       needs.check_actor.outputs.proceed == 'true' &&
       needs.find-ticket-packages.result == 'success'
-    uses: ultimaker/cura-workflows/.github/workflows/unit-test.yml@main
+    uses: ultimaker/cura-workflows/.github/workflows/unit-test.yml@CURA-13226_slice-speed-benchmark-tooling
```
HellAholic:
```suggestion
    uses: ultimaker/cura-workflows/.github/workflows/unit-test.yml@main
```

PR #21695, opened 2026-07-03
Add value translation logic for the user change model

HellAholic:
Updated the value shown to the user in the changed settings dialog to line up with the labels shown to the users on the UI.

1. If it is a -1 for not overridden, it should turn into Auto.
2. If it is a -1 because the setting numerical value, such as expand distance, is that value, then it should remain unchanged.

Before:
image

After:
image

CURA-10005

### DIFF
--- a/cura/API/Interface/Settings.py
+++ b/cura/API/Interface/Settings.py
@@ -3,11 +3,14 @@
 
 from dataclasses import asdict
 
-from typing import cast, Dict, TYPE_CHECKING, Any
+from typing import cast, Dict, Optional, TYPE_CHECKING, Any
 
+from UM.i18n import i18nCatalog
 from UM.Settings.InstanceContainer import InstanceContainer
 from UM.Settings.SettingFunction import SettingFunction
 from cura.Settings.GlobalStack import GlobalStack
+
+catalog = i18nCatalog("cura")
 
 if TYPE_CHECKING:
     from cura.CuraApplication import CuraApplication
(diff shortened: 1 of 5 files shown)


cura/Machines/Models/UserChangesModel.py
```diff
@@ -35,6 +37,37 @@ def __init__(self, parent = None):
...
+            global_stack = Application.getInstance().getMachineManager().activeMachine
+            if global_stack and 0 <= int_value < len(global_stack.extruderList):
+                return global_stack.extruderList[int_value].getName()
+            return str(int_value + 1)
+
+        if setting_type == "enum":
+            options = stack.getProperty(setting_key, "options")
+            if options:
+                str_value = str(value)
+                option_label = options.get(str_value) or options.get(value)
```
casperlamboo: This should never go into the `or` case right?

cura/Machines/Models/UserChangesModel.py
```diff
@@ -35,6 +37,37 @@ def __init__(self, parent = None):
 
         self._update()
 
+    def _getDisplayValue(self, setting_key: str, value, stack) -> str:
```
casperlamboo, line 38: This function wholly makes sense, form a code quality perspective I would give it an approve. I'm just wondering if there isn't similar code else where that can be reused?

casperlamboo (approved):

cura/Machines/Models/UserChangesModel.py
HellAholic: Yes, defensive programming that would not be defending, since python xD

cura/Machines/Models/UserChangesModel.py
casperlamboo: Some places that feel similar

https://github.com/Ultimaker/Cura/blob/c1f9cf7f5c749e81ef604d1bfbb6f11316235eca/plugins/3MFWriter/SettingsExportModel.py#L121-L147

https://github.com/Ultimaker/Cura/blob/c1f9cf7f5c749e81ef604d1bfbb6f11316235eca/plugins/3MFReader/SpecificSettingsModel.py#L31-L49

Not the biggest issue. I'm not very dogmatic in my DRY principles, so if the functions are not naturally combinable consider this PR approved.

HellAholic: I'll see if I can just turn it into a generalized api to be called from all 3 places.
When loading a UCP for example we're not showing the translations, so it might be worth the effort to also fix it for CURA-11799

cura/API/Interface/Settings.py
```diff
@@ -109,6 +112,43 @@ def getSliceMetadata(self) -> Dict[str, Dict[str, Dict[str, str]]]:
 
         return settings
 
+    def getSettingDisplayValue(self, setting_key: str, value, stack, i18n_catalog: Optional[i18nCatalog] = None) -> str:
```
casperlamboo, line 115: Why is catalog an argument? Can't we use the internal one, or should plugins be able to provide their own?

casperlamboo (approved):

cura/API/Interface/Settings.py
HellAholic: The idea was to allow plugin providing their own mainly. For the plugins that extend settings for example, but might be a bit of an over-design/complication as we don't have many (any) with custom translations. But also should be usable from Uranium for example.

Mostly flexibility for future.

PR #21707, opened 2026-07-10
[CURA-11257] extend linter to guard against circular dependency chains

HellAholic:
This pull request introduces a new linter to detect circular dependencies in printer definition and profile files, improves the formula typo detection logic, and adds workflow and configuration updates to support these features.

**Circular dependency detection and formula validation improvements:**

*New linter and checks:*

- Added a new `CircularDependency` linter (`printer-linter/src/printerlinter/linters/circular_dependency.py`) to detect and report cycles in setting value dependencies across definition and profile files. This prevents infinite recursion at runtime due to circular references.

*Workflow improvements:*

- Updated the `printer-linter-pr-diagnose` GitHub Actions workflow to support running the linter on all files for a given printer (via the `printer` input), not just on PR diffs.

*Formula typo detection improvements:*

- Improved the formula typo detection logic to avoid false positives by skipping Python keywords and builtins when checking for possible setting name typos in formulas.
- Enhanced the error messages and correction suggestions for formula typos to be more specific and context-aware.

*Configuration changes:*

- Disabled the `diagnostic-definition-redundant-override` check (set to `false`) in `.printer-linter` due to current engine requirements for calculating print estimation times.

CURA-11257


.github/workflows/printer-linter-pr-diagnose.yml
```diff
@@ -28,6 +38,32 @@ jobs:
...
 
+      - name: Collect printer files specified by input (UM only, optional)
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.printer != '' }}
+        env:
+          PRINTER_INPUT: ${{ github.event.inputs.printer }}
+        run: |
+          if ! [[ "$PRINTER_INPUT" =~ ^[a-zA-Z0-9_]+$ ]]; then
+            echo "::error::Invalid printer name '${PRINTER_INPUT}'. Only letters, digits, and underscores are allowed."
+            exit 1
+          fi
```
wawanbreton, line 50: Is it actually useful since it will properly fait at the next step anyway ?
The security concern by copilot is actually quite irrelevant since it concerns only the manual execution of the workflow, which can be done only by people that have rights on the repository...

.github/workflows/printer-linter-pr-diagnose.yml
```diff
@@ -28,6 +38,32 @@ jobs:
...
+            -o -path "resources/extruders/${PRINTER_INPUT}_*.def.json" \
+            -o -path "resources/quality/${PRINTER_INPUT}/*.inst.cfg" \
+            -o -path "resources/variants/${PRINTER_INPUT}*.inst.cfg" \
+            -o -path "resources/intent/${PRINTER_INPUT}/*.inst.cfg" \
+          \) -type f | tr '\n' ' ')
+          if [ -z "${FILES// }" ]; then
+            echo "::error::No resource files found for printer '${PRINTER_INPUT}'. Verify the printer name is correct."
+            exit 1
+          fi
+          echo "PRINTER_FILES=$FILES" >> "$GITHUB_ENV"
```
wawanbreton: Prefer using inputs/outputs instead of the environment, because this one keeps growing and can be polluted. inputs/outputs are local to a step/job and explicitely referenced.
But this is mostly a personal preference, feel free to ignore.

.github/workflows/printer-linter-pr-diagnose.yml
```diff
@@ -28,6 +38,32 @@ jobs:
...
+
+      - name: Diagnose printer file(s) specified by input (UM only, optional)
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.printer != '' }}
+        run: python printer-linter/src/terminal.py --diagnose --report printer-linter-result/fixes.yml $PRINTER_FILES
+
       - name: Diagnose file(s)
         if: env.GIT_DIFF && !env.MATCHED_FILES
```
wawanbreton, line 69: So this step will run also if the previous one runs ?

printer-linter/src/printerlinter/linters/circular_dependency.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def check(self) -> Iterator[Diagnostic]:
+        if self._settings["checks"].get("diagnostic-circular-dependency", False):
+            for check in self.checkCircularDependencies():
+                yield check
+        yield
+
+    # ------------------------------------------------------------------
+    # Public check entry-point
+    # ------------------------------------------------------------------
```
wawanbreton: Those comments make the file unnecessary long IMHO, and that does not really match the syntax of the other files

printer-linter/src/printerlinter/linters/circular_dependency.py
```diff
@@ -0,0 +1,313 @@
...
+        """Build a directed dependency graph from a setting-name -> formula map."""
+        graph: Dict[str, Set[str]] = {}
+        for key, formula in value_map.items():
+            refs = self._extract_references(formula)
+            if refs:
+                graph[key] = refs
+        return graph
+
+    def _detect_cycle(self, graph: Dict[str, Set[str]]) -> Optional[List[str]]:
+        """DFS-based cycle detection.
```
wawanbreton:
```suggestion
       """Depth-First-Search-based cycle detection.
```

printer-linter/src/printerlinter/linters/circular_dependency.py
```diff
@@ -0,0 +1,313 @@
...
+
+        Returns the cycle as a list of node names where the first and last element
+        are the same (e.g. ``['a', 'b', 'a']``), or ``None`` if no cycle exists.
+        Iteration order is sorted for deterministic output.
+        """
+        visited: Set[str] = set()
+
+        def dfs(node: str, path: List[str], path_set: Set[str]) -> Optional[List[str]]:
```
wawanbreton:
```suggestion
        def depth_first_search(node: str, path: List[str], path_set: Set[str]) -> Optional[List[str]]:
```
And the call(s) should also be changed obviously

printer-linter/src/printerlinter/linters/circular_dependency.py
```diff
@@ -0,0 +1,313 @@
...
+                    self._extract_values_from_settings_tree(value["children"], result)
+
+    # ------------------------------------------------------------------
+    # Filesystem helpers
+    # ------------------------------------------------------------------
+
+    def _find_definitions_dir(self) -> Optional[Path]:
+        """Walk up the directory tree to locate ``resources/definitions``."""
+        search = self._file.parent
+        for _ in range(10):
```
wawanbreton: That looks weird 😃 could you just go up until the root is reached?

wawanbreton (commented):
Somme suggestion, but looks overall good

.github/workflows/printer-linter-pr-diagnose.yml
HellAholic: Mostly for better error handling rather than the security concern part. If the name pattern is incorrect, starting a search and then finding nothing is a bit of a waste.
- Clarifies the input is valid/invalid -> for manual workflow run this is useful because typos can happen
- Fails/stops before the search -> saves like 2 seconds 😝

.github/workflows/printer-linter-pr-diagnose.yml
HellAholic: I am on board with the inputs/outputs, but the previous structure was on github env. I'll see if I can move everything to outputs.

.github/workflows/printer-linter-pr-diagnose.yml
HellAholic: This one (and the next) is skipped for the manual trigger ones with printer definitions. See example run
https://github.com/Ultimaker/Cura/actions/runs/29097039622/job/86376066649
Skips both Diagnose file(s) and Check Deleted File(s)

printer-linter/src/printerlinter/linters/circular_dependency.py
HellAholic: 👍 
updated in ec5fd54a30258f272b96ce76cb41f5a43be01103

printer-linter/src/printerlinter/linters/circular_dependency.py
HellAholic: 👍 
Updated in fb8257a8a020865b8e8096da9c132264c7546dd7

printer-linter/src/printerlinter/linters/circular_dependency.py
HellAholic: 👍 
Updated in fb8257a8a020865b8e8096da9c132264c7546dd7

printer-linter/src/printerlinter/linters/circular_dependency.py
HellAholic: 👍
Removed in d29c646aa415244c45655b97940cd84942df544c

.github/workflows/printer-linter-pr-diagnose.yml
HellAholic: Changed to outputs in 97a517889067d829a972b7eda9dd12f823e183b0

.github/workflows/printer-linter-pr-diagnose.yml
wawanbreton: Ok, I don't understand how, but it does :smiley:

wawanbreton (approved):

PR #21711, opened 2026-07-13
[CURA-11353] Allow for multi extruder - material default config

HellAholic:
Allow multiple material and nozzle default config for multi extruder printers.

CURA-11353

### DIFF
--- a/cura/Machines/MachineNode.py
+++ b/cura/Machines/MachineNode.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2024 UltiMaker
 # Cura is released under the terms of the LGPLv3 or higher.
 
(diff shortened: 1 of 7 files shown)


rburema (approved):
I really like this approach, despite (or even because?) it's 'feeling' a bit hacky (on first blush only); and feels like something I'd come up with when I'm having a good day.

rburema:
Oh one thing I do want to mention; if you somehow have time left in the day; there's some unit-test classes that go over this functionality (TestMachineNode.py and TestVariantNode.py) that test this functionality, but currently only for the 'non-list' case.

Optional, please don't feel obliged ;-)

PR #21713, opened 2026-07-14
Fix various issues with menu bar on macos

casperlamboo:
- Disabled state was not properly applied
- Save options were not properly updated after options were added/removed

CURA-13240


casperlamboo:
I've added quite some statements akin to `object.text += " ";`, this is forcing an update that was otherwise not happening.

resources/qml/Actions.qml
```diff
@@ -73,6 +73,11 @@ Item
 
     property alias configureSettingVisibility: configureSettingVisibilityAction
 
+    property alias save: saveAction
+    property alias saveUCP: saveUCPAction
+    property alias export_: exportAction
```
wawanbreton: Why the `_` here ?

wawanbreton (approved):
it does sound a bit hacky, but from what I understood, it is the best solution, so 👍 
Also I see you did remove a lot of existing hacks, so that compensates 😃

resources/qml/Actions.qml
casperlamboo: `export` is a reserved name

wawanbreton: Ah ok :smile: then maybe `exportSomething` would be more explicit ?

fieldOfView: Perhaps document that? (in the code)

casperlamboo: https://github.com/Ultimaker/Cura/pull/21713/commits/4abcada91c2b61111ca8282582bac868d8e2a391

HellAholic (approved):

PR #21714, opened 2026-07-14
CURA-13250 Define settings for chamfered support base

wawanbreton:
This PRs adds a few settings to control the new support (tree and normal) base, which can be enabled independantly inside and/or outside the actual support area. The base makes the bottom of the support much sturdier and also improves adhesion to the buildplate, so it is now possible to use higher supports, or tree supports with less trunks.

In order to ensure retro-compatility and consistency, the `support_brim_enable` setting is now actually the "Support Inner Base" and a new `support_outer_brim_enable` is added for the "Support Outer Base".

Also to note: both the inner and outer base outside widths are capped to their corresponding brim widths. Thus when using a brim, the user should set the brim width first, then the base width. When no brim is used, the values can be set without this restriction.

CURA-13250
Requires https://github.com/Ultimaker/CuraEngine/pull/2346


resources/definitions/fdmprinter.def.json
```diff
@@ -5926,6 +5937,128 @@
...
+                            "default_value": 0,
+                            "minimum_value": "0",
+                            "maximum_value": "machine_height",
+                            "limit_to_extruder": "support_extruder_nr",
+                            "settable_per_mesh": false,
+                            "settable_per_extruder": true
+                        }
+                    }
+                },
+                "support_base_inside_width":
```
casperlamboo: Why do base the "Support Base Height" settings and "Support Base Slope" settings have parent, children but the "Support Base Width" settings don't have a parent?

casperlamboo: And what is the difference between this setting and "Enable Inner Support Base"?

wawanbreton: > Why do base the "Support Base Height" settings and "Support Base Slope" settings have parent, children but the "Support Base Width" settings don't have a parent?

For the height and the slope, it does make sense to have the same value for the inner and outer bases, so you just set the value of the parent, and it sets the value of both the children.
For the width, they will almost never have the same value, since they depend on the inner brim width and global brim width, so having a parent value doesn't make sense here.

> And what is the difference between this setting and "Enable Inner Support Base"?

In practice, you can indeed disable the inner base by settings a 0 width (or height), however for the retro-compatibility I needed to keep the "Enable Outer Support Base" anyway, so for the sake of consistency it seemed logical to also have a "Enable Inner Support Base". I do think it makes some sense as a user, that the whole feature can be enabled/disabled at once.

casperlamboo: > For the width, they will almost never have the same value, since they depend on the inner brim width and global brim width, so having a parent value doesn't make sense here.

We can still have child values that differ from it's parent or siblings, see default profile for s3 for example.
Screenshot 2026-07-17 at 13 34 58

Benefit of adding this parent child relationship is that it nicely groups the values in the settings.

wawanbreton: This example is unrelated, the `top_bottom_thickness` and `wall_thickness` do not share a parenting relationship, at best a formula that makes one depend on the other.

I do remember a discussion with Jaime about the parenting. We had an external PR that added a lot of parenting for grouping purposes. Although looking good in the UI, it was rejected because from what I understood, parenting is really only for a setting that can have multiple more-defined values of itself. For instance, `wall_extruder_nr` is a parent of `wall_0_extruder_nr` and `wall_x_extruder_nr`. The idea is that you can set them both at once by using the parent, or use different values with the children, in which case the parent becomes useless (and is actually greyed out).

So if we had a `Support Base Width` with `Support Inner Base Width` and `Support Outer Base Width` as children, what would be the `value` of the `Support Base Width` ? The innner/outer also don't have the same limits, so they are impossible to combine into a parent.

casperlamboo: > I do remember a discussion with Jaime about the parenting. We

When adding a parent<->child setting, then the parent setting should never be used by the engine and the child setting should dictate how the setting is used. They can  very well have different limits etc.

wawanbreton: Then could you make a proposal for the parent definition ? I fail to see how this can work :thinking:

casperlamboo: https://github.com/Ultimaker/Cura/pull/21714/commits/fe67663d4e8a1e201f02c16dd5584959d01d1bdc

HellAholic (approved):

PR #21744, opened 2026-08-07
Cura 13289

HellAholic:
This pull request introduces a new user preference that allows users to choose whether file open and save dialogs always start at fixed default locations or remember the last used folder. The implementation includes both backend preference handling and frontend UI elements in the Preferences dialog, giving users control over dialog path behavior and allowing them to set default load/save directories.
image
image

Requires:
https://github.com/Ultimaker/Uranium/pull/1045

CURA-13289

### DIFF
--- a/cura/CuraApplication.py
+++ b/cura/CuraApplication.py
@@ -642,6 +642,8 @@ class CuraApplication(QtApplication):
 
             preferences.addPreference("local_file/%s" % key, os.path.expanduser("~/"))
 
+        preferences.addPreference("local_file/use_fixed_dialog_paths", False)
+
         preferences.setDefault("local_file/last_used_type", "text/x-gcode")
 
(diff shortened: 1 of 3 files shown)


GregValiant:
I loaded up the three Cura files and the Uranium file.  Checking the setting does not persist.  When I close the Preferences dialog the setting reverts to "unchecked".
(The Machine Settings dialog had the same sort of thing going on.)

HellAholic:
Since the change is in the core of the application and not a plugin, they would not reflect by copying the files over. The core part changes (and the Uranium ones - except the plugins in Uranium) are included when you create a new build. So while the QML changes show (resources folder is loaded during application run), changes like the curaApplication.py would not work.

GregValiant:
Thanks for explaining.  It works in the build you sent over.

Another victory for the "Make Cura Nice" initiative.

casperlamboo (approved):

PR #21749, opened 2026-08-11
Cura 13252 smooth tree support branches

wawanbreton:
Adds a new settings to enable tree support branches smoothing

CURA-13252
Requires https://github.com/Ultimaker/CuraEngine/pull/2360


resources/definitions/fdmprinter.def.json
```diff
@@ -5436,6 +5436,20 @@
                     "settable_per_mesh": true,
                     "settable_per_extruder": true
                 },
+                "support_tree_smooth_layers":
+                {
+                    "label": "Branch Smoothing Layers",
+                    "description": "The branches of the support will be smoothed on an average on the given number of layers. Smoother branches have more inter-layer adhesion and will be be sturdier, but a very large value can bend the structure too close to the model and have it lose its consistency.",
```
rburema:
```suggestion
                    "description": "The width of the branches of the support-tree will be smoothed; the branch diameter will be an average of a rolling window specified by the number of layers. Smoother branches have more inter-layer adhesion and will be be sturdier, but a very large value can bend the structure too close to the model and have it lose its consistency.",
```

wawanbreton: Hmm, that is not correct :thinking: the width doesn't change, only the position of the circles centers. But then there is definitely something missing... I will try to reword it.

rburema: I did this review first, then went on to check the engine code; I didn't go back to correct this 😅

wawanbreton: Does the new version sound better to you ?

rburema: Yes, this reads well.

PR #21758, opened 2026-08-18
Fix Windows file reveal command

HellAholic:
Use a quoted Explorer command on Windows when revealing a file in the file manager, avoiding path parsing issues with `explorer /select`.

CURA-13295


rburema (approved):
LGTM!

PR #21760, opened 2026-08-19
CURA-13243 Make sure all dialogs appear centered

wawanbreton:
Make sure we create all the top-level dialogs using the proper method that registers their transient parent.

Requires https://github.com/Ultimaker/Uranium/pull/1046

CURA-13243


rburema (approved):
LGTM!

PR #21766, opened 2026-08-24
CURA-9839 add bridging and overhang line type color scheme

wawanbreton:
This PR allows displaying some print segments with a specific color depending on attributes set by the engine, e.g. overhanging or bridging.

It also reorganizes the print types color legend by re-using the tooltips containing detailed segment color. This way we can have more colors and not overload the legend.

Requires https://github.com/Ultimaker/CuraEngine/pull/2366
Requires https://github.com/Ultimaker/pyArcus/pull/17
Requires https://github.com/Ultimaker/Uranium/pull/1047
CURA-9839


cura/PrintSegmentAttributes.py
```diff
@@ -0,0 +1,15 @@
...
+
+class PrintSegmentAttributes(IntFlag):
+    '''
+    Print attributes are flags that can be added to some print segments to indicate that they have been processed
+    a specific way, e.g. by using overhanging or bridging settings
+    This enumeration has an equivalent in CuraEngine/include/PrintSegmentAttributes.h
+    '''
+    NoAttribute = 0
+    Overhanging = 0x1
+    Bridging = 0x2
```
rburema: I'm not sure if overhang and bridging can co-exist ... seems like something is either overhanging, or bridging. But I can see the upsides of a construction like this for future enhancements.

rburema: Also, you can do true bit notation in Python (and C++ as well these days) like `0b10` for decimal 2 (for example).

cura/LayerPolygon.py
```diff
@@ -74,7 +78,22 @@ def __init__(self, extruder: int, line_types: numpy.ndarray, data: numpy.ndarray
         # Buffering the colors shouldn't be necessary as it is not
         # re-used and can save a lot of memory usage.
         self._color_map = LayerPolygon.getColorMap()
-        self._colors: numpy.ndarray = self._color_map[self._types]
+
+        raw_colors = []
+        for idx in range(min(len(self._types), len(print_attributes))):
```
rburema, line 83: If this is fast enough in practice, then I'm OK with it, but I think there's a reason everything else here works with nearly raw numpy-arrays.

rburema (approved):
Not sure we need the flags at the moment, but could be useful in the future.

LGTM!

cura/PrintSegmentAttributes.py
wawanbreton: Currently it does happen that lines are both bridging and overhanging, even though it should not be the case :smile: but that was mostly designed like this for further tagging indeed.
I'll add he binary notation as for the C++ part

cura/LayerPolygon.py
wawanbreton: I did try very hard to do it the numpy-way, which would indeed be faster, but couldn't figure it out :grimacing: if you have a proposal, I would be very grateful...

PR #21780, opened 2026-09-01
[CURA-13312] Add timestamp to alpha builds for better segregation

HellAholic:
Add a timestamp to the internal/alpha builds. So the builds are differentiated more with Month-Day-Hour-Minute-Second. This will make sure the prompt for Import / Start fresh will always show up for the internal/alpha builds.

CURA-13312


wawanbreton (approved):
I'm a bit afraid of side-effects on other workflows that may expect this string with a specific format, but let's merge it and see what breaks 😄

PR #21781, opened 2026-09-01
CURA-13267 file menu rework

wawanbreton:
This PR simplifies the logic behind the "File" menu. Since the "Save project" and "Load project" menus are dynamic depending on the loaded plugins and their enabled states, it is quite difficult to make it work reliably on all platforms.

The new logic is the following:
* The Python model classes give a list of `QObject`-based objects to the QML views
* The `Cura.qml` listens to the models and create unique and global `Action`s that are parented to the main window. This is the most reliable way to have the actions work globally for all platforms.
* The menus create their items based on the `Action`s, instead of the model. Since this is more standard behavior, it simplified the glue-ing logic. The menus now contain only display code, not the logic.
* The actions triggers are handled globally, which avoid code duplication

CURA-13267
Requires https://github.com/Ultimaker/Uranium/pull/1048


resources/qml/Cura.qml
```diff
@@ -1014,4 +1026,90 @@ UM.MainWindow
...
+            Cura.Actions.saveWorkspaceActions.splice(index, 1);
+        }
+    }
+
+    Instantiator
+    {
+        model: CuraApplication.getExtrudersModel()
+        Action
+        {
+            text: "%1: %2 - %3".arg(model.name).arg(model.material).arg(model.variant)
```
rburema:
```suggestion
            text: "${model.name}: ${model.material} - ${model.variant}"
```

resources/qml/Actions.qml
```diff
@@ -85,6 +85,35 @@ Item
...
+    {
+        for(var extraOpenFileAction of extraOpenFileActions)
+        {
+            if(extraOpenFileAction.enabled)
+            {
+                return true;
+            }
+        }
+        return false;
+    }
```
rburema:
```suggestion
    property bool hasExtraOpenFileActions: extraOpenFileActions.some((x) => x.enabled)
```

resources/qml/Actions.qml
```diff
@@ -85,6 +85,35 @@ Item
...
+        let enabled_count = 0;
+        for(var saveWorkspaceAction of saveWorkspaceActions)
+        {
+            if(saveWorkspaceAction.enabled)
+            {
+                enabled_count++;
+            }
+        }
+        return enabled_count > 1;
+    }
```
rburema:
```suggestion
    property bool hasExtraSaveWorkspaceActions: saveWorkspaceActions.filter((x) => x.enabled).length > 1
```

resources/qml/Menus/ContextMenu.qml
```diff
@@ -38,16 +38,14 @@ Cura.Menu
 
     Instantiator
     {
-        model: CuraApplication.getExtrudersModel()
+        model: Cura.Actions.setExtruderActions
```
rburema, line 41: The naming of this confuses me a little. It may be better to show it like `extruderActionsSet` instead -- otherwise I'm thinking of something related to setting extruders.

rburema (approved):
I have some small suggestions, but otherwise I think this looks good!

resources/qml/Menus/ContextMenu.qml
wawanbreton: Oh yes indeed, now you say it, it sound very much like a setter method. I'll change it.

resources/qml/Actions.qml
wawanbreton: Much nicer 😄

resources/qml/Actions.qml
wawanbreton: JS is really not my thing 😅

resources/qml/Cura.qml
wawanbreton: FYI, syntax was not entirely correct: 06b4f3272a8a5d20b9f8c648c65f5b8462cb13a5

resources/qml/Menus/ContextMenu.qml
wawanbreton: Hmm, actually, this **is** a list of actions which purposes are to assign the print extruder to a specific model. The best name I could come with would then be `actionsSetExtruder`, but it is not consistent with the other list of actions.

wawanbreton:
@rburema Please also review the associated Uranium PR https://github.com/Ultimaker/Uranium/pull/1048
I forgot to mention it initially

PR #21782, opened 2026-09-02
[CURA-13291] sharpen wall bridging conditions

rburema:
Front-end part needed for https://github.com/Ultimaker/CuraEngine/pull/2368 -- see that PR for details.

### DIFF
--- a/resources/definitions/fdmprinter.def.json
+++ b/resources/definitions/fdmprinter.def.json
@@ -9351,6 +9351,19 @@
                     "settable_per_extruder": false,
                     "settable_per_meshgroup": false
                 },
+                "bridge_wall_max_deviation":
+                {
+                    "label": "Maximum Bridge Deviation",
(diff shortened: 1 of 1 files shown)


resources/definitions/fdmprinter.def.json
```diff
@@ -9351,6 +9351,29 @@
...
+                },
+                "bridge_wall_max_deviation":
+                {
+                    "label": "Maximum Bridge Deviation",
+                    "description": "Bridge-walls shouldn't deviate from straight lines more than this, anything that falls outside of the area of maximum allowed deviation (defined by the aligned bounding box around the anchor points) isn't considered a bridging wall. Skin lines are already straight.",
+                    "unit": "mm",
+                    "type": "float",
+                    "minimum_value": "0",
+                    "default_value": 5,
+                    "value": "line_width / 2",
```
wawanbreton, line 9372: It looks a bit weird that the `default_value` and the `value` are that much different.

rburema: What would you propose here?
I think we could set it to what the value is now, but I'm not sure.

wawanbreton: That is what I usually do: set the `default_value` so that it matches the most common `value`. In this case, would be 0.2

wawanbreton (approved):

PR #21797, opened 2026-09-10
[CURA-13259] Port part of model-checking over to libuvula for speed.

rburema:
We needed the speed-boost. In isolation this is in the ballpark of 5x -- we'd need to get the face-connectivity working too to get it to like 15x, but that isn't working yet.

( Ported to: https://github.com/Ultimaker/libUvula/pull/18 )


wawanbreton (approved):
LGTM !

PR #10232, opened 2021-08-04
Adding Official Profiles for LulzBot Machines

rhendersonLulzbot:
We are wanting to add official machine profiles for the Lulzbot TAZ 6, Mini 2, Workhorse, Pro/ProXT, and SideKick 289/747.

PR #21805, opened 2026-09-16
Cura 13336

HellAholic:
PR moved from the original one:
We are wanting to add official machine profiles for the Lulzbot TAZ 6, Mini 2, Workhorse, Pro/ProXT, and SideKick 289/747.
https://github.com/Ultimaker/Cura/pull/10232
on behalf of `rhendersonLulzbot`


resources/definitions/lulzbot_base.def.json
```diff
@@ -0,0 +1,123 @@
...
+    "inherits": "fdmprinter",
+    "metadata":
+    {
+        "visible": false,
+        "author": "FAME3D Base",
+        "manufacturer": "LulzBot",
+        "file_formats": "text/x-gcode",
+        "firmware_last_version": "2.0.6.7",
+        "first_start_actions": [ "MachineSettingsAction" ],
+        "has_machine_quality": true,
```
HellAholic: As there are no quality files this is redundant. When the quality files are added this can be turned back on to true.
```suggestion
        "has_machine_quality": false,
```

GregValiant:
Without the Quality files, the warnings can't be dismissed, and no new Profiles can be created?

image

HellAholic:
> Without the Quality files, the warnings can't be dismissed, and no new Profiles can be created?

It's the line `"has_machine_quality": true,` in the base profile.
That indicates to Cura that there are quality files specific to each machine and it should find them. If the line is set to false or possibly removed, then it would fallback to the fdm generic quality profiles and users can customize them.

The setting version also needs to be bumped up but that's a minor thing for the variants.

GregValiant:
In the base definition file, setting "has_machine_quality" to "false" makes things look much better.

resources/definitions/lulzbot_base.def.json
HellAholic:
```suggestion
        "has_machine_quality": false,
```

PR #21806, opened 2026-09-16
CURA-13339 Add a few new settings to expert category

wawanbreton:
Add some of the new settings to the advanced/expert category. Most of those we have added recently are very complex and not to be tuned by end-users unless they have a very good understanding of the engine, so I suggest adding only those, which are more user-facing and easy enough to understand.

`support_brim_enable` is not strictly speaking a new setting, but it is even more important now that we have the inner and outer bases for the support.

CURA-13339


rburema (approved):
LGTM!

PR #21813, opened 2026-09-21
Make all support settings always available

wawanbreton:
Now that we have support painting, we can have "auto-generate support" disabled, no support mesh present, and still generate support, so the support-related settings need to be accessible all the time.

An other option to handle that would be to create a `support_painting_present` on the same basis as the existing `support_meshes_present`. This way the settings would become available as soon as support painting would exist. However, I do believe this behavior is more user-friendly, and also consistent with the recommended settings view.

Note about `remove_empty_first_layers`: this setting is supposed to be available only when support is fully disabled, which cannot really happen now. So I have removed the `enabled` property, and the value is actually enabled when appropriate in the engine. I have then modified the description to indicate that. This technically breaks string freeze, but I am pretty sure nobody reads the description of this setting. At worst they will have it in English.

CURA-13343


resources/definitions/fdmprinter.def.json
```diff
@@ -4529,7 +4525,7 @@
                             "maximum_value_warning": "50",
                             "default_value": 20,
                             "value": "jerk_print",
-                            "enabled": "resolveOrValue('jerk_enabled') and (support_enable or support_meshes_present)",
+                            "enabled": "resolveOrValue('jerk_enabled')",
```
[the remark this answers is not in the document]
wawanbreton: I would let them as they are. Those are really edge-cases, and `support_enable` is set to `True` by default for these profiles anyway.
There are some similar and more relevant cases in the S8 profiles, but again I would not touch them, because with support painting you don't really know what has been painted (as you are never sure that activating support will actually make some in the end)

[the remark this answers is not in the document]
rburema: Side question: Why is that (the enabled) there (in the makerbot file) in the first place though? Since it sounds like it just does the same as the file it's inheriting from anyway.

Anyway, I agree in this case. Enabled just shows it to the user; the engine will still get the actual values of the other properties -- for makerbots especially we don't really want people to deviate from the prefab settings anyway, and the most important thing is to not risk changing too much before release.

(Maybe this is what we meant with 'review resources' in the release planning though: Check if profiles are still up to date with the changed base-files.)

resources/definitions/fdmprinter.def.json
```diff
@@ -5911,8 +5901,7 @@
                     "minimum_value": "0",
                     "maximum_value_warning": "100",
                     "default_value": 15,
-                    "value": "15 if support_enable and support_structure == 'normal' else 0 if support_enable and support_structure == 'tree' else 15",
-                    "enabled": "support_enable or support_meshes_present",
+                    "value": "15 if support_structure == 'normal' else 0 if support_structure == 'tree' else 15",
```
rburema:
```suggestion
                    "value": "0 if support_structure == 'tree' else 15",
```
The fallback was the same as 'normal', and 'tree'/'normal' are the only two options, so completely redundant.

resources/definitions/fdmprinter.def.json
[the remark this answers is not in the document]
wawanbreton: Well that is a good question, and I don't have the answer... I don't see anyone fine-tuning the jerk for the support indeed, that is an extremely advanced settings... But technically it makes some sense: if jerk and support are enabled, then you should be allowed to fine-tune the support jerk.. After that, the visibility settings should make it either visible or not.

resources/definitions/fdmprinter.def.json
```diff
@@ -6291,7 +6271,7 @@
                     "limit_to_extruder": "support_interface_extruder_nr if support_bottom_enable else support_infill_extruder_nr",
                     "minimum_value": "0",
                     "maximum_value_warning": "10.0",
-                    "enabled": "((support_enable and support_structure == 'normal') or support_meshes_present) and support_bottom_stair_step_height > 0",
+                    "enabled": "(support_structure == 'normal' or support_meshes_present) and support_bottom_stair_step_height > 0",
```
rburema, line 6274: I think this might be basically the only (if not, one of the very few) places where we'd still need 'support meshes present'. (Moreover, if trees rest on the model, I could see some version of stair-stepping being a thing there -- it's just not implemented at the moment.)

rburema (approved):
We should probably be looking into completely removing `support_meshes_present`, as now with the painting we (almost) always have to assume that's true anyway.

resources/definitions/fdmprinter.def.json
wawanbreton: yes indeed !

resources/definitions/fdmprinter.def.json
wawanbreton: There are indeed a few cases where this variable is used independently of the `support_enable`, so it makes (some) sense to keep it.

